PR #155164, opened 2025-06-04
Provide access to the cudaGraph_t underlying a CUDAGraph.

galv:
There are a few considerations here:

1. A user might want to modify the cudaGraph_t either during the stream capture or after the stream capture (but before instantiation). This draft implements modification after stream capture only, though support could be added for modification during stream capture by applying
https://github.com/pytorch/pytorch/pull/140979/files#diff-d7302d133bb5e0890fc94de9aeea4d9d442555a3b40772c9db10edb5cf36a35cR391-R404

2. Previously, the cudaGraph_t would be destroyed before the end of capture_end() unless the user had previously called enable_debug_mode(). There is no way to implement this correctly without removing this restriction, or forcing the user to always call enable_debug_mode(). However, enable_debug_mode() is a confusing API (despite being an instance method, it would modify a static global variable; thus, putting one CUDAGraph object into debug mode puts all of them into debug mode, which is not acceptable in my opinion). Therefore, I made enable_debug_mode() into a no-op. This means that the CPU memory usage will increase after this change. I think this is likely to be fine.

3. No python bindings yet. These should be easy to add. It is probably worthwhile to take some time to make sure that the returned cudaGraph_t can be converted into the cuda-python cudaGraph_t in a reasonable, hopefully type-safe, manner (but without making cuda-python a dependency of pytorch), since I imagine most users will use the pip cuda-python package to make modifications.

4. There are two foot guns:

   a. The cudaGraph_t returned by raw_cuda_graph() is not owned by the user, so it will be destroyed once the owning CUDAGraph is destroyed (or calls reset()).

   b. The following seuquence won't work as intended:

```
g = torch.cuda.CUDAGraph()
with torch.cuda.graph(g):
    foo()
g.replay()
raw_graph = g.raw_cuda_graph()
modify(raw_graph)
g.replay()
```

This won't work because the user must call instantiate() again after modifying cudaGraph_t. You could add a "safety" mechanism by traversing the cudaGraph_t to create a hash and seeing if the hash changes between calls to replay(), but this is likely way too expensive.

I think these two foot guns are probably okay given that this a bit of an experts' API.

Fixes #155106

cc @mcarilli @ezyang @eellison @penguinwu @BoyuanFeng

PR #161294, opened 2025-08-22
[CUDAGraph] Add getter for cuda graph exec

galv:
This is far simpler than #155164 since we never destroy the cudaGraphExec_t.

The request comes from TRT-LLM specifically. The motivation is that some power users would like to mutate specific kernel parameters via APIs like `cudaGraphExec*SetParams` after a cuda graph has been instantiated. For example, a common request has been to be able to change the sequence length of attention kernels, after having captured a graph for the largest possible sequence length. It turns out that the host overhead you eliminate via cuda graphs in LLM inference ends up causing an increase in computation time when you size your kernels to the maximum possible sequence length (which I believe is done in both TRT-LLM and vLLM). Attention is the most problematic kernel because its computation time is quadratic in the sequence length, rather than linear.

This can work if your attention kernel can work for arbitrary shapes (this is not the case for all attention implementations! Many of them specialize with templates), and you have a persistent kernel that allocates only as many blocks as you have SM's (so you don't have to figure out how many blocks to allocate for a specific sequence length). Using a conditional SWITCH node is a better generic approach to this problem, but that requires more infrastructure work.

Note that this requires knowledge of the exact location of the value in your kernel's parameter buffer to mutate. It won't work with arbitrary stream capture code whose kernels you don't know before hand. So I expect this code path to be rarely used.

Testing:

```
pytest -s -k raw_graph_exec test/test_cuda.py
```

cc @mcarilli @ezyang @eellison @penguinwu @BoyuanFeng


test/test_cuda.py
```diff
@@ -3639,6 +3639,26 @@ def test_cuda_graph_raw_graph_keep_graph_false(self):
...
+        not TEST_CUDA_GRAPH or not TEST_CUDA_PYTHON_BINDINGS,
+        "CUDA >= 11.0 or ROCM >= 5.3 required for graphs, cuda-bindings must be installed",
+    )
+    def test_cuda_graph_raw_graph_exec_keep_graph_false(self):
+        import cuda.bindings.runtime as cudart
+
+        graph = torch.cuda.CUDAGraph(keep_graph=False)
+        x = torch.zeros([2000], device="cuda")
+        y = torch.ones([2000], device="cuda")
+        with torch.cuda.graph(graph, capture_error_mode="relaxed"):
```
ngimel: ooc why relaxed mode here?

test/test_cuda.py
```diff
@@ -3666,6 +3686,33 @@ def test_cuda_graph_raw_graph(self):
 
         graph.replay()
 
+    @unittest.skipIf(
+        not TEST_CUDA_GRAPH or not TEST_CUDA_PYTHON_BINDINGS,
+        "CUDA >= 11.0 or ROCM >= 5.3 required for graphs, cuda-bindings must be installed",
+    )
+    def test_cuda_graph_raw_graph_exec(self):
```
ngimel: seems like this test can be parametrized on keep_graph, there's very little difference

BoyuanFeng:
Just curious.

1. Could you provide a pytorch eager example on how to use this feature with attention ops? Many users would be interested.
2. For the variable length attention use case. FWIW, users (e.g., vllm) usually capture CUDAGraphs for each batch size as multiples of 8 and pad inputs to nearest batch size. What's the advantage of the proposed feature?
3. Any latency estimation of updating this config during runtime?

test/test_cuda.py
galv: I haven't done any "warmup" in this case, and it's possible to run this test before any other test. On the chance that torch.ops.aten.add does some kind of "graph unsafe" operation (I don't remember if this is the case to be honest, but you can see I did this with the keep_graph tests), doing stream capture in "relaxed" mode prevents these unsafe operations from causing an error.

test/test_cuda.py
galv: Oh yes this is right. Will do.

test/test_cuda.py
galv: if you want me to use warmup first and then do thread local mode, that is fine with me. Just say so. I don't really like to sweat details like this.

ngimel (approved):

BoyuanFeng (approved):

eellison (approved):

eqy (approved):

ngimel:
@pytorch merge -i

PR #161299, opened 2025-08-22
[Draft][CUDA] Handling edge cases of comparing the runtime device arch with the built version of the pytorch binary

nWEIdia:
Fixes misleading warning messages when running on sm12x devices using binaries built with sm120. 
PyTorch binary built with sm120 is compatible with e.g. sm121, so no need for the warning of incompatibility. 

Also allow the 'matched_cuda_warn' message to show when e.g. the user is running a binary built with only sm90 on sm12x, so that the user would be prompted to get a build which supports e.g. sm120. 

Test Plan: download the built binaries and run on Spark; THOR; B200; RTX5090; A10G (if CI is not sufficient); 

cc @ptrblck @tinglvv @atalman @malfet @eqy


eqy (approved):

torch/cuda/__init__.py
```diff
@@ -296,8 +296,8 @@ def _check_capability():
                 matched_arches = ""
                 for arch, arch_info in CUDA_ARCHES_SUPPORTED.items():
                     if (
-                        current_arch >= arch_info["min"]
-                        and current_arch <= arch_info["max"]
+                        cur_arch_major >= arch_info["min"]
+                        and cur_arch_major <= arch_info["max"]
```
Skylion007: Arch_info["min"] <= cur_arch_major <= arch_info["max"] is a tiny bit more efficient and readable.

torch/cuda/__init__.py
```diff
@@ -279,7 +279,7 @@ def _check_capability():
                 (_extract_arch_version(arch) for arch in torch.cuda.get_arch_list()),
                 default=50,
             )
-            if current_arch < min_arch or current_arch > max_arch:
+            if cur_arch_major < min_arch or cur_arch_major > max_arch:
                 warnings.warn(
```
Skylion007, line 306: Please add an explicit stack level arg while editing this so people know where the warning message is from.

nWEIdia: explicit stack level arg ? Could you please clarify?

torch/cuda/__init__.py
nWEIdia: Thanks, will do.

torch/cuda/__init__.py
nWEIdia: Do you have an example? Thanks!

atalman (approved):
lgtm. Thank you

malfet:
@atalman , @nWEIdia can you please share an example of such warning? Because indeed it will fix it for sm12, but will suppress for older gpus, where such incompatibility existed (say sm80 and sm89 need different cubins, aren't they?)

torch/cuda/__init__.py
justinchuby: https://docs.astral.sh/ruff/rules/no-explicit-stacklevel/

malfet (changes requested):
IMO it would be good to have something like `are_arches_compatible(compiled_tuple, present_tuple) -> bool` and just call `if not any(are_cuda_arch_compatible(gpu_sm, compiled_sm) for compiled_sm in COMPILED_SMS):

PR #161292, opened 2025-08-22
Stop accessing func._schema in _python_dispatch.correct_storage_aliasing

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #161455
* #161432
* #161329
* #161328
* #161317
* #161315
* #161308
* #161304
* __->__ #161292
* #161301

func._schema is a pybind, accessing the arguments/returns is expensive, we have no reason to do it anyway, and even though #161301 makes accessing the arguments/returns less expensive, this still seems to improve performance.

PR #161301, opened 2025-08-22
Fix forced copying def_property_readonly for FunctionSchema & friends

swolchok:
* #161455
* #161432
* #161329
* #161328
* #161317
* #161315
* #161308
* #161304
* #161292
* __->__ #161301

This took me a bit to figure out and I'm pretty sure I've looked at
this code before. Pybind uses
`return_value_policy::reference_internal` for `def_property`, which
[causes the owning object to be kept alive for the lifespan of the
return
value](https://pybind11.readthedocs.io/en/stable/advanced/functions.html),
allowing the getter to safely avoid copying the property
value. However, lambdas act like they return `auto`, not
`decltype(auto)`, so our lambdas themselves were forcing copies!

Testing: observed std::vector<Argument> copying disappear in Linux
perf profile of someOpInfo._schema.arguments/returns (in
_python_dispatch.correct_storage_aliasing).


torch/csrc/jit/python/init.cpp
```diff
@@ -1959,14 +1959,25 @@ void initJITBindings(PyObject* module) {
            bool,
            bool>())
       .def_property_readonly(
-          "name", [](const FunctionSchema& self) { return self.name(); })
+          "name",
+          [](const FunctionSchema& self) -> decltype(auto) {
```
Skylion007: May want to add a comment about why the explicit rvalue is important here even if it looks like it's implicitly inferred. An overzealous refactor could easily see these as redundant and remove them. I for sure wouldn't know the difference without an explanation in the comments.

Skylion007 (approved):

torch/csrc/jit/python/init.cpp
Skylion007: You can also just explicitly specify the return value type as another arg to def can't you?

swolchok: > You can also just explicitly specify the return value type as another arg to def can't you?

I would not be surprised if an attempt to do that ended up trying to take a const reference to a temporary. Being explicit should be simple and effective.

(next ghstack update will add a comment)

swolchok:
cancelled merge because I think there's a more straightforward fix

malfet (approved):
Can we have some sort of generic linter for those?

wconstab (approved):

swolchok:
> Can we have some sort of generic linter for those?

you could probably use clang-based tooling to look for 
1) lambdas passed as argument to a pybind def* family function (to cut false positives)
2) that return a single expression
3) where the type of the expression is a reference
4) but the return type of the function is a value

PR #161304, opened 2025-08-22
Use `is`, not ==, to check exact type matches in _python_dispatch

swolchok:
* #161455
* #161432
* #161329
* #161328
* #161317
* #161315
* #161308
* __->__ #161304
* #161292
* #161301

`is` checks object identity and is more efficient. Google seems to confirm it is the correct way to do an exact type check.


Skylion007 (approved):
We have a flake8/ruff rule for this btw, but it's not enabled.

Skylion007 (approved):

torch/utils/_python_dispatch.py
```diff
@@ -406,7 +406,7 @@ def is_traceable_wrapper_subclass_type(t: type) -> TypeIs[type[TensorWithFlatten
     """Same as above, but takes a type argument instead of an instance."""
     return (
         issubclass(t, torch.Tensor)
-        and t != torch.Tensor
+        and t is not torch.Tensor
```
Skylion007, line 470: Ruff rule is E721 and can be silenced by either using isinstance or type is. We should enable it after we migrate all the instances. We are pretty consistent about not wanting to handle subclasses in a lot of inductor so mostly should be "is" overloads.

malfet (approved):

bdhirsh (approved):

PR #161309, opened 2025-08-22
[SymmMem] Allow handle->buffer to point to tensor's data_ptr

kwen2501:
* __->__ #161309

This PR is intended to guard against a case where a user passes `x[1]` to our ops.
If we use storage ptr to search in the `symm_mem` map, `x[1]` would end up with a hit, returning the same handle as `x[0]`. But `handle->buffer` points to `x[0]` address instead of `x[1]`. Since most of our ops today feed these buffer addresses into the kernel, so I think it'd be better for them to be distinctive.

Using `data_ptr` as key would create more entries in the `symm_mem` map (hopefully not by much), but I think that can be compensated by reducing the handle creation cost (later PRs).

(Only changed for NVSHMEM backend.)

### DIFF
--- a/torch/csrc/distributed/c10d/symm_mem/NCCLSymmetricMemory.cu
+++ b/torch/csrc/distributed/c10d/symm_mem/NCCLSymmetricMemory.cu
@@ -269,15 +269,20 @@ class NCCLSymmetricMemoryAllocator : public SymmetricMemoryAllocator {
   };
 
   c10::intrusive_ptr<SymmetricMemory> rendezvous(
-      void* ptr,  // data_ptr() of the tensor
+      const at::Tensor& tensor,
       const std::optional<std::string>& group_name) override {
     TORCH_CHECK(group_name.has_value(), "group_name must be provided");
+
+    // `ptr` is tensor data's starting address
+    auto ptr = tensor.data_ptr();
+    auto symm_mem_key = std::make_tuple(ptr, *group_name);
     {
-      auto it = symm_mems_.find(std::make_tuple(ptr, *group_name));
+      auto it = symm_mems_.find(symm_mem_key);
       if (it != symm_mems_.end()) {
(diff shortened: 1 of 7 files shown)


ngimel:
>This PR is intended to guard against a case where a user passes x[1] to our ops.

Why should we guard against this? We support it and there currently tests for cuda backend that exercise this functionality

ngimel:
Also having different behavior for nvshmem backend and cuda backend is confusing?

kwen2501:
> > This PR is intended to guard against a case where a user passes x[1] to our ops.
> 
> Why should we guard against this? We support it and there currently tests for cuda backend that exercise this functionality

Sorry about the wording. By "guard" I mean support that use case. 

Okay, sounds good that nvshmem_triton already uses `data_ptr`. I can modify the rest of ops to use it too.

Just so that we are on the same page, `handle->buffers_dev_[peer]` would point to the **base** address of the corresponding allocation on `peer`, and it would be kernel writer's responsibility to adjust it with offset, correct?

torch/csrc/distributed/c10d/symm_mem/NCCLSymmetricMemory.cu
```diff
@@ -269,15 +269,20 @@ class NCCLSymmetricMemoryAllocator : public SymmetricMemoryAllocator {
   };
 
   c10::intrusive_ptr<SymmetricMemory> rendezvous(
-      void* ptr,  // data_ptr() of the tensor
+      const at::Tensor& tensor,
       const std::optional<std::string>& group_name) override {
     TORCH_CHECK(group_name.has_value(), "group_name must be provided");
+
+    // `ptr` is tensor data's starting address
+    auto ptr = tensor.data_ptr();
```
Skylion007, line 277: Should be be data_ptr const_data_ptr?

PR #160709, opened 2025-08-15
Do not incorrectly chain each of the strings as iterables

ezyang:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.11.0) (oldest at bottom):
* __->__ #160709

Signed-off-by: Edward Yang <ezyang@meta.com>

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta

PR #161311, opened 2025-08-22
[DeviceMesh] Clarifying flatten use case

fduwjj:
* __->__ #161311

Since we are in the middle of big refactoring and simplying the bookkeeping for device mesh. We found an interesting bug inside DeviceMesh flatten implementation. Here is the finding:
1. In unit test, we assume users can call `dp_cp_mesh._flatten()` many times but no backend will be created (aka cached).
2. From the implementation of slicing, we actually throw exception erroring out doing the `_flatten` more than once. But there is bug which was partially fixed in https://github.com/pytorch/pytorch/pull/160709 but it does not fixed the check for the case when we call the `_flatten` twice.

What's more important question to ask is, what behavior we want for `_flatten`? Do we allow calling `_flatten` multiple times (with same mesh_name)? I think we should, why?
1. We allow slicing for the same mesh_name or name_list multiple times, and we cache the PG behinds. Although we will return a new device mesh object everytime, when we compare them they are all the same (according to __eq__).
2. We actually cached the flattened mesh today inside `root_to_flatten_mapping` and actually do the early return but that  line will never be reached if we error out before that.

Also we should allow a no-op for flatten a 1D mesh into itself's mesh_dim_name, I added a unit test for it.


torch/distributed/device_mesh.py
```diff
@@ -189,7 +189,7 @@ def create_flatten_mesh(
             self.flatten_name_to_root_dims.setdefault(root_mesh, {})
             invalid_dim_names = chain(
                 list(not_none(root_mesh.mesh_dim_names)),
-                *self.flatten_name_to_root_dims[root_mesh].keys(),
```
fduwjj, line 171: Since this PR is for discussion and illustration purpose, this change is not my final version. My proposal is to directly remove this line.

test/distributed/test_device_mesh.py
```diff
@@ -843,11 +853,12 @@ def test_flatten_mesh_3d(self):
         ]
         self.assertEqual(flatten_mesh_root_dims, (0, 1))
 
-        ref_pg_count = _world.group_count
         # Calling flatten again should not create a new pg.
-        flattened_dp_cp_mesh_2 = dp_cp_mesh._flatten()
-        self.assertEqual(flattened_dp_cp_mesh, flattened_dp_cp_mesh_2)
-        self.assertEqual(ref_pg_count, _world.group_count)
```
fduwjj: Since this PR is for discussion and illustration purpose, this change is not my final version. My proposal is to still keep this test intact.

test/distributed/test_device_mesh.py
```diff
@@ -834,6 +834,11 @@ def test_flatten_mesh_3d(self):
         # Test flatten contiguous dims
         dp_cp_mesh = mesh_3d["dp", "cp"]
         flattened_dp_cp_mesh = dp_cp_mesh._flatten()
+        with self.assertRaisesRegex(
+            RuntimeError,
+            "already exists for submesh of the DeviceMesh",
```
wconstab, line 846: i am confused. are you trying to test that we throw an error if someone flattens twice?  or are trying to enable flattening twice to work?  the PR desc suggested you want to enable flatten twice to work.

fduwjj: no this is for the test when users trying to flatten into an existing name, for example, if we have a mesh like (2,2,2) with names ("dp", "cp", "tp") then mesh["dp", "cp"]._flatten("dp") is invalid and we don't have a test case for it previously

torch/distributed/device_mesh.py
```diff
@@ -185,12 +185,15 @@ def create_flatten_mesh(
             if not mesh_dim_name:
                 mesh_dim_name = "_".join(not_none(device_mesh.mesh_dim_names))
 
+            # Flatten a 1D device mesh into its original mesh_dim_name will return itself.
+            if device_mesh.ndim == 1 and mesh_dim_name in not_none(
+                device_mesh.mesh_dim_names
+            ):
+                return device_mesh
+
```
fegin, line 172: hmm, what if users flatten 1D but with a different name? What's the behavior we should expect?

fduwjj: Then we throw exception that is an invalid use right? That's why we have the second check after and.

wconstab: Thinking about the case where users flattened the (2,4) mesh and somewhere else flattened the (4,2) mesh- in their mind there might not be a clear relationship between the 2 (8,) meshes they created. Perhaps they want to use different names for them, for the logical use case for those meshes. However we should still allow the flatten and reuse the PG if possible. Maybe this is not a common case, but I also don't see the reason we need to error for flattening twice with different names. That's my thinking anyway.

fduwjj: @wconstab  I see, that is outside the scope of this PR. For this one I just want to fix the current behavior for now.

fegin (approved):

Issue #159779, opened 2025-08-04
Enable CUDA 13.0 binaries


tinglvv:
### 🚀 The feature, motivation and pitch

CUDA 13.0 is released this morning, creating issue tracker for CUDA 13.0 binaries enablement.

malfet:
It would be nice to have a short description why updating to CUDA-13 is helpful. I.e. it might be that it improves perf, adds new features, new HW capabilities etc

tinglvv:
> > Supporting new HW capabilities: sm_103 (B300/GB300), sm_110 (Jetson Thor), sm_120 (Geforce 5000 series), sm_121(DGX SPARK). Notes: sm_121 will be compatible with sm_120 binaries, sm_10x will be compatible with sm_100 binaries.
> 
> [@tinglvv](https://github.com/tinglvv) Aren't sm_120 already supported by CUDA-12.8 and rest is already supported by CUDA-12.9, isn't it? (Or can you clarify what HW were added as part of CUDA-12.9?)

Thanks for catching it, yes sm_120 is already supported in CUDA 12.8. sm_121 is the newly supported arch. Modified the description.

nWEIdia:
https://developer.nvidia.com/blog/whats-new-and-important-in-cuda-toolkit-13-0/#wheel_package_changes_to_cuda_130 
it is also time to evaluate this.

johnnynunez:
cudnn 9.12 is out @atalman @malfet

eqy:
> > Hi! I managed to get it to build 2.9.0a0+gitc7609e1 with 13.0.48 Cuda13 over the weekend, took at least a dozen times compiling to work out all the errors, but the changes needed are actually pretty overall just really nitpicky. I haven't contributed before but I could post a PR or share a diff if it would help! Tested to confirm it works with some of the new CUDA13 features specifically just fine and is super fast on 5090-blackwell!
> 
> [@michaelw9999](https://github.com/michaelw9999) Congrats, do you have detailed notes? I'm having performance issues on 5090, `backward` is much slower than 4090, seems 5090 is not fully supported by 12.8/12.9, and cuda 13 explicitly mentioned supporting RTX 50 series. FYI, my test script and terminal output:
> 
> import torch, torch.nn as nn, torch.nn.functional as F
> import numpy as np
> import time
> 
> batch_size, seqlen, i_dim, hidden  = 2048,128,38,32
> if __name__ == '__main__':
>     print(torch.cuda.get_device_name())
[...]

PR #157791, opened 2025-07-08
[BE]: Reduce binary size 40% using aggressive fatbin compression.

Skylion007:
NVCC apparently has a [compression-mode flag](https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/#compress-mode-default-size-speed-balance-none-compress-mode) to tell it how you want to compress the fatbinary since 12.4. This mode defaults to speed (pick a low compression mode that loads the file quickly). Since we are running into PyPi size issues, this will allow us to upload smaller wheel files.

From: https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#compress-mode-default-size-speed-balance-none-compress-mode
```
size
Uses a compression mode more focused on reduced binary size, at the cost of compression and decompression time.
```

Up to 37.2%  reduction in binary size with virtually no drawback (except potentially a little slower loading of the .so at PyTorch startup).

694 MB for CUDA 12.9 builds with 6.0;7.0;7.5;8.0;8.6;9.0;10.0;12.0+PTX
vs
1.08GB for CUDA 12.9 builds with 7.5;8.0;8.6;9.0;10.0;12.0+PTX

CUDA 12.9 ***694MB*** vs ***1.08GB***

CUDA 12.8 ***604MB*** vs ***845MB***

This ends up saving PyPi.org approximately 19.6 PiB of bandwidth per month for the CUDA 12.9 case. 

This will also allow us to add back CUDA 12.8 12.0+PTX which will make the package forward compatible on newer GPUs. Undoing the need for PR https://github.com/pytorch/pytorch/pull/157516 and https://github.com/pytorch/pytorch/pull/157634 

Screenshot 2025-07-08 at 5 36 44 PM

More details can be found in Nvidia's technical blog for CUDA 12.4: https://developer.nvidia.com/blog/runtime-fatbin-creation-using-the-nvidia-cuda-toolkit-12-4-compiler/

Note: This PR has been merged as https://github.com/pytorch/pytorch/pull/161316

cc @ptrblck @msaroufim @eqy @jerryzh168

PR #161316, opened 2025-08-22
Use -compress-mode=size for CUDA 13 build for binary size reduction

tinglvv:
https://github.com/pytorch/pytorch/issues/159779

CUDA 13 added the support for --compress-mode flag for nvcc across all drivers of CUDA 13.X toolkits, enabling the possibility to use --compress-mode=size for significant size reduction (~71% less for CUDA Math APIs for example). https://developer.nvidia.com/blog/whats-new-and-important-in-cuda-toolkit-13-0/

Why we have to add for CUDA 13 only, quote from @ptrblck : Any usage of --compress-mode=size/balance will drop the support of older CUDA drivers and will bump the min. driver requirement to CUDA 12.4. https://github.com/pytorch/pytorch/pull/157791#issuecomment-3058027353

Default for CUDA 13 will be --compress-mode=balance which gives smaller binaries than LZ4 speed mode used in previous CUDA versions.

Related - https://github.com/pytorch/pytorch/pull/157791 

cc @ptrblck @nWEIdia @atalman @malfet

### DIFF
--- a/.ci/manywheel/build_cuda.sh
+++ b/.ci/manywheel/build_cuda.sh
@@ -120,9 +120,8 @@ if [[ $CUDA_VERSION == 12* || $CUDA_VERSION == 13* ]]; then
     TORCH_NVCC_FLAGS="-Xfatbin -compress-all --threads 2"
     # Compress the fatbin with -compress-mode=size for CUDA 13
     if [[ $CUDA_VERSION == 13* ]]; then
-        TORCH_NVCC_FLAGS+=" -compress-mode=size"
+        export TORCH_NVCC_FLAGS="$TORCH_NVCC_FLAGS -compress-mode=size"
     fi
-    export TORCH_NVCC_FLAGS
     if [[ -z "$PYTORCH_EXTRA_INSTALL_REQUIREMENTS" ]]; then
         echo "Bundling with cudnn and cublas."
         DEPS_LIST+=(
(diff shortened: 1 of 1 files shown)


nWEIdia (approved):

.ci/manywheel/build_cuda.sh
```diff
@@ -117,7 +117,11 @@ DEPS_SONAME=(
 if [[ $CUDA_VERSION == 12* || $CUDA_VERSION == 13* ]]; then
     export USE_STATIC_CUDNN=0
     # Try parallelizing nvcc as well
-    export TORCH_NVCC_FLAGS="-Xfatbin -compress-all --threads 2"
+    TORCH_NVCC_FLAGS="-Xfatbin -compress-all --threads 2"
+    # Compress the fatbin with -compress-mode=size for CUDA 13
+    if [[ $CUDA_VERSION == 13* ]]; then
+        export TORCH_NVCC_FLAGS="$TORCH_NVCC_FLAGS -compress-mode=size"
```
Skylion007, line 123: Should we use compress-mode size or balance is my question? I guess size even if it makes the CI a bit slower.

Skylion007 (approved):

.ci/manywheel/build_cuda.sh
tinglvv: "balance" is the default mode for CUDA 13, so let's use the "size" option. if CI is too slow we can adjust.

tinglvv:
Wheel size saving comparson:
for [manywheel-py3_9-cuda13_0] for example
after: 579 MB in https://github.com/pytorch/pytorch/actions/runs/17167063966/job/48709750243
before: 725 MB in https://github.com/pytorch/pytorch/actions/runs/17172898848

146 MB saved — ~20.1% smaller

PR #161320, opened 2025-08-22
Add stable parallel_for

mikaylagawarecki:
The current state of the world is that there are two implementations of torch's parallel interface, OpenMP and ParallelNative. `INTRA_OP_PARALLEL` (which is used to is gate whether the "parallel logic" in `parallel_for` is used, see below) is defined if
1. `AT_PARALLEL_OPENMP = 1` (at libtorch build time per the generated ATen/Config.h) + `_OPENMP` is defined at extension build time (meaning that **both libtorch and extension compile/link against OPENMP**)
2. `AT_PARALLEL_OPENMP=0 && AT_PARALLEL_NATIVE = 1` (at libtorch build time per the generated ATen/Config.h)

https://github.com/pytorch/pytorch/blob/e0cb1848d0fd9fb4467ad8b844c565aea5071838/aten/src/ATen/Parallel-inl.h#L9-L43

The approach taken in this PR is to paste the implementation of `parallel_for` from `ATen/Parallel-inl.h `into `torch/csrc/stable/ops.h` with the following modifications:

For perf, we want the function passed to `parallel_for` to be inlined all the way into `invoke_parallel`
- This is possible for the OpenMP implementation which templates F just like how `parallel_for` does
https://github.com/pytorch/pytorch/blob/001e1d263746ae9d121d9c8cf55bc87f777d9dba/aten/src/ATen/ParallelOpenMP.h#L14-L53
We paste the implementation of `invoke_parallel` into `torch::stable::internal` **with a modification that `ThreadIdGuard` (the only non-headeronly piece) uses the shimmed version. We do not move it to torch/headeronly due to the ThreadIdGuard limitation**

When I compile the call I added in `kernel.cpp`,
```cpp
torch::stable::parallel_for(
      0, size, grain_size, [data_ptr](int64_t begin, int64_t end) {
        for (int64_t i = begin; i < end; i++) {
          int thread_id = aoti_torch_get_thread_num();
          data_ptr[i] = i | (static_cast<int64_t>(thread_id) << 32);
        }
      });
```
I can find the following in the objdump, which I think indicates that the function is getting inlined correctly into `invoke_parallel` for this path

Screenshot 2025-10-08 at 7 03 49 PM

- This is not possible for the ParallelNative implementation 
    - takes in an `std::function` for `f`
    - Is defined in a cpp (and relies on other non-headeronly functions)
For the above two reasons, we shim the ParallelNative version of `invoke_parallel`
https://github.com/pytorch/pytorch/blob/71aefd5595834dd97f38aa978ee32abbd13ac3d6/aten/src/ATen/ParallelNative.cpp#L144-L199

The rest of the APIs are shimmed
- `at::internal::lazy_init_num_threads()` --> `aoti_torch_lazy_init_num_threads`
    Reason for shimming: The [implementation of `at::internal::lazy_init_num_threads()`](https://github.com/pytorch/pytorch/blob/e0cb1848d0fd9fb4467ad8b844c565aea5071838/aten/src/ATen/Parallel.h#L32-L38) calls `at::init_num_threads` which is not header-only
- `at::in_parallel_region` --> `aoti_torch_in_parallel_region`
    Reason for shimming:  The [OpenMP implementation](https://github.com/pytorch/pytorch/blob/e0cb1848d0fd9fb4467ad8b844c565aea5071838/aten/src/ATen/ParallelOpenMP.cpp#L94-L100) is defined in a .cpp and depends on whether OPENMP is linked against at libtorch build time. The [ParallelNative implementation](https://github.com/pytorch/pytorch/blob/e0cb1848d0fd9fb4467ad8b844c565aea5071838/aten/src/ATen/ParallelNative.cpp#L266-L276) is defined in a .cpp and depends on whether `c10_MOBILE` is defined at libtorch build time
- `at::get_num_threads` --> `aoti_torch_get_num_threads`
    Reason for shimming: Similar story to `in_parallel_region`, [`ParallelNative` impl](https://github.com/pytorch/pytorch/blob/e0cb1848d0fd9fb4467ad8b844c565aea5071838/aten/src/ATen/ParallelNative.cpp#L241-L260), [`OpenMP` impl](https://github.com/pytorch/pytorch/blob/e0cb1848d0fd9fb4467ad8b844c565aea5071838/aten/src/ATen/ParallelOpenMP.cpp#L75-L82)
- `ThreadIdGuard` --> `aoti_torch_create_thread_id_guard` and `aoti_torch_delete_thread_id_guard`, with a C++ wrapper `torch::stable::ThreadIdGuard`
   Reason for shimming: Depends on `set_thread_num` which is not header-only [ThreadIdGuard impl](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/Parallel.h?fbclid=IwY2xjawNTpHBleHRuA2FlbQIxMQBicmlkETBhRjNHRG5BZEZIcjRTdHVzAR6LlnxdN6zn2HJlVDeoUyYJBHLZKidAmH_wiEJ7CbBE5bF56_4-WaltmlBOEw_aem_iArU_QX6AZQZeBizz6EEJQ#L42-L50)
- `c10::ParallelGuard` --> `aoti_torch_create_parallel_guard`, `aoti_torch_delete_parallel_guard`, `aoti_torch_parallel_guard_is_enabled` with a C++ wrapper `torch::stable::ParallelGuard`
   Reason for shimming: Has a cpp file [[ParallelGuard.cpp](https://github.com/pytorch/pytorch/blob/main/c10/util/ParallelGuard.cpp?fbclid=IwY2xjawNTpLxleHRuA2FlbQIxMQBicmlkETBhRjNHRG5BZEZIcjRTdHVzAR7k_w2Ob695Dy7w_WYPK9vsiMEycutaGMeNkwsp_m0x8Y2FbyWMA1QVtvsH7Q_aem_07IlbboUfzDpqB07wU27IA)]

* __->__ #161320
* #165698


torch/headeronly/util/ParallelGuard.h
```diff
@@ -0,0 +1,33 @@
...
+    thread_local static bool in_at_parallel = false;
+    return in_at_parallel;
+  }
+
+ public:
+  static inline bool is_enabled() {
+    return get_in_at_parallel();
+  }
+
+  inline ParallelGuard(bool state) : previous_state_(is_enabled()) {
```
Skylion007: This can ctor can be remove an invocation of the reference getter by using std::exchange

Skylion007: Previous_state_(std::exchange(get_in_at_parallel(), state)) {}

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
```diff
@@ -574,3 +574,52 @@ STABLE_TORCH_LIBRARY_IMPL(libtorch_agnostic, CompositeExplicitAutograd, m) {
...
+
+  torch::stable::zero_(tensor);
+
+  // Use parallel_for to fill each element with its index
+  // If using a parallel path, the thread id is encoded in the upper 32 bits
+  torch::stable::parallel_for(
+      0, size, grain_size, [data_ptr](int64_t begin, int64_t end) {
+        for (int64_t i = begin; i < end; i++) {
+          int thread_id = aoti_torch_get_thread_num();
+          data_ptr[i] = i | (static_cast<int64_t>(thread_id) << 32);
```
swolchok, line 605: should we check somewhere that i <= UINT32_MAX?

torch/csrc/stable/ops.h
```diff
@@ -235,4 +244,147 @@ inline torch::stable::Tensor clone(const torch::stable::Tensor& self) {
...
+
+// We expect this to be the stable version of the OpenMP variant of the
+// invoke_parallel op with identical semantics to the existing invoke_parallel.
+// This is copy pasted from aten/src/ATen/ParallelOpenMP.h except that we
+// replace ThreadIdGuard with the shim-ed version.
+// Requiring the extension to have _OPENMP defined to use invoke_parallel
+// matches the existing semantic.
+#ifdef _OPENMP
+template <typename F>
+inline void invoke_parallel(
```
swolchok, line 261: OK, so if I understand correctly, this implementation isn't actually a stable ABI to anything, it's just an OpenMP-based invoke_parallel implementation that will get built into clients' binaries. That's fine, so long as there are no concerns about clients using gcc/clang when pytorch uses the other one (they have different OpenMP implementations that they use by default), and the overall process getting linked against two different OpenMP support libraries, and that somehow causing problems. Those problems, to the extent they are real (which I don't know), already exist for customers that want to use OpenMP in their extension anyway; however here we are sort of advertising/pushing OpenMP to them and so we should make sure that that's actually a reasonable thing to do.

torch/csrc/stable/ops.h
```diff
@@ -235,4 +244,147 @@ inline torch::stable::Tensor clone(const torch::stable::Tensor& self) {
...
+      false,
+      "Attempting to call torch::stable::invoke_parallel "
+      "without _OPENMP. Internal error, should not have gotten here");
+}
+#endif // _OPENMP
+
+// For the ParallelNative path, this helps with converting C++ lambdas
+// etc. to a C-style function pointer expected by the C-shim
+template <typename F>
+struct Trampoline {
```
swolchok, line 314: Sure, this is basically a minimal [llvm::function_ref](https://github.com/llvm/llvm-project/blob/main/llvm/include/llvm/ADT/STLFunctionalExtras.h#L2) so I'm fine with it. Note that they moved to intptr_t from `void*` 11 years ago because some compilers warn on these casts. https://github.com/llvm/llvm-project/commit/36e1295a2b9ed828a3e19fc2d1229c51a8a08eb7

torch/csrc/inductor/aoti_torch/c/shim.h
```diff
@@ -422,6 +428,44 @@ aoti_torch_cpu__wrapped_quantized_linear_prepacked(
...
+aoti_torch_create_parallel_guard(bool state, ParallelGuardHandle* ret_guard);
+
+AOTI_TORCH_EXPORT AOTITorchError
+aoti_torch_delete_parallel_guard(ParallelGuardHandle guard);
+
+AOTI_TORCH_EXPORT bool aoti_torch_parallel_guard_is_enabled();
+
+// invoke_parallel function, only intended for use for the
+// AT_PARALLEL_NATIVE path, where the function is not inlined.
+typedef void (*AOTIParallelLambda)(int64_t begin, int64_t end, void* ctx);
```
swolchok: "lambda" is a misnomer here; this is a real actual function pointer. I would just call it AOTIParallelFunc or AOTIParallelFunction.

torch/csrc/stable/ops.h
mikaylagawarecki: Yes I think your understanding is right, the only reason I call it "stable" is because it uses a shimmed-ThreadIdGuard (though perhaps that is a misnomer)

> Those problems, to the extent they are real (which I don't know), already exist for customers that want to use OpenMP in their extension anyway

Yes exactly! 😄 

> however here we are sort of advertising/pushing OpenMP to them and so we should make sure that that's actually a reasonable thing to do.

Makes sense, my intent mostly for "extension writers to be able to use `parallel_for` in the same way the used to". But I see what you mean that having this here might incentivize them to try to use OpenMP. 

I think this is the first op we've added in torch/csrc/stable that the compiler toolchain issues apply to (cc @janeyx99 is that right?). Would it be reasonable if I:
- add more explicit documentation here + in our user facing docs re the potential issues users might run into if they use different compilers 
- (if helpful) put all the openmp related code in some "more private" looking file, rather than exposing it directly in ops.h

torch/csrc/inductor/aoti_torch/c/shim.h
```diff
@@ -172,6 +172,12 @@ AOTI_TORCH_EXPORT AOTITorchError aoti_torch_scalar_to_tensor_complex128(
 AOTI_TORCH_EXPORT bool aoti_torch_grad_mode_is_enabled();
 AOTI_TORCH_EXPORT void aoti_torch_grad_mode_set_enabled(bool enabled);
 
+// Check if INTRA_OP_PARALLEL is defined
+AOTI_TORCH_EXPORT bool aoti_torch_get_intra_op_parallel_enabled();
+
+// Value of AT_PARALLEL_OPENMP
+AOTI_TORCH_EXPORT bool aoti_torch_get_parallel_openmp_enabled();
```
malfet: Should we try to mimic naming patterns of existing AOTI APIs (See pattern on line 172) as well as `torch.backends.openmp.is_available()`?  (`get_` prefix implies there must be a matching `set_` API, but OpenMP is either enabled or not, isn't it?) 
```suggestion
AOTI_TORCH_EXPORT bool aoti_torch_openmp_is_available();
```

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
```diff
@@ -574,3 +574,52 @@ STABLE_TORCH_LIBRARY_IMPL(libtorch_agnostic, CompositeExplicitAutograd, m) {
...
+      1,
+      sizes,
+      strides,
+      aoti_torch_dtype_int64(),
+      aoti_torch_device_type_cpu(),
+      0,
+      &tensor_handle);
+
+  Tensor tensor(tensor_handle);
+  int64_t* data_ptr = reinterpret_cast<int64_t*>(tensor.data_ptr());
```
malfet, line 594: Nit
```suggestion
  auto* data_ptr = reinterpret_cast<int64_t*>(tensor.data_ptr());
```

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
```diff
@@ -574,3 +574,52 @@ STABLE_TORCH_LIBRARY_IMPL(libtorch_agnostic, CompositeExplicitAutograd, m) {
...
+
+  torch::stable::zero_(tensor);
+
+  // Use parallel_for to fill each element with its index
+  // If using a parallel path, the thread id is encoded in the upper 32 bits
+  torch::stable::parallel_for(
+      0, size, grain_size, [data_ptr](int64_t begin, int64_t end) {
+        for (int64_t i = begin; i < end; i++) {
```
malfet: Why not use `c10::irange` there? (or at the very least use auto for `i`)
```suggestion
        for (const auto i : c10::irange(begin, end)) {
```

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
```diff
@@ -574,3 +574,52 @@ STABLE_TORCH_LIBRARY_IMPL(libtorch_agnostic, CompositeExplicitAutograd, m) {
...
+      });
+
+  return tensor;
+}
+
+void boxed_test_parallel_for(
+    StableIValue* stack,
+    uint64_t num_args,
+    uint64_t num_outputs) {
+  Tensor res = test_parallel_for(to<int64_t>(stack[0]), to<int64_t>(stack[1]));
```
malfet, line 616:
```suggestion
  auto& res = test_parallel_for(to<int64_t>(stack[0]), to<int64_t>(stack[1]));
```

test/cpp_extensions/libtorch_agnostic_extension/setup.py
```diff
@@ -35,6 +36,16 @@ def get_extension():
     extra_compile_args = {
         "cxx": ["-fdiagnostics-color=always"],
     }
+    extra_link_args = []
+
+    # Note that adding this flag does not mean extension's parallel_for will
+    # always use OPENMP path, OpenMP path will only be used if (1) AND (2)
+    # (1) libtorch was built with OpenMP
+    # (2) extension compiles and links with -fopenmp
+    # macOS clang does not support -fopenmp so we need to skip it
```
malfet: This is an incorrect statement, it does, option simply needs to be wrapped with `-Xcompiler -fopenmp`

And indeed, if we want some sort of an abstraction there, shouldn't we have helper function, say `torch.utils.cpp_extensions.get_openmp_flags()` (which I think already exist)

test/cpp_extensions/libtorch_agnostic_extension/test/test_libtorch_agnostic.py
```diff
@@ -367,6 +369,33 @@ def test_my_clone(self, device):
...
 
+        @onlyCPU
+        # TODO: Debug this:
+        # Dynamo failed to run FX node with fake tensors:
+        # call_function libtorch_agnostic.test_parallel_for.default(*(100, 10), **{}):
+        # got RuntimeError('libtorch_agnostic::test_parallel_for() expected at most
+        # 2 argument(s) but received 3 argument(s).
+        # Declaration: libtorch_agnostic::test_parallel_for(int size, int grain_size) -> Tensor')
+        @xfailIfTorchDynamo
+        @skipIf(IS_MACOS, "Default Apple clang/g++ on macos doesn't have -fopenmp flag")
```
malfet: Again, this is a lie, it does support openmp

torch/csrc/inductor/aoti_torch/c/shim.h
```diff
@@ -422,6 +428,44 @@ aoti_torch_cpu__wrapped_quantized_linear_prepacked(
 
 AOTI_TORCH_EXPORT AOTITorchError aoti_torch_zero_(AtenTensorHandle self);
 
+// parallel utilities
+AOTI_TORCH_EXPORT void aoti_torch_lazy_init_num_threads();
+AOTI_TORCH_EXPORT bool aoti_torch_in_parallel_region();
+AOTI_TORCH_EXPORT int32_t aoti_torch_get_num_threads();
+AOTI_TORCH_EXPORT int32_t aoti_torch_get_thread_num();
```
malfet: Those two sounds like a bad name choices, may be 2nd one should be `get_thread_idx`
```suggestion
AOTI_TORCH_EXPORT int32_t aoti_torch_get_num_threads();
AOTI_TORCH_EXPORT int32_t aoti_torch_get_thread_idx();
```

torch/csrc/inductor/aoti_torch/c/shim.h
```diff
@@ -422,6 +428,44 @@ aoti_torch_cpu__wrapped_quantized_linear_prepacked(
 
 AOTI_TORCH_EXPORT AOTITorchError aoti_torch_zero_(AtenTensorHandle self);
 
+// parallel utilities
+AOTI_TORCH_EXPORT void aoti_torch_lazy_init_num_threads();
```
malfet: This API looks quite confusing, may be some explanation why this is needed would be good? (For example, I don't know myself why, on all non-emedded OSes it's pretty safe to query num cores at the initialization time)

torch/csrc/inductor/aoti_torch/shim_common.cpp
```diff
@@ -1728,3 +1730,96 @@ AOTITorchError aoti_torch_get_current_device_index(int32_t* ret_device_index) {
...
+// ABI stable parallel utilities implementations
+void aoti_torch_lazy_init_num_threads() {
+  at::internal::lazy_init_num_threads();
+}
+
+bool aoti_torch_in_parallel_region() {
+  return at::in_parallel_region();
+}
+
+int32_t aoti_torch_get_num_threads() {
```
malfet: Kind of curious - why does it need to be a 32-bit signed integer? Num threads is always a positive number, isn't it? Moreover, there aren't a single platform that PyTorch supports right now that would work faster if return value is 32-bit rather than 64-bit one

torch/csrc/inductor/aoti_torch/shim_common.cpp
```diff
@@ -1728,3 +1730,96 @@ AOTITorchError aoti_torch_get_current_device_index(int32_t* ret_device_index) {
...
+AOTITorchError aoti_torch_invoke_parallel(
+    int64_t begin,
+    int64_t end,
+    int64_t grain_size,
+    AOTIParallelLambda lambda,
+    void* ctx) {
+  AOTI_TORCH_CONVERT_EXCEPTION_TO_ERROR_CODE({
+    TORCH_CHECK(
+        AT_PARALLEL_NATIVE,
+        "Only use aoti_torch_invoke_parallel if libtorch built with AT_PARALLEL_NATIVE=1");
```
malfet: If this is a compile time check, shouldn't it be changed to
````
#if !AT_PARALLEL_NATIVE
  TORCH_CHECK(false, ....);
#endif

File: test/cpp_extensions/libtorch_agnostic_extension/setup.py [hunk 16]
```diff
@@ -35,6 +36,16 @@ def get_extension():
...
+    if sys.platform != "darwin":
+        extra_compile_args["cxx"].extend(["-fopenmp", "-D_OPENMP"])
```
malfet: Why do you need to define _OPENMP here, isn't it a compiler's job? See https://godbolt.org/z/7dEcn1vfP

Also, you'll need to pass a different flag if you are on Windows

torch/csrc/stable/parallel_utils.h
```diff
@@ -0,0 +1,55 @@
+#pragma once
+
+#include <torch/csrc/inductor/aoti_torch/c/shim.h>
```
malfet: I'm somewhat new to the codebase, but this header name makes me a bit uncofortable. Are stable API simply piggy backing on some of AOTI shim definitions? And is the intention to move it later into torch/csrc/stable folder?

torch/csrc/stable/ops.h
```diff
@@ -235,4 +244,147 @@ inline torch::stable::Tensor clone(const torch::stable::Tensor& self) {
   return to<torch::stable::Tensor>(stack[0]);
 }
 
+namespace internal {
+
+// Copied from aten/src/ATen/Parallel.h
+inline int64_t divup(int64_t x, int64_t y) {
```
malfet: I think this is 5th copy of this function that I'm seeing in the codebase. Why not move it to say `torch/csrs/stable/utils.h`? Also what's wrong with the template
```
template<typename T>
intline T divup(T x, T y) {
...
}
```

malfet (commented):
Looks like `torch::stable::internal::parallel_for` is just a copy of respective implementations from `ParallelOpenMP.h` and `ParallelNative.h`. If this is the case, why aren't you deleting the implementaiton there and just make them use this "Stable" implementation?

torch/csrc/stable/parallel_utils.h
mikaylagawarecki: Yes `torch/csrc/stable` mainly provides C++ wrappers around the AOTI shim definitions (my understanding is some of these are intended to be more ergonomic e.g. do memory management that the C header does not, provide kwarg defaults that the C header can't etc.). I don't think there is an intention to move it later into `torch/csrc/stable` cc @janeyx99 for why

torch/csrc/stable/ops.h
mikaylagawarecki: Good point, let me add it to `torch/headeronly` and make the rest of the libtorch include that

torch/csrc/inductor/aoti_torch/shim_common.cpp
mikaylagawarecki: I was following the declaration that used `int`

https://github.com/pytorch/pytorch/blob/ffe3cb226a5724ec9b0ba7a2d8b8ebd0e18760de/aten/src/ATen/Parallel.h#L20

but happy to change this to uint64_t

torch/csrc/inductor/aoti_torch/c/shim.h
mikaylagawarecki: Following the naming in the declaration, but good point, I will make this change

https://github.com/pytorch/pytorch/blob/ffe3cb226a5724ec9b0ba7a2d8b8ebd0e18760de/aten/src/ATen/Parallel.h#L24

test/cpp_extensions/libtorch_agnostic_extension/setup.py
mikaylagawarecki: ooh thank you, let me try this

Sorry, I saw other errors in the codebase that seemed to corroborate this so I assumed it was the case

https://github.com/pytorch/pytorch/blob/ffe3cb226a5724ec9b0ba7a2d8b8ebd0e18760de/torch/_inductor/cpp_builder.py#L606-L615

torch/csrc/inductor/aoti_torch/c/shim.h
mikaylagawarecki: Will add a comment that it's only for use by parallel_for in torch/csrc/stable!

torch/csrc/inductor/aoti_torch/c/shim.h
mikaylagawarecki: Good point, thank you!

test/cpp_extensions/libtorch_agnostic_extension/setup.py
mikaylagawarecki: I think windows supports this too https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-1/fopenmp.html

torch/csrc/stable/parallel_utils.h
janeyx99: Yea, I can see why the headers give discomfort. I think based on offline discussion, we should start moving our shims to a non-aoti file and go from there.

torch/csrc/stable/ops.h
```diff
@@ -235,4 +246,144 @@ inline torch::stable::Tensor clone(const torch::stable::Tensor& self) {
...
+      num_threads = std::min(num_threads, divup((end - begin), grain_size));
+    }
+
+    int64_t tid = omp_get_thread_num();
+    int64_t chunk_size = divup((end - begin), num_threads);
+    int64_t begin_tid = begin + tid * chunk_size;
+    if (begin_tid < end) {
+      try {
+        ThreadIdGuard tid_guard(static_cast<uint64_t>(tid));
+        f(begin_tid, std::min(end, chunk_size + begin_tid));
```
mikaylagawarecki, line 285: @swolchok This is where I thought we could inline f to

PR #161323, opened 2025-08-22
[ONNX] Refactor torchscript based exporter

justinchuby:
* #161546
* #161449
* __->__ #161323

Refactor torchscript based exporter logic to move them to a single (private) location for better code management. Original public module and method apis are preserved.

- Updated module paths in `torch/csrc/autograd/python_function.cpp` accordingly
- Removed `check_onnx_broadcast` from `torch/autograd/_functions/utils.py` because it is private&unused

@albanD / @soulitzer could you review changes in `torch/csrc/autograd/python_function.cpp` and
`torch/autograd/_functions/utils.py`? Thanks!

## BC Breaking
- **Deprecated members in `torch.onnx.verification` are removed**

cc @titaiwangms @albanD

Differential Revision: [D81236421](https://our.internmc.facebook.com/intern/diff/D81236421)


justinchuby:
Create symbolic_helper.py, define all in all files

soulitzer:
Changes in `torch/csrc/autograd/python_function.cpp` and
`torch/autograd/_functions/utils.py` sgtm

albanD (commented):
Why are some of the files not detected as moves? Did you change a few of these files?

The other changes look good.

justinchuby:
> Why are some of the files not detected as moves? Did you change a few of these files?
> 
> The other changes look good.

I duplicated the symbolic_* files and modified the original to import from the duplicated / private files. Maybe github only detects moves when the original file is removed?

test/onnx/test_utility_funs.py
```diff
@@ -86,86 +84,6 @@ def _model_to_graph(
         return graph, params_dict, torch_out
 
 
-@common_utils.instantiate_parametrized_tests
```
xadupre, line 89: Did you move this code somewhere else?

justinchuby: We removed because the api being tested was removed.

docs/source/onnx_verification.md
```diff
@@ -11,23 +12,3 @@
...
-
-<!-- Some deprecated members are not publicly shown -->
-```{eval-rst}
-.. py:class:: check_export_model_diff
-.. py:class:: GraphInfo
-.. py:class:: GraphInfoPrettyPrinter
-.. py:class:: OnnxBackend
-.. py:class:: OnnxTestCaseRepro
-.. py:class:: VerificationOptions
-.. py:function:: find_mismatch
```
titaiwangms, line 31: Looks like these still exist. Shouldn't we keep them until we remove them?

titaiwangms (approved):

docs/source/onnx_verification.md
justinchuby: They were removed from the public interface. This way we don’t break old code but stop advertising for users. 

In any case, these apis are removed in the subsequent PR.

angelayi (approved):

PR #161343, opened 2025-08-23
[inductor][ez] pass template rather than template.uid

coconutruben:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162017
* #161534
* #161350
* #161351
* #162238
* #161349
* #161348
* #161347
* #161346
* #161345
* #161344
* __->__ #161343
* #161342
* #161341
* #161340
* #162075

# why

- simpler interface
- enables future of extracting more things out of the template e.g. a
  hash

# what

V.choices.get_mm_configs now takes the whole template rather than just
the template.uid

# testing

```
python3 -bb -m pytest test/inductor/test_max_autotune.py -v
```

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov

Differential Revision: [D81520576](https://our.internmc.facebook.com/intern/diff/D81520576)

PR #161342, opened 2025-08-23
[inductor][aten] treat like a template in GEMMs

coconutruben:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162017
* #161534
* #161350
* #161351
* #162238
* #161349
* #161348
* #161347
* #161346
* #161345
* #161344
* #161343
* __->__ #161342
* #161341
* #161340
* #162075

# why

- central point to analyze and override all generated choices

# what

- add a pseudo heuristic for aten that just yields a single, empty
  kwargs
- add a pseudo heuristic with the bias_addmm logic for it
- add an addmm specific heuristic that yields a single choice, but
  also expands it with alpha and beta kwargs

- replace all the aten.bind calls with V.choices.get_mm_configs
  using the now matching API for aten

# testing

```
python3 -bb -m pytest test/inductor/test_max_autotune.py -v
```

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @Lucaskabela

Differential Revision: [D81520580](https://our.internmc.facebook.com/intern/diff/D81520580)

PR #161341, opened 2025-08-23
[inductor][ez] add template/externchoice uid

coconutruben:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162017
* #161534
* #161350
* #161351
* #162238
* #161349
* #161348
* #161347
* #161346
* #161345
* #161344
* #161343
* #161342
* __->__ #161341
* #161340
* #162075

# why

- to have a central registry of templates/externkernelchoice
  to match them to heuristics etc, they need unique names
- mm is both the triton template name and the aten_mm name

# what

- add a uid() to KernelTemplate/ExternKernelChoice that returns name
- override in ExternKernel to prepend "aten::"
- override in TritonTemplate to prepend "triton::"

This id is just use to find template heuristics, so it has no other
impact

# testing

```
python3 -bb -m pytest test/inductor/test_max_autotune.py -v
```

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov

Differential Revision: [D81520579](https://our.internmc.facebook.com/intern/diff/D81520579)

PR #161340, opened 2025-08-23
[inductor] move scaled_mm input nodes logic

coconutruben:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162017
* #161534
* #161350
* #161351
* #162238
* #161349
* #161348
* #161347
* #161346
* #161345
* #161344
* #161343
* #161342
* #161341
* __->__ #161340
* #162075

# why

- a step towards a unified interface for all choices, where any
  adjustment to nodes (e.g. unsqueezing) happens as part of
  choice specific preprocessing, behind a common point

# what

- move the unsqueeze logic for triton nodes for scaled_mm inside
  the new hookup for adjusting the kernel inputs for template
  heuristics

# testing

```
python3 -bb -m pytest test/inductor/test_max_autotune.py -v -k "scale"
```

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov

Differential Revision: [D81520582](https://our.internmc.facebook.com/intern/diff/D81520582)

PR #161344, opened 2025-08-23
[inductor] move max-autotune logic inside V.choices.get_mm_configs

coconutruben:
* #162017
* #161534
* #161350
* #161351
* #162238
* #161349
* #161348
* #161347
* #161346
* #161345
* __->__ #161344
* #161343
* #161342
* #161341
* #161340
* #162075

# why

- heuristics providers know decide whether to (or which choices to add)
  in the max-autotune case
- enables an eventual override point to gracefully fallback to the
  standard behavior

# what

- max-autotune is determined inside V.choices.get_mm_configs
  because it's mm only right now, we can just do
  `config.max_autotune or config.max_autotune_gemm`
  a TODO indicates that this can change in the future when this
  expands to more templates

# testing

```
```

Differential Revision: [D81520573](https://our.internmc.facebook.com/intern/diff/D81520573)


jansel (approved):

coconutruben:
@coconutruben has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81520573).

coconutruben:
@coconutruben has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81520573).

torch/_inductor/choices.py
```diff
@@ -123,6 +123,9 @@ def get_mm_configs(
         Yields:
             Template parameter dictionaries ready for maybe_append_choice
         """
+        # TODO(coconutruben): once this supports more than just GEMMs, we need to pass in
+        # the max-autotune bool, rather than inferring it here
+        max_autotune = config.max_autotune or config.max_autotune_gemm
```
jansel: Rather than duplicating this logic many times throughout the PR, let's have the config check happen inside `get_template_configs`.  Add a helper method to that class to do the check.

torch/_inductor/utils.py
```diff
@@ -1651,7 +1651,6 @@ def use_triton_template(
             )
             or (layout.device.type == "cpu" and layout.dtype in layout_dtypes)
         )
-        and (config.max_autotune or config.max_autotune_gemm)
```
jansel, line 1654: This is semantics changing.  Have you searched the codebase for every usage of this function?

Maybe it would be cleaner to have two versions of this function, or a boolean flag to disable the check.

This PR is going from a single place to control this heuristic, to the heuristic being duplicated all over.

jansel (changes requested):

torch/_inductor/utils.py
coconutruben: you're right. This is now taking a bool to skip the test and only the relevant parts are using that bool now. The check is moved into the heuristics themselves now

torch/_inductor/choices.py
coconutruben: done! the TemplateConfigHeuristics are now running with a should_run that determines whether they should run at all, and that's overwritten in the triton heuristics (and decompose k) case to be `config.max_autotune or config.max_autotune_gemm` in a base class

jansel (approved):

coconutruben:
@coconutruben has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81520573).

coconutruben:
@coconutruben has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81520573).

PR #161350, opened 2025-08-23
[inductor] leverage template stacking in V.choices.get_mm_configs

coconutruben:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162017
* #161534
* #161468
* #162293
* __->__ #161350
* #161351

# why

- now everything is in place to just gather templates and run
  the V.choices.get_mm_configs once per op
- enables any overrides inside V.choices.get_mm_configs to
  have a full view of the options for an op, not just for
  one template

# what

- replace multiple calls to V.choices.get_mm_configs with
  calls to gather the active templates, and then using those
  in a single call

# testing

```
python3 -bb -m pytest test/inductor/test_max_autotune.py -v
```

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov

Differential Revision: [D81520571](https://our.internmc.facebook.com/intern/diff/D81520571)

PR #161351, opened 2025-08-23
[inductor] FlexibleLayout for ExternKernelChoice for mms

coconutruben:
* #162017
* #161534
* #161468
* #162293
* #161350
* __->__ #161351

# why

- if we only use ExternKernelChoice we're not doing any codegen
- if we're not doing any codegen, we can use a FlexibleLayout
  here, and provide deeper passes more chances to change it

# what

- if all the kernel template choices (KTC) are with a ExternKernelChoice
  template, we switch to a FlexibleLayout before generating the choice
- add a test to make sure that works as intended (FlexibleLayout for
  only extern, and FixedLayout if Triton is involved)
    
- caveats:
    - because CPP, CUTLASS, and CK are not using
       V.choices.get_mm_configs yet, we turn off the optimization
       if either of those backends are in use. This will be relaxed
       once they support this too
    - because Triton templates are still using their own calls
       (not a single call) to get_mm_configs, it's also turned
       off there. The next diff unifies Triton + ATEN to a single
       call to get_mm_configs and that in turn allows the optimization
       there too

# testing

```
```

Differential Revision: [D81520584](https://our.internmc.facebook.com/intern/diff/D81520584)


torch/_inductor/choices.py
```diff
@@ -141,6 +143,38 @@ def _adjust_mm_configs(
...
+            return False
+        if _use_autotune_backend("CUTLASS"):
+            return False
+        if _use_autotune_backend("CK") or _use_autotune_backend("CKTILE"):
+            return False
+        if _use_autotune_backend("CPP"):
+            return False
+        return all(
+            isinstance(ktc.template, ExternKernelChoice) for ktc in adjusted_choices
+        )
```
eellison: Wouldnt this check be sufficient ? Why do we need the other checks as well ?

torch/_inductor/choices.py
```diff
@@ -220,13 +249,21 @@ def get_mm_configs(
...
             max_autotune,
         )
-        choices = []
-        # Third pass: Get adjusted choices and collect non-None ChoiceCaller objects
-        for ktc in adjusted_choices:
-            if ktc.choice is not None:
-                choices.append(ktc.choice)
-
-        return choices
+        # Layout optimization: if all choices are ExternKernelChoice and layout is FixedLayout, convert to FlexibleLayout
```
eellison, line 282: This is typically the reverse order of how layout optimization goes. would it be possible to keep the layouts Flexible, then make them Fixed ?

torch/_inductor/choices.py
coconutruben: it's because CK/CUTLASS/CPP don't go through `get_mm_configs` yet, so we don't know that the choices we're looking at here are _all_ the choices that we're ultimately comparing. only ATEN/Triton go through here

torch/_inductor/choices.py
coconutruben: this is also related to everyone going through this point. We make the layout fixed in here https://github.com/pytorch/pytorch/blob/main/torch/_inductor/kernel/mm_common.py#L48 so we'd need to refactor/duplicate that logic to avoid touching other callers for now. If okay, I'd suggest to do this here, and swap it around once every backend goes through this point. At that point, we know it's safe to do and we can defer layout making decision to the end wdyt?

coconutruben:
@coconutruben has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81520584).

coconutruben:
@coconutruben has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81520584).

torch/_inductor/choices.py
```diff
@@ -226,18 +260,21 @@ def get_mm_configs(
...
+        if self._need_to_fix_layout(adjusted_choices, op_name):
+            layout = kernel_inputs.output_layout(flexible=False)
+            for ktc in adjusted_choices:
+                ktc.layout = layout
+                # for good measure, delete the cached ChoiceCaller from the ktc if it existed.
+                # ExternKernelChoice are cheap to generate
+                if hasattr(ktc, "_choice"):
+                    del ktc._choice
```
eellison, line 290: Is this still necessary?

torch/_inductor/choices.py
```diff
@@ -176,16 +174,57 @@ def get_ktc(
...
+        # TODO(coconutruben): remove this once TRITON,CPP,CK,CUTLASS are supported
+        if _use_autotune_backend("TRITON"):
+            return True
+        if _use_autotune_backend("CUTLASS"):
+            return True
+        if _use_autotune_backend("CK") or _use_autotune_backend("CKTILE"):
+            return True
+        if _use_autotune_backend("CPP"):
+            return True
+        return any(
```
eellison, line 234: Is this check sufficient ? Or do we need the above checks as well

eellison (approved):

torch/_inductor/choices.py
coconutruben: we need the above checks until all the backends go through get_mm_configs, otherwise we don't know if there are more choices (that might have codegen). The next PR removes the TRITON check for instance

torch/_inductor/choices.py
coconutruben: yes, though that should already be None in this case. since we're swapping out the layout, we don't know what layout any previously generated ChoiceCaller was made with, so this avoids a stale cached `_choice`

jansel (approved):

PR #161355, opened 2025-08-23
[dynamo][hops] Remove const outputs from the speculated subgraph

anijain2305:
* __->__ #161355


torch/_dynamo/variables/higher_order_ops.py
```diff
@@ -74,10 +74,16 @@
...
+    Contains the treespec of the output of the speculated subgraph, and the
+    information to mask out the constant values from the output during
+    flattening and inserting them back during unflattening. Cleaning up
+    constants from the graph makes the graph simpler for AOTDispatcher and
+    Inductor.
     """
 
     treespec: pytree.TreeSpec
+    const_mask: Optional[list[bool]] = None
+    const_values: Optional[list[Any]] = None
```
zou3519: Can you describe what the scheme is for `mask` and `values` and document any invariants?

Also a post_init for this dataclass to check the invariants might be helpful

torch/_dynamo/external_utils.py
```diff
@@ -249,3 +249,25 @@ def wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _R:
...
+def filter_out_const_values(tup: tuple[Any, ...], mask: list[bool]) -> tuple[Any, ...]:
+    out = []
+    for mask_idx, m in enumerate(mask):
+        if not m:
+            out.append(tup[mask_idx])
+    return tuple(out)
+
+
+def insert_const_values_with_mask(
+    tup: tuple[Any, ...], mask: list[bool], values: tuple[Any, ...]
```
zou3519: so, mask is always the same length as values? Is there a filler value inside `values` when the mask is False at some elements?

If there are invariants then let's assert them somewhere

torch/_dynamo/variables/higher_order_ops.py
```diff
@@ -1070,6 +1109,8 @@ def speculate_branch(branch):
                 "cond",
                 source_target=self.value,
                 should_flatten_outputs=True,
+                # TODO - removing consts from control flow ops need more work
+                remove_consts_from_outputs=False,
```
zou3519, line 1137: File an issue please

torch/_dynamo/variables/higher_order_ops.py
```diff
@@ -646,6 +661,7 @@ def speculate_subgraph(
     set_subgraph_inputs="automatic",
     restore_side_effects=True,
     should_flatten_outputs=False,
+    remove_consts_from_outputs=True,
```
zou3519, line 678: Need documentation. this list has gotten too long.
- remove_consts_from_outputs -> this only applies when should_flatten_outputs=True

zou3519: I'm not sure this should be True by default, but I haven't thought very hard about this. For example, this now allows any constant outputs to be returned from a HOP. @ydwu4 thoughts?

zou3519 (commented):
this look reasonable to me. Two main comments:
1) I'm not sure this flag should be on by default for all speculate_subgraph. It makes sense to give it to invoke_subgraph
2) the checkpoint test failure is suspicious

torch/_dynamo/external_utils.py
anijain2305: I will assert the invariants. The mask is either True or False, and if the value is False, we look at values at that index to find the constant value saved earlier.

torch/_dynamo/variables/higher_order_ops.py
ydwu4: I think it's OK to filter out the constants by default. For contorl flow operators, we have some checks around the input and output specs to be consistent e.g. for while_loop, scan, map. 

For other hops I think the in out spec restrictions are more relaxed so filtering them out might be the expected behavior and potentially allow more subgraphs to be captured.

zou3519 (commented):
seems fine but could we dig into new checkpoint test failure before merging?

zou3519 (approved):
stamp to unblock

anijain2305:
The failing checkpoint test is not related to these changes. Its a pre-existing issue that got exposed because this PR makes the test go further than before

```

    def test_checkpoint_non_tensor_inputs_outputs(self):
        def foo(t1, t2, scale, t3):
            t4 = t1 + t2 * t3
            t5 = t1 * t2 + t3
            t4 *= scale
            t5 *= scale
            return scale, t4, None, True, t5, "bar", t1

        t1 = torch.rand(10, requires_grad=True)
        t2 = torch.rand(10, requires_grad=True)
        t3 = torch.rand(10)
        scale = random.randint(0, 10)
        # res = checkpoint(foo, t1, t2, scale, t3, use_reentrant=True)
        res = foo(t1, t2, scale, t3)
        self.assertEqual(scale, res[0])
        self.assertEqual((t1 + t2 * t3) * scale, res[1])
        self.assertEqual(None, res[2])
        self.assertEqual(True, res[3])
        self.assertEqual((t1 * t2 + t3) * scale, res[4])
        self.assertEqual("bar", res[5])
        self.assertEqual(t1, res[6])

        # Validate running backward.
        res[1].sum().backward(retain_graph=True)
        res[4].sum().backward(retain_graph=True)
        res[6].sum().backward()
        with self.assertRaisesRegex(
            RuntimeError, "Trying to backward through the graph a second time"
        ):
            res[6].sum().backward()
  ```

The test fails in the last 4 lines where it was expecting an assertion to be thrown. I tried removing checkpoint (see the commented line) and the test still fails. The error is somewhere with double backward not raising an error with compile, and unrelated to compile.

PR #161370, opened 2025-08-24
Remove the CUPTI  CMake check for kineto

cyyever:
This PR removes the CUPTI check because kineto has always linked to `CUDA::cupti`.


Skylion007 (approved):

sraikund16:
@cyyever it looks like it is still failing with the same signature :(

sraikund16:
@cyyever  I did notice the note in this PR https://github.com/pytorch/kineto/pull/1109: 
```
Note: Windows shared builds don't work because no C++ symbol exported, this failure doesn't affect PyTorch builds because it always uses static builds.. This serious failure should be fixed in a later PR.
```

Did you plan on fixing it before bumping the version?

sraikund16:
Looks like it is breaking for another reason now, so the cmakefile fix worked. Going to fix the lambda that is causing issues

PR #161374, opened 2025-08-24
Use `itertools.batched` if available within `new_subgroups`

evanwporter:
Uses the new `itertools.batched` (`cp3.12`) functionality if its available instead of list comprehension. Something worth considering is if we should move the try except outside of the function so it doesn't have to check every time `new_subgroups` is called.

### DIFF
--- a/torch/distributed/distributed_c10d.py
+++ b/torch/distributed/distributed_c10d.py
@@ -5458,7 +5458,7 @@ def new_subgroups(
         ranks_per_subgroup_list = [
             ranks[i : i + group_size] for i in range(0, len(ranks), group_size)
         ]
-    
+
     return new_subgroups_by_enumeration(
         ranks_per_subgroup_list,
         timeout=timeout,
--- a/torch/distributed/distributed_c10d.py
+++ b/torch/distributed/distributed_c10d.py
@@ -5433,6 +5433,7 @@ def new_subgroups(
                 "please pass in 'group_size' correctly."
             )
         group_size = torch.cuda.device_count()
+
     if group_size <= 0:
         raise ValueError(f"The arg 'group_size' ({group_size}) must be positive")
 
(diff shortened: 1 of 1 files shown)


torch/distributed/distributed_c10d.py
```diff
@@ -5446,11 +5447,21 @@ def new_subgroups(
...
+        _use_batched = True
+    except ImportError:
+        _use_batched = False
+
     ranks = get_process_group_ranks(group=group)
-    ranks_per_subgroup_list = [
-        ranks[i : i + group_size] for i in range(0, len(ranks), group_size)
-    ]
+    if _use_batched:
+        ranks_per_subgroup_list = [list(chunk) for chunk in batched(ranks, group_size)]
```
Skylion007:
```suggestion
        ranks_per_subgroup_list = list(map(list, batched(ranks, group_size)))
```
Might be marginally faster

Skylion007 (approved):

torch/distributed/distributed_c10d.py
```diff
@@ -5446,11 +5447,21 @@ def new_subgroups(
...
+        from itertools import batched
+        _use_batched = True
+    except ImportError:
+        _use_batched = False
+
     ranks = get_process_group_ranks(group=group)
-    ranks_per_subgroup_list = [
-        ranks[i : i + group_size] for i in range(0, len(ranks), group_size)
-    ]
+    if _use_batched:
```
Skylion007: Sounds like something we might just want to add backport pollyfill in a util file TBH.

torch/distributed/distributed_c10d.py
```diff
@@ -5446,11 +5447,21 @@ def new_subgroups(
             f"The world size ({world_size}) must be divisible by '{group_size=}'"
         )
 
-    # TODO: Use itertools.batched(get_process_group_ranks(group=group), group_size) instead when Python 3.12 is supported.
+    # TODO can be dropped once compat for <=cp3.11 is dropped
+    try:
```
Skylion007: Actually, this dynamic import is rather inefficient. Why not just use a python version check. the other advantage is when the minimum Python version is removed, the lintertools should flag this if else.

Skylion007 (changes requested):

torch/distributed/distributed_c10d.py
evanwporter: yeah it is expensive.

evanwporter: the newest fix should deal with the lint issues as well

Skylion007 (approved):

evanwporter:
merge -i

Skylion007 (approved):
@pytorchbot merge -i

PR #161378, opened 2025-08-24
[NVIDIA] Generate aarch64 binaries for blackwell family

johnnynunez:
Include Thor, Spark and GB300. When Orin be SBSA, we can include it
Fixes #161377

@Skylion007 @tinglvv @ptrblck @nWEIdia

### DIFF
--- a/.ci/aarch64_linux/aarch64_ci_build.sh
+++ b/.ci/aarch64_linux/aarch64_ci_build.sh
@@ -7,7 +7,7 @@ if [[ "$GPU_ARCH_VERSION" == *"12.9"* ]]; then
     export TORCH_CUDA_ARCH_LIST="8.0;9.0;10.0;12.0"
 fi
 if [[ "$GPU_ARCH_VERSION" == *"13.0"* ]]; then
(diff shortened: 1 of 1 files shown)


.ci/aarch64_linux/aarch64_ci_build.sh
```diff
@@ -6,6 +6,9 @@ GPU_ARCH_VERSION=${GPU_ARCH_VERSION:-}
 if [[ "$GPU_ARCH_VERSION" == *"12.9"* ]]; then
     export TORCH_CUDA_ARCH_LIST="8.0;9.0;10.0;12.0"
```
alexeldeib, line 7: 10.3 is supported in 12.9 fwiw :) it seems to work

https://docs.nvidia.com/cuda/archive/12.9.0/cuda-toolkit-release-notes/index.html#new-features

johnnynunez: Yes, also spark. The only really new is Thor eith OpenRM driver in cuda 13

alexeldeib: maybe dumb question, why only add for 13 not 12.9? downstream code will still need to bump for cu13 too, no?

PR #161388, opened 2025-08-25
[pytorch] Leverage `unordered_map.try_emplace()` to simplify code

tsunghsienlee:
Summary: Because [`unordered_map.try_emplace()`](https://en.cppreference.com/w/cpp/container/unordered_map/try_emplace.html) does not invoke value's constructor if key is already existed, this matches with the previous the behavior on checking the key's existence first, and then instantiate the value.

Test Plan:
OSS CI & tests

Rollback Plan:

Differential Revision: D80916349

### DIFF
--- a/aten/src/ATen/native/ForeachUtils.h
+++ b/aten/src/ATen/native/ForeachUtils.h
@@ -338,36 +338,34 @@ inline FlatMap _group_tensors_by_first_tensors_device_and_dtype(
               }
             }),
         "Tensors of the same index must be on the same device and the same dtype except `step` tensors that can be CPU and float32/64 notwithstanding");
-    if (!grouped_tensors_with_indices.count(key)) {
-      grouped_tensors_with_indices.insert(
-          {key,
-           TensorsAndIndicesT{
-               [&]() -> nested_optional_tensorvec_t {
-                 nested_optional_tensorvec_t nested_tensorvec;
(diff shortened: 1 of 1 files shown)


janeyx99 (approved):
this looks ok

aten/src/ATen/native/ForeachUtils.h
```diff
@@ -338,36 +338,34 @@ inline FlatMap _group_tensors_by_first_tensors_device_and_dtype(
...
+              for (const auto& i : c10::irange(num_lists)) {
+                std::vector<std::optional<at::Tensor>> tensors;
+                if (!nested_tensorlist[i].empty()) {
+                  // NB: num_tensors is the max possible length for any of
+                  // the inner lists of tensor references. Reserving the max
+                  // trades memory for perf. This should not have significant
+                  // impact.
+                  tensors.reserve(num_tensors);
+                }
+                nested_tensorvec.emplace_back(tensors);
```
Skylion007:
```suggestion
                nested_tensorvec.emplace_back(std::move(tensors));
```

tsunghsienlee: Done, originally I don't want to make any functional change, and this probably is done automatically by C++ compiler optimization due to the scope of `tensors` is going to end at the end of this block; however, explicitly `std::move()` here should cause no harm for sure.

PR #161394, opened 2025-08-25
Update the operator benchmarking, to benchmark using torch.compile

jainapurva:
This pull request enhances the PyTorch operator benchmarking suite by introducing support for benchmarking with `torch.compile` mode, in addition to existing Eager and JIT. It also adds peak memory measurement (fwd/bwd pass); improves the output format in JSON to be used by dashboard for reporting; and introduce some more CLI options. The new CLI flags introduced are:

- Added `--use-compile` CLI argument and corresponding logic to run benchmarks using `torch.compile`, including mutual exclusivity with `--use-jit`
- Added `--benchmark-name` argument for customizing the benchmark name in output 
- Updated default value for `--output-json-for-dashboard` to `benchmark-results.json` for more predictable output file name

Sample command to run a single operator: 
`python -m pt.mm_test --use-compile`


benchmarks/operator_benchmark/pt/mm_test.py
```diff
@@ -26,7 +26,7 @@
     M=[8, 128],
     N=[32, 64],
     K=[256, 512],
-    device=["cpu", "cuda"],
+    device=["cuda"],
```
jbschlosser: should cpu actually be removed here?

benchmarks/operator_benchmark/benchmark_runner.py
```diff
@@ -1,4 +1,5 @@
 import argparse
+from ast import parse
```
jbschlosser: is this import needed?

benchmarks/operator_benchmark/benchmark_core.py
```diff
@@ -354,8 +367,17 @@ def _measure_time(self, launch_test, test_case, iters, print_per_iter):
         """
         curr_test_total_time = 0
         time_trace = []
+        peak_bytes = None
+        # TODO: Add conditional to measure peak memory usage
```
jbschlosser: is the idea that measuring peak memory usage slows things down? or is there some other reason why it would be beneficial to disable this?

benchmarks/operator_benchmark/benchmark_core.py
```diff
@@ -330,6 +339,10 @@ def _launch_forward(self, test_case, iters, print_per_iter):
         func = test_case.run_forward
         if self.use_jit:
             func = test_case.run_jit_forward
+        elif self.use_compile:
+            func = test_case.run_compile_forward
+        else:
+            pass
```
jbschlosser: empty else doesn't seem needed - was this an LLM suggestion? :p

benchmarks/operator_benchmark/benchmark_core.py
```diff
@@ -254,9 +257,11 @@ def _print_perf_result(self, reported_run_time_us, test_case):
                     )
                 print()
             else:
-                print(f"{mode} Execution Time (us) : {reported_run_time_us[0]:.3f}\n")
+                print(f"{mode} Execution Time (us) : {reported_run_time_us[0]:.3f}")
+            if "peak_bytes" in kwargs:
+                print(f"Peak Memory (bytes) : {kwargs['peak_bytes']}\n")
 
-    def _perf_result_to_dict(self, reported_run_time_us, test_case):
+    def _perf_result_to_dict(self, reported_run_time_us, test_case, **kwargs):
```
jbschlosser: rather than utilize a `kwargs` here, I'll suggest two args for this func: `results` and `test_case` where `results` is e.g. a dict with all the relevant info

benchmarks/operator_benchmark/benchmark_pytorch.py
```diff
@@ -3,6 +3,12 @@
 
 import torch
 
+# Import the C++ extension to register the _consume operator
+try:
+    import benchmark_cpp_extension  # noqa: F401
+except ImportError:
+    # If the extension isn't built, define a simple Python fallback
```
jbschlosser: what exactly happens here if the extension isn't built? it seems like the message will be printed and the script will continue, but I'm not sure the implications. should the script error out for this case?

benchmarks/operator_benchmark/benchmark_pytorch.py
```diff
@@ -71,6 +77,17 @@ def forward_consume(self, iters: int):
         for _ in range(iters):
             torch.ops.operator_benchmark._consume(self.forward_impl())
 
+    @torch._dynamo.optimize("inductor")
```
jbschlosser: why decorate this here? it seems the outer function is already decorated with `torch.compile`, no?

jbschlosser (commented):
thanks for the hard work here! :)

I've got a few small clarification questions but it's mostly looking good. will the changes here also handle torch.compile + backwards pass?

benchmarks/operator_benchmark/pt/mm_test.py
jainapurva: No, this was just for testing. Working on cleaning the PR

benchmarks/operator_benchmark/benchmark_core.py
jainapurva: Yes, empty else doesn't seem right. This logic is no longer being used. Missed it while clean up

benchmarks/operator_benchmark/benchmark_core.py
jainapurva: Skipping cpu for now, as memory numbers for cpu aren't accurate

jbschlosser (approved):
Looks great, thanks for addressing my concerns!

jeffdaily:
Is there already a public dashboard somewhere like https://hud.pytorch.org/ that will be presenting historical data collected from operator benchmarks?

PR #161399, opened 2025-08-25
[ATen][CUDA] Add family conditional for CUTLASS matmuls

Aidyn-A:
This PR adds a custom CMake command to compile CUTLASS kernels with family conditionals. The necessity of custom command arises because NVCC does not allow compiling with both `sm_100` and `sm_100f` flags enabled at the same time. Unfortunately, `sm_100` is stored in a global variable `CMAKE_CUDA_FLAGS` which cannot be adjusted for particular targets.
Th flags `sm_100f` and `sm_120f` are important to make those matmuls work on B300/GB300 and DGX Spark.

cc @ptrblck @msaroufim @eqy @jerryzh168 @manuelcandales @SherlockNoMad @angelayi


cmake/Codegen.cmake
```diff
@@ -107,26 +111,73 @@ if(INTERN_BUILD_ATEN_OPS)
...
+        endif()
       endforeach()
       list(JOIN _file_compile_flags " " _file_compile_flags)
-
-      set_source_files_properties(${file} PROPERTIES COMPILE_FLAGS "${_file_compile_flags}")
+      get_filename_component(file_we ${file} NAME_WE)
+      set(file_obj "${CMAKE_BINARY_DIR}/caffe2/CMakeFiles/torch_cuda.dir/__/aten/src/ATen/native/cuda/${file_we}.cu.o")
+      set(custom_compile_command "")
+      set(custom_compile_command "${CMAKE_CUDA_COMPILER} -forward-unknown-to-host-compiler -fPIC")
+      list(APPEND custom_compile_command "-O2 --expt-relaxed-constexpr -DNDEBUG")
```
cyyever: O2 and NDEBUG should not appear here as they are controlled by the CMake build type.

alexsamardzic: LGTM.  What is the advantage of creating a whole custom command instead of just adding needed arch(s) flags?

eqy: @alexsamardzic it is not really an advantage, it is a necessary workaround.
From the description:
> The necessity of custom command arises because NVCC does not allow compiling with both sm_100 and sm_100f flags enabled at the same time. Unfortunately, sm_100 is stored in a global variable CMAKE_CUDA_FLAGS which cannot be adjusted for particular targets.

Basically `TORCH_CUDA_ARCH_LIST=10.0 12.0` etc., forces `sm100` and `sm120` to be added to `CMAKE_CUDA_FLAGS` which from `nvcc`'s POV conflicts with trying to add `sm100f` and `sm120f`. We previously didn't have this issue as we only needed arch-conditionals such as `sm90a` and `sm100a` which do not conflict with their non-arch-specific targets but family conditions are now added to save binary size (e.g., between `sm100` and `sm103`). The other workaround would be to change to `TORCH_CUDA_ARCH_LIST=10.0f 12.0f` etc., but that exposes the wonkiness to a more user-facing interface and also `TORCH_CUDA_ARCH_LIST=10.0+PTX` etc., would not work.

In the future we could also consider adding something like `sm103a` instead at the expense of some additional binary size, but CUTLASS did not recognize `sm103a` at the time of writing.

alexsamardzic: > > Unfortunately, sm_100 is stored in a global variable CMAKE_CUDA_FLAGS which cannot be adjusted for particular targets.

I see - so doing `REGEX_REPLACE` instead of `APPEND` would not do?

eqy: Right, I don't think we can mutate the `CMAKE_CUDA_FLAGS` here unfortunately

alexsamardzic (approved):

cmake/Codegen.cmake
```diff
@@ -87,10 +87,14 @@ if(INTERN_BUILD_ATEN_OPS)
     # for example, it will enable building for SM 90a in case PyTorch
     # built for SM 90, etc.  For examples of how to use the function,
     # see below the function itself.
```
drisspg, line 89: can we update the comments

cmake/Codegen.cmake
drisspg: I still don't love this, is it not possible to get the existing includes + compile options have 1 source of truth

Aidyn-A: Perhaps you are right, I will see if I can re-use `ATen_CUDA_INCLUDE` here:
https://github.com/pytorch/pytorch/blob/9a12bab0d305ffa8ee482126af8e76b9daaeec77/aten/src/ATen/CMakeLists.txt#L636-L637

Aidyn-A: Hmm, I do not think I can utilize `ATen_CUDA_INCLUDE` because the `Codegen.cmake` gets called earlier than `aten`.
https://github.com/pytorch/pytorch/blob/b16d3f4c8c01d461c2f01064e9ca5fa2b33f5cf1/caffe2/CMakeLists.txt#L2
https://github.com/pytorch/pytorch/blob/b16d3f4c8c01d461c2f01064e9ca5fa2b33f5cf1/caffe2/CMakeLists.txt#L47
So the `ATen_CUDA_INCLUDE` simply does not exist when it calls `_BUILD_FOR_ADDITIONAL_ARCHS` :unamused:

Aidyn-A: >O2 and NDEBUG

Alright, I am querying them from `CMAKE_CXX_FLAGS`. Only the include flags are left hard-coded (see https://github.com/pytorch/pytorch/pull/161399#discussion_r2318285866)

Aidyn-A:
A good news, CUTLASS 4.2 will have the support for [`sm_103` and `sm_103a`](https://github.com/NVIDIA/cutlass/blob/b2dd65dc864e09688245b316ac46c4a6cd07e15c/include/cutlass/arch/config.h#L176-L194), so instead of having this monstrosity, all we will need to do is to add `sm_103a` when the CUTLASS pin is updated to v4.2. I have confirmed it by re-building PyTorch locally on B300. But we will have to wait until they finalize the branch, as CUTLASS v4.2 is not fully baked yet.

Aidyn-A:
Closing in favor of #162956

PR #161400, opened 2025-08-25
fix _clone_meta stride computation for torch.preserve_format

morrison-turnansky:
Fixes #161010

fixed stride issue for cloning meta tensor

### DIFF
--- a/torch/_prims/__init__.py
+++ b/torch/_prims/__init__.py
@@ -694,7 +694,7 @@ def _clone_meta(
         )
 
     # memory_format == torch.preserve_format
-    strides = input.stride() #strides = utils.compute_elementwise_output_strides(input) #
+    strides = input.stride()
     return torch.empty_strided(
         input.shape,
         strides,
--- a/torch/_prims/__init__.py
+++ b/torch/_prims/__init__.py
@@ -694,7 +694,10 @@ def _clone_meta(
         )
 
     # memory_format == torch.preserve_format
-    strides = input.stride()
+    if torch._prims_common.is_non_overlapping_and_dense(input):
+        strides = input.stride()
+    else:
+        strides = input.contiguous().stride()
     return torch.empty_strided(
         input.shape,
         strides,
--- a/torch/_prims/__init__.py
+++ b/torch/_prims/__init__.py
@@ -694,10 +694,7 @@ def _clone_meta(
         )
 
     # memory_format == torch.preserve_format
-    if torch._prims_common.is_non_overlapping_and_dense(input):
-        strides = input.stride()
-    else:
-        strides = input.contiguous().stride()
+    strides = utils.compute_elementwise_output_strides(input)
     return torch.empty_strided(
         input.shape,
         strides,
--- a/torch/_prims/__init__.py
+++ b/torch/_prims/__init__.py
@@ -694,7 +694,7 @@ def _clone_meta(
         )
 
     # memory_format == torch.preserve_format
-    strides = utils.compute_elementwise_output_strides(input)
+    strides = input.stride() #strides = utils.compute_elementwise_output_strides(input) #
     return torch.empty_strided(
         input.shape,
         strides,
--- a/torch/_prims_common/__init__.py
+++ b/torch/_prims_common/__init__.py
@@ -655,6 +655,12 @@ def compute_elementwise_output_strides(*tensors) -> tuple[int, ...]:
     if len(tensors) == 0:
         return ()
 
+    if len(tensors) == 1:
+        if torch._prims_common.is_non_overlapping_and_dense(tensors[0]):
+            return tensors[0].stride()
+        else:
+            return tensors[0].contiguous().stride()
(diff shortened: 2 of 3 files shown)


morrison-turnansky:
I was directed to include you for review. Thank you in advance, @zou3519

zou3519 (dismissed):
Thank you for the PR, I think the semantics of preserve_format are a bit more complicated. Could you take a look please?

torch/_prims/__init__.py
```diff
@@ -694,7 +694,10 @@ def _clone_meta(
         )
 
     # memory_format == torch.preserve_format
-    strides = utils.compute_elementwise_output_strides(input)
+    if torch._prims_common.is_non_overlapping_and_dense(input):
+        strides = input.stride()
+    else:
+        strides = input.contiguous().stride()
```
zou3519: I'm not completely sure this is correct. Also, assuming that it is correct, we should update utils.compute_elementwise_output_strides -- other operators run into the same problem.

This is the function we use to compute strides when the input is not "non_overlapping_and_dense": https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/TensorIterator.cpp#L1276.

If we think it is exactly `input.contiguous.stride()`, then we should add some unittests to check that

zou3519: EDIT: .contiguous calls exactly that function so it is actually the correct thing. So my ask is that we should update `utils.compute_elementwise_output_strides` to call 
```py
if torch._prims_common.is_non_overlapping_and_dense(input):
    strides = input.stride()
else:
    strides = input.contiguous().stride()
```

morrison-turnansky: @zou3519 Thank you for checking the correct behavior. I updated ```utils.compute_elementwise_output_strides``` to call 

```python 
if torch._prims_common.is_non_overlapping_and_dense(input):
    strides = input.stride()
else:
    strides = input.contiguous().stride()
```
exactly when 1 tensor is given.
For the case of multiple tensors, it was not clear how to adapt this without more substantial changes. There is quite a bit of logic in ```compute_elementwise_output_logical_to_physical_perm```. Let me know if you want  me to update the behavior there as well.

zou3519: No, let's fix the one tensor case first.

torch/_prims_common/__init__.py
```diff
@@ -655,6 +655,12 @@ def compute_elementwise_output_strides(*tensors) -> tuple[int, ...]:
     if len(tensors) == 0:
         return ()
 
+    if len(tensors) == 1:
+        if torch._prims_common.is_non_overlapping_and_dense(tensors[0]):
+            return tensors[0].stride()
+        else:
+            return tensors[0].contiguous().stride()
```
zou3519, line 662: @laithsakka @bobrenjc93 any dynamic shapes issues around implementing this like this? Otherwise, if the tests pass, I will assume we can ship this.

torch/_prims/__init__.py
morrison-turnansky: After we get this merged, I can look into the multiple tensor case if you'd like.

torch/_prims_common/__init__.py
morrison-turnansky: I took a look at failing tests, still figuring it out. It is an issue that was introduced  by the change with inductor.

```python
def test_clone_not_memory_dense():
    def foo():
        x = torch.randn(10, 8).t()[::2, ::2]
        y = x.clone()
        return y
    y = foo()
    assert y.stride() == (1, 4)
    print("uncompiled")
    y = torch.compile(foo, backend="eager")()
    print("eager")
    assert y.stride() == (1, 4)
    y = torch.compile(foo, backend="aot_eager")()
    print("aot_eager")
    assert y.stride() == (1, 4)
    y = torch.compile(foo, backend="inductor")()
    print("inductor")
    print(y.stride())
    assert y.stride() == (1, 4)
```

Lucaskabela:
Can you try rebasing and running the test again?

torch/_prims_common/__init__.py
```diff
@@ -655,6 +655,12 @@ def compute_elementwise_output_strides(*tensors) -> tuple[int, ...]:
     if len(tensors) == 0:
         return ()
 
+    if len(tensors) == 1:
+        if torch._prims_common.is_non_overlapping_and_dense(tensors[0]):
```
laithsakka, line 659: we prob should rename def is_non_overlapping_and_dense to 
is_non_overlapping_and_dense_or_false. its fine its reasonable definition of unbacked semantics its probably would be good to ensure inductor behave similarly with unbacked inputs.

morrison-turnansky: @laithsakka Thank you for the note, these commits were merged on PR, https://github.com/pytorch/pytorch/pull/163017. However that PR did not fully capture all the correct behavior. When the future work is implemented, I will make a note of this preserved naming scheme.

PR #161404, opened 2025-08-25
Invert unary read and write for fusion

eellison:
* __->__ #161404

For [this repro](https://gist.github.com/eellison/75a99616a0fcca0436316bbfd8987fae) enables fusion of `to_blocked` with the prior `to_mx` calculation, so that there is only a single kernel per tensor, resulting in a 10% speedup of the non conversion code (need to update my local devserver to 12.9 to time the matmul as well).

The `to_mx` kernel has a contiguous write:

```Py
op6_op7: FusedSchedulerNode(SchedulerNode,SchedulerNode)
op6_op7.writes = [MemoryDep('buf6', c0, {c0: 2097152}), MemoryDep('buf7', c0, {c0: 67108864})]
op6_op7.unmet_dependencies = []
op6_op7.met_dependencies = [MemoryDep('arg1_1', c0, {c0: 67108864})]
op6_op7.outputs = [
    buf6: ComputedBuffer
    buf6.layout = FixedLayout('cuda:0', torch.float32, size=[8192, 256], stride=[256, 1])
    buf6.users = [
        NodeUser(node=SchedulerNode(name='op7'), can_inplace=False, is_weak=False),
        NodeUser(node=SchedulerNode(name='op9'), can_inplace=False, is_weak=False),
    ]
    buf7: ComputedBuffer
    buf7.layout = FixedLayout('cuda:0', torch.float8_e4m3fn, size=[8192, 256, 32], stride=[8192, 32, 1])
    buf7.users = [NodeUser(node=ExternKernelSchedulerNode(name='op10'), can_inplace=False, is_weak=False)]
]
```

While the `to_blocked` has a single discontiguous read and a single contiguous write.

```Py
op9: SchedulerNode(ComputedBuffer)
op9.writes = [MemoryDep('buf9', c0, {c0: 2097152})]
op9.unmet_dependencies = [   MemoryDep('buf6', 32768*((c0//32768)) + 8192*(((ModularIndexing(c0, 1, 16))//4)) + 256*(ModularIndexing(c0, 16, 32)) + 4*(ModularIndexing(c0, 512, 64)) + (ModularIndexing(ModularIndexing(c0, 1, 16), 1, 4)), {c0: 2097152})]
op9.met_dependencies = []
op9.outputs = [
    buf9: ComputedBuffer
    buf9.layout = FixedLayout('cuda:0', torch.float8_e8m0fnu, size=[2097152], stride=[1])
    buf9.users = [NodeUser(node=ExternKernelSchedulerNode(name='op10'), can_inplace=False, is_weak=False)]
]
```

To enable fusion, we invert the read, giving op9 and contiguous read and discontiguous write. More explanation here: https://gist.github.com/eellison/6f9f4a7ec10a860150b15b719f9285a9

[Tlparse with this optimization](https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/eellison/custom/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000). 

[Tlparse without this optimization](https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/eellison/custom/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000).

cc @msaroufim @jerryzh168 @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @mlazos


test/inductor/test_fp8.py
```diff
@@ -589,6 +591,46 @@ def linear(x_fp8, x_inverse_scale, w_t_fp8, w_inverse_scale, bias):
         self.assertEqual(y_compiled.dtype, dtype)
         torch.testing.assert_close(y_eager, y_compiled, rtol=1e-2, atol=0.07)
 
+    @unittest.skipIf(not PLATFORM_SUPPORTS_FP8, f8_msg)
+    def test_mx_fusion(self):
+        from torchao.prototype.mx_formats.mx_tensor import MXTensor
```
drisspg: dont think this will be avail in torch cI

test/inductor/test_loop_ordering.py
```diff
@@ -1046,6 +1047,90 @@ def fn(nodes):
...
+            # Wrong coefficient ratios: 4 ≠ 1×2 ❌
+            (4 * ModularIndexing(p, 1, 8) + ModularIndexing(p, 8, 2), False, None),
+            (
+                100 * FloorDiv(p, 100) + 7 * ModularIndexing(p, 1, 100),
+                False,
+                None,
+            ),  # Wrong ratios
+            (FloorDiv(p, 100) + FloorDiv(p, 10) + p, False, None),  # Overlapping ranges
+            (p**2 + 10 * p + 1, False, None),  # Quadratic
+            (sympy.sin(p) + sympy.cos(p), False, None),  # Trigonometric
```
drisspg, line 1285: I trust you and the AI confirmed these tests are actually helpful

test/inductor/test_loop_ordering.py
```diff
@@ -1046,6 +1047,90 @@ def fn(nodes):
...
+            if should_invert:
+                self.assertIsNotNone(reconstruction, f"Expected invertible: {expr}")
+                # Test correctness on sample values
+                test_vals = [0, 1, test_range - 1] if test_range else []
+                for val in test_vals:
+                    forward = int(expr.subs(p, val))
+                    recovered = int(reconstruction.subs(p, forward))
+                    self.assertEqual(val, recovered)
+            else:
+                self.assertIsNone(reconstruction, f"Expected non-invertible: {expr}")
```
drisspg, line 1296: generate_inverse_formula returns None when it can't invert ?

torch/_inductor/config.py
```diff
@@ -598,6 +598,8 @@ def use_autoheuristic(name: str) -> bool:
     os.environ.get("TORCHINDUCTOR_LOOP_ORDERING_AFTER_FUSION", "0") == "1"
 )
 
+loop_index_inversion_in_fusion: bool = True
```
drisspg, line 686: lil doc block would be nice :)

drisspg: even if it just points to : https://github.com/pytorch/pytorch/pull/161404/files#diff-52e228e64281bf00dceb40bf50bb277c7905da4bfd31af34579d9774ed5f5d0aR24

eellison:
@drisspg, yes lets do it, will clean up today/tomorrow and submit for review.

test/inductor/test_loop_ordering.py
```diff
@@ -1163,6 +1164,113 @@ def fn(nodes):
...
+        super().setUpClass()
+
+        gm = torch.fx.symbolic_trace(lambda: 0)
+        graph = GraphLowering(gm)
+        graph.scheduler = MockScheduler
+        cls._exit_stack = contextlib.ExitStack()
+        cls._exit_stack.enter_context(V.set_graph_handler(graph))
+
+    def _check_expr(self, expr, reconstruction, val_range):
+        import numpy as np
```
drisspg, line 1204: we already have a dep on numpy right?

test/inductor/test_loop_ordering.py
```diff
@@ -1163,6 +1164,113 @@ def fn(nodes):
...
+        )
+        reconstruction_numpy_ready = reconstruction.replace(
+            FloorDiv, floordiv_replacement
+        ).replace(ModularIndexing, modularindexing_replacement)
+
+        # Now lambdify with standard numpy
+        forward_func = lambdify(p0, expr_numpy_ready, modules="numpy")
+        inverse_func = lambdify(p0, reconstruction_numpy_ready, modules="numpy")
+
+        test_values = np.arange(0, val_range, dtype=np.int64)
```
drisspg, line 1230: is this needed acutally?

test/inductor/test_loop_ordering.py
```diff
@@ -1163,6 +1164,113 @@ def fn(nodes):
...
+                100 * FloorDiv(p, 100)
+                + 10 * FloorDiv(ModularIndexing(p, 1, 100), 10)
+                + ModularIndexing(p, 1, 10),
+                True,
+                1000,
+            ),
+            (4 * p, False, 64),  # expr and inverse not bijections
+            # when sorted, invertible
+            (ModularIndexing(p, 1, 10) + 10 * ModularIndexing(p, 10, 10), True, None),
+            # Wrong coefficient ratios: 4 ≠ 1×2 ❌
```
drisspg: emojis actually caused one of my PR so get reverted once

test/inductor/test_loop_ordering.py
eellison: in test but not in core i believe

test/inductor/test_loop_ordering.py
eellison: otherwise, the `2097152` values takes too long to run, and it's nice to check all of them instead of just a subset

test/inductor/test_fp8.py
```diff
@@ -749,6 +750,333 @@ def linear(x_fp8, x_inverse_scale, w_t_fp8, w_inverse_scale, bias):
         self.assertEqual(y_compiled.dtype, dtype)
         torch.testing.assert_close(y_eager, y_compiled, rtol=5e-2, atol=0.07)
 
+    @unittest.skipIf(not PLATFORM_SUPPORTS_FP8, f8_msg)
+    def test_mx_fusion(self):
+        def forward(
+            arg0_1: "f32[8192, 8192][8192, 1]cuda:0",
+            arg1_1: "f32[8192, 8192][8192, 1]cuda:0",
+        ):
+            view: "f32[8192, 256, 32][8192, 32, 1]cuda:0" = (
```
shunting314: Why not put the original pytorch code (rather than the generated fx graph) in the test?

test/inductor/test_fp8.py
```diff
@@ -749,6 +750,333 @@ def linear(x_fp8, x_inverse_scale, w_t_fp8, w_inverse_scale, bias):
...
+
+        torch.manual_seed(42)
+        A = torch.randn(M, K, dtype=torch.float32, device=device)
+        B = torch.randn(K, N, dtype=torch.float32, device=device)
+
+        f_c = torch.compile(fullgraph=True)(forward)
+
+        out, code = run_and_get_code(f_c, A, B)
+        self.assertEqual(out, forward(A, B))
+        FileCheck().check("call").check_count(".run(", 2, exactly=True).run(code[0])
```
shunting314: More robust to check for '@triton.jit' or check the kernel count metric. Since if people run the test with config.benchmark_kernels enabled, there will be more matches.

torch/_inductor/scheduler.py
```diff
@@ -4320,6 +4456,12 @@ def can_fuse(
             smaller_node.expand_dimension_for_pointwise_node(expand_dim, expand_size)
             shared_data_score = self.score_fusion_memory(node1, node2)
 
+        if (
+            config.loop_index_inversion_in_fusion
```
shunting314, line 4852: Does this have to be run in every round of fusion? If only run it in the final round is fine, add a condition check for 'can_reroder' like the LOAF branch.

Maybe we need rename 'can_reorder' to 'is_final_round' since the meaning has been expanded here.

shunting314 (approved):

test/inductor/test_fp8.py
eellison: no torch ao dependency

shunting314: Can the dependencies being duplicate here? Based on the fx graph, I think it should not be that large amount of code duplication.

torch/_inductor/scheduler.py
eellison: done!

torch/_inductor/scheduler.py
```diff
@@ -4229,6 +4232,148 @@ def decide_fusion_fail_reason(
...
+        node1_write = node1_writes[node2_read.name]
+
+        if not isinstance(node1_write, MemoryDep):
+            return -1
+
+        # We are checking for compatibility with the normalized node1 write
+        # then modifying node2 reads/writes. since the node1 write will be just used
+        # for compatibility, while node2 will be used in actual modification, just
+        # normalize node1 not node2.
+        node1_write = node1_write.normalize()
```
eellison, line 4376: cc @shunting314 this was only change i needed to make

test/inductor/test_fp8.py
```diff
@@ -937,6 +937,234 @@ def linear(x_fp8, x_inverse_scale, w_t_fp8, w_inverse_scale, bias):
...
+            )
+
+            for seed in range(42):
+                input_values.clear()
+                torch.manual_seed(seed)
+                # without dividing, outputs get way too large
+                A = torch.randn(M, K, dtype=torch.float32, device=device)
+                B = torch.randn(K, N, dtype=torch.float32, device=device)
+
+                # Uses fake_scaled_mm custom op (no CUDA 12.8 needed!)
```
eellison, line 1158: cc @drisspg @vkuzo  I get this failure on ci that neither shunting nor myself get locally..
```
FAILED [1.8686s] inductor/test_fp8.py::TestFP8Lowering::test_mx_fusion - AssertionError: Tensor-likes are not 'equal'!

Mismatched elements: 271686 / 67108864 (0.4%)
The first mismatched element is at index (0, 33).
idx 0 seed 0

To execute this test, run the following from the base repo dir:
    python test/inductor/test_fp8.py TestFP8Lowering.test_mx_fusion

This message can be suppressed by setting PYTORCH_PRINT_REPRO_ON_FAILURE=0
!!!!!!!!!!!!!!!!!!!!!!!!!! stopping after 1 failures !!!!!!!!!!!!!!!!!!!!!!!!!!!
================= 1 failed, 245 deselected, 2 rerun in 16.34s ==================
Got exit code 1
FAILED CONSISTENTLY: test/inductor/test_fp8.py::TestFP8Lowering::test_mx_fusion
Stopping at first consistent failure
The following tests failed consistently: ['test/inductor/test_fp8.py::TestFP8Lowering::test_mx_fusion']
```

vkuzo: if I understand correctly, this tests that compile does not change numerics, and doesn't actually test that the pattern is fused into one kernel?

if yes, then agreed that the failure is problematic but also seems orthogonal to this PR?  I wonder if this test by itself would pass without the PR?

test/inductor/test_fp8.py
```diff
@@ -937,6 +937,234 @@ def linear(x_fp8, x_inverse_scale, w_t_fp8, w_inverse_scale, bias):
...
+                _ = forward(A, B)
+                f_c(A, B)
+
+                assert len(input_values) == 2
+                for i in range(4):
+                    self.assertEqual(
+                        input_values[0][i], input_values[1][i], msg=f"idx {i} seed {seed}"
```
vkuzo: curious on the reason of saving data in global `input_values` instead of just saving it in local vars and comparing directly?

test/inductor/test_fp8.py
vkuzo: can we also either save the bad inputs to disk or just print out the first 64 values of data and first 2 values of scale to stdout on failure? If first mismatch is in index 33 then maybe a small slice would give a local repro.

test/inductor/test_fp8.py
eellison: Since we're not actually running the mx precision, I was worried that the incorrectness was due to running weird random data and getting nans/infs. So i'm saving the input values. also part of this is me debugging why we're getting errors on ci and not locally.

test/inductor/test_fp8.py
eellison: > and doesn't actually test that the pattern is fused into one kernel?

We do check there is one kernel per input value.

PR #161405, opened 2025-08-25
[inductor] Runtime estimations: use nccl estimator; mm only benchmark mode

IvanKobzarev:
* __->__ #161405

During comms reordering , sink wait iterative observed previous runtime estimations pretty off for collectives and mms.

Adding optional usage of:
- c10d.time_estimator for collectives, which is based on NCCL estimator

Benchmark mode only for matmuls, as they are highly dependent on mm backend

- The logic mostly copied from Ruisi's PRs for inductor simple_fsdp https://github.com/pytorch/pytorch/pull/157572

This estimations corrections are in default `BaseSchedulerNode.estimate_runtime()`

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben

Differential Revision: [D81152294](https://our.internmc.facebook.com/intern/diff/D81152294)


torch/_inductor/comm_analysis.py
```diff
@@ -49,6 +55,8 @@ def get_collective_type(node: ir.IRNode) -> NCCL_COLL:
         return NCCL_COLL.ALL_GATHER
     elif "reduce_scatter" in kernel_name:
         return NCCL_COLL.REDUCE_SCATTER
+    elif "torch.ops._dtensor.shard_dim_alltoall.default" in kernel_name:
+        return NCCL_COLL.ALL_TO_ALL
```
Microve, line 59: Do we support "torch.ops._c10d_functional.all_to_all_single.default"?

torch/_inductor/comm_analysis.py
```diff
@@ -158,6 +166,123 @@ class NCCL_PROTO(IntEnum):
...
+    if name == "torch.ops._c10d_functional.all_gather_into_tensor.default":
+        fn = torch.ops._c10d_functional.all_gather_into_tensor
+        return fn, {
+            "input": in_t,
+            "group_size": pg_size,
+            "group_name": pg_name,
+        }
+    elif name == "torch.ops._c10d_functional.all_gather_into_tensor_out.default":
+        # TODO: use real all_gather_into_tensor_out
+        fn = torch.ops._c10d_functional.all_gather_into_tensor
```
Microve: Curious why we use `all_gather_into_tensor` here for `all_gather_into_tensor_out`?

torch/_inductor/comm_analysis.py
```diff
@@ -158,6 +166,123 @@ class NCCL_PROTO(IntEnum):
...
+        return fn, {
+            "input": in_t,
+            "reduce_op": reduce_op,
+            "group_name": pg_name,
+        }
+    elif name == "torch.ops._c10d_functional.all_to_all_single.default":
+        fn = torch.ops._c10d_functional.all_to_all_single
+        # Artificial uniform split assumption,
+        # which can be not true in case of uneven sharding.
+        split_sizes = [in_t.size(0) // pg_size] * pg_size
```
Microve: if in_t.size(0) is 5 and pg_size is 2, then split_sizes is [2,2]? If so, it only covers 4 not 5?

torch/_inductor/comm_analysis.py
```diff
@@ -237,9 +364,13 @@ def estimate_nccl_collective_runtime(node: ir.IRNode) -> float:
             nInterSteps = 2 * nNodes
         else:
             nInterSteps = 0
-    elif coll in (NCCL_COLL.REDUCE_SCATTER, NCCL_COLL.ALL_GATHER):
+    elif coll in (NCCL_COLL.REDUCE_SCATTER, NCCL_COLL.ALL_GATHER, NCCL_COLL.ALL_TO_ALL):
         nInterSteps = nNodes - 1
 
+    tensor_size_mult = 1.0
+    if coll == NCCL_COLL.ALL_TO_ALL:
+        tensor_size_mult = 2.0 / group_size
```
Microve: Why are 2.0 divided by `group_size`?

torch/_inductor/comm_analysis.py
```diff
@@ -158,6 +166,123 @@ class NCCL_PROTO(IntEnum):
...
+        }
+    elif name == "torch.ops._c10d_functional.all_gather_into_tensor_out.default":
+        # TODO: use real all_gather_into_tensor_out
+        fn = torch.ops._c10d_functional.all_gather_into_tensor
+        return fn, {
+            "input": in_t,
+            "group_size": pg_size,
+            "group_name": pg_name,
+        }
+    elif name == "torch.ops._c10d_functional.reduce_scatter_tensor.default":
```
eellison: I think the way you generating input nodes below generically would be better here as well

torch/_inductor/scheduler.py
```diff
@@ -895,6 +913,91 @@ def get_prologue_template_epilogue(
...
+        return None
+    fn = mms_fns[python_kernel_name]
+    if not isinstance(snode.node, ir.ExternKernel):
+        return None
+    args = snode.node.inputs
+    args = snode.node.fill_non_provided_args(
+        [*args, *snode.node.constant_args], snode.node.kwargs
+    )
+    kwargs = snode.node.kwargs
+    flat_args, flat_args_pytree_spec = pytree.tree_flatten((args, kwargs))
```
eellison: These are the same utilities we could have used above for getting the nccl inputs.

torch/_inductor/scheduler.py
```diff
@@ -895,6 +913,91 @@ def get_prologue_template_epilogue(
...
+        end_event.record(torch.cuda.current_stream())
+        cpu_end = time.time()
+        torch.cuda.synchronize()
+        cpu_time = cpu_end - cpu_start
+        total_op_time = start_event.elapsed_time(end_event) - cpu_time
+        mean_op_time_ms = total_op_time / num_iters
+        del flat_args
+        mean_op_time_ns = mean_op_time_ms * 1e6
+        cache.put(cache_key, mean_op_time_ns)
+        return mean_op_time_ns
```
eellison: Could we reuse `benchmark_gpu` here ?

torch/_inductor/scheduler.py
```diff
@@ -895,6 +913,91 @@ def get_prologue_template_epilogue(
...
+
+    def get(self, cache_key: Any) -> Optional[float]:
+        return self.dict.get(cache_key, None)
+
+    def put(self, cache_key: Any, value: float) -> float:
+        self.dict[cache_key] = value
+        return value
+
+
+def estimate_runtime_benchmark(snode: BaseSchedulerNode) -> Optional[float]:
```
eellison: Have you tried `benchmark_fused_nodes` ? this should already accomplish what this function does

torch/_inductor/scheduler.py
```diff
@@ -2060,6 +2163,7 @@ def _init(self, nodes: list[ir.Operation]) -> None:
         super().__init__()
         V.graph.scheduler = self
         self.backends: dict[torch.device, BaseScheduling] = {}
+        self.estimate_runtime_cache = EstimateRuntimeCache()
```
eellison: Can we make this a machine local cache ? See: https://github.com/pytorch/pytorch/blob/1f820de639c75a1562d3fb03f160439f853ae07b/torch/_inductor/fx_passes/pad_mm.py#L250-L268

eellison (commented):
few comments

torch/_inductor/comm_analysis.py
IvanKobzarev: Yes, It should be supported, but I have not tested it.

torch/_inductor/comm_analysis.py
IvanKobzarev: I reused arguments parsing for  `all_gather_into_tensor`, 
as collective work should be the same.

But in _out variant we do not have memory allocation

torch/_inductor/comm_analysis.py
IvanKobzarev: Yeah, just added hacky multiplier to be closer to the values during benchmarking.

ruisizhang123 (commented):
I found there is a `config.estimate_op_runtime`, which allows users to parse customized op estimation function to inductor: https://github.com/pytorch/pytorch/blob/b7e207ca9f046ddd716076965a0cce403ba99052/torch/_inductor/comms.py#L1227-L1236.


It might make more sense to have `runtime_estimations_use_nccl_lib_estimations` and `runtime_estimations_mms_benchmark`  as a callable function to `estimate_op_runtime`?

IvanKobzarev:
@IvanKobzarev has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81152294).

torch/_inductor/comm_analysis.py
```diff
@@ -158,6 +166,49 @@ class NCCL_PROTO(IntEnum):
...
+        # -1000 constant is NCCL return in case of error during estimations.
+        # Observed it for all_to_all estimations.
+        if est_time_us < 0:
+            return None
+        est_time_ns = est_time_us * 1e3
+        return est_time_ns
+    except Exception as e:
+        # NCCL estimator can fail
+        log.info(e)
+        return None
```
eellison, line 202: Would it it make sense to more narrowly try ... catch around the nccl estimation? so that we do not catch unrelated exceptions?

torch/_inductor/config.py
```diff
@@ -413,6 +413,10 @@ def prologue_fusion_enabled() -> bool:
 # for built-in estimation function, pass in "default"; for user-defined estimation function, pass in the function handle
 estimate_op_runtime = "default"
 
+runtime_estimations_use_nccl_lib_estimations: bool = False
```
eellison: would it make sense to have a distributed/comms config class ?

torch/_inductor/comm_analysis.py
eellison: The rest of this code is a direct port of the nccl logic

```
    The following heuristics are copied from https://github.com/NVIDIA/nccl/blob/master/src/graph/tuning.cc.
    We aim to estimate the runtime as accurately as possible.
```
does it make sense to keep that property ?

eellison: also, i dont see AlltoAll here for whatevr reason.. https://github.com/NVIDIA/nccl/blob/master/src/graph/tuning.cc

torch/_inductor/scheduler.py
```diff
@@ -898,6 +923,78 @@ def get_prologue_template_epilogue(
...
+            fn(*args, **kwargs)
+        end_event.record(torch.cuda.current_stream())
+        cpu_end = time.time()
+        torch.cuda.synchronize()
+        cpu_time = cpu_end - cpu_start
+        total_op_time = start_event.elapsed_time(end_event) - cpu_time
+        mean_op_time_ms = total_op_time / num_iters
+        mean_op_time_ns = mean_op_time_ms * 1e6
+        cache.set_value(cache_key, value=mean_op_time_ns)
+        return mean_op_time_ns
```
eellison: Bump - why do we need to reimplement a timer function inline. Can you use `do_bench`? This will do things that make the calculation more accurate like cache clearing etc (in this impl the inputs might already be in cache which will distort timing)

torch/_inductor/utils.py
```diff
@@ -3574,3 +3575,38 @@ def python_subprocess_env() -> dict[str, str]:
         env["PYTHONHOME"] = sysconfig.get_path("data")
 
     return env
+
+
+def snode_args_kwargs(snode: BaseSchedulerNode):  # type: ignore[no-untyped-def]
```
eellison: `-> tuple[List[Any], dict[str, Any]]` ?

torch/_inductor/scheduler.py
```diff
@@ -2182,6 +2279,31 @@ def _init(self, nodes: list[ir.Operation]) -> None:
                 assign_memory_planning_info_for_scheduler_buffers(
                     self.nodes, self.name_to_buf
                 )
+
+            runtime_estimations = None
+            if config.runtime_estimations_align_across_all_distributed_ranks:
```
eellison: We shouldnt need to do this if `runtime_estimations_mms_benchmark` is not set, since the other estimations are deterministic.

torch/_inductor/scheduler.py
```diff
@@ -2182,6 +2279,31 @@ def _init(self, nodes: list[ir.Operation]) -> None:
...
+                dist.all_gather_object(
+                    gathered_runtime_estimations, list(runtime_estimations.values()), pg
+                )
+                median_runtime_estimations = torch.median(
+                    torch.tensor(gathered_runtime_estimations), dim=0
+                ).values.tolist()
+                for idx, (key, value) in enumerate(runtime_estimations.items()):
+                    self.nodes[
+                        idx
+                    ].override_estimated_runtime = median_runtime_estimations[idx]
```
eellison: Nit: can you factor this out to a helper ? the `Scheduler.__init__` is already very long, and people who are trying to understand the Scheduler shouldnt need to go through details of runtime estimation

test/distributed/test_inductor_collectives.py
```diff
@@ -1821,6 +1823,8 @@ def _reorder_communication_preserving_peak_memory(
                 ],
                 "allow_buffer_reuse": False,
                 "test_configs.track_memory_lifecycle": "error",
+                "runtime_estimations_align_across_all_distributed_ranks": True,
```
eellison: One thing you can do to test this is add a post fusion custom pass, and check that the estimations have been set on the mms. 

I added a test like that here https://github.com/pytorch/pytorch/blob/9491d289b329e4ba4a9f5f5b1be7960671bb7840/test/inductor/test_loop_ordering.py#L790-L812

torch/_inductor/comm_analysis.py
```diff
@@ -158,6 +166,49 @@ class NCCL_PROTO(IntEnum):
...
+            return None
+
+        from torch.distributed.distributed_c10d import _resolve_process_group
+
+        pg_name = kernel.constant_args[-1]  # type: ignore[attr-defined]
+        pg = _resolve_process_group(pg_name)
+        rank: int = torch.distributed.get_rank(pg)
+        # TODO(ivankobzarev): Figure out how we can use time estimations,
+        # without cuda allocations.
+        device = torch.device(f"cuda:{rank}")
```
eellison: What happens today ?

torch/_inductor/scheduler.py
eellison: The convention around the codebase is milliseconds. See: https://github.com/pytorch/pytorch/blob/9491d289b329e4ba4a9f5f5b1be7960671bb7840/torch/_inductor/runtime/benchmarking.py#L148-L173 and https://github.com/triton-lang/triton/blob/0a93c966ed599a9932f16d5a622814ceb994c0da/python/triton/testing.py#L127-L136

eellison (commented):
looks good but a few comments

torch/_inductor/config.py
IvanKobzarev: Yes, agree, I will introduce torch._inductor.config_comms.py, moving in this PR new configs.
But in following PRs I will move all distributed configs there.

torch/_inductor/scheduler.py
eellison: although, i see that the collective timing uses nano seconds, so we'll need to be careful

IvanKobzarev: Yes, I will align all the estimations to ms.

torch/_inductor/comm_analysis.py
IvanKobzarev: Now we allocate real cuda tensor for this estimation. I tried "meta" tensor and it failed in estimator. I guess we can fix it on distributed side.

torch/_inductor/scheduler.py
IvanKobzarev: Converted all sources of time to ms.

test/distributed/test_inductor_collectives.py
IvanKobzarev: Thanks.
Added the checks,
reused customly set pass in `reorder_for_compute_comm_overlap_passes` ( in test_reorder_peak_memory_bucketed in test_inductor_collectives.py)

To check the effect of `runtime_estimations_align_across_all_distributed_ranks` and `runtime_estimations_mms_benchmark`

IvanKobzarev:
@IvanKobzarev has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81152294).

IvanKobzarev:
@IvanKobzarev has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81152294).

torch/_inductor/config.py
```diff
@@ -413,6 +413,8 @@ def prologue_fusion_enabled() -> bool:
 # for built-in estimation function, pass in "default"; for user-defined estimation function, pass in the function handle
 estimate_op_runtime = "default"
 
+runtime_estimations_mms_benchmark: bool = False
```
eellison, line 419: nit: config menu

eellison (approved):

PR #161408, opened 2025-08-25
Forward fix periodic vision build

atalman:
Trying to forward fix: https://github.com/pytorch/pytorch/issues/161358 use SM 80 architecture by default


zou3519 (approved):

huydhn (approved):
I have updated the PR to properly handle nogpu case where `TORCH_CUDA_ARCH_LIST` couldn't be deduced from `nvidia-smi`.

PR #161410, opened 2025-08-25
Fix issue with FloorDiv handling

nandesuka:
Summary:
The launch grid calculation is generated as string and evaluated to get sympy expressions. 

The default "python" mode generates expressions using negative integer division behaviour in python. This creates problems when we try generate generic code for representing the launch grid calculation at the FX graph level.

In this diff we:

1. Add a new mode "python_slow" which achieves the same ceildiv behaviour that is generally applicable to integer division
2. Move and generalize the logic for replacing sympy.floor to CeilDiv

Test Plan:
CI

Rollback Plan:

Differential Revision: D80744150

### DIFF
--- a/test/inductor/test_fxir_backend.py
+++ b/test/inductor/test_fxir_backend.py
@@ -402,6 +402,57 @@ class FxirTestCase(InductorTestCase):
             ]
             self.assertEqual(placeholder.meta["val"], symbol)
 
+    @parametrize(
+        "shape",
+        [
+            (20,),
+            (50, 30),
+            (50, 30, 40),
+        ],
+    )
+    @torch._inductor.config.patch(
+        {
+            "pad_dynamic_shapes": True,
+            "comprehensive_padding": True,
+            "padding_alignment_bytes": 32,
+            "pad_outputs": True,
+        }
+    )
+    def test_dynamic_shapes_with_padding(self, shape):
+        """
+        Test a graph with dynamic shapes with padding.
+        """
+
+        def get_input(shape):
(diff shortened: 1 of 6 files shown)


jansel (approved):

test/inductor/test_fxir_backend.py
```diff
@@ -400,6 +400,57 @@ def get_input():
...
+        (empty_strided,) = gm.graph.find_nodes(
+            op="call_function", target=torch.empty_strided
+        )
+        example_tensor = empty_strided.meta["val"]
+        symbolic_dims = example_tensor.shape
+        symbolic_strides = example_tensor.stride()
+
+        align_elems = 32 // args[0].dtype.itemsize
+        expected_strides = [1 for _ in range(len(shape))]
+        for i in range(len(shape) - 1, 0, -1):
```
blaine-rister, line 448: Thanks for the thorough testing!

torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -2980,14 +2980,14 @@ class GridExpr:
     """Generate code for grid size expressions in launcher"""
 
     inductor_meta: dict[str, Any]
-    mode: Literal["python", "cpp"] = "python"
+    mode: Literal["python", "cpp", "python_slow"] = "python"
```
blaine-rister, line 3082: Is `"python_slow"` mode used by the FX converter? I'm trying to figure out where that's configured.

blaine-rister (approved):
LGTM!

PR #161412, opened 2025-08-25
NLLLoss: validate target is 0D when input is 1D

### LINKED ISSUE #157420: nll_loss gives result when both input and target are 1D tensor
### 🐛 Describe the bug

According to the [official document](https://docs.pytorch.org/docs/2.7/generated/torch.nn.functional.nll_loss.html#torch.nn.functional.nll_loss), nll_loss can compute case `input.shape=(N, C)` and `target.shape=(N,)`, where `N` is batch number and `C` is numbert of classes, in this case `input.dim=2` and `target.dim=1`.
When N=1, it can be simplified to `input.shape=(C), target.shape=()`, in this case `input.dim=1, target.dim=0`. 
So what is the meaning of case `input.dim=1, target.dim=1`? nll_loss gives me the result of this case, the sample code are as following:
```python
import torch

input_tensor = torch.randn((10,)).cuda()
target_tensor = torch.randint(0, 5, (10,)).cuda()

loss = F.nll_loss(input_tensor, target_tensor)

print("Final Loss:", loss.item())
```

### Versions

torch: 2.6.0.dev20241112+cu121

mansiag05:
Add a shape check in nll_loss_forward to error out when both input and target are 1D. Added a unit test to cover the incompatible 1D/1D case.

Fixes #157420

cc @albanD @ngimel @peterbell10 @cyyever @kurtamohler

### DIFF
--- a/aten/src/ATen/native/LossNLL.cpp
+++ b/aten/src/ATen/native/LossNLL.cpp
@@ -57,7 +57,7 @@ TORCH_META_FUNC(nll_loss_forward)
       target.sizes(),
       ")")
 
-  TORCH_CHECK(
+  TORCH_CHECK_VALUE(
         !(self.dim() == 1 && target.dim() == 1),
         "Expected target to be 0D when input is 1D, but got target.dim =", target.dim());
 
--- a/aten/src/ATen/native/LossNLL.cpp
+++ b/aten/src/ATen/native/LossNLL.cpp
@@ -47,19 +47,19 @@ TORCH_META_FUNC(nll_loss_forward)
   TORCH_CHECK(
       target.dim() <= 1,
       "0D or 1D target tensor expected, multi-target not supported");
-
-  auto no_batch_dim = self.dim() == 1  && target.dim() == 0;
+  if (self.dim() == 1 && target.dim() == 1) {
+      TORCH_CHECK_VALUE(
+          target.size(0) == 1,
+          "For 1D input, 1D target must have size 1, but got target size: ",
+          target.size(0));
+  }
   TORCH_CHECK(
-      no_batch_dim || (self.size(0) == target.size(0)),
+      self.dim() == 1 || (self.size(0) == target.size(0)),
       "size mismatch (got input: ",
       self.sizes(),
       ", target: ",
       target.sizes(),
       ")")
-
-  TORCH_CHECK_VALUE(
-        !(self.dim() == 1 && target.dim() == 1),
-        "Expected target to be 0D when input is 1D, but got target.dim =", target.dim());
 
   const auto n_classes = self.size(-1);
 
--- a/aten/src/ATen/native/LossNLL.cpp
+++ b/aten/src/ATen/native/LossNLL.cpp
@@ -57,6 +57,10 @@ TORCH_META_FUNC(nll_loss_forward)
       target.sizes(),
       ")")
 
+  TORCH_CHECK(
+        !(self.dim() == 1 && target.dim() == 1),
+        "Expected target to be 0D when input is 1D, but got target.dim =", target.dim());
+
   const auto n_classes = self.size(-1);
 
   TORCH_CHECK(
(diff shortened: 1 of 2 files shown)


aten/src/ATen/native/LossNLL.cpp
```diff
@@ -57,6 +57,10 @@ TORCH_META_FUNC(nll_loss_forward)
       target.sizes(),
       ")")
 
+  TORCH_CHECK(
```
Skylion007:
```suggestion
  TORCH_CHECK_VALUE(

File: test/test_nn.py [hunk 2]
```diff
@@ -12043,6 +12043,13 @@ def test_softmax_bfloat16(self, device):
             # test softmax with large input value which causes exp() to overflow
             _test_bfloat16_ops(self, torch.nn.Softmax(dim=dim), device, inp_dims=(16, 33, 15, 16), prec=0.05, scale_factor=1000.0)
 
+    def test_nll_loss_incompatible_1d_input_1d_target(self, device):
+        x = torch.randn((10,), device=device)  # 1D input: single sample, 10 classes
+        t = torch.zeros((10,), dtype=torch.int64, device=device)  # 1D target: 10 class indices
+
+        with self.assertRaisesRegex(RuntimeError, "Expected target to be 0D when input is 1D"):
```
Skylion007:
```suggestion
        with self.assertRaisesRegex(ValueError, "Expected target to be 0D when input is 1D"):

**Review by Skylion007 — DISMISSED** [review]

**Review by ngimel — CHANGES_REQUESTED** [review]
Please don't submit BC-breaking changes. When you think documentation contradicts current behavior, changes should be made to documentation, not behavior. In this case, there is logic in implementation that sets batch_size to 1 if input is 1d, and there was a specific check in the file you were changing https://github.com/pytorch/pytorch/pull/161412/files#diff-0b104a4612220d611b2c5d76272c28ddb21efbd8f9b9ae29d7f5b6c9e2a9be1fR53 to make sure inputs are valid

**Comment by mansiag05** [comment]
Hello all, thanks for the reviews and clarifications.
I wanted to follow up with an example to clarify the edge case we are discussing.

**Examples:**

_Valid case:_
`input = torch.randn(10)        # [C] -> Internally [1,C]
target = torch.tensor(3)       # scalar, class 3 is the correct label.
F.nll_loss(input, target)      # works`

- Conceptually, nll_loss expects input.shape = [N, C], target.shape = [N].
- When input is 1D ([C]), PyTorch internally treats it as a batch of size 1 ([1, C]). This means the target should be scalar (or [1]) to match the single batch sample.

_Invalid case:_
`# 1D input and 1D target of same length
input1 = torch.randn(10)                           # shape [10] -> internally [1,10]
target1 = torch.randint(0, 10, (10,), dtype=torch.long)  # shape [10]
F.nll_loss(input1, target1)`

- Here PyTorch currently does not raise an error, even though logically this is invalid: a batch of size 1 cannot have 10 labels.
- This seems to be a quirk in the current shape-checking logic: the implementation treats the 1D input as batch size 1 internally but does not fully enforce that the target must be scalar. So it accidentally accepts [10] as a target, even though it doesn’t align with the “batch size =1” interpretation.

**Next Step**

Given this, what would be the recommended approach for the PR?

1. Update the documentation to clearly explain that for 1D input, only scalar targets are valid?
2. Or additionally introduce a deprecation warning / stricter error check for the invalid 1D input + 1D target case?

Happy to adjust the PR based on guidance.

**Comment by ngimel** [comment]
neither. We could make the 1d-1d shape check correct, but we won't limit target to 1d, and we won'd deprecate.

File: aten/src/ATen/native/LossNLL.cpp [hunk 3]
```diff
@@ -57,6 +57,10 @@ TORCH_META_FUNC(nll_loss_forward)
       target.sizes(),
       ")")
 
+  TORCH_CHECK_VALUE(
+        !(self.dim() == 1 && target.dim() == 1),
```
ngimel: per discussion, today we support case where self.dim and target.dim are 1 and we will keep supporting it.

ngimel (approved):

t-vi:
PyTorch used to allow 1d size 0 target with 1d size 0 input, is it intentional that this PR blocks this? It's a corner case, just wondering. I guess part of it is that there is no 0 element 0 dim tensor, but that's just how dimensions work.

PR #161413, opened 2025-08-25
[ez][CI] GIve the linux check job a name that isn't linux-job

clee2000:
Reason:
The default name is linux-job, which gets put in the linux category on HUD, but this isn't really a linux related job.  Renaming it like this will make it go into the "other" category on HUD

Other options:
Change the grouping code in test-infra


huydhn (approved):
Good catch :)

seemethere (approved):

PR #161414, opened 2025-08-25
[dynamic shapes] unbacked-safe slicing

pianpwk:
Summary:
Generates new unbacked symbols for slice output size & storage offset, when appropriate semantics are unclear. Teaches inductor to codegen the slice with flexible semantics.

Test Plan:
contbuild & OSS CI, see https://hud.pytorch.org/commit/pytorch/pytorch/56218d85e2da09d9ede3809718ec989c2151632c

Rollback Plan:

Differential Revision: D80948073

### DIFF
--- a/torch/_inductor/lowering.py
+++ b/torch/_inductor/lowering.py
@@ -1201,9 +1201,119 @@ def permute(x, dims):
 
 @register_lowering(aten.slice, type_promotion_kind=None)
 def slice_(x, dim=0, start=0, end=2**63, step=1, clamp=True):
+    """
+    Lowers a slice call, creating ExternKernels for the output size & storage offset symbols,
+    if the indices are unbacked and appropriate semantics aren't known.
+    If they are known (indices are static/backed/unbacked with info), a SliceView is created.
+    """
+
+    from torch.fx.experimental.symbolic_shapes import (
+        CallMethodKey,
+        resolve_unbacked_bindings,
+    )
+
     assert isinstance(x, TensorBox)
     dim = _validate_dim(x, dim, 0)
-    return TensorBox(ir.SliceView.create(x.data, dim, start, end, step, clamp=clamp))
+    size = x.get_size()[dim]
+    step = sympy.expand(step)
+    assert isinstance(step, sympy.Expr) or step > 0, step
+
+    # maybe apply slice optimization
+    try:
+        if (
+            start == 0
+            and V.graph.sizevars.statically_known_leq(size, end)
+            and step == 1
+        ):
+            return x
+    except TypeError:
+        pass
+
+    # try to avoid dynamic (unbacked) slice
+    def compute_slice_index(index, size, default=None):
+        if index is None:
+            return default
+
+        fn = lambda x: V.graph.sizevars.guard_or_false(x)  # noqa: E731
+        index = sympy.expand(index)
+        size = sympy.expand(size)
+        if fn(sympy.Ge(index, 0)) and fn(sympy.Le(index, size)):
+            return index
+        elif fn(sympy.Lt(index, 0)) and fn(sympy.Ge(index, -size)):
+            return index + size
+        elif fn(sympy.Gt(index, size)):
+            return size
+        elif fn(sympy.Lt(index, -size)):
+            return 0
+        return None
+
+    start_index, end_index = None, None
+    ambiguous_slice = clamp
+    if ambiguous_slice:
+        start_index = compute_slice_index(start, size, 0)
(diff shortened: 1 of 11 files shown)


torch/_inductor/lowering.py
```diff
@@ -1172,9 +1172,110 @@ def permute(x, dims):
...
+                sym_storage = sym
+
+    if V.graph.current_node is None or not clamp or (sym_size is None and sym_storage is None):
+        return TensorBox(
+            ir.SliceView.create(x.data, dim, start, end, step, clamp=clamp)
+        )  # go to SliceView/ReinterpretView
+
+    # realize to get strides/storage offset
+    if x.maybe_get_layout() is None:
+        x.realize()
```
pianpwk: @eellison I'm not very familiar with inductor lowering, so I wasn't sure if this was the right approach; the goal was to call as_strided with codegened sizes/strides/offset, but later figured I had to call `realize()`, on inputs like `ir.Pointwise`.

maybe this case should be lowering in a different way?

torch/_subclasses/fake_impls.py
```diff
@@ -746,6 +747,88 @@ def _padded_dense_to_jagged_forward(fake_mode, func, padded, offsets, total_L=No
...
+
+    # size
+    new_size = None
+    if start_index is not None and end_index is not None:
+        if guard_or_false(end_index >= start_index):
+            new_size = (end_index - start_index + step - 1) // step
+        elif guard_or_false(start_index >= end_index):
+            new_size = 0
+
+    # create unbacked if case unknown
```
laithsakka, line 807: nit: unknown -> undecided due to unbacked shapes.

torch/_inductor/lowering.py
```diff
@@ -1172,9 +1172,110 @@ def permute(x, dims):
 
 @register_lowering(aten.slice, type_promotion_kind=None)
 def slice_(x, dim=0, start=0, end=2**63, step=1, clamp=True):
+    """
+    Lowers a slice call, creating ExternKernels for the output size & storage offset symbols,
+    if the indices are unbacked and appropriate semantics aren't known.
```
laithsakka, line 1206: nit: semantics are the same for known and unknown. 

maybe:
creating ExternKernels to compute the output size & storage offset symbols dynamically if they  cant be determined at compile time because indices are unbacked.

torch/_inductor/lowering.py
```diff
@@ -1172,9 +1172,110 @@ def permute(x, dims):
...
+    # maybe apply slice optimization
+    try:
+        if (
+            start == 0
+            and V.graph.sizevars.statically_known_leq(size, end)
+            and step == 1
+        ):
+            return x
+    except TypeError:
+        pass
```
laithsakka, line 1230: why would a type error be thrown? statically_known_leq should be resilient to unbacked.

torch/_inductor/lowering.py
```diff
@@ -1172,9 +1172,110 @@ def permute(x, dims):
...
+        assert len(unbacked_bindings) <= 2, unbacked_bindings
+        sym_size, sym_storage = None, None
+        for sym, keypath in unbacked_bindings.items():
+            if keypath == (CallMethodKey("size"), pytree.SequenceKey(dim)):
+                sym_size = sym
+            elif keypath == (CallMethodKey("storage_offset"),):
+                sym_storage = sym
+
+    if V.graph.current_node is None or not clamp or (sym_size is None and sym_storage is None):
+        return TensorBox(
```
laithsakka, line 1264: // This path is taken if no unbacked symbols has been allocated meaning that we know how to determine things statically and no need for dynamic computations.

torch/_inductor/lowering.py
```diff
@@ -1172,9 +1172,110 @@ def permute(x, dims):
...
+        if (
+            start == 0
+            and V.graph.sizevars.statically_known_leq(size, end)
+            and step == 1
+        ):
+            return x
+    except TypeError:
+        pass
+
+    # try to avoid dynamic (unbacked) slice
```
laithsakka, line 1232: this does not seem the appropriate line for this comment?

torch/_inductor/lowering.py
```diff
@@ -1172,9 +1172,110 @@ def permute(x, dims):
...
+    def compute_slice_index(index, size, default=None):
```
laithsakka, line 1233: this is only called once with default being 0? 
sounds like we never call it with index None  in this version?

laithsakka (approved):
approved but please address my comments before landing

PR #161420, opened 2025-08-25
[Inductor] Improve RoPE

BoyuanFeng:
This PR fuses ROPE from 2 kernels into 1 kernel.

Shape:
```
q: [B, Hq, S, D]
k: [B, Hkv, S, D]
```

`Hq=32, Hkv=8, D=128` following Llama3 setting.

image


torch/_inductor/config.py
```diff
@@ -762,6 +762,10 @@ def decide_worker_start_method() -> str:
 
 worker_start_method: str = decide_worker_start_method()
 
+# Threshold to decide if a kernel has small memory access in bytes
+# Default value is 16 MB which is arbitrarily selected.
+small_memory_access_threshold: int = 16777216
```
exclamaforte, line 773: could we run some rough benchmarks on this threshold for rope if you haven't? It would be good to know in general.

BoyuanFeng: Yes we can see the phase out
image

torch/_inductor/loop_body.py
```diff
@@ -223,6 +223,53 @@ def merge_loops(self) -> LoopBody:
...
+        )
+
+        def new_body(*indices: Sequence[sympy.Expr]) -> Any:
+            index = [*itertools.chain.from_iterable(indices)]
+            assert len(index) == len(iter_size) + len(reduce_size)
+            iter_idx = index[: len(iter_size)]
+            reduce_idx = index[len(iter_size) :]
+
+            new_iter_idx = list(iter_idx)
+            new_iter_idx[dimension] = iter_idx[dimension] % original_range
```
eellison, line 255: We are re-indexing and re-storing duplicate indices. Can you compare the performance masking instead ? i think that's the only reason you had to set a small_memory_access_threshold

eellison (commented):
nice to get perf speedup, but would you mind posting the generated code and comparing perf using masks instead ? we shouldn't need to do redundant reads and writes.

torch/_inductor/loop_body.py
BoyuanFeng: I tried to use ops.masked:
```python
        def new_body(*indices: Sequence[sympy.Expr]) -> Any:
            index = [*itertools.chain.from_iterable(indices)]
            assert len(index) == len(iter_size) + len(reduce_size)
            iter_idx = index[: len(iter_size)]
            reduce_idx = index[len(iter_size) :]

            mask = ops.lt(
                ops.index_expr(iter_idx[dimension], torch.int64),
                ops.index_expr(original_range, torch.int64),
            )
            return ops.masked(mask, lambda: old_body(iter_idx, reduce_idx), 0)
```
However, `lambda: old_body(iter_idx, reduce_idx)`'s graph looks like: 
```
    ...
    %store : [num_users=1] = call_method[target=store](args = (%ops, buf1, %get_index_7, %add, None), kwargs = {})
    return store
```

It returns store which is None. Triton codegen of `masked` requires the returned values of body to be NOT `None`. 
https://github.com/pytorch/pytorch/blob/0e2c8af5a660264e181b98f2c0bcae89535733ad/torch/_inductor/codegen/triton.py#L1538-L1544

So I used mod as a workaround for now. I can support None-return value of `masked` in the next PR.

BoyuanFeng: Comparing code before and after this pr: [diff](https://www.internalfb.com/intern/diffing/?paste_number=1924262733)

eellison: Are you against waiting to land until we have the masked version ? My concern is that the most important optimization inductor does is reduce read/writes to global memory. Here we are increasing it, with a config value that has been special cased to rope. we could also land with the config off by default. 

cc @shunting314 for other thoughts

shunting314: I'm wondering how cudagraphs could make any difference on the perf number here. From the perf results, it looks like the difference is at microsecond  level (?) which seems to match the launch overhead.

Also one thing to look out is compilation time increase. A dashboard run would be nice.

BoyuanFeng: @shunting314 all numbers are measured with cudagraph. More details [here](https://docs.google.com/document/d/1K_1PNWE37vwJmZ8nDkmbEbKoz7H5xuCAV0XAhbvV-Rw/edit?tab=t.0#heading=h.cxe4zgj8lpjf).

[Dashboard run](https://hud.pytorch.org/benchmark/compilers?dashboard=torchinductor&startTime=Wed%2C%2027%20Aug%202025%2018%3A24%3A44%20GMT&stopTime=Wed%2C%2003%20Sep%202025%2018%3A24%3A44%20GMT&granularity=hour&mode=inference&dtype=bfloat16&deviceName=cuda%20(h100)&lBranch=bf/rope&lCommit=89a05d2f60d27a10587f77a081d25c6d9614412e&rBranch=main&rCommit=c83cbd2f2a2de2e3258f07de77d8740743df6d2d) shows neutral on compilation time and speedup.
image

shunting314:
Sorry I still feel a bit skeptical here, can you clarify a  bit?
1. The main motivation here is to make sure the two kernels that applying rope to Q and K can be fused, right? But the common memory access for these 2 kernels is the frequency-tensor (sine/cosine etc) which would be broadcasted and is small (compared to Q/K). 
2. The benefit of less number of kernels can be mostly achieved by cudagraphs
3. expanding ir.Node shape in general sounds a bit tricky and need be careful to not hurt perf

shunting314 (approved):

shunting314:
Also make sure to address elias's comment above before tuning this on by default

ProExpertProg:
> The benefit of less number of kernels can be mostly achieved by cudagraphs

@shunting314 we have found that not to be the case in vLLM, and the extra kernel call is expensive.

PR #161328, opened 2025-08-23
Avoid double hash lookup in torch._library.simple_registry

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #161455
* #161432
* #161329
* __->__ #161328
* #161317
* #161315
* #161308
* #161304
* #161292
* #161301

Not a huge cost, but free win is free.

PR #161432, opened 2025-08-25
Stop trying to intern arguments in PyObject_FastGetAttrString

swolchok:
* #161455
* __->__ #161432
* #161329
* #161328
* #161317
* #161315
* #161308
* #161304
* #161292
* #161301

If we want them interned, we should intern at callsites. (The numpy reference has bit rotted; see https://github.com/numpy/numpy/commit/b222eb66c79b8eccba39f46f020ed8303614a87f#diff-6bdb6105198083838f51c57b55b3a49472ed23043bb40018f1ea41138e687163)

Profiling a simple torchdispatch benchmark with perf before/after seems to show that time spent copying std::strings and interning Python strings is gone, though there is some noise and the improvement is very small.


ezyang (approved):

swolchok:
couple macos-py3-arm64 failures with the build leaving third-party/kineto dirty. HUD seems to indicate these are flaking on main. normally I'd bypass but it's a long weekend so I'm rerunning

Issue #140457, opened 2024-11-12
Inductor inappropriately tries to fuse scalar views of a CPU tensor into GPU kernels.


Chillee:
### 🐛 Describe the bug

```
import torch

@torch.compile
def f(a, b):
    return a + b[0]

f(torch.randn(20, device='cuda'), torch.randn(4, device='cpu'))
```
```
    self.launch(*args, **kwargs)
ValueError: Pointer argument (at 1) cannot be accessed from Triton (cpu tensor?)
```

cc: @desert

### Versions

N/A

cc @chauhang @penguinwu @voznesenskym @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @aakhundov @coconutruben @ezyang @yf225 @ColinPeppler @desertfire

etaf:
Hi, @Chillee @eellison @shunting314 may I know anyone is looking into this issue?

PR #161447, opened 2025-08-25
[Inductor] Fix cross-device scalar lowering - cpu scalar with cuda tensor fails in torch.compile

karthickai:
* __->__ #161447

This PR fixes bug in TorchInductor where cross-device scalar indexing fails during compilation, causing discrepancies from eager mode behavior.

Fixes: #140457 


test/inductor/test_torchinductor.py
```diff
@@ -13833,6 +13833,74 @@ def fn(m, inp):
...
+    @requires_cuda_and_triton
+    def test_gpu_scalar_with_gpu_tensor(self):
+        def fn(a, b):
+            return a[0] + b
+
+        a = torch.rand(20, device=GPU_TYPE)
+        b = torch.rand(20, device=GPU_TYPE)
+
+        eager = fn(a, b)
+        compiled = torch.compile(fn, backend="inductor")(a, b)
```
mlazos: can you verify the number of compiled kernels is as expected? (1 for the GPU kernels, 2 for the cpu-gpu kernel etc.)

torch/_inductor/lowering.py
```diff
@@ -348,6 +352,27 @@ def transform_args(
             args[args_indices[0]] if args_indices else kwargs[kwargs_indices[0]]
         ).get_device()
 
+        def _is_copy_cpu_scalar(x: TensorBox) -> TensorBox:
```
mlazos: Can you rename this to `_maybe_copy_cpu_scalar` and perhaps extract it as a utility function?

mlazos: I also just thought of an alternative approach, could you just try realizing the scalar argument instead of the copy? This might work because then the scheduler won't fuse and we won't even need the copy.

mlazos: Yeah realizing won't work if it's an InputBuffer, we would still need the copy. This sounds alright with me.

torch/_inductor/lowering.py
```diff
@@ -348,6 +352,27 @@ def transform_args(
...
+            size = x.get_size()
+            cur_device = x.get_device()
+            if (
+                cur_device is not None
+                and cur_device.type == "cpu"
+                and cur_device != device
+                and isinstance(x.data, ir.ReinterpretView)
+                and (len(size) == 0 or (len(size) == 1 and size[0] == 1))
```
mlazos: Does this handle dynamic shapes though? You might need to check the size using V.graph.sizevars (search for that string elsewhere in the codebase)

torch/_inductor/lowering.py
karthickai: I made the changes, moved maybe_copy_cpu_scalar to a utility function.

torch/_inductor/lowering.py
karthickai: Thanks, updated to handle dynamic shapes and added a dynamic shape test case.

test/inductor/test_torchinductor.py
karthickai: Added a kernel count check in the test case, for gpu it generates 1 kernel and for cpu it also generates 1. 

```python

def fn(a, b):
    return a + b[0]

a = torch.rand(20, device="cpu")
b = torch.rand(4, device="cpu")

torch.compile(fn, backend="inductor")(a, b)
```
Generated output code:
```python
# AOT ID: ['0_inference']
from ctypes import c_void_p, c_long, c_int
import torch
import math
import random
import os
import tempfile
from math import inf, nan
from cmath import nanj
from torch._inductor.hooks import run_intermediate_hooks
from torch._inductor.utils import maybe_profile
from torch._inductor.codegen.memory_planning import _align as align
from torch import device, empty_strided
from torch._inductor.async_compile import AsyncCompile
from torch._inductor.select_algorithm import extern_kernels

aten = torch.ops.aten
inductor_ops = torch.ops.inductor
_quantized = torch.ops._quantized
assert_size_stride = torch._C._dynamo.guards.assert_size_stride
assert_alignment = torch._C._dynamo.guards.assert_alignment
empty_strided_cpu = torch._C._dynamo.guards._empty_strided_cpu
empty_strided_cpu_pinned = torch._C._dynamo.guards._empty_strided_cpu_pinned
empty_strided_cuda = torch._C._dynamo.guards._empty_strided_cuda
empty_strided_xpu = torch._C._dynamo.guards._empty_strided_xpu
empty_strided_mtia = torch._C._dynamo.guards._empty_strided_mtia
reinterpret_tensor = torch._C._dynamo.guards._reinterpret_tensor
alloc_from_pool = torch.ops.inductor._alloc_from_pool
async_compile = AsyncCompile()
empty_strided_p2p = torch._C._distributed_c10d._SymmetricMemory.empty_strided_p2p


cpp_fused_add_select_0 = async_compile.cpp_pybinding(['const float*', 'const float*', 'float*'], '''
#include <torch/csrc/inductor/cpp_prefix.h>
extern "C"  void  kernel(const float* in_ptr0,
                       const float* in_ptr1,
                       float* out_ptr0)
{
    {
        for(int64_t x0=static_cast<int64_t>(0L); x0<static_cast<int64_t>(20L); x0+=static_cast<int64_t>(16L))
        {
            {
                if(C10_LIKELY(x0 >= static_cast<int64_t>(0) && x0 < static_cast<int64_t>(16L)))
                {
                    auto tmp0 = at::vec::Vectorized<float>::loadu(in_ptr0 + static_cast<int64_t>(x0), static_cast<int64_t>(16));
                    auto tmp1 = in_ptr1[static_cast<int64_t>(0L)];
                    auto tmp2 = at::vec::Vectorized<float>(tmp1);
                    auto tmp3 = tmp0 + tmp2;
                    tmp3.store(out_ptr0 + static_cast<int64_t>(x0));
                }
                if(C10_UNLIKELY(x0 >= static_cast<int64_t>(16L) && x0 < static_cast<int64_t>(20L)))
                {
                    auto tmp0 = at::vec::Vectorized<float>::loadu(in_ptr0 + static_cast<int64_t>(x0), static_cast<int64_t>(4L));
                    auto tmp1 = in_ptr1[static_cast<int64_t>(0L)];
                    auto tmp2 = at::vec::Vectorized<float>(tmp1);
                    auto tmp3 = tmp0 + tmp2;
                    tmp3.store(out_ptr0 + static_cast<int64_t>(x0), static_cast<int64_t>(4L));
                }
            }
        }
    }
}
''')


async_compile.wait(globals())
del async_compile

class Runner:
    def __init__(self, partitions):
        self.partitions = partitions

    def recursively_apply_fns(self, fns):
        new_callables = []
        for fn, c in zip(fns, self.partitions):
            new_callables.append(fn(c))
        self.partitions = new_callables

    def call(self, args):
        arg0_1, arg1_1 = args
        args.clear()
        assert_size_stride(arg0_1, (4, ), (1, ))
        assert_size_stride(arg1_1, (20, ), (1, ))
        buf0 = empty_strided_cpu((20, ), (1, ), torch.float32)
        cpp_fused_add_select_0(arg1_1, arg0_1, buf0)
        del arg0_1
        del arg1_1
        return (buf0, )

runner = Runner(partitions=[])
call = runner.call
recursively_apply_fns = runner.recursively_apply_fns


def benchmark_compiled_module(times=10, repeat=10):
    from torch._dynamo.testing import rand_strided
    from torch._inductor.utils import print_performance
    arg0_1 = rand_strided((4, ), (1, ), device='cpu', dtype=torch.float32)
    arg1_1 = rand_strided((20, ), (1, ), device='cpu', dtype=torch.float32)
    fn = lambda: call([arg0_1, arg1_1])
    return print_performance(fn, times=times, repeat=repeat)


if __name__ == "__main__":
    from torch._inductor.wrapper_benchmark import compiled_module_main
    compiled_module_main('None', benchmark_compiled_module)

```

mlazos (approved):

PR #161452, opened 2025-08-25
inductor: Log the specific triton kernel that fails

c00w:
* #161561
* __->__ #161452

Added a optional name argument to SubprocPool.submit.

We record this in a dictionary, and when raising exceptions, add the name.
We manage the lifecycle the same as the pending futures.

Added a specific testcase to make sure this logs correctly.

### DIFF
--- a/torch/_inductor/async_compile.py
+++ b/torch/_inductor/async_compile.py
@@ -457,7 +457,7 @@ class AsyncCompile:
                 try:
                     kernel, elapsed_us = task.result()
                 except SubprocException as e:
-                    raise e.with_kernel_name(kernel_name) from e
+                    raise e.with_name(kernel_name) from e
 
                 # Now that we've compiled, we should clear the future
                 # so it can't be used again
--- a/torch/_inductor/async_compile.py
+++ b/torch/_inductor/async_compile.py
@@ -447,7 +447,7 @@ class AsyncCompile:
                 load_kernel,
                 extra_env,
                 extra_config,
-                **({name: kernel_name} if self.process_pool is SubprocPool else {}),
+                **({"name": kernel_name} if self.process_pool is SubprocPool else {}),
             )
 
             def get_result() -> CachingAutotuner:
--- a/torch/_inductor/async_compile.py
+++ b/torch/_inductor/async_compile.py
@@ -447,7 +447,7 @@ class AsyncCompile:
                 load_kernel,
                 extra_env,
                 extra_config,
-                **({name=kernel_name} if self.process_pool is SubprocPool else {}),
+                **({name: kernel_name} if self.process_pool is SubprocPool else {}),
             )
 
             def get_result() -> CachingAutotuner:
--- a/torch/_inductor/async_compile.py
+++ b/torch/_inductor/async_compile.py
@@ -38,7 +38,11 @@ from torch._inductor.codecache import (
     StaticAutotunerFuture,
     torch_key,
 )
-from torch._inductor.compile_worker.subproc_pool import AnyPool, SubprocPool
+from torch._inductor.compile_worker.subproc_pool import (
+    AnyPool,
+    SubprocException,
+    SubprocPool,
+)
 from torch._inductor.compile_worker.tracked_process_pool import (
     TrackedProcessPoolExecutor,
 )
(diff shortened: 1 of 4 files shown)


torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -188,7 +191,11 @@ def __init__(
         self.read_thread.start()
 
     def submit(
-        self, job_fn: Callable[_P, _T], *args: _P.args, **kwargs: _P.kwargs
+        self,
```
masnesral: The lint warning seems important?

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -198,6 +205,7 @@ def submit(
             job_id = next(self.job_id_count)
             self.pending_futures[job_id] = future = Future()
         future.set_running_or_notify_cancel()
+        self.names[job_id] = name
```
masnesral: Are updates to dictionaries atomic in Python? Seems like this should happen inside the lock? (it is below at least)

torch/_inductor/async_compile.py
```diff
@@ -447,6 +447,7 @@ def reload_kernel_in_parent():
                 load_kernel,
                 extra_env,
                 extra_config,
+                **({"name": kernel_name} if self.process_pool is SubprocPool else {}),
```
masnesral: I guess I'm struggling to have a strong opinion here, but an alternative implementation would be to catch the exception below in `get_result()` when calling `task.result()`, and augmenting the exception message with the kernel name (which we have there because of the closure). I probably would have done it that way only because it should be a lot fewer lines of code. But I don't know if it's actually better. I guess it works if the process pool is not a SubprocPool? Eh whatever.

masnesral (changes requested):
Question about a lint warning and lock usage

torch/_inductor/compile_worker/subproc_pool.py
c00w: No idea why that wasn't in the lock, fixed.

torch/_inductor/compile_worker/subproc_pool.py
c00w: It's wrong, but unexpressible,. Either way deleted all this code.

torch/_inductor/async_compile.py
c00w: I like your implementation better :) (In particular the initial implementation is unexpressible with mypy typing :( ).

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -90,8 +90,13 @@ class SubprocException(Exception):
...
     """
 
-    def __init__(self, details: str) -> None:
-        super().__init__(f"An exception occurred in a subprocess:\n\n{details}")
+    def __init__(self, details: str, name: str = "<unknown>") -> None:
+        super().__init__(
+            f"An exception occurred in a subprocess:\n\nName={name}\n{details}"
+        )
+
+    def with_name(self, name: str):
```
masnesral: Lint?

torch/_inductor/async_compile.py
```diff
@@ -450,7 +454,11 @@ def reload_kernel_in_parent():
             )
 
             def get_result() -> CachingAutotuner:
-                kernel, elapsed_us = task.result()
+                try:
+                    kernel, elapsed_us = task.result()
+                except SubprocException as e:
+                    raise e.with_kernel_name(kernel_name) from e
```
masnesral: This is named with_name below, right?

masnesral (changes requested):
LGTM, but I think you have a bug in the name. Also, did you lose the unit test on purpose? It seemed legit on the previous impl

torch/_inductor/compile_worker/subproc_pool.py
c00w: Fixed.,

torch/_inductor/async_compile.py
c00w: Fixed.

c00w:
Unittest was on SubprocPool - Wrote a equivalent one on AsyncCompile.

masnesral (approved):

PR #161454, opened 2025-08-26
[ONNX] Drop draft_export in exporter API

titaiwangms:
If onnx exporter fallbacks to draft_export with big models, this is taking forever for users, and possibly spam the printout, which keeps users from their stack trace with strict=False.

We could consider make another API for draft_export as debugging tool, or combine it with report=True when "model is small"?

cc @justinchuby


torch/onnx/_internal/exporter/_core.py
```diff
@@ -1340,8 +1340,6 @@ def export(
                 export_status.torch_export_non_strict = result.success
             elif strategy_class is _capture_strategies.TorchExportStrictStrategy:
                 export_status.torch_export_strict = result.success
-            elif strategy_class is _capture_strategies.TorchExportDraftExportStrategy:
```
justinchuby, line 1343: This is only report right? Do we want to update the logic as well?

titaiwangms: Oh I missed it.

justinchuby (approved):

PR #161455, opened 2025-08-26
PythonArgs::toBool: order cheap mutually exclusive checks first

swolchok:
* __->__ #161455
* #161432
* #161329
* #161328
* #161317
* #161315
* #161308
* #161304
* #161292
* #161301

symbools are not identical with Py_True or PyFalse, so we can do those cheap checks first and at least get plain old bools to go fast.


Skylion007 (approved):

torch/csrc/utils/python_arg_parser.h
```diff
@@ -1059,13 +1059,20 @@ inline double PythonArgs::toDouble(int i) {
...
 
 inline bool PythonArgs::toBool(int i) {
-  if (!args[i])
+  if (!args[i]) {
     return signature.params[i].default_bool;
+  }
+  if (args[i] == Py_True) {
+    return true;
+  }
+  if (args[i] == Py_False) {
```
swolchok, line 1068: I wonder if we should also check for Py_None here

Issue #161456, opened 2025-08-26
[dynamo][hop] speculate_subgraph doesn't work with subclass inputs


xmfan:
### 🐛 Describe the bug

We have explicit checks for FakeTensor

```python
RuntimeError: Process 0 exited with error code 10 and exception:
Traceback (most recent call last):
  File "/home/xmfan/core/a/pytorch/torch/testing/_internal/common_distributed.py", line 816, in run_test
    getattr(self, test_name)()
  File "/home/xmfan/core/a/pytorch/torch/testing/_internal/common_distributed.py", line 670, in wrapper
    fn()
  File "/home/xmfan/core/a/pytorch/torch/testing/_internal/common_utils.py", line 3224, in wrapper
    method(*args, **kwargs)
  File "/home/xmfan/core/a/pytorch/torch/testing/_internal/distributed/_tensor/common_dtensor.py", line 490, in wrapper
    raise e
  File "/home/xmfan/core/a/pytorch/torch/testing/_internal/distributed/_tensor/common_dtensor.py", line 487, in wrapper
    func(self, *args, **kwargs)  # type: ignore[misc]
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xmfan/core/a/pytorch/test/distributed/tensor/experimental/test_local_map.py", line 213, in test_local_map_in_placements
    Y_dt = local_mm_forward(X_dt, W_dt)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xmfan/core/a/pytorch/torch/_dynamo/eval_frame.py", line 804, in compile_wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/xmfan/core/a/pytorch/torch/_dynamo/convert_frame.py", line 1720, in __call__
    result = self._torchdynamo_orig_backend(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/xmfan/core/a/pytorch/torch/_dynamo/convert_frame.py", line 1473, in __call__
    result = self._inner_convert(
             ^^^^^^^^^^^^^^^^^^^^
  File "/home/xmfan/core/a/pytorch/torch/_dynamo/convert_frame.py", line 693, in __call__
    result = _compile(
             ^^^^^^^^^
  File "/home/xmfan/core/a/pytorch/torch/_dynamo/convert_frame.py", line 1280, in _compile
[...]

PR #161458, opened 2025-08-26
[dynamo][hop] Introduce Local Map HOP

xmfan:
* #163133
* #162702
* __->__ #161458

Can't actually deploy it because of: https://github.com/pytorch/pytorch/issues/161456


test/higher_order_ops/test_local_map.py
```diff
@@ -0,0 +1,123 @@
...
+    ),
+    redistribute_inputs=True,
+    in_grad_placements=None,
+    device_mesh=None,
+)
+def cp_decorated(query, key, value):
+    return context_parallel_attention(query, key, value)
+
+
+cp_function = local_map(
```
wconstab: dead code? (But i think the block above is the real deal and the test is ok)

test/higher_order_ops/test_local_map.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+def context_parallel_attention(query, key, value):
+    out = F.scaled_dot_product_attention(
+        query=query, key=key, value=value, is_causal=False
+    )
+    return out
+
+
+@local_map(
```
wconstab: so IIUC the overall UX for this is that users use local_map as usual, and it 'just works' with torch compile- but for the time being, due to speculate inputs issue, you have to specifically enable this `LocalMapWrappedHigherOrderVariable` ctx, right?

xmfan: yes, the code in this PR is to have dynamo trace the local map into a subgraph instead of into the main graph. this gives other components an easier time grabbing the nodes related to the local_map as a single unit (subgraph call, and the metadata lives on the call_function node targeting the subgraph).

if the inputs are DTensors we run into the speculate inputs issue, makes no sense to release this outside of AP rn

test/higher_order_ops/test_local_map.py
xmfan: im gonna test the dynamo mapping for both

torch/_dynamo/variables/higher_order_ops.py
```diff
@@ -3436,6 +3437,114 @@ def _call_function(
...
+
+    # Subclasses aren't supported by speculate_subgraph yet
+    # So this HOP is only usable with plain tensors
+    _enabled = False
+
+    @classmethod
+    @contextlib.contextmanager
+    def enable(cls):
+        """Context manager to temporarily enable local map wrapping.
+        Will be removed when speculate_subgraph supports subclass inputs.
```
ydwu4: nit: maybe link the issue?

torch/_higher_order_ops/local_map.py
```diff
@@ -0,0 +1,388 @@
...
+
+@local_map_hop.py_impl(FakeTensorMode)
+def fake_mode_key(
+    mode: FakeTensorMode,
+    fw_gm: GraphModule,
+    *args: Any,
+    **kwargs: Any,
+) -> tuple[torch.Tensor]:
+    with mode:
+        return fw_gm(*args, **kwargs)
```
ydwu4, line 273: Is it possible that fw_gm might create new unbacked symbols e.g. nonzero or .item()? If we want to support this case, can check how while_loop is done here.

torch/_higher_order_ops/local_map.py
```diff
@@ -0,0 +1,388 @@
...
+    fw_gm: GraphModule,
+    *args: Any,
+    **kwargs: Any,
+) -> tuple[torch.Tensor]:
+    return fw_gm(*args, **kwargs)
+
+
+# Running HOP in eager with real tensors
+@local_map_hop_backward.py_impl(torch._C.DispatchKey.CPU)
+@local_map_hop_backward.py_impl(torch._C.DispatchKey.CUDA)
```
ydwu4: Can Implement DispatchKey.CompositeExplicitAutograd key instead of cpu/cuda

torch/_higher_order_ops/local_map.py
```diff
@@ -0,0 +1,388 @@
...
+    @staticmethod
+    def backward(
+        ctx: Any, *_grads: tuple[torch.Tensor]
+    ) -> tuple[Optional[torch.Tensor], ...]:
+        saved_activations = saved_tensors_and_symints(ctx)
+        with torch._C._AutoDispatchBelowAutograd():
+            # Filter out grads that are None or do not require_grad.
+            # The AOTAutograd utils we rely on force this assumption.
+            grads = [_grads[i] for i in ctx.filtered_grads_idx]
+            grad_ins = local_map_hop_backward(ctx.bw_gm, *saved_activations, *grads)
```
ydwu4: It would be simpler to maintain if we implement the backward to be just local_map_hop instead of having a separate backward hop. The only difference seems to be we have a assertion for  _DEFER_INLINING instead of an inlining logic in autograd key, which seems ok to remove?

torch/_higher_order_ops/local_map.py
xmfan: Sounds good, we are in static shapes land for a bit

ydwu4 (approved):
the create_hop_fw_bw is a bit messy but we can clean it up later.

PR #161008, opened 2025-08-19
[SymmMem] Support memory pool (NVSHMEM and NCCL backends)

kwen2501:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #161090
* __->__ #161008

Hooking SymmetricMemory Allocator to MemPool so that user can create symmetric tensors with regular `torch.zeros`, `torch.arange` etc factories. Also so that our ops can have functional variants that create `out` tensors on symmetric memory.

To end users, this PR supports a python UI as follows:
```
allocator = symm_mem.get_mempool_allocator(device)
mempool = torch.cuda.MemPool(allocator)
with torch.cuda.use_mem_pool(mempool):
    tensor = torch.arange(numel, dtype=dtype, device=device)
```

Note: this feature does not support the `CUDASymmetricMemoryAllocator` yet bc this backend requires a `group` argument which is not provided by MemPool invocation. The other two allocators (`NVSHMEMSymmetricMemoryAllocator` and `NCCLSymmetricMemoryAllocator` do not require this arg).

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta

PR #161471, opened 2025-08-26
[2/N][SymmMem] Add MemPool allocator and tests

kwen2501:
* #161533
* #161532
* __->__ #161471

(Porting most of #161008)

Hooking SymmetricMemory Allocator to MemPool so that user can create symmetric tensors with regular `torch.zeros`, `torch.arange` etc factories. Also so that our ops can have functional variants that create `out` tensors on symmetric memory.

To end users, this PR supports a python UI as follows:
```
allocator = symm_mem.get_mempool_allocator(device)
mempool = torch.cuda.MemPool(allocator)
with torch.cuda.use_mem_pool(mempool):
    tensor = torch.arange(numel, dtype=dtype, device=device)
```

Added tests for both use cases above.

Differential Revision: [](https://our.internmc.facebook.com/intern/diff/)


test/distributed/test_nvshmem.py
```diff
@@ -65,6 +65,58 @@ def foo():
...
+
+        with torch.cuda.use_mem_pool(mempool):
+            x = x0 + self.rank
+            y = torch.mm(x, w)
+
+        # y should be a symm tensor
+        torch.ops.symm_mem.nvshmem_broadcast(y, group_name)
+        expected = torch.mm(x0, w)
+        self.assertEqual(y, expected)
+
```
ngimel, line 119: can you also add the test for offset functionality?

ngimel (commented):
I take it this should work only for nvshmem and nccl backends, can you make cuda backend error out if someone tries to use a mempool?

kwen2501:
@ngimel thanks. Added a `TORCH_CEHCK` now.

ngimel (approved):

PR #161508, opened 2025-08-26
[Dynamo] Fix weakref.proxy error when `torch.compile`

### LINKED ISSUE #159258: `torch.compile` crashes when accessing a `weakref.proxy` inside a compiled method
## Repro Script

```python
import torch
import torch.nn as nn
import weakref

class MyModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = nn.Linear(10, 10)
        self.trainer = None  # Will later hold a weakref.proxy object

    def forward(self, x):
        return self.linear(x)

    def log(self):
        # This line triggers a crash inside torch.compile
        if self.trainer.foo:
            print("OK")

class Trainer:
    def __init__(self):
        self.foo = True

def main():
    model = MyModel()
    trainer = Trainer()
    model.trainer = weakref.proxy(trainer)

    # Works in eager mode
    model.log()

    # Crashes in compiled mode
    compiled_log = torch.compile(model.log)
    compiled_log()

if __name__ == "__main__":
    main()
```

## Error Output (Truncated)
```
OK
E0728 torch/_guards.py: Error while creating guard:
  Name: "L['self'].trainer"
  Create Function: TYPE_MATCH
    File ".../torch/_guards.py", line XXX, in create
      return self.create_fn(builder, self)
    ...
    File ".../torch/_dynamo/guards.py", line 2276, in _set_guard_export_info
      obj_ref = weakref.ref(guarded_object)
TypeError: cannot create weak reference to 'weakref.ProxyType' object

  ...
torch._dynamo.exc.InternalTorchDynamoError:
  TypeError: cannot create weak reference to 'weakref.ProxyType' object
```

## My Analysis
+ The crash seems to come from TYPE_MATCH attempting to create a weak reference to an object that is already a weakref.ProxyType.

+ In eager mode, this is totally fine, and the code runs without any problems.

+ However, under torch.compile, it fails during guard construction.

### Versions

```
PyTorch version: 2.7.1a0+gite2d141d
Is debug build: True

CMake version: version 3.22.1
Libc version: glibc-2.35

Is CUDA available: True
```

can-gaa-hou:
Fixes #159258

The error occurs when we attempt to create a weak reference from a weak reference proxy.
https://github.com/pytorch/pytorch/blob/e9d42b3880dcdbd823bbdc9370c8b0b3af0ba2e3/torch/_dynamo/guards.py#L2910-L2915

In fact, we shouldn't create a weak reference from another reference or proxy, as it would check in CPython.
https://github.com/python/cpython/blob/f60f8225ed146a8f9b5fbf1eeed3474782127ea8/Objects/weakrefobject.c#L410-L418

However, `__weakrefoffset__` is not equal to **0** when the `guarded_object` is in `weakref.ProxyTypes`, and it will wrongly create a weak reference for the `weakref.ProxyTypes`. I think this could be a bug from CPython, but we can prevent it by adding more weakref type checks (`weakref.ProxyTypes` contains `weakref.ProxyType` and `weakref.CallableProxyType`) here.

### DIFF
--- a/test/dynamo/test_repros.py
+++ b/test/dynamo/test_repros.py
@@ -5006,6 +5006,27 @@ def forward(self, s77 : torch.SymInt, s27 : torch.SymInt, L_x_ : torch.Tensor):
         res = opt_fn(x_weak, weight, y)
         self.assertEqual(ref, res)
 
+    # https://github.com/pytorch/pytorch/issues/159258
+    def test_weakref_proxy(self):
+        class DummyTrainer:
+            def __init__(self, x):
+                self.foo = x
+
+        class DummyModel(nn.Module):
+            def __init__(self):
+                super().__init__()
+                self.trainer = None
+
+            def foo(self):
+                return self.trainer.foo
+
+        x = torch.randn(4)
+        model = DummyModel()
+        trainer = DummyTrainer(x)
+        model.trainer = weakref.proxy(trainer)
+        compiled_foo = torch.compile(model.foo, backend="eager", fullgraph=True)
+        self.assertEqual(compiled_foo(), x)
+
     def test_weakref_reconstruct(self):
         def fn(x_weak, weight, y):
             y = torch.sin(y)
--- a/torch/_dynamo/guards.py
+++ b/torch/_dynamo/guards.py
@@ -2912,7 +2912,9 @@ class GuardBuilder(GuardBuilderBase):
             getattr(guarded_object.__class__, "__weakrefoffset__", 0) != 0
         )
         # See D64140537 for why we are checking for tuple.
-        if supports_weakref and not isinstance(guarded_object, (enum.Enum, tuple)):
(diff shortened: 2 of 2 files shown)


Lucaskabela (changes requested):
Thanks for fixing! Before approving, can we add a test from the original issue showing this no longer repros? It can be added to 
`test/dynamo/test_repros.py` with a link to the issue :) Thank you and feel free to ping me once done

anijain2305 (changes requested):
Thanks for the fix. Please add a test in `test/dynamo/test_repros.py`. Rest LGTM.

Lucaskabela (approved):
Thanks for the changes and incorporating tests!

anijain2305 (approved):

malfet (approved):

Lucaskabela:
Seems a PR was recently reverted so should work if we rerun actually

PR #161531, opened 2025-08-26
Fix silent ImportError handling in torch._dynamo.utils to preserve critical error messages

haochengxia:
## Problem
Previously, the exception handling in `torch/_dynamo/utils.py` used a bare `except ImportError: pass` which silently suppressed **all** import errors, including critical system-level issues. This made debugging extremely difficult when encountering real import problems.

**Example of hidden error that was previously silenced:**
```
ImportError: /lib64/libstdc++.so.6: version `GLIBCXX_3.4.29' not found
```

This type of error indicates missing system dependencies (like incompatible glibc/libstdc++ versions) but was completely hidden from users, making it nearly impossible to diagnose environment issues.

## Root Cause
The original code treated all import failures the same way:
```python
except ImportError:
    pass  # This hides ALL import errors, including critical ones
```

This approach conflates two very different scenarios:
1. **Optional modules not available** (expected, should be handled gracefully)
2. **Critical import failures** (system issues, missing dependencies - should be reported)

## Solution
Implemented proper exception handling that distinguishes between different types of import failures:

```python
except ModuleNotFoundError as e:
    # Optional modules may not be available in all configurations
    logging.getLogger(__name__).debug("Optional module not found: %s", e)
except ImportError as e:
    # Re-raise critical import errors (e.g., missing dynamic libraries like cxxabi)
    logging.getLogger(__name__).error("Critical import error: %s", e)
    raise
```

## Key Improvements

1. **Preserves critical error information**: System-level import failures (missing .so files, incompatible library versions) are now properly reported with full error messages and tracebacks

2. **Follows PyTorch logging conventions**: 
   - Uses `logging.getLogger(__name__)` instead of `warnings.warn()`
   - Uses appropriate log levels (`debug` vs `error`)
   - Uses structured logging format with `%s` placeholders

3. **Better categorization**:
   - `ModuleNotFoundError`: Truly missing optional modules → logged at debug level
   - `ImportError`: Critical system issues → logged as error and re-raised

4. **Maintains backward compatibility**: Optional module handling still works as expected, but critical errors are no longer silenced

## Impact
- **Debugging**: Users will now see meaningful error messages for system-level import issues
- **Development**: Easier to diagnose environment setup problems
- **Production**: No unnecessary noise (optional module warnings only appear in debug mode)
- **Consistency**: Follows established PyTorch logging patterns used throughout the codebase

## Testing
The change maintains existing functionality for optional imports while ensuring critical import errors are properly propagated. This is a debugging/observability improvement that doesn't change the core logic.

### DIFF
--- a/torch/_dynamo/utils.py
+++ b/torch/_dynamo/utils.py
@@ -148,8 +148,13 @@ try:
 
         NP_TO_TNP_MODULE = {}
     from torch._subclasses.fake_tensor import FakeTensor, is_fake, maybe_get_fake_mode
-except ImportError:
-    pass
+except ModuleNotFoundError as e:
+    # Optional modules may not be available in all configurations
+    logging.getLogger(__name__).debug("Optional module not found: %s", e)
+except ImportError as e:
+    # Re-raise critical import errors (e.g., missing dynamic libraries like cxxabi)
+    logging.getLogger(__name__).error("Critical import error: %s", e)
+    raise
 
 
 T = TypeVar("T")


Lucaskabela:
I am not familiar with why this import pass was added - cc @xuzhao9  is there any additional context here?

Logging is a good option but not sure if reraising will break something or not so hoping @xuzhao9  has the answer :)

PR #161533, opened 2025-08-26
[4/N][SymmMem] Add `get_remote_tensor` + move up `get_buffer` and `get_signal_pad`

kwen2501:

`get_remote_tensor `: return a symmetric tensor given a peer rank.

The difference between `get_buffer` API and `get_remote_tensor` API:
- the former accepts an offset, whereas the latter doesn't
- the latter returns a symmetric tensor at `hdl.offset` on `peer`.

As a refactorization, this PR also moves the implementation of `get_buffer` and `get_signal_pad` to the `SymmetricMemory` level as their code is common to all backends.

### DIFF
--- a/torch/csrc/distributed/c10d/symm_mem/NVSHMEMSymmetricMemory.cu
+++ b/torch/csrc/distributed/c10d/symm_mem/NVSHMEMSymmetricMemory.cu
@@ -204,6 +204,9 @@ class NVSHMEMSymmetricMemory : public SymmetricMemory {
       c10::IntArrayRef sizes,
       c10::ScalarType dtype,
       int64_t storage_offset) override {
+    auto peer_ptr = pai_->buffers_[rank];
+    TORCH_CHECK(peer_ptr != nullptr,
+        "Cannot get buffer across nodes, my rank: ", get_rank(), ", peer: ", rank);
     // TODO: deduplicate
     const size_t numel = std::accumulate(
         sizes.begin(),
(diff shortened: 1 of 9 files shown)


torch/csrc/distributed/c10d/symm_mem/NVSHMEMSymmetricMemory.cu
```diff
@@ -229,6 +232,23 @@ class NVSHMEMSymmetricMemory : public SymmetricMemory {
...
 
+  at::Tensor get_remote_tensor(
+      int peer,
+      c10::IntArrayRef sizes,
+      c10::ScalarType dtype) override {
+    auto peer_ptr = pai_->buffers_[peer];
+    TORCH_CHECK(peer_ptr != nullptr,
+        "Cannot get peer tensor across nodes, my rank: ", get_rank(), ", peer: ", peer);
+    // Peer tensor's offset is the same as my offset
+    auto data_ptr = reinterpret_cast<uint8_t*>(peer_ptr) + offset_;
```
ngimel: you are not checking that you have enough space in the buffer, like get_buffer does. Would make sense to have one function that checks this, into which both get_buffer and get_remote_tensor will call

kwen2501: Do you think handle should remember tensor shape and dtype? If so I can remove the `size` argument from this API, returning a tensor of exactly same shape as the tensor that the handle comes from.

kwen2501: Never mind. Perhaps the local tensor will experience some `view` ops which will changes its shape. In this case, the user may want to provide the new shape to get a corresponding tensor directly.

kwen2501:
@ngimel I did a refactor so that both methods call into a `_get_buffer_at_byte_offset` util (which checks everything)

ngimel (commented):
Now `get_buffer` function is exactly the same for all backends - can we move it to parent class, and add `get_remote_tensor` to parent class also? No need to limit it to NVShmem

ngimel (approved):

PR #161534, opened 2025-08-26
[inductor][addmm] incorporate into new get_mm_configs properly

coconutruben:
* __->__ #161534

# why

- addmm aten running with an expanded version of bias vs the regular
  bias sometimes causes numerics differences
- to avoid this for now, we make addmm aten use inp vs inp_expanded
  depending on if we're in max-autotune or not, matching the previous
  logic

# what

- pass unexpanded bias (inp)
- let template (heuristics) that it to be expanded (ATen in not max-autotune, Triton always) expand it

# testing

```
python3 -bb -m pytest test/inductor/test_aot_inductor.py::AOTInductorTestABICompatibleCpu::test_aoti_debug_printer_codegen_cpu
```

Differential Revision: [D81520581](https://our.internmc.facebook.com/intern/diff/D81520581)


coconutruben:
@coconutruben has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81520581).

coconutruben:
@coconutruben has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81520581).

eellison (commented):
just for my understanding - how is this different than adjust_kernel_inputs? adjust_kernel_inputs should only be applying views, right ?

coconutruben:
@coconutruben has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81520581).

coconutruben:
@coconutruben has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81520581).

coconutruben:
@coconutruben has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81520581).

coconutruben:
@coconutruben has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81520581).

jansel (approved):

torch/_inductor/template_heuristics/aten.py
```diff
@@ -64,9 +64,61 @@ def get_extra_kwargs(
...
+        # through a different decomposition
+        max_autotune = inductor_config.max_autotune or inductor_config.max_autotune_gemm
+        assert isinstance(kernel_inputs, MMKernelInputs)
+        if not max_autotune:
+            nodes = kernel_inputs.nodes()
+            bias = nodes[0]
+            from ..ir import as_storage_and_layout, ReinterpretView
+
+            # remove the expansion from the bias
+            bias, old_layout = as_storage_and_layout(bias)
```
jansel: I think you need to guard this with `is_storage_and_layout`.   ExternKernel.realize_input supports all things.

jansel (changes requested):
Failing tests?

torch/_inductor/template_heuristics/aten.py
coconutruben: @jansel can you take a look at the implementation now inside lowering? I want something that undoes what `realize(expand(x))` does or said differently, I want `compress(realize(expand(x)))` to be equivalent to `realize(x)`. does what I built achieve that?

coconutruben: we can disregard this. The whole approach is wrong, you cannot remove the expansion, because you don't know if you (the mm lowering) added the expansion or it was added elsewhere. The right thing to do is to not expand it, and let all the templates/implementations that need to expand it, expand it. I'm closing out this comment, let's take a look at the new implementation where we pass an unexpanded bias, and the configurations that need the expanded bias expand it themselves

PR #161535, opened 2025-08-26
Fix dev container configuration

pbezglasny:
Fix the configuration of dev container:
- update version of CUDA to match with NVIDIA repo
- Set working python to venv, set owner of the venv folder to vscode user(now there is an issue that vscode user cannot install build packages to system interpreter)

Tested working with CPU and CUDA environments with a built dev container image from scratch.


wdvr (approved):

drisspg (approved):
@pytorchbot merge

Issue #161167, opened 2025-08-21
[RFC] Deprecate Python 3.9 in CI/CD


atalman:
### 🐛 Describe the bug

Similar to https://github.com/pytorch/pytorch/issues/120718

After addition of Python 3.14 to CD, CI to be completed.

We want to start process for deprecation of Python 3.9 from Pytorch CI and nightly builds. 
End of Life for Python 3.9 is October 2025 Please see: https://devguide.python.org/versions/
According to our policy: According to our policy: https://github.com/pytorch/rfcs/blob/master/RFC-0038-cpython-support.md : 
- Deprecate during the release process of the last PyTorch release before the oldest version of python goes EOL.

 Proposal is to Deprecate Python 3.9 for Release 2.10. hence deprecation plan is as follows:

### Update Plan

| Feature | Target Date |
| --- | --- |
| Deprecate Python 3.9 from nightly CD | Sept  2025 |
| Deprecate Python 3.9 from nightly CI  | Sept/Oct 2025 |
| Release torch 2.9 with Python 3.9 support | Oct 2025 | <- Here we should announce deprecation of Python 3.9 for next release
| Release torch 2.10 without Python 3.9 support | TBD |

cc @malfet @seemethere @pytorch/pytorch-dev-infra @albanD @ezyang

### Versions

2.10

atalman:
Revised, to use Option 2 as our option for Release 2.9

PR #161536, opened 2025-08-26
[BE] Consolidate inductor benchmark Docker images and rename jobs

huydhn:
We have 4 different version of inductor benchmark Docker images used in CI at the moment:

1. `pytorch-linux-jammy-cuda12.8-cudnn9-py3-gcc9-inductor-benchmarks` is used by almost all inductor jobs including nightly benchmark
2. `pytorch-linux-jammy-cuda12.8-cudnn9-py3.12-gcc9-inductor-benchmarks` runs inductor unit tests with python 3.12
3. `pytorch-linux-jammy-cuda12.8-cudnn9-py3.13-gcc9-inductor-benchmarks` runs inductor unit tests with python 3.13
4. `pytorch-linux-jammy-py3-gcc11-inductor-benchmarks` runs inductor unit tests on CPU

My proposal here is to clean up (2) and (3) and to keep (1) under the same setup from https://ghcr.io/pytorch/torchbench.  Simplicity is the key here as inductor workflows are getting more and more complex:
1. Unit tests for Python variant like 3.12 and 3.13 were useful when they were first added to CI.  They are much less useful now.  [Flambeau](https://hud.pytorch.org/flambeau/s/3876ec7b-43f0-42c6-bfbf-899035e5bb77) shows a 0.97 correlation between them.  And we are also moving to 3.14 nowadays.  I want to choose 3.12 for (1), but will do this separately.  This is also what TorchBench and vLLM are using on CI.
1. We are gradually cleaning up 3.9 on CI https://github.com/pytorch/pytorch/issues/161167

Another BE change here is to rename the jobs various inductor workflows because I think names like `linux-jammy-cuda12_8-py3_10-gcc9-inductor-build` is too long and confusing to look at, better just use human-friendly names like `inductor-build`.  Other information is already spelled out in the build environment.


zou3519:
> My proposal here is to clean up (2) and (3) and to keep (1) under the same setup from https://ghcr.io/pytorch/torchbench. Simplicity is the key here as inductor workflows are getting more and more complex:

To clarify: does "clean up (2) and (3)" mean "delete (2) and (3)", or "deduplicate (2) and (3)", or something else?

zou3519:
> It's dedup (2) and (3), we only need one here according to the correlation data I see. Do we want to maintain 2 variants of inductor unit tests job for 3.12 and 3.13?

imo inductor benchmark tests should just run on one version of python, if the thing we are testing is Inductor performance then the python version doesn't matter. Will get feedback from the team too.

zou3519 (approved):

PR #160888, opened 2025-08-18
Allow branching on whether you are in AC region with is_checkpoint_enabled

soulitzer:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162434
* #160883
* __->__ #160888

Also test with SAC?

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela

PR #161541, opened 2025-08-26
Support Triton kernels in SAC region

soulitzer:
* #160883
* #160888
* __->__ #161541

SAC interaction with triton kernel:
- In eager, triton ops are not dispatchable, and so it is always ignored by SAC,  i.e., always recomputed.
- In compile, although we wrap triton kernels into HOPs, allowing us to intercept them, we still recompute by default rather than save by default, so that compile maintains the invariant of using less memory than eager.
- If you want to do something else (e.g. save the output of your triton kernel) you should wrap it in a custom op.

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela @fegin


torch/_higher_order_ops/triton_kernel_wrap.py
```diff
@@ -1342,6 +1344,9 @@ def triton_kernel_wrapper_functional_functionalize(
 triton_kernel_wrapper_functional.fallthrough(DispatchKey.AutogradCUDA)
 triton_kernel_wrapper_functional.fallthrough(DispatchKey.AutogradCPU)
```
soulitzer, line 1345: I wonder when the functional HOP is used over mutation HOP? Should we register to both?

drisspg (approved):

torch/_higher_order_ops/triton_kernel_wrap.py
ydwu4: after functionalization key, the mutation hop becomes functional hop. The functional hop is not user-facing.

zou3519 (approved):

xmfan (approved):

PR #161449, opened 2025-08-25
[ONNX] Remove unused logic from internal verification module

justinchuby:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #161546
* __->__ #161449
* #161323

Signed-off-by: Justin Chu <justinchuby@users.noreply.github.com>

cc @titaiwangms

PR #161546, opened 2025-08-26
[ONNX] Remove private members from torch.onnx

justinchuby:
* __->__ #161546
* #161449
* #161323

Remove import of two functions

- _run_symbolic_function
- _run_symbolic_method

to the `torch.onnx` namespace.

Signed-off-by: Justin Chu <justinchuby@users.noreply.github.com>

cc @titaiwangms


torch/csrc/jit/passes/onnx.cpp
```diff
@@ -260,7 +260,8 @@ void NodeToONNX(
     ::torch::onnx::OperatorExportTypes operator_export_type,
     py::dict& env,
     py::set& values_in_env) {
-  py::object onnx = py::module::import("torch.onnx");
+  py::object onnx_utils =
+      py::module::import("torch.onnx._internal.torchscript_exporter.utils");
```
titaiwangms, line 264: Curious why this is a necessary change? We know torchscript is the only exporter consuming this.

titaiwangms (approved):

torch/csrc/jit/passes/onnx.cpp
justinchuby: This is because the utils module in `torch.onnx` is deprecated and is merely an alias of the internal module, not advertised to users.

PR #161547, opened 2025-08-26
[Flop Counter] Support tracing a decomposed custom triton kernel in torch op

Lucaskabela:
Add a new API for registering triton kernels, and provide an example of how to escape hatch triton op to leverage this API

Context: For some custom `torch.library.op`, the operator decomposes into one or more custom kernels.  To date, there has not been a good system for having these counted by FlopCounterMode, as the function will decompose to the wrapped kernels, and then check the registry.

To support this use case, we add a custom API, `register_flop_formula_for_triton_kernel` and demonstrate how it is used with `register_torch_dispatch`.  Currently, this is done manually, but as a followup, we will support inbuilt support for this.

### Unit Test
```
python test/test_flop_counter.py TestFlopCounter.test_flop_counter_custom_triton
```

```
.
----------------------------------------------------------------------
Ran 1 test in 0.005s
```


test_flop.py
```diff
@@ -0,0 +1,63 @@
+import torch
+from torch.utils.flop_counter import (
+    _FlopCounterMode,
+    FlopCounterMode,
+    register_flop_formula_for_triton_kernel,
+)
+
```
zou3519: tests go into the test/ folder and should use unittest

torch/utils/flop_counter.py
```diff
@@ -54,6 +54,60 @@ def register(target):
...
     return register_fun
 
+
+triton_flop_registry: dict[Any, Any] = {}
+def _register_flop_formula_for_triton_kernel(targets, get_raw=False) -> Callable[[Callable[_P, _T]], Callable[_P, _T]]:
+    """
+    This experimental API is used to register flop formula for triton kernel in `targets`
+
+    This is separate from `register_flop_formula` because triton kernel
+    currently gets decomposed into `triton_kernel_wrapper_functional(kernel)`
```
zou3519, line 64: Any thoughts on combining the two APIs? (@bdhirsh @davidberard98 ?)

torch/utils/flop_counter.py
```diff
@@ -782,15 +841,14 @@ def _execute_with_isolated_flop_counting(self, branch_fn, operands):
...
-
-        # The flop counter for cond counts the upper bound of flops.
-        # For example, if a matmul is executed 2 times in true branch
-        # but only 1 time in the false branch, the flop counter will
-        # record the larger number of flops, i.e. 2 times.
-        if func is torch.ops.higher_order.cond:
-
+        if func is torch.ops.higher_order.triton_kernel_wrapper_mutation:
+            # Special case - look in the triton flop registry for the kernel
+            return self.counter._count_flops(kwargs["kernel"], None, args, kwargs)
```
zou3519: We should return this for both triton_kernel_wrapper_mutation and triton_kernel_wrapper_functional

torch/utils/flop_counter.py
zou3519: Separate seems fine for now if we're hesitating

torch/_higher_order_ops/triton_kernel_wrap.py
```diff
@@ -976,8 +976,10 @@ def __call__(
         grid: list["TritonGridType"],
         tma_descriptor_metadata: TMADescriptorMetadata,
         kwargs: dict[str, Any],
+        kernel: Optional["TritonKernelType"] = None,
     ) -> Any:
         return super().__call__(
+            kernel=kernel,
```
zou3519: You don't need to pass in `kernel`, you can find the kernel via looking it up via `kernel_idx` in the lookup table

zou3519 (commented):
seems reasonable. Main feedback is that we shouldn't need to add an extra `kernel` argument to triton_kernel_wrapper_mutation

PR #161559, opened 2025-08-26
[dynamic shapes] remove _maybe_guard_rel warnings

pianpwk:
Fixes #161051

stops warnings arising from `torch._check(sym_or(...))` calls

The issue happens when:
- SDPA calls expand
- expand calls this check, signifying either expand or no-op semantics
https://github.com/pytorch/pytorch/blob/47ecd2042f2f99368cd2d60731d23f4437eecd1d/torch/_refs/__init__.py#L3051-L3052
- The `sym_or` is passed to `_maybe_guard_rel` to process any upper/lower bound implications for axioms we know are True
- `_maybe_guard_rel` was extended to handle `And` expressions as well as relations, so users could write `torch._check(sym_and(a >= 0, a <= N, ...))` easily. For other types, we raised this warning:
https://github.com/pytorch/pytorch/blob/47ecd2042f2f99368cd2d60731d23f4437eecd1d/torch/fx/experimental/symbolic_shapes.py#L6845-L6852

The change only applies `_maybe_guard_rel` to `And` and relation expressions, not calling it for the `sym_or` case

### DIFF
--- a/torch/fx/experimental/symbolic_shapes.py
+++ b/torch/fx/experimental/symbolic_shapes.py
@@ -6846,11 +6846,8 @@ class ShapeEnv:
             for arg in expr.args:
                 self._maybe_guard_rel(arg)
             return
-        elif not isinstance(expr, sympy.Rel):
-            log.warning(
-                "_maybe_guard_rel() was called on non-relation expression %s", expr
-            )
-            return
+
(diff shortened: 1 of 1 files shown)


ezyang (changes requested):
I am not buying it. What if I have an And which contains non Rel insides

PR #161564, opened 2025-08-26
fold to mm when row dimentions is small

yushangdi:
Summary:
Fix https://github.com/pytorch/pytorch/issues/159346

When we have a matmul like [M, B, K].transpose(0, 1) x [K, N], we should fold when B is small, and M, K, N are large. This could happen when we use MultiheadAttention with batch_first=True.

Folding to mm can speed up`F.linear(x, weight, bias)` by a lot.

Some benchmark numbers (all without max-autotune). If we use max-autotune, I think it'll only give folding to mm more benefits, because bmm triton template is limited to int32 indexing, so we have to fall back to aten bmm when BSxMxNxK is large.

Float16:

| BS |  M   |   N   |   K   | time_fold | time_no_fold | speedup  |
|  4 | 512  | 8192  | 8192  |   1.274   |    41.216    | 32.3516  |
| 64 | 512  | 2048  | 2048  |   1.618   |    1.751     | 1.0822   |
|  4 | 1024 | 2048  | 2048  |   0.576   |    2.253     | 3.9115   |

Float32:

| BS |  M   |   N   |   K   | time_fold | time_no_fold | speedup  |
|  4 | 512  | 8192  | 2048  |   4.059   |    22.967    | 5.6583   |
| 64 | 512  | 2048  | 2048  |  15.763   |    29.486    | 1.8706   |
|  4 | 1024 | 2048  | 2048  |   2.272   |    9.124     | 4.0158   |

You can find more benchmarking results in https://docs.google.com/spreadsheets/d/1d79g3CCE9Tn59JaAbxiNVYfpRxNyVDbYeWGVurqOses/edit?gid=1040716098#gid=1040716098.

In this PR, I set the numbers to be quite conservative, i.e. we only fold to mm if we're sure we can get performance gain from folding. This is to prevent any perf regression. But I think we could be more aggressive in our folding decision based on the benchmarking result.

Another potential work is to enable folding when the batch size is a dynamic symbol. Currently, we only fold when all shape sizes are integers. This means it's hard to take advantage of this when shape size are symints.

```
        # shape_env = s.node.shape_env
        # if shape_env is None:
        #     return False
        # return shape_env.var_to_range[s.node.expr].issubset(ValueRanges(0, 32))
```

Finally, note that in this implementation, we avoid any copy by using transpose. However, this means that the stride of the output after decomposition could be different from the output of the original op.

Test Plan:
```
buck run mode/dev-nosan fbcode//caffe2/test/inductor:test_inductor_cuda -- -r  test_folding_bmm_to_mm
```

Rollback Plan:

Differential Revision: D80385541

### DIFF
--- a/test/inductor/test_torchinductor.py
+++ b/test/inductor/test_torchinductor.py
@@ -13879,6 +13879,28 @@ def forward(self, arg0_1: "Sym(s77)", arg1_1: "Sym(s27)", arg2_1: "Sym(s53)", ar
         inp = torch.randn(100, 100, device=self.device)
         self.assertTrue(CommonTemplate._is_triggering_buffer_reuse(fn, m, inp))
 
+    
+    def test_folding_bmm_to_mm(self):
+        @torch.compile
+        def linear(x, weight, bias):
+            return F.linear(x, weight, bias) 
+
+
+        weight = torch.randn(2048, 2048, device=self.device)
+        bias = torch.randn(2048, device=self.device)
+
+        bs = 2
+        # we do a transpose here to make a non-contiguous. This can happen in nn.MultiheadAttention when batch_size_first=True
+        x = torch.randn(bs, 512, 2048, device=self.device).transpose(0, 1) # -> (512, bs, 2048)
+
+        _result, code = run_and_get_code(linear, x, weight, bias)
+        res = F.linear(x, weight, bias) 
+
+        # since bs is small, we should fold bmm to mm for better performance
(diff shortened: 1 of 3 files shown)


ngimel:
this is perf problem in eager too, so can't we fix both?

eellison (commented):
clearing up review queue - this should either be a change to eager as well, or a joint graph/post graph change so that eager & compile have consistent striding.

PR #161565, opened 2025-08-26
[1/2]Add summary report for vllm build

yangw-dev:
* #161585
* __->__ #161565

Demo Run
https://github.com/pytorch/pytorch/actions/runs/17259533323?pr=161565

image


.ci/lumen_cli/cli/lib/common/gh_summary.py
```diff
@@ -0,0 +1,229 @@
+from __future__ import annotations
```
huydhn, line 1: Have you tried to explore the option to use a jinja template instead?  The logic from here could be simplified with it, and the report template is also easier to maintain if it is getting more complex

.ci/lumen_cli/cli/lib/core/vllm/vllm_build.py
```diff
@@ -153,18 +159,50 @@ def run(self):
...
+        write_gh_step_summary(
+            f"**Vllm Commit**: [{vllm_commit}](https://github.com/vllm-project/vllm/commit/{vllm_commit})"
+        )
+        torch_sha = os.getenv("GITHUB_SHA")
+
+        if torch_sha:  # only can grab this in github action
+            write_gh_step_summary(
+                f"**Pytorch Commit**: [{torch_sha}](https://github.com/pytorch/pytorch/commit/{torch_sha})]"
+            )
+        vllm_artifact_dir = inputs.output_dir / "wheels"
```
huydhn, line 188: I think we could see if the output from https://github.com/vllm-project/vllm/blob/main/vllm/collect_env.py could be used here.  People are familiar with its output

.ci/lumen_cli/cli/lib/common/gh_summary.py
yangw-dev: the current use case is ok now, I try to avoid jinja since it can be hard to maintain eventually.
let's revisit it if we have more complicated use cases, I can see we can have some template for this,
but right now let's just write the markdown

.ci/lumen_cli/cli/lib/core/vllm/vllm_build.py
yangw-dev: that can be a followup!
The main thing i want to output here are just those packages and make sure torch versions are not overrwrite

.ci/lumen_cli/cli/lib/common/gh_summary.py
yangw-dev: actually second thoughts, i think i can use jinja

.ci/lumen_cli/cli/lib/common/gh_summary.py
```diff
@@ -0,0 +1,143 @@
...
+) -> bool:
+    f = Path(output_dir) / freeze_file
+    if not f.exists():
+        return False
+    content = f.read_text(encoding="utf-8").strip()
+    md = render_content(content, title=title, lang=code_lang)
+    return write_gh_step_summary(md)
+
+
+def summarize_wheels(path: Path, title: str = "Wheels", max_depth: int = 3):
```
yangw-dev, line 107: @huydhn  thx for the suggestion of ninja, altough i can put everything in a big summary template, i want to make those as module, and people can reuse for most cases, and generate their own template if it's complicated

huydhn (approved):

PR #161572, opened 2025-08-26
Update Kineto submodule

sraikund16:
Differential Revision: D81087601


cyyever (approved):

cyyever:
@sraikund16 To fix the error we could use `auto log_event = [=, this]() {` in CuptiActivityProfiler.cpp

sraikund16:
> @sraikund16 To fix the error we could use `auto log_event = [=, this]() {` in CuptiActivityProfiler.cpp

I just changed it to explicitly capture instead of using `=`

sraikund16:
@cyyever looks like it is failing again for the same linker issue. Do we need https://github.com/pytorch/kineto/pull/1131 to fix it?

aaronenyeshi (approved):

Issue #159906, opened 2025-08-05
New at::HostAllocator interface prevents using more than one allocator implementation for a device type


galv:
I'm not happy with the changes in #151431, @guangyey . The primary problem is that your current registration API  prevents using more than one allocator for a particular device.

While you can use the "priority" field to override an existing allocator, there isn't a way to "pop off the stack" the current allocator to begin using the previous allocator again. I am trying to rebase my work in https://github.com/pytorch/pytorch/pull/146924, which originally used a separate subclass from the normal CUDACachingHostAllocator. This is much cleaner than putting every new functionality we want into the same original class.

I am going to try to rework your design to allow more than one allocator for a given device type. The ability to use pluggable allocators (like we can for GPU allocations today) is necessary for other future work like the ability to allocate host memory with special properties like numa locality anyway. If you have any comments, let me know.

galv:
There are three situations I've encountered where we want to use different allocators on the fly:

1. cudaMallocFromPoolAsync() now supports host allocations as of CUDA 13.0 I believe, so we can now use that for host memory allocations, in the same way that there is a cudaMallocAsync allocator in pytorch.
2. We need to use a different allocator implementation for stream capture, because detection of whether a block has been freed must use a fundamentally different algorithm than querying cudaEvents. That work was started in #146924
3. Allocations with particular NUMA affinity cannot be done with cudaHostAlloc https://github.com/pytorch/pytorch/issues/160420

It's pretty clear to me that there is a need for the interface to change to allow for more than one caching host allocator within a single program. @guangyey I will be making a PR shortly with a new version of #146924 rebased on top of your recent changes.

In it, you will see that I use an implementation pattern where the singleton caching host allocator has pointers to many separate implementations, and the user can choose which implementation to use.

guangyey:
I don’t think that supporting multiple allocators and having a unified allocator interface are mutually exclusive.
Can’t we have multiple allocators, such as “native” and “graph,” that both inherit from HostAllocator? And the user can choose which implementation to use.

PR #158352, opened 2025-07-15
[CUDA] Reuse blocks with record_stream during CUDA Graph capture in the CUDACachingAllocator

eee4017:
## Introduction

During CUDA Graph capture, the CUDA caching allocator currently defers reclaiming blocks until capture ends. This is because CUDA forbids querying events recorded during capture (the CUDA operation is not executed during the capture stage), so the allocator cannot use its normal event-based logic. However, capture records an DAG (we call it **capturing graph**) of work. We can use the capturing graph to determine when a block’s old lifetime is fully before future work, and safely reuse it within the same capture.

This PR adds an experimental flag `graph_capture_record_stream_reuse: True|False (default: False)`. When enabled, the allocator inserts lightweight free markers and uses capture ordering to decide if a freed block is safe to reuse during capture. If the proof cannot be established, we fall back to the existing post-capture path.

## Terms

* **Free marker**: A capture-legal no-op (created with `cudaGraphAddEmptyNode`) inserted after the last captured use of the block on each stream that used it. 
* **Terminal**: The set of the lastest operations of the stream (or the capturing graph). Any newly captured op on that stream will attach after all nodes in this set. For a stream currently capturing, it is the set of nodes returned in `dependencies_out` by `cudaStreamGetCaptureInfo`. 

### Strong Rule (Graph-Wide Safety)

This rule provides a universal guarantee that a block is safe for reuse by any stream in the graph.

> A block is safe to reuse if every free marker is a predecessor of every terminal of all active streams in the graph.

Why it's safe:

This rule establishes a strict global ordering. Since any new operation on any stream must be appended after that stream's terminals, this condition guarantees that the block's new lifetime begins only after its old lifetime has completely ended everywhere. This prevents lifetime overlaps when the graph is replayed, ensuring correctness.

### Per-stream Rule (A Practical Optimization)

The strong rule, while safe, is often unnecessarily restrictive. The `DeviceCachingAllocator` introduces a crucial constraint that allows for a simpler check.

In `DeviceCachingAllocator`, `get_free_block` only returns blocks whose `block->stream == p.stream()`. In other words, we never reuse a block on a stream different from the allocation stream. This means we don't need to verify safety across the entire graph. We only need to confirm that the block is safe to reuse from the perspective of its own allocation stream.

> Reuse a block for allocations on stream S if every free marker is a predecessor of every node in the terminal set of S.

[...]

PR #146924, opened 2025-02-11
Support pin_memory() during CUDA stream capture.

galv:
This code previously did not work:

```python
import torch

def test():
    graph = torch.cuda.CUDAGraph()

    with torch.cuda.graph(graph, capture_error_mode="global"):
        data = torch.randn(8)
        data_gpu = torch.randn(8, device="cuda")
        data = data.pin_memory()
        data_gpu.to(data, non_blocking=True)

    graph2 = torch.cuda.CUDAGraph()
    with torch.cuda.graph(graph2, capture_error_mode="global"):
        data2 = torch.randn(8)
        data2_gpu = torch.randn(8, device="cuda")
        data2 = data2.pin_memory()
        data2_gpu.to(data2, non_blocking=True)

if __name__ == "__main__":
    test()
```

We use events to signal when a particular usage of a pinned host memory block has completed. Every time we call pin_memory(), cudaEventQuery() gets called to see if we we can reuse existing blocks rather than allocating new blocks. cudaEventQuery() is not allowed during stream capture unless we set the thread to relaxed capture mode. (This is safe in this case so long as we make sure that the pinned buffer is live until the cuda graph is destroyed.)

I haven't fully thought this through. I need to make sure that a pinned memory tensor does in fact stay live until its corresponding cuda graph is destroyed. (I haven't done this yet!)

Draft.

Discovered in https://github.com/pytorch/pytorch/pull/146145#issuecomment-2650061333

Not a high priority, but I wanted to start to figure out what proper support might look like.

PR #161583, opened 2025-08-27
Memory pool support for Caching Host Allocator.

galv:
Includes basic support for stream capture to a cuda graph of pinned host memory to the memory pool.

We need a concept of memory pools in order to support a few things:

1. Support of torch.Tensor.pin_memory() during cuda stream capture
2. Custom allocators, which have a few niche use cases described here: https://github.com/pytorch/pytorch/issues/159906#issuecomment-3190768836

Note that (1) is most important right now, and therefore I have not added support for (2) at this time.

https://github.com/pytorch/pytorch/pull/146924 was my original attempt at this several months ago, where I thought that the major problem was how to implement a graph search algorithm inside the cuda graph to detect whether a given call to record_event() could be killed. In hindsight, this is not important in almost all use cases of interest today. In addition, this implementation also prevented me from recycling memory across seaprate cuda graphs sharing a memory pool (oops). Finally, it seems that there is preliminary evidence from https://github.com/pytorch/pytorch/pull/158352#issuecomment-3215947565 that this graph search on the cuda graph can increase stream capture time substantially. I'm sure that is solvable, but I don't want to hold off on this. Instead, I use the current approach in CUDACachingAllocator.cpp: If a block is used on only one stream, and that stream is the stream that allocated it, then it can be reused. Otherwise, it can never be reused.

The initial use case of this is for offloading cached activations from the forward pass to the host, in order to reduce device memory pressure. Running these training workloads with cuda graphs allows them to run a few percentage points faster. Note that this use case does not ever recycle memory, since all allocations happen in the forward pass, and all frees happen in the backward pass.

@guangyey please give me your feedback on this. I had a choice between making breaking changes to CachingHostAllocatorImpl, which will break your XPU backend and potentially anyone else using your interface (is anyone?), or making a duplicate of it. I did the latter, since I needed to add a new field `was_allocated_during_stream_capture_ ` to the Block template argument. This field needs to be inspected during the allocate() routine, in a way that cannot. Furthermore, I think that keeping implementation separate is good for when I eventually add support for reusing blocks during stream capture (see https://github.com/pytorch/pytorch/pull/146924/files#diff-5d3beb56bf9b6f380f91d5f6f063480ce2e14ca15c415d59d153436018089223R396-R404 for details on the algorithm), which will require two things: (1) usage of cuda graph APIs that exist in cuda but not sycl and (2) eagerly creating events when record_event() is called (which does not fit into the current implementation of CachingHostAllocatorImpl and its virtual methods at all).


ngimel:
>If a block is used on only one stream, and that stream is the stream that allocated it, then it can be reused. Otherwise, it can never be reused.

It's not good enough for host allocator - current events are to prevent reuse on the cpu while previous copy has not finished. Typical usecase:
```
def fn(iter):
    x = torch.empty(1024, pin_memory=True).fill_(iter) # prepare some data
    return x.to("cuda")
    
x_c0 = fn(0)
# at this point x went out of scope, next call to fn will reuse the memory, but the copy has not even started yet!
x_c1 = fn(1)
# by the time copies finally happen, both x_c0 and x_c1 will have "1", if nothing else happened to pinned memory in between
```

galv:
@ngimel well that's embarrassing for me. I got so obsessed with offloading activations to host for training that I didn't think about anything else. Thank you. I will need to think this over for a bit.

guangyey (commented):
Thanks for your mention.
I noticed that in this PR, the host caching allocator has multiple implementations, and each implementation maps to a `MempoolId_t`. Why not introduce a `PrivatePool` concept similar to what `DeviceCachingAllocator` uses?
That way, the host caching allocator would only need a single implementation, and each implementation could internally manage multiple `PrivatePool`s. This approach may not break the original design and would also allow us to reuse existing code.
IMO, aside from differences in runtime APIs and some subtle behavioral variations, the core logic for managing pools/blocks is largely consistent across accelerators. It may be worthwhile to consolidate this logic into shareable code, which would also provide other backends an opportunity to benefit.
In addition, several accelerators provide similar graph modes, such as XPUGraph (coming soon), [NPUGraph](https://github.com/Ascend/pytorch/blob/474452ca754d41abe2d3c682aeeecb53faa6bb18/torch_npu/npu/graphs.py#L178), [MLUGraph](https://github.com/Cambricon/torch_mlu/blob/d578002c0863bfa1ad30fae6b99e4fb3a79dff48/torch_mlu/mlu/graphs.py#L44), and others. Their behavior is quite similar to `CUDAGraph`, so the allocator logic should remain the same and could be made shareable.

galv:
> IMO, aside from differences in runtime APIs and some subtle behavioral variations, the core logic for managing pools/blocks is largely consistent across accelerators.

My point is that this was true when you made your initial work, but it isn't true today for the following reason:

- In order to recycle memory fully during stream capture to a graph, we cannot do the "lazy" method of creating an event and recording it only at the end of a stream. This was a decision made back in torch7 IIRC as an attempt to try to minimize the overhead of cudaEventCreate(), which can synchronize the device. So even if you ignore the cuda stream capture issue, it is not crazy to think that you might want to use the non-lazy method today (since created events are now actually cached via the EventPool). Maybe there will be a performance reason for this for one accelerator, but not for another.

It's possible that Natalia's comment above may invalidate my need to make a lot of big changes. I am still trying to evaluate fully and I am slow :(

Again, I think that making the at::HostAllocator abstract interface is good. 

My problem is with prescribing how this interface is implemented via insisting upon usage of at::CachingHostAllocatorInterface and an Impl class that implements an implicit interface via this macro: https://github.com/pytorch/pytorch/blob/47ecd2042f2f99368cd2d60731d23f4437eecd1d/aten/src/ATen/core/CachingHostAllocator.h#L710-L717 

Following this suggestion makes it extremely hard to change the implementation in non-trivial ways because, as soon as you need to introduce a new abstract method to do something (e.g., I need to call getCurrentCUDAStream() at one point, so I would need to wrap this in a virtual method that would be added here: https://github.com/pytorch/pytorch/blob/2efcf9d0815fd5c729abcbb5622746184b19c3cc/aten/src/ATen/core/CachingHostAllocator.h#L602-L621), you will break at compile time all other users of this interface unfortunately. When these implementations live outside of pytorch's main repo, we are basically out of luck for making sure there is no compile time breakage. I suppose you could add a "default" implementation which simply throws an exception for API's that don't exist for other implementations, but it would still temporarily break other users before they provide a proper implementation, in case calling this virtual method is unavoidable in CachingHostAllocatorImpl.

> I noticed that in this PR, the host caching allocator has multiple implementations, and each implementation maps to a MempoolId_t. Why not introduce a PrivatePool concept similar to what DeviceCachingAllocator uses?

I originally considered doing that. But this approach effectively does the same thing as that, but in a fairly simple way. It also allows each pool to use its own mutexes in a way that will minimize cross-talk across threads. (You do need to use a mutex in CachingHostAllocatorInterface to handle selecting the proper impl, but you can do that via a reader-write lock, aka std::shared_mutex, fortunately.) Since we want pinned host memory allocations to be allocated to the same numa node as the current thread, which (should) be on a numa node next to its current GPU, I think this is a fairly good design that involves minimal changes to CachingHostAllocatorImpl, since it doesn't need to be aware of what a memory pool id is. Ultimately, I am not attached to one particular design, but I think this one is fairly straightforward to understand since it essentially separates out memory pools from the actual allocation logic.

galv:
@ngimel Okay, thanks a lot for the pointer. I can write up a more mathematically precise document if desired, but I think my conclusion is as follows:

1. If a cuda graph's operations _read_ from pinned host memory, that pinned host memory cannot be recycled. This is because cuda stream capture is a form of lazy execution: The actual read from host memory won't happen until after cudaGraphLaunch() is called. So we must keep every host memory allocation used by a cuda graph alive as long as the cuda graph's memory pool is alive, with no opportunity for reuse at all.
2. If a cuda graph's operations _write_ to pinned host memory, that pinned host memory *can* be recycled. This is due to the fact that the CPU cannot ever write GPU memory (unless you are using cudaMallocManaged, which we are not), and because all GPU operations are going to be associated with a stream (thus, they will be captured in the cuda graph). Consider this:

```
import torch

host_data_ptrs = []

def fn(i):
    # d2h copy
    x_cpu = torch.empty(1024, device="cpu", pin_memory=True)
    x_gpu = torch.empty(1024, device="cuda").fill_(i)
    x_cpu.copy_(x_gpu, non_blocking=True)
    host_data_ptrs.append(x_cpu.data_ptr())
    return x_gpu

# warmup
x0_cpu = fn(0)
x1_cpu = fn(1)

g = torch.cuda.CUDAGraph()
with torch.cuda.graph(g, capture_error_mode="thread_local"):
    x0_cpu_graph = fn(0)
    x1_cpu_graph = fn(1)

g.replay()
torch.cuda.synchronize()

# It is safe to reuse this memory
assert host_data_ptrs[-1] == host_data_ptrs[-2]
```

2. (cont'd) However, there is not much point of writing to host memory inside of a cuda graph if you are going to delete it before the host even reads from it.

Some possible use cases come to mind:

- We can write data-dependent shapes or boolean conditions to host memory, have the host allocate memory (or select the next cuda graph to run), and then run the next cuda graph, assuming that we are using the same memory pool for all cuda graphs.
- More importantly, if you simply want to use pinned host memory as a scratch space to save gpu memory that you will then bring back to GPU memory later, then this can be helpful because you can reuse that pinned memory (e.g., for saving cached activations on host during interleaved pipeline parallelism, where it is no longer true that all forward passes happen before all backward passes), but only if we somehow could mark that the read from memory could be reused. In order to mark the read as "safe", we would have to "prove" that the CPU memory was never touched by the CPU. There is no obvious way to do this with today's infrastructure, though it isn't impossible (if your program is a single fxgraph, we could do it with something like inductor graph partition).

The first case is niche, and the second case is hard to implement without blowing up the scope of this. (Note that we would need to expand the record_event() API to document whether the host memory is being read or written as well.)

Nevertheless, the point still stands that being able to pin memory during stream capture is helpful for the sake of increasing memory capacity during training by offloading cached activations to host. I propose this simple path forward: Never recycle any host memory in a memory pool if it *ever* is used during stream capture. This would not hamper the main use case at all. This might make @guangyey happy because it means very little will change to CachingHostAllocatorImpl, such that I can use the original implementation without duplicating it. (But I still think the point stands that there is potential for a divergence to happen in the future.)

ngimel:
Yeah I agree with a simple path forward if it enables important usecases, my comment was we should avoid breaking the world, which recycling memory had a full potential of doing. Lmk when this PR is ready for review.

galv:
Okay, so I discussed a bit internally. If we allocate host memory that the host is banned from touching (we can mark the read and write bits to 0 in the page table on the host to guarantee this), we can actually recycle memory via graph search algorithms. This would allow interleaved pipeline parallelism to be able to reuse memory.

This is possible as of CUDA 13.0, since the cuMemCreate and cuMemSetAccess APIs together allow us to allocate memory on the host (that is device readable and writable but not readable or writable by the host):

> Added support for cuMemCreate and cudaMallocAsync on the host, with CU_MEM_LOCATION_TYPE_HOST.

https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html#id2

I'm not planning to do this anytime soon. We would need (1) a pluggable allocator API like @syed-ahmed made for CUDACachingAllocator and (2) to change the logic for detecting when memory blocks are reusable depending upon if a particular allocator is used. This is non-trivial and not needed by me or anyone internally right this instant. But I figure it was worth noting here. Honestly, @guangyey, I know I've been belaboring this point a few times, but it's going to get more and more difficult to keep the CachingHostAllocatorImpl and CachingHostAllocatorInterface classes "generic" as time goes on so I unfortunately will likely have to break away from using them for the CUDA implementation at some point.

guangyey:
Currently, we see two possible approaches here:
- Option 1 (current design): The host caching allocator has multiple `CachingHostAllocatorImpl`s where each impl maps a memory pool associated with a `MempoolId_t`;
- Option 2 (our proposal): Following the device caching allocator design, the host caching allocator keeps a single `CachingHostAllocatorImpl` and introduces the concept of `PrivatePool` to manage multiple memory pools.

Option 1 risks diverging the codebase, while Option 2 keeps the implementation generic and shareable across backends.
That said, if the PyTorch community prefers CUDA-specific code, we are fine with it. However, we believe Option 2 provides a healthier long-term path for the community by keeping the generic design.

galv:
@guangyey Thank you for the summary message. I will go ahead and evaluate option 2 for this PR, in an effort to reduce the amount of new code, once I revamp it to no longer try to memory block recycling. I do understand that there is clear benefit to maintaining abstractions so that the implementations can be similar, especially when they are quite complicated (e.g., see all the locks here). I appreciate you bearing my long messages.

guangyey:
@galv Thanks a lot for the thoughtful evaluation. Let me know if there’s anything I can help with during your design.

PR #161586, opened 2025-08-27
Add inline fast paths for SymInt operators

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #161586
* #161466

If SymInt::maybe_as_int() returns non-empty, then we get an inline
fast path. The philosophy here (as with the previous PR) is to
preserve performance in the "plain old ints" case.

Observed time spent in SymInt functions in computeStorageNBytes to
drop (and not cost shift elsewhere in the function) after this change,
profiling detach() using code similar to the benchmark from #160580
and Linux perf.

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben

Differential Revision: [D81530107](https://our.internmc.facebook.com/intern/diff/D81530107)

PR #161466, opened 2025-08-26
Outline SymInt::maybe_as_int_slow_path

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #161586
* __->__ #161466

Keeps SymInt::maybe_as_int small enough to inline.

Differential Revision: [D81530097](https://our.internmc.facebook.com/intern/diff/D81530097)

PR #161588, opened 2025-08-27
Add C++ function to accelerate DTensor.__new__

swolchok:
* #161695
* #161694
* #161693
* #161692
* #161635
* #161634
* #161633
* #161622
* #161596
* #161595
* #161591
* #161590
* __->__ #161588
* #161586
* #161466
* #161455
* #161432
* #161329
* #161328
* #161317
* #161315
* #161308
* #161304
* #161292
* #161301
* #161300
* #161286

As the new comment explains, pybinding DispatchKeySet is expensive; one native function avoids the number of DispatchKeySets we incur pybind overhead for to 1.

Reviewers please advise on a better location for the new torch._C function; I just threw it in Module.cpp because I didn't know where to put it.


swolchok:
It might be better to write a specialized torch.Tensor._make_dtensor variant of _make_wrapper_subclass. Will look into that next. Can do this as an incremental step though.

swolchok:
> It might be better to write a specialized torch.Tensor._make_dtensor variant of _make_wrapper_subclass. Will look into that next. Can do this as an incremental step though.

yeah, that approach speeds up benchmark significantly, will put it up soonish

ezyang (approved):
By the way, most wrapper subclasses need to handle negative/conjugate correctly (DTensor had a fix bodged on just for it because I was lazy), so there might be a bigger change that eliminates the problem for all subclass users without another one off function

torch/csrc/Module.cpp
```diff
@@ -2756,6 +2756,22 @@ Call this whenever a new thread is created in order to propagate values from
   torch::set_disabled_torch_dispatch_impl(
       PyObject_GetAttrString(module, "_disabled_torch_dispatch_impl"));
   ASSERT_TRUE(torch::disabled_torch_dispatch_impl() != nullptr);
+
+  // DTensor acceleration.
```
swolchok, line 2760: @ezyang where's the right place for this code to live? make a new .cpp file like all the other ones that are partially initializing torch._C?

albanD: Here is ok, other dispatch/keyset utils are in torch/csrc/utils/python_dispatch.cpp so that would be good as well.
But not a blocker for sure.

albanD (approved):

bdhirsh (approved):

PR #161595, opened 2025-08-27
Add C++ function for torch.distributed.tensor._op_schema.is_view_op

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162337
* #162336
* #161695
* #162298
* #162220
* #162219
* #162218
* #161693
* #161692
* #161634
* #161633
* #161596
* __->__ #161595
* #161591
* #161590
* #161586
* #161466

This seems to have been an especially slow one because of the repeated pybind access (schema is a pybind, as is arguments, and then we hit each argument). It's still ~~1% of total benchmark runtime because of the repeated single pybind function call, but that's a lot better.

Differential Revision: [D81530095](https://our.internmc.facebook.com/intern/diff/D81530095)

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @EikanWang @jgong5 @wenzhe-nrv @sanchitintel

PR #161591, opened 2025-08-27
Don't call check_has_torch_dispatch in THPVariable_NewWithVar if we already know

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162337
* #162336
* #161695
* #162298
* #162220
* #162219
* #162218
* #161693
* #161692
* #161634
* #161633
* #161596
* #161595
* __->__ #161591
* #161590
* #161586
* #161466

We already know when we're called from make_wrapper_subclass or make_dtensor. The check isn't particularly cheap.

Differential Revision: [D81530099](https://our.internmc.facebook.com/intern/diff/D81530099)

PR #161596, opened 2025-08-27
Remove logger.debug statements in DTensor dispatch

swolchok:
* #162337
* __->__ #161596
* #162218
* #162220
* #162219
* #161692
* #161634
* #161633
* #161595
* #161591

These seem to have been costing us 5-10 usec per detach (out of ~~95 usec total).  If they need to ship let's talk about requirements and how we can make this more efficient given that we would prefer if an entire DTensor op could finish in 10 usec.

Differential Revision: [D81530106](https://our.internmc.facebook.com/intern/diff/D81530106)


ezyang (approved):
I am grudgingly approving this. Is there any way we can have a debug statement along these lines without nuking perf?

torch/distributed/tensor/_dispatch.py
```diff
@@ -144,7 +145,6 @@ def dispatch(
 
         # extract local tensor and sharding infos to a OpInfo
         op_info = self.unwrap_to_op_info(op_call, args, kwargs)
-        logger.debug("Dispatching op_call: %s", op_info.schema)
```
Skylion007, line 153: These should not have that much overhead, do they really?

Skylion007 (approved):
Massive loss for logging. How unfortunate the overhead is so high even with lazy string substitution.

Skylion007:
> > I am grudgingly approving this. Is there any way we can have a debug statement along these lines without nuking perf?
> 
> `if DEBUG: print("whatever")` comes to mind I guess?

```python
if logger.isEnabledFor(logging.debug):
```
is what you want.

torch/distributed/tensor/_dispatch.py
wconstab, line 153: This log can be quite useful. But I'm sure we can figure another way to do this. Would it be mostly addressed if we just gated this under a flag for now? I'm surprised it's so expensive though.

swolchok:
> > > I am grudgingly approving this. Is there any way we can have a debug statement along these lines without nuking perf?
> > 
> > 
> > `if DEBUG: print("whatever")` comes to mind I guess?
> 
> ```python
> if logger.isEnabledFor(logging.debug):
>    pass
> ```
> 
> is what you want.

this check is still most of a microsecond:

```
timeit.timeit(stmt="logger.isEnabledFor(logging.DEBUG)", setup="import logging; logger = logging.getLogger('test')")
0.7279661330394447
```

Python is slow. :(

torch/distributed/tensor/_dispatch.py
swolchok: > do they really

My measurements are noisy but per the summary it looked to me like they cost 5-10 usec per dispatch. timeit says the cost for a single trivial disabled log is most of a microsecond (result is total seconds to run the benchmark and it's 1 million iterations by default), and this was multiple logs that also pass arguments.

```
>>> timeit.timeit(stmt="logger.debug('hello')", setup="import logging; logger = logging.getLogger('test')")
0.8429189168382436
```

swolchok:
@swolchok has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81530106).

Skylion007:
> > > > I am grudgingly approving this. Is there any way we can have a debug statement along these lines without nuking perf?
> > > 
> > > 
> > > `if DEBUG: print("whatever")` comes to mind I guess?
> > 
> > 
> > ```python
> > if logger.isEnabledFor(logging.debug):
> >    pass
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > is what you want.
> 
> this check is still most of a microsecond:
> 
> ```
> timeit.timeit(stmt="logger.isEnabledFor(logging.DEBUG)", setup="import logging; logger = logging.getLogger('test')")
> 0.7279661330394447
> ```
> 
> Python is slow. :(

True, but we can cache it for both calls if a single call is a fast enough speedup.

ezyang:
I wouldn't over-index on the test case for the logs; it is likely there to make sure people don't accidentally remove the log, as opposed to saying YOU MUST NOT REMOVE THE LOG (but as usual, the history is the truth)

PR #161604, opened 2025-08-27
port distributed tensor test files for Intel GPU

wincent8:
In this pr, we port test/distributed/tensor test filesfor Intel GPU
We could enable Intel GPU with following methods and try the best to keep the original code styles:

Use torch.accelerator for general gpu
Skip the case if running on xpu which has known issues

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @tianyu-l @XilunWu

### DIFF
--- a/test/distributed/tensor/test_random_ops.py
+++ b/test/distributed/tensor/test_random_ops.py
@@ -24,8 +24,8 @@ from torch.distributed.tensor._random import (
 from torch.distributed.tensor._utils import compute_local_shape_and_global_offset
 from torch.distributed.tensor.debug import CommDebugMode
 from torch.distributed.tensor.parallel import ColwiseParallel, parallelize_module
-from torch.testing._internal.common_utils import run_tests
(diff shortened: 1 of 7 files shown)


guangyey (approved):
LGTM.

test/distributed/tensor/test_random_ops.py
```diff
@@ -592,8 +593,9 @@ class DistTensorRandomOpsTest3D(DTensorTestBase):
     def world_size(self):
         return 8
 
-    @with_comms
     @skip_if_lt_x_gpu(8)
+    @skipXPUIf(True, "Skip it due to XPU CI machine limitation")
```
guangyey: What is the machine limitation, the number of machines, or is the test slow?

wincent8: need 8 cards， thanks

d4l3k (approved):
LGTM

PR #161617, opened 2025-08-27
Fix Inductor Periodic

zou3519:
* __->__ #161617

Models are now passing accuracy. # of graph breaks is larger because
these were not actually tested in CI (if the model fails accuracy we
do not assert on # of graph breaks).


anijain2305 (approved):

zou3519:
wrong label 🤦

PR #161622, opened 2025-08-27
Use specific Tuple/List APIs instead of PySequence in is_int_or_symint_list

swolchok:
* #161695
* #161694
* #161693
* #161692
* #161635
* #161634
* #161633
* __->__ #161622
* #161596
* #161595
* #161591
* #161590
* #161586
* #161466
* #161455
* #161432
* #161329
* #161328
* #161317
* #161315
* #161308
* #161304
* #161292
* #161301

These APIs are faster and, since we can use the container-specific unchecked indexing APIs  don't repeat bounds checks.


swolchok:
this seems to be causing segfaults, rethinking

ezyang (approved):

PR #161635, opened 2025-08-27
PythonArgParser::symintlist{,Optional} should use SymDimVector

swolchok:
* __->__ #161635
* #162281

They're usually/always used for sizes & strides, which is what SymDimVector is for (saves heap allocation of the list itself). Had to patch OptionalArray because there's a bunch of generated code that wants to convert these to SymIntArrayRef and the generator isn't easy to patch to manually wrap. Clear but small improvement in perf on "detach DTensor in a loop" benchmark; we aren't heap-allocating symdimlists anymore, though there's still some cost with destroying these because SymInt needs cleanup.

Differential Revision: [D81530103](https://our.internmc.facebook.com/intern/diff/D81530103)


Skylion007 (approved):

fffrog (approved):

swolchok:
this one needs debugging (`python test/test_torch.py TestTorch.test_tensor_set`); kicking it to top of stack for now

PR #161641, opened 2025-08-27
Ensure that tensors are contiguous before using no-graph MPS impl

rebeccajae:
Fixes #161640

Check if tensors are contiguous before using the no-graph implementation. Using the script in the issue above with this change I get expected results.

```
MPS contiguous result sample: tensor([ 1.3600, -2.9516,  1.3207, -3.5132,  1.7061], device='mps:0')
MPS non-contig result sample: tensor([ 1.3600, -2.9516,  1.3207, -3.5132,  1.7061], device='mps:0')
CPU non-contig result sample: tensor([ 1.3600, -2.9516,  1.3207, -3.5132,  1.7061])
```

### DIFF
--- a/test/test_mps.py
+++ b/test/test_mps.py
@@ -1244,6 +1244,21 @@ class TestMPS(TestCaseMPS):
             torch.nn.functional.linear(torch.rand(size, device='cpu'),
                                        torch.rand(size, device='mps'))
 
+    def test_linear_non_contiguous(self):
+        # Regression test for https://github.com/pytorch/pytorch/issues/161640
+        # Slice tensors to force non-contiguity
+        large_weight = torch.randn(12, 8, device='mps')
+        weight_sliced = large_weight[::2, ::1]
+        weight_contiguous_equiv = weight_sliced.contiguous()
+        input_s = torch.randn(2, 8, device='mps')
+        result_sliced = torch.nn.functional.linear(input_s, weight_sliced)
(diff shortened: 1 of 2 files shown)


malfet (approved):
Looks good to me, would it be possible to add a regression test for it?

rebeccajae:
> Looks good to me, would it be possible to add a regression test for it?

Added regression test.

test/test_mps.py
```diff
@@ -1244,6 +1244,21 @@ def test_linear_errors(self):
...
+        weight_sliced = large_weight[::2, ::1]
+        weight_contiguous_equiv = weight_sliced.contiguous()
+        input_s = torch.randn(2, 8, device='mps')
+        result_sliced = torch.nn.functional.linear(input_s, weight_sliced)
+        result_contig = torch.nn.functional.linear(input_s, weight_contiguous_equiv)
+        input_cpu = input_s.cpu()
+        weight_sliced_cpu = weight_sliced.cpu()
+        result_sliced_cpu = torch.nn.functional.linear(input_cpu, weight_sliced_cpu)
+        self.assertEqual(result_contig, result_sliced, atol=1e-5, rtol=1e-5)
+        self.assertEqual(result_sliced.to("cpu"), result_sliced_cpu, atol=1e-5, rtol=1e-5)
```
malfet: No need to copy results on CPU, comparing the two on GPU feels fine
```suggestion
        self.assertEqual(result_contig, result_sliced)
```

PR #161643, opened 2025-08-27
Apply Triton tensor descriptor for flex-decoding for performance

EikanWang:
* __->__ #161643

cc @voznesenskym @penguinwu @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben


torch/_inductor/kernel/flex/templates/flex_decode.py.jinja
```diff
@@ -220,4 +238,4 @@
 
     mask = (idx_m < Q_LEN) & (idx_d < V_HEAD_DIM)
     acc = acc.reshape(G, BLOCK_M_PER_HQ, V_HEAD_DIM)
-    {{store_output(("idx_z", "idx_t", "idx_hq", "idx_m", "idx_d"), "acc", "mask")}}
\ No newline at end of file
+    {{store_output(("idx_z", "idx_t", "idx_hq", "idx_m", "idx_d"), "acc", "mask")}}
```
EikanWang: Only format change.

whitneywhtsang:
Similar to flex attention, this PR uses tensor descriptor in flex decoding when `USE_TMA` is defined.

drisspg:
can you add a flex-decode specific test

EikanWang:
Oh, my bad—I thought flex-decode TMA already had some test cases

EikanWang:
@drisspg, may I know your thoughts on whether this PR should be imported for Meta's internal CI?

test/inductor/test_flex_decoding.py
```diff
@@ -825,6 +829,57 @@ def test_builtin_score_mods_different_block_size(
         )
         self.run_test(score_mod, dtype, block_mask=block_mask, device=device)
 
+    @common_utils.parametrize("dtype", test_dtypes_fast)
+    @common_utils.parametrize("use_tma", test_tma_options)
+    def test_tma_decoding(self, device, dtype: torch.dtype, use_tma):
```
drisspg: nit just skip the test if device does not support tma, no need to parametrize if there is 1 option

test/inductor/test_flex_decoding.py
```diff
@@ -825,6 +829,57 @@ def test_builtin_score_mods_different_block_size(
...
+            1,
+            n_heads,
+            seq_len,
+            head_dim,
+            device=device,
+            dtype=dtype,
+            requires_grad=False,
+        )
+
+        score_mod = _generate_alibi_bias(8)
```
drisspg: this should be the num_heads

test/inductor/test_flex_decoding.py
drisspg: actualyl can you just update run_test to accept kernel_options and then add parametrize over this
https://github.com/pytorch/pytorch/blob/2fed4fb464d87fe7cc2ff646ec2bb8052e76c729/test/inductor/test_flex_decoding.py#L786

EikanWang: I just tried to ensure we are on the page: Do you mean to apply TMA for all the test cases for flex-decoding or just use run_test with kernel option for this test case? @drisspg

drisspg: For now lets just update run_test to accept kernel params and use that as the tester for this specific test

EikanWang: The newly added test case will be skipped if `has_triton_tma_device` is `False`, and the `False` scenario should have been covered by other test cases.

test/inductor/test_flex_decoding.py
```diff
@@ -846,6 +852,28 @@ def test_builtin_score_mods_different_block_size(
         )
         self.run_test(score_mod, dtype, block_mask=block_mask, device=device)
 
+    @unittest.skipIf(not has_triton_tma_device(), "Skip when TMA is not available")
+    @common_utils.parametrize("dtype", test_dtypes_fast)
+    def test_tma_decoding(self, device, dtype: torch.dtype):
+        n_heads, head_dim, seq_len = 4, 16, 128
+
+        score_mod = _generate_alibi_bias(n_heads)
+        kernel_options = {"USE_TMA": True}
```
EikanWang, line 861: The newly added test case will be skipped if `has_triton_tma_device` is `False`, and the `False` scenario should have been covered by other test cases.

EikanWang:
> @EikanWang this patch doesn't enable USE_TMA for FlexDecoding for XPU. Is there another path that enable that opt-in ?

@etiotto , Enabling USE_TMA is part of the integration logic. Users can decide if USE_TMA should be enabled or not.

drisspg (approved):
Thanks

Issue #158772, opened 2025-07-21
Compile failing with NamedTuple._replace


ameya98:
### 🐛 Describe the bug

A simple example:
```python
import torch
from typing import NamedTuple


class ModelOutput(NamedTuple):
    logits: torch.Tensor
    loss: torch.Tensor


def fn(tup: ModelOutput) -> torch.Tensor:
    tup = tup._replace(logits=tup.logits * 2)
    return tup.logits + tup.loss

tup = ModelOutput(logits=torch.tensor([1.0, 2.0]), loss=torch.tensor(0.5))
compiled_fn = torch.compile(fn, fullgraph=True, dynamic=True)
result = compiled_fn(tup)
```

This fails to compile.

### Error logs

```
Unsupported: Attempted to inline function marked as skipped
  Explanation: Dynamo developers have intentionally marked that the function `ModelOutput._replace` should not be traced.
  Hint: Avoid calling the function `ModelOutput._replace`.
  Hint: Remove the function `ModelOutput._replace` or the file `/homefs/home/daigavaa/miniforge3/envs/jamun/lib/python3.11/collections/__init__.py` from torch/_dynamo/trace_rules.py. More graph breaks may occur as a result of attempting to trace into the function.
  Hint: Please file an issue to PyTorch.

  Developer debug context: qualname: ModelOutput._replace, name: _replace, filename: `/homefs/home/daigavaa/miniforge3/envs/jamun/lib/python3.11/collections/__init__.py`, skip reason: skipped according trace_rules.lookup SKIP_DIRS
```

### Versions

```
Collecting environment information...
PyTorch version: 2.7.1+cu126
Is debug build: False
CUDA used to build PyTorch: 12.6
ROCM used to build PyTorch: N/A

OS: Amazon Linux 2023.6.20250211 (x86_64)
GCC version: (GCC) 11.4.1 20230605 (Red Hat 11.4.1-2)
Clang version: Could not collect
CMake version: version 3.22.2
Libc version: glibc-2.34

Python version: 3.11.11 | packaged by conda-forge | (main, Dec  5 2024, 14:17:24) [GCC 13.3.0] (64-bit runtime)
Python platform: Linux-6.1.127-135.201.amzn2023.x86_64-x86_64-with-glibc2.34
Is CUDA available: True
CUDA runtime version: Could not collect
CUDA_MODULE_LOADING set to: LAZY
GPU models and configuration: GPU 0: NVIDIA A10G
[...]

PR #161645, opened 2025-08-27
NamedTuple: Allow side effects for dynamic attributes

### LINKED ISSUE #161610: NamedTuple Dynamic Objects do not persist via torch compile and eager backend
### 🐛 Describe the bug

Code to reproduce bug:
Expected behavior is both objects result in the same print statements. Actual behavior compiled function results in error when trying to get extra info. 
```python
import torch
from typing import NamedTuple

class MyNamedTuple(NamedTuple):
    first: torch.Tensor
    second: torch.Tensor

class MyNamedTupleSubclass(MyNamedTuple):
    pass

def fn(tup: MyNamedTuple) -> torch.Tensor:
    extra_info = torch.tensor(4.0)
    tup.extra_info = extra_info  # Add dynamic attribute
    return tup

print("\nTesting NamedTuple with __setattr__:")
extended_tup = MyNamedTupleSubclass(first=torch.tensor([2.0]), second=torch.tensor(1.0))
setattr_result = fn(extended_tup)
print(f"NamedTuple __setattr__ result: {setattr_result.extra_info}")


print("\nTesting NamedTuple with __setattr__:")
extended_tup = MyNamedTupleSubclass(first=torch.tensor([2.0]), second=torch.tensor(1.0))
setattr_result = torch.compile(fn, backend="eager")(extended_tup)
print(f"NamedTuple __setattr__ result: {setattr_result.extra_info}")
```

### Error logs

  File "/workspaces/pytorch/temp_scripts/temp.py", line 25, in <module>
    print(f"NamedTuple __setattr__ result: {setattr_result.extra_info}")
                                            ^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'MyNamedTupleSubclass' object has no attribute 'extra_info'

### Versions

PyTorch version: 2.9.0a0+gite7e270a
Is debug build: False

OS: Fedora Linux 41 (Container Image) (x86_64)
GCC version: (GCC) 14.3.1 20250523 (Red Hat 14.3.1-1)
CMake version: version 4.1.0
Libc version: glibc-2.40

Python version: 3.12.11 | packaged by conda-forge | (main, Jun  4 2025, 14:45:31) [GCC 13.3.0] (64-bit runtime)
Python platform: Linux-6.15.9-201.fc42.x86_64-x86_64-with-glibc2.40
Is CUDA available: False
CUDA runtime version: No CUDA
cuDNN version: No CUDA
Is XPU available: False
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Architecture:                            x86_64
CPU op-mode(s):                          32-bit, 64-bit
Address sizes:                           46 bits physical, 48 bits virtual
Byte Order:                              Little Endian
CPU(s):                                  22
On-line CPU(s) list:                     0-21
Vendor ID:                               GenuineIntel
Model name:                              Intel(R) Core(TM) Ultra 7 165H
CPU family:                              6
Model:                                   170
Thread(s) per core:                      2
Core(s) per socket:                      16
Socket(s):                               1
Stepping:                                4
CPU(s) scaling MHz:                      32%
CPU max MHz:                             5000.0000
CPU min MHz:                             400.0000
BogoMIPS:                                6144.00
Flags:                                   fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc art arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf tsc_known_freq pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm 3dnowprefetch cpuid_fault epb ssbd ibrs ibpb stibp ibrs_enhanced tpr_shadow flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid rdseed adx smap clflushopt clwb intel_pt sha_ni xsaveopt xsavec xgetbv1 xsaves split_lock_detect user_shstk avx_vnni dtherm ida arat pln pts hwp hwp_notify hwp_act_window hwp_epp hwp_pkg_req hfi vnmi umip pku ospke waitpkg gfni vaes vpclmulqdq rdpid bus_lock_detect movdiri movdir64b fsrm md_clear serialize pconfig arch_lbr ibt flush_l1d arch_capabilities
Virtualization:                          VT-x
L1d cache:                               544 KiB (14 instances)
L1i cache:                               896 KiB (14 instances)
L2 cache:                                18 MiB (9 instances)
L3 cache:                                24 MiB (1 instance)
NUMA node(s):                            1
NUMA node0 CPU(s):                       0-21
Vulnerability Gather data sampling:      Not affected
Vulnerability Ghostwrite:                Not affected
Vulnerability Indirect target selection: Not affected
Vulnerability Itlb multihit:             Not affected
Vulnerability L1tf:                      Not affected
Vulnerability Mds:                       Not affected
Vulnerability Meltdown:                  Not affected
Vulnerability Mmio stale data:           Not affected
Vulnerability Reg file data sampling:    Not affected
Vulnerability Retbleed:                  Not affected
Vulnerability Spec rstack overflow:      Not affected
Vulnerability Spec store bypass:         Mitigation; Speculative Store Bypass disabled via prctl
Vulnerability Spectre v1:                Mitigation; usercopy/swapgs barriers and __user pointer sanitization
Vulnerability Spectre v2:                Mitigation; Enhanced / Automatic IBRS; IBPB conditional; PBRSB-eIBRS Not affected; BHI BHI_DIS_S
Vulnerability Srbds:                     Not affected
Vulnerability Tsa:                       Not affected
Vulnerability Tsx async abort:           Not affected

[pip3] flake8==7.3.0
[pip3] flake8-pyi==25.5.0
[pip3] intel-cmplr-lib-ur==2025.2.1
[pip3] intel-openmp==2025.2.1
[pip3] mkl-include==2025.2.0
[pip3] mkl-static==2025.2.0
[pip3] mypy==1.16.0
[pip3] mypy_extensions==1.1.0
[pip3] numpy==2.1.0
[pip3] optree==0.13.0
[pip3] tbb==2022.2.0
[pip3] tbb-devel==2022.2.0
[pip3] tcmlib==1.4.0
[pip3] torch==2.9.0a0+gite7e270a
[pip3] umf==0.11.0
[conda] intel-cmplr-lib-ur               2025.2.1            pypi_0              pypi
[conda] intel-openmp                     2025.2.1            pypi_0              pypi
[conda] mkl-include                      2025.2.0            pypi_0              pypi
[conda] mkl-static                       2025.2.0            pypi_0              pypi
[conda] numpy                            2.1.0               pypi_0              pypi
[conda] optree                           0.13.0              pypi_0              pypi
[conda] tbb                              2022.2.0            pypi_0              pypi
[conda] tbb-devel                        2022.2.0            pypi_0              pypi
[conda] tcmlib                           1.4.0               pypi_0              pypi
[conda] torch                            2.9.0a0+gite7e270a  pypi_0              pypi
[conda] torchfix                         0.4.0               pypi_0              pypi
[conda] umf                              0.11.0              pypi_0              pypi

cc @ezyang @gchanan @zou3519 @kadeng @msaroufim @chauhang @penguinwu @voznesenskym @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @amjames @Lucaskabela

morrison-turnansky:
I confirmed that the tracing was correct i.e. NamedTupleVariable had the correct dynamic attribute added to it.

The problem was that NamedTupleVariable was always marked as immutable. This does not reflect the behavior of namedtuple. 

Subclasses of namedtuple may be mutable, so when a NamedTupleVariable is derived from a subclass that is mutable, I made NamedTupleVariable mutable as well. Then side_effects correctly updates the returned object. 

Fixes #161610

### DIFF
--- a/torch/_dynamo/variables/builder.py
+++ b/torch/_dynamo/variables/builder.py
@@ -707,7 +707,10 @@ class VariableBuilder:
             result = NamedTupleVariable(
                 output, tuple_cls=type(value), source=self.source
             )
-            return result
+            if result.is_structseq() or result.tuple_cls.__bases__ == (tuple,):
+                return result
+
+            return self.tx.output.side_effects.track_object_existing(value, result)
         elif istype(value, (dict, collections.defaultdict, collections.OrderedDict)):
             self.install_guards(GuardBuilder.TYPE_MATCH)
             all_const = all(ConstantVariable.is_literal(k) for k in value.keys())
--- a/torch/_dynamo/variables/builder.py
+++ b/torch/_dynamo/variables/builder.py
@@ -707,9 +707,6 @@ class VariableBuilder:
             result = NamedTupleVariable(
                 output, tuple_cls=type(value), source=self.source
             )
-            if result.tuple_cls.__bases__ == (tuple,):
-                return result
-
             return self.tx.output.side_effects.track_object_existing(value, result)
         elif istype(value, (dict, collections.defaultdict, collections.OrderedDict)):
             self.install_guards(GuardBuilder.TYPE_MATCH)
--- a/torch/_dynamo/variables/builder.py
+++ b/torch/_dynamo/variables/builder.py
@@ -707,7 +707,7 @@ class VariableBuilder:
             result = NamedTupleVariable(
                 output, tuple_cls=type(value), source=self.source
             )
-            if result.is_structseq() or result.tuple_cls.__bases__ == (tuple,):
+            if result.tuple_cls.__bases__ == (tuple,):
                 return result
 
             return self.tx.output.side_effects.track_object_existing(value, result)
--- a/torch/_dynamo/variables/lists.py
+++ b/torch/_dynamo/variables/lists.py
@@ -1129,7 +1129,14 @@ class NamedTupleVariable(TupleVariable):
             ):
                 raise_observed_exception(AttributeError, tx)
             # Subclass of namedtuple type can have dynamic attributes
+            # STEP 1: Ensure proper mutation type for attribute mutations
+            from .base import AttributeMutationExisting
+
+            self.mutation_type = AttributeMutationExisting()
+
             tx.output.side_effects.mutation(self)
+            # STEP 3: Store the attribute mutation (CRITICAL - this was missing!)
+            tx.output.side_effects.store_attr(self, attr, value)
             self.dynamic_attributes[attr] = value
             return ConstantVariable.create(None)
         return super().call_method(tx, name, args, kwargs)
--- a/torch/_dynamo/variables/lists.py
+++ b/torch/_dynamo/variables/lists.py
@@ -1076,9 +1076,22 @@ class NamedTupleVariable(TupleVariable):
     def as_python_constant(self):
         if self.is_structseq():
             # StructSequenceType(iterable)
-            return self.python_type()([x.as_python_constant() for x in self.items])
-        # NamedTupleType(*iterable)
-        return self.python_type()(*[x.as_python_constant() for x in self.items])
+            result = self.python_type()([x.as_python_constant() for x in self.items])
+        else:
+            # NamedTupleType(*iterable)
+            result = self.python_type()(*[x.as_python_constant() for x in self.items])
+        
+        # Apply dynamic attributes if any were set
+        if self.dynamic_attributes:
+            for attr_name, attr_value in self.dynamic_attributes.items():
+                # Convert VariableTracker to Python constant if needed
+                if hasattr(attr_value, 'as_python_constant'):
+                    python_value = attr_value.as_python_constant()
+                else:
+                    python_value = attr_value
+                setattr(result, attr_name, python_value)
+        
+        return result
 
     def as_proxy(self):
         assert self.python_type() is not SizeVariable
@@ -1089,6 +1102,12 @@ class NamedTupleVariable(TupleVariable):
         return self.python_type()(*self._as_proxy())
 
     def reconstruct(self, codegen: "PyCodegen") -> None:
+        # If we have dynamic attributes, we need to use constant loading
+        if self.dynamic_attributes:
+            codegen.append_output(
+                codegen.create_load_const_unchecked(self.as_python_constant())
+            )
+            return
         # Constructors:
         #   StructSequenceType(iterable)
         #   NamedTupleType(*iterable)
@@ -1128,12 +1147,17 @@ class NamedTupleVariable(TupleVariable):
                 or attr in self.fields()
             ):
                 raise_observed_exception(AttributeError, tx)
-            # Subclass of namedtuple type can have dynamic attributes
-            from .base import AttributeMutationExisting
-
-            self.mutation_type = AttributeMutationExisting()
+            
+                  # Handle mutation tracking based on variable type
+            if self.mutation_type is None:
+                from .base import AttributeMutationNew
+                self.mutation_type = AttributeMutationNew()
+            
+            # Always call mutation to mark as modified
             tx.output.side_effects.mutation(self)
             tx.output.side_effects.store_attr(self, attr, value)
+            
+            # Always update our internal tracking
             self.dynamic_attributes[attr] = value
             return ConstantVariable.create(None)
         return super().call_method(tx, name, args, kwargs)
--- a/torch/_dynamo/variables/lists.py
+++ b/torch/_dynamo/variables/lists.py
@@ -1128,6 +1128,7 @@ class NamedTupleVariable(TupleVariable):
                 or attr in self.fields()
             ):
                 raise_observed_exception(AttributeError, tx)
+            # Subclass of namedtuple type can have dynamic attributes
             from .base import AttributeMutationExisting
             self.mutation_type = AttributeMutationExisting()
             tx.output.side_effects.mutation(self)
--- a/torch/_dynamo/variables/lists.py
+++ b/torch/_dynamo/variables/lists.py
@@ -1137,18 +1137,11 @@ class NamedTupleVariable(TupleVariable):
             + create_call_function(1, False)
         )
 
-        # Handle dynamic attributes through store_attr_mutations
-        from torch._dynamo.symbolic_convert import InstructionTranslator
-
-        tx = InstructionTranslator.current_tx()
-        if tx.output.side_effects.has_pending_mutation(self):
-            for name, value in tx.output.side_effects.store_attr_mutations[
-                self
-            ].items():
-                codegen.dup_top()
-                codegen(value)
-                codegen.extend_output(create_rot_n(2))
-                codegen.store_attr(name)
+        for name, value in self.dynamic_attributes.items():
+            codegen.dup_top()
+            codegen(value)
+            codegen.extend_output(create_rot_n(2))
+            codegen.store_attr(name)
 
     def call_method(
         self,
--- a/torch/_dynamo/variables/lists.py
+++ b/torch/_dynamo/variables/lists.py
@@ -25,7 +25,11 @@ import torch
 import torch.fx
 
 from .. import graph_break_hints, polyfills, variables
-from ..bytecode_transformation import create_call_function, create_instruction
+from ..bytecode_transformation import (
+    create_call_function,
+    create_instruction,
+    create_rot_n,
+)
 from ..exc import raise_observed_exception, unimplemented_v2
 from ..source import AttrSource, NamedTupleFieldsSource
 from ..utils import (
@@ -1114,12 +1118,7 @@ class NamedTupleVariable(TupleVariable):
         return self.python_type()(*self._as_proxy())
 
     def reconstruct(self, codegen: "PyCodegen") -> None:
-        # If we have dynamic attributes, we need to use constant loading
-        if self.dynamic_attributes:
-            codegen.append_output(
-                codegen.create_load_const_unchecked(self.as_python_constant())
-            )
-            return
+        # Always reconstruct the NamedTuple normally first
         # Constructors:
         #   StructSequenceType(iterable)
         #   NamedTupleType(*iterable)
@@ -1137,6 +1136,19 @@ class NamedTupleVariable(TupleVariable):
             ]
             + create_call_function(1, False)
         )
+
+        # Handle dynamic attributes through store_attr_mutations
+        from torch._dynamo.symbolic_convert import InstructionTranslator
+
+        tx = InstructionTranslator.current_tx()
+        if tx.output.side_effects.has_pending_mutation(self):
+            for name, value in tx.output.side_effects.store_attr_mutations[
+                self
+            ].items():
+                codegen.dup_top()
+                codegen(value)
+                codegen.extend_output(create_rot_n(2))
+                codegen.store_attr(name)
 
     def call_method(
         self,
--- a/torch/_dynamo/variables/lists.py
+++ b/torch/_dynamo/variables/lists.py
@@ -1128,14 +1128,9 @@ class NamedTupleVariable(TupleVariable):
                 or attr in self.fields()
             ):
                 raise_observed_exception(AttributeError, tx)
-            # Subclass of namedtuple type can have dynamic attributes
-            # STEP 1: Ensure proper mutation type for attribute mutations
             from .base import AttributeMutationExisting
-
             self.mutation_type = AttributeMutationExisting()
-
             tx.output.side_effects.mutation(self)
-            # STEP 3: Store the attribute mutation (CRITICAL - this was missing!)
             tx.output.side_effects.store_attr(self, attr, value)
             self.dynamic_attributes[attr] = value
             return ConstantVariable.create(None)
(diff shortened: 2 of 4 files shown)


torch/_dynamo/variables/lists.py
```diff
@@ -1129,7 +1129,11 @@ def call_method(
             ):
                 raise_observed_exception(AttributeError, tx)
             # Subclass of namedtuple type can have dynamic attributes
+            from .base import AttributeMutationExisting
+
+            self.mutation_type = AttributeMutationExisting()
```
anijain2305: Thanks for raising the correctness issue, and sending a fix.

1) Do you need this after using `self.tx.output.side_effects.track_object_existing(value, result)` in builder.py?

2) We also need to support case where the named tuple is created inside the function getting compiled, and then getting returned. In that case, the mutation_type should be `AttributeMutationNew`. Can you add a test case for that as well?

anijain2305 (changes requested):

torch/_dynamo/variables/lists.py
morrison-turnansky: For 1), you are correct that is no longer necessary. 
For 2), I ran a test case, and it is failing for basically the same reason. This makes sense. I will add a fix for this as well as they go through different code paths.

morrison-turnansky: @anijain2305 I put in a fix for the sourceless case as well. Respective tests are also added. Please let me know you want any additional changes

torch/_dynamo/variables/lists.py
```diff
@@ -1129,7 +1148,12 @@ def call_method(
             ):
                 raise_observed_exception(AttributeError, tx)
             # Subclass of namedtuple type can have dynamic attributes
+            if self.source is None:
+                from .base import AttributeMutationNew
+
+                self.mutation_type = AttributeMutationNew()
```
anijain2305: So this should happen when we initialize the Variable tracker through the `kwargs` member. Can you put a breakpoint in the `__init__` to directly update the place where we build the variable tracker for NamedTuple . This is similar to your change of `self.tx.output.side_effects.track_object_existing(value, result)` but for variable trackers that are created inside the compiled frame.

torch/_dynamo/variables/lists.py
```diff
@@ -1089,6 +1102,12 @@ def as_proxy(self):
         return self.python_type()(*self._as_proxy())
 
     def reconstruct(self, codegen: "PyCodegen") -> None:
+        # If we have dynamic attributes, we need to use constant loading
+        if self.dynamic_attributes:
+            codegen.append_output(
+                codegen.create_load_const_unchecked(self.as_python_constant())
+            )
+            return
```
anijain2305: This might not be correct. You could have a mutable object in the NamedTuple which (1) might not be a converted into a constant, or (2) even if it could, we should use `source` to reconstruct the object so that id of the object is same.

anijain2305 (commented):
If supporting NamedTuple creation inside the compiled frame and then returning is hard, we can do a graph break and not support it in this PR. 

Let me know - this PR is little  more complicated than I anticipated because most of the code around tuples is written with immutability in mind, and named tuple creation in CPython is not a standard process (CPython generates some string, and then `eval` it on the fly), so using a `UserDefinedObjectVariable` is also not readily available.

torch/_dynamo/variables/lists.py
morrison-turnansky: Added mutation logic in ```NamedTupleVariable.__init__```. I believe that is where you wanted it as opposed to ```VariableTracker.__init__```.

torch/_dynamo/variables/lists.py
morrison-turnansky: Thank you for the comments. This PR in general has really helped me get a handle on this part of the code base. 

 I am trying to clarify what you mean by source here.

When passing a subclass of namedtuple to the function, the NamedTupleVariable.source is defined as one would expect. I have confirmed that the returned object has the same object ID as the original. In particular, the original object is directly altered.

However,  when we construct the subclass of  NamedTuple within compilation as opposed to passing it to the function, NamedTupleVariable.source is None as are the dynamic attributes' respective source.  I followed patterns elsewhere for reconstruction of object attributes for this case. This fixes 1). For 2), I have avoided the copy with this change by directly adding the dynamic attributes to the reconstructed bytecode stack.

morrison-turnansky:
Let me know if these changes were close to what you wanted. I am happy to keep iterating on this, but I am also fine to just introduce a graph break.

torch/_dynamo/variables/lists.py
```diff
@@ -1076,9 +1092,22 @@ def python_type(self):
...
+            result = self.python_type()(*[x.as_python_constant() for x in self.items])
+
+        # Apply dynamic attributes if any were set
+        if self.dynamic_attributes:
+            for attr_name, attr_value in self.dynamic_attributes.items():
+                # Convert VariableTracker to Python constant if needed
+                if hasattr(attr_value, "as_python_constant"):
+                    python_value = attr_value.as_python_constant()
+                else:
+                    python_value = attr_value
```
anijain2305: Should we raise `NotImplemetedError` here?

torch/_dynamo/variables/lists.py
```diff
@@ -1048,10 +1052,22 @@ class NamedTupleVariable(TupleVariable):
...
+            mutation_type = None
+
+            if self.is_namedtuple() and tuple_cls.__bases__ != (tuple,):
+                from .base import AttributeMutationNew
+
+                mutation_type = AttributeMutationNew()
+
+            super().__init__(items, mutation_type=mutation_type, **kwargs)
+        else:
+            super().__init__(items, **kwargs)
```
anijain2305: Can you see where this is coming from and just set the `mutation_type` there (similar to builder.py)?

My concern is that if later on someone sets the `mutation_type` in kwargs, it will be ignored because of this code.

anijain2305 (changes requested):
Very close to be done. Thanks for sending the PR.

torch/_dynamo/variables/lists.py
morrison-turnansky: That makes sense.  I could see how the current implementation could cause problems later. Update to raise error.

torch/_dynamo/variables/lists.py
morrison-turnansky: Now  I understand what you had meant. I moved this into moved this user defined variable call function which is where the original constructor call happens for this case. @anijain2305 all issues should be resolved now, please let me know if there is anything else.

torch/_dynamo/variables/lists.py
```diff
@@ -1130,6 +1156,7 @@ def call_method(
                 raise_observed_exception(AttributeError, tx)
             # Subclass of namedtuple type can have dynamic attributes
             tx.output.side_effects.mutation(self)
+            tx.output.side_effects.store_attr(self, attr, value)
```
StrongerXi: Could you put this under a `if self.source:` check? Because IIUC here the `side_effects.store_attr` is just for codegen-ing the STORE_ATTR for dynamic attributes of pre-existing named tuple subclass objects.

For newly created ones, you are handling it in `reconstruct` above. So `side_effects.store_attr` for these cases aren't really needed, and might create unexpected consequences due to information duplication.

Overall, I think an ideal fix to this and #158772 is to make `NamedTupleVariable` a subclass of `UserDefinedTupleVariable`, or even remove `NamedTupleVariable` altogether and treat it as a `UserDefinedTupleVariable`, although I think the latter is hard due to structseq, I was trying it a few months ago. This is ideal because it basically removes all the code dup we have in `NamedTupleVariable`, which are basically old ways of emulating the logic we already had in `UserDefinedObjectVariable` and `SideEffect`. But that might be another rabbit hole, so I'm just making a note in case anyone becomes interested in tackling this.

torch/_dynamo/variables/user_defined.py
```diff
@@ -716,7 +716,16 @@ def deque_signature(iterable=None, maxlen=None):
 
                 assert all(x is not None for x in items)
 
-            return variables.NamedTupleVariable(items, self.value)
+            # Modify mutability of namedtuple for sourcelesss instantiations.
+            mutation_type = None
+            if self.value.__bases__ != (tuple,):
```
StrongerXi: Ditto.

torch/_dynamo/variables/builder.py
```diff
@@ -707,7 +707,10 @@ def from_tensor():
             result = NamedTupleVariable(
                 output, tuple_cls=type(value), source=self.source
             )
-            return result
+            if result.is_structseq() or result.tuple_cls.__bases__ == (tuple,):
```
StrongerXi: I think this mirrors 
https://github.com/pytorch/pytorch/blob/b4827d1486b2ea15cbf5b324752a6c04b43160fe/torch/_dynamo/variables/lists.py#L1122-L1126

But do we really need to check them upfront, if we are already handling them in the `call_method`?

StrongerXi (commented):
Thanks for the PR! Looks mostly good to me. Dropped a few comments. Let's see what the tests say.

torch/_dynamo/variables/builder.py
morrison-turnansky: Change made.

torch/_dynamo/variables/lists.py
morrison-turnansky: @StrongerXi I made the changes you had specified. 

As far as the larger change, let's get these fixes in, and then I can take a look at the larger change if you want.

torch/_dynamo/variables/lists.py
```diff
@@ -1130,6 +1156,14 @@ def call_method(
                 raise_observed_exception(AttributeError, tx)
             # Subclass of namedtuple type can have dynamic attributes
             tx.output.side_effects.mutation(self)
+            if self.source:
+                tx.output.side_effects.track_object_existing(self.source, self)
```
StrongerXi: I thought you had this and `self.mutation_type` below somewhere in `builder.py` and `user_defined.py` (`UserClassObjectVariable.call_function`)? That's actually the preferred approach since we'd like to set `mutation_type` as close to VariableTracker creation as possible, rather than changing them dynamically.

I probably created some confusion, I meant you can keep these modifications in those files, while removing their `is_structseq` checks, since we are already checking those in `call_method` here.

torch/_dynamo/variables/lists.py
StrongerXi: Yep that sounds great, thank you!

torch/_dynamo/variables/lists.py
morrison-turnansky: All good this is what source control and frequent commits are for. Yes I has misunderstood original comment, and I agree with you from a design perspective. Removed respective ```is_structseq``` checks.

StrongerXi: Thanks, one last bit, remove the tuple checks as well, since we are already checking for it in `call_method`, just like `is_structseq`:
https://github.com/pytorch/pytorch/blob/b4827d1486b2ea15cbf5b324752a6c04b43160fe/torch/_dynamo/variables/lists.py#L1125-L1126

Then it looks solid!

morrison-turnansky: @StrongerXi Great, and done.

StrongerXi: Sorry, I mean removing it from the newly added code, rather than the existing `call_method` code path.

morrison-turnansky: @StrongerXi  Okay, I have made the specified changes. Just to be clear since all of the checks are in call_method, NamedTupleVariable will be mutable when created via builder or user_defined. This seems like incorrect behavior.  On the other hand, the only mutating method currently is ```__setattr__```, so this shouldn't create a bug immediately. I suppose this general issue with maintainability and repeating code is one of the reasons why you want to integrate NamedTupleVariable with ```UserDefinedTupleVariable```.

StrongerXi: > I suppose this general issue with maintainability and repeating code is one of the reasons why you want to integrate NamedTupleVariable with UserDefinedTupleVariable.

Yep. For historical context, we had a lot of specialized VariableTracker subclasses like `NamedTupleVariable` because we couldn't trace into these python objects (and their methods) easily back then. Now that we've significantly improved the coverage of `torch.compile` on python language features, a lot of these specialized subclasses could be removed (with variable amount of efforts).

morrison-turnansky: thanks for the context!

morrison-turnansky:
@anijain2305 Please review and/or resolve requested change.

StrongerXi:
Looks great, merging, thank you!

StrongerXi (approved):

anijain2305 (approved):

PR #161647, opened 2025-08-27
[Flex] Fix float16 default config 128 headdim

drisspg:
* __->__ #161647

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Chillee @yanboliang @BoyuanFeng


torch/_inductor/template_heuristics.py
```diff
@@ -797,7 +797,7 @@ def __init__(self) -> None:
             (torch.bfloat16, 128): FlexConfig(128, 64, 3, 8),
             (torch.bfloat16, 256): FlexConfig(64, 32, 3, 4),
             (torch.float16, 64): FlexConfig(128, 128, 3, 4),
-            (torch.float16, 128): FlexConfig(128, 128, 3, 8),
+            (torch.float16, 128): FlexConfig(128, 64, 3, 8),
```
v0i0: would there ever be a reason for float16 config to be different from bfloat16 config? i see they are different for sm100 above too

v0i0 (approved):

torch/_inductor/template_heuristics.py
drisspg: That is a good question .. I think they should be the same, let me do a sweep between the two to see if there is coverage / perf differences between the two settings

drisspg: Baseline: num_stages_2
Change:   num_stages_3
Geometric Mean Speedup (change over baseline): 0.9951x
Geometric Mean % Change: -0.49%
Median Speedup (change over baseline): 0.9977x
Speedup Std Dev: 0.0123


essentially within noise, bumped to 3 since in theory feels like if we can pipeline more why not

PR #161651, opened 2025-08-27
[ONNX][submodule] Update onnx and protobuf

titaiwangms:
Fixes #ISSUE_NUMBER


titaiwangms:
We will keep using ONNX==1.18.0 for existing torchscript exporter.

Issue #161368, opened 2025-08-24
Incorrect offsets for 2D uneven sharding in DTensor in `compute_local_shape_and_global_offsetproduces`


KimbingNg:
### 🐛 Describe the bug

```python
# torchrun --nproc_per_node=8 2d_uneven_sharding_issue.py

import torch
import loguru
from torch import nn
from torch import distributed as dist
import torch.distributed.checkpoint as dcp
from torch.distributed.checkpoint.stateful import Stateful
from torch.distributed.device_mesh import DeviceMesh, init_device_mesh
from torch.distributed.tensor import DTensor, Replicate, Shard
from torch.distributed.tensor._utils import compute_local_shape_and_global_offset

dist.init_process_group(backend='nccl', init_method='env://')
torch.cuda.set_device(dist.get_rank())

tensor = torch.randn(2)


# 1D uneven sharding produces correct offset
mesh = DeviceMesh(device_type="cuda", mesh=[0, 1, 2, 3, 4, 5, 6, 7])
placements = [Shard(0)]
dtensor = DTensor.from_local(tensor, mesh).redistribute(placements=placements)
size, offset = compute_local_shape_and_global_offset(dtensor.shape, mesh, placements)
loguru.logger.info(f'[Rank {dist.get_rank()}] dtensor: {size=} {dtensor.to_local().shape} | {offset=}')

# =================================================================================================
dist.barrier() # optimize output
# =================================================================================================

# 2D uneven sharding produces incorrect offset
mesh = DeviceMesh(device_type="cuda", mesh=[[0, 1, 2, 3], [4, 5, 6, 7]])
placements = [Shard(0), Shard(0)]
dtensor2 = DTensor.from_local(tensor, mesh).redistribute(placements=placements)
size, offset = compute_local_shape_and_global_offset(dtensor2.shape, mesh, placements)
loguru.logger.info(f'[Rank {dist.get_rank()}] dtensor2: {size=} {dtensor2.to_local().shape} | {offset=}')
```

output:
```
[Rank 0] dtensor: size=(1,) torch.Size([1]) | offset=(0,)
[Rank 1] dtensor: size=(1,) torch.Size([1]) | offset=(1,)
[Rank 2] dtensor: size=(0,) torch.Size([0]) | offset=(2,)
[Rank 3] dtensor: size=(0,) torch.Size([0]) | offset=(2,)
[Rank 4] dtensor: size=(0,) torch.Size([0]) | offset=(2,)
[...]

PR #161652, opened 2025-08-27
Issue Fix: incorrect offset computation for double-sharded tensors in compute_local_shape_and_global_offset

davidizzle:
### Fix incorrect offset computation for double-sharded tensors in compute_local_shape_and_global_offset

**Fixes Issue #161368**

This PR fixes a bug in torch.distributed.tensor._utils.compute_local_shape_and_global_offset where offsets were computed incorrectly when a tensor was sharded multiple times along the same dimension (e.g., placements like [Shard(0), Shard(0)] on a 2D mesh).
Previously, the function only handled single-dimension sharding correctly. In the double-shard case, it produced bogus offsets (e.g., many ranks reporting identical offsets or non-contiguous slices). This broke correctness when redistributing tensors across uneven meshes.

I have slightly changed the logic so as to calculate a composite index in a way that is tolerant to this. 
In essence, we first need to:
1.  Coalesce sharding by grouping placements by tensor dimensions, in case they are repeated (as in the issue referenced)
2. Compute composite indices across mesh axes
3. Correctly assign local shapes and global offsets even with multiple shards on the same dimension

All the rest has remained the same. 
I have tentatively added a quick regression test which reproduces the issue, aiming to ensure that local shards align with the expected slices of the global tensor, preventing future regressions.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @tianyu-l @XilunWu @SherlockNoMad @ezyang @kwen2501 @jiayisuse @osalpekar @lw @angelayi

### DIFF
--- a/test/distributed/tensor/test_utils.py
+++ b/test/distributed/tensor/test_utils.py
@@ -272,6 +272,23 @@ class UtilTest(DTensorTestBase):
                 )
 
     @with_comms
+    def test_regression_double_shard_offset(self):
+        # Repro for GitHub issue #161368, unit test should pass after fix
+        mesh = init_device_mesh(self.device_type, (2, 4))  # uneven 2D mesh
+        global_tensor = torch.arange(2)
+        placements = [Shard(0), Shard(0)]
+        dtensor = distribute_tensor(global_tensor, mesh, placements)
+
+        local_size, global_offset = compute_local_shape_and_global_offset(
+            global_tensor.size(), mesh, placements
+        )
+
+        # Expect correct slicing (no bogus offsets like before)
+        dim = self._compute_start_end_offsets(global_offset, local_size, 1)
+        d0s, d0e = dim[0]
+        self.assertEqual(dtensor.to_local(), global_tensor[d0s:d0e])
+
+    @with_comms
     def test_fsdp_tp_meta_compute(self):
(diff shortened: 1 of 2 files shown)


davidizzle:
Thanks, will fix quickly

ezyang:
@zpcore This is redistribute related, I wonder if you are willing to take a look. This is a good example of "hmm, a proof would be good"

test/distributed/tensor/test_utils.py
```diff
@@ -271,6 +271,23 @@ def test_compute_local_shape_and_global_offset_2D(self):
                     global_tensor[dim0_start:dim0_end, dim1_start:dim1_end],
                 )
 
+    @with_comms
+    def test_regression_double_shard_offset(self):
```
zpcore, line 275: Hmm, the test passed w/o the logic change in this PR. I left a comment here just to double check: https://github.com/pytorch/pytorch/issues/161368#issuecomment-3269101582

davidizzle: Yes, I noticed this — I've been meaning to get back to this test, but haven't had the time yet. I'll try to repro issue #161368

PR #161658, opened 2025-08-27
Error when TORCH_STABLE_ONLY is defined in TensorBase.h

mikaylagawarecki:
* __->__ #161658


aten/src/ATen/core/TensorBase.h
```diff
@@ -1,5 +1,10 @@
 #pragma once
 
+#ifdef TORCH_STABLE_ONLY
```
albanD, line 11: Please add a comment here explaining what this flag is for and link to the test to see how we expect it to be used.

albanD (approved):
nit on adding a comment, Sounds like a good basic solution!

PR #161659, opened 2025-08-27
Guard config copy for pickle errors

oulgen:
* __->__ #161659

Differential Revision: [D81168335](https://our.internmc.facebook.com/intern/diff/D81168335)

### DIFF
--- a/torch/_dynamo/utils.py
+++ b/torch/_dynamo/utils.py
@@ -1557,9 +1557,14 @@ def _scrubbed_inductor_config_for_logging() -> Optional[str]:
 
     keys_to_scrub: set[Any] = set()
     inductor_conf_str = None
-    inductor_config_copy = (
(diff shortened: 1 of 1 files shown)


zou3519 (approved):
can we add a test? Make a custom pass with a non-serializable object and put it on the config?

oulgen:
> can we add a test? Make a custom pass with a non-serializable object and put it on the config?

so.. this is actually why it took so long, i couldn't figure out how to write a test.
Config gets copy.deepcopy'ed so, I wrote a class like this
```
class PickleCopy:
    def __init__(self, data):
        self.data = data

    def __copy__(self):
        return pickle.loads(pickle.dumps(self))

    def __deepcopy__(self, memo):
        return pickle.loads(pickle.dumps(self))
```

and then added this to config but couldnt repro it. Maybe there's some fbcode version specific thing?

PR #116473, opened 2023-12-27
Export get_printoptions and printoptions

jackm321:
Summary: Expose `torch.get_printoptions` and `torch.printoptions`

Test Plan:
```
print(f"torch.get_printoptions: {torch.get_printoptions}")
print(f"torch.printoptions: {torch.printoptions}")
```
prints:
```
torch.get_printoptions: <function get_printoptions at 0x7fadd2e18160>
torch.printoptions: <function printoptions at 0x7fadd2e18310>
```

Reviewed By: yuhc

Differential Revision: D52239250

PR #161665, opened 2025-08-27
Make get_printoptions and printoptions public

i-aki-y:
This PR proposes to make the following three related functions consistently public:

```
set_printoptions
get_printoptions
printoptions
```

Currently, only set_printoptions is exposed publicly, which may be confusing to users who likely expect all three to have the same visibility.
This PR aligns the visibility of `get_printoptions` and `printoptions` with `set_printoptions`, making all three publicly accessible.

A similar PR was previously submitted but later closed due to inactivity:

#116473

For this work, I referred to the following documentation: https://github.com/pytorch/pytorch/wiki/Public-API-definition-and-documentation

Note: `set_printoptions` remains unchanged in this PR. As a result, its `__module__` differs from the other two:

```
set_printoptions.__module__  # → "torch._tensor_str"
get_printoptions.__module__  # → "torch"
printoptions.__module__      # → "torch"
```

Personally, I think it would be cleaner that they have the same `__module__` values. Should we consider updating `set_printoptions.__module__` as well?


albanD (approved):
Thanks!

PR #161670, opened 2025-08-28
[dynamo] Graph break on on user-defined class in compiled region

### LINKED ISSUE #161562: `set_fullgraph(False)` actually skips the frame if a graph break happens
### 🐛 Describe the bug

I think I might be getting this wrong, but reporting just in case.

The code below is a simple reproducer that would produce a graph break in the absence of `set_fullgraph(False)`. I expect set_fullgraph to split the graph into two parts (`y := x + 10` and `f.x - y`), but once a graph break occurs, the frame is skipped and the code runs on eager.

```python
import torch

@torch.compile(backend="eager", fullgraph=True)
def fn(x):
    y = x + 10

    with torch._dynamo.set_fullgraph(False):
        class Foo:
            def __init__(self, x):
                self.x = x

    f = Foo(x)
    return f.x - y

x = torch.tensor([1.0])
y = fn(x)
print(y)
```

It looks like it is missing a `@break_graph_if_unsupported` in the `LOAD_BUILD_CLASS` opcode?

### Versions

PyTorch main branch

cc @chauhang @penguinwu @voznesenskym @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @amjames @Lucaskabela

rtimpe:
* __->__ #161670

Currently, user-defined classes inside of a compiled frame will cause the whole
frame to be skipped by dynamo.  This change defers the Unsupported exception
until the __build_class__ builtin is actually called, which allows a graph break
to be inserted.


test/dynamo_expected_failures/CPython313-test_bool-BoolTest.test_blocked
rtimpe, line 1: This test (`test_small_stability`, which was deleted) was already passing before this PR.  I'm not sure why

test/test_fx.py
```diff
@@ -198,6 +198,7 @@ def side_effect_func(x: torch.Tensor):
     print(x)
 
 
+@skipIfTorchDynamo("Not suitable for dynamo")
```
rtimpe: Some of these are failing during setup, but many define custom classes, so I think it's just easier to skip them all.  If there's a desire to run these tests under dynamo, I can look at them more carefully

guilhermeleobas:
@williamwen42, can you review this one? Rob implemented the `LOAD_BUILD_CLASS` opcode, and we're now graph breaking when builtins.__build_class__() actually happens. This should avoid the frame being skipped because of `set_fullgraph(False)` region.

@rtimpe, we need to make sure the CPython tests are actually being traced by Dynamo and not skipped.

williamwen42 (approved):

test/test_fx.py
guilhermeleobas: Put these failures on `test/dynamo_expected_failures`

guilhermeleobas:
@rtimpe can you rebase this PR? You can do this with `@pytorchbot rebase` command

test/test_fx.py
rtimpe: Done.  Note that two of the tests had to go in `dynamo_skips` though (one was passing despite being expected to fail and the other was passing on python 3.10 but not 3.13)

guilhermeleobas (approved):

PR #161677, opened 2025-08-28
[SymmMEM] Fix test_empty_strided_p2p_persistent

fegin:
* __->__ #161677
* #161676

test_empty_strided_p2p_persistent allocates persistent symm memory tensors. However, it uses the same alloc_id for different tests, which could cause troubles if these tests are ran under the same process. This PR fixes the issue by using a different alloc_id for different test.

https://github.com/pytorch/pytorch/pull/161668 should also fix the issue but we can land this PR for a safer test.

cc @H-Huang @awgu @wanchaol @fduwjj @wz337 @wconstab @d4l3k @pragupta


test/distributed/test_symmetric_memory.py
```diff
@@ -152,24 +152,27 @@ def test_empty_strided_p2p_persistent(self, set_device: bool) -> None:
         self._init_process(set_device)
         enable_symm_mem_for_group(dist.group.WORLD.group_name)
 
+        alloc_id = 42 + int(set_device)
```
kwen2501: nit: we can make `alloc_id` a random number, in case `int(set_device)` looks weird

kwen2501 (approved):

PR #161678, opened 2025-08-28
Add ciflow/vllm to vLLM commit hash update PR(s)

huydhn:
As it should be, otherwise, PR(s) like https://github.com/pytorch/pytorch/pull/161121 were merged without the signals it needed.


atalman (approved):

huydhn:
Caught by @clee2000 in https://github.com/pytorch/pytorch/pull/163353, so let's close this and keep the new behavior

PR #161687, opened 2025-08-28
[ROCm] revamp miopen integration

ethanwee1:
Update sources under ATen/miopen and ATen/native/miopen to align with best practices. Avoid reshape_ calls inside backward operations.

cc @jeffdaily @sunway513 @jithunnair-amd @pruthvistony @ROCmSupport @dllehr-amd @jataylo @hongxiayang @naromero77amd @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @albanD


jeffdaily:
> Why skipping PR sanity check is ok here? We would expect a script to repro the change/codemod and regular file move usually. cc @seemethere

Did I add the wrong label?  PR SIZE is 2598, so over the 2000 limit.  This PR is very difficult to break into smaller pieces, and a big chunk of the diff is removing miopen depthwise redundant code.

jeffdaily (approved):

PR #161688, opened 2025-08-28
export: add explicit decomposition for aten.expand_copy and unit test

### LINKED ISSUE #161080: torch.export fails for expand_copy when implicit=True
### 🐛 Describe the bug

```python
import torch
import torch.nn as nn

class ExpandModel(nn.Module):
    def __init__(self):
        super(ExpandModel, self).__init__()

    def forward(self, x, implicit):
        return torch.expand_copy(x, [3, 3], implicit=implicit)

model = ExpandModel()
x = torch.ones([3])

# this succeeds
model(x, False)
model(x, True)
torch.export.export(model, (x, False))

# this fails
# TypeError: expand() got an unexpected keyword argument 'implicit'
torch.export.export(model, (x, True))
```

### Versions

nightly

cc @chauhang @penguinwu @avikchaudhuri @gmagogsfm @zhxchen17 @tugsbayasgalan @angelayi @suo @ydwu4

albertw7711:
Fixes #161080 
torch.export.export fails with TypeError: expand() got an unexpected keyword argument 'implicit' when calling torch.expand_copy(..., implicit=True). This happened because expand_copy = _make_copy_from_view(aten.expand) register aten. expand as the decomposition path for aten.expand_copy, which doesn’t accept the implicit argument.

I have added an explicit a decomposition for aten.expand_copy in torch/_decomp/decompositions.py to ignore the implicit argument, and a simple unit test to demonstrate the bug being fixed.


albertw7711:
hi sorry I am still not very familiar with the pr process. I will fix the other issues when I wake up tomorrow. Thanks.

torch/_decomp/decompositions.py
```diff
@@ -770,6 +770,19 @@ def slice_forward(
         return self.as_strided(sizes, strides, storage_offset)
 
 
+@register_decomposition(aten.expand_copy)
+@out_wrapper()
+def expand_copy(self, size, *, implicit: bool = False):
```
angelayi: [expand](https://github.com/pytorch/pytorch/blob/a8270dd1248569e667861b2a128417537228a899/aten/src/ATen/native/native_functions.yaml#L2645) also takes in implicit as a flag, so I think we should just add the flag to its decomposition

test/export/test_expand_copy.py
```diff
@@ -0,0 +1,27 @@
...
+import pytest
+
+class ExpandModel(nn.Module):
+    def __init__(self):
+        super().__init__()
+
+    def forward(self, x, implicit):
+        return torch.expand_copy(x, [3, 3], implicit=implicit)
+
+def test_expand_copy_export_handles_implicit_true():
```
angelayi: can we add this to test/export/test_export.py instead of creating a new file?

albertw7711: done!

torch/_decomp/decompositions.py
albertw7711: hi I have spent several hours on this trying to trace down the root cause and attempting various fixes. However, it seems that the issue lies within python function binding. The schema is working fine as modifying it does not change anything and it was propagated properly throughout all the modules I have traced through. It was very interesting as it seems that the schema is effective in identifying unrecognized argument, but if the implicit keyword is used, the process would go further and break because it just randomly stops following the schema.

The latest commit I have submitted just straight up dereference the implicit argument since it's not used by aten anyway.

I hope this helps, sorry for not being able to do more.

can-gaa-hou: I think all you need to do is just add the `implicit` flag here and set a default value for it
https://github.com/pytorch/pytorch/blob/4acdbb8311f760513556e2e4fdd7bfd88c225e52/torch/_refs/__init__.py#L3016-L3019

Both `expand` and `expand_copy` accept `implicit` as an input, and since `expand` makes a copy view from `expand_copy`, we can fix it in the same place.

If there is another issue, you can paste the output here or start a new issue in the community :)

albertw7711: thank you so much @can-gaa-hou! This fixes the problem completely and I will make sure to learn from this experience.
@angelayi @can-gaa-hou I think the pr can be merged now to close [#161688](https://github.com/pytorch/pytorch/issues/161080).

test/export/test_export.py
```diff
@@ -13647,6 +13647,28 @@ def forward(self, x, y):
...
+        # These should succeed
+        model(x, False)
+        model(x, True)
+        torch.export.export(model, (x, False))
+
+        # This used to fail with TypeError; now should succeed
+        try:
+            torch.export.export(model, (x, True))
+        except TypeError as e:
+            self.fail(f"expand_copy export with implicit=True raised TypeError: {e}")
```
angelayi: i think we can delete this part :)

test/export/test_export.py
```diff
@@ -13647,6 +13647,28 @@ def forward(self, x, y):
...
+            def forward(self, x, implicit):
+                return torch.expand_copy(x, [3, 3], implicit=implicit)
+
+        model = ExpandModel()
+        x = torch.ones([3])
+
+        # These should succeed
+        model(x, False)
+        model(x, True)
+        torch.export.export(model, (x, False))
```
angelayi:
```suggestion
        export(model, (x, False))
```

angelayi (approved):
thanks!

can-gaa-hou (approved):

albertw7711:
@angelayi hi I really hope my first pr here can be accepted I am planning on making future contributions as well so if the code is still not up to standard please let me know much appreciated! Please approve the review otherwise.

angelayi (approved):

albertw7711:
Thanks everyone for being supportive! I will start working on a new issue.

PR #161690, opened 2025-08-28
[Intel GPU] Fix XPU SDPA default priority_order UT fail

### LINKED ISSUE #161483: DISABLED test_default_priority_order_xpu (__main__.TestSDPAXpuOnlyXPU)
Platforms: xpu

This test was disabled because it is failing on main branch ([recent examples](https://torch-ci.com/failure?failureCaptures=%5B%22test_transformers.py%3A%3ATestSDPAXpuOnlyXPU%3A%3Atest_default_priority_order_xpu%22%5D)).

LuFinch:
Fixes #161483

When the whole `test/test_transformers.py` file is run, the case `test_default_priority_order` can pass because other xpu cases would call SDPA so that the priority order is set by https://github.com/pytorch/pytorch/blob/eec876deb659fe667aac2d97a48d7451c3e88dee/aten/src/ATen/native/mkldnn/xpu/Attention.cpp#L98-L112

However, when the case `test_default_priority_order` is run separately, the priority order is unset so that this case would fail. This PR fix this case.

### DIFF
--- a/test/test_transformers.py
+++ b/test/test_transformers.py
@@ -4269,6 +4269,12 @@ class TestSDPAXpuOnly(NNTestCase):
     def test_default_priority_order(self, device):
         # The default priority order of xpu is overrideable, math, flash, efficient, cudnn
         # For xpu backend, we need to make sure that overrideable > math > flash
+        dtype = torch.bfloat16
+        shape = SdpaShape(1, 1, 1, 1)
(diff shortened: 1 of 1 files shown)


guangyey (approved):
Good catch!

drisspg (approved):

PR #161634, opened 2025-08-27
fastpath type Tensor in THPVariable_NewWithVar

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162220
* #162219
* #161692
* __->__ #161634
* #161633
* #161595
* #161591

It is cheap to do an exact check against Tensor and much faster when it works (PyType_IsSubtype does not have this fastpath, I checked [source](https://github.com/python/cpython/blob/9ee0214b5dd982ac9fbe18dcce0e8787456e29af/Objects/typeobject.c#L2889)). Spot-checked in perf on detach-DTensor-in-a-loop benchmark; small win but clear.

Differential Revision: [D81530101](https://our.internmc.facebook.com/intern/diff/D81530101)

PR #161633, opened 2025-08-27
Avoid redundant PyTuple_GetSize call in _maybe_handle_torch_function

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162220
* #162219
* #161692
* #161634
* __->__ #161633
* #161595
* #161591

py::args::size() calls PyTuple_GetSize. Compiler can't know the two calls will always return the same result, so we have to consolidate them ourselves.

Differential Revision: [D81530096](https://our.internmc.facebook.com/intern/diff/D81530096)

cc @EikanWang @jgong5 @wenzhe-nrv @sanchitintel

PR #161694, opened 2025-08-28
Remove unnecessary asserts in _correct_storage_aliasing

swolchok:
* #162220
* #162219
* #162218
* __->__ #161694
* #161693
* #161692
* #161634
* #161633
* #161596
* #161595
* #161591
* #161590
* #161586
* #161466

These don't seem necessary to get reasonable errors:
- func was already passed to get_alias_info earlier in the calling function
- all we require for args and outs is the ability to index into them, which will cleanly throw if it fails anyway

Differential Revision: [D81530108](https://our.internmc.facebook.com/intern/diff/D81530108)


torch/utils/_python_dispatch.py
```diff
@@ -526,9 +526,6 @@ def _correct_storage_aliasing(func, schema_info, args, outs):
     It does this by unsafely overwriting the storage field of the output tensor
     to be the same storage as the input.
     """
-    assert isinstance(func, torch._ops.OpOverload)
```
Skylion007, line 528: If we are going to remove this, can we at least add back the typing info to func signature?

swolchok: add back? it was never there. I can add it and see if linters complain though

Skylion007: @swolchok Yeah, but the typing would have at least been partially enabled for the function scope. This is useful info and we should add it the type def and see if the linters complain.

swolchok:
@swolchok has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81530108).

PR #161695, opened 2025-08-28
Port OpSchema.__post_init__ and OpSchema._recompute_comparison_key to C++

swolchok:
* #163030
* #162990
* #162968
* #162508
* __->__ #161695

I initially didn't see good results porting this, but it was apparently because of pybind11 function calling overhead. (pybind11's object-handling primitives seem fine enough.) I'm interested in setting up nanobind, but this demonstrates it's not blocking.

Differential Revision: [D81530102](https://our.internmc.facebook.com/intern/diff/D81530102)


swolchok:
@swolchok has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81530102).

torch/csrc/autograd/python_variable.cpp
```diff
@@ -854,6 +854,226 @@ static PyObject* THPVariable_make_dtensor(
...
+  if (dtensor_interned_strings.s == nullptr) {                            \
+    return false;                                                         \
+  }
+
+  FOR_EACH_DTENSOR_INTERNED_STRING(INTERN_DTENSOR_STRING);
+#undef INTERN_DTENSOR_STRING
+  return true;
+}
+
+static bool DTensor_OpSchema_recompute_comparison_key_impl(
```
ezyang, line 939: don't forget to xref with the duped Python code

ezyang: oh nvm you deleted the Python code

torch/csrc/autograd/python_variable.cpp
```diff
@@ -854,6 +854,226 @@ static PyObject* THPVariable_make_dtensor(
...
+    PyObject* self,
+    const py::tuple& args_schema) {
+  py::object static_kwargkey;
+  Py_ssize_t static_argnum;
+  const py::handle self_handle = py::handle(self);
+  const py::handle schema_info =
+      self_handle.attr(dtensor_interned_strings.schema_info);
+  if (PyObject_Not(schema_info.ptr())) {
```
ezyang: Not handling failure here?

Should we make a mini-function binding library that translates -1 into exception raises? Or would that cause the same slowdown pybind11 has?

swolchok: whoops. I couldn't find an efficient pybind11 wrapper for either PyObject_IsTrue or PyObject_Not and I forgot it could fail.

> make a mini-function binding library

doesn't seem like a measured response; pybind11 has wrappers for a lot of potential stuff we might want, just not all of it.

torch/distributed/tensor/_op_schema.py
```diff
@@ -386,14 +392,7 @@ def __str__(self) -> str:
         return f"Op(op={self.op}, args_schema={', '.join(args_schema)} @ mesh: {mesh_shape})"
 
     def __post_init__(self) -> None:
-        has_symints = False
-        for a in self.args_schema:
-            if isinstance(a, DTensorSpec) and a.tensor_meta is not None:
-                if any(isinstance(s, torch.SymInt) for s in a.tensor_meta.shape):
-                    has_symints = True
-                    break
-        self.has_symints = has_symints
```
ezyang, line 395: I suspect the has_symints check here isn't as tight as it could be CC @bdhirsh. One easy improvement is to just test if we are compiling (or fake tensor mode or any of the non eager perf cases) and just bypass this entirely, since we know it cannot occur. If there isn't a way to consistently figure this out quickly we should. (Probably the main issue is  fake tensor value based dispatch?)

torch/csrc/autograd/python_variable.cpp
```diff
@@ -854,6 +854,226 @@ static PyObject* THPVariable_make_dtensor(
...
+  }
+
+  FOR_EACH_DTENSOR_INTERNED_STRING(INTERN_DTENSOR_STRING);
+#undef INTERN_DTENSOR_STRING
+  return true;
+}
+
+static bool DTensor_OpSchema_recompute_comparison_key_impl(
+    PyObject* self,
+    const py::tuple& args_schema) {
```
ezyang, line 941: FWIW this code would probably be obsoleted once you get to computing the key from IValue directly

torch/csrc/autograd/python_variable.cpp
```diff
@@ -854,6 +854,226 @@ static PyObject* THPVariable_make_dtensor(
   END_HANDLE_TH_ERRORS
 }
 
+static py::handle get_dtensor_spec_class() {
+#if IS_PYBIND_2_13_PLUS
```
ezyang, line 858: Oh no, what's going on LOL

torch/csrc/autograd/python_variable.cpp
```diff
@@ -854,6 +854,226 @@ static PyObject* THPVariable_make_dtensor(
...
+                                                        .attr("distributed")
+                                                        .attr("tensor")
+                                                        .attr("_dtensor_spec")
+                                                        .attr("DTensorSpec"))
+                                             .release();
+  return dtensor_spec_class;
+#endif
+}
+
+static bool arg_type_tensor_or_tensor_list_like(py::handle arg) {
```
ezyang, line 881: Technically dtensor_spec? I guess you kept consistent with old

ezyang (commented):
Do you want a fine grained review for memory mgmt and cpython error handling? Ow seems fine

torch/csrc/autograd/python_variable.cpp
swolchok: This seems to be a required general pattern, you can see it elsewhere in the codebase. Personally I would like to just force using pybind11 3.x (to which we upgraded recently) but I assume Linux distros are trying to use their packaged pybind? It's header-only so that doesn't seem high-leverage to me, but then again I don't maintain a Linux distribution or even any packages for one.

swolchok:
> Do you want a fine grained review for memory mgmt and cpython error handling? Ow seems fine

I think I have updates queued for this PR

torch/csrc/autograd/python_variable.cpp
swolchok: I think I'm fine with building scaffolding to support incremental development and migration.

swolchok:
python test/distributed/tensor/test_dtensor_compile.py TestDTensorCompile.test_dtensor_different_gradient_placement is succeeding locally. Hopefully that means it will start succeeding when I update.

torch/csrc/autograd/python_variable.cpp
```diff
@@ -854,6 +854,238 @@ static PyObject* THPVariable_make_dtensor(
...
```
ezyang, line 881: Do you have objections to keeping the old Python implementation around in comments as a reference, or do you think it will bitrot immediately?

ezyang:
How good do you think the preexisting testing here is? We are looking at some (orthogonal) parts of DTensor where the testing is definitely NOT enough and I'm wondering if this applies here or not.

torch/csrc/autograd/python_variable.cpp
```diff
@@ -854,6 +854,238 @@ static PyObject* THPVariable_make_dtensor(
...
+  } else {
+    static_argnum = py::cast<size_t>(
+        schema_info.attr(dtensor_interned_strings.static_argnum));
+    static_kwargkey =
+        schema_info.attr(dtensor_interned_strings.static_kwargkey);
+  }
+  c10::SmallVector<py::object, 8> args_to_hash;
+  size_t idx = 0;
+  for (const auto& e : args_schema) {
+    if (idx >= static_argnum || arg_type_tensor_or_tensor_list_like(e)) {
```
ezyang, line 959: reordered cond, nice

torch/csrc/autograd/python_variable.cpp
```diff
@@ -854,6 +854,238 @@ static PyObject* THPVariable_make_dtensor(
...
+        self_handle.attr(dtensor_interned_strings.kwargs_schema);
+    if (!PyDict_Check(raw_kwargs_schema.ptr())) {
+      PyErr_SetString(PyExc_TypeError, "self.kwargs_schema must be a dict!");
+      return false;
+    }
+    py::tuple kwargs_to_hash(static_kwargkey_list.size());
+    int idx = 0;
+    auto kwargs_schema = py::reinterpret_borrow<py::dict>(raw_kwargs_schema);
+    for (const auto& k : static_kwargkey_list) {
+      PyObject* item = PyDict_GetItem(kwargs_schema.ptr(), k.ptr());
```
ezyang: safer to call PyDict_GetItemWithError

torch/csrc/autograd/python_variable.cpp
```diff
@@ -854,6 +854,238 @@ static PyObject* THPVariable_make_dtensor(
...
+      if (item) {
+        kwargs_to_hash[idx++] = py::reinterpret_borrow<py::object>(item);
+      } else {
+        kwargs_to_hash[idx++] = py::none();
+      }
+    }
+    PyObject* comparison_key = PyTuple_Pack(
```
ezyang: Should test that PyTuple_Pack succeeded, and ditto below

torch/csrc/autograd/python_variable.cpp
```diff
@@ -854,6 +854,238 @@ static PyObject* THPVariable_make_dtensor(
...
+  }
+  py::tuple args_to_hash_tup(args_to_hash.size());
+  for (const auto idx : c10::irange(args_to_hash.size())) {
+    args_to_hash_tup[idx] = std::move(args_to_hash[idx]);
+  }
+  if (!static_kwargkey.is_none()) {
+    if (!PyList_Check(static_kwargkey.ptr())) {
+      PyErr_SetString(
+          PyExc_TypeError, "self.schema_info.static_kwargkey must be a list!");
+      return false;
```
ezyang, line 978: It might be more uniform with the rest of the codebase if we just raised exceptions when this happens, but I guess this is not too difficult to get right.

torch/csrc/autograd/python_variable.cpp
```diff
@@ -854,6 +854,238 @@ static PyObject* THPVariable_make_dtensor(
...
+  }
+  py::tuple args_schema = py::reinterpret_borrow<py::tuple>(raw_args_schema);
+  if (!DTensor_OpSchema_recompute_comparison_key_impl(self, args_schema)) {
+    return nullptr;
+  }
+
+  const auto dtensor_spec_class = get_dtensor_spec_class();
+  bool has_symints = false;
+  for (const auto& a : args_schema) {
+    if (Py_TYPE(a.ptr()) != (PyTypeObject*)(dtensor_spec_class.ptr()) &&
```
ezyang, line 1059: direct type test first, interesting that this helps!

ezyang (approved):
I reviewed logic equivalence and CPython API usage, I did NOT review memory management

torch/csrc/autograd/python_variable.cpp
swolchok: commented-out code generally bitrots. if we want it around we would need to make it modally enabled and add tests

torch/csrc/autograd/python_variable.cpp
swolchok: oh wow I didn't fully realize what was going on with PyDict_GetItem

torch/csrc/autograd/python_variable.cpp
swolchok: yeah I briefly considered upstreaming extensions to PyFloat_Check etc. to first do the exact checks because they're way cheaper. the CPython API functions can't be inlined so there's function call overhead and they also don't themselves start with the direct test.

PR #161700, opened 2025-08-28
[ROCm] OffsetCalc Unroll Optimization

amd-hhashemi:
Our compiler is generating inefficient code for the offsetCalc in certain situations.
The root-cause for this needs to be identified. For now specialized unrolling based on 'dims' notably helps perf.

### DIFF
--- a/aten/src/ATen/cuda/detail/OffsetCalculator.cuh
+++ b/aten/src/ATen/cuda/detail/OffsetCalculator.cuh
@@ -47,7 +47,7 @@ struct OffsetCalculator {
     offset_type offsets;
 
 #if defined(USE_ROCM)
-    if (dims <= 3) {
+    if ((dims > 0) && (dims <= 3)) {
       auto divmod = sizes_[0].divmod(linear_idx);
 #pragma unroll
       for (int arg = 0; arg < NARGS; arg++)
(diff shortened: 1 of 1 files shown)


aten/src/ATen/cuda/detail/OffsetCalculator.cuh
```diff
@@ -45,6 +45,30 @@ struct OffsetCalculator {
 
   C10_HOST_DEVICE offset_type get(index_t linear_idx) const {
     offset_type offsets;
+
+#if defined(USE_ROCM)
+    if (dims <= 3) {
```
jerrymannil: i think it should be `(dims > 0  && (dims <= 3)`
if dims = 0, then we should fallback to default case

aten/src/ATen/cuda/detail/OffsetCalculator.cuh
```diff
@@ -45,6 +45,24 @@ struct OffsetCalculator {
...
+
+#if defined(USE_ROCM)
+    if ((dims > 0) && (dims <= 2)) {
+      auto divmod = sizes_[0].divmod(linear_idx);
+#pragma unroll
+      for (int arg = 0; arg < NARGS; arg++)
+        offsets[arg] = divmod.mod * strides_[0][arg];
+      if (dims >= 2) {
+        divmod = sizes_[1].divmod(divmod.div);
+#pragma unroll
```
Skylion007, line 57: Do you want to add a comment to link to the compiler issue?

amd-hhashemi: We're investigating what is exactly happening in compiler here. Will provide an issue number later when we've rootcaused.

jeffdaily (approved):

PR #147433, opened 2025-02-19
[Inductor] Avoid tensor slice overflow for large step

DDEle:
Fixes #147071

Currently, if step is a value very close to INT64_MAX, the calculation of slice output length will overflow. This PR tries to fix this problem and thus fix #147071.

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @yf225 @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov

PR #161707, opened 2025-08-28
[indexing] Prevent integer overflow from large step values in C++

thenumberouscode:
Fixes https://github.com/pytorch/pytorch/issues/160868
hmmm, I found an existing fix PR after I've finished this one. For reference, the old PR was
https://github.com/pytorch/pytorch/pull/147433/files. 

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @mlazos @chenyang78 @leslie-fang-intel

### DIFF
--- a/aten/src/ATen/native/TensorShape.cpp
+++ b/aten/src/ATen/native/TensorShape.cpp
@@ -3127,7 +3127,9 @@ Tensor slice(
   }
   auto storage_offset = self.storage_offset() + start_val * strides[dim];
   auto len = end_val - start_val;
-  sizes[dim] = (len + step - 1) / step; // round-up
+  // NB: (len + step - 1) / step is equivalent implementation,
+  // but len + step could overflow if step or len is very large
+  sizes[dim] = (len == 0) ? 0 : (1 + (len - 1) / step); // safely round up
   strides[dim] *= step;
 
   Tensor result;
--- a/test/inductor/test_torchinductor.py
+++ b/test/inductor/test_torchinductor.py
@@ -4343,6 +4343,31 @@ class CommonTemplate:
         self.assertEqual(torch.compile(fn)(x1, y), fn(x1, y))
         self.assertEqual(torch.compile(fn)(x2, y), fn(x2, y))
 
+    def test_slice_copy(self):
+        class Model(nn.Module):
+            def __init__(self, start=449, step=(2**63 - 1)):
+                super().__init__()
+                self.start = start
+                self.step = step
+
+            def forward(self, x: torch.Tensor):
+                sliced = torch.slice_copy(
+                    x, dim=0, start=self.start, end=None, step=self.step
+                )
+                return torch.reciprocal(sliced)
+
+        with config.patch({"implicit_fallbacks": True}):
+            # bad case
+            self.common(
+                Model(),
+                (torch.randn(875),),
+            )
+            # normal case
+            self.common(
+                Model(step=10),
+                (torch.randn(875),),
+            )
+
     def test_slice1(self):
         def fn(a):
             return (
--- a/torch/_decomp/decompositions.py
+++ b/torch/_decomp/decompositions.py
@@ -759,7 +759,7 @@ def slice_forward(
 
     storage_offset = self.storage_offset() + start_val * strides[dim]
     len = end_val - start_val
-    sizes[dim] = (len + step - 1) // step
+    sizes[dim] = -(len // -step)  # round-up, avoiding overflow
     strides[dim] *= step
 
     if self.is_quantized:


leslie-fang-intel:
cc @CaoE to take a look

leslie-fang-intel (dismissed):

thenumberouscode:
@leslie-fang-intel Just rebased with main. Should we retry the workflow? I've run the failed unit cases in my local env, and all passed

CaoE (dismissed):

mlazos (dismissed):

thenumberouscode:
@mlazos Please run the workflow again, thank.

thenumberouscode:
@mlazos Could you please review the PR again? thx.

thenumberouscode:
@leslie-fang-intel @CaoE  Could you please to review the PR again? thanks.

thenumberouscode:
@leslie-fang-intel @CaoE Could you please to review the PR again? thanks.

test/inductor/test_torchinductor.py
```diff
@@ -4309,6 +4309,42 @@ def fn(x, y):
...
+                (torch.randn(875),),
+            )
+            # normal case
+            self.common(
+                Model(step=10),
+                (torch.randn(875),),
+            )
+
+    def test_slice_dynamic_shape(self):
+        class LinearRel(torch.nn.Module):
```
leslie-fang-intel: This test seems not test with a large step value.

thenumberouscode: No, this test is not designed to validate large step values. 
Its purpose is to verify that the PR functions correctly in dynamic shapes scenarios within torch compile. The test was added in response to a previous workflow failure, see https://hud.pytorch.org/pytorch/pytorch/commit/c140bf217f5ca5071ab9dbc1bcf9d4006242f44a

aten/src/ATen/native/TensorShape.cpp
```diff
@@ -3063,7 +3063,7 @@ Tensor slice(
   }
   auto storage_offset = self.storage_offset() + start_val * strides[dim];
   auto len = end_val - start_val;
-  sizes[dim] = (len + step - 1) / step; // round-up
+  sizes[dim] = (len / step) + (len % step != 0); // safely round-up
```
CaoE: Can we avoid % by using like `auto v = len / step;  auto r = len - v * step;  return v + (r != 0);` for performance ?

CaoE: It is better to add some comments. What if `step` < 0 ? I see the comment: `TODO: support negative strides`.

test/inductor/test_torchinductor.py
thenumberouscode: Remove this test as it's redundant - all tests in this file are automatically run with dynamic shapes.

aten/src/ATen/native/TensorShape.cpp
thenumberouscode: 1. Yes, I have not used % in the latest code.
2. Sure, I added comments,  the torch api doc says step will be positive. so we can ignore the negative steps.

thenumberouscode:
@leslie-fang-intel @CaoE  Ping

thenumberouscode:
@CaoE The workflow failure was due to an unrelated test case, as shown [here](https://github.com/pytorch/pytorch/actions/runs/18483211084/job/53136419431?pr=161707). 

I've rebased onto the latest main branch. Could you please rerun the workflow? Thank you!

torch/_decomp/decompositions.py
```diff
@@ -759,7 +759,7 @@ def slice_forward(
 
     storage_offset = self.storage_offset() + start_val * strides[dim]
     len = end_val - start_val
-    sizes[dim] = (len + step - 1) // step
+    sizes[dim] = -(len // -step)  # round-up, avoiding overflow
```
CaoE, line 762: Can the test cover the relevant changes in both python and cpp?

thenumberouscode: Yes. Both cpp and python code would be covered by the unit test.

CaoE (approved):

thenumberouscode:
@pytorchbot merge
@leslie-fang-intel @CaoE Thanks for the review

leslie-fang-intel (approved):

aten/src/ATen/native/TensorShape.cpp
```diff
@@ -3079,7 +3079,9 @@ Tensor slice(
   }
   auto storage_offset = self.storage_offset() + start_val * strides[dim];
   auto len = end_val - start_val;
-  sizes[dim] = (len + step - 1) / step; // round-up
+  // NB: (len + step - 1) / step is equivalent implementation,
+  // but len + step could overflow if step or len is very large
+  sizes[dim] = (len == 0) ? 0 : (1 + (len - 1) / step); // safely round up
```
jansel, line 3132: Why do we need the `len==0` check here?

Is there a test for what this change is fixing?  I only see an inductor test.

torch/_decomp/decompositions.py
jansel: This pattern is incorrect in C++ because of the different round down versus round to zero semantics.  I want to double check we have test coverage for this.

jansel (changes requested):

aten/src/ATen/native/TensorShape.cpp
thenumberouscode: The inductor test can cover this C++ code.
slice_copy operation would decompose into two separate operations: slice and copy. The slice operation would be invoked as torch.aten.ops.slice(xxx) in the inductor-generated code. The code shown above is part of the slice operation's implementation.

jansel: I still don't understand what this change is fixing.  Isn't everything int64 here?  If you revert this change (but not the other one) does the test you added fail?

torch/_decomp/decompositions.py
jansel: I tested this locally, and the added tests still pass even if I revert this change.  Please add a test for what this is fixing.

Interestingly it looks like the other change *is* needed.  Though I'm a bit confused at how this is an integer overflow issue.  How is `len` getting close to INT64_MAX with such small input tensors?

jansel (changes requested):

aten/src/ATen/native/TensorShape.cpp
thenumberouscode: @jansel Let me explain the change in detail:
The current C++ code implements the ceiling division of len by step using the expression (len + step - 1) / step, where both len and step are of type int64_t. However, when either len or step is sufficiently large (for instance, close to INT64_MAX), the addition len + step may overflow. This overflow can lead to unexpected behavior, as demonstrated in the [issue](https://github.com/pytorch/pytorch/issues/160868).

In that specific case, with len being 426 and step equal to 2^63 - 1, their sum overflows the int64_t range. Consequently, (len + step - 1) / step evaluates to 0 instead of the correct value of 1. As a result, the slice operation returns an empty tensor.

In contrast, the equivalent Python expression (len + step - 1) // step works correctly because Python integers do not overflow. This discrepancy causes a conflict: while the Python result instructs the Inductor to access the tensor, the slice operation in C++ returns null, ultimately leading to a core dump.

torch/_decomp/decompositions.py
thenumberouscode: 1. Got it, I will investigate why the unit test can pass without the change.
2. len is small indeed, but step is close to INT64_MAX, and we would use `(len + step - 1) / step` to get the dim size of the sliced tensor.

aten/src/ATen/native/TensorShape.cpp
thenumberouscode: Our modification aims to implement a more robust method for computing the ceiling of len / step. Specifically, I'll address your previous questions:

Why do we need the len == 0 check here?

The C++ implementation now uses: `sizes[dim] = (len == 0) ? 0 : (1 + (len - 1) / step);`,The expression 1 + (len - 1) / step is more reliable than (len + step - 1) / step because it avoids potential overflow from addition. The len == 0 check is necessary because the result should always be 0 when len is zero, since ceil(0 / any_value) is always 0. This check also prevents the special case where we would incorrectly add 1 to a negative value when len is 0.

How unit tests verify both C++ and Python code:

The unit tests validate both the C++ implementation and Python behavior through an Inductor test. Although you have pointed out the test did not work as expected. I will paste the code generated by the inductor later.

thenumberouscode:
@jansel Would you be available to discuss this PR on Slack?

PR #161710, opened 2025-08-28
Update the heuristic for AArch64 bgemm

karmeh01:
Updates heuristic for bgmm. Based on [these changes](https://github.com/pytorch/pytorch/compare/main...Mousius:pytorch:bgemm?expand=1)

```
import torch
import time

# Set below to True to use cases selected by only one of the hueristics.
USE_ONLY_DIVERGENT_TEST_CASES = True   
BATCH_SIZES  = [ 1, 8, 32 ]
M_DIMS       = [ 4, 8, 16, 32, 64, 256, 512, 1024, 2048 ]
N_DIMS       = [ 4, 8, 16, 32, 64, 256, 512, 1024, 2048 ]
K_DIMS       = [ 4, 8, 16, 32, 64, 256, 512, 1024, 2048 ]
ITERS        = 40
DTYPE        = torch.bfloat16

def old_heuristic(m, n, k):
    is_above_min_dims = m > 8 and n > 8 and k > 8
    is_above_min_size = m * n * k <= 16 * 16 * 16
    return is_above_min_dims and is_above_min_size

def new_heuristic(m, n, k):
    is_above_min_dims = m >= 8 and n >= 8 and k >= 8
    heuristic = m <= 8 or k <= 8 or n >= 1024
    return heuristic and is_above_min_dims

def generate_test_cases():
   test_cases = []
   for b in BATCH_SIZES:
       for m in M_DIMS:
           for n in N_DIMS:
                   for k in K_DIMS:
                       if USE_ONLY_DIVERGENT_TEST_CASES:
                           if old_heuristic(m, n, k) != new_heuristic(m, n, k):
                               test_cases.append([b, m, n, k])
                       else:
                           test_cases.append([b, m, n, k])
   return test_cases

def test(x, y):
   for _ in range(5):
       torch.bmm(x, y)
   perf = 0.0
   for _ in range(ITERS):
       start = time.time()
       torch.bmm(x, y)
       end = time.time()
       perf += (end - start) / ITERS
   return perf

def main():
    device = torch.device("cpu")

    print(f"{'b':<10}{'m':<10}{'n':<10}{'k':<10}{'time (s)':10}")
    cumulative_mean_time = 0.0
    last_b = None

    for b, m, n, k in generate_test_cases():
        x = torch.randn((b, m, n), dtype=DTYPE, device=device)
        y = torch.randn((b, n, k), dtype=DTYPE, device=device)
        mean_time = test(x, y)
        cumulative_mean_time += mean_time
        print(f"{b:<10}{m:<10}{n:<10}{k:<10}{mean_time:10.3e}")
    print(f"Cumulative mean time = {cumulative_mean_time:.4f} s")


if __name__ == "__main__":
   main()
```

From the script we see that cumulative mean time from all test cases (at 16 threads) is:

2.0007 s for the old heuristic
0.7480 s for the new heuristic

@aditew01 

### DIFF
--- a/.ci/docker/common/install_openblas.sh
+++ b/.ci/docker/common/install_openblas.sh
@@ -4,7 +4,7 @@
 set -ex
 
 cd /
-git clone https://github.com/Mousius/OpenBLAS.git -b "bgemm-optimisation" --depth 1 --shallow-submodules
+git clone https://github.com/OpenMathLib/OpenBLAS.git -b "${OPENBLAS_VERSION:-v0.3.30}" --depth 1 --shallow-submodules
 
 OPENBLAS_CHECKOUT_DIR="OpenBLAS"
 OPENBLAS_BUILD_FLAGS="
--- a/.ci/docker/common/install_openblas.sh
+++ b/.ci/docker/common/install_openblas.sh
@@ -4,9 +4,10 @@
 set -ex
 
 cd /
-git clone https://github.com/OpenMathLib/OpenBLAS.git -b "${OPENBLAS_VERSION:-v0.3.30}" --depth 1 --shallow-submodules
+git clone https://github.com/OpenMathLib/OpenBLAS.git
 
 OPENBLAS_CHECKOUT_DIR="OpenBLAS"
+git -C "${OPENBLAS_CHECKOUT_DIR}" checkout 5e43ba948c3cb35864c7b0953b8dd02374dd3967
 OPENBLAS_BUILD_FLAGS="
 NUM_THREADS=128
 USE_OPENMP=1
--- a/.ci/docker/common/install_openblas.sh
+++ b/.ci/docker/common/install_openblas.sh
@@ -4,7 +4,7 @@
 set -ex
 
 cd /
-git clone https://github.com/OpenMathLib/OpenBLAS.git -b "${OPENBLAS_VERSION:-v0.3.30}" --depth 1 --shallow-submodules
+git clone https://github.com/Mousius/OpenBLAS.git -b "bgemm-optimisation" --depth 1 --shallow-submodules
 
 OPENBLAS_CHECKOUT_DIR="OpenBLAS"
 OPENBLAS_BUILD_FLAGS="
--- a/aten/src/ATen/native/CPUBlas.cpp
+++ b/aten/src/ATen/native/CPUBlas.cpp
@@ -20,6 +20,14 @@ extern "C" void dgemm_(char *transa, char *transb, int *m, int *n, int *k, doubl
 extern "C" void sgemm_(char *transa, char *transb, int *m, int *n, int *k, float *alpha, const float *a, int *lda, const float *b, int *ldb, float *beta, float *c, int *ldc);
 extern "C" void cgemm_(char *transa, char *transb, int *m, int *n, int *k, void *alpha, const void *a, int *lda, const void *b, int *ldb, void *beta, void *c, int *ldc);
 extern "C" void zgemm_(char *transa, char *transb, int *m, int *n, int *k, void *alpha, const void *a, int *lda, const void *b, int *ldb, void *beta, void *c, int *ldc);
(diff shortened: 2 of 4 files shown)


.ci/docker/common/install_openblas.sh
```diff
@@ -4,7 +4,7 @@
 set -ex
 
 cd /
-git clone https://github.com/OpenMathLib/OpenBLAS.git -b "${OPENBLAS_VERSION:-v0.3.30}" --depth 1 --shallow-submodules
+git clone https://github.com/Mousius/OpenBLAS.git -b "bgemm-optimisation" --depth 1 --shallow-submodules
```
michalowski-arm: Is this intended to go with this PR? I assumed we prefer to stick to official releases.

aten/src/ATen/native/mkldnn/Matmul.cpp
```diff
@@ -157,7 +157,7 @@ mkldnn_gemm(
   bool bf32_usable = std::is_same_v<scalar_t, float> && use_mkldnn_bf32_matmul();
   bool tf32_usable = std::is_same_v<scalar_t, float> && use_mkldnn_tf32_matmul();
   if ( !(bf16_usable || fp16_usable || bf32_usable || tf32_usable) ||
-      (m * n * k <= 16 * 16 * 16) || (alpha == 0.0f)) {
+      ((n * m <= 16384) && (n * k >= 4 * 2048 / m)) || (m * n * k <= 16 * 16 * 16) || (alpha == 0.0f)) {
```
michalowski-arm: I think `m * n * k >= 8192` for the second condition would be clearer and still does the same thing. Other than that, maybe it's best to keep the new heuristic under an `AT_MKLDNN_ACL_ENABLED()` guard. Unless I'm reading this wrong, it would currently unintentionally apply to the x86 path as well. 

And I'm wondering if we want to keep the the old `m * n * k <= 16 * 16 * 16` bit as part of the heuristic? Right now it causes some unusual behavior where for example: (16, 16, 16) gets dispatched to OpenBlas, then (20, 20, 16) to ACL, then (24, 24, 16) to OpenBLAS again, and finally (36, 36, 16) to ACL. So we oscillate between the impls as the dimensions increase. I feel that's a bit unusual, I would expect that below certain sizes OpenBlas is faster, but otherwise ACL is better.

michalowski-arm (changes requested):

.ci/docker/common/install_openblas.sh
karmeh01: Fixed 👍

aten/src/ATen/native/mkldnn/Matmul.cpp
karmeh01: The heuristic here was outdated, but yes you are correct. It's been removed in the updated heuristic

aten/src/ATen/native/CPUBlas.cpp
```diff
@@ -353,19 +361,30 @@ void gemm(
...
    }
 #endif
-#if AT_BUILD_WITH_BLAS() && defined(BLAS_HAS_SBGEMM)
+#if AT_BUILD_WITH_BLAS() && (defined(BLAS_HAS_SBGEMM) || defined(BLAS_HAS_BGEMM))
    if (use_blas_gemm(transa, transb, m, n, k, lda, ldb, ldc)) {
       int m_ = m, n_ = n, k_ = k, lda_ = lda, ldb_ = ldb, ldc_ = ldc;
       char transa_ = to_blas(transa), transb_ = to_blas(transb);
-      float alpha_ = alpha, beta_ = beta;
-      int c_size = n_ * m_;
+#if defined(BLAS_HAS_BGEMM)
```
aditew01, line 368: unless we checkout to a more recent commit, we cannot pickup BGEMM kernels in OpenBLAS. Is that going to be a seperate PR ?

PR #161716, opened 2025-08-28
[Lowering] Fix the edge case of empty subgraph split due to dataclass node

yuhuishi-convect:
Summary: Fix the edge case by allowing `call_function` nodes with no deps as graph entry (starter_nodes) in the splitter.

Test Plan:
The test shall pass in the current diff (after fix), and fail in the parent diff (before fix)

```
buck test mode/opt //glow/fb/fx/lowering:split_tests -- test_dataclass_as_graph_entry
```

Rollback Plan:

Differential Revision: D81232435

### DIFF
--- a/test/fx/test_fx_split.py
+++ b/test/fx/test_fx_split.py
@@ -1,10 +1,26 @@
 # Owner(s): ["module: fx"]
 
+import dataclasses
 from collections import defaultdict
 
 import torch
+import torch.fx.passes.operator_support as op_support
+import torch.fx.passes.splitter_base as splitter_base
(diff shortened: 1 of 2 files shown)


ezyang (changes requested):
Needs test

yuhuishi-convect:
> Needs test

there are some internal test suites that cover this file. [D81231937](https://www.internalfb.com/diff/D81231937)
but we are lacking tests in the public repo. 
@ezyang would you recommend porting some internal tests here?

ezyang (approved):

ezyang (approved):

PR #161407, opened 2025-08-25
create torch._grouped_mm fallback path with for loops / bmm

vkuzo:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162159
* #162059
* #161717
* __->__ #161407

Summary:

Creates a fallback path for `torch._grouped_mm`, using the naive for
loop implementation (or bmm).

For the sake of keeping the PR small, this PR only enables SM80+ (CUDA
capability 8.0 and up), since I am testing this on an A100 machine. In
future PRs, we can increase the coverage of the fallback to:
1. float32 and float16, which will extend the GPU coverage
2. cpu

Test Plan:

```bash
pytest test/test_matmul_cuda.py -s -k test_grouped_gemm_2d_3d -x
pytest test/test_matmul_cuda.py -s -k test_grouped_gemm_3d_2d -x
pytest test/test_matmul_cuda.py -s -k test_grouped_gemm_2d_2d -x
pytest test/test_matmul_cuda.py -s -k test_grouped_gemm_3d_3d -x
```

Reviewers:

Subscribers:

Tasks:

Tags:

PR #161717, opened 2025-08-28
move `_grouped_mm` fallback to composite explicit autograd

vkuzo:
* #162159
* #162059
* __->__ #161717
* #161407

Summary:

Moves the `torch._grouped_mm` fallback from cuda-only code to a place
where it can be used by multiple backends. Specifically:
1. make the fallback path and util functions reusable and move them to
   `ATen/native/GroupedMMUtils.h`
2. register a backend-agnostic kernel to composite explicit autograd key
3. refactor the grouped_mm tests to their own test case and enable CPU

At the end of this PR, here is the support matrix:
* CUDA SM90+: fast path with test coverage (no change)
* CUDA SM80+: fallback with test coverage (no change)
* CPU: fallback works, but without test coverage (new in this PR)
* other SM versions and other backends: will probably already work, but
  let's leave this to future PRs
* float32/float16: will probably already work, but let's leave this to
  future PRs

Test Plan:

```bash
pytest test/test_matmul_cuda.py -s -k test_grouped_gemm -x
```

Reviewers:

Subscribers:

Tasks:

Tags:


test/test_matmul_cuda.py
```diff
@@ -298,6 +298,165 @@ def _expand_to_batch(t: torch.Tensor):
...
+
+    @onlyCUDA
+    @skipIfRocm
+    @parametrize("input_dtype", [torch.float32, torch.float16, torch.bfloat16])
+    @parametrize("M", [1, 32, 64])
+    @parametrize("N", [1, 32, 64])
+    @parametrize("K", [1, 32, 64])
+    @parametrize("batch_size", [None, 1, 16])
+    @parametrize("backend", ["cublas", "cublaslt"])
+    def test_mm_bmm_dtype_overload(self, input_dtype, M, N, K, batch_size, backend):
```
vkuzo: I didn't touch this code, this is git incorrectly vizualizing the actual changes, which are moving the `grouped_mm` functions out to a new test class.  If needed, can probably figure out how to undo this to get better history.

test/test_matmul_cuda.py
```diff
@@ -2165,6 +2164,7 @@ def run_test(
             )
 
 instantiate_device_type_tests(TestMatmulCuda, globals(), except_for="cpu")
+instantiate_device_type_tests(TestGroupedMM, globals())
```
vkuzo: technically this is no longer `test_matmul_cuda` since we are including CPU, but IMO this is ok.  If not, I'll move it out.

test/test_matmul_cuda.py
vkuzo: this is annoying, will rework before land

aten/src/ATen/native/GroupedMMUtils.h
```diff
@@ -0,0 +1,162 @@
...
+      // this case is not actually encountered for MoE gemms
+      TORCH_CHECK(offs->size(0) == mat_a.size(0), "matrix batch sizes have to match");
+      out_size = {mat_a.size(1), mat_b.size(1)};
+    } else { // regular bmm
+      TORCH_CHECK(mat_a.size(0) == mat_b.size(0), "batched dimension has to match");
+      out_size = {mat_a.size(0), mat_a.size(1), mat_b.size(-1)};
+    }
+  }
+
+  const auto out_dtype_ = out_dtype.value_or(kBFloat16);
```
drisspg, line 62: is this only on the fast path?

test/test_matmul_cuda.py
drisspg: hmmm, I do think that this is a lil confusing, I think the better thing to do here is maybe just test cuda.

And then add an entry to common_method_invocations (opinfos) to test the generic path

test/test_matmul_cuda.py
```diff
@@ -495,8 +651,8 @@ def test_grouped_gemm_3d_2d(self, strided, a_row_major, b_row_major):
 
     @unittest.skipIf(TEST_WITH_ROCM, "ROCm doesn't support CUTLASS")
     @xfailIfSM100OrLater
-    # TODO(future PR): enable compile for SM80+
-    @unittest.skipIf(not SM90OrLater, "Grouped gemm supported on SM90")
+    # TODO(future PR): enable compile for torch._grouped_mm fallback path
+    @unittest.skipIf(not SM90OrLater, "Grouped gemm with compile supported on SM90")
```
drisspg, line 499: OOC why does this error today, I figured it would fallback to the aten impl

vkuzo: max autotune tries to generate grouped_mm inductor code and fails, likely some more gating is needed somewhere in the inductor codebase

aten/src/ATen/native/GroupedMMUtils.h
```diff
@@ -0,0 +1,162 @@
...
+  const auto last_dim = out_size.size() - 1;
+  const auto alignment = 16 / c10::elementSize(out_dtype_);
+  const int64_t size_padded = (out_size[last_dim] + alignment - 1) / alignment * alignment;
+  std::vector<int64_t> out_stride;
+  if (a_is_2d != b_is_2d) {
+    out_stride = {size_padded, 1};
+  } else {
+    out_stride = {out_size[1] * size_padded, size_padded, 1};
+  }
+  return at::empty_strided(out_size, out_stride, mat_a.options().dtype(out_dtype_));
```
Skylion007, line 77: Shouldn't some of these args be std::moved or something?

aten/src/ATen/native/GroupedMMUtils.h
```diff
@@ -0,0 +1,162 @@
...
+namespace at::native {
+
+bool check_valid_strides_and_return_transposed(const Tensor& mat) {
+  IntArrayRef tensor_strides = mat.strides();
+  IntArrayRef tensor_sizes = mat.sizes();
+  int end_dim = mat.dim() - 1;
+  int alignment = 16 / mat.element_size();
+  TORCH_CHECK(uint64_t(mat.data_ptr()) % 16 ==0, "expected data_ptr to be aligned to 16 bytes\n");
+  if ((tensor_strides[end_dim - 1] == 1) && (tensor_strides[end_dim] >= std::max<int64_t>(1, tensor_sizes[end_dim - 1]))) {
+    TORCH_CHECK(tensor_strides[end_dim] % alignment == 0, "strides should be multiple of 16 bytes");
```
Skylion007, line 26: Shouldn't some of these TORCH_CHECK_VALUE

aten/src/ATen/native/GroupedMMUtils.h
vkuzo: this PR does not relax the bf16 restriction, so keeping it for now.  We can relax in the next PR in the stack.

aten/src/ATen/native/GroupedMMUtils.h
vkuzo: for now this code is moved without changes, seems ok to change in a follow-up

aten/src/ATen/native/GroupedMMUtils.h
vkuzo: for now this code is moved without changes, seems ok to change in a follow-up

test/test_matmul_cuda.py
vkuzo: sg, kept testing to CUDA only in this PR.  We can add an OpInfo further in the stack.

aten/src/ATen/native/Blas.cpp
```diff
@@ -332,4 +333,22 @@ _scaled_mm_cpu(const Tensor& mat_a, const Tensor& mat_b,
   return _scaled_mm_out_cpu(mat_a, mat_b, scale_a, scale_b, bias, scale_result, out_dtype, use_fast_accum, out);
 }
 
+// TODO(vasiliy, future PR): figure out why we need to declare this function, when
```
vkuzo, line 336: would love a tip on this

ngimel: what error are you getting?

ngimel (approved):

aten/src/ATen/native/Blas.cpp
drisspg: I think you might need to add `#include <ATen/ops/_grouped_mm.h>` in the `#ifndef AT_PER_OPERATOR_HEADERS` branch, since there is where the official aten impl lives

drisspg (approved):

aten/src/ATen/native/Blas.cpp
vkuzo: just tried it, no luck: https://github.com/pytorch/pytorch/actions/runs/17463747903/job/49594335338?pr=162159

PR #161740, opened 2025-08-28
Add support for tensor stride load and save

nautsimon:
Summary:
We add a new path for MTIA tensors that will deserialize tensors with strides preserved. As MTIA Allocator is not exposed externally, we cannot simply let MTIA be handled by the typical CPU/CUDA path. To solve this, we add a variant of _rebuild_device_tensor_from_cpu_tensor, that tries to rebuild the tensor with as_strided to preserve strides from the original tensor.

Uses the tests written from the previous diff to verify strides can be saved and loaded on MTIA.

Reviewed By: egienvalue

Differential Revision: D78512965

cc @egienvalue

### DIFF
--- a/torch/_tensor.py
+++ b/torch/_tensor.py
@@ -329,8 +329,20 @@ class Tensor(torch._C.TensorBase):
         materialize_fake_tensors = (
             torch.serialization._serialization_tls.materialize_fake_tensors
         )
-
-        if self.device.type in ["xla", "maia", "mtia"] or (
+        if self.device.type in ["mtia"] or (
+            not torch._C._has_storage(self)
+            and self.device.type == torch._C._get_privateuse1_backend_name()
+        ):
+            if skip_data:
+                raise RuntimeError(
+                    "Cannot serialize tensors on backends with no storage under skip_data context manager"
+                )
+            cpu_tensor = self.cpu()
+            return (
+                torch._utils._rebuild_device_tensor_from_cpu_tensor_with_strides,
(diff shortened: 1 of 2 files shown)


torch/_tensor.py
```diff
@@ -329,8 +329,20 @@ def _reduce_ex_internal(self, proto):
...
+            if skip_data:
+                raise RuntimeError(
+                    "Cannot serialize tensors on backends with no storage under skip_data context manager"
+                )
+            cpu_tensor = self.cpu()
+            return (
+                torch._utils._rebuild_device_tensor_from_cpu_tensor_with_strides,
+                (cpu_tensor, self.dtype, str(self.device), self.stride(), self.storage_offset(), self.requires_grad),
+            )
+        if self.device.type in ["xla", "maia"] or (
```
malfet, line 345: Why duplicate all that code, rather than add special `if self.devic.dtype == "mtia": return XYZ` check on line 354?

malfet (changes requested):
Why only restore with strides for MTIA, but not for XLA or maia?

torch/_tensor.py
```diff
@@ -329,8 +329,20 @@ def _reduce_ex_internal(self, proto):
         materialize_fake_tensors = (
             torch.serialization._serialization_tls.materialize_fake_tensors
         )
-
-        if self.device.type in ["xla", "maia", "mtia"] or (
+        if self.device.type in ["mtia"] or (
+            not torch._C._has_storage(self)
```
albanD, line 333: Changing mtia is fine as I expect you test it. But everything else should remain unchanged!

albanD (changes requested):

PR #161739, opened 2025-08-28
[dynamo] rename set_fullgraph to error_on_graph_break

williamwen42:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162027
* #162023
* #161747
* __->__ #161739

Renaming `set_fullgraph` to `error_on_graph_break` for now. There are no semantic differences yet. In a followup PR, we will introduce a new `torch.compile` option `error_on_graph_break` that has lower priority than `fullgraph` so that `fullgraph` really returns 1 graph.

I could keep `set_fullgraph` as a deprecated alias for `error_on_graph_break` for now, but I'm hoping that won't be necessary since it's still private API (there are no internal callsites yet, and there are no significant OSS callsites yet).

 cc @albanD @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela @mlazos @guilhermeleobas @xmfan as primary users for `set_fullgraph`

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela

PR #161747, opened 2025-08-28
[dynamo] change error_on_graph_break/fullgraph semantics

williamwen42:
* #162027
* #162023
* __->__ #161747
* #161739

This PR implements the semantics change to `torch._dynamo.error_on_graph_break`:
- ~`torch.compile` now has a new `error_on_graph_break` kwarg that serves as a lower-priority toggle for erroring/continuing on graph breaks~
- `error_on_graph_break` is a new internal `torch.compile `setting that is lower-priority than `fullgraph`. It allows the user to toggle erroring/continuing on graph breaks.
- `error_on_graph_break` does nothing when `fullgraph=True`
- `error_on_graph_break` does NOT guarantee a single graph

Followup [DONE]: need to change the programming model docs to reflect the 3 graph break modes for compilation:
- `fullgraph=True`: enforce one graph, no graph breaks, cannot be toggled
- `fullgraph=False, error_on_graph_break=True`: errors on graph breaks, latter can be toggled during compile time
- `fullgraph=False, error_on_graph_break=False`: resumes tracing on graph breaks, latter can be toggled during compile time

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela @mlazos


test/dynamo/test_decorators.py
```diff
@@ -1953,34 +1953,85 @@ def f3():
...
 
-        @torch.compile(backend="eager", fullgraph=True)
+        @torch.compile(backend="eager", error_on_graph_break=True)
         def f2(x):
             return inner_f2(x)
 
         self.assertEqual(f2(inp), inp + 3)
 
+    def test_error_on_graph_break_fullgraph(self):
+        inp = torch.ones(3)
```
Lucaskabela, line 1999: Might be worth commenting that this aims to test the top row of our programming model (interaction with fullgraph=True and error_on_grpah_break values)

I think making that explicit here would be helpful

torch/__init__.py
```diff
@@ -2499,6 +2502,13 @@ def compile(
         in the function that it will optimize. If True, then we require that the entire function be
         capturable into a single graph. If this is not possible (that is, if there are graph breaks),
         then this will raise an error.
+       error_on_graph_break (bool): If `fullgraph` is set, then this arg does nothing
```
Lucaskabela: once we have some comprehensive documentation on this, might be worth linking to an example that shows when users would want to use this setting

mlazos:
My thinking on this is that if this isn't a super common use case, it shouldn't be a kwarg due to the confusion with fullgraph (which I think comparatively is pretty common). If this is a power-user API I think a decorator makes more sense. If you're okay with this could we hold off on the kwarg until we see how often it would be needed (e.g. seeing people add the decorator to the top-level function)

williamwen42:
@mlazos The idea going forward is that `fullgraph=True` is for power users/frameworks that _really_ want to enforce no graph breaks/one graph guarantee. `error_on_graph_break` is a setting for the average user to disallow most graph breaks, but allow ones that are difficult to deal with.

williamwen42:
@mlazos I removed `error_on_graph_break` as a kwarg to `torch.compile` - we can add it later again if deemed helpful.

test/dynamo_expected_failures/CPython313-test_ordered_dict-CPythonBuiltinDictTests.test_highly_nested
guilhermeleobas, line 1: Based on the source code, this one should fail due to `support.gc_collect()`

```python
def test_highly_nested_subclass(self):
    # Issues 25395 and 35983: test that the trashcan mechanism works
    # correctly for OrderedDict: deleting a highly nested OrderDict
    # should not crash Python.
    OrderedDict = self.OrderedDict
    deleted = []
    with torch._dynamo.set_fullgraph(fullgraph=False):
        class MyOD(OrderedDict):
            def __del__(self):
                deleted.append(self.i)
    obj = None
    for i in range(100):
        obj = MyOD([(None, obj)])
        obj.i = i
    del obj
    support.gc_collect()
    self.assertEqual(deleted, list(reversed(range(100))))
```

test/dynamo/test_modes.py
```diff
@@ -93,7 +93,8 @@ def __torch_dispatch__(self, func, types, args, kwargs=None):
                 return func(*args, **kwargs)
 
         # test e2e, with Inductor, as smoketest.
-        @torch.compile(fullgraph=True, backend="inductor")
```
mlazos, line 96: Why this test? I'm curious haha

mlazos (approved):

test/dynamo/test_modes.py
williamwen42: It was a test that was added after `set_fullgraph` was merged - in that change, `fullgraph=True` went through the same convert_frame.py path as `fullgraph=False`, which increments `counters["frames"]["total"]`. This PR reverted that change so that the counter is no longer incremented.

PR #85447, opened 2022-09-21
[CUBLAS][CUDA GRAPHS] (re-re-open of #83461) Explicitly set the workspace for cuBLAS handles

eqy:
Now includes @dagitses 's optimizations and fixes for teardown

CC @ngimel @ptrblck

PR #161749, opened 2025-08-28
[cuBLAS] update cuBLAS determinism docs, remove workspace requirement checks

eqy:
Since CUDA 11.x (need to update the docs for this, current PR is saying 12.2 which is incorrect) we've been allocating cuBLAS workspaces explicitly per handle/stream combination https://github.com/pytorch/pytorch/pull/85447

According to the cuBLAS documentation, this appears to be sufficient for determinism without any explicit workspace requirements to e.g., `:4096:8` or `:16:8` as was previously expressed in PyTorch docs https://docs.nvidia.com/cuda/cublas/#results-reproducibility

Planning to add an explicit determinism test as well...

cc @ptrblck @msaroufim @jerryzh168 @csarofeen @xwang233 @mruberry @kurtamohler


docs/source/notes/randomness.rst
```diff
@@ -125,10 +125,12 @@ deterministic implementation will be used::
             [[ 0.1509,  1.8027],
              [ 0.0333, -1.1444]]], device='cuda:0')
 
-Furthermore, if you are using CUDA tensors, and your CUDA version is 10.2 or greater, you
+Furthermore, if you are using CUDA tensors, and your CUDA version is between 10.2 and 11.0 you
```
ngimel: no longer supported, just remove this sentence

ngimel (approved):

eqy:
Sure, let's see CI signal after removing the old determinisic alert test

ngimel:
ciflow/H100 is still run on trunk (see on HUD), if it doesn't report existing failures that's a problem (and looks like it doesn't).

PR #161750, opened 2025-08-28
Fix linalg.norm: ensure dtype is respected on CUDA

Navya1707:
Previously, torch.linalg.norm ignored the dtype argument on CUDA, always promoting
to float64. This PR aligns behavior with torch.norm by honoring dtype.
Added tests for float16/float32 on CPU and CUDA.

### DIFF
--- a/aten/src/ATen/native/LinearAlgebra.cpp
+++ b/aten/src/ATen/native/LinearAlgebra.cpp
@@ -3002,7 +3002,15 @@ Tensor& linalg_matrix_norm_out(
 }
 
 // Numerical or None norms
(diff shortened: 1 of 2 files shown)


lezcano (commented):
Linalg norm simply dispatches to linalg.{vector,matrix}_norm and these handle the dtype correctly here

https://github.com/pytorch/pytorch/blob/0371c6aee514c1c5f9fd9169e6ecf07f80f3913c/aten/src/ATen/native/LinearAlgebra.cpp#L241-L244

so I don't understand how this is failing.

PR #161755, opened 2025-08-29
make einsum produce contiguous inputs in more cases

### LINKED ISSUE #161729: Einsum produces transposed output for some cases
```
import torch
import numpy as np

torch.set_default_device("cuda")


torch.set_default_device("cuda")
batch, in_dim, out_dim = 128, 1024, 4096
x = torch.randn(batch, in_dim, dtype=torch.float)
w = torch.randn(out_dim, in_dim, dtype=torch.float)

def linear(x, w):
    return F.linear(x, w)

out = linear(x, w)
print(out.shape, out.stride())
out = torch.einsum("fd,bd->bf", w, x)
print(out.shape, out.stride())
out_np = np.einsum("fd,bd->bf", w.cpu().numpy(), x.cpu().numpy())
print(out_np.shape, out_np.strides)
```
outputs 
```
torch.Size([128, 4096]) (4096, 1)
torch.Size([128, 4096]) (1, 128)
(128, 4096) (16384, 4)
```
`linear` and `numpy` produce contiguous output, `einsum` produces transposed output, there are probably other cases when this happens

cc @jianyuh @nikitaved @mruberry @walterddr @xwang233 @Lezcano

ngimel:
Fixes #161729 
Written by codex
This won't produce contiguous inputs for all einsum applications, because we flatten all right-only and left-only dimensions, so if right and left operand dimensions are interleaved in output, we cannot (with current algo) produce contiguous output, however, for common cases like in the linked issue it works. Let's see what CI says

### DIFF
--- a/aten/src/ATen/native/Linear.cpp
+++ b/aten/src/ATen/native/Linear.cpp
@@ -185,6 +185,17 @@ static Tensor sumproduct_pair(const Tensor& left_, const Tensor& right_, IntArra
   // right:  "lro, summed, ro" permuted with rpermutation and the three flattened
   // then the permuted output is a view of bmm(left, right)
   // finally, opermutation reverts the permutation to the original order of dimensions
+  // By default the output is "lro, lo, 1-for-summed-dims, ro" with original shape dimensions.
+  // However, if all dimensions from the right operand appear before those from the left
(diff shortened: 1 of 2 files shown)


malfet (approved):

albanD (approved):
SGTM !

PR #161756, opened 2025-08-29
use host+device_id to make sure devices are unique in rendezvous request

ngimel:
Per title, for NVL72 systems where devices with the same indices on multiple hosts are within the same nvlink domain


kwen2501 (approved):
LGTM!

PR #128519, opened 2024-06-12
Add new parameter for gen_pyi.py to make it more configureable.

qlzh727:
This will make the script more flexible for the directory where it is executed.

PR #161772, opened 2025-08-29
Add new parameter for gen_pyi.py to make it more configureable.

0xjeffro:
This is a reposting of PR #128519.
This change is important to how we maintain PyTorch at Google.

From the previous PR:
"
This will make the script more flexible for the directory where it is executed.
...
We plan to use the deprecated_yaml from a blaze genrule that invokes pyi.py. As the input to the pyi.py, genrule requires the input file to be explicitly listed out. When we feed the value of tools/autograd/deprecated.yaml to genrule, it failed to resolve since tools/autograd is a package from blaze perspective. Any file under a blaze package will a proper blaze target to be access.
"

cc @haifeng-jin

### DIFF
--- a/tools/autograd/build.bzl
+++ b/tools/autograd/build.bzl
@@ -12,3 +12,9 @@ def define_targets(rules):
             "//torchgen",
         ],
     )
+
+    rules.filegroup(
+        name = "deprecated_yaml",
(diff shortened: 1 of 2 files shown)


albanD (approved):
SGTM!

PR #161603, opened 2025-08-27
[OpenReg] Rename cpu_fallback_blacklist to cpu_fallback_blocklist

fffrog:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #160101
* #160100
* #161773
* #160099
* __->__ #161603

As the title stated.

Related Infos: https://github.com/pytorch/pytorch/pull/158644#discussion_r2301460839

PR #160101, opened 2025-08-07
[OpenReg] Improve the Event and Stream capabilities of DeviceGuardImplInterface

fffrog:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162826
* __->__ #160101
* #161918
* #161917

**Changes:**

- Based on `OpenRegStream` and `OpenRegEvent`, we improve the implementation of Device Guard for `OpenReg`
- Add some related testcases

PR #160100, opened 2025-08-07
[OpenReg] Integrate Event&Stream from OpenReg Backend into PyTorch

fffrog:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #160101
* __->__ #160100
* #161773
* #160099
* #161603

We integrated the openreg backend’s `Stream` and `Event` into PyTorch, all of which are similar
to other accelerators like `CUDA`, `XPUs`, etc.

PR #160099, opened 2025-08-07
[OpenReg] Add Event&Stream Support for OpenReg Backend

fffrog:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #160101
* #160100
* #161773
* __->__ #160099
* #161603

Referring to the signatures and functions of `Stream` and `Event` in CUDA, we use CPU multithreading
and conditional variables to implement equivalent capabilities as the underlying foundation of torch_openreg.

**Changes:**

- Add stream capabilities for OpenReg
- Add event capabilities for OpenReg
- Add kernel launch entrypoint for OpenReg
- Add testcases about stream and event for OpenReg
- Add example for OpenReg

PR #161773, opened 2025-08-29
[OpenReg] Add tests of device and memory for OpenReg

fffrog:
* #160101
* #160100
* __->__ #161773
* #160099
* #161603

As the title stated.


albanD (approved):
Thx.
Do these run in CI? Or only manually?

Also we can focus on python tests for overall behavior as I think that everything should be accessible from python one way or the other.

PR #161770, opened 2025-08-29
[do NOT land] replace _StridedShard w/ sharding order in FSDP2

XilunWu:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #161775
* #161704
* __->__ #161770

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci

PR #161704, opened 2025-08-28
[WIP] add util for computing local shape and global offset from DTensorSpec with extra tensor_sharding field

XilunWu:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #161775
* __->__ #161704
* #161770

**Test**
`pytest test/distributed/checkpoint/fsdp/test_fsdp_dsd.py -s -k fsdp2`

seems problematic:
image

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci

PR #161775, opened 2025-08-29
[WIP] change DCP code to use the new util

XilunWu:
* __->__ #161775
* #161704
* #161770

**Test**
`pytest test/distributed/checkpoint/fsdp/test_fsdp_dsd.py -s -k fsdp2`

Before changing DCP local shape and global offset compute:
image
After:
<insert here>

### DIFF
--- a/torch/distributed/checkpoint/planner_helpers.py
+++ b/torch/distributed/checkpoint/planner_helpers.py
@@ -8,7 +8,9 @@ from torch._utils import _get_device_module
 from torch.distributed._shard.metadata import ShardMetadata
 from torch.distributed._shard.sharded_tensor import ShardedTensor
 from torch.distributed.tensor import DTensor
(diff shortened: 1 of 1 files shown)


XilunWu:
Fail:
test_save_with_fsdp2_tp_and_load_with_tp -> reshard wrong (save=full) -> check it

Pass:
test_save_with_fsdp1_and_load_with_fsdp2
test_save_with_fsdp1_and_load_with_fsdp2_tp -> load right, reshard wrong = wrong, but why it still passes? Hypothesis: load_full_state_dict uses distribute_tensor. (Load wrong reshard wrong) -> check it
test_save_with_tp_and_load_with_fsdp2_tp

Redistribute is necessary for full_state_dict option because it “_gather_state_dict” (called from get_model_state_dict) calls redistribute.

PR #161780, opened 2025-08-29
Fix _scaled_grouped_mm not reported as unsupported on SM100.

alexsamardzic:
* __->__ #161780


aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1665,7 +1665,7 @@ const std::optional<at::Tensor>& bias,
 const std::optional<at::Tensor>& scale_result,
 std::optional<c10::ScalarType> out_dtype,
 bool use_fast_accum) {
-  bool allowed_device = _scaled_mm_allowed_device();
+  bool allowed_device = _scaled_mm_allowed_device(true, false);
```
danielvegamyhre: nit: maybe make these named variables, so the code is readable without looking up the function signature:

```C++
bool sm90_only = true;
bool sm100_only = false;
bool allowed_device = _scaled_mm_allowed_device(sm90_only, sm100_only);
```

danielvegamyhre (approved):

aten/src/ATen/native/cuda/Blas.cpp
alexsamardzic: I've made the change to mimic `_grouped_mm`, where args are designated through comments.  The names probably should be changed, maybe to put "supported" instead of "only".

ngimel (approved):

Skylion007 (approved):

eqy (approved):

PR #161797, opened 2025-08-29
Run vLLM tests on all trunk commits before 2.9 branch cut

huydhn:
This makes it easier to bisect issue now given that we don't have lots of time.

### DIFF
--- a/.github/workflows/vllm.yml
+++ b/.github/workflows/vllm.yml
@@ -2,6 +2,9 @@ name: vllm-test
 
 on:
   push:
+    branches:
+      - main
+      - release/*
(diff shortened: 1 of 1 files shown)


yangw-dev (approved):

huydhn:
Let me do some maths to see how much CI will increase because of this, just to be sure

huydhn:
cc @zou3519 As @yangw-dev has move some multi-gpu test off h100, I can land this now https://github.com/pytorch/pytorch/pull/162292 and we will have vLLM signals for all PyTorch trunk commit for the time being

PR #161805, opened 2025-08-29
[AOTI] Add Windows-compatible implementation of the mmap-related funcs

xuhancn:
Add Windows-compatible implementation of the mmap-related functions.

These code was validated on the small developing project: https://github.com/xuhancn/cross_os_mmap?tab=readme-ov-file#cross_os_mmap

cc @peterjc123 @mszhanyi @skyline75489 @nbcsm @iremyux @Blackhex @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10

### DIFF
--- a/torch/csrc/inductor/aoti_runtime/model_base.h
+++ b/torch/csrc/inductor/aoti_runtime/model_base.h
@@ -1,11 +1,12 @@
 #pragma once
 #ifdef _WIN32
 #include <Windows.h>
+#include <functional> // std::function
+#ifdef USE_MMAP_SELF
 #include <errno.h>
 #include <fcntl.h>
 #include <io.h>
 #include <sys/stat.h>
-#include <functional> // std::function
 
 #define PROT_READ 0x1
 #define PROT_WRITE 0x2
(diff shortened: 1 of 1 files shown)


torch/csrc/inductor/aoti_runtime/model_base.h
```diff
@@ -39,6 +39,266 @@
     }                                 \
   } while (0)
 
+#ifdef _WIN32
```
angelayi, line 2: maybe we can delete the `#ifdef _WIN32` code above?

torch/csrc/inductor/aoti_runtime/model_base.h
```diff
@@ -39,6 +39,266 @@
...
+  }
+
+  return fd;
+}
+
+int close(int fd) {
+  return _close(fd);
+}
+
+void* mmap(
```
angelayi, line 185: should this be under `#ifdef USE_MMAP_SELF`?

torch/csrc/inductor/aoti_runtime/model_base.h
xuhancn: merged Window OS related code together.

torch/csrc/inductor/aoti_runtime/model_base.h
xuhancn: Corrent code is right, `mmap` is only not implemented on `Windows`, it is existing on `Linux` and `MacOS`.

xuhancn:
Hi @angelayi 

It fixed failed mmap UT on Windows.
image

angelayi (approved):

PR #161806, opened 2025-08-29
Investigate moving from constrain_to_fx_strides to constrain_to_fake_tensors

### LINKED ISSUE #153489: at::Tag::needs_fixed_stride_order doesn't work with lists?
This function doesn't seem to handle (tuple, list): https://github.com/pytorch/pytorch/blob/5ff2cb8587005272d32a3c2adfa60b78f19cd31e/torch/_inductor/lowering.py#L2460-L2468

I haven't tested it out yet.

constrain_to_fake_tensors does (https://github.com/pytorch/pytorch/blob/5ff2cb8587005272d32a3c2adfa60b78f19cd31e/torch/_inductor/lowering.py#L2434-L2450) which implies that this logic is indeed missing.

benjaminglass1:
* __->__ #161806

Fixes #153489
Fixes #159155

### DIFF
--- a/torch/_inductor/kernel/mm.py
+++ b/torch/_inductor/kernel/mm.py
@@ -29,7 +29,7 @@ from ..ir import Buffer, FlexibleLayout, is_triton, Layout
 from ..kernel_inputs import MMKernelInputs
 from ..lowering import (
     add_layout_constraint,
-    constrain_to_fx_strides,
+    constrain_to_fake_tensors,
     lowerings as L,
     register_lowering,
 )
@@ -1123,7 +1123,7 @@ def tuned_sparse_semi_structured_mm(
     )
 
 
-add_layout_constraint(aten._scaled_mm.default, constrain_to_fx_strides)
+add_layout_constraint(aten._scaled_mm.default, constrain_to_fake_tensors)
 
 
 @register_lowering(aten._scaled_mm.default, type_promotion_kind=None)  # type: ignore[misc]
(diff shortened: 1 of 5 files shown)


torch/_inductor/lowering.py
```diff
@@ -2745,7 +2747,7 @@ def is_aligned(x):
 make_fallback(aten._grouped_mm, require_dense)
 
 # Need templated kernel. Probably impossible to write efficiently
-make_fallback(aten.convolution_backward, constrain_to_fx_strides)
+make_fallback(aten.convolution_backward, constrain_to_fake_tensors)
```
eellison, line 2750: Pretty sure for backward we want to keep the same dimension dense, but otherwise, we do not need to match exactly (current behavior)

torch/_inductor/kernel/mm.py
```diff
@@ -1123,7 +1123,7 @@ def tuned_sparse_semi_structured_mm(
     )
 
 
-add_layout_constraint(aten._scaled_mm.default, constrain_to_fx_strides)
+add_layout_constraint(aten._scaled_mm.default, constrain_to_fake_tensors)
```
eellison, line 1126: same here i believe

torch/_inductor/lowering.py
benjaminglass1: Gotcha! I was experimenting with what would happen if I just switched all of them, but clearly that didn't work. :-) I'll do a second pass and see if I can more precisely audit which needs to be changed.

PR #161813, opened 2025-08-29
[CI/CD] Windows set git config --global core.ignorecase false

atalman:
Make sure git on windows have core.ignorecase false


malfet (approved):
Probably fine, but can't we move this logic into `checkout-pytorch` action? Or `setup-win` in test-infra?

PR #161040, opened 2025-08-20
[Inductor] Prune configs that require more shared memory than the hardware limit

wychi:
Summary:
This diff removes configs that require more shared memory than the hardware limit, which causes the following compilation error:
```
No valid triton configs. OutOfMemoryError: out of resource: triton_mm Required: 327680 Hardware limit:232448 Reducing block sizes or `num_stages` may help.
```

Test Plan:
```
buck2 test mode/dev-nosan fbcode//caffe2/test/inductor:max_autotune -- test_max_autotune_prune_choices -v 1,stderr
```

Rollback Plan:

Differential Revision: D80594562

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben

PR #161816, opened 2025-08-29
[Reland][Inductor] Prune configs that require more shared memory than the hardware limit

wychi:
* __->__ #161816

Summary:
This is a re-land of [PR161040](https://github.com/pytorch/pytorch/pull/161040), which had previously caused test failures on AMD GPUs. The tests are now configured to target only NVIDIA GPUs.

This diff removes configurations that exceed the hardware shared memory limit, which causes the following compilation error:
```
No valid triton configs. OutOfMemoryError: out of resource: triton_mm Required: 327680 Hardware limit:232448 Reducing block sizes or `num_stages` may help.
```

Test Plan:

```
pytest test/inductor/test_max_autotune.py
pytest test/inductor/test_triton_heuristics.py

```

Reviewers:

Subscribers:

Tasks:

Tags:

Differential Revision: [D81400958](https://our.internmc.facebook.com/intern/diff/D81400958)

### DIFF
--- a/torch/_inductor/template_heuristics/triton.py
+++ b/torch/_inductor/template_heuristics/triton.py
@@ -541,34 +541,68 @@ class BaseConfigHeuristic(metaclass=BaseHeuristicSingleton):
 
         return scaled_configs
 
+    def _get_exceeding_shared_memory_checker(
+        self,
+    ) -> Optional[Callable[[BaseConfig, int], bool]]:
+        """
+        Returns a function that checks whether a given configuration exceeds the available shared memory for the device.
+        If the device does not report available shared memory, returns None.
+        """
+
+        try:
+            device = torch.cuda.current_device()
+            props = torch.cuda.get_device_properties(device)
+            if not hasattr(props, "shared_memory_per_block_optin"):  # for NVidia GPUs
+                return None
+            sm_available = int(props.shared_memory_per_block_optin)
+        except Exception:
+            # If CUDA is not available or properties cannot be queried, return None
+            return None
+
+        def exceeds(gemm_config: BaseConfig, dtype_size: int) -> bool:
+            shared_mem_accum = dtype_size * (
+                gemm_config.block_m * gemm_config.block_k
+                + gemm_config.block_n * gemm_config.block_k
+            )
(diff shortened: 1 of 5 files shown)


torch/_inductor/template_heuristics/triton.py
```diff
@@ -541,34 +541,68 @@ def _scale_mm_configs(
...
 
+    def _get_exceeding_shared_memory_checker(
+        self,
+    ) -> Optional[Callable[[BaseConfig, int], bool]]:
+        """
+        Returns a function that checks whether a given configuration exceeds the available shared memory for the device.
+        If the device does not report available shared memory, returns None.
+        """
+
+        try:
```
coconutruben, line 552: nit: I think you can simplify this by just trusting the try/except

torch/_inductor/template_heuristics/triton.py
```diff
@@ -541,34 +541,68 @@ def _scale_mm_configs(
...
+            device = torch.cuda.current_device()
+            props = torch.cuda.get_device_properties(device)
+            if not hasattr(props, "shared_memory_per_block_optin"):  # for NVidia GPUs
+                return None
+            sm_available = int(props.shared_memory_per_block_optin)
+        except Exception:
+            # If CUDA is not available or properties cannot be queried, return None
+            return None
+
+        def exceeds(gemm_config: BaseConfig, dtype_size: int) -> bool:
```
coconutruben, line 563: nit: can you add a TODO here to make a BaseDeviceConfigHeuristics and have CUDA as an implementation of that? this is fine for the PR now, but I don't like how many if/else we have on cuda vs rocm vs whatever, this should happen at the class level

coconutruben (approved):

coconutruben:
code is fine to me, let's make sure all the devices pass the tests

wychi:
@wychi has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D81400958).

wychi:
close this PR as it seems cause some problems to CI tools. Will restart a new PR from pure OSS workflow.

PR #161819, opened 2025-08-29
 Fix coalesce analysis to work with symints

eellison:
* __->__ #161819

Fix coalesce analysis to work with symints

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @aditew01 @voznesenskym @penguinwu @EikanWang @Guobing-Chen @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @chenyang78


torch/_inductor/tiling_utils.py
```diff
@@ -227,7 +234,7 @@ def get_pw_red_splits(
             (n._body.reduce_vars, n._body.sizes[1]),
         )  # type: ignore[return-value]
 
-    assert sympy_product(n._body.sizes[0]) == pointwise_numel * red_numel  # type: ignore[operator]
+    assert get_hint(sympy_product(n._body.sizes[0])) == get_hint(pointwise_numel * red_numel) # type: ignore[operator]
```
eellison: Sometimes, sympy is unable to prove that these variables are equivalent when they are complex expressions - this is just asserting an invariant we should have elsewhere, based on how we fuse.

torch/_inductor/tiling_utils.py
```diff
@@ -320,6 +327,9 @@ def get_node_splits(self) -> tuple[Split, Split]:
         if len(self.all_node_sizes) == 1:
             return next(iter(self.all_node_sizes))
 
+        if len(self.pw_split_options) == 0:
+            return ((self.pointwise_numel,), (self.red_numel,))
```
eellison, line 367: Similarly, this might have happened if we did a poor job finding splits, because of potential sympy difficulty in inferring divisibility

torch/_inductor/tiling_utils.py
```diff
@@ -531,11 +541,18 @@ def extract_normalized_read_writes(
...
+        try:
+            new_ranges, return_getters_groups = (
+                torch._inductor.codegen.simd.SIMDKernel._split_iteration_ranges(
+                    groups, lengths
+                )
             )
-        )
+        except torch._inductor.codegen.simd.CantSplit:
+            # occasionally with dynamic shapes, we will be unable to prove
+            # divisibility
```
eellison, line 563: Same thing here..

torch/utils/_sympy/functions.py
```diff
@@ -89,7 +89,8 @@
 def _is_symbols_binary_summation(expr: sympy.Expr) -> bool:
     # No need to check that two args are not the same, since expr is pr-optimized but we do it anyway.
     return (
-        expr.is_Add
+        isinstance(expr, sympy.Expr)
+        and expr.is_Add
         and len(expr._args) == 2
```
eellison, line 97: Bug fix - for whatever reason somehow an int was making its way here..

torch/utils/_sympy/functions.py
```diff
@@ -237,6 +238,9 @@ def eval(
             return base
         if base.is_integer and equal_valued(divisor, -1):
             return sympy.Mul(base, -1)
+        if base is divisor:
+            return sympy.S.One
```
eellison: Fix for divisibility

torch/_inductor/tiling_utils.py
eellison: And these invariants come from:

https://github.com/pytorch/pytorch/blob/d8e6b2fddc54c748d976e8f0ebe4b63ebe36d85b/torch/_inductor/codegen/simd.py#L1184

https://github.com/pytorch/pytorch/blob/d8e6b2fddc54c748d976e8f0ebe4b63ebe36d85b/torch/_inductor/codegen/simd.py#L1184

and 

https://github.com/pytorch/pytorch/blob/d8e6b2fddc54c748d976e8f0ebe4b63ebe36d85b/torch/_inductor/codegen/simd.py#L1260

test/inductor/test_loop_ordering.py
```diff
@@ -948,21 +956,23 @@ def fn(nodes):
             if not downcast_transposed_v:
                 self.assertEqual(cont_reads, t_reads * 3)
             else:
-                self.assertEqual(cont_reads, t_reads * 1.5)
```
shunting314, line 951: Just curious, did this cause test failure before or `t_reads` happens to be an even number and the test passes?

torch/_inductor/codegen/simd.py
```diff
@@ -2672,13 +2672,17 @@ def compute_tiling_strategy(
         pw_ranges = [ranges[v] for v in all_iter_vars]
         red_ranges = [ranges[v] for v in all_red_vars]
 
+        # Sometimes dynamic shapes is unable to prove equality without hint
```
shunting314, line 2675: Do you have an example? Using hint may make the torch._check pass just for some hint while it may fail for other hints.

torch/_inductor/codegen/simd.py
```diff
@@ -3113,4 +3117,7 @@ def is_good_size(s):
 
 
 class CantSplit(Exception):
-    pass
+    def __init__(self, expr, remaining):
```
shunting314, line 3120: nit: an `__str__` saying 'expr' can not be evenly divided by `remaining` would be nice.

torch/_inductor/tiling_utils.py
```diff
@@ -67,7 +67,7 @@ def solve_for_tiling(expr: sympy.Expr) -> Optional[sympy.Expr]:
     To simplify things for sympy, we'll try just x * y == 1, check x(1) and x(0).
     """
 
-    if len(expr.free_symbols) == 0:
+    if len(expr.free_symbols) != 1:
```
shunting314, line 70: too restrictive?

torch/_inductor/tiling_utils.py
```diff
@@ -198,8 +198,12 @@ def find_coalesced_var(
             variables[v] = 0
         else:
             variables[v] = get_hint(v)
+    try:
+        zero_index = sympy_subs(index, variables)
+    except:
+        print(index, variables)
+        raise
```
shunting314: nit: catch the specific type of exceptions

test/inductor/test_loop_ordering.py
eellison: not actually sue, let me look at this

torch/_inductor/codegen/simd.py
eellison: this is post inferring splits, so, we should already know that these are equal. its just enforcing invariants / debugging. e.g. the results of the hint are not actually used.

torch/_inductor/tiling_utils.py
eellison: yea, the print is errant, let me update

torch/_inductor/codegen/simd.py
laithsakka: if we know for sure that they are always the same, why do you need  to compare hints?
seems like you want to do here
V.graph.sizevars.check_equals( sympy_product(pw_ranges) ,  pointwise_numel)?

since we operate on sympy expressions , this shall add runtime assertions for unbacked and guard for backed(but its ok to guard since you said we already know its true).

eellison: It's just to document invariants.  again, we don't actually need to guard etc, because the result is unused.

test/inductor/test_loop_ordering.py
eellison: updated, this was an unnecessary change.

torch/_inductor/tiling_utils.py
eellison: in this case, if there are multiple free symbols, then inducing a split won't cause coalescing.

torch/utils/_sympy/functions.py
```diff
@@ -242,6 +243,9 @@ def eval(cls, base: sympy.Integer, divisor: sympy.Integer) -> sympy.Basic | None
             return base
         if base.is_integer and equal_valued(divisor, -1):
             return sympy.Mul(base, -1)
+        if base is divisor:
```
shunting314: Should `base == divisor` being more general since that support the case that `base` is an expression rather than a bare symbol

shunting314 (approved):

torch/utils/_sympy/functions.py
eellison: Seemingly, this is already handled.. but i agree that == is more general
>>> import sympy
>>> x = sympy.Symbol("x")
>>> (x * 4) is (x * 4)
True

shunting314: Nice, maybe sympy has cache that helps for basic cases. But check a slightly more complex cases:
```
>>> a = x + 1
>>> b = x + 2
>>> c = b - 1
>>> a
x + 1
>>> a is c
False
>>> c
x + 1
>>> a == c
True
```

eellison: yea, sg, updated

Issue #161822, opened 2025-08-29
[CUDA][cuBLAS] #125888 introduces measurable CPU overhead for matmuls


eqy:
### 🐛 Describe the bug

Short repro:
```python
import torch
import time

warmup = 128
iters = 16384

a = torch.zeros(512, 512, device='cuda', dtype=torch.bfloat16)
for _ in range(warmup):
    torch.matmul(a, a)

torch.cuda.synchronize()
t0 = time.perf_counter()
for _ in range(iters):
    torch.matmul(a, a)
torch.cuda.synchronize()
t1 = time.perf_counter()
print(f"{1e6 * (t1 - t0)/iters}")
```

microbenchmark runs @  `53e0b9c3936176521ed8d71c00abd5b7499057c2`
```
8.559683067232982
8.514835997175396
8.353039788744354
8.471890282635286
8.251526651292806
```

microbenchmark runs @  `de45c5f673ce261e9a82c54280beeda36cff640e` (previous commit)
```
8.313276055105234
8.274250603790279
8.362368646430696
8.318946299823438
8.450675863969082 
```

Especially painful as currently _all_ matmuls that grab a cuBLAS handle call the `at::globalContext().float32Precision()` function which appears to do a minimum of `.find()` calls, string comparisons, and validation. Will follow up with a PR to skip this for non-fp32 matmuls but longer term we would like to reduce the overhead here.

### Versions

current `main`

cc @msaroufim @jerryzh168 @ptrblck @albanD

eqy:
> [@eqy](https://github.com/eqy) I was trying to get an estimate of how much perf improvement we can get this was and the numbers are not very encouraging. I built PyTorch locally (RelWithDebInfo, x86_64) and tested three configs:
> 
> 1. Commit before the offending commit - **9.35 ticks**
> 2. With the offending commit - **9.96 ticks**
> 3. All calls to `validate_fp32_prec` and `check_fp32_prec_backend_and_op` removed - **9.8 ticks**
> 
> I also found that the measurements are very noisy and change a lot across runs. I increased the number of iterations but there is still significant variation (±0.4 tick). I'll keep looking and report if I find anything.

Thanks! Unfortunately yes I also observed that the measurements were quite noisy but the slowdown seems real.
[...]

PR #125888, opened 2024-05-10
refine fp32 precision api

zhuhaozhe:
Based on the [conversation](https://github.com/pytorch/pytorch/issues/121791), we plan to drop the "highest, high, medium" to represent fp32  internal computation data types . Instead, we will directly use the algorithm to represent it.

#### Pros
 - The names are more informative. 'tf32' is more informative than a simple "high".
 - Easier to extend new algorithm like `tf32x3` 
#### Cons
 - "HIGHEST, HIGH, MEDIUM" indicated the relative precision between different algorithms. However, we can have more documents to discuss them.

### We provide a layered structure for backends/operators.
('f32' is short for 'fp32_precision')
image

### We provide 3 fp32 compute precision can be set:
 - **"ieee"**: Not allowed to use any other internal computation data types .
 - **"tf32"**: Allowed to use tf32 as internal computation data types.
 - **"bf16"**: Allowed to use bf16 as internal computation data types.
 - **"none"**:  Precision's are not set. Can be override by its father node.

### Overriding Precision Settings
Child node can be override by its father node if it is set to default.
For current default settings:
```
backend = generic, op = all, precision setting = none
    backend = cuda, op = all, precision setting = none
        backend = cuda, op = conv, precision setting = tf32
        backend = cuda, op = rnn, precision setting = tf32
        backend = cuda, op = matmul, precision setting = none
    backend = matmul, op = all, precision setting = none
        backend = matmul, op = conv, precision setting = none
        backend = matmul, op = rnn, precision setting = none
        backend = matmul, op = matmul, precision setting = none
```
 - If the user set `torch.backends.mkldnn.fp32_precision="bf16"`, his child nodes `torch.backends.mkldnn.matmul.fp32_precision` / `torch.backends.mkldnn.conv.fp32_precision` / `torch.backends.mkldnn.rnn.fp32_precision` will also be override to "bf16".
 - If the user set `torch.backends.fp32_precision="bf16"`,  `torch.backends.mkldnn.fp32_precision` and his child nodes will also we override to "bf16".

### Backward Compatible
Since new API allow user to have more fine-grained control. There will be some conflict. For example, previous `torch.backends.cudnn.allow_tf32` are not enough to represent the status for `torch.backends.cudnn.rnn.fp32_precision="ieee"` and `torch.backends.cudnn.conv.fp32_precision="tf32"`. Therefore, our goal for backward compatible is
 - If the user only uses previous APIs, it will work as previous expectations.
[...]

PR #161823, opened 2025-08-29
[CUDA][cuBLAS][TF32] Skip checking TF32 options and setting cuBLAS handle TF32 modes for other dtypes

eqy:
For #161822

Basically #125888 appears to introduce measurable CPU overhead due to TF32 precision setting checks. Unfortunately one of the checks is in `getCurrentCUDABlasHandle`, which is called preceding _every_ cuBLAS matmul (and in a few other places such as workspace setup). We don't need to do this for non-`float32` matmuls so this PR is to alleviate the performance hit where it hurts the most (smaller dtypes that have faster matmuls) at the cost of some copypasta.

Some microbenchmark runs with this PR (microseconds):
```
7.812199214640714 
8.07804053692962 
7.865882366786536
7.898942214978888
8.018492849259928
```
and without
```
8.222943563396257
8.129948014357069
8.184711361991504
8.28010104214627
8.266569921033806
```

script:
```
import torch
import time

warmup = 128
iters = 16384

a = torch.zeros(512, 512, device='cuda', dtype=torch.bfloat16)
for _ in range(warmup):
    torch.matmul(a, a)

torch.cuda.synchronize()
t0 = time.perf_counter()
for _ in range(iters):
    torch.matmul(a, a)
torch.cuda.synchronize()
t1 = time.perf_counter()
print(f"{1e6 * (t1 - t0)/iters}")
```

Longer term we'd prefer making `float32Precision` faster (better data structures, less validation, etc.?)

cc @ptrblck @msaroufim @jerryzh168 @csarofeen @xwang233 @zasdfgbnm

### DIFF
--- a/aten/src/ATen/cuda/CUDABlas.cpp
+++ b/aten/src/ATen/cuda/CUDABlas.cpp
@@ -586,6 +586,12 @@ void bgemm_internal_cublas<float>(CUDABLAS_BGEMM_ARGTYPES(float)) {
   // See Note [Writing Nondeterministic Operations]
   globalContext().alertCuBLASConfigNotDeterministic();
   cublasHandle_t handle = at::cuda::getCurrentCUDABlasHandle();
+  if (!NoTF32Guard::should_disable_tf32() &&
+      at::globalContext().float32Precision("cuda", "matmul") == "tf32") {
+    TORCH_CUDABLAS_CHECK(cublasSetMathMode(handle, CUBLAS_TF32_TENSOR_OP_MATH));
+  } else {
+    TORCH_CUDABLAS_CHECK(cublasSetMathMode(handle, CUBLAS_DEFAULT_MATH));
+  }
   cublasOperation_t opa = _cublasOpFromChar(transa);
   cublasOperation_t opb = _cublasOpFromChar(transb);
   _cublasAdjustLdLevel3(transa, transb, m, n, k, &lda, &ldb, &ldc);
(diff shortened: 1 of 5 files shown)


aten/src/ATen/cuda/CUDABlas.cpp
```diff
@@ -586,6 +586,12 @@ void bgemm_internal_cublas<float>(CUDABLAS_BGEMM_ARGTYPES(float)) {
   // See Note [Writing Nondeterministic Operations]
   globalContext().alertCuBLASConfigNotDeterministic();
   cublasHandle_t handle = at::cuda::getCurrentCUDABlasHandle();
+  if (!NoTF32Guard::should_disable_tf32() &&
```
Skylion007: Why not just delegate this to an inline function instead of copy paste

malfet (changes requested):
This feels mildly wrong... I.e. replacing one check with 30+ in different places are bound to cause errors. Where overhead is coming from? String comparison? Than it should be replaced by enum

eqy:
> This feels mildly wrong... I.e. replacing one check with 30+ in different places are bound to cause errors. Where overhead is coming from? String comparison? Than it should be replaced by enum

I brought this up in the original PR, there's multiple reasons string comparison is used rather than `enum`, but it's not that simple either as there are multiple levels of string comparison e.g., for backend and for precision https://github.com/pytorch/pytorch/pull/125888

Additionally there's a lot of validation steps rather than simply returning values than before. I think with the changes to inline it would be cleaner than before.

ngimel:
Should we revert #125888 until we come up with better design? It seems like it created a lot of problems without solving any

PR #161825, opened 2025-08-29
C++ API handle optimizer defaults 

### LINKED ISSUE #141884: Default optimizer options are ignored in the C++ API
### 🐛 Describe the bug

This standalone C++ program creates an optimizer with 3 param groups of a single tensor each. The tensor are initialized to 1. It also has default options that set the learning rate to 0, and the param groups are not setting their learning rate. It then computes a backward pass and runs an optimization step, printing the tensors modified by the optimizer:
```
#include <torch/torch.h>
#include <iostream>

using namespace torch;

int main() {
  Tensor input = torch::ones({1, 3});
  Tensor w = torch::ones({2, 3});
  Tensor b = torch::ones({1, 2});
  Tensor output;

  input.set_requires_grad(true);
  w.set_requires_grad(true);
  b.set_requires_grad(true);   

  std::vector<optim::OptimizerParamGroup> param_groups;

  // Param group for input tensor
  auto *input_adam_opts = new optim::AdamOptions();
  input_adam_opts->weight_decay(0.11);
  optim::OptimizerOptions *input_opts = input_adam_opts;
  param_groups.push_back(
    optim::OptimizerParamGroup(
      std::vector<Tensor>({input}),
      std::unique_ptr<optim::OptimizerOptions>(input_opts)
    )
  );

  // Param group for w tensor
  auto *w_adam_opts = new optim::AdamOptions();
  w_adam_opts->weight_decay(0.22);
  optim::OptimizerOptions *w_opts = w_adam_opts;
  param_groups.push_back(
    optim::OptimizerParamGroup(
      std::vector<Tensor>({w}),
      std::unique_ptr<optim::OptimizerOptions>(w_opts)
    )
  );

  // Param group for b tensor
  auto *b_adam_opts = new optim::AdamOptions();
  b_adam_opts->weight_decay(0.33);
  optim::OptimizerOptions *b_opts = b_adam_opts;
  param_groups.push_back(
    optim::OptimizerParamGroup(
      std::vector<Tensor>({b}),
      std::unique_ptr<optim::OptimizerOptions>(b_opts)
    )
  );

  // Default options and optimizer creation
  optim::AdamOptions default_opts;
  default_opts.lr(0);
  auto optimizer = optim::Adam(param_groups, default_opts);

  output = at::linear(input, w, b);
  Tensor outgrad = torch::ones({1, 2});
  output.backward(outgrad);

  optimizer.step();
  std::cout << "input:" << std::endl;
  std::cout << input << "\n";
  std::cout << "w:" << std::endl;
  std::cout << w << "\n";
  std::cout << "b:" << std::endl;
  std::cout << b << "\n";

  std::cout << "\n";

  std::vector<torch::optim::OptimizerParamGroup> retrieved_param_groups = 
    optimizer.param_groups();

  auto *group_0_opts = static_cast<torch::optim::AdamOptions*>(
    &(retrieved_param_groups[0].options())
  );
  std::cout << "input weight decay:\n";
  std::cout << group_0_opts->weight_decay() <<"\n";
  std::cout << "input learning rate:\n";
  std::cout << group_0_opts->lr() <<"\n";

  std::cout << "\n";

  auto *group_1_opts = static_cast<torch::optim::AdamOptions*>(
    &(retrieved_param_groups[1].options())
  );
  std::cout << "w weight decay:\n";
  std::cout << group_1_opts->weight_decay() <<"\n";
  std::cout << "w learning rate:\n";
  std::cout << group_1_opts->lr() <<"\n";

  std::cout << "\n";

  auto *group_2_opts = static_cast<torch::optim::AdamOptions*>(
    &(retrieved_param_groups[2].options())
  );
  std::cout << "b weight decay:\n";
  std::cout << group_2_opts->weight_decay() <<"\n";
  std::cout << "b learning rate:\n";
  std::cout << group_2_opts->lr() <<"\n";
}
```

The output is:
```
input:
 0.9990  0.9990  0.9990
[ CPUFloatType{1,3} ]
w:
 0.9990  0.9990  0.9990
 0.9990  0.9990  0.9990
[ CPUFloatType{2,3} ]
b:
 0.9990  0.9990
[ CPUFloatType{1,2} ]

input weight decay:
0.11
input learning rate:
0.001

w weight decay:
0.22
w learning rate:
0.001

b weight decay:
0.33
b learning rate:
0.001
```

So it can be seen that setting the learning rate to 0 in the default options is ignored, the tensor have changed after the step. And retrieving the param group options shows that they are using a learning rate of 0.001.

### Versions

I have built from source using commit `23c0d2689e4b49858787adc6ca2c5032ac221b96`. This is the `cmake` command:

```
cmake \
	-DCMAKE_BUILD_TYPE=Release \
	-DBUILD_PYTHON=False \
	-DCMAKE_INSTALL_PREFIX=$home_dir/torch \
	-DTORCH_BUILD_VERSION=2.6.0a0+git23c0d26 \
	-DUSE_NUMPY=False \
```

This is the output of the `collect_env.py` script:

```
PyTorch version: N/A
Is debug build: N/A
CUDA used to build PyTorch: N/A

OS: Ubuntu 24.04.1 LTS (x86_64)
GCC version: (Ubuntu 13.2.0-23ubuntu4) 13.2.0
CMake version: version 3.31.1
Libc version: glibc-2.39

Python version: 3.12.3 (main, Nov  6 2024, 18:32:19) [GCC 13.2.0] (64-bit runtime)
Python platform: Linux-6.8.0-49-generic-x86_64-with-glibc2.39
Is CUDA available: N/A
CUDA runtime version: Could not collect
GPU models and configuration: GPU 0: NVIDIA GeForce GTX 1650
Nvidia driver version: 550.120
cuDNN version: Could not collect
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: N/A

CPU:
Address sizes:                        39 bits physical, 48 bits virtual
CPU(s):                               16
On-line CPU(s) list:                  0-15
Model name:                           Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
Model:                                158
Core(s) per socket:                   8
Stepping:                             13
CPU(s) scaling MHz:                   85%
CPU max MHz:                          5000.0000
CPU min MHz:                          800.0000
BogoMIPS:                             4800.00
Flags:                                fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc art arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni pclmulqdq dtes64 monitor ds_cpl vmx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm 3dnowprefetch cpuid_fault epb ssbd ibrs ibpb stibp ibrs_enhanced tpr_shadow flexpriority ept vpid ept_ad fsgsbase tsc_adjust sgx bmi1 avx2 smep bmi2 erms invpcid mpx rdseed adx smap clflushopt intel_pt xsaveopt xsavec xgetbv1 xsaves dtherm ida arat pln pts hwp hwp_notify hwp_act_window hwp_epp vnmi sgx_lc md_clear flush_l1d arch_capabilities
Virtualization:                       VT-x
L1d cache:                            256 KiB (8 instances)
L1i cache:                            256 KiB (8 instances)
L2 cache:                             2 MiB (8 instances)
L3 cache:                             16 MiB (1 instance)
NUMA node0 CPU(s):                    0-15
Vulnerability Gather data sampling:   Mitigation; Microcode
Vulnerability Itlb multihit:          KVM: Mitigation: VMX disabled
Vulnerability L1tf:                   Not affected
Vulnerability Mds:                    Not affected
Vulnerability Meltdown:               Not affected
Vulnerability Mmio stale data:        Mitigation; Clear CPU buffers; SMT vulnerable
Vulnerability Reg file data sampling: Not affected
Vulnerability Retbleed:               Mitigation; Enhanced IBRS
Vulnerability Spec store bypass:      Mitigation; Speculative Store Bypass disabled via prctl
Vulnerability Spectre v1:             Mitigation; usercopy/swapgs barriers and __user pointer sanitization
Vulnerability Spectre v2:             Mitigation; Enhanced / Automatic IBRS; IBPB conditional; RSB filling; PBRSB-eIBRS SW sequence; BHI SW loop, KVM SW loop
Vulnerability Srbds:                  Mitigation; Microcode
Vulnerability Tsx async abort:        Mitigation; TSX disabled

[pip3] numpy==1.26.3
[pip3] nvidia-cublas-cu11==11.11.3.6
[pip3] nvidia-cuda-cupti-cu11==11.8.87
[pip3] nvidia-cuda-nvrtc-cu11==11.8.89
[pip3] nvidia-cuda-runtime-cu11==11.8.89
[pip3] nvidia-cudnn-cu11==9.1.0.70
[pip3] nvidia-cufft-cu11==10.9.0.58
[pip3] nvidia-curand-cu11==10.3.0.86
[pip3] nvidia-cusolver-cu11==11.4.1.48
[pip3] nvidia-cusparse-cu11==11.7.5.86
[pip3] nvidia-nccl-cu11==2.20.5
[pip3] nvidia-nvtx-cu11==11.8.86
[pip3] torch==2.4.1+cu118
[pip3] torchaudio==2.4.1+cu118
[pip3] torchvision==0.19.1+cu118
[pip3] triton==3.0.0
[conda] Could not collect
```

cc @vincentqb @jbschlosser @albanD @janeyx99 @crcrpar

stmcgovern:
Fixes #141884 

This fixes the issue for all optimizers and parameter options. 
A member function `overwrite_from` is added to the optimizer base class. Each optimizer then implements this function for comparing their accepted parameters to defaults. A SFINAE approach to handle the different optimizer parameters generically (in optimizer.h only) was evaluated, but I think this is easier to review and maintain.

This mirrors the Python API up to one edge case. An example of the edge case is provided below. 

Python can distinguish between 1) Key not present in dict = "not specified"  and 2) Key present in dict = "explicitly set". The C++ implementation cannot.
The issue hinges on whether or not to track if a particular parameter was set by the user explicitly or not (discrepancy in the case when the constructor default is explicitly passed in). 

To track this seems like it will take more intervention than would be worth it (modify TORCH_ARG to keep track, use std::optional for the parameter types, use bitset tracking) and was not pursued in the current PR. I'm happy to alter the design if appropriate. 

### Example of edge case hinging on CONSTRUCTOR DEFAULTS vs OPTIMIZER DEFAULTS 

1. CONSTRUCTOR DEFAULTS:
   These are the values you get when calling AdamOptions()
   AdamOptions().lr() = 0.001
   AdamOptions().weight_decay() = 0
   AdamOptions().eps() = 1e-08

2. OPTIMIZER DEFAULTS:
   These are the values the user chose when creating the optimizer
   User's optimizer defaults:
   optimizer.lr() = 0.005
   optimizer.weight_decay() = 0.1
   optimizer.eps() = 1e-07

3. THE PROBLEM SCENARIO:
   User wants to add a parameter group with explicit weight_decay=0.0
   User sets: weight_decay(0)

4. THE CONFUSION:
   Constructor default weight_decay: 0
   User's explicit weight_decay:     0
   Are they equal? YES

   Since they're equal, our overwrite_from() logic thinks:
   "User didn't set weight_decay explicitly, use optimizer default"

5. CURRENT BEHAVIOR:
   Final weight_decay: 0.1
   User expected:      0
   Match? ❌ NO

=== KEY INSIGHT ===
Constructor defaults are built into the C++ class definition.
Optimizer defaults are chosen by the user at runtime. We want to respect the user intention.


stmcgovern:
Hi @janeyx99, this is the PR I mentioned in issue #141884 . How can I link the PR to the issue? I thought the "Fixes #number" does that...

test/cpp/api/optim.cpp
```diff
@@ -564,3 +566,332 @@ TEST(OptimTest, CheckLRChange_ReduceLROnPlateau_Adam) {
...
+
+  Adam optimizer(param_groups, defaults);
+
+  // Check Group 1: weight_decay preserved, others inherited
+  auto& group1_opts =
+      static_cast<AdamOptions&>(optimizer.param_groups()[0].options());
+  ASSERT_NEAR(group1_opts.lr(), 0.002, 1e-6); // Inherited
+  ASSERT_EQ(group1_opts.betas(), std::make_tuple(0.8, 0.88)); // Inherited
+  ASSERT_NEAR(group1_opts.eps(), 1e-12, 1e-15); // Inherited
+  ASSERT_NEAR(group1_opts.weight_decay(), 0.11, 1e-6); // Preserved
```
janeyx99: How come these can't be ASSERT_EQ?

test/cpp/api/optim.cpp
```diff
@@ -564,3 +566,332 @@ TEST(OptimTest, CheckLRChange_ReduceLROnPlateau_Adam) {
...
+      static_cast<SGDOptions&>(optimizer.param_groups()[1].options());
+  ASSERT_NEAR(group2_opts.lr(), 0.02, 1e-6); // Preserved
+  ASSERT_NEAR(group2_opts.momentum(), 0.9, 1e-6); // Inherited
+  ASSERT_NEAR(group2_opts.dampening(), 0.0, 1e-6); // Inherited
+  ASSERT_NEAR(group2_opts.weight_decay(), 0.05, 1e-6); // Inherited
+  ASSERT_TRUE(group2_opts.nesterov()); // Inherited
+}
+
+TEST(OptimTest, MergeWithDefaultOptions_AdamW) {
+  torch::manual_seed(0);
```
janeyx99: is this important to the test? the actual params won't matter, right?

torch/csrc/api/src/optim/optimizer.cpp
```diff
@@ -94,17 +94,32 @@ void OptimizerOptions::serialize(
       "You must override it in your subclass of torch::optim::OptimizerCloneableOptions<YourOptimizerOptions>.");
 }
 
+void OptimizerOptions::overwrite_from(const OptimizerOptions& source) {
```
janeyx99: Hi! Some highlevel qs:
How come we need a whole new overwrite_from API?
From the toplevel I would expect us to fix the base class so that the user specified defaults override the original defaults and then are used in add_param_group, without the need for adding a new API.

janeyx99 (commented):
Thanks for working on the fix, this is an interesting bug indeed. I left a comment on the overall approach. Furthermore, since these tests don't run on CI--could you post a paste of the C++ test results locally?

test/cpp/api/optim.cpp
stmcgovern: changed. use in serialization tests left.

test/cpp/api/optim.cpp
stmcgovern: right. removed

torch/csrc/api/src/optim/optimizer.cpp
stmcgovern: maybe we don't :) . I've tried to provide a fix to the base class, without adding a new API, but don't see a way to do it without one new virtual function call.

stmcgovern:
Hi @janeyx99, thanks very much for your feedback. I've taken each one of your comments into consideration.
I've tried to find a complete fix to the base class, without adding a new API. The core problem is that we need a way to track whether the parameter field was explicitly set or not (to mimic Python dict behavior). I started with a function pointer registry and string based checking in add_param_group. Trying to move as much to compile time as possible, 1) led to the CRTP pattern for static dispatch and 2) avoiding string hash checking, with the bitset tracking, was adopted for performance and ease of serialization. 
This uses a new macro (which I also don't like), but seems to solve the problem in an efficient and complete way (python api parity), without introducing a new API.

Here are the local optimizer tests (including the new ones). Is there a better way with gtest to see what the test actually ran (not just pass/fail)? dropping the filter and running 1020 tests pass too.

```
(base) [root@49023e5b8e19 pytorch]# ./build/bin/test_api --gtest_filter="*Optim*" -v
CUDA not available. Disabling CUDA and MultiCUDA tests
Note: Google Test filter = *Optim*-*_CUDA:*_MultiCUDA
[==========] Running 51 tests from 2 test suites.
[----------] Global test environment set-up.
[----------] 44 tests from OptimTest
[ RUN      ] OptimTest.OptimizerAccessors
[       OK ] OptimTest.OptimizerAccessors (1 ms)
[ RUN      ] OptimTest.OldInterface
[       OK ] OptimTest.OldInterface (0 ms)
[ RUN      ] OptimTest.XORConvergence_SGD
[       OK ] OptimTest.XORConvergence_SGD (655 ms)
[ RUN      ] OptimTest.XORConvergence_LBFGS
[       OK ] OptimTest.XORConvergence_LBFGS (434 ms)
[ RUN      ] OptimTest.XORConvergence_Adagrad
[       OK ] OptimTest.XORConvergence_Adagrad (250 ms)
[ RUN      ] OptimTest.XORConvergence_RMSprop
[       OK ] OptimTest.XORConvergence_RMSprop (236 ms)
[ RUN      ] OptimTest.XORConvergence_RMSpropWithMomentum
[       OK ] OptimTest.XORConvergence_RMSpropWithMomentum (703 ms)
[ RUN      ] OptimTest.XORConvergence_Adam
[       OK ] OptimTest.XORConvergence_Adam (264 ms)
[ RUN      ] OptimTest.XORConvergence_AdamWithAmsgrad
[       OK ] OptimTest.XORConvergence_AdamWithAmsgrad (278 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_Adam
[       OK ] OptimTest.ProducesPyTorchValues_Adam (92 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_AdamWithWeightDecay
[       OK ] OptimTest.ProducesPyTorchValues_AdamWithWeightDecay (96 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_AdamWithWeightDecayAndAMSGrad
[       OK ] OptimTest.ProducesPyTorchValues_AdamWithWeightDecayAndAMSGrad (99 ms)
[ RUN      ] OptimTest.XORConvergence_AdamW
[       OK ] OptimTest.XORConvergence_AdamW (267 ms)
[ RUN      ] OptimTest.XORConvergence_AdamWWithAmsgrad
[       OK ] OptimTest.XORConvergence_AdamWWithAmsgrad (266 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_AdamW
[       OK ] OptimTest.ProducesPyTorchValues_AdamW (94 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_AdamWWithoutWeightDecay
[       OK ] OptimTest.ProducesPyTorchValues_AdamWWithoutWeightDecay (95 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_AdamWWithAMSGrad
[       OK ] OptimTest.ProducesPyTorchValues_AdamWWithAMSGrad (99 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_Adagrad
[       OK ] OptimTest.ProducesPyTorchValues_Adagrad (78 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_AdagradWithWeightDecay
[       OK ] OptimTest.ProducesPyTorchValues_AdagradWithWeightDecay (83 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_AdagradWithWeightDecayAndLRDecay
[       OK ] OptimTest.ProducesPyTorchValues_AdagradWithWeightDecayAndLRDecay (83 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_RMSprop
[       OK ] OptimTest.ProducesPyTorchValues_RMSprop (82 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_RMSpropWithWeightDecay
[       OK ] OptimTest.ProducesPyTorchValues_RMSpropWithWeightDecay (86 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_RMSpropWithWeightDecayAndCentered
[       OK ] OptimTest.ProducesPyTorchValues_RMSpropWithWeightDecayAndCentered (94 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_RMSpropWithWeightDecayAndCenteredAndMomentum
[       OK ] OptimTest.ProducesPyTorchValues_RMSpropWithWeightDecayAndCenteredAndMomentum (99 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_SGD
[       OK ] OptimTest.ProducesPyTorchValues_SGD (69 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_SGDWithWeightDecay
[       OK ] OptimTest.ProducesPyTorchValues_SGDWithWeightDecay (71 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_SGDWithWeightDecayAndMomentum
[       OK ] OptimTest.ProducesPyTorchValues_SGDWithWeightDecayAndMomentum (72 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_SGDWithWeightDecayAndNesterovMomentum
[       OK ] OptimTest.ProducesPyTorchValues_SGDWithWeightDecayAndNesterovMomentum (80 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_LBFGS
[       OK ] OptimTest.ProducesPyTorchValues_LBFGS (68 ms)
[ RUN      ] OptimTest.ProducesPyTorchValues_LBFGS_with_line_search
[       OK ] OptimTest.ProducesPyTorchValues_LBFGS_with_line_search (298 ms)
[ RUN      ] OptimTest.ZeroGrad
[       OK ] OptimTest.ZeroGrad (0 ms)
[ RUN      ] OptimTest.ExternalVectorOfParameters
[       OK ] OptimTest.ExternalVectorOfParameters (0 ms)
[ RUN      ] OptimTest.AddParameter_LBFGS
[       OK ] OptimTest.AddParameter_LBFGS (0 ms)
[ RUN      ] OptimTest.CheckLRChange_StepLR_Adam
[       OK ] OptimTest.CheckLRChange_StepLR_Adam (0 ms)
[ RUN      ] OptimTest.CheckLRChange_ReduceLROnPlateau_Adam
[       OK ] OptimTest.CheckLRChange_ReduceLROnPlateau_Adam (0 ms)
[ RUN      ] OptimTest.MergeWithDefaultOptions_Adam
[       OK ] OptimTest.MergeWithDefaultOptions_Adam (0 ms)
[ RUN      ] OptimTest.MergeWithDefaultOptions_SGD
[       OK ] OptimTest.MergeWithDefaultOptions_SGD (0 ms)
[ RUN      ] OptimTest.MergeWithDefaultOptions_AdamW
[       OK ] OptimTest.MergeWithDefaultOptions_AdamW (0 ms)
[ RUN      ] OptimTest.MergeWithDefaultOptions_Adagrad
[       OK ] OptimTest.MergeWithDefaultOptions_Adagrad (0 ms)
[ RUN      ] OptimTest.MergeWithDefaultOptions_RMSprop
[       OK ] OptimTest.MergeWithDefaultOptions_RMSprop (0 ms)
[ RUN      ] OptimTest.MergeWithDefaultOptions_LBFGS
[       OK ] OptimTest.MergeWithDefaultOptions_LBFGS (0 ms)
[ RUN      ] OptimTest.MergeWithDefaultOptions_NoOptionsInheritance
[       OK ] OptimTest.MergeWithDefaultOptions_NoOptionsInheritance (0 ms)
[ RUN      ] OptimTest.SerializationPreservesFieldTracking_Adam
[       OK ] OptimTest.SerializationPreservesFieldTracking_Adam (9 ms)
[ RUN      ] OptimTest.SerializationPreservesFieldTracking_SGD
[       OK ] OptimTest.SerializationPreservesFieldTracking_SGD (0 ms)
[----------] 44 tests from OptimTest (5217 ms total)

[----------] 7 tests from SerializeTest
[ RUN      ] SerializeTest.Optim
[       OK ] SerializeTest.Optim (1 ms)
[ RUN      ] SerializeTest.Optim_Adagrad
[       OK ] SerializeTest.Optim_Adagrad (1 ms)
[ RUN      ] SerializeTest.Optim_SGD
[       OK ] SerializeTest.Optim_SGD (1 ms)
[ RUN      ] SerializeTest.Optim_Adam
[       OK ] SerializeTest.Optim_Adam (1 ms)
[ RUN      ] SerializeTest.Optim_AdamW
[       OK ] SerializeTest.Optim_AdamW (1 ms)
[ RUN      ] SerializeTest.Optim_RMSprop
[       OK ] SerializeTest.Optim_RMSprop (1 ms)
[ RUN      ] SerializeTest.Optim_LBFGS
[       OK ] SerializeTest.Optim_LBFGS (1 ms)
[----------] 7 tests from SerializeTest (12 ms total)

[----------] Global test environment tear-down
[==========] 51 tests from 2 test suites ran. (5230 ms total)
[  PASSED  ] 51 tests.
(base) [root@49023e5b8e19 pytorch]# 
```

torch/csrc/api/include/torch/optim/optimizer.h
```diff
@@ -51,19 +52,47 @@ class OptimizerCloneableParamState : public OptimizerParamState {
...
+    return static_cast<uint32_t>(explicitly_set_fields_.to_ulong());
+  }
+
+  void set_field_mask(uint32_t mask) {
+    explicitly_set_fields_ = std::bitset<32>(mask);
+  }
+
+ protected:
+  void mark_field_as_set(size_t field_id) {
+    if (field_id < 32) {
```
janeyx99: Comment why we need to check < 32 here (bitset len) + we should probably error in the else case.

stmcgovern:
Thanks for your comments @janeyx99 . That makes sense! I certainly agree that users should not have to be aware of these implementation details. I wasn't sure how much the runtime performance cost impacted your review. will revisit and simplify.

stmcgovern:
Hi @janeyx99, here is my preferred approach. It does everything in optimizer.h/cpp, doesn't introduce any new API and most work is done at compile time. I've tried to follow how c10 uses template metaprogramming style. C++20 concepts can help smooth out some of the  boilerplate, when they become available in PyTorch.  Local tests are passing. Please let me know what you think. Thanks!

torch/csrc/api/include/torch/optim/optimizer.h
```diff
@@ -186,6 +489,43 @@ class TORCH_API Optimizer {
   /// Deserializes the optimizer state from the given `archive`.
   virtual void load(serialize::InputArchive& archive);
 
+ private:
+  /// Helper function to try merging for a specific optimizer type
+  template <typename OptimizerType>
+  static bool try_merge_optimizer_type(
```
janeyx99: make these private by prefixing with _!
```suggestion
  static bool _try_merge_optimizer_type(
```

torch/csrc/api/include/torch/optim/optimizer.h
```diff
@@ -66,12 +67,314 @@ class TORCH_API OptimizerOptions {
...
+        result->max_iter(user_options.max_iter());
+      }
+    }
+    if constexpr (has_max_eval<Derived>::value) {
+      if (user_options.max_eval() != constructor_defaults.max_eval()) {
+        result->max_eval(user_options.max_eval());
+      }
+    }
+    if constexpr (has_tolerance_grad<Derived>::value) {
+      if (user_options.tolerance_grad() !=
```
janeyx99: Asking for my own understanding: what happens if the user doesn't specify tolerance_grad here? is user_options.tolerance_grad() always existent?

If it is, why don't we just always set the result->attr to user_options.attr()?

janeyx99 (commented):
Much nicer! Can we privatize all the helpers?

torch/csrc/api/include/torch/optim/optimizer.h
stmcgovern: changed

torch/csrc/api/include/torch/optim/optimizer.h
stmcgovern: outdated

torch/csrc/api/include/torch/optim/optimizer.h
stmcgovern: great questions. 
1) yes it's always set - either as user specified or as a constructor default (which won't get used, since the derived optimizer defaults will be selected in the merge process).  Unspecified fields in user_options contain constructor defaults, but we don't use them aside from comparison.

2) if we always set the result->attr to user_options.attr(), then we would end up with constructor defaults in the final parameters. This is not what the user expects. User intent would be that when overriding a default, it does not affect the values of the unspecified fields (which should be the optimizer defaults, not the constructor defaults). That matches the python api expectations.

janeyx99: Ooooh the optimizer defaults and the constructor defaults are different.

torch/csrc/api/include/torch/optim/optimizer.h
```diff
@@ -66,12 +67,314 @@ class TORCH_API OptimizerOptions {
   virtual void set_lr(const double lr);
 };
 
-template <typename Derived>
+// Forward declarations for optimizer option types
```
janeyx99, line 70: Can we make the following classes and structs private as well?

torch/csrc/api/include/torch/optim/optimizer.h
```diff
@@ -66,12 +67,314 @@ class TORCH_API OptimizerOptions {
...
+template <class Derived>
 class OptimizerCloneableOptions : public OptimizerOptions {
  private:
   std::unique_ptr<OptimizerOptions> clone() const override {
     return std::make_unique<Derived>(static_cast<const Derived&>(*this));
   }
+
+  // SFINAE field detection - detects optimizer fields using public accessor methods
+  template <class T, class Enable = void>
+  struct has_lr : std::false_type {};
```
janeyx99: These structs too

janeyx99 (dismissed):
Please do privatize as much as possible so we are not inadvertently growing our API surface.

torch/csrc/api/include/torch/optim/optimizer.h
stmcgovern: These helper structs are in the private part of the class OptimizerCloneableOptions. 
Do you just want me to prefix all the implementation stuff I've added with an underscore? Is that just a style convention (happy to follow) or is there some codegen or python binding specific transformation?

torch/csrc/api/include/torch/optim/optimizer.h
stmcgovern: since these are forward declarations, I'm inclined to not change the style to use the prefix (would need touching all optimizer files). I do have to change them to be struct (which resolves the inconsistency causing clang build failure).

stmcgovern:
OK I think this is ready. Thanks for your feedback and help @janeyx99 !

janeyx99 (dismissed):

PR #161829, opened 2025-08-29
[Inductor] update exp codegen for better precision

BoyuanFeng:
Prior to this PR, we have:
```
[Default Behavior] uses `tl.math.exp({x})`:
eager diff: tensor(2.6935e-06, device='cuda:0', dtype=torch.float64)
compile diff: tensor(9.2757e-06, device='cuda:0', dtype=torch.float64)
eager_latency:0.0013996509159580942, compile_latency:0.0013981951951980592


TORCHINDUCTOR_USE_FAST_MATH=1 uses `tl.extra.libdevice.exp2(tmp0 * 1.4426950408889634)`:
eager diff: tensor(2.2315e-06, device='cuda:0', dtype=torch.float64)
compile diff: tensor(3.5329e-06, device='cuda:0', dtype=torch.float64)
eager_latency:0.0013982331859319662, compile_latency:0.0013824134564199367


Update inductor to use `tl.extra.libdevice.exp(tmp0)`:
eager diff: tensor(2.3421e-06, device='cuda:0', dtype=torch.float64)
compile diff: tensor(2.3421e-06, device='cuda:0', dtype=torch.float64)
eager_latency:0.0014109122834153282, compile_latency:0.0014062877025520593
```

Since `tl.extra.libdevice.exp` leads to both better precision and on-par latency, we use it by default now.

Note that `tl.extra.libdevice.exp` used to have a perf issue in [January 2025](https://github.com/triton-lang/triton/issues/5735) since it used due to `ex2.approx.f32` instead of `ex2.approx.ftz.f32`. So `tl.extra.libdevice.exp2(tmp0 * 1.4426950408889634)` was used as a workaround. I double checked that the issue is resolved and `tl.extra.libdevice.exp` also uses [ex2.approx.ftz.f32](https://github.com/triton-lang/triton/issues/5735#issuecomment-3238421293) today.

### DIFF
--- a/torch/_inductor/codegen/triton.py
+++ b/torch/_inductor/codegen/triton.py
@@ -1061,10 +1061,7 @@ class TritonOverrides(OpOverrides):
         Check https://github.com/triton-lang/triton/issues/5735 for
         more details.
         """
-        if config.use_fast_math:
-            return f"libdevice.exp2({x} * {TritonOverrides._LOG_2_E})"
-        else:
-            return f"tl_math.exp({x})"
+        return f"libdevice.exp({x})"
 
     @staticmethod
     @maybe_upcast_float32()
--- a/torch/_inductor/codegen/triton.py
+++ b/torch/_inductor/codegen/triton.py
@@ -1061,7 +1061,10 @@ class TritonOverrides(OpOverrides):
         Check https://github.com/triton-lang/triton/issues/5735 for
         more details.
         """
-        return f"libdevice.exp({x})"
+        if config.use_fast_math:
+            return f"libdevice.exp({x})"
+        else:
+            return f"tl_math.exp({x})"
 
     @staticmethod
     @maybe_upcast_float32()
--- a/torch/_inductor/codegen/triton.py
+++ b/torch/_inductor/codegen/triton.py
@@ -1062,9 +1062,9 @@ class TritonOverrides(OpOverrides):
         more details.
         """
         if config.use_fast_math:
-            return f"libdevice.exp({x})"
-        else:
             return f"tl_math.exp({x})"
+        else:
+            return f"libdevice.exp({x})"
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


torch/_inductor/codegen/triton.py
```diff
@@ -1061,10 +1061,7 @@ def exp(x):
         Check https://github.com/triton-lang/triton/issues/5735 for
         more details.
         """
-        if config.use_fast_math:
-            return f"libdevice.exp2({x} * {TritonOverrides._LOG_2_E})"
```
jansel, line 1065: If you just change this line does it fix the issue?  I think still having a `use_fast_math` option would be useful.

jansel (changes requested):

torch/_inductor/codegen/triton.py
BoyuanFeng: `libdevice.exp` gives the same numerics as eager (3.6707e-06 difference from fp64), which is better than `tl_math.exp` (7.1521e-06 difference from fp64). To fix the issue, we need to use `libdevice.exp` by default. I updated to keep `use_fast_math` option  and turn it on by default.

torch/_inductor/codegen/triton.py
```diff
@@ -1062,7 +1062,7 @@ def exp(x):
         more details.
         """
         if config.use_fast_math:
```
jansel, line 1064:
```suggestion
        if not config.use_fast_math:
```

I think these options are swapped (this is likely the root cause of the bug).

Fast math should be off by default.

jansel (changes requested):

torch/_inductor/codegen/triton.py
BoyuanFeng: yes agree

jansel (approved):

PR #161840, opened 2025-08-30
[BE] Move git config setup to checkout action

malfet:
Not sure why it was duplicated between build and test in the first place

And let's use the same set of setting across all repos


.github/actions/checkout-pytorch/action.yml
```diff
@@ -23,6 +23,17 @@ runs:
       id: check_container_runner
       run: echo "IN_CONTAINER_RUNNER=$(if [ -f /.inarc ] || [ -f /.incontainer ]; then echo true ; else echo false; fi)" >> "$GITHUB_OUTPUT"
 
+      - name: Enable git long paths and symlinks on Windows and disable fsmonitor daemon
```
atalman, line 26: We should probably change the comment and don't mention windows since this will be applied to all platforms

PR #161866, opened 2025-08-31
[dynamo] Error when attempting to skip due to compilation under torch dispatch mode while fullgraph=True

priyank-p0:
Fixes #161790
fix implements error message in case when dynamo tries to skip on fullgraph = True. Added test to test_modes.py.

### DIFF
--- a/torch/_dynamo/convert_frame.py
+++ b/torch/_dynamo/convert_frame.py
@@ -1820,9 +1820,8 @@ class CatchErrorsWrapper:
                 exc.unimplemented_v2(
                     gb_type="TorchDispatchMode with fullgraph=True",
                     context="Dynamo attempted to skip frame due to TorchDispatchMode",
-                    explanation="fullgraph=True was specified, which requires compilation of the entire function. "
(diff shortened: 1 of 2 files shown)


torch/_dynamo/convert_frame.py
```diff
@@ -1801,16 +1801,34 @@ def __call__(
...
             or is_skipfile
             or config.disable
-            or (
-                should_skip_due_to_torch_dispatch_mode()
-                and not getattr(self._torchdynamo_orig_backend, "_export", False)
-            )
+            or should_skip_for_torch_dispatch
         ):
+            # Check if we're trying to skip due to TorchDispatchMode with fullgraph=True
+            if should_skip_for_torch_dispatch and _get_error_on_graph_break():
```
williamwen42, line 1818: We should thread `fullgraph/one_graph` from eval_frame.py instead of relying on `_get_error_on_graph_break`.

williamwen42 (commented):
This is enough to handle the case where we skip due to compilation in a torch dispatch mode.

More work needs to be done to deal with skips in general under fullgraph=True, which I will get around to doing in a followup.

PR #161876, opened 2025-09-01
[Bug] Add more boundary check for FractionalMaxPool3d

can-gaa-hou:
This PR aims to fix the bug mentioned at [#161853](https://github.com/pytorch/pytorch/issues/161853#issuecomment-3240695121)

cc @albanD @jbschlosser @mikaylagawarecki

### DIFF
--- a/aten/src/ATen/native/FractionalMaxPool3d.cpp
+++ b/aten/src/ATen/native/FractionalMaxPool3d.cpp
@@ -67,13 +67,13 @@ TORCH_PRECOMPUTE_META_FUNC(fractional_max_pool3d)(
   int64_t inputH = input_.size(heightDim);
   int64_t inputW = input_.size(widthDim);
 
-  TORCH_CHECK(outputT + poolSizeT - 1 < inputT,
+  TORCH_CHECK((poolSizeT <= inputT) && (outputT + poolSizeT - 1 < inputT),
            "fractional_max_pool3d_out(): pool time ", poolSizeT,
            " too large relative to input time ", inputT);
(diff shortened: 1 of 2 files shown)


malfet (approved):
LGTM, but I can envision next series of errors there, when someone passes negative pool size. Can you fix this one as well?
And may be one can refactor error checking to be common between MaxPool3d and FractionalMaxPool3d?

malfet:
> @pytorchbot label "topic: not user facing"

@can-gaa-hou why do you think it's not user facing? bugfixes are indeed user facing

can-gaa-hou:
> > @pytorchbot label "topic: not user facing"
> 
> @can-gaa-hou why do you think it's not user facing? bugfixes are indeed user facing

My bad. I won't add this next time.

can-gaa-hou:
> LGTM, but I can envision next series of errors there, when someone passes negative pool size. Can you fix this one as well? And may be one can refactor error checking to be common between MaxPool3d and FractionalMaxPool3d?

The negative pool size has been checked before calling the aten operator.
https://github.com/pytorch/pytorch/blob/9a665ca3c472384e9d722bddba79e5a7680f1abd/torch/nn/modules/pooling.py#L1059-L1063

Also, the MaxPool3d operator already includes these checks.
https://github.com/pytorch/pytorch/blob/9a665ca3c472384e9d722bddba79e5a7680f1abd/aten/src/ATen/native/DilatedMaxPool3d.cpp#L32-L36

I have added more tests for these two ops.

PR #122995, opened 2024-03-29
[dtensor] improve new factory strategy

tianyu-l:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #122997
* #122996
* __->__ #122995

Previously, the new tensor out of the "new factory" all become replicated.
With this PR, if the new tensor has the same shape as the old tensor **and** the shape can be evenly sharded, then the old spec is inherited and preferred.

To accommodate this when the old tensor has sharded placements, the input args for local computation (size, stride) need to be adjusted.

cc @mrshenli @pritamdamania87 @zhaojuanmao @satgera @gqchen @aazzolini @osalpekar @jiayisuse @H-Huang @kwen2501 @awgu @penguinwu @fegin @XilunWu @wanchaol @fduwjj @wz337 @wconstab @yf225 @chauhang @d4l3k @rohan-varma

PR #161882, opened 2025-09-01
[DTensor] select strategy with no redistribute when redistribute cost is 0

tianyu-l:
* __->__ #161882

Before this PR, the `_select_strategy` always selects the first strategy with minimum redistribute cost. This causes unexpected behavior when
- multiple strategies have 0 redistribute costs
- the first one with 0 redistribute cost may perform local chunking

E.g. in memory efficient SDPA, the default orders of candidate strategies have a `Shard(2)` one before the `Replicate()` one. https://github.com/pytorch/pytorch/blob/main/torch/distributed/tensor/_ops/_matrix_ops.py#L500-L512
When the input is `Replicate()`, `_select_strategy` will pick the `Shard(2)` strategy and do local chunking first, before local computation. This is clearly unexpected to users.

In this PR, we improve `_select_strategy` so that when multiple strategies have 0 redistribute cost, we prioritize the one which keeps input unchanged.

### DIFF
--- a/torch/distributed/tensor/_sharding_prop.py
+++ b/torch/distributed/tensor/_sharding_prop.py
@@ -348,7 +348,7 @@ class ShardingPropagator:
 
             if isinstance(op_strategy, OpStrategy):
                 # single Op strategy
-                output_strategy = self._select_strategy(op_strategy)
+                output_strategy = self._select_strategy(op_strategy, op_schema)
 
                 # check if we need to redistribute the input
                 needs_redistribute = False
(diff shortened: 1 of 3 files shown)


torch/distributed/tensor/_sharding_prop.py
```diff
@@ -556,21 +556,56 @@ def propagate_op_sharding_non_cached(self, op_schema: OpSchema) -> OutputShardin
...
             assert op_spec.redistribute_cost is not None, (
                 "must set redistribute cost each OpSpec!"
             )
             redistribute_cost = sum(chain.from_iterable(op_spec.redistribute_cost))
             op_spec_costs.append(redistribute_cost)
 
+            # If there's no redistribute cost, we record the index of the strategy
+            # which doesn't need redistribute.
+            # TODO: Currently this only applies to OpStrategy selection. Requires extra
+            # logic to make it work for TupleStrategy, if needed.
```
ezyang, line 578: Bah, another good reason to figure out how to get rid of TupleStrategy lol (no action needed)

torch/distributed/tensor/_sharding_prop.py
```diff
@@ -556,21 +556,56 @@ def propagate_op_sharding_non_cached(self, op_schema: OpSchema) -> OutputShardin
...
+                        op_spec.output_spec
+                        if op_spec.input_specs is None
+                        else op_spec.input_specs[spec_idx]
+                    )
+                    if input_spec.placements != desired_spec.placements:
+                        needs_redistribute = True
+                        break
+
+                if not needs_redistribute:
+                    no_redistribute_strategy_index = strategy_idx
```
ezyang, line 592: Note to myself: since we always check this every loop iteration, the LAST strategy that has no comms and no local compute is picked. But this one should always be unique.

torch/distributed/tensor/_sharding_prop.py
```diff
@@ -556,21 +556,56 @@ def propagate_op_sharding_non_cached(self, op_schema: OpSchema) -> OutputShardin
...
             )
             redistribute_cost = sum(chain.from_iterable(op_spec.redistribute_cost))
             op_spec_costs.append(redistribute_cost)
 
+            # If there's no redistribute cost, we record the index of the strategy
+            # which doesn't need redistribute.
+            # TODO: Currently this only applies to OpStrategy selection. Requires extra
+            # logic to make it work for TupleStrategy, if needed.
+            if op_schema is not None and redistribute_cost == 0:
+                needs_redistribute = False
```
ezyang, line 580: It might be marginally clearer to name this needs_local_redistribute but I wouldn't block the PR on this.

torch/distributed/tensor/_sharding_prop.py
```diff
@@ -556,21 +556,56 @@ def propagate_op_sharding_non_cached(self, op_schema: OpSchema) -> OutputShardin
...
+                    no_redistribute_strategy_index = strategy_idx
+
         # for eager execution, we just select the one with the minimal redistribute cost
-        return strategy.strategies[op_spec_costs.index(min(op_spec_costs))]
+        min_cost = min(op_spec_costs)
+        if min_cost < 0:
+            # If there's negative cost, we select the one with the minimal cost,
+            # even if this means we need to redistribute, e.g. via local chunking.
+            # E.g. this can happen for ops in self.op_to_shape_and_stride_idx
+            # when the inputs / outputs are sharded.
```
ezyang, line 600: For my information, what exactly does it to have negative cost?! Seems suspicious haha.

torch/distributed/tensor/_sharding_prop.py
```diff
@@ -556,21 +556,56 @@ def propagate_op_sharding_non_cached(self, op_schema: OpSchema) -> OutputShardin
...
+            # If there's no redistribute cost, we record the index of the strategy
+            # which doesn't need redistribute.
+            # TODO: Currently this only applies to OpStrategy selection. Requires extra
+            # logic to make it work for TupleStrategy, if needed.
+            if op_schema is not None and redistribute_cost == 0:
+                needs_redistribute = False
+                for spec_idx, input_spec in enumerate(op_schema.args_spec):
+                    desired_spec = (
+                        op_spec.output_spec
+                        if op_spec.input_specs is None
```
ezyang, line 584: For my education, when can this be None?

ezyang (approved):
Seems fine, thanks!

torch/distributed/tensor/_sharding_prop.py
tianyu-l: Yes, I believe this `if` is at most hit once. I can't early exit because there can be negative costs.

torch/distributed/tensor/_sharding_prop.py
tianyu-l: See my earlier PR https://github.com/pytorch/pytorch/pull/122995

The point is, for new factory ops, even when there is the option to do no local (or global) redistribute, we would favor something with local redistribute.

torch/distributed/tensor/_sharding_prop.py
tianyu-l: I think for trivial `OpSpec` we have the option to not assign `input_specs`, e.g. https://github.com/pytorch/pytorch/blob/main/torch/distributed/tensor/_ops/_tensor_ops.py#L147

It hurts some readability IMO.

torch/distributed/tensor/_sharding_prop.py
tianyu-l: Oh pls forgive me for easy merging without addressing this.
Plus, this is consistent with `needs_redistribute` in the main function.

PR #161885, opened 2025-09-01
[BUG]Fixed handle cannot be hit in the cache in the IPC ExpandableSegment

mengph:
Fixed the bug that handle cannot be hit in the ipcMemHandle_to_devptr cache in the IPC scenario of ExpandableSegment.

Fixes #161884


albanD (approved):
Thanks for the fix !

PR #160214, opened 2025-08-08
Port amax to stable ABI

samanklesaria:
To enable porting torchaudio to the stable ABI, we need the `amax` operation to be accessible. This PR ports the op and provides tests that it behaves correctly. 

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @mcarilli @ptrblck @leslie-fang-intel @jgong5 @voznesenskym @penguinwu @EikanWang @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela

PR #161891, opened 2025-09-01
[STABLE ABI] Add mutable_data_ptr() and const_data_ptr() methods to torch::stable::Tensor.

pearu:
This ghstack is a prerequisite for porting torchaudio C++ extensions to use torch stable ABI, see https://github.com/pytorch/audio/issues/4074, https://github.com/pytorch/audio/issues/4075, https://github.com/pytorch/audio/issues/4076, https://github.com/pytorch/audio/issues/4077, https://github.com/pytorch/audio/issues/4078

* __->__ #161891
* #167772


mikaylagawarecki (commented):
This seems to make sense to me as a convenience utility and matches what's in TensorBase.h

Do add some lightweight tests (the file structure for testing can be seen in https://github.com/pytorch/pytorch/pull/160214/files)

pearu:
@mikaylagawarecki I have added tests for `mutable_data_ptr` and `const_data_ptr` Tensor methods.

@janeyx99 I have addressed the concern about semantics by introducing `aoti_torch_get_{mutable,const}_data_ptr` shim functions that call `{mutable,const}_data_ptr` methods of the `TensorBase` class so that stable Tensor `{mutable,const}_data_ptr` will have the corresponding semantics.

Please review.

torch/csrc/stable/tensor_struct.h
```diff
@@ -74,12 +74,30 @@ class Tensor {
...
+  // Do not add new uses of data_ptr(), use const_data_ptr() if
+  // possible, mutable_data_ptr() otherwise.
   void* data_ptr() const {
     void* data_ptr;
     TORCH_ERROR_CODE_CHECK(aoti_torch_get_data_ptr(ath_.get(), &data_ptr));
     return data_ptr;
   }
 
+  template <typename T>
+  T* mutable_data_ptr() const {
```
janeyx99: What's the difference between having the template version be on the implementation directly versus what we have in TensorBase.h:

```
  const void* const_data_ptr() const {
    return this->unsafeGetTensorImpl()->data();
  }

  void* mutable_data_ptr() const {
    return this->unsafeGetTensorImpl()->mutable_data();
  }
...
template <typename T, std::enable_if_t<!std::is_const_v<T>, int> = 0>
  const T* const_data_ptr() const;

  template <typename T, std::enable_if_t<std::is_const_v<T>, int> = 0>
  const std::remove_const_t<T>* const_data_ptr() const;

  template <typename T>
  T* mutable_data_ptr() const;
```
?

janeyx99 (commented):
I believe this is a convenience function, a "nice-to-have" over a "must-have", so we can take our time making sure that we land the right thing here. It is important to get this right as we want to minimize churn in a directory we label as stable.

torch/csrc/stable/tensor_struct.h
pearu: The TensorBase.h implementation of `template mutable_data_ptr` is defined in aten/src/ATen/templates/TensorMethods.cpp. The difference between that and `mutable_data_ptr` here appears to be in the level of correctness or safety: `template mutable_data_ptr` does not allow types that are incompatible with the tensor dtype. For instance, if `t` is a `at::Tensor` with a dtype, say, Int32 then `t.mutable_data_ptr<float>` leads to an exception while if `t` is a `torch::stable::Tensor`, there will be no exception.

So, to make these semantically identical, we would need to check if `T` and `dtype` are compatible. What do you think?

pearu: This PR now exposes non-template methods `mutable_data_ptr` and `const_data_ptr`. The corresponding template methods (with proper type checking) will be implemented in a follow-up.

janeyx99: I think we should add the templating in the same PR to match the semantics of the nonstable version.

pearu: Note that the template versions of `mutable/const_data_ptr` that were implemented in this PR before, lacked type-checking that is required for matching the semantics.
I planned to implement the type-checking in a separate PR because it involves porting https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/templates/TensorMethods.cpp to stable but it's fine with me to have all of it in a single PR as well.

pearu: > I think we should add the templating in the same PR to match the semantics of the nonstable version.

This is now done in this PR.

torch/csrc/stable/tensor_struct.h
```diff
@@ -74,12 +74,34 @@ class Tensor {
...
+    TORCH_ERROR_CODE_CHECK(
+        aoti_torch_get_const_data_ptr(ath_.get(), &data_ptr));
+    return data_ptr;
+  }
+
+  template <typename T>
+  T* mutable_data_ptr() const;
+
+  template <typename T>
+  const T* const_data_ptr() const;
```
mikaylagawarecki, line 107: Could you help me understand why we don't need to mirror the templates in TensorBase.h please

https://github.com/pytorch/pytorch/blob/dfc8a1c5ddc8401197e9ab546e03b0f745edc27b/aten/src/ATen/core/TensorBase.h#L630-L634

torch/csrc/stable/ScalarType.h
```diff
@@ -0,0 +1,50 @@
+#pragma once
+
+/*
```
mikaylagawarecki: Could you extend the comment to describe the rationale for this file (iiuc that the below mirrors  `AT_FORALL_SCALAR_TYPES_WITH_COMPLEX_AND_QINTS` and `toString` for ScalarType but the set of ScalarTypes supported by `torch::stable::Tensor.scalar_type` is a subset of that supported by libtorch)

test/cpp_extensions/libtorch_agnostic_extension/test/test_libtorch_agnostic.py
```diff
@@ -14,11 +14,38 @@
     install_cpp_extension,
     IS_WINDOWS,
     run_tests,
+    skipIfTorchDynamo,
     TestCase,
     xfailIfTorchDynamo,
 )
 
 
+def get_supported_dtypes():
```
mikaylagawarecki, line 23: do we have any hope of programmatically generating this?

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
```diff
@@ -476,14 +476,59 @@ void boxed_test_default_constructor(
...
+
+void boxed_get_any_data_ptr(
+    StableIValue* stack,
+    uint64_t num_args,
+    uint64_t num_outputs) {
+  uint64_t res = get_any_data_ptr(to<Tensor>(stack[0]), to<bool>(stack[1]));
+  stack[0] = from(res);
+}
+
+uint64_t get_template_any_data_ptr(Tensor t, Tensor r, bool mutable_) {
```
mikaylagawarecki: nit: could the ScalarType be passed directly? why do we need to use a tensor to carry it

mikaylagawarecki (commented):
I think this mostly makes sense to me and my comments are mostly nits/questions

To check my understanding this PR is mostly following the existing conventions in [TensorMethods.cpp](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/templates/TensorMethods.cpp?fbclid=IwY2xjawNc99VleHRuA2FlbQIxMQBicmlkETBhRjNHRG5BZEZIcjRTdHVzAR5JAPrn0XYLgLJnJn-Sn2do4mMsV5ezMJWDtEBXLmN9WEL09RhyvMMrGIeXUQ_aem_YaYcz2RarDfpZnsemQrKPw) to handle the `<T>` templates for `const/mutable_data_ptr<T>`, with the caveat that the version copied into `torch/csrc/stable` it's rewritten to only support types that `torch::stable::Tensor.scalar_type()` supports, is that right?

torch/csrc/stable/tensor_struct.h
pearu: This PR adds `const_data_ptr<T>` to stable while assuming that `T` is not `const` --- this corresponds to torchaudio use case.

I'll add the const `T` support for completeness.

torch/csrc/stable/ScalarType.h
pearu: The stable `toString` support is now in torch core, so this comment and `toString` implementation is obsolete here, I'll remove these.

torch/csrc/stable/tensor_struct.h
mikaylagawarecki: ok, thank you, if you haven't already added it I think it's reasonable to put this off to a later PR, I just wanted to clarify why!

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
pearu: yes, passing ScalarType works now (after moving ScalarType to headeronly). I'll update tests!

test/cpp_extensions/libtorch_agnostic_extension/test/test_libtorch_agnostic.py
pearu: The full list of supported dtypes for data ptr methods is defined by
https://github.com/pytorch/pytorch/blob/cfb3a6b3dac3e176cc4dc52f99c1e0e5dd483e6d/aten/src/ATen/templates/TensorMethods.cpp#L45-L49
that includes dtypes that are not supported by `FromImpl<ScalarType>::call`, for instance, `Float8_e8m0fnu`. So creating a `get_supported_dtypes() -> Scalar[]` utility function within `libtorch_agnostic_extension` is blocked by this. Creating a `get_supported_dtypes() -> Tensor[]` function (where the tensors in the result list contain the bytes of the corresponding dtype names) would be possible but I am not sure if the involved complexity would be worth it.

I would say that there is hope when stable ABI will implement support for all dtypes. At the moment, it discards unused/deprecated/otherwise-unusual dtypes in many places so that the set of supported dtypes must be restricted here.

torch/csrc/stable/ScalarType.h
pearu: This file is now removed.

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
pearu: Updated.

torch/csrc/stable/tensor_inl.h
```diff
@@ -33,4 +33,40 @@ inline Device Tensor::device() const {
...
+        PRED(stype, torch::headeronly::ScalarType::name), \
+        "expected scalar type " #name " but found ",      \
+        torch::headeronly::toString(stype));              \
+    return static_cast<const T*>(const_data_ptr());       \
+  }
+
+#define _PRED(S1, S2) S1 == S2
+#define _PRED_QINT(S1, S2) S1 == S2 || toUnderlying(S1) == S2
+#define DEFINE_CAST(T, name) DEFINE_DATA_PTR_CAST(T, name, _PRED)
+#define DEFINE_CAST_QINT(T, name) DEFINE_DATA_PTR_CAST(T, name, _PRED_QINT)
```
mikaylagawarecki: How does this work for qints given that theses methods call scalar_type and the from/to in `stable::Tensor::scalar_type()` 
https://github.com/pytorch/pytorch/blob/40e6f090d91026947fbec92a42564ad492f37eae/torch/csrc/stable/tensor_inl.h#L18-L23

will fall into this case
https://github.com/pytorch/pytorch/blob/40e6f090d91026947fbec92a42564ad492f37eae/torch/csrc/stable/stableivalue_conversions.h#L122-L124

Let's remove the qint case given that we don't support it for scalar_type? (and it also isn't tested above)

mikaylagawarecki (commented):
just one comment about qints, otherwise this looks good to go

torch/csrc/stable/tensor_inl.h
pearu: Indeed, `const/mutable_data_ptr` methods will raise `Not yet supported...` runtime exception on tensors with qint dtypes. However, without the `scalar_type` check the methods would be fully functional.

I'll remove the qint case for now.

pearu: Done.

mikaylagawarecki (approved):
Thanks!

torch/csrc/inductor/aoti_torch/shim_common.cpp
```diff
@@ -281,6 +281,24 @@ AOTITorchError aoti_torch_get_data_ptr(
   });
 }
 
+AOTITorchError aoti_torch_get_const_data_ptr(
```
mikaylagawarecki, line 284: oh wait! @pearu I am very sorry for only catching this now, but we have moved to adding to our own [shim.h](https://github.com/pytorch/pytorch/blob/main/torch/csrc/stable/c/shim.h) and [shim_common.cpp](https://github.com/pytorch/pytorch/blob/main/torch/csrc/shim_common.cpp)

The new naming convention there is `torch_foo` rather than `aoti_torch_foo` and the purpose of this is

1. to decouple from aoti in future development
2.  to enforce our versioning system in this new shim 

Could you move these to the new shim please (follow-up PR is ok if it's a hassle)

mikaylagawarecki: @pearu actually, I am planning to merge a pretty big test refactor, in order to avoid too much hassle for you, let's merge this and I will fix this in a follow up

pearu: Dropping "aoti" for "stable" makes sense.

pearu: > @pearu actually, I am planning to merge a pretty big test refactor, in order to avoid too much hassle for you, let's merge this and I will fix this in a follow up

Thanks!

PR #161894, opened 2025-09-01
[STABLE ABI] Add device_type and device_index optional arguments to new_empty.

pearu:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #161911
* __->__ #161894
* #164123
* #163973
* #161891

PR #161892, opened 2025-09-01
[STABLE ABI] Add sizes() and strides() methods to torch::stable::Tensor

pearu:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #161897
* #161967
* #161911
* #161894
* __->__ #161892
* #163973
* #161891

PR #161895, opened 2025-09-01
[STABLE ABI] Add copy_ operation.

pearu:
* #161897
* #161967
* #161911
* #161894
* #161892
* #161891
* #161896
* __->__ #161895


torch/csrc/stable/ops.h
```diff
@@ -201,4 +201,20 @@ inline Tensor zero_(Tensor& self) {
   return to<Tensor>(stack[0]);
 }
 
+// We expect this to be the stable version of the copy_ op with
+// identical semantics to the existing copy_ op (except that it will
+// not be called as a tensor method but only as a function
+// i.e. copy_(dst, src) not dst.zero_(src)).
```
janeyx99:
```suggestion
// identical semantics to the existing copy_ op
```
We can write the except note as a general statement somewhere, but let's not make this specific op different as the note applies generally.

janeyx99 (approved):
looks p good overall! thanks! One comment about the docs.

torch/csrc/stable/ops.h
pearu: Suggestion applied.

janeyx99 (approved):

PR #155367, opened 2025-06-06
torch::stable::Tensor beginnings, mainly mem mgmt

janeyx99:
```
// The torch::stable::Tensor class is a highlevel C++ header-only wrapper around
// the C shim Tensor APIs. We've modeled this class after TensorBase, as custom
// op kernels only really need to interact with Tensor metadata (think sizes,
// strides, device, dtype). Other functions on Tensor (like empty_like) should
// live like the ATen op that they are and exist outside of this struct.
//
// There are several goals of this class over AtenTensorHandle and
// RAIIAtenTensorHandle:
// 1. torch::stable::Tensor is a nicer UX much closer to torch::Tensor than the
//    C APIs with AtenTensorHandle. Under the hood we still call to these C shim
//    APIs to preserve stability.
// 2. RAIIAtenTensorHandle boils down to a uniq_ptr that forces the user to pass
//    around ownership. This makes it difficult to pass one input into 2
//    different functions, e.g., doing something like c = a(t) + b(t) for
//    stable::Tensor t. Thus, we use a shared_ptr here.
```

This PR:
- exemplifies the above
- adds test cases in libtorch_agnostic to make sure the file actually works
- includes the results of a battle with template specialization

Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #156126
* #155977
* __->__ #155367

PR #161896, opened 2025-09-01
[STABLE ABI] Add clone method to torch::stable::Tensor

pearu:
* #161897
* #161967
* #161911
* #161894
* #161892
* #161891
* __->__ #161896


torch/csrc/stable/tensor_struct.h
```diff
@@ -189,6 +189,12 @@ class Tensor {
     return defined;
   }
 
+  Tensor clone() const {
```
mikaylagawarecki: Is `clone` a method on at::Tensor/at::TensorBase? If not I think the guiding principle (https://github.com/pytorch/pytorch/pull/155367#issuecomment-2960051820) is that we want this one to live in ops.h

pearu: `clone` is now a ops.h function.

torch/csrc/stable/ops.h
```diff
@@ -217,4 +217,13 @@ inline Tensor copy_(
   return to<Tensor>(stack[0]);
 }
 
+// We expect this to be the stable version of the clone op that takes
+// in no kwargs (memory_format). We will add kwargs support in the
+// future.
+inline Tensor clone(const Tensor& self) {
+  AtenTensorHandle ret = nullptr;
+  TORCH_ERROR_CODE_CHECK(aoti_torch_clone(self.get(), &ret));
```
janeyx99: We'd eventually rewrite this with aoti_torch_call_dispatcher (when we add MemoryFormat support) so I'd prefer if this PR did that now (and just passed in an empty value for the third arg).

janeyx99 (commented):
This PR is close!

torch/csrc/stable/ops.h
pearu: It looks like MemoryFormat is already supported in stable, so I added the optional `memory_format` argument to clone.

janeyx99: Hm it's not supported as MemoryFormat on the user side (it's still int32_t), so we shouldn't add it for now.

torch/csrc/stable/ops.h
```diff
@@ -215,4 +217,15 @@ inline Tensor copy_(
   return to<Tensor>(stack[0]);
 }
 
+// We expect this to be the stable version of the clone op.
+inline Tensor clone(
+    const Tensor& self,
+    std::optional<c10::MemoryFormat> memory_format = std::nullopt) {
```
janeyx99: Let's not support the optional kwarg just yet

torch/csrc/stable/ops.h
pearu: Now using aoti_torch_call_dispatcher and removed memory_format support.

torch/csrc/stable/ops.h
pearu: Optional memory_format support is now removed.

janeyx99 (approved):
thank u!

PR #161900, opened 2025-09-01
Update misleading torch.sparse_coo_tensor error check

### LINKED ISSUE #160622: torch.sparse_coo_tensor fails for valid 1D case with misleading dimension error
### 🐛 Describe the bug

import torch

def generate_cartesian_product(*tensors):
    if not all(isinstance(t, torch.Tensor) for t in tensors):
        raise ValueError("All inputs must be torch.Tensor")
    if not all(t.dim() == 1 for t in tensors):
        raise ValueError("All tensors must be 1-dimensional")
    if not tensors[0].is_sparse:
        tensors = (tensors[0].to_sparse(),) + tensors[1:]
    result = torch.cartesian_prod(*tensors)
    print(result)
    return result

# Create first tensor as sparse
indices = torch.tensor([[0], [1], [2]])
values = torch.tensor([1, 2, 3])
tensor1 = torch.sparse_coo_tensor(indices, values, size=(3,))  # Fails here

tensor2 = torch.tensor([4, 5])
generate_cartesian_product(tensor1, tensor2)

### Versions

  File "repro.py", line 18, in <module>
    tensor1 = torch.sparse_coo_tensor(indices, values, size=(3,))
RuntimeError: number of dimensions must be sparse_dim (3) + dense_dim (0), but got 1

cc @alexsamardzic @nikitaved @pearu @cpuhrsch @amjames @bhosmer @jcaip

parsshar-RH:
Fixes #160622

### Summary
Updated the misleading torch.sparse_coo_tensor error check to provide clear context. 
earlier:
`RuntimeError: number of dimensions must be sparse_dim (3) + dense_dim (0), but got 1`

Updated:
`RuntimeError: 'len(size) == sparse_dim + dense_dim' is not satisfied: len(size) = 1, sparse_dim = 3, dense_dim = 0`

**Impacts:**

- Comprehensive error message that will improve developer experience.
- module: sparse 

cc @alexsamardzic @nikitaved @pearu @cpuhrsch @amjames @bhosmer @jcaip

### DIFF
--- a/aten/src/ATen/SparseTensorImpl.h
+++ b/aten/src/ATen/SparseTensorImpl.h
@@ -133,12 +133,12 @@ struct TORCH_API SparseTensorImpl : public TensorImpl {
         "resize_ called on tensor with symbolic shape")
     TORCH_CHECK(
         sparse_dim + dense_dim == static_cast<int64_t>(size.size()),
-        "number of dimensions must be sparse_dim (",
+        "'len(size) == sparse_dim + dense_dim' is not satisfied: len(size) = ",
+        size.size(),
+        ", sparse_dim = ",
         sparse_dim,
-        ") + dense_dim (",
-        dense_dim,
-        "), but got ",
-        size.size());
+        ", dense_dim = ",
+        dense_dim);
     if (nnz() > 0) {
       [[maybe_unused]] auto constexpr alt_options_msg =
           "You could try the following options:\n\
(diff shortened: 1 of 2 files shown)


parsshar-RH:
@salilsdesai @kimishpatel @digantdesai @jianyuh

Please review the PR
Thanks

nikitaved (approved):
I like the update. Thank you! @pearu , the call is ultimately yours as you were active in the discussion.

pearu (approved):
LGTM! Thanks, @parsshar-RH!

Btw, apparently we don't have tests for exceptions that may be raised from sparse COO tensors construction. For sparse compressed tensors we have
https://github.com/pytorch/pytorch/blob/b477fb106f4350428c2e94ee94680a740ed6a52b/test/test_sparse_csr.py#L876
However, I don't expect that the missing tests for sparse COO tensors covering all invalid input cases should be implemented in this PR, but it would be a nice follow-up that would reveal if there exists more exception messages that could be improved similar to this PR.

parsshar-RH:
@pearu @nikitaved 

Thanks for the valuable feedback and reviews.
I appreciate you pointing out the lack of exception tests for sparse COO tensors.
I will address the feedback in a follow-up PR

Thanks

PR #161901, opened 2025-09-01
constant_pad_nd: allow zero-sized dims with mixed positive/negative pads; added regression test

### LINKED ISSUE #161014: Inconsistent constant_pad_nd behavior with negative padding

PluvioXO:
Fixes #161014

### Summary
`aten::constant_pad_nd` incorrectly errors when mixed positive/negative pads
produce a zero-sized dimension (e.g., on a [5, 3] tensor, pads [-1, -2, 1, 1]
should yield shape [7, 0]). Zero-sized dims are valid; only negative sizes should error.

### What changed
- Relax size check from `> 0` to `>= 0`:
  - Python decomposition: torch/_refs/__init__.py
  - Native implementation: aten/src/ATen/native/PadNd.cpp
- Added regression tests: test/test_constant_pad_nd.py

### Test plan
- `pytest -q test/test_constant_pad_nd.py` passes locally.
- Verified:
  - `constant_pad_nd(x, [-1, -2]) -> (5, 0)` still OK
  - `constant_pad_nd(x, [-1, -2, 1, 1]) -> (7, 0)` now OK
  - Truly negative result (e.g., [-2, -2]) still raises.

### Notes
- Behavior aligns eager and decomposition/meta paths.

### DIFF
--- a/aten/src/ATen/native/PadNd.cpp
+++ b/aten/src/ATen/native/PadNd.cpp
@@ -73,7 +73,7 @@ Tensor constant_pad_nd(const Tensor& self, IntArrayRef pad, const Scalar& value)
     for (const auto i : c10::irange((size_t)l_pad)) {
         auto pad_idx = pad.size() - ((i + 1) * 2);
         auto new_dim = input_sizes[l_diff + i] + pad[pad_idx] + pad[pad_idx + 1];
-        TORCH_CHECK(new_dim > 0, "The input size ", input_sizes[l_diff + i], ", plus negative padding ",
+        TORCH_CHECK(new_dim >= 0, "The input size ", input_sizes[l_diff + i], ", plus negative padding ",
                  pad[pad_idx], " and ", pad[pad_idx + 1], " resulted in a negative output size, "
                  "which is invalid. Check dimension ", l_diff + i, " of your input.");
         new_shape.emplace_back(new_dim);
--- a/test/test_constant_pad_nd.py
+++ b/test/test_constant_pad_nd.py
@@ -29,7 +29,7 @@ class TestConstantPadNd(TestCase):
         assert z.numel() == 0
 
     @onlyNativeDeviceTypes
-    @dtypes(*floating_types)
+    @dtypes(*floating_types())
     def test_constant_pad_nd_devices(self, device, dtype):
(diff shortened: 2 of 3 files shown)


PluvioXO:
Hi! This PR fixes the constant_pad_nd mixed positive/negative padding case (allows zero-sized dims).
Could a maintainer please add the appropriate labels (e.g., module: error checking as on issue #161014)?
This change is not BC-breaking. Thanks!

malfet (changes requested):
Change is probably OK, but please stick to regular testing patterns rather than invent new ones (i.e. classes with standard Pytorch decorator for dtypes/devices)

PluvioXO:
my apologies, should all be fixed.

Issue #5040, opened 2018-02-05
Give a better error when we run out of shared memory, instead of "RuntimeError: DataLoader worker (pid 13) is killed by signal: Bus error."


miraclewkf:
When I set `num_workers=1` or other value greater than 0 in `torch.utils.data.DataLoader`, I get this error.
The detail of the error:
``` 
Traceback (most recent call last):
  File "/opt/project/train.py", line 150, in <module>
    dataset_sizes=dataset_sizes)
  File "/opt/project/train.py", line 51, in train_model
    outputs = model(inputs)
  File "/opt/conda/envs/pytorch-py3.6/lib/python3.6/site-packages/torch/nn/modules/module.py", line 357, in __call__
    result = self.forward(*input, **kwargs)
  File "/opt/conda/envs/pytorch-py3.6/lib/python3.6/site-packages/torch/nn/parallel/data_parallel.py", line 64, in forward
    inputs, kwargs = self.scatter(inputs, kwargs, self.device_ids)
  File "/opt/conda/envs/pytorch-py3.6/lib/python3.6/site-packages/torch/nn/parallel/data_parallel.py", line 75, in scatter
    return scatter_kwargs(inputs, kwargs, device_ids, dim=self.dim)
  File "/opt/conda/envs/pytorch-py3.6/lib/python3.6/site-packages/torch/nn/parallel/scatter_gather.py", line 30, in scatter_kwargs
    inputs = scatter(inputs, target_gpus, dim) if inputs else []
  File "/opt/conda/envs/pytorch-py3.6/lib/python3.6/site-packages/torch/nn/parallel/scatter_gather.py", line 25, in scatter
    return scatter_map(inputs)
  File "/opt/conda/envs/pytorch-py3.6/lib/python3.6/site-packages/torch/nn/parallel/scatter_gather.py", line 18, in scatter_map
    return list(zip(*map(scatter_map, obj)))
  File "/opt/conda/envs/pytorch-py3.6/lib/python3.6/site-packages/torch/nn/parallel/scatter_gather.py", line 15, in scatter_map
    return Scatter.apply(target_gpus, None, dim, obj)
  File "/opt/conda/envs/pytorch-py3.6/lib/python3.6/site-packages/torch/nn/parallel/_functions.py", line 68, in forward
    outputs = comm.scatter(input, ctx.target_gpus, ctx.chunk_sizes, ctx.dim, streams)
[...]

PR #161910, opened 2025-09-01
Use posix_fallocate() to reserve disk space for shared memory

wenjianhn:
Shared memory is allocated by creating a file in /dev/shm (by default) that can run out of space. Pytorch reserves the file size by calling ftruncate() that creates a sparse file, so it succeeds even if sufficient disk space is not available.

This could lead to a situation when a shared memory region is successfully created but a subsequent access to a shared memory page results in SIGBUS due to the disk being full.

Using posix_fallocate() instead of ftruncate() eliminates this problem because the former syscall always allocates space and it returns an error if the disk is full.

Related to https://github.com/pytorch/pytorch/issues/5040


aten/src/ATen/MapAllocator.cpp
```diff
@@ -292,6 +292,28 @@ MapAllocator::MapAllocator(WithFd, std::string_view filename, int fd, int flags,
           if (ftruncate(fd, static_cast<off_t>(size)) == -1) {
             TORCH_CHECK(false, "unable to resize file <", filename_, "> to the right size: ", c10::utils::str_error(errno), " (", errno, ")");
           }
+
+#if !defined(__APPLE__) && !defined(__ANDROID__)
```
Skylion007: Would it be a better to use a CMake check function exists? Really no better way to check if it exists? Or no more portable way to preallocate it?

wenjianhn: Added `CHECK_FUNCTION_EXISTS(posix_fallocate HAVE_POSIX_FALLOCATE)`. Thanks

mikaylagawarecki (approved):
Thanks

PR #161914, opened 2025-09-01
[easy] Handle Autotuners in get_triton_source_codes_for_gm

jamesjwu:
* __->__ #161914

Some triton kernels are autotuners, in that case, grab the function from the autotuner.


torch/_functorch/_aot_autograd/autograd_cache.py
```diff
@@ -331,6 +333,11 @@ def get_triton_source_codes_from_gm(
         )
 
         for kernel in triton_kernels:
+            from triton.runtime.autotuner import Autotuner
+
+            if isinstance(kernel, Autotuner):
+                # Grab the Inner JITFunction
+                kernel = kernel.fn
```
zou3519, line 340: this is for caching, right? Do we need to include the autotuner configs somewhere?

jamesjwu: We're not caching it, we're just using it to add the source code of triton kernels to the hash key. Don't think the specific configs matter in this case

oulgen (approved):

PR #161924, opened 2025-09-01
[inductor] Fix removal of constexpr args from the launcher signature

kundaMwiza:
Fixes the case described below which occurs when:
- A user `torch.compile`s a function that uses a triton kernel.
- `TORCHINDUCTOR_DUMP_LAUNCH_PARAMS=1` .

Problem:

If the user defined triton kernel is not autotuned:

```python
import os
os.environ["TORCHINDUCTOR_DUMP_LAUNCH_PARAMS"] = "1"
@triton.jit
def kernel(..., BLOCK_SIZE: tl.constexpr):
    ...
@torch.compile
def fn(..)
    kernel[..](..., 128)

fn(..)
```

Then In `triton_heuristics. _interpret_args_grid`, `filter_signature` function:

```python
        def filtered_signature() -> list[str]:
                # constexprs are not passed in as args
                return [
                    x
                    for x in self.triton_meta["signature"].keys()
                    if x not in cfg.kwargs.keys()
                ]
```

because `triton.autotune` is not used on the the `triton.jit` function, `cfg` above will be empty, and so `BLOCK_SIZE` will not be removed from the signature even though it is constexpr, even though it is removed from the arguments that are passed in to `interpret_args_grid`. This results in a mismatch between the number of parameters in the signature and the number of arguments, which leads to the error `NameError: name '_grid_2' is not defined`. 

Fix:

Use the triton jit kernel `constexprs` for args to remove.  Not sure if this is a good fix so suggestions are welcome.

Test plan: 

Added a parameter to an existing triton kernel to test for this edge case

### DIFF
--- a/torch/_inductor/runtime/triton_heuristics.py
+++ b/torch/_inductor/runtime/triton_heuristics.py
@@ -1314,7 +1314,10 @@ class CachingAutotuner(KernelInterface):
                 return [
                     x
                     for x in self.triton_meta["signature"].keys()
+                    # Also include constants that are None or equal to 1 as these
+                    # are marked as constexpr
                     if x not in cfg.kwargs.keys()
+                    and x not in self.triton_meta["constants"]
                 ]
         else:
 
@@ -1424,12 +1427,23 @@ class CompileResult(Generic[_T]):
         if triton_version_uses_attrs_dict():
             call_args = arg_names
             def_args = arg_names
-            implicit_constants = OrderedSet(
-                (
-                    "num_warps",
-                    "num_stages",
+            equals_to_1 = OrderedSet(
+                [
+                    c
+                    for c in compile_meta["constants"]
+                    if compile_meta["constants"][c] == 1
+                ]
+            )
+            implicit_constants = (
+                OrderedSet(
+                    (
+                        "num_warps",
+                        "num_stages",
+                    )
                 )
-            ).union(OrderedSet(k for k in known_constants))
+                .union(OrderedSet(k for k in known_constants))
+                .union(equals_to_1)
+            )
             if implicit_constants := implicit_constants & OrderedSet(
                 compile_meta["constants"].keys()
             ):
--- a/torch/_inductor/runtime/triton_heuristics.py
+++ b/torch/_inductor/runtime/triton_heuristics.py
@@ -1313,11 +1313,8 @@ class CachingAutotuner(KernelInterface):
                 # constexprs are not passed in as args
                 return [
                     x
-                    for x in self.triton_meta["signature"].keys()
-                    # Also include constants that are None or equal to 1 as these
-                    # are marked as constexpr
-                    if x not in cfg.kwargs.keys()
-                    and x not in self.triton_meta["constants"]
+                    for i, x in enumerate(self.triton_meta["signature"].keys())
+                    if x not in cfg.kwargs.keys() and i not in self.fn.constexprs
                 ]
         else:
 
@@ -1427,23 +1424,12 @@ class CompileResult(Generic[_T]):
         if triton_version_uses_attrs_dict():
             call_args = arg_names
             def_args = arg_names
-            equals_to_1 = OrderedSet(
-                [
-                    c
-                    for c in compile_meta["constants"]
-                    if compile_meta["constants"][c] == 1
-                ]
-            )
-            implicit_constants = (
-                OrderedSet(
-                    (
-                        "num_warps",
-                        "num_stages",
-                    )
+            implicit_constants = OrderedSet(
+                (
+                    "num_warps",
+                    "num_stages",
                 )
-                .union(OrderedSet(k for k in known_constants))
-                .union(equals_to_1)
-            )
+            ).union(OrderedSet(k for k in known_constants))
             if implicit_constants := implicit_constants & OrderedSet(
                 compile_meta["constants"].keys()
             ):
--- a/torch/_inductor/runtime/triton_heuristics.py
+++ b/torch/_inductor/runtime/triton_heuristics.py
@@ -1314,7 +1314,7 @@ class CachingAutotuner(KernelInterface):
                 return [
                     x
                     for i, x in enumerate(self.triton_meta["signature"].keys())
-                    if x not in cfg.kwargs.keys() and i not in self.fn.constexprs
+                    if i not in self.fn.constexprs
                 ]
         else:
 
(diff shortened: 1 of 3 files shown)


kundaMwiza:
@davidberard98 Can I please get a review for this PR? Thanks

kundaMwiza:
Example reproducer: 

```python
import torch
import triton
import triton.language as tl

import os

os.environ["TORCHINDUCTOR_DUMP_LAUNCH_PARAMS"] = "1"


def test_reprex() -> None:

    @triton.jit
    def elementwise_unary_1d_kernel(
        a_ptr,  # noqa: ANN001
        c_ptr,  # noqa: ANN001
        M,  # noqa: ANN001,N803
        none_arg,
        stride_m,  # noqa: ANN001
        BLOCK_SIZE_M: tl.constexpr,  # pyright: ignore [reportArgumentType] # noqa: N803
    ) -> None:

        pid_m = tl.program_id(0)

        a_blk_ptr = tl.make_block_ptr(
            base=a_ptr,
            shape=(M,),
            strides=(stride_m,),
            offsets=(pid_m * BLOCK_SIZE_M,),
            block_shape=(BLOCK_SIZE_M,),
            order=(0,),
        )
        c_blk_ptr = tl.make_block_ptr(
            base=c_ptr,
            shape=(M,),
            strides=(stride_m,),
            offsets=(pid_m * BLOCK_SIZE_M,),
            block_shape=(BLOCK_SIZE_M,),
            order=(0,),
        )

        a = tl.load(a_blk_ptr, boundary_check=(0,), padding_option="zero")
        if none_arg is None:
            c = tl.exp(a)
        else:
            c = a * 2
        tl.store(c_blk_ptr, c, boundary_check=(0,))

    @torch.compile
    def run_elementwise_unary_1d(a: torch.Tensor) -> torch.Tensor:
        assert a.is_contiguous(), "Matrix A must be contiguous"
        m = a.shape[0]
        c = torch.empty((m,), device=a.device, dtype=a.dtype)

        def grid(meta: dict) -> tuple[int]:
            return (triton.cdiv(m, meta["BLOCK_SIZE_M"]),)

        elementwise_unary_1d_kernel[grid](a, c, m, None, a.stride(0), 512)  # pyright: ignore [reportArgumentType]

        return c

    _triton_output = run_elementwise_unary_1d(torch.randn((128,), device="cuda"))
```

davidberard98 (approved):
Thanks for the fix!

torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -1280,11 +1280,23 @@ def _interpret_args_grid(
 
             def filtered_signature() -> list[str]:
                 # constexprs are not passed in as args
-                return [
```
kundaMwiza, line 1314: @davidberard98 For clarity, is the intention here to remove only parameters that were explicitly declared `constexpr`? Or should it also include equals-to-1 and `None` constants? I can see that if `triton_version_uses_attrs_dict()` evaluates to True, then its fine to not pass constants to the kernel so long as they're constexpr. Is that correct? if so instead of the more complex code that I've added we can just use `self.triton_meta["constants"]`. This will however require updating code in other places that currently doesn't remove equals-to-1 or `None` args from the signature (and launcher signature).

Explaination for the new code:

Torch compiling the following code:

```python
import os
import torch
os.environ["TRITON_INTERPRET"] = "1"

@torch.compile()
def foo(x):
    return x + 1

foo(torch.rand([256], device="cuda"))
```

Then `@triton.jit` on the generated kernel will return an `InterpretedFunction`, not a `JitFunction`. An `InterpretedFunction` does not have constexprs, but since the kernel is created by inductor, block sizes are definately there. We could use `self.triton_meta["constants"]` as suggested above, but that also includes equals-to-1 and `None` constexpr args, not just the block sizes. Also dynamo currently [does not](https://github.com/pytorch/pytorch/blob/9491d289b329e4ba4a9f5f5b1be7960671bb7840/torch/_dynamo/variables/builder.py#L1166) wrap a user defined triton kernel when `TRITON_INTERPRET=1` since it is of type `InterpretedFunction` so this code cannot be reached

davidberard98: Hey @kundaMwiza , apologies for the delay (and sorry for the short response, I've been busy preparing the triton release)

> or clarity, is the intention here to remove only parameters that were explicitly declared constexpr? Or should it also include equals-to-1 and None constants?

tbh I don't remember all the details of equals-to-1 and None; but the important part is that we just need to get the native calls to the triton.jit function to work correctly (and for all the args to be passed in the right location).

Generally, since the tests are passing, I think this change would be okay! But I have one other request - can you test it with triton 3.2 to make sure that the tests still pass? (you could do this one of two ways: either open another PR with this change + a change to `.ci/docker/ci_commit_pins/triton.txt`, or locally install triton 3.2 and test the `test_triton_kernels.py` test and verify that everything passes)

kundaMwiza: @davidberard98  

I ran this locally using pytest on a machine with an A100 GPU and the newly added tests pass:
```shell
PASSED [0.6926s] test_triton_kernels.py::KernelTests::test_triton_kernel_empty_autotune_config_dict_backend_aot_eager
PASSED [0.6556s] test_triton_kernels.py::KernelTests::test_triton_kernel_empty_autotune_config_dict_backend_eager
PASSED [0.6019s] test_triton_kernels.py::KernelTests::test_triton_kernel_empty_autotune_config_dict_backend_inductor
PASSED [0.5591s] test_triton_kernels.py::KernelTests::test_triton_kernel_equal_to_1_arg_dump_launch_params_0_dynamic_False
PASSED [1.8776s] test_triton_kernels.py::KernelTests::test_triton_kernel_equal_to_1_arg_dump_launch_params_0_dynamic_True
PASSED [0.5879s] test_triton_kernels.py::KernelTests::test_triton_kernel_equal_to_1_arg_dump_launch_params_1_dynamic_False
PASSED [1.8701s] test_triton_kernels.py::KernelTests::test_triton_kernel_equal_to_1_arg_dump_launch_params_1_dynamic_True
PASSED [0.5798s] test_triton_kernels.py::KernelTests::test_triton_kernel_equal_to_1_float_arg_dynamic_False
```

All the other tests pass as well.

davidberard98: Awesome, thanks! I think we can probably land this then!

PR #161926, opened 2025-09-01
Selectify pytorch platform args

jaejunku:
Summary: Platform args was a buck1 concept that we decided to port over to buck2 in order to make the migration easier. However, platforms args existing in the repo blocks some buck modernization like modefile free efforts, so we're trying to get rid of the usage.

Test Plan:
CI

Rollback Plan:

Differential Revision: D81451267


albanD (approved):
SGTM

PR #161936, opened 2025-09-02
Fix _scaled_dot_product_attention_math inconsistent dtypes

CaoE:
`_scaled_dot_product_attention_math` may throw: `RuntimeError: expected mat1 and mat2 to have the same dtype, but got: float != c10::BFloat16` when mask is a high precision tensor https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/transformers/attention.cpp#L925. The add operator will make `attn`  ​a float tensor, resulting in inconsistent data types error https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/transformers/attention.cpp#L906.


Valentine233 (approved):

CaoE:
@mruberry Could you please review this PR ? Thanks.

CaoE:
@drisspg Could you please review this PR ? Thanks.

drisspg:
Why are you allowing the attention-bias to not be the same dtype as Q,k,V?

drisspg:
Which implementation of Flash are you referring to? The Gpu variant does not accept a attention bias. Efficient attention and cudnn both do, can you add tests showing that they accept this mixed dtype?

CaoE:
> Which implementation of Flash are you referring to? 

I mean https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/transformers/sdp_utils_cpp.h#L81 the check for flash attention backend does not include mask_attn.  If `scaled_dot_product_attention` fallback to `_scaled_dot_product_attention_math` we may encounter such a case: The data types of matmul inputs may not be consistent.

test/test_transformers.py
```diff
@@ -1249,6 +1249,33 @@ def func():
 
             self.assertRaises(RuntimeError, func)
 
+    @sdpa_kernel(backends=[SDPBackend.MATH])
```
drisspg: can you test all backends that supporting masking

CaoE: Added `SDPBackend.EFFICIENT_ATTENTION` and `SDPBackend.CUDNN_ATTENTION`. `EFFICIENT_ATTENTION` checks the dtype of mask and `CUDNN_ATTENTION` does not check and it can pass this test.
`MATH` can't pass the test without the changes of this PR on CUDA.

test/test_transformers.py
```diff
@@ -1249,6 +1249,45 @@ def func():
 
             self.assertRaises(RuntimeError, func)
 
+    @parametrize("kernel", [SDPBackend.MATH, SDPBackend.EFFICIENT_ATTENTION, SDPBackend.CUDNN_ATTENTION])
+    def test_scaled_dot_product_attention_matmul_inconsistent_dtypes(self, device, kernel):
+        orig_status = torch._C._get_math_sdp_allow_fp16_bf16_reduction()
```
drisspg: this patching should be in a toop level decorator

test/test_transformers.py
```diff
@@ -1249,6 +1249,45 @@ def func():
...
+                return torch.rand(shape, device=device, dtype=dtype)
+
+            size = (1, 4, 2, 8)
+            query = rand_tensor(*size)
+            key = rand_tensor(*size)
+            value = rand_tensor(*size)
+            # Create float mask to generate mixed dtype inputs for matmul
+            mask = torch.rand((1, 4, 2, 2), device=device, dtype=torch.float)
+            compiled_fn = torch.compile(fn)
+            if kernel == SDPBackend.EFFICIENT_ATTENTION and "cuda" in str(device):
```
drisspg, line 1285: If you would like to support this semantic for sdpa, you need to make updates to every kernel that can support masking. Tt does not sense to have a split semantic between cpu and cuda when there need not be

CaoE: Regarding the issue of unified mask handling, this requires alignment between CUDA, CPU, and SDP variants. For example, whether to allow the dtype of the mask to be different from that of qkv (which may cause regression), or to convert the mask type to the qkv type. This has a wide range of impacts. Do you mind if we can unify the mask issues in a future PR?

test/test_transformers.py
CaoE: Use `with math_sdp_allow_fp16_bf16_reduction(True):` instead.

test/test_transformers.py
drisspg: Yup this is the crux of the issue. I think allowing for fp32 attention_mask tensors w/ lower precision q,k,v is not an unreasonable semantic. But its not clear to me which backends support this new semantic. I would like a few things before accepting this PR:
1. I would like some estimation on what it would take to support this semantic for the efficient attention and cudnn backend. I would like to know if this would require new kernels, and if so how much the expected binary size increase would be.
2. Ideally if the work from 1 is small then I would like to see PR stack adding those implementations.
3. If the research from 1 proves to be a significant amount of work than, I would like there to be a unified filter function in sdp_utils that each backend can use to gate allowing of this feature, and tests showing which backends are enabled for this new semantic and which will throw this unified error

CaoE: Cpu flash attention supports this semantic. CUDNN attention also supports this semantic in the test. @eqy Could you please help to confirm whether cudnn attention fully supports mixed types of attn_bias and qvk? Thanks !
Efficient attention needs to modify the kernel https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/transformers/cuda/mem_eff_attention/kernel_forward.h#L137 and it doesn't seem to be a small work. I am not familiar with cuda efficient attention. Please correct me if I have any misunderstandings.

PR #161950, opened 2025-09-02
[DTensor] forbid view ops to redistribute when local split is impossible

tianyu-l:
* __->__ #161950

This PR is a followup to https://github.com/pytorch/pytorch/pull/149764.

In that PR, it only forbids illegal view due to `Flatten`; this PR also forbids illegal view caused by `Split`.

This PR also updates the error message to be less about internal implementation details, which users may find confusing.

### DIFF
--- a/torch/distributed/tensor/_ops/_view_ops.py
+++ b/torch/distributed/tensor/_ops/_view_ops.py
@@ -519,45 +519,48 @@ def propagate_shape_and_sharding(
                 return i, placement
         return None, None
 
+    # NOTE: This function has three responsibilities:
+    # 1. determine "theoretically" if an output dimension can be sharded, i.e. fill the shardable_dims map
+    # 2. determine "theoretically" the corresponding input dimension to shard on, via return value
+    # 3. throw an error when strict_view is enabled and we cannot shard an output dimension
+    # 1 and 2 doesn't require the info of whether current input is sharded.
+    # 3 requires that info, to decide whether we can error out. Maybe we can refactor
+    # to make this function purely "theoretical".
     def get_in_dim_to_shard(cmd: DimSpec) -> Optional[InputDim]:
-        # TODO(whc) this helper is pretty hard to understand, at least it should be better documented if not refactored
         if isinstance(cmd, InputDim):
             return cmd
         elif isinstance(cmd, Flatten):
             for i, dim in enumerate(cmd.input_dims):
-                if isinstance(dim, InputDim):
-                    can_shard_dim = True
-                    shard_mesh_dim, shard_placement = (
-                        maybe_get_shard_mesh_dim_and_placement(dim)
-                    )
-                    input_sharded = shard_mesh_dim is not None
-                    if i > 0:
-                        can_shard_dim = False
-                        if strict_view and input_sharded:
-                            raise RuntimeError(
-                                f"Attempted to flatten sharded dimension {i}, ",
-                                "but only the leftmost dim of a Flatten can be sharded.",
-                            )
-                    elif input_sharded:
-                        assert (
-                            shard_placement is not None and shard_mesh_dim is not None
+                # so far all Flatten is always composed of InputDims; revisit this if needed
+                assert isinstance(dim, InputDim)
+                can_shard_dim = True
+                shard_mesh_dim, shard_placement = (
+                    maybe_get_shard_mesh_dim_and_placement(dim)
+                )
+                input_sharded = shard_mesh_dim is not None
+                if i > 0:
+                    can_shard_dim = False
+                    if strict_view and input_sharded:
+                        raise RuntimeError(
+                            f"Attempted to flatten multiple dimensions, with dimension {dim.input_dim} being sharded. ",
+                            "It cannot be performed without redistribution, which is disallowed by the current operator.",
                         )
-                        tensor_dim_size = global_input_shape[shard_placement.dim]
-                        mesh_dim_size = mesh_sizes[shard_mesh_dim]
-                        if tensor_dim_size % mesh_dim_size != 0:
-                            can_shard_dim = False
-                            if strict_view:
-                                raise RuntimeError(
-                                    f"Attempted to flatten unevenly sharded dimension {i}, "
-                                    "which would require resharding the input. "
-                                    "Please explicitly redistribute the tensor instead."
-                                )
+                elif input_sharded:
+                    assert shard_placement is not None and shard_mesh_dim is not None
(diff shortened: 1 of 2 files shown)


torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -519,45 +519,48 @@ def maybe_get_shard_mesh_dim_and_placement(
...
+    # 1 and 2 doesn't require the info of whether current input is sharded.
+    # 3 requires that info, to decide whether we can error out. Maybe we can refactor
+    # to make this function purely "theoretical".
     def get_in_dim_to_shard(cmd: DimSpec) -> Optional[InputDim]:
-        # TODO(whc) this helper is pretty hard to understand, at least it should be better documented if not refactored
         if isinstance(cmd, InputDim):
             return cmd
         elif isinstance(cmd, Flatten):
             for i, dim in enumerate(cmd.input_dims):
-                if isinstance(dim, InputDim):
```
tianyu-l, line 528: Here I changed this `if` to an assertion.

torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -519,45 +519,48 @@ def maybe_get_shard_mesh_dim_and_placement(
...
-                    can_shard_dim = True
-                    shard_mesh_dim, shard_placement = (
-                        maybe_get_shard_mesh_dim_and_placement(dim)
-                    )
-                    input_sharded = shard_mesh_dim is not None
-                    if i > 0:
+                # so far all Flatten is always composed of InputDims; revisit this if needed
+                assert isinstance(dim, InputDim)
```
wconstab, line 535: Better to add an assertion message similar to the comment?

torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -519,45 +519,48 @@ def maybe_get_shard_mesh_dim_and_placement(
...
+                can_shard_dim = True
+                shard_mesh_dim, shard_placement = (
+                    maybe_get_shard_mesh_dim_and_placement(dim)
+                )
+                input_sharded = shard_mesh_dim is not None
+                if i > 0:
```
wconstab, line 541: @zpcore fyi when you land ordered sharding I think you'll have to change this logic to be about sharding order not implicit left to right

torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -576,6 +579,14 @@ def get_in_dim_to_shard(cmd: DimSpec) -> Optional[InputDim]:
                     out_size % mesh_dim_size == 0 for mesh_dim_size in mesh_sizes
                 ]
 
+                shard_mesh_dim, _ = maybe_get_shard_mesh_dim_and_placement(in_dim)
+                if strict_view and shard_mesh_dim is not None:
+                    if not shardable_dims[in_dim.input_dim][shard_mesh_dim]:
+                        raise RuntimeError(
+                            f"Attempted to split the sharded dimension {in_dim.input_dim} into multiple subdimensions. ",
+                            "It cannot be performed without redistribution, which is disallowed by the current operator.",
```
wconstab, line 587: what would the semantics be by the way?  (assuming we did the redistribution for you) - we don't let you have 2 different tensor dims sharded on the same mesh dim today do we?  so it basically means either (a) you're banned from splitting any sharded dim since the output placements would be ambiguous, or (b) we have one of the output dims become replicated, and we have a convention for which one stays sharded?

torch/distributed/tensor/_ops/_view_ops.py
zpcore: Thanks, noted. I think the whole function `propagate_shape_and_sharding` assumes we can only operate on the "leftmost" dimension as in the comment.

torch/distributed/tensor/_ops/_view_ops.py
tianyu-l: So let's see what happens when we split a dim into two dims, with a simple example.

Before view: size (4,), `Shard(0)`
After view: size (2, 2)

Case 1: Without distribution, say num devices = 2
After view, the placement is still `Shard(0)`. The result sharding is determined by the actual placement without redistribute. It's less about convention, but more about fact.

Case 2: needs redistribute, with num devices = 4
While one might think after view, the results is a legal sharding `[S(0), S(1)]` assuming the underlying mesh is 2D, we actually don't have such a 2D mesh / are not allowed to create one on the fly.
With original 1D mesh, there is no expression for such placement, i.e.
> we don't let you have 2 different tensor dims sharded on the same mesh dim today

So we have to redistribute before view, from `S(0)` to `Replicate()`. After view, it is still `Replicate()`.

wconstab: Ahh I see, you are right, I was not thinking correctly about case 1

torch/distributed/tensor/_ops/_view_ops.py
ezyang: Review with whitespace changes hidden!

torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -519,45 +519,48 @@ def maybe_get_shard_mesh_dim_and_placement(
...
+                shard_mesh_dim, shard_placement = (
+                    maybe_get_shard_mesh_dim_and_placement(dim)
+                )
+                input_sharded = shard_mesh_dim is not None
+                if i > 0:
+                    can_shard_dim = False
+                    if strict_view and input_sharded:
+                        raise RuntimeError(
+                            f"Attempted to flatten multiple dimensions, with dimension {dim.input_dim} being sharded. ",
+                            "It cannot be performed without redistribution, which is disallowed by the current operator.",
```
ezyang, line 546: @weifengpy This is exactly what we want to relax; if we support a hierarchical placement then this error message is obsolete.

ezyang (approved):

torch/distributed/tensor/_ops/_view_ops.py
ezyang: I enjoyed this example, it was good to think about.

torch/distributed/tensor/_ops/_view_ops.py
tianyu-l: @ezyang 
wait while it sounds appealing, +-indentation can change logic, no?

torch/distributed/tensor/_ops/_view_ops.py
weifengpy: great - now I have an concrete example

PR #161967, opened 2025-09-02
[STABLE ABI] Add accessor template method to torch::stable::Tensor

pearu:
Note: this PR relies on ATen/core/TensorAccessor.h but it does not use `ArrayRef.h`-dependent features. Is there a need to port a lightweight version of `TensorAccessor` to torch stable ABI?

* #161897
* __->__ #161967
* #161911
* #161894
* #161892
* #163973
* #161891


torch/csrc/stable/tensor_struct.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <ATen/core/TensorAccessor.h>
```
mikaylagawarecki, line 3: I believe the principle is we should not include from ATen in this file

As you point out in the PR description, the use of non-header only ArrayRef seems to be the main issue that prevents TensorAccessor from being header-only

From earlier discussion, I do believe we want to provide a lightweight TensorAccessor for convenience. I believe Sam might have ported some version of one directly in torchaudio in his draft PRs 😅 

Similar to my comment on the size/stride PR, let's hash this one out when @janeyx99  is back

PR #161983, opened 2025-09-02
[SymmMem] Increase minimum nthreads to cover sync needs in NVL72

kwen2501:
* #162003
* #161984
* __->__ #161983

`sync_remote_blocks` maps threads to peers. Previously min nthreads is warp size, which is too small to cover NVL72. Bumping it.

### DIFF
--- a/torch/csrc/distributed/c10d/symm_mem/CUDASymmetricMemoryOps.cu
+++ b/torch/csrc/distributed/c10d/symm_mem/CUDASymmetricMemoryOps.cu
@@ -19,6 +19,9 @@
 #include <torch/csrc/distributed/c10d/symm_mem/CUDASymmetricMemory-inl.h>
 #include <torch/csrc/distributed/c10d/symm_mem/CUDASymmetricMemory.hpp>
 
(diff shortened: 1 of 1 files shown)


torch/csrc/distributed/c10d/symm_mem/CUDASymmetricMemoryOps.cu
```diff
@@ -112,9 +115,9 @@ void init_elementwise_launch_config(
 
   if (numel_per_split <= max_num_threads * numel_per_thread) {
     num_blocks = 1;
-    num_threads = at::round_up(
-        at::ceil_div(numel_per_split, numel_per_thread),
-        static_cast<size_t>(at::cuda::warp_size()));
+    num_threads = at::ceil_div(numel_per_split, numel_per_thread);
+    num_threads = max(num_threads, MAX_CUDA_P2P_DOMAIN_SIZE);
```
ngimel: Better to pass world_size here and make max equal to world size

kwen2501: Thanks. Fixed :)

ngimel (approved):

PR #161987, opened 2025-09-02
Check ASAN/UBSAN coverage in c10 the same way we do ATen & csrc

swolchok:
* __->__ #161987

I'm not certain we have ASAN configured properly for c10. Let's make sure we do.


swolchok:
Looks like ASAN is on but UBSAN is not!

PR #162003, opened 2025-09-02
[SymmMem] Better tuning of A2AV based on accurate node boundary

kwen2501:
* __->__ #162003

Use `world_within_direct_access()` to distinguish intra- vs inter- node. 
Previously we assumed a fixed node size of 8, which is not true for NVL72.

Also added env var `TORCH_SYMMMEM_NBLOCKS` for control.


torch/csrc/distributed/c10d/symm_mem/env.hpp
```diff
@@ -0,0 +1,19 @@
+#include <cstdlib>
+#include <cstring>
+
+namespace c10d::symmetric_memory {
+
+static int getenv_nblocks() {
+  static int num_blocks = -1; // Uninitialized
+  if (num_blocks == -1) {
+    const char* str = getenv("TORCH_SYMMMEM_NBLOCKS");
```
Skylion007: Don't we have a c10 version that is thread safe?

torch/csrc/distributed/c10d/symm_mem/env.hpp
```diff
@@ -0,0 +1,19 @@
...
+    if (str && strlen(str) > 0) {
```
Skylion007: Use Std::strlen and std::atoi

torch/csrc/distributed/c10d/symm_mem/env.hpp
kwen2501: Done!

torch/csrc/distributed/c10d/symm_mem/env.hpp
kwen2501: Done!

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
```diff
@@ -280,6 +280,21 @@ __global__ void allToAllV(void *send_data, void *recv_data, int64_t* in_out_spli
 #endif
 }
 
+static int get_a2a_nblocks(size_t size, int world_size, bool intra_node) {
+  // Check user setting first
+  int num_blocks = c10d::symmetric_memory::getenv_nblocks();
+  if (num_blocks > 0) {  // set by user
+    return num_blocks;
```
ngimel, line 287: I think you want to treat it as max_blocks, and not just return

ngimel (approved):

torch/csrc/distributed/c10d/symm_mem/env.hpp
```diff
@@ -0,0 +1,18 @@
...
+    if (str.has_value()) {
+      num_blocks = std::stoi(str.value());
+    } else {
+      num_blocks = -2; // Not set
+    }
+  }
+  return num_blocks;
+}
+
+} // namespace c10d::symmetric_memory
```
fduwjj, line 18: you want to make linter happy here.

fduwjj (approved):

PR #162007, opened 2025-09-02
[MPS] enable cat op for sparse

Isalia20:
Enable cat op for sparse on MPS

cc @kulinseth @albanD @malfet @DenisVieriu97 @jhavukainen


malfet (dismissed):

malfet (approved):
Revert were bogus

PR #162009, opened 2025-09-02
build: Add fallback commands to setup.py

seemethere:
* #162010
* __->__ #162009

Adds fallback commands for the following:
* python setup.py install
* python setup.py develop

Ideally these should just work and should provide backwards compat.

Thought process here is that multiple people rely on these commands and just because setuptools wants to drop support for this I don't think a lot of our downstream users who build from source are expecting these to be gone.

This should provide some room for developers to move away from these commands until we have a unified frontend for doing all of these commands that should abstract most of these away.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -422,7 +422,10 @@ for i, arg in enumerate(sys.argv):
         EMIT_BUILD_WARNING = True
     if arg == "develop":
         print(
-            "WARNING: Redirecting 'python setup.py develop' to 'pip install -e . -v --no-build-isolation', for more info see https://github.com/pytorch/pytorch/issues/152276",
+            (
+                "WARNING: Redirecting 'python setup.py develop' to 'pip install -e . -v --no-build-isolation',"
+                " for more info see https://github.com/pytorch/pytorch/issues/152276"
+            ),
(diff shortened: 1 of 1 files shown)


clee2000 (approved):

setup.py
```diff
@@ -420,6 +420,33 @@ def _get_package_path(package_name: str) -> Path:
     if arg == "rebuild" or arg == "build":
         arg = "build"  # rebuild is gone, make it build
         EMIT_BUILD_WARNING = True
+    if arg == "develop":
+        print(
+            "Redirecting 'python setup.py develop' to 'pip install -e . -v --no-build-isolation'",
```
atalman: Should we print warning here:
```
print_box(
            """
            WARNING: Redirecting 'python setup.py develop' to 'pip install -e . -v --no-build-isolation'
            """
        )
```

atalman (approved):

setup.py
seemethere: I would use print_box but it's defined further down in the file which would require me to move it which I feel like is a bit much for this PR so going to leave as is for now.

PR #162011, opened 2025-09-02
[SymmMem] Replace unnecessary nvshmemx_collective_launch

kwen2501:
* __->__ #162011

The `nvshmemx_collective_launch` function must be used when the CUDA kernels use NVSHMEM synchronization or collective APIs (e.g., nvshmem_wait, nvshmem_barrier, nvshmem_barrier_all, or any other collective operation). CUDA kernels that do not use synchronizing NVSHMEM APIs are not required to be launched by this API. 

Also added `NVSHMEM_CHECK` or `CUDA_CHECK` around the launch calls.

### DIFF
--- a/torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cuh
+++ b/torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cuh
@@ -3,7 +3,15 @@
 #include <c10/macros/Macros.h>
 #include <ATen/ATen.h>
 
-#define NVSHMEM_CHECK(stmt, msg)                                             \
+#define NVSHMEM_CHECK(stmt)                                             \
+  do {                                                                  \
+    int result = (stmt);                                                \
+    TORCH_CHECK(                                                        \
+        result == 0,                                                    \
+        "NVSHMEM call failed, error code: " + std::to_string(result));  \
+  } while (0)
+
+#define NVSHMEM_CHECK_MSG(stmt, msg)                                         \
   do {                                                                       \
     int result = (stmt);                                                     \
     TORCH_CHECK(                                                             \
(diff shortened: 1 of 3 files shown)


torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cuh
```diff
@@ -3,7 +3,15 @@
 #include <c10/macros/Macros.h>
 #include <ATen/ATen.h>
 
-#define NVSHMEM_CHECK(stmt, msg)                                             \
+#define NVSHMEM_CHECK(stmt)                                             \
+  do {                                                                  \
+    int result = (stmt);                                                \
+    TORCH_CHECK(                                                        \
+        result == 0,                                                    \
+        "NVSHMEM call failed, error code: " + std::to_string(result));  \
```
Skylion007, line 11: Use lazy string construction and add the to_string as the third argument to the macro instead of concatenation it eagerly.

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cuh
```diff
@@ -3,7 +3,15 @@
...
+  } while (0)
+
+#define NVSHMEM_CHECK_MSG(stmt, msg)                                         \
```
Skylion007, line 14: This one should be fixed too with the eager string concatenation. Otherwise we add a lot of overhead for the fastpath of no exception

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cuh
kwen2501: @Skylion007 Sorry, following up -- what API do you refer to by "lazy string construction"?

Skylion007: If I recall, TORCH_CHECK takes arbitrary arguments where all the last n arguments are assembled into a string and concatenated/parsed only when the CHECK fails.

Just do , std::to_string instead of + std::to_string

kwen2501: I see, thanks!

PR #162015, opened 2025-09-02
[c10d] Lessen density of barrier warning

kwen2501:
* __->__ #162015

Warnings are great, but too dense when there are many ranks.

### DIFF
--- a/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
+++ b/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
@@ -5075,7 +5075,7 @@ c10::DeviceIndex ProcessGroupNCCL::guessDeviceId() const {
   // offset wrt the device id if intra-node GPUs are sharded into multiple
   // dimensions.
   int devIdx = globalRank() % localDeviceCount_;
-  if (rank_ == 0) {
+  if (devIdx == 0) { // only log on first rank of each node
     LOG(WARNING) << c10::str(
         "Guessing device ID based on global rank. ",
         "This can cause a hang if rank to GPU mapping is heterogeneous. ",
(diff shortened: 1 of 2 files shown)


torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
```diff
@@ -5075,14 +5075,12 @@ c10::DeviceIndex ProcessGroupNCCL::guessDeviceId() const {
...
   int devIdx = globalRank() % localDeviceCount_;
-  LOG(WARNING)
-      << logPrefix()
-      << c10::str(
-             " using GPU ",
-             devIdx,
-             " as device used by this process is currently unknown. ",
-             "This can potentially cause a hang if this rank to GPU mapping is incorrect. ",
-             "You can specify device_id in init_process_group() to force use of a particular device.");
+  if (rank_ == 0) {
```
d4l3k: Should we log once per host? i.e. check `LOCAL_RANK == 0` instead?

In someways if we don't log on every host it makes it very easy for this warning to be ignored

d4l3k (approved):
accepting since it's just warnings

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
kwen2501: Good point, thanks

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
```diff
@@ -5075,14 +5075,12 @@ c10::DeviceIndex ProcessGroupNCCL::guessDeviceId() const {
...
-      << c10::str(
-             " using GPU ",
-             devIdx,
-             " as device used by this process is currently unknown. ",
-             "This can potentially cause a hang if this rank to GPU mapping is incorrect. ",
-             "You can specify device_id in init_process_group() to force use of a particular device.");
+  if (devIdx == 0) { // only log on first rank of each node
+    LOG(WARNING) << c10::str(
+        "Guessing device ID based on global rank. ",
+        "This can cause a hang if rank to GPU mapping is heterogeneous. ",
```
H-Huang, line 5081: nit: "rank to GPU mapping is heterogeneous." -> "hosts have differing number of GPUs" sounds clearer to me?

H-Huang (approved):
LGTM!

PR #162016, opened 2025-09-02
Support custom callback functions in schedule

H-Huang:
* #164982
* #164976
* #164962
* __->__ #162016

This is going to be used in https://github.com/pytorch/torchtitan/issues/1682

Add a `register_custom_function` to the `_PipelineScheduleRuntime` which allows users to implement any custom function to replace the runtime operation dynamically.

The signature of the callback should look like:

```python
class _CustomFunctionProtocol(Protocol):
    def __call__(self, action: _Action, ctx: _PipelineContext) -> None: ...
```

`_PipelineContext` contains a reference to the schedule which is executing the operations.

### Testing

Added a test which adds custom methods for `FORWARD` and `OVERLAP_F_B` which are just the same implementations as those used in the default schedule runtime. Check that the schedule can still run, numerics are correct, and the callbacks are executed the correct number of times.

cc @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @ezyang


torch/distributed/pipelining/schedules.py
```diff
@@ -11,7 +11,7 @@
 from collections.abc import Callable
 from enum import Enum
 from functools import lru_cache
-from typing import Any, NamedTuple, Optional, Union
+from typing import Any, Callable, NamedTuple, Optional, Protocol, Union
```
fegin: `typing.Callable` is deprecated since 3.10. Consider use `collections.abc.Callable`.

torch/distributed/pipelining/schedules.py
```diff
@@ -2069,7 +2123,27 @@ def _perform_action(action: _Action) -> None:
...
+                    mb_indices = []
+                    if action.sub_actions is not None:
+                        for sub_a in action.sub_actions:
+                            stages.append(stage_index_to_stage[sub_a.stage_index])
+                            assert sub_a.microbatch_index is not None
+                            mb_indices.append(sub_a.microbatch_index)
+                    else:
+                        stages.append(stage_index_to_stage[action.stage_index])
+                        assert action.microbatch_index is not None
+                        mb_indices.append(action.microbatch_index)
```
fegin:
```suggestion
				   sub_actions = [action] if action.sub_actions is None else action.sub_actions
                    for sub_a in action.sub_actions:
                        stages.append(stage_index_to_stage[sub_a.stage_index])
                        assert sub_a.microbatch_index is not None
                        mb_indices.append(sub_a.microbatch_index)
```

torch/distributed/pipelining/schedules.py
```diff
@@ -1757,6 +1777,44 @@ class _PipelineScheduleRuntime(PipelineScheduleMulti):
...
+    ) -> None:
+        """
+        Register a custom function to be executed for a specific computation type.
+
+        Args:
+            computation_type: The computation type for which to register the custom function
+            custom_function: The function to execute when this computation type is encountered.
+                Must have signature: (stage: _PipelineStageBase, mb_index: int, *args, **kwargs) -> None
+        """
+        # Ensure that the computation type is valid
```
fegin, line 1804: We should also check if the `computation_type` is already registered.

torch/distributed/pipelining/schedules.py
```diff
@@ -1749,6 +1749,26 @@ def _step_microbatches(
...
+        arg_mbs: Optional[list[tuple]] = None,
+        kwarg_mbs: Optional[list[dict]] = None,
+        target_mbs: Optional[list] = None,
+        losses: Optional[list] = None,
+    ):
+        self.schedule_ref = schedule_ref
+        self.arg_mbs = arg_mbs
+        self.kwarg_mbs = kwarg_mbs
+        self.target_mbs = target_mbs
+        self.losses = losses
```
fegin, line 1765: Should we check the list length to ensure they are either None or match the length?

test/distributed/pipelining/test_schedule_multiproc.py
```diff
@@ -758,6 +766,194 @@ def test_v_shape_schedules(self, schedule_class, use_new_runtime):
...
+                stage_index_to_stage[backward_stage_index - 1].set_local_bwd_input(
+                    backward_stage.get_local_bwd_output(backward_mb_index),
+                    backward_mb_index,
+                )
+
+        # Add the callback for FORWARD computation type
+
+        base_schedule.register_custom_function(FORWARD, forward_callback)
+        base_schedule.register_custom_function(OVERLAP_F_B, overlap_callback)
+
```
fegin, line 912: I'm confused by this test. We didn't update `callback_calls` in `overlap_callback`. Do we expect `overlap_callback` to be called or not? And if yes, what's the test for `overlap_callback`?

torch/distributed/pipelining/schedules.py
H-Huang: the inputs lengths were validated earlier so no need to repeat

torch/distributed/pipelining/schedules.py
H-Huang: I will add a warning if that is the case!

test/distributed/pipelining/test_schedule_multiproc.py
H-Huang: It is because `overlap_callback` calls `forward_callback` in its implementation. To make this less confusing I will use two counters and test that both are being called.

torch/distributed/pipelining/schedules.py
```diff
@@ -1757,6 +1777,53 @@ class _PipelineScheduleRuntime(PipelineScheduleMulti):
...
+            OVERLAP_F_B,
+        ):
+            raise ValueError(
+                f"Invalid computation type {computation_type}. Only FORWARD, FULL_BACKWARD, \
+BACKWARD_INPUT, BACKWARD_WEIGHT, and OVERLAP_F_B are supported."
+            )
+
+        # Check if computation_type is already registered
+        if computation_type in self._comp_type_to_function_map:
+            logger.warning(
```
fegin, line 1819: Should we error out or warning? This seems to be an unexpected mistake from users.

fegin (approved):
LGTM

torch/distributed/pipelining/schedules.py
H-Huang: I'm thinking about whether users would want to change their custom function between steps, doesn't seem like this would be the case so maybe error-ing out is better. Let me think about it more

PR #162020, opened 2025-09-02
[inductor] Fix int64 from MutationOutput Buffer

yushangdi:
Summary:
When we have a user defined triton kernel, it marks the mutated outputs as `MutationOutput` with a NoneLayout. This MutationOutput may later be used as input to another inductor-generated triton kernel.

When we determine whether to use int32 or int64 for the inductor generated triton kernel, we need to look at the number of elements for all buffers involved. If one of the buffer is a MutationOutput, we should still consider it's number of elements, instead of skipping it.

To get a hint on the MutationOutput size, we look at the buffers corresponding to `mutation_names` in MutationOutput.

Test Plan:
```
buck run mode/opt  fbcode//caffe2/test/inductor:test_aot_inductor -- -r test_autotune_int64_user_defined_triton_kernel
```

Differential Revision: D81530083

### DIFF
--- a/torch/_inductor/codegen/simd.py
+++ b/torch/_inductor/codegen/simd.py
@@ -1423,6 +1423,15 @@ class SIMDScheduling(BaseScheduling):
             if buf.has_tensor_output()
         ]
 
+        for buf in buffers:
+            if not buf.has_tensor_output() and isinstance(buf, ir.MutationOutput):
+                mutated_bufs = buf.get_mutation_buffers()
+                buf_sizes += [
+                    buf.get_layout().storage_size()
+                    for buf in mutated_bufs
+                    if buf.has_tensor_output()
+                ]
+
         if not all(expr_fits_within_32bit(size) for size in buf_sizes):
             return False
 
--- a/test/inductor/test_aot_inductor.py
+++ b/test/inductor/test_aot_inductor.py
@@ -29,6 +29,7 @@ from torch._inductor.utils import (
     maybe_aoti_standalone_config,
     run_and_get_cpp_code,
 )
+from torch._library import capture_triton
 from torch._utils_internal import full_aoti_runtime_assert
 from torch.ao.quantization.quantize_pt2e import convert_pt2e, prepare_pt2e
 from torch.ao.quantization.quantizer.x86_inductor_quantizer import X86InductorQuantizer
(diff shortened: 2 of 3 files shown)


davidberard98 (approved):
Awesome, thanks! LGTM with some caveats:

- CI failures look legit - could you try to shrink the size of the test so that it doesn't OOM? (e.g. use int8 tensors) Also, you may need to do some skips depending on whether triton is available, etc.
- maybe give @eellison ~a day to review in case there's something I'm missing in the IR/SIMD changes?

torch/_inductor/codegen/simd.py
```diff
@@ -1423,6 +1423,15 @@ def can_use_32bit_indexing(
             if buf.has_tensor_output()
         ]
 
+        for buf in buffers:
```
eellison, line 1426: any buffer that is a mutation output should also be an input. do you know why the input mutated buffer was not being accounted for in the buffers that are iterated over [here](https://github.com/yushangdi/pytorch/blob/5b5e0698a5f560decb9bbdd150ed7b0622eb7777/torch/_inductor/codegen/simd.py#L1218-L1241) ?

yushangdi: @eellison I'm not sure. We just get two buffers in `select_index_dtype`, here `buf2` is the outupt buffer and `buf1` is the input buffer for the inductor-generated triton kernel, and it's the `MutationOutput` buffer. 

```
(Pdb) self.scheduler_nodes()
(SchedulerNode(name='op2'),)
(Pdb) self.scheduler_nodes()[0].get_buffer_names()
OrderedSet(['buf2'])
(Pdb) self.scheduler_nodes()[0].used_buffer_names()
OrderedSet(['buf1', 'buf2'])
(Pdb) buffers
[ComputedBuffer(name='buf2', layout=FixedLayout('cuda:0', torch.bfloat16, size=[s97, 512], stride=[512, 1]), data=Pointwise(device=device(type='cuda', index=0), dtype=torch.bfloat16, inner_fn=<function make_pointwise.<locals>.inner.<locals>.inner_fn at 0x7f545d964d60>, ranges=[s97, 512])), MutationOutput(name='buf1', layout=NoneLayout(device=device(type='cuda', index=0), size=[0], stride=[0]))]
```

yushangdi: The mutation output is an input for the UserDefinedTritonKernel, but here the inductor generated triton kernel takes the output of the UserDefinedTritonKernel and has the mutation output has an input. 

More specifically, `buf2` is the `writes` and `buf1`(the MutationOutput) is the reads.

yushangdi: An alternative fix is to change the line here: https://github.com/pytorch/pytorch/blob/9491d289b329e4ba4a9f5f5b1be7960671bb7840/torch/_inductor/ir.py#L6862

to 

```
 self.mutation_outputs = [
            MutationOutput(buf.layout, buf, self)
            for buf in self.mutable_args
        ]
```

I don't know why it's set to NoneLayout initially, maybe because the output's layout can potentially change?

eellison (approved):

PR #162021, opened 2025-09-02
[FSDP][Collectives] skipping reduce_scatter when world size is 1

anshul-si:
**Summary:** In its current state, FSDP collectives uses cuda synchronizations and communication ops regardless of what the world size is. However, now that replicate will use FSDP, there will be instances where group size = 1 and these synchronizations and ops will be used needlessly. I have updated fsdp_collectives to skip reduce_scatter in the foreach_reduce API when world_size ‎ = 1. I have created edited a test that uses CommDebugMode to verify that the reduce_scatter has been removed. I also edited an affected test which used 1-way FSDP by verifying and changing its assert statements for CommDebugMode. I have also added a test command. 

**Test Cases**
1. pytest test/distributed/_composable/fsdp/test_fully_shard_training.py -k test_train_parity_single_worldsize1
2. pytest test/distributed/_composable/test_composability/test_2d_composability.py -k test_tp_with_fsdp_offloading

* __->__ #162021


torch/distributed/fsdp/_fully_shard/_fsdp_collectives.py
```diff
@@ -1,3 +1,5 @@
+# mypy: disable-error-code=possibly-undefined
```
Skylion007: I think just caching world_size > 1 as a Boolean might fix this error code. Also better to have this inline type ignore  right?

mori360 (approved):

PR #161305, opened 2025-08-22
[cuBLASLt][FP8] `cuBLASLt` appears to support float8 rowwise-scaling on H100

eqy:
Following #157905 I think the macro around
```
  TORCH_INTERNAL_ASSERT(use_rowwise == false, "rowwise scaled_gemm not supported with blaslt");
```
was never updated and this would cause `float8` tests to fail. Also it appears the `Lt` accepts two inputs with `e4m3` and `e5m2` dtypes simultaneously, so removing that check here as well...

CC @lw 

cc @ptrblck @msaroufim @jerryzh168 @csarofeen @xwang233 @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben

PR #162022, opened 2025-09-02
[FP8][cuBLAS][H100] only test fp32 outputs for rowwise `_scaled_mm` on H100

eqy:
only cuBLAS supports float32 output and cuBLAS only supports rowwise for SM 9.0

Intended to land after #161305

### DIFF
--- a/test/test_matmul_cuda.py
+++ b/test/test_matmul_cuda.py
@@ -1353,22 +1353,34 @@ class TestFP8Matmul(TestCase):
         x_fp8 = to_fp8_saturated(x * x_scales, e4m3_type)
         y_fp8 = to_fp8_saturated(y * y_scales, e4m3_type)
 
-        # Calculate actual F8 mm
-        out_scaled_mm = mm_float8(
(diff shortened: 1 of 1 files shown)


test/test_matmul_cuda.py
```diff
@@ -1353,22 +1353,34 @@ def test_scaled_mm_vs_emulated_row_wise(self, base_dtype):
...
+            if base_dtype in {torch.bfloat16, torch.float16}:
+                atol, rtol = 7e-2, 7e-2
+            else:
+                atol, rtol = 2e-3, 2e-3
+
+            self.assertEqual(out_scaled_mm, out_emulated, atol=atol, rtol=rtol)
+
+        # only cuBLAS supports rowwise with fp32 output and cuBLAS only supports
+        # rowwise on SM 9.0
+        if torch.cuda.get_device_capability != (9, 0) and output_dtype == torch.float:
```
ngimel, line 1376: but H100 is sm_90 so the test should not have been failing?

eqy: This is added by the PR and checks that it indeed fails on non-sm90?

ngimel (approved):

PR #161732, opened 2025-08-28
[scan] materialize combine_fn in forward add more autograd tests 

ydwu4:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162754
* __->__ #161732
* #162025
* #161808
* #161664
* #161557

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben

PR #162025, opened 2025-09-02
[hop] refactor check input alias and mutation to be a graph pass

ydwu4:
* #162754
* #161732
* __->__ #162025
* #161808
* #161664
* #161557


torch/_higher_order_ops/utils.py
```diff
@@ -919,145 +919,75 @@ def _tensor_storage(t) -> StorageWeakRef:
...
-                    t.size(),
-                    t.stride(),
-                    dtype=t.dtype,
-                    requires_grad=t.requires_grad,
-                    device=t.device,
-                )
-            return t
+    def _get_example_value(n):
+        if not isinstance(n, torch.fx.Node):
+            return n
```
zou3519, line 931: can n be a tuple of nodes?

zou3519 (approved):

torch/_higher_order_ops/utils.py
ydwu4: for both placeholders and output nodes we've flattened them so it wouldn't be a tuple

PR #162026, opened 2025-09-02
[SymmMem] Increase signal pad size for NVL72

kwen2501:
* __->__ #162026

so that the signal calls do not step on each other's foot.

### DIFF
--- a/torch/csrc/distributed/c10d/symm_mem/CUDASymmetricMemoryTypes.hpp
+++ b/torch/csrc/distributed/c10d/symm_mem/CUDASymmetricMemoryTypes.hpp
@@ -1,8 +1,18 @@
 #pragma once
 
+#include <cstdint>
+
 namespace c10d::symmetric_memory {
 
-constexpr size_t signal_pad_size = 2048;
+// Covers NVL72
+constexpr int max_cuda_p2p_domain_size = 72;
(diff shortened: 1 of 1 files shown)


torch/csrc/distributed/c10d/symm_mem/CUDASymmetricMemoryTypes.hpp
```diff
@@ -1,8 +1,18 @@
...
 
+#include <cstdint>
+
 namespace c10d::symmetric_memory {
 
-constexpr size_t signal_pad_size = 2048;
+// Covers NVL72
+constexpr int max_cuda_p2p_domain_size = 72;
+// Maximum number of channels
+constexpr int symm_max_nblocks = 32;
```
ngimel, line 10: do we ever increase blocks to 32?

ngimel (approved):

torch/csrc/distributed/c10d/symm_mem/CUDASymmetricMemoryTypes.hpp
kwen2501: We may need to do so soon for our ops on Blackwell (NV18) due to its large bw.
I will pull a PR on that later.

PR #162033, opened 2025-09-03
[RFC] Switch all of c10d to use .h instead of .hpp

ezyang:
* __->__ #162033

It would have saved me some grief in our internal build system if the headers
were .h instead of .hpp, since the .hpp were missed by some .h only globs.  I
did this with claude code so no big deal if it is rejected.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @albanD


ezyang:
Removing the hpp headers is technically bc breaking. I don't know how prevalent usage of them outside of PyTorch is. In this PR I left stub headers for BC

wconstab:
A bit hard to load the diff, where did you put the hpp stubs? I first thought maybe you don't have to edit all the include sites since you have stubs, and this tests our bc claim too. But this would defeat the purpose of the rfc 🤦‍♂️

ezyang:
Oh it is possible I forgot to git add them

albanD (approved):
Spot checked, sounds good!

lw:
Stretch goal: could we leverage this opportunity to also switch from CamelCase to snake_case for the file names, to align with (the majority of) the rest of PyTorch?

PR #162040, opened 2025-09-03
[AsyncTP] Fixes AsyncMM

fegin:
* #162041
* __->__ #162040

The original implementation set beta to be 1, which cause the out (C) being added to the the output. Thus if the output is not initialized as zero beforehand, the output can be incorrect.

Removing the alpha and beta fixes the issue.

Thanks @ngimel to figure out the root cause.


danielvegamyhre (approved):

fegin:
@ngimel You are right, I can always reproduce the error with the two options you mentioned. But that requires `CUBLAS_WORKSPACE_CONFIG=:4096:8` to be exported before running the test. So I guess we cannot add that to the thest.

ngimel:
huh cc @eqy, exporting this `CUBLAS_WORKSPACE_CONFIG=:4096:8` env var is not longer necessary for deterministic mode. Are you getting any errors without it? If it's just warnings you can ignore.

ngimel:
Ok, this requirement will be removed once https://github.com/pytorch/pytorch/pull/161749 is landed

fegin:
> You probably can do torch.set_deterministic_debug_mode("warn") + torch.utils.deterministic.fill_uninitialized_memory=True and that way you'd still get the tests to fail, but it won't require env variable even today

This works. Add this to the test and landing the PR.

PR #162041, opened 2025-09-03
[AsyncTP] Use assertEqual instead of allClose for bf16 tests

fegin:
* __->__ #162041
* #162040

The async tp result and regular MM result are very close. If we adjust the allclose threshold, the test succeeds. This seems to indicate that the error is from numerical error of low precision.


test/distributed/test_symmetric_memory.py
```diff
@@ -555,7 +555,10 @@ def test_fused_scaled_matmul_reduce_scatter(
                 )
 
         assert outputs[0].stride() == outputs[1].stride()
-        assert torch.allclose(outputs[0], outputs[1]), (outputs[0], outputs[1])
+        assert torch.allclose(outputs[0], outputs[1], rtol=1e-3, atol=1e-3), (
```
danielvegamyhre: This is less restrictive than the default atol=1e-8, rtol=1e-5 (https://docs.pytorch.org/docs/stable/generated/torch.allclose.html). Why are we relaxing the threshold?

danielvegamyhre (approved):

test/distributed/test_symmetric_memory.py
ngimel: for bfloat16 the tolerance should be higher, also if we used `self.assertEqual` the tolerances would be dtype-dependent and the test would likely pass

fegin: `self.assertEqual` does make the test pass! Change this PR to use `self. assertEqual`.

ngimel (approved):

PR #159106, opened 2025-07-25
[DRAFT] Update CTC Loss gradients to be wrt. raw inputs

redwrasse:
(Maybe) Fixes #52241

Draft PR updating CTC loss gradients to be wrt. raw inputs, using the trick that CTC loss gradients wrt before and after `log(softmax(x))` differ by a factor of `exp(log_probs)`.

The CPU changes I built and tested locally, appear to be passing OpInfo tests. The CUDA and CUDNN code I'm less familiar with and don't have an efficient build server, hoping to rely on Pytorch CI to test those changes ( cc @soulitzer ).

PR #162042, opened 2025-09-03
Update ctc loss docs float32 input required for CuDNN

redwrasse:
Discovered while working on https://github.com/pytorch/pytorch/pull/159106 the non-obvious requirement that inputs must be float32 to use CuDNN (https://github.com/pytorch/pytorch/pull/159106#issuecomment-3189981705), otherwise the native CUDA implementation is called.

Updates the docs.

### DIFF
--- a/torch/nn/modules/loss.py
+++ b/torch/nn/modules/loss.py
@@ -2062,10 +2062,11 @@ class CTCLoss(_Loss):
         https://www.cs.toronto.edu/~graves/icml_2006.pdf
 
     Note:
-        In order to use CuDNN, the following must be satisfied: :attr:`targets` must be
+        In order to use CuDNN, the following must be satisfied: the :attr:`targets` must be
         in concatenated format, all :attr:`input_lengths` must be `T`.  :math:`blank=0`,
         :attr:`target_lengths` :math:`\leq 256`, the integer arguments must be of
-        dtype :attr:`torch.int32`.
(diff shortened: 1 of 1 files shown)


torch/nn/modules/loss.py
```diff
@@ -2062,10 +2062,11 @@ class CTCLoss(_Loss):
         https://www.cs.toronto.edu/~graves/icml_2006.pdf
 
     Note:
-        In order to use CuDNN, the following must be satisfied: :attr:`targets` must be
+        In order to use CuDNN, the following must be satisfied: the :attr:`targets` must be
         in concatenated format, all :attr:`input_lengths` must be `T`.  :math:`blank=0`,
         :attr:`target_lengths` :math:`\leq 256`, the integer arguments must be of
-        dtype :attr:`torch.int32`.
+        dtype :attr:`torch.int32`, and the :attr:`input` itself must be of
```
mikaylagawarecki: To match the name used above under the Shape: section

```suggestion
        dtype :attr:`torch.int32`, and the :attr:`log_probs` itself must be of
```

mikaylagawarecki (commented):
Seems right according to 

https://github.com/pytorch/pytorch/blob/2dd529df0092799f68ee7afcf52338276906706a/aten/src/ATen/native/cudnn/LossCTC.cpp#L98

Just a nit

mikaylagawarecki (approved):

PR #153287, opened 2025-05-09
Enable manywheel build and smoke test on main branch for ROCm

jithunnair-amd:
Fixes issue of not discovering breakage of ROCm wheel builds until the nightly job runs e.g. https://github.com/pytorch/pytorch/pull/153253

cc @jeffdaily @sunway513 @pruthvistony @ROCmSupport @dllehr-amd @jataylo @hongxiayang @naromero77amd

PR #162044, opened 2025-09-03
[ROCm] Use MI325 (gfx942) runners for binary smoke testing

jithunnair-amd:
### Motivation

* MI250 Cirrascale runners are currently having network timeout leading to huge queueing of binary smoke test jobs: 
image

* MI210 Hollywood runners (with runner names such as `pytorch-rocm-hw-*`) are not suitable for these jobs, because they seem to take much longer to download artifacts: https://github.com/pytorch/pytorch/pull/153287#issuecomment-2918420345 (this is why these jobs were specifically targeting Cirrascale runners). However, it doesn't seem like Cirrascale runners are necessarily doing much better either e.g. [this recent build](https://github.com/pytorch/pytorch/actions/runs/17332256791/job/49231006755).
* Moving to MI325 runners should address the stability part at least, while also reducing load on limited MI2xx runner capacity. 
* However, I'm not sure if the MI325 runners will do any better on the artifact download part (this may need to be investigated more) cc @jeffdaily @sunway513 @pruthvistony @ROCmSupport @dllehr-amd @jataylo @hongxiayang @naromero77amd @amdfaa

* Also removing `ciflow/binaries` and `ciflow/binaries_wheel` label/tag triggers for `generated-linux-binary-manywheel-rocm-main.yml` because we already trigger ROCm binary build/test jobs via these labels/tags in `generated-linux-binary-manywheel-nightly.yml`. And for developers who want to trigger ROCm binary build/test jobs on their PRs, they can use the `ciflow/rocm-mi300` label/tag as per this PR.

### TODOs (cc @amdfaa):
* Check that the workflow runs successfully on the MI325 runners in this PR. Note how long the test jobs take esp. the "Download Build Artifacts" step
* Once this PR is merged, clear the queue of jobs targeting `linux.rocm.gpu.mi250`

cc @jeffdaily @sunway513 @pruthvistony @ROCmSupport @dllehr-amd @jataylo @hongxiayang @naromero77amd

### DIFF
--- a/.github/scripts/generate_ci_workflows.py
+++ b/.github/scripts/generate_ci_workflows.py
@@ -22,7 +22,7 @@ LABEL_CIFLOW_BINARIES = "ciflow/binaries"
 LABEL_CIFLOW_PERIODIC = "ciflow/periodic"
 LABEL_CIFLOW_BINARIES_LIBTORCH = "ciflow/binaries_libtorch"
 LABEL_CIFLOW_BINARIES_WHEEL = "ciflow/binaries_wheel"
-LABEL_CIFLOW_ROCM = "ciflow/rocm"
+LABEL_CIFLOW_ROCM = "ciflow/rocm-mi300"
 
 
 @dataclass
--- a/.github/scripts/generate_ci_workflows.py
+++ b/.github/scripts/generate_ci_workflows.py
@@ -139,8 +139,6 @@ ROCM_SMOKE_WORKFLOWS = [
         ),
         ciflow_config=CIFlowConfig(
             labels={
-                LABEL_CIFLOW_BINARIES,
-                LABEL_CIFLOW_BINARIES_WHEEL,
                 LABEL_CIFLOW_ROCM,
             },
             isolated_workflow=True,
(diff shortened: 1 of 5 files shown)


jeffdaily (approved):

jeffdaily (approved):

jeanschmidt:
This PR won't fix problems for jobs that block viable/strict. Moving from `linux.rocm.gpu.mi250` to `linux.rocm.gpu.gfx942.1` is very likely insufficient for those jobs.

The main problem is that `linux.rocm.gpu.gfx942.1` is heavily queued right now, adding more jobs to run on them will just amplify the queue time and increase even more the delay to those runners.

I am in discussion with other members and evaluating temporarily disable those jobs until we can grow the fleet with some spare capacity in order to avoid queuing and absorb peaks even if partially down.

jithunnair-amd:
Here's an example of a failing run for MI300: https://github.com/pytorch/pytorch/pull/162044#issuecomment-3254427869
https://github.com/pytorch/pytorch/actions/runs/17469551170/job/49630020383

jithunnair-amd:
Closing in favor of https://github.com/pytorch/pytorch/pull/175784

PR #149782, opened 2025-03-21
[graph partition] support splitting on custom ops

BoyuanFeng:
This PR adds support for graph partition on custom ops. Land after #149458.

### API
This PR provides a new API to set torch._C.Tag.cudagraph_unsafe tag for custom ops. This tag will be used for graph partition.

Example usage:

```python
@torch.library.custom_op(
    "mylib::mysin",
    mutates_args=["out_list"],
    schema="(Tensor x, Tensor(a!)[]? out_list) -> Tensor",
    tags=(torch._C.Tag.cudagraph_unsafe,),
)
def mysin(x, out_list) -> torch.Tensor:
    r = x.sin()
    if out_list is not None:
        out_list[0].copy_(r)
    return r

@mysin.register_fake
def _(x, out_list) -> torch.Tensor:
    return torch.empty_like(x)
```

### Example
In this example, 1 torch-compiled region has 3 cudagraphs after splitting on 2 custom ops.

image

Code to repro:
```python
import torch

torch._inductor.config.graph_partition = True

@torch.library.custom_op(
    "mylib::movement",
    mutates_args=(),
    tags=(torch._C.Tag.cudagraph_unsafe,),
)
def movement(pic: torch.Tensor) -> torch.Tensor:
    img = pic.cpu()
    cropped_img = (img + 1) * 2
    return cropped_img.cuda() / 255.0

@movement.register_fake
def _(pic):
    return torch.empty_like(pic)

@torch.library.custom_op(
    "mylib::modify",
    mutates_args=(),
    tags=(torch._C.Tag.cudagraph_unsafe,),
)
def modify(pic: torch.Tensor) -> torch.Tensor:
    pic1 = pic + 1
    pic1_cpu = (pic1.cpu() + 1) * 2
    return pic1_cpu.cuda() + pic

@modify.register_fake
def _(pic):
    return torch.empty_like(pic)

@torch.library.custom_op("mylib::transform", mutates_args=())
def transform(pic: torch.Tensor) -> torch.Tensor:
    return (pic + 1) * 2

@transform.register_fake
def _(pic):
    return torch.empty_like(pic)

img = torch.randn(3, 64, 64, device="cuda")

def f(img):
    x = (img + 10) * 2
    y = movement(x)
    z = y + 1
    u = transform(z)
    v = 2 * u + 1
    out = modify(v)
    return out + 1

compiled_f = torch.compile(f, fullgraph=True)

eager_out = f(img)
compiled_out = compiled_f(img)

[...]

PR #162047, opened 2025-09-03
Add missing `tags` parameter to `custom_op` overload signatures

SigureMo:
It appears to be an omission in #149782.

### DIFF
--- a/torch/_library/custom_ops.py
+++ b/torch/_library/custom_ops.py
@@ -28,7 +28,7 @@ def custom_op(
     mutates_args: Union[str, Iterable[str]],
     device_types: device_types_t = None,
     schema: Optional[str] = None,
-    tags: Sequence[_C.Tag] | None = None,
+    tags: Optional[Sequence[_C.Tag]] = None,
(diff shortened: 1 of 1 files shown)


zou3519 (approved):

BoyuanFeng (approved):

torch/_library/custom_ops.py
```diff
@@ -28,6 +28,7 @@ def custom_op(
     mutates_args: Union[str, Iterable[str]],
     device_types: device_types_t = None,
     schema: Optional[str] = None,
+    tags: Sequence[_C.Tag] | None = None,
```
BoyuanFeng:
```suggestion
    tags: Optional[Sequence[_C.Tag]] = None,
```

torch/_library/custom_ops.py
```diff
@@ -40,6 +41,7 @@ def custom_op(
...
```
BoyuanFeng:
```suggestion
    tags: Optional[Sequence[_C.Tag]] = None,
```

PR #162052, opened 2025-09-03
[ROCm][inductor] Codegen support for fast_tanhf

jataylo:
Improve tanh performance on ROCm when the data type is float32 or lower precision. For float64, retain the current implementation.

Requires commits that will be present in Triton 3.6. 

cc @jeffdaily @sunway513 @jithunnair-amd @pruthvistony @ROCmSupport @hongxiayang @naromero77amd @pragupta @jerrymannil @xinyazhang @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @mlazos @dllehr-amd

### DIFF
--- a/torch/_inductor/codegen/triton.py
+++ b/torch/_inductor/codegen/triton.py
@@ -27,6 +27,7 @@ from torch._dynamo.utils import identity, preserve_rng_state
 from torch._prims_common import is_integer_dtype
 from torch.utils._ordered_set import OrderedSet
 from torch.utils._sympy.functions import CeilDiv, FloorDiv, ModularIndexing
+from torch.utils._sympy.value_ranges import bound_sympy
 from torch.utils._triton import has_triton_package, has_triton_stable_tma_api, get_triton_version
 
 from ...utils._sympy.symbol import free_symbol_is_type, prefix_str, symbol_is_type, SymT
--- a/torch/_inductor/codegen/triton.py
+++ b/torch/_inductor/codegen/triton.py
@@ -1682,7 +1682,10 @@ class TritonOverrides(OpOverrides):
     @staticmethod
     @maybe_upcast_float32()
     def tanh(x):
-        return f"libdevice.tanh({x})"
+        if torch.version.hip:
+            return f"libdevice.fast_tanhf({x})"
+        else:
+            return f"libdevice.tanh({x})"
 
     @staticmethod
     @maybe_upcast_float32()
--- a/torch/_inductor/codegen/triton.py
+++ b/torch/_inductor/codegen/triton.py
@@ -1692,9 +1692,11 @@ class TritonOverrides(OpOverrides):
         else:
             dtype = None
         if (
-            torch.version.hip
+            config.use_fast_math
+            and torch.version.hip
             and get_triton_version() > (3, 5)
             and dtype != torch.float64
+            and dtype is not None
         ):
             # Requires upstream Triton 3.6+ for latest fast_tanhf support
             # https://github.com/triton-lang/triton/pull/8551
--- a/torch/_inductor/codegen/triton.py
+++ b/torch/_inductor/codegen/triton.py
@@ -1686,7 +1686,11 @@ class TritonOverrides(OpOverrides):
     @staticmethod
     @maybe_upcast_float32()
     def tanh(x):
-        dtype = V.kernel.cse.varname_map.get(x).dtype
+        cse_var = V.kernel.cse.varname_map.get(x)
+        if cse_var and hasattr(cse_var, "dtype"):
+            dtype = cse_var.dtype
+        else:
+            dtype = None
         if (
             torch.version.hip
             and get_triton_version() > (3, 5)
--- a/torch/_inductor/codegen/triton.py
+++ b/torch/_inductor/codegen/triton.py
@@ -27,9 +27,11 @@ from torch._dynamo.utils import identity, preserve_rng_state
 from torch._prims_common import is_integer_dtype
 from torch.utils._ordered_set import OrderedSet
 from torch.utils._sympy.functions import CeilDiv, FloorDiv, ModularIndexing
-from torch.utils._sympy.value_ranges import bound_sympy
-from torch.utils._triton import has_triton_package, has_triton_stable_tma_api, get_triton_version
-
+from torch.utils._triton import (
+    get_triton_version,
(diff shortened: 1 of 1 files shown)


jeffdaily (approved):

torch/_inductor/codegen/triton.py
```diff
@@ -1325,7 +1325,13 @@ def tan(x):
     @staticmethod
     @maybe_upcast_float32()
     def tanh(x):
-        return f"libdevice.tanh({x})"
+        if config.use_fast_math and torch.version.hip:
+            if get_triton_version() > (3, 4):
```
naromero77amd: Just to double-check, do we want this to work for the case when you are on triton-lang main branch and on a commit that is after Triton 3,4 cut?

naromero77amd (approved):

torch/_inductor/codegen/triton.py
```diff
@@ -26,7 +26,8 @@
 from torch._prims_common import is_integer_dtype
 from torch.utils._ordered_set import OrderedSet
 from torch.utils._sympy.functions import CeilDiv, FloorDiv, ModularIndexing
-from torch.utils._triton import has_triton_package, has_triton_stable_tma_api
+from torch.utils._sympy.value_ranges import bound_sympy
```
naromero77amd: Is this `import bound_sympy` something that got committed by mistake?

naromero77amd (commented):
I think there is an import statement that need to be removed?

torch/_inductor/codegen/triton.py
```diff
@@ -1646,7 +1650,21 @@ def tan(x):
     @staticmethod
     @maybe_upcast_float32()
     def tanh(x):
-        return f"libdevice.tanh({x})"
+        cse_var = V.kernel.cse.varname_map.get(x)
```
eellison, line 1689: Can you query the dtype more directly ? I dont think we need to go through varname map.

See: `isinstance(var, CSEVariable) and var.dtype ...`

eellison (approved):

torch/_inductor/codegen/triton.py
```diff
@@ -1646,7 +1650,21 @@ def tan(x):
...
+        else:
+            dtype = None
+        if (
+            torch.version.hip
+            and get_triton_version() > (3, 4)
+            and dtype != torch.float64
+        ):
+            # Requires upstream Triton 3.6+ for latest fast_tanhf support
+            # https://github.com/triton-lang/triton/pull/8551
+            return f"libdevice.fast_tanhf({x})"
```
eellison, line 1703: does this change numerics ? should we avoid dispatching to this in certain settings ?  cc @v0i0

jataylo: I don't believe so, @xiaohuguo2023 should be able to comment more on implementation side.

We can set off the inductor benchmarks as a sanity check

torch/_inductor/codegen/triton.py
naromero77amd: The `@maybe_up_cast_float32()` decorator function is return a string-ified version of `x` that is potentially upcast as well (to.float32). Because `x` is stringified, I cannot directly access the dtype.

torch/_inductor/codegen/triton.py
xiaohuguo2023: fast_tanhf is fp32 only

eellison: Sure, but does it change numerics ?

torch/_inductor/codegen/triton.py
eellison: We make maybe_up_cast_float32 return a CSEVariable if that helps.  but up to you.

test/inductor/test_op_dtype_prop.py
```diff
@@ -208,6 +208,9 @@ def test_dtype_aware_codegen(self, op_name: str, load_upcast_to_fp32, input_dtyp
             # torch.sqrt lowers to tl.sqrt_rn after switching away from libdevice.sqrt
             "sqrt": "sqrt_rn",
         }
+        # ROCm uses fast_tahnf for everything input types that are not float64
```
shunting314: typo: fast_tahnf -> fast_tanhf

torch/_inductor/codegen/triton.py
```diff
@@ -1646,7 +1650,21 @@ def tan(x):
     @staticmethod
     @maybe_upcast_float32()
     def tanh(x):
-        return f"libdevice.tanh({x})"
+        cse_var = V.kernel.cse.varname_map.get(x)
+        if cse_var and hasattr(cse_var, "dtype"):
+            dtype = cse_var.dtype
+        else:
+            dtype = None
```
shunting314, line 1693: Did you see this branch happen? We probably want to skip fast tanhf if dtype is None.

shunting314 (commented):
What's the behavior if fast_tanhf get float64 inputs? It's just having bad perf or it would fail?

torch/_inductor/codegen/triton.py
```diff
@@ -1646,7 +1650,21 @@ def tan(x):
...
+        if (
+            torch.version.hip
+            and get_triton_version() > (3, 4)
+            and dtype != torch.float64
```
v0i0, line 1698: let's guard with `config.use_fast_math`

torch/_inductor/codegen/triton.py
xiaohuguo2023: compare with pytorch ref, the error is round 10^-7

torch/_inductor/codegen/triton.py
naromero77amd: It was mostly there as a precaution and because I thought the linter would complain about it otherwise.

It doesn't seem to be possible to feed a constant into tanh through the PyTorch API. We can add it as a skip in the `if`-statement on line 1658.

mlazos (approved):

v0i0 (approved):

naromero77amd:
@shunting314 Are there any further comments on the PR?

shunting314 (approved):

PR #162053, opened 2025-09-03
[Inductor] Naive foreach autotune support

jataylo:
Initial autotuning support for foreach kernels, 4x improvement for some kernels in internal workload. More improvements can surely be made here in the future. Removing num_warps for definition to enable autotune support in generated wrapper code.

Before:
triton_for_fused_18.kd 🔍 | 4.986 ms | 4.986 ms | 2.493 ms | 2 |  
triton_for_fused_6.kd 🔍 | 0.098 ms | 0.098 ms | 0.049 ms | 2 |  
triton_for_fused_7.kd 🔍 | 0.036 ms | 0.036 ms | 0.018 ms | 2 |  

After:
triton_for_fused_18.kd 🔍 | 1.273 ms | 1.273 ms | 0.636 ms | 2 |  
triton_for_fused_6.kd 🔍 | 0.044 ms | 0.044 ms | 0.022 ms | 2 |  
triton_for_fused_7.kd 🔍 | 0.024 ms | 0.024 ms | 0.012 ms | 2 |  

num_warps=8 default due to https://github.com/pytorch/pytorch/blob/main/torch/_inductor/codegen/triton_combo_kernel.py#L374

### DIFF
--- a/torch/_inductor/runtime/triton_heuristics.py
+++ b/torch/_inductor/runtime/triton_heuristics.py
@@ -3537,7 +3537,7 @@ def foreach(triton_meta, filename=None, inductor_meta=None):
     configs = []
 
     # Naive autotuning path for num_warps
-    if not inductor_meta.get("autotune_pointwise", True) and not (
+    if not (
         inductor_meta.get("max_autotune") or inductor_meta.get("max_autotune_pointwise")
     ):
         configs.append(triton.Config({}, num_stages=1, num_warps=8))
--- a/torch/_inductor/runtime/triton_heuristics.py
+++ b/torch/_inductor/runtime/triton_heuristics.py
@@ -3554,6 +3554,7 @@ def foreach(triton_meta, filename=None, inductor_meta=None):
         filename=filename,
     )
 
+
 @dataclasses.dataclass
 class GridExpr:
     """Generate code for grid size expressions in launcher"""
--- a/torch/_inductor/runtime/triton_heuristics.py
+++ b/torch/_inductor/runtime/triton_heuristics.py
@@ -3537,8 +3537,9 @@ def foreach(triton_meta, filename=None, inductor_meta=None):
     configs = []
 
     # Naive autotuning path for num_warps
-    if disable_pointwise_autotuning(inductor_meta) and not (
-        inductor_meta.get("max_autotune") or inductor_meta.get("max_autotune_pointwise")
+    if not inductor_meta.get("autotune_pointwise", True) and not (
+        inductor_meta.get("max_autotune")
+        or inductor_meta.get("max_autotune_pointwise")
     ):
         configs.append(triton.Config({}, num_stages=1, num_warps=8))
     else:
--- a/torch/_inductor/runtime/triton_heuristics.py
+++ b/torch/_inductor/runtime/triton_heuristics.py
@@ -3530,19 +3530,30 @@ def user_autotune(
     )
 
 
-def foreach(triton_meta, num_warps, filename=None, inductor_meta=None):
+def foreach(triton_meta, filename=None, inductor_meta=None):
     """
     Compile a triton foreach kernel
     """
+    configs = []
+    
+    # Naive autotuning path for num_warps
+    if disable_pointwise_autotuning(inductor_meta) and not (
+        inductor_meta.get("max_autotune") or
+        inductor_meta.get("max_autotune_pointwise")
+    ):
+        configs.append(triton.Config({}, num_stages=1, num_warps=8))
+    else:
+        for warps in [1, 2, 4, 8]:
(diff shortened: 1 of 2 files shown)


torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -3060,20 +3060,31 @@ def user_autotune(
...
+    configs = []
+
+    # Naive autotuning path for num_warps
+    if disable_pointwise_autotuning(inductor_meta) and not (
+        inductor_meta.get("max_autotune") or
+        inductor_meta.get("max_autotune_pointwise")
+    ):
+        configs.append(triton.Config({}, num_stages=1, num_warps=8))
+    else:
+        for warps in [1, 2, 4, 8]:
```
eellison, line 3545: Do you have any more details on the kernels that were improved ?

jataylo: Let me dig that out

mlazos (approved):

naromero77amd (approved):

eellison:
@jataylo can we revert this ? or turn it on only for max-autotune ? the optimizers are simple, bandwidth bound kernels and shouldnt need to autotune. or we can turn it on only for amd.

mlazos:
@jataylo I think only turning it on for max autotune is a good approach. @eellison mentioned offline that this is likely because we are creating a ton of copies of all of the parameters during benchmarking because they're mutable, so I don't think there is an easier optimization here.

jataylo:
@eellison @mlazos interesting, in theory this should only affect max autotune, or this was at least our intention. We will take a look. Also happy to have this AMD exclusive, but we also don't want to affect baseline compilation time on our side, lets figure out logic issue cc: @naromero77amd

naromero77amd:
@eellison Any additional comments before we re-land PR?

jataylo:
Moving this back to draft as I've noticed an issue with the PR. In none autotune path we're now defaulting num_warps=8 rather than taking as input, this could potentially affect performance. Taking a look.

jataylo:
@mlazos @eellison No compilation regression and confirmed there will be no perf regression from this change as the default didnt change from num_warps 8, we good to merge?

mlazos:
Yes, sounds good

mlazos (approved):

jeffdaily (approved):

mlazos:
I think if we just only have one config in the else statement we're good to reland, lmk if that's okay.

naromero77amd:
> I think if we just only have one config in the else statement we're good to reland, lmk if that's okay.

The else statement needs to have more configs, you wouldn't know apriori which is the fastest config. The point is to try more configs, this is what is doing for reductions, persistent reductions, etc.

There should be no increased compile time with autotune disabled, since there is no one config is the same one as before by construction.

More details would be helpful with respect to the slowdown which is being observed.

naromero77amd:
cc: @jataylo 

@mlazos Currently, `TORCHINDUCTOR_MAX_AUTOTUNE_POINTWISE=1` enables autotuning of the following Triton kernels:
- pointwise
- reduction
- persistent reduction
- foreach (with this PR)

`TORCHINDUCTOR_MAX_AUTOTUNE=1` would include GEMMs and convolutions that are much more time consuming to tune.

The logic is that `TORCHINDUCTOR_MAX_AUTOTUNE_POINTWISE` is low tuning overhead and high reward.

So, I would recommend against only activating foreach tuning with `TORCHINDUCTOR_MAX_AUTOTUNE=1`.

mlazos:
> cc: @jataylo
> 
> @mlazos Currently, `TORCHINDUCTOR_MAX_AUTOTUNE_POINTWISE=1` enables autotuning of the following Triton kernels:
> 
> * pointwise
> * reduction
> * persistent reduction
> * foreach (with this PR)
> 
> `TORCHINDUCTOR_MAX_AUTOTUNE=1` would include GEMMs and convolutions that are much more time consuming to tune.
> 
> The logic is that `TORCHINDUCTOR_MAX_AUTOTUNE_POINTWISE` is low tuning overhead and high reward.
> 
> So, I would recommend against only activating foreach tuning with `TORCHINDUCTOR_MAX_AUTOTUNE=1`.

@naromero77amd  Yeah I think adding an additional flag TORCHINDUCTOR_MAX_AUTOTUNE_COMBOKERNEL is perhaps a better solution. The main reason being is that 1) it takes a long time because these kernels can be large w/ optimizers and 2) it can cause OOMs when autotuning the optimizer. This is because the optimizer kernels we generate are inplace mutations which induce a clone when we autotune (we can't mutate the actual params). what do you think?

mlazos:
I talked to the internal team, they were able to remove the flag. Relanding! Thank you for your patience.

PR #162059, opened 2025-09-03
enable float32 and float16 in `torch._grouped_mm` fallback

vkuzo:
* #162159
* __->__ #162059
* #161717
* #161407

Summary:

Enables `torch.float32` and `torch.float16` options in
`torch._grouped_mm`. Note that the fast path is only enabled if `mat_a`,
`mat_b`, and `out_dtype` are `torch.bfloat16`.

Saving for future PRs:
1. enabling testing on more platforms
2. supporting out_dtype != mat_a.dtype
3. opinfo
4. better compile support

Test Plan:

```bash
// on A100 and H100
pytest test/test_matmul_cuda.py -s -k test_grouped_gemm -x
// on H100
pytest test/test_matmul_cuda.py -s -k test_scaled_grouped_gemm -x
```

Reviewers:

Subscribers:

Tasks:

Tags:


aten/src/ATen/native/GroupedMMUtils.h
```diff
@@ -59,8 +59,9 @@ std::optional<c10::ScalarType> out_dtype
     }
   }
 
-  const auto out_dtype_ = out_dtype.value_or(kBFloat16);
-  TORCH_CHECK(out_dtype_ == kBFloat16, "Only bf16 high precision output types are supported for grouped gemm");
+  const auto out_dtype_ = out_dtype.value_or(mat_a.scalar_type());
```
ngimel: this is a behavior change for fp8 gemm? There it we don't specify dtype bfloat16 would be assumed, not mat_a scalar type

vkuzo: thanks, fixed!  I'll also test this locally on a B200, since we don't have coverage in CI.

eqy:
Should the compute-capability of tests be gated if #161407 is only for sm80+

ngimel (approved):

eqy (approved):

PR #161856, opened 2025-08-31
Improve Allocator Config warning/error message

guangyey:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #161856
* #161788
* #161787
* #161786
* #161594

PR #160666, opened 2025-08-14
[PyTorch][CachingAllocatorConfig] back out D79620246 and D79620264

joshuuuasu:
Summary:
Reverting D79620246 [Refactor CUDAAllocatorConfig to reuse AcceleratorAllocatorConfig (#150312)] and D79620264 [Deprecate overleap functions in CUDAAllocatorConfig, use AcceleratorAllocatorConfig instead (#156165)].

Context: Sigrid Predictor AMD binary release pipeline is currently blocked by failing test `sl_m897405705_slatest_amd_mi300xg1` due to metrics `thrift.process_time.p99.60` significant increase since August 5: https://www.internalfb.com/services/ipnext_ip_prod/deployment/ship/container?pacing_namespace=ipnext_ip_prod%2Fdeployment%2Fcontainer&control_plane=icsp&environment=ipnext&selected_pipeline_names[0]=binary_sigrid_predictor_amd

This has lead to S551328: Sigrid HH binaries release SLA violation SEV3.

Auto multi-sect has been done: https://www.internalfb.com/lab/attribution/session/bb078c5e-51d7-487e-93b4-d714fa0e13ca, and has identified D79620246 as the blame diff. D79620264 introduced changes on top of the culprit, thus it needs to be reverted as well to get back to a working state. This back-out fixed the issue, as shown in test plan.

As a follow-up for me or pytorch folks, further investigation is needed. This issue might be caused by the cuda/rocm caching allocator config (specifically roundup_power2_divisions) not properly set in this diff.

Test Plan:
With this diff: **0.23% increase (negligible)** in `thrift.process_time.p99.60`

https://www.internalfb.com/servicelab/experiment/5901332363

Current trunk: **47.11 % increase (significant)** in `thrift.process_time.p99.60`

https://www.internalfb.com/servicelab/experiment/5901331454

Rollback Plan:

Reviewed By: XuanQi, aeolyus, 842974287, sayitmemory

Differential Revision: D80217492

PR #161786, opened 2025-08-29
[Reland] Refactor CUDAAllocatorConfig to reuse AcceleratorAllocatorConfig

guangyey:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #161856
* #161788
* #161787
* __->__ #161786
* #161594
# Motivation
This is a reland of https://github.com/pytorch/pytorch/pull/150312

PR #162066, opened 2025-09-03
[Reland] Refactor CUDAAllocatorConfig to reuse AcceleratorAllocatorConfig

guangyey:
# Motivation
This is a regular PR to faciliate re-import for internal testing...

# Additional Context
This is a reland of the stack of https://github.com/pytorch/pytorch/pull/161856

# Additional Context
As mentioned in this [comment](https://github.com/pytorch/pytorch/pull/160666#issuecomment-3260221878), @joshuuuasu figured out that the regression is caused by `round_size` in `CUDACachingAllocator`.
To test the performance of `round_size`, I compared `round_size_old` (using `CUDAAllocatorConfig`) and `round_size_new` (using `AcceleratorAllocatorConfig`)
```cpp
size_t round_size_old(size_t size) {
  using c10::cuda::CUDACachingAllocator::Native::kMinBlockSize;
  if (size < kMinBlockSize) {
    return kMinBlockSize;
  } else {
    auto divisions = CUDACachingAllocator::CUDAAllocatorConfig::roundup_power2_divisions(size);
    if (divisions > 1 && size > (kMinBlockSize * divisions)) {
      return roundup_power2_next_division(size, divisions);
    } else {
      return kMinBlockSize * ((size + kMinBlockSize - 1) / kMinBlockSize);
    }
  }
}

size_t round_size_new(size_t size) {
  using c10::cuda::CUDACachingAllocator::Native::kMinBlockSize;
  if (size < kMinBlockSize) {
    return kMinBlockSize;
  } else {
    auto divisions = CachingAllocator::AcceleratorAllocatorConfig::roundup_power2_divisions(size);
    if (divisions > 1 && size > (kMinBlockSize * divisions)) {
      return roundup_power2_next_division(size, divisions);
    } else {
      return kMinBlockSize * ((size + kMinBlockSize - 1) / kMinBlockSize);
    }
  }
}
```
I bound them to `_cuda_round_size_old` and `_cuda_round_size_new`, each measuring the time for 100 calls.
```cpp
m.def("_cuda_round_size_old", [](size_t size) {
    size_t result;
    auto start = std::chrono::high_resolution_clock::now();
    for(auto i = 0; i < 100; i++) {
      result = c10::cuda::round_size_old(size + 512*i);
    }
    auto end = std::chrono::high_resolution_clock::now();
    std::cout << "time cost = " << std::chrono::duration_cast<std::chrono::nanoseconds>(end - start).count() << " ns" << std::endl;
    return result;
  });

  m.def("_cuda_round_size_new", [](size_t size) {
    size_t result;
    auto start = std::chrono::high_resolution_clock::now();
    for(auto i = 0; i < 100; i++) {
      result = c10::cuda::round_size_new(size + 512*i);
    }
    auto end = std::chrono::high_resolution_clock::now();
    std::cout << "time cost = " << std::chrono::duration_cast<std::chrono::nanoseconds>(end - start).count() << " ns" << std::endl;
    return result;
  });
```
With `PYTORCH_CUDA_ALLOC_CONF="roundup_power2_divisions:[1024:4,2048:1, >:1]"`, the results running on the following test script were:
```python
import torch

for _ in range(10):
    # only run one function per test.
    torch._C._cuda_round_size_old(1000)
    # torch._C._cuda_round_size_new(1000)
```
`_cuda_round_size_old` first call `~9800 ns`, subsequent calls `~820 ns` (100 times):
```bash
time cost = 9800 ns
time cost = 910 ns
time cost = 820 ns
time cost = 830 ns
time cost = 820 ns
time cost = 820 ns
time cost = 830 ns
time cost = 830 ns
time cost = 820 ns
time cost = 820 ns
```
`_cuda_round_size_new` first call `~42470 ns`, subsequent calls `~770 ns` (100 times):
```bash
time cost = 42470 ns
time cost = 840 ns
time cost = 770 ns
time cost = 770 ns
time cost = 780 ns
time cost = 770 ns
time cost = 780 ns
time cost = 780 ns
time cost = 770 ns
time cost = 780 ns
```
So the steady-state performance is `~8 ns` per call for both, but the first invocation of `round_size_new` shows ~4x higher overhead. This is because `round_size_new` would like to initialize a static single `AcceleratorAllocatorConfig` and `CUDAAllocatorConfig`, however, `round_size_old` would only initialize a static single `CUDAAllocatorConfig`. Another reason is `AcceleratorAllocatorConfig` needs to initialize by parsing both `PYTORCH_ALLOC_CONF` and `PYTORCH_CUDA_ALLOC_CONF`, while `CUDAAllocatorConfig` only parses `PYTORCH_CUDA_ALLOC_CONF`. If I instead set `PYTORCH_ALLOC_CONF="roundup_power2_divisions:[1024:4,2048:1, >:1]"`, the first-call overhead of `round_size_new` drops to ~23030 ns (roughly 2×). It sounds reasonable because `round_size_new` needs to initialize two static single instances than one static single instance initialized by `round_size_old`.
```bash
time cost = 23030 ns
time cost = 860 ns
time cost = 770 ns
time cost = 780 ns
time cost = 770 ns
time cost = 780 ns
time cost = 780 ns
time cost = 780 ns
time cost = 780 ns
time cost = 770 ns
```
Given this, I believe the observed regression in metrics `thrift.process_time.p99.60` on `sl_m897405705_slatest_amd_mi300xg1` is primarily due to the first-call initialization cost. Since subsequent calls take only ~8 ns, the steady-state performance should not be impacted. Based on this analysis, and considering @joshuuuasu's [precondition](https://github.com/pytorch/pytorch/pull/160666#issuecomment-3260221878), I suspect it is not reasonable to test metrics `thrift.process_time.p99.60` on `sl_m897405705_slatest_amd_mi300xg1`? 


guangyey:
Hi @atalman, following Aaron's suggestion in https://github.com/pytorch/pytorch/pull/161786#discussion_r2310714584, I updated the code to raise `ValueError` instead of `RuntimeError`. In addition, I added a deprecated warning message for `torch.cuda.memory._set_allocator_settings`, so it may emit a warning instead of the expected `Error`.
Could you help adjust the internal test case accordingly, as shown in [this change](https://github.com/pytorch/pytorch/pull/162066/files#diff-893b1eea27352f336f4cd832919e48d721e4e90186e63400b8596db6b82e7450R4548-R4577). This should help ensure the internal tests pass.

guangyey:
@joshuuuasu I replied and commented [here](https://github.com/pytorch/pytorch/pull/160666#issuecomment-3270102469)

guangyey:
The only code difference is `CUDAAllocatorConfig` uses the following code to calculate `index`, see https://github.com/pytorch/pytorch/blob/f68de58c9dc635f4a89c1f1f322fe83e6fc96a6a/c10/cuda/CUDAAllocatorConfig.cpp#L47-L49
However, `AcceleratorAllocatorConfig` uses the following code to calculate `index`
https://github.com/pytorch/pytorch/blob/2c5a3d7e6085b08abb0e0aa685933dd0875180f7/c10/core/AllocatorConfig.cpp#L52-L55
Could this difference be the cause of the performance drop on AMD machine?
@joshuuuasu Could you help have a try?

PR #162072, opened 2025-09-03
Relax fences for intrusive ptr's refcnt

mcfi:
Summary: Relax fences for intrusive ptr's refcnt dec op for performance testing.

lock needs acquire when the op succeeds and relaxed if the op is not. In addition, the expire call and the following refcnt reads were merged to remove one extra read.

incref does not need any fences because the caller should already have a valid reference. use_count follows the same reasoning.

decref only needs a release fence to make sure every write op prior to it has finished. When the refcnt goes to zero, there should be a acquire fence to make sure no read op reads stale data before the object is destructed. However, microbenchmark showed that the optimal fence for decref is not performing noticeably better than the current decref with acq-rel, so we keep decref as-is.

This change should have no material impact on x86, but for Arm64 (and other CPUs with weak memory models), it should boost performance.

### DIFF
--- a/c10/util/intrusive_ptr.h
+++ b/c10/util/intrusive_ptr.h
@@ -196,20 +196,25 @@ TTarget* assign_ptr_(TTarget* rhs) {
   }
 }
 
-// Increment needs to be acquire-release to make use_count() and
-// unique() reliable.
+// The only requirement for refcount increment is that it happens-before
+// decrement, so no additional memory ordering is needed.
 inline uint32_t atomic_refcount_increment(std::atomic<uint32_t>& refcount) {
-  return refcount.fetch_add(1, std::memory_order_acq_rel) + 1;
+  return refcount.fetch_add(1, std::memory_order_relaxed) + 1;
 }
 
-// weak_use_count() is only used for testing, so we don't need it to
-// be reliable. Relaxed should be fine.
 inline uint32_t atomic_weakcount_increment(std::atomic<uint32_t>& weakcount) {
   return weakcount.fetch_add(1, std::memory_order_relaxed) + 1;
 }
 
-// Both decrements need to be acquire-release for correctness. See
-// e.g. std::shared_ptr implementation.
+// The requirement is that all modifications to the managed object happen-before
+// invocation of the managed object destructor, and that allocation of the
+// managed object storage happens-before deallocation of the storage.
+//
+// To get this ordering, all non-final decrements must synchronize-with the
+// final decrement. So all non-final decrements have to store-release while the
+// final decrement has to load-acquire, either directly or with the help of
+// fences. But it's easiest just to have all decrements be acq-rel. And it turns
+// out, on modern architectures and chips, it's also fastest.
 inline uint32_t atomic_refcount_decrement(std::atomic<uint32_t>& refcount) {
   return refcount.fetch_sub(1, std::memory_order_acq_rel) - 1;
 }
@@ -332,7 +337,7 @@ class intrusive_ptr final {
   intrusive_ptr() noexcept
       : intrusive_ptr(NullType::singleton(), raw::DontIncreaseRefcount{}) {}
 
-  intrusive_ptr(std::nullptr_t) noexcept
+  /* implicit */ intrusive_ptr(std::nullptr_t) noexcept
       : intrusive_ptr(NullType::singleton(), raw::DontIncreaseRefcount{}) {}
 
   // This constructor will not increase the ref counter for you.
@@ -445,14 +450,14 @@ class intrusive_ptr final {
     if (target_ == NullType::singleton()) {
       return 0;
     }
-    return target_->refcount_.load(std::memory_order_acquire);
+    return target_->refcount_.load(std::memory_order_relaxed);
   }
 
   uint32_t weak_use_count() const noexcept {
     if (target_ == NullType::singleton()) {
       return 0;
     }
-    return target_->weakcount_.load(std::memory_order_acquire);
+    return target_->weakcount_.load(std::memory_order_relaxed);
   }
 
   bool unique() const noexcept {
@@ -851,14 +856,14 @@ class weak_intrusive_ptr final {
       return 0;
     }
     return target_->refcount_.load(
-        std::memory_order_acquire); // refcount, not weakcount!
+        std::memory_order_relaxed); // refcount, not weakcount!
   }
 
   uint32_t weak_use_count() const noexcept {
     if (target_ == NullType::singleton()) {
       return 0;
     }
-    return target_->weakcount_.load(std::memory_order_acquire);
+    return target_->weakcount_.load(std::memory_order_relaxed);
   }
 
   bool expired() const noexcept {
@@ -866,18 +871,22 @@ class weak_intrusive_ptr final {
   }
 
   intrusive_ptr<TTarget, NullType> lock() const noexcept {
-    if (expired()) {
+    if (target_ == NullType::singleton()) {
       return intrusive_ptr<TTarget, NullType>();
     } else {
-      auto refcount = target_->refcount_.load(std::memory_order_seq_cst);
+      auto refcount = target_->refcount_.load(std::memory_order_relaxed);
       do {
         if (refcount == 0) {
           // Object already destructed, no strong references left anymore.
           // Return nullptr.
           return intrusive_ptr<TTarget, NullType>();
         }
-      } while (
-          !target_->refcount_.compare_exchange_weak(refcount, refcount + 1));
+      } while (!target_->refcount_.compare_exchange_weak(
+          refcount,
+          refcount + 1,
+          std::memory_order_acquire,
+          std::memory_order_relaxed));
+
       return intrusive_ptr<TTarget, NullType>(
           target_, raw::DontIncreaseRefcount{});
     }


c10/util/intrusive_ptr.h
```diff
@@ -208,14 +206,13 @@ inline uint32_t atomic_weakcount_increment(std::atomic<uint32_t>& weakcount) {
   return weakcount.fetch_add(1, std::memory_order_relaxed) + 1;
 }
 
-// Both decrements need to be acquire-release for correctness. See
-// e.g. std::shared_ptr implementation.
+// Both decrements need to be release for correctness.
 inline uint32_t atomic_refcount_decrement(std::atomic<uint32_t>& refcount) {
-  return refcount.fetch_sub(1, std::memory_order_acq_rel) - 1;
+  return refcount.fetch_sub(1, std::memory_order_release) - 1;
```
swolchok: why is libc++'s shared_ptr is using acquire/release for this if release is OK? https://github.com/llvm/llvm-project/blob/2729284db1b8c79f14190070c2487a3d9b0687b6/libcxx/include/__memory/shared_count.h#L65

c10/util/intrusive_ptr.h
```diff
@@ -196,10 +196,8 @@ TTarget* assign_ptr_(TTarget* rhs) {
   }
 }
 
-// Increment needs to be acquire-release to make use_count() and
-// unique() reliable.
 inline uint32_t atomic_refcount_increment(std::atomic<uint32_t>& refcount) {
-  return refcount.fetch_add(1, std::memory_order_acq_rel) + 1;
+  return refcount.fetch_add(1, std::memory_order_relaxed) + 1;
```
swolchok, line 202: I think it's probably OK to sacrifice use_count() and unique() reliability because the behavior matches std::shared_ptr's typical behavior. I messed with this once already in https://github.com/pytorch/pytorch/pull/48221/ and I went with the reliable use_count/unique because the previous behavior was seq_cst.

However, we should probably get a second opinion from one or more of @ezyang / @malfet / @albanD.

c10/util/intrusive_ptr.h
```diff
@@ -208,14 +206,13 @@ inline uint32_t atomic_weakcount_increment(std::atomic<uint32_t>& weakcount) {
...
-// e.g. std::shared_ptr implementation.
+// Both decrements need to be release for correctness.
 inline uint32_t atomic_refcount_decrement(std::atomic<uint32_t>& refcount) {
-  return refcount.fetch_sub(1, std::memory_order_acq_rel) - 1;
+  return refcount.fetch_sub(1, std::memory_order_release) - 1;
 }
 
 inline uint32_t atomic_weakcount_decrement(std::atomic<uint32_t>& weakcount) {
-  return weakcount.fetch_sub(1, std::memory_order_acq_rel) - 1;
+  return weakcount.fetch_sub(1, std::memory_order_release) - 1;
```
swolchok: ditto

c10/util/intrusive_ptr.h
```diff
@@ -283,15 +280,21 @@ class intrusive_ptr final {
       // See comment above about weakcount. As long as refcount>0,
       // weakcount is one larger than the actual number of weak references.
       // So we need to decrement it here.
+      // The loads needs to be seq_cst to prevent it from being reordered
+      // before the prior refcount decrement. This essentially makes it
+      // an acquire fence.
       bool should_delete =
-          target_->weakcount_.load(std::memory_order_acquire) == 1;
+          target_->weakcount_.load(std::memory_order_seq_cst) == 1;
```
swolchok: the old acquire behavior seems to match libc++: https://github.com/llvm/llvm-project/blob/fee69dde12643c616209b0ec667280024a130a89/libcxx/src/memory.cpp#L81

I am hesitant to pioneer new reference counting smart pointer atomicity semantics.

c10/util/intrusive_ptr.h
```diff
@@ -283,15 +280,21 @@ class intrusive_ptr final {
...
       if (!should_delete) {
         // justification for const_cast: release_resources is basically a
         // destructor and a destructor always mutates the object, even for const
         // objects. NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
         const_cast<std::remove_const_t<TTarget>*>(target_)->release_resources();
         should_delete =
             detail::atomic_weakcount_decrement(target_->weakcount_) == 0;
+        if (should_delete) {
+          std::atomic_thread_fence(std::memory_order_acquire);
+        }
```
swolchok: haven't thought about this carefully but I'm given to understand that standalone fences are usually wrong, and would need to think carefully to deviate from typical behavior

c10/util/intrusive_ptr.h
```diff
@@ -445,14 +448,14 @@ class intrusive_ptr final {
     if (target_ == NullType::singleton()) {
       return 0;
     }
-    return target_->refcount_.load(std::memory_order_acquire);
+    return target_->refcount_.load(std::memory_order_relaxed);
```
swolchok, line 453: this one is fine per above.

I'm stopping detailed feedback here; my overall request is to provide at least one detailed reference to a battle-tested reference-counting smart pointer implementation when changing the fences for each change. I'm satisfied for use_count and unique() in particular because I know those were more aggressive than standard C++.

swolchok (changes requested):

ezyang:
Scott I will happily defer to you on this

c10/util/intrusive_ptr.h
mcfi: the decref op can be:
1. acq_rel, or
2. rel in cases which the refcnt is not 0, and rel followed by an acquire fence if refcnt is 0.
In both cases, when the refcnt hits 0, none of the resource release is reordered before the last writer to the object, which is needed for correctness..
My change implements the second (except that I used a seq_cst read following a release, which is used as an acquire fence), which should be less restrictive than libcpp's because the common case is when refcnt does not hit zero.
That said, perhaps we should change libcpp as well :)

c10/util/intrusive_ptr.h
mcfi: Please see my comments above.

c10/util/intrusive_ptr.h
mcfi: if the decref op is release, we can't use acquire here because the CPU might reorder the read before the released store. There should be an acquire fence before the read, we can combine the acquire fence with the read op in a seq_cst read, which is exactly what I'm doing here.

c10/util/intrusive_ptr.h
mcfi: This acquire fence is needed for correctness since the decref only has the release semantics. Otherwise, target_ destruction might be reordered before the refcnt hits zero.

c10/util/intrusive_ptr.h
albanD: We do rely on use_count/unique being valid in things like the autograd engine. This can be used over multiple threads when considering worker threads, but I assume races are going to be unlikely. In the worst case there, we will do extra work (cloning instead of re-using the Tensor).

We also leverage unicity for the PyObject resurrection logic. Which is pretty cursed as well. But again, I guess unlikely to race this way.

swolchok: @albanD, what makes race conditions hard to debug is that they are unlikely. At scale, unlikely things happen all the time. if unique() is load-bearing for correctness in the PyObject resurrection logic that makes me less inclined to do this.

WonderfulVoid: Returning a value from `atomic_refcount_increment()` makes it a load. Changing the semantics from load-acquire to load-relaxed would affect any user which checks the return value and then depends on observing any related memory updates. What do we know about usage of this function?
If the return value isn't used, then better to remove it and the weakening from acquire to relaxed shouldn't matter.
Correspondingly, `atomic_refcount_increment()` is also a store with release semantics. We need to know if any user depends on that store-release. It could synchronize-with any other load-acquire from this refcount.
Ideally, the specification of this function should specify what memory ordering semantics it has. Perhaps we need a new function `atomic_refcount_increment_rlx()` which does not have any memory ordering semantics and use that in suitable places.

mcfi: @albanD @swolchok can you show me a concrete example of how unique() needs to be load-bearing for correctness in PyObject resurrection logic? For unique(), if the caller is a strong ptr, it guarantees that use_count() >= 1 and the object is alive; if the caller is a weak ptr, unique() is not guaranteed at all and the caller has to call lock to upgrade to a strong ptr for a more reliable unique call and lock has acquire, which brings us back to the strong ptr case. In either case, I don't see any necessity of having use_count() require acquire.

c10/util/intrusive_ptr.h
mcfi: Actually, my microbenchmark on Arm server chips didn't show much difference between 1 and 2, so I just reverted the changes for decref.

c10/util/intrusive_ptr.h
mcfi: > Returning a value from `atomic_refcount_increment()` makes it a load. Changing the semantics from load-acquire to load-relaxed would affect any user which checks the return value and then depends on observing any related memory updates. What do we know about usage of this function? If the return value isn't used, then better to remove it and the weakening from acquire to relaxed shouldn't matter. Correspondingly, `atomic_refcount_increment()` is also a store with release semantics. We need to know if any user depends on that store-release. It could synchronize-with any other load-acquire from this refcount. Ideally, the specification of this function should specify what memory ordering semantics it has. Perhaps we need a new function `atomic_refcount_increment_rlx()` which does not have any memory ordering semantics and use that in suitable places.

Very good point. If you look at the file, atomic_refcount_increment returns to incref and retain_, neither of which returns the incremented value, so there is no user for the returned value. Same for atomic_weakcount_increment.

yfeldblum (approved):
From feedback left internally.
* Primary benefit of this PR is from relaxing the memory-order in increment, in addition to the memory-orders in use-count and lock.
* But splitting out a fence in decrement is likely a pessimization on modern hardware architectures. This part is reverted.

c10/util/intrusive_ptr.h
mcfi: This requested change is moot because I reverted my change at that line.

swolchok (approved):
current set of changes appears to be only "stop trying to make use_count/unique reliable", which is fine enough. I'm nervous about the autograd stuff but I've never seen it before and @albanD was OK with it so I'm not going to block here.

PR #162080, opened 2025-09-03
ci: Disable all jobs that require mi250

seemethere:
* __->__ #162080

We don't have nodes for these right now and this is causing us to queue
forever. If this is still the case after a week then I'm deleting all of
these.


jeffdaily:
@seemethere we need to switch the binary smoketests from mi250 to mi300.  See https://github.com/pytorch/pytorch/pull/162044/files.

huydhn (approved):

jithunnair-amd:
@seemethere Moving to Draft until you have acknowledged [Jeff's comment](https://github.com/pytorch/pytorch/pull/162080#issuecomment-3250318678)

PR #162084, opened 2025-09-03
use sym_or instead of any to avoid dde in calc_conv_nd_return_shape

laithsakka:
* #162109
* #162099
* __->__ #162084

### DIFF
--- a/torch/_meta_registrations.py
+++ b/torch/_meta_registrations.py
@@ -2373,9 +2373,10 @@ def calc_conv_nd_return_shape(
             ret_shape.append(
                 _formula(dims[i], padding[i], dilation[i], kernel_size[i], stride[i])
             )
+    from torch.fx.experimental.symbolic_shapes import sym_or
 
     torch._check(
-        any(x > 0 for x in ret_shape[2:]),
+        sym_or(*[x > 0 for x in ret_shape[2:]]),
         lambda: f"Given input size per channel: {list(dims)}. "
         f"Calculated output size per channel: {ret_shape[2:]}. "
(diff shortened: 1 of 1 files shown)


torch/_meta_registrations.py
```diff
@@ -2373,9 +2373,10 @@ def _formula_transposed(ln: int, p: int, d: int, k: int, s: int, op: int) -> int
             ret_shape.append(
                 _formula(dims[i], padding[i], dilation[i], kernel_size[i], stride[i])
             )
+    from torch.fx.experimental.symbolic_shapes import sym_or
 
     torch._check(
-        any(x > 0 for x in ret_shape[2:]),
+        sym_or(*[x > 0 for x in ret_shape[2:]]),
```
aorenste, line 2379: nit:
```suggestion
        sym_or(*(x > 0 for x in ret_shape[2:])),
```
The original way (with square brackets) creates an intermediate list which is then converted into a tuple for the varargs. This way just creates the tuple directly.

aorenste: How does this avoid the DDE? Does `torch._check` not evaluate syms as bool or something weird? Or is it that you won't get a DDE as long as ONE of the ret_shapes are definitely > 0 so the others don't get evaluated?

laithsakka: sym_or creates a single expressions, torch._check have special handling for unhinted input expressions it convert them to runtime assertions
but when we use any each expressions is evaluated separately outside the torch._check.

aorenste (approved):

PR #162086, opened 2025-09-03
[CI][ROCM] Disable rocm build tests

jeanschmidt:
There is a queue time of 23 hours for `linux.rocm.gpu.mi250` and 12h for `linux.rocm.gpu.gfx942.1`

With https://github.com/pytorch/pytorch/pull/162044 more jobs will be running on `linux.rocm.gpu.gfx942.1` and this fleet is clearly not capable of handling the load.

So the goal here is to remove smoke tests for builds that uses ROCM in order to reduce the pressure in the fleet and reduce the lag between viable/strict.

Once the fleet is increased we can revert this PR

### DIFF
--- a/.github/templates/linux_binary_build_workflow.yml.j2
+++ b/.github/templates/linux_binary_build_workflow.yml.j2
@@ -96,7 +96,7 @@ jobs:
 
   {%- if config["gpu_arch_type"] != "cuda-aarch64" %}
   !{{ config["build_name"] }}-test:  # Testing
-    if: ${{ github.repository_owner == 'pytorch' }}
+    {% if config["gpu_arch_type"] == "rocm" %}if: false{% else %}if: ${{ github.repository_owner == 'pytorch' }}{% endif %}
     needs:
       - !{{ config["build_name"] }}-build
       - get-label-type
--- a/.github/workflows/generated-linux-binary-libtorch-nightly.yml
+++ b/.github/workflows/generated-linux-binary-libtorch-nightly.yml
@@ -338,7 +338,7 @@ jobs:
     secrets:
       github-token: ${{ secrets.GITHUB_TOKEN }}
   libtorch-rocm6_3-shared-with-deps-release-test:  # Testing
-    if: ${{ github.repository_owner == 'pytorch' }}
+    if: false
     needs:
       - libtorch-rocm6_3-shared-with-deps-release-build
       - get-label-type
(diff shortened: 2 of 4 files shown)


seemethere (commented):
Looks mostly good to me. Would want Jeff's opinion here as well. 

Another thing I have a question about is if this affects our ability to upload binaries. Might be good to check that the test isn't still a dependency for upload.

jeanschmidt:
Ohh, if upload is done by the tests step, that would be a deal breaker, in this case, maybe we need to think in a more elaborate approach.

But I still land firmly on the side of reducing the dependency on ROCm runners for now for critical flows like builds and nightly

jithunnair-amd:
Can we observe for the rest of today how my changes in https://github.com/pytorch/pytorch/pull/162044 impact the TTS for binary smoke test jobs on ROCm, and if it seems to not be sufficient, we could disable only `.github/workflows/generated-linux-binary-manywheel-rocm-main.yml`? Also, what exactly defines which workflows get considered as gating for `viable/strict`? The intention with `.github/workflows/generated-linux-binary-manywheel-rocm-main.yml` was to provide us a valuable granular signal for ROCm binary builds breaking, but it doesn't have to block `viable/strict`.

jeanschmidt:
Hi @jithunnair-amd, sure,

All jobs that are tests on build, block the update of viable/strict branch. This is a VERY critical workflow in PyTorch. Let me explain:

developers are supposed to base their PRs from the viable/strict branch, that contain only the last 'green' status on main for the critical signals (lint, trunk, pull, build). If viable/strict is lagging behind main branch, people can't trust this branch and are forced to develop in top of main otherwise the risk of merge conflicts grows too much. But then they can't trust their signals that well.

Another critical path for this are nightly builds, that are on top of viable/strict. So, if there are too many hours of delay between a commit in main and the time it takes to get its signal, this delays nightly builds and everyone who uses those builds gets a older version of the code.

This is why, 3h queue time is not acceptable for this use case. And we should strive to both have a consistently zero queue time due to spare capacity (in case of an outage bring some of those instances offline).

I am open to find compromises, we have a parallel discussion into other approach to ignore tests status to update viable/strict and accept the risk of having broken nightly builds for ROCm. But this is less ideal.

Still, I am closing this PR as this solution is not correct, given that the upload depends on the test, so, this PR is actually going to break more than fix.

PR #162090, opened 2025-09-03
[aot-precompile] default-filter global guards

dolpm:
if the user doesn't provide their own guard filter fn, we should by default filter global guards.

pytest test/dynamo/test_aot_compile.py


zhxchen17 (approved):

zhxchen17:
@dolpm seems a rebase is needed

PR #162091, opened 2025-09-03
Fix unclear CUBLAS error in torch._scaled_mm with FP8 inputs missing out_dtype

### LINKED ISSUE #160816: torch._scaled_mm crash with cublas error when out_dtype is not specified
## Summary
- `torch._scaled_mm` support for DeepSeekV3 style FP8 blockwise GEMMs was recently added in https://github.com/pytorch/pytorch/pull/158037
- I found that if `out_dtype` keyword arg is not explicitly specified, we get an inscrutable cublas error (see below).
- Explicitly specifying the out_dtype arg resolves the issue, but we should have either (1) a sensible default like fp32, or (2) a more helpful user-facing error message.

```python

Error:
  File "/home/danvm/ao/benchmarks/blockwise_fp8_training/bench_gemms.py", line 116, in run_experiment
    warmup(
    ~~~~~~^
        torch._scaled_mm,
        ^^^^^^^^^^^^^^^^^
    ...<3 lines>...
        1.0 / B_t_s,
        ^^^^^^^^^^^^
    )
    ^
  File "/home/danvm/ao/benchmarks/blockwise_fp8_training/bench_gemms.py", line 82, in warmup
    func(*args, **kwargs)
    ~~~~^^^^^^^^^^^^^^^^^
RuntimeError: CUDA error: CUBLAS_STATUS_INVALID_VALUE when calling `cublasLtMatmulAlgoGetHeuristic( ltHandle, computeDesc.descriptor(), Adesc.descriptor(), Bdesc.descriptor(), Cdesc.descriptor(), Ddesc.descriptor(), preference.descriptor(), 1, &heuristicResult, &returnedResult)`
```

cc @lw 

### Versions

- pytorch main branch build from source with CUDA 12.9 (commit: 8fe4b3f84814e0733b50c3fed5b9f9db22c87c15)

dsashidh:
Fixes #160816 

When using FP8 inputs with torch._scaled_mm (for example DeepSeekV3 style FP8 blockwise GEMMs) without specifying out_dtype, users get an uninformative error: 

RuntimeError: CUDA error: CUBLAS_STATUS_INVALID_VALUE when calling `cublasLtMatmulAlgoGetHeuristic( ltHandle, computeDesc.descriptor(), Adesc.descriptor(), Bdesc.descriptor(), Cdesc.descriptor(), Ddesc.descriptor(), preference.descriptor(), 1, &heuristicResult, &returnedResult)

My change adds validation in aten/src/ATen/native/cuda/Blas.cpp that catches this case early and provides a helpful error message:

torch._scaled_mm: when using FP8 inputs please pass out_dtype (e.g., torch.float16 or torch.float32). Omitting it leads to an unsupported configuration in the CUDA backend.

I tested this by creating FP8 tensors with torch.float8_e4m3fn and calling torch._scaled_mm without the out_dtype parameter. The new error message now appears instead of the CUBLAS error.

### DIFF
--- a/aten/src/ATen/native/cuda/Blas.cpp
+++ b/aten/src/ATen/native/cuda/Blas.cpp
@@ -1246,6 +1246,14 @@ _scaled_mm_out_cuda(const Tensor& mat1, const Tensor& mat2,
   TORCH_CHECK(
       mat1.sizes()[1] == mat2.sizes()[0], "mat1 and mat2 shapes cannot be multiplied (",
       mat1.sizes()[0], "x", mat1.sizes()[1], " and ", mat2.sizes()[0], "x", mat2.sizes()[1], ")");
+  
+  // DeepSeek blockwise scaling types
+  std::unordered_set<ScalingType> deepseek_recipes = {
+      ScalingType::BlockWise128x128,
+      ScalingType::BlockWise1x128,
+      ScalingType::BlockWise1x32,
+      ScalingType::BlockWise1x16
+  };
 
   // Check what type of scaling we are doing based on inputs. This list is sorted
   // by decreasing priority. We prefer "simpler" schemes as they are supported
@@ -1262,7 +1270,10 @@ _scaled_mm_out_cuda(const Tensor& mat1, const Tensor& mat2,
       std::make_pair(ScalingType::BlockWise1x16, ScalingType::BlockWise1x16)
     },
     mat1, mat2, scale_a, scale_b);
-
+  if (deepseek_recipes.count(scaling_choice_a) || deepseek_recipes.count(scaling_choice_b)) {
+    TORCH_CHECK(out_dtype.has_value(),
+                "When using DeepSeek recipes (e.g., BlockWise128x128, BlockWise1x128, BlockWise1x32, BlockWise1x16), you must specify the out_dtype (e.g., torch.float16 or torch.float32).");
+  }
   TORCH_CHECK(!scale_result || (scale_result->numel() == 1 && scale_result->scalar_type() == kFloat),
        "scale_result must be a float scalar");
   TORCH_CHECK(!bias || bias->numel() == mat2.sizes()[1], "Bias must be size ", mat2.sizes()[1],
--- a/test/test_matmul_cuda.py
+++ b/test/test_matmul_cuda.py
@@ -1327,6 +1327,29 @@ class TestFP8Matmul(TestCase):
         torch.testing.assert_close(out_scaled_mm, out_emulated, atol=atol, rtol=rtol)
 
     @onlyCUDA
+    @unittest.skipIf(not PLATFORM_SUPPORTS_FP8, f8_msg)
+    @parametrize("base_dtype", [torch.float8_e4m3fn, torch.float8_e5m2])
+    def test_scaled_mm_deepseek_pattern_missing_out_dtype_error(self, device, base_dtype):
+       def ceil_div(a, b):
+           return (a + b - 1) // b
+       m = k = n = 256
+       a = torch.empty((m, k), device=device, dtype=base_dtype)
+       b = torch.empty((k, n), device=device, dtype=base_dtype)
+       scale_a_size = (m, ceil_div(k, 128))
+       scale_a_stride = (1, m)
+       scale_a_storage = torch.empty(scale_a_size[0] * scale_a_size[1], device=device, dtype=torch.float32)
+       scale_a = scale_a_storage.as_strided(scale_a_size, scale_a_stride)
+       scale_b_size = (n, ceil_div(k, 128)) 
+       scale_b_stride = (1, n) 
+       scale_b_storage = torch.empty(scale_b_size[0] * scale_b_size[1], device=device, dtype=torch.float32)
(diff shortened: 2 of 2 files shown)


dsashidh:
> Reading a bit more closely why doesn't ` const auto out_dtype_ = out_dtype.value_or(mat_a.scalar_type());` save us here, is it because it's picking an unsupported output type (FP8)?
> 
> Would it make sense to try and avoid dispatching to cuBLAS for this case instead?

To your first question, yes that's correct. When out_dtype isn't provided, out_dtype.value_or(mat_a.scalar_type()) uses the input tensor's type as the fallback. Since mat_a is FP8, it defaults to FP8 output, which creates an unsupported FP8->FP8 configuration that cuBLAS can't handle.

And to answer your second question, I think requiring explicit out_dtype is the right approach rather than avoiding cuBLAS dispatch. FP8 ->FP8 output would be a poor default due to significant precision loss, and users working with FP8 should be intentional about their output precision. The explicit requirement ensures users deliberately choose a reasonable output type rather than accidentally getting low-precision FP8 results.

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1651,6 +1651,13 @@ _scaled_mm_cuda(const Tensor& mat_a, const Tensor& mat_b,
           const std::optional<at::Tensor>& scale_result,
           std::optional<c10::ScalarType> out_dtype,
           bool use_fast_accum) {
+  TORCH_CHECK(
```
drisspg: Instead of doing it this way can you instead update the deepseek recipe to have this error message, scaled_mm is only useful for float8 types and fp4 types. So this implicitly is making our optional<out_dtype> not really optional.

We can changes this for the public API 

cc @slayton58

drisspg (changes requested):
comment

aten/src/ATen/native/cuda/Blas.cpp
lw: I commented on the issue but I'll repeat it here: do we know of anyone actually wanting a fp8 output from scaled-mm? If so, can we disable this and, in fact, make `out_dtype` or `out` a mandatory argument? I think this is the same thing @drisspg proposed.

slayton58: @lw I haven't found anyone so far - discussions have shown that people would be ok with a fusable conversion after a `_scaled_mm` call, and not require direct low-precision output.

`out_dtype` being a mandatory argument, with default (assume `torch.bfloat16`) seems reasonable to me.

dsashidh: Hi @drisspg I tried to find the DeepSeek recipe you mentioned to move the error handling there, but I'm having trouble locating it. I searched through the codebase for torch._scaled_mm calls that don't specify out_dtype (excluding test files), and all the ones I found in the PyTorch repo already handling it correctly:
grep -r "torch._scaled_mm" --include="*.py" --exclude-dir="test" . | grep -v "out_dtype"
The files that came up:

- third_party/fbgemm/fbgemm_gpu/experimental/gen_ai/bench/quantize_ops.py already specifies out_dtype=torch.bfloat16
- third_party/aiter/gradlib/gradlib/GemmTuner.py already specifies out_dtype=self.outdtype
- third_party/aiter/aiter/tuned_gemm.py already specifies out_dtype=otype
- torch/cuda/tunable.py already specifies out_dtype=dtypeC
- torch/distributed/_symmetric_memory/init.py passes out_dtype as a positional argument

The original error traceback points to /home/danvm/ao/benchmarks/blockwise_fp8_training/bench_gemms.py, which is external user code.
Could you please clarifies which specific file or function you meant by DeepSeek recipe? Or would you prefer I go with the alternative approach of making out_dtype a mandatory parameter with a default value like torch.bfloat16? Thank you for your time

drisspg: https://github.com/pytorch/pytorch/blob/264e7f68a095b5b2f10e7ca66a3bbb16f6e56e28/aten/src/ATen/native/cuda/Blas.cpp#L1254-L1266

we get the joint scaling recpies here so if this falls into one of the deepseek recipes e.g. the 128x ones than lets make sure a user passes in an outdtype

dsashidh: @drisspg Thanks for your feedback! I’ve made the updates. Hope this aligns with you had in mind. Please let me know if I should adjust anything further.

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1246,6 +1246,14 @@ _scaled_mm_out_cuda(const Tensor& mat1, const Tensor& mat2,
   TORCH_CHECK(
       mat1.sizes()[1] == mat2.sizes()[0], "mat1 and mat2 shapes cannot be multiplied (",
       mat1.sizes()[0], "x", mat1.sizes()[1], " and ", mat2.sizes()[0], "x", mat2.sizes()[1], ")");
+  
+  // DeepSeek blockwise scaling types
+  std::unordered_set<ScalingType> deepseek_recipes = {
```
drisspg, line 1251: Did you mean to close this, ? alsoonly the 128 ones are neede this chec, can you just call it hopperx128

PR #162095, opened 2025-09-03
[Torch FX] Run ensures on PassResult instead of the input graph_module.

abeakkas:
Summary: We need to run the ensures on res.graph_module in case a pass creates a new graph_module instead of modifying in place.

Test Plan:
CI

Rollback Plan:

Differential Revision: D81617603


ezyang:
Needs test

PR #162097, opened 2025-09-03
[PGO] handle PGO profile merges

pianpwk:
Avoid merges from extra PGO key, if same source has different rank. Unlikely to happen (needs code hash match & source variable type to change), but being safe.

Differential Revision: D81299840


bobrenjc93 (approved):
please fix lint before landing

PR #162099, opened 2025-09-03
make should_swap more dde friendly

laithsakka:
* #162354
* #162212
* __->__ #162099

unblock customers for common cases with DDE ,until @pianpwk  land the change to should_swap https://github.com/pytorch/pytorch/pull/160473.

### DIFF
--- a/torch/_prims_common/__init__.py
+++ b/torch/_prims_common/__init__.py
@@ -542,7 +542,10 @@ def is_non_overlapping_and_dense(a: Tensor) -> bool:
 def compute_elementwise_output_logical_to_physical_perm(
     *tensors, _skip_checks=False
 ) -> list[int]:
-    from torch.fx.experimental.symbolic_shapes import guard_size_oblivious
+    from torch.fx.experimental.symbolic_shapes import (
+        guard_or_false,
+        guard_size_oblivious,
+    )
 
     if not _skip_checks and len(tensors) == 0:
         msg = "Can't compute elementwise output strides for zero tensors!"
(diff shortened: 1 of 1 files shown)


torch/_prims_common/__init__.py
```diff
@@ -601,10 +604,26 @@ def should_swap(idx_a, idx_b):
...
-            if guard_size_oblivious(stride_a > stride_b):
+            if not guard_or_false(stride_b == 1) and guard_size_oblivious(
+                stride_a < stride_b
+            ):
+                return -1
+
+            # imitates if stride_a > stride_b : return 1 but
+            # when stride_b = 1, we want stride_a > stride_b to be TRUE
+            # when stride_a = 1, we want stride_a > stride_b to be FALSE
+            if guard_or_false(stride_b == 1):
```
aorenste: I think this would be a little clearer to move this above the previous check (`if not guard_or_false...`) - and then you don't need to check `guard_or_false(stride_b == 1)` twice...

torch/_prims_common/__init__.py
```diff
@@ -601,10 +604,26 @@ def should_swap(idx_a, idx_b):
...
+                return 1
+
+            if not guard_or_false(stride_a == 1) and guard_size_oblivious(
```
aorenste: Isn't this `guard_or_false(stride_a == 1)` guaranteed to be False due to line 610?

laithsakka: right I was just doing this independently let me update it.

aorenste (approved):

torch/_prims_common/__init__.py
laithsakka: seems like this change caused issued idk why let me look more into the new failures.

PR #162108, opened 2025-09-03
[MPS] Add `native_dropout` and `native_dropout_backward`

kurtamohler:
* __->__ #162108

Fixes #162002


aten/src/ATen/native/mps/operations/Dropout.mm
```diff
@@ -0,0 +1,52 @@
...
+
+static Tensor native_dropout_mask_and_scale(const Tensor& input, const Tensor& mask, float scale) {
+  auto output = at::empty_like(input);
+  mps::binary_op_kernel("native_dropout_mask_and_scale", input, mask, output, scale);
+  return output;
+}
+
+std::tuple<Tensor, Tensor> native_dropout_mps(const Tensor& input, double p, std::optional<bool> train) {
+  if (input.numel() == 0) {
+    return std::make_tuple(input, at::empty_like(input, input.options()));
```
malfet: Nit (also, shouldn't mask be `bool`?
```suggestion
    return {input.clone(), at::empty_like(input, input.options().dtype(kBool))};
```

aten/src/ATen/native/mps/operations/Dropout.mm
```diff
@@ -0,0 +1,52 @@
...
+namespace at::native {
+
+static Tensor native_dropout_mask_and_scale(const Tensor& input, const Tensor& mask, float scale) {
+  auto output = at::empty_like(input);
+  mps::binary_op_kernel("native_dropout_mask_and_scale", input, mask, output, scale);
+  return output;
+}
+
+std::tuple<Tensor, Tensor> native_dropout_mps(const Tensor& input, double p, std::optional<bool> train) {
+  if (input.numel() == 0) {
```
malfet:
```suggestion
  if (input.numel() == 0 || !train.value_or(false) || p == 0) {
```

aten/src/ATen/native/mps/operations/Dropout.mm
```diff
@@ -0,0 +1,52 @@
...
+    float p_comp = 1.0f - p;
+    mask = at::empty_like(input, input.options().dtype(c10::kBool));
+    mask.bernoulli_(p_comp);
+    auto scale = p_comp == 0 ? 0.0f : 1.0f / p_comp;
+    output = native_dropout_mask_and_scale(input, mask, scale);
+  } else {
+    mask = at::ones_like(input, input.options().dtype(c10::kBool));
+    output = input.clone();
+  }
+  return std::make_tuple(output, mask);
```
malfet:
```suggestion
  return {output, mask};
```

aten/src/ATen/native/mps/operations/Dropout.mm
```diff
@@ -0,0 +1,52 @@
...
+
+std::tuple<Tensor, Tensor> native_dropout_mps(const Tensor& input, double p, std::optional<bool> train) {
+  if (input.numel() == 0) {
+    return std::make_tuple(input, at::empty_like(input, input.options()));
+  }
+
+  Tensor mask;
+  Tensor output;
+
+  if (!train.has_value() || *train) {
```
malfet: Nit
```suggestion
  if (!train.value_or(false)) {
```

aten/src/ATen/native/mps/operations/Dropout.mm
Skylion007:
```suggestion
  return {std::move(output), std::move(mask)};

**Comment by kurtamohler** [comment]
I think the macos-m2-15 failure looks like it's due to an infrastructure issue. The macos-m1-14 one passes

File: aten/src/ATen/native/mps/operations/Dropout.mm [hunk 5]
```diff
@@ -0,0 +1,45 @@
...
+  if (input.numel() == 0 || !train.value_or(false) || p == 0) {
+    return {input.clone(), at::ones_like(input, input.options().dtype(c10::kBool))};
+  }
+
+  float p_comp = 1.0f - p;
+  Tensor mask = at::empty_like(input, input.options().dtype(c10::kBool));
+  mask.bernoulli_(p_comp);
+  auto scale = p_comp == 0 ? 0.0f : 1.0f / p_comp;
+  Tensor output = native_dropout_mask_and_scale(input, mask, scale);
+  return {output, mask};
```
Skylion007: Std::move output and std::move mask here

test/test_mps.py
```diff
@@ -7524,6 +7524,40 @@ def test_bernoulli(self):
             uniq = mps_out.unique()
             self.assertEqual(uniq, torch.arange(2, device='mps', dtype=dtype))
 
+    @parametrize("dtype", [torch.float16, torch.bfloat16, torch.float32])
+    @parametrize("device", ['mps', 'cpu'])
+    def test_dropout(self, dtype, device):
```
malfet: How long does it take to run this test?

malfet (approved):

test/test_mps.py
kurtamohler: About 4 seconds or so to run with all combinations. I think it's safe to reduce the input sizes by a factor of 10 to cut down runtime to less than a second. I just made it arbitrarily large to make it very unlikely for the computed distribution to mismatch the `p` arg. I'll also remove the CPU run.

kurtamohler: With that update it now runs all combinations of the test in about 0.5 seconds

PR #162109, opened 2025-09-03
rewrite __maybe_broadcast should_expand check for unbacked

laithsakka:
* __->__ #162109
* #162099

### DIFF
--- a/torch/_prims_common/__init__.py
+++ b/torch/_prims_common/__init__.py
@@ -107,6 +107,7 @@ CustomOutParamAnnotation = "__custom_out_param__"
 
 
 def same_shape(a: ShapeType, b: ShapeType, *, allow_rhs_unbacked=False) -> bool:
+    from torch.fx.experimental.symbolic_shapes import guard_or_true
     if len(a) != len(b):
         return False
 
@@ -115,7 +116,8 @@ def same_shape(a: ShapeType, b: ShapeType, *, allow_rhs_unbacked=False) -> bool:
             if isinstance(y, torch.SymInt):
                 continue
 
-        if x != y:
+        # if we do not know, then they are not the same.
+        if guard_or_true(x != y):
             return False
 
     return True
--- a/torch/_prims_common/__init__.py
+++ b/torch/_prims_common/__init__.py
@@ -107,24 +107,15 @@ CustomOutParamAnnotation = "__custom_out_param__"
 
 
 def same_shape(a: ShapeType, b: ShapeType, *, allow_rhs_unbacked=False) -> bool:
-    from torch.fx.experimental.symbolic_shapes import guard_size_oblivious
-
     if len(a) != len(b):
         return False
 
     for x, y in zip(a, b):
         if allow_rhs_unbacked:
-            # TODO: We should check that the symbols are consistent
-            # with each other
             if isinstance(y, torch.SymInt):
                 continue
-        # NB: Naively, you would not expect to have to do an oblivious guard
-        # here because there is seemingly no broadcasting here, but in fact we
-        # use this in some situations to determine if we need to do an expand
-        # on the tensor because they don't line up, so you can definitely end
-        # up trying to prove u0 != 1 in this situation.  See
-        # python test/test_proxy_tensor.py -k test_cumsum_unbacked
-        if guard_size_oblivious(x != y):
+
+        if x != y:
             return False
 
     return True
--- a/torch/_prims_common/__init__.py
+++ b/torch/_prims_common/__init__.py
@@ -108,6 +108,7 @@ CustomOutParamAnnotation = "__custom_out_param__"
 
 def same_shape(a: ShapeType, b: ShapeType, *, allow_rhs_unbacked=False) -> bool:
     from torch.fx.experimental.symbolic_shapes import guard_or_true
+
     if len(a) != len(b):
         return False
 
--- a/torch/_refs/__init__.py
+++ b/torch/_refs/__init__.py
@@ -470,7 +470,7 @@ def _maybe_broadcast(*args, preserve_cpu_scalar_tensors=True):
             if guard_or_false(sym_and(x == 1, y == 1)):
                 pass
             elif guard_or_false(sym_or(x == 1, y == 1)):
-                # assume braodcasting.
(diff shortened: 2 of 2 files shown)


torch/_prims_common/__init__.py
```diff
@@ -107,24 +107,15 @@ def __rmul__(self, other: Any) -> typing.Self: ...
...
                 continue
-        # NB: Naively, you would not expect to have to do an oblivious guard
-        # here because there is seemingly no broadcasting here, but in fact we
-        # use this in some situations to determine if we need to do an expand
-        # on the tensor because they don't line up, so you can definitely end
-        # up trying to prove u0 != 1 in this situation.  See
-        # python test/test_proxy_tensor.py -k test_cumsum_unbacked
-        if guard_size_oblivious(x != y):
+
+        if x != y:
```
laithsakka: we could use guard_or_false here also, but (assume they are not the same shape if we cant tell)
but had to separate bcs for broadcast we want to assume they are the same. unlike when we call this from 
a function that is intended to compare tensors.

torch/_refs/__init__.py
```diff
@@ -449,6 +449,20 @@ def _maybe_broadcast(*args, preserve_cpu_scalar_tensors=True):
...
+    def should_expand(a: ShapeType, b: ShapeType) -> bool:
+        from torch.fx.experimental.symbolic_shapes import guard_or_true
+
+        if len(a) != len(b):
+            return True
+
+        for x, y in zip(a, b):
+            if guard_or_true(x != y):
+                return True
+            # assume no braodcasting for unabcked.
```
aorenste: nit: 2x spelling

torch/_refs/__init__.py
```diff
@@ -449,6 +449,20 @@ def _maybe_broadcast(*args, preserve_cpu_scalar_tensors=True):
...
+            torch._check(x == y)
```
aorenste: How could this torch._check() ever be false? I thought guard_or_true() will guarantee it through a guarded value.

torch/_prims_common/__init__.py
ColinPeppler: since we're no longer using this for broadcast check, should we be using `guard_or_true` if we're assuming they're not the same shape?

laithsakka: right. assume they are not the same guard_or_true, if we shall handle shall be used idk if we shall do it here or wait for actual use case.

torch/_refs/__init__.py
laithsakka: if x= u0 and y = u1 then we cant evaluate x!=y. 
in that case we defer this to a runtime assertion using the 
   torch._check(x == y)

The actual result at runtime could be True or False based on concrete values of u0 and u1 . 
make sense?

torch/_refs/__init__.py
laithsakka: why the linter did not catch it lol
i shall not rely on it i guess.

torch/_refs/__init__.py
aorenste: If we can't evaluate x!=y then won't guard_or_true return true and return?

laithsakka: ops i need the check inside lol

laithsakka: ops i deleted your comment by mistake lol it was 
if this is True we would have returned

laithsakka: @aorenste i got something wrong here lol you are right.

torch/_refs/__init__.py
```diff
@@ -458,7 +490,7 @@ def __maybe_broadcast(x, shape):
             if preserve_cpu_scalar_tensors and utils.is_cpu_scalar_tensor(x):
                 return x
 
-            if not utils.same_shape(x.shape, common_shape):
+            if should_expand(x.shape, common_shape):
```
aorenste, line 493: It seems like the old code would expand if it couldn't prove the shapes were the same but the new code assumes that unbacked are equal and thus doesn't expand w/ a runtime check. Is that the right thing to do? Or should we just always expand unbacked?
(The old code was "weird", right? Like "u0 vs u1" would not expand but "s0/constant vs u1" WOULD expand?)
i.e. Should the new code really do something like add "handle_broadcast" to same_shape() and then "if not same_shape(x.shape, common_shape, broadcast=True)" then expand?

torch/_refs/__init__.py
```diff
@@ -449,6 +449,38 @@ def _maybe_broadcast(*args, preserve_cpu_scalar_tensors=True):
...
+                # We know they are not the same.
+                return True
+
+            # They are the same or we do not know if they are the same or not.
+            # 1==1 no-broadcast
+            # u0==1 and 1==u0 cases. We broadcast!
+            if guard_or_false(sym_and(x == 1, y == 1)):
+                pass
+            elif guard_or_false(sym_or(x == 1, y == 1)):
+                # assume braodcasting.
```
aorenste: nit: spelling

aorenste (commented):
Is it possible to add a test for this behavior? (I assume there's tests for existing behavior but maybe something that can tickle the behavior that previously was not handled properly)

torch/_refs/__init__.py
laithsakka: the old code assumes that trying to handle u0 ==1 in case in which we would want to expand) guard size oblivious will say well assume u0>2 hence expand
the new code preserve that handling for u0==1 but also address u0==u1 in this case what shall we do?
previously  guard size oblivious woudl throw DDE, in this new code we will assume u0 and u1 equal.

I am not sure if for this function specifically we would receive that, but this is generally how we handle broadcasting case. (I hope we do that right in all locations now lol) /

aorenste:
> > Is it possible to add a test for this behavior? (I assume there's tests for existing behavior but maybe something that can tickle the behavior that previously was not handled properly)
> 
> I can add a unit test that test the should_expand function but I am not sure how do an e2e test tbh, would that be sufficient?

E2e would be better... I'm less concerned w/ the details of should_expand vs keeping the new behavior from back-sliding.

aorenste (approved):
Approving. Think about adding an e2e unit test - but that can definitely be in a separate PR.

Issue #83069, opened 2022-08-09
__dlpack__() should normalize strides


mattip:
### 🐛 Describe the bug

When converting a tensor to dlpack, the strides of single-value tensors (`t.shape==(1,)`) should be normalized to `(1,)`. This is the fallout conclusion from #82610.

Here is a demo of what is wrong:
```
>>> from torch.utils import dlpack
>>> a = torch.rand(16)
>>> b = a[::3]
>>> c = b[:1]
>>> c.shape
torch.Size([1])
>>> c.stride()
(3,)             # <- shape is (1,), strides is not (1,)
>>> capsule = dlpack.to_dlpack(c)
>>> d = dlpack.from_dlpack(capsule)
>>> d.shape
torch.Size([1])
>>> d.stride()
(3,)             # <- the weird strides is still here. It should be (1,)
```

### Versions

Collecting environment information...
PyTorch version: 1.10.0a0+git2aa19f3
Is debug build: False
CUDA used to build PyTorch: Could not collect
ROCM used to build PyTorch: N/A

OS: Ubuntu 20.04.4 LTS (x86_64)
GCC version: (GCC) 9.4.0
Clang version: 12.0.1 (https://github.com/conda-forge/clangdev-feedstock 83cefdc6bfd937aa14138cf601945c6d9fc9a490)
CMake version: version 3.21.1
Libc version: glibc-2.31

Python version: 3.9.6 | packaged by conda-forge | (default, Jul 11 2021, 03:39:48)  [GCC 9.3.0] (64-bit runtime)
Python platform: Linux-5.4.0-122-generic-x86_64-with-glibc2.31
Is CUDA available: False
CUDA runtime version: 11.2.152
GPU models and configuration: GPU 0: NVIDIA GeForce GTX 1660 SUPER
Nvidia driver version: 470.141.03
cuDNN version: /usr/lib/x86_64-linux-gnu/libcudnn.so.7.6.5
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: False

Versions of relevant libraries:
[pip3] mypy==0.812
[pip3] mypy-extensions==0.4.3
[pip3] numpy==1.21.2
[pip3] torch==1.10.0a0+git2aa19f3
[pip3] torchaudio==0.10.0a0+8a347b6
[conda] magma-cuda112             2.5.2                         1    pytorch
[conda] mkl                       2021.3.0           h726a3e6_557    conda-forge
[...]

PR #162111, opened 2025-09-04
[DLPACK] Optimize toDLPack Conversion Speed

### LINKED ISSUE #162113: ToDLPack Speed Regression
### 🐛 Describe the bug

Previously in https://github.com/pytorch/pytorch/issues/83069, the toDLPack converter introduces a normalization step that changes the strides to 1 when shape[i] == 1

This step, however, calls as_strided during toDLPack, and can slow down the toDLPack about 3x. This causes PyTorch's DLPack conversion to be around 0.6 us overhead per call from the < 0.2us.

Most of the framework handles toDLPack conversion around 0.1 - 0.2 us level, which helps to have good normal eager mode conversion with torch

cc @msaroufim @jerryzh168

tqchen:
Previously in gh-83069, the toDLPack converter introduces a normalization step that changes the strides to 1 when shape[i] == 1

This step, however, calls as_strided during toDLPack, and can slow down the toDLPack about 3x. This causes PyTorch's DLPack conversion to be around 0.6 us overhead per call from the < 0.2us.

This PR updates the logic by adding a need_normalize_strides check, to first confirm if the strides normalization is necessary. In most common cases, when the tensor is continguous, such normalization is not necessary.

We confirmed that having this additional step would recover the speed of toDLPack to below 0.2us and can help significantly speedup eager mode integration of DLPack with PyTorch.

If we detect that there is normalization needs, the older path will be invoked.

Fixes #162113

### DIFF
--- a/aten/src/ATen/DLConvertor.cpp
+++ b/aten/src/ATen/DLConvertor.cpp
@@ -308,17 +308,44 @@ void fillVersion<DLManagedTensorVersioned>(
 // constructed out of ATen tensor
 template <class T>
 T* toDLPackImpl(const Tensor& src) {
-  // create a new tensor with possibly normalized strides
-  // gh-83069
-  auto shape = src.sizes();
-  auto strides = src.strides().vec();
-  for (int i = 0; i < src.dim(); i++) {
-    if (shape[i] < 2) {
-      strides[i] = 1;
+  auto view = src;
+
+  // Detect whether there is need to normalize the strides
+  // Background: gh-83069
+  //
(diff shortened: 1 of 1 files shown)


tqchen:
Benchmark, on AMD Ryzen:

```
torch.utils.dlpack.to_dlpack[old]             6.162643432617187e-07 sec/call. 
to_dlpack[new]                                          1.8970966339111327e-07 sec/call
```

msaroufim (approved):
@pytorchbot merge

PR #162112, opened 2025-09-04
[ROCm] fix miopen batchnorm changing output format

amdfaa:
It was found that the integration of miopen batchnorm was causing the output to always be in default contig memory format even when the input was channels last.  This also unskips a number of related unit tests.


jeffdaily (approved):

test/test_nn.py
```diff
@@ -5198,9 +5213,8 @@ def test_batchnorm(self, dims, mode, memory_format, ref_backend, mixed, dtype):
                 # https://github.com/pytorch/pytorch/issues/156513
                 self.skipTest("bfloat16 NCHW train failed due to native tolerance issue")
 
-            if self._testMethodName == "test_batchnorm_3D_train_NCHW_vs_native_mixed_float16" \
-                    and _get_torch_rocm_version() < (7, 0):
-                self.skipTest("3D float16 NCHW train failed on ROCm<7.0")
```
jithunnair-amd, line 5203: @dnikolaev-amd @jeffdaily Why remove the ROCm7.0 condition? Does this test fail even on ROCm7.0?

dnikolaev-amd: @jithunnair-amd this test failed on ROCm7.0+
```
AssertionError: Tensor-likes are not close!

Mismatched elements: 1 / 8 (12.5%)
Greatest absolute difference: 0.021703720092773438 at index (4,) (up to 1e-05 allowed)
Greatest relative difference: 0.0013534484896808863 at index (4,) (up to 1.3e-06 allowed)
```
Looks like another native BN accuracy issue:
- test_batchnorm_3D_train_NCHW_vs_native_mixed_float16 - failed
- test_batchnorm_3D_train_NCHW_vs_cpu_mixed_float16 - passed

PR #162117, opened 2025-09-04
Making batching rule for F.embedding DTensor-aware

zou3519:
* __->__ #162117

`vmap(F.embedding)(DTensor, DTensor)` was failing because F.embedding's
batching rule generates a new tensor via at::arange, at::arange
generates a regular tensor, and DTensor rightfully errors on mixed
DTensor-regular Tensor operations.

This PR fixes the problem by activating DTensor implicit replication on
just the at::arange and the subsequent add operation.

In order to accomplish this I move the DTensor implicit replication flag
to C++ (most batching rules are in C++).

Test Plan:
- new test


torchgen/gen_vmap_plumbing.py
```diff
@@ -238,10 +242,28 @@ def gen_vmap_plumbing(native_function: NativeFunction) -> str | None:
...
+    # Consider batched_tensor + arange(n). This would work via broadcasting.
+    # However, if we changed this into batched_tensor + batched_tensor.new_arange(n),
+    # then we would actually allocate a repeated tensor for the arange (it
+    # would be a batched tensor), and we'd be using more memory than necessary,
+    # unless we implement some sort of lazy batchedtensor materialization thing.
     return f"""\
 template <typename batch_rule_t, batch_rule_t batch_rule>
 {sig.decl(name=schema.name.unambiguous_name() + "_generated_plumbing")} {{
   c10::impl::ExcludeDispatchKeyGuard guard(DispatchKey::FuncTorchBatched);
+  at::DTensorAllowImplicitReplication guard2;
```
ezyang: Do you then have to disallow implicit replication after you're done doing the vmap rule?

zou3519: Yes. And that does happen (if I'm understanding the question correctly), `DTensorAllowImplicitReplication` is a RAII guard that returns implicit replication to the previous state and that happens after dispatch gets run. Here's some example generated code:

image

zou3519: Oh was the question "should we turn off implicit replication if we pop vmap off the stack"?

torchgen/gen_vmap_plumbing.py
```diff
@@ -238,10 +242,28 @@ def gen_vmap_plumbing(native_function: NativeFunction) -> str | None:
     bdims_all_none_case = gen_case_where_all_bdims_are_none(sig, schema, cur_level_var)
 
     wrapped_returns = gen_returns(returns, cur_level_var, results_var)
+    # NOTE: [vmap and DTensor Implicit Replication]
+    # For all batching rules, we assume that new tensors generated by factory
+    # functions are meant to be replicated by DTensor. This saves us
```
bdhirsh: I'm trying to think through whether this is always safe to do. The example in my head is - let's say we are calling into this code:
```
aten.mm(BatchedTensor(DTensor), BatchedTensor(torch.Tensor)
```

the vmap rule will eventually desugar (I think) into a call to:
```
aten.bmm(DTensor, torch.Tensor)
```

And at this point, we probably want implicit replication to be turned back off? It is expected behavior for that second op to raise an error in eager.

Does that make sense? I could imagine two workarounds:

(1) instead of putting the RAII guard in the vmap rule itself, we could hook into the `FuncTorchDynamicLayerBackMode` and `FuncTorchDynamicLayerFrontMode` keys to turn implicit replication back off underneath each transform

(2) or we could be more targeted about only turning implicit replication on for specific vmap rules that call tensor constructors before redispatching

torchgen/gen_vmap_plumbing.py
zou3519: Brian had the same question with an example and I see the problem now

torchgen/gen_vmap_plumbing.py
zou3519: Okay I see the problem.

(1) is not going to work, we need DTensorAllowImplicitReplication on during DTensor operations

So we're gonna need to do (2)

zou3519: For Option (2), do you think there is anything wrong with changing the batching rule to look like:

(old)
https://github.com/pytorch/pytorch/blob/e532c9d4f1cdcbc1ea9628f55b9813e77847bdc7/aten/src/ATen/functorch/BatchRulesModules.cpp#L45-L50

(new)
```py
  auto indices_ = moveBatchDimToFront(indices, indices_bdim);
  {
    DTensorAllowImplicitReplication guard;
    const auto range = getStepTensor(indices, batch_size, num_embeddings);
    indices_ = indices_ + range;
  }
  auto result = at::embedding_symint(weight_, indices_, std::move(padding_idx), scale_grad_by_freq, sparse);
  return std::make_tuple(std::move(result), 0);
```
Otherwise we are going to need a DTensor-specific new_arange function

zou3519: The main thing that I think can go wrong is if you have BatchedTensor(OtherSubclass(DTensor)) as `indices_` and somehow the OtherSubclass's torch_dispatch is wrong and causes a mixed DTensor-Tensor operation.

I don't really want a DTensor-specific new_arange function because the semantics seem unclear to all Tensor subclasses that override it.

bdhirsh: > The main thing that I think can go wrong is if you have BatchedTensor(OtherSubclass(DTensor)) as indices_ and somehow the OtherSubclass's torch_dispatch is wrong and causes a mixed DTensor-Tensor operation.

Hmm. I guess can interpose between the call to `aten.add(torch.Tensor, OtherSubclass(DTensor))` and do "anything". So the bad outcome is something like... `OtherSubclass` constructs its own temporary tensor that we know should *not* be replicated as part of intercepting `aten.add`, but we have implicit replicate turned on during the entire `aten.add` call.

I agree that there is some risk, although I can't think of any way to do this that isn't extremely synthetic (on top of that - in practice, we have only ever seen cases of `DTensor(InnerSubclass)`, not the other way around).

build_variables.bzl
```diff
@@ -1087,6 +1087,7 @@ aten_cpu_source_non_codegen_list = [
     "aten/src/ATen/DeviceAccelerator.cpp",
     "aten/src/ATen/Context.cpp",
     "aten/src/ATen/DLConvertor.cpp",
+    "aten/src/ATen/DTensorState.cpp",
```
bdhirsh, line 1090: oh god non-globbed ATen file sources

bdhirsh (approved):
sgtm!

PR #162134, opened 2025-09-04
Add Module API shim and nativeRT implementation.

JacobSzwejbka:
Introduce the initial draft of a shared module shim layer. The intention is for various inference engines like ExecuTorch and nativeRT to be able to share this and provide a unified api for inference. Combined with a unified AoT flow. The exact details of which runner are being used where can be hidden and allows the runtime stack more flexibility in using the best runtime for a given model without having to force a change in user code.

### DIFF
--- a/torch/csrc/executorch/shim/module_shim.h
+++ b/torch/csrc/executorch/shim/module_shim.h
@@ -62,7 +62,7 @@ experimental_torch_module_num_outputs(ModuleHandle handle, uint64_t* ret_value);
 
 AOTI_TORCH_EXPORT AOTITorchError experimental_torch_module_forward_flattened(
     ModuleHandle handle,
-    const TypedStableIValue* args,
+    TypedStableIValue* args,
     uint64_t num_args,
     TypedStableIValue* ret_values,
     uint64_t num_outputs);
--- a/torch/csrc/executorch/shim/module_shim.h
+++ b/torch/csrc/executorch/shim/module_shim.h
@@ -1,8 +1,8 @@
 #pragma once
 
-#include <cstdint>
 #include <torch/csrc/inductor/aoti_torch/c/macros.h>
 #include <torch/csrc/inductor/aoti_torch/c/shim.h>
+#include <cstdint>
 
 // This header defines a prototype stable C API for certain Module
 // functionality. It is inspired by:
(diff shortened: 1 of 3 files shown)


torch/csrc/executorch/shim/module_shim.h
```diff
@@ -0,0 +1,72 @@
...
+  StableIValue val;
+  StableIValueTag tag;
+};
+
+struct ModuleOpaque;
+using ModuleHandle = ModuleOpaque*;
+
+AOTI_TORCH_EXPORT AOTITorchError experimental_torch_load_module_from_file(
+    const char* package_path,
+    uint64_t package_path_len,
```
albanD, line 52: Ho we don't have zero terminated strings?

torch/csrc/executorch/shim/module_shim.h
```diff
@@ -0,0 +1,72 @@
...
+experimental_torch_delete_module_object(ModuleHandle handle);
+
+AOTI_TORCH_EXPORT AOTITorchError
+experimental_torch_module_num_outputs(ModuleHandle handle, uint64_t* ret_value);
+
+AOTI_TORCH_EXPORT AOTITorchError experimental_torch_module_forward_flattened(
+    ModuleHandle handle,
+    const TypedStableIValue* args,
+    uint64_t num_args,
+    TypedStableIValue* ret_values,
```
albanD, line 67: Should we use the same stack for in/out? Like we do for other dispatcher pieces?

torch/csrc/executorch/shim/module_shim.h
```diff
@@ -0,0 +1,72 @@
...
+
+AOTI_TORCH_EXPORT AOTITorchError
+experimental_torch_delete_module_object(ModuleHandle handle);
+
+AOTI_TORCH_EXPORT AOTITorchError
+experimental_torch_module_num_outputs(ModuleHandle handle, uint64_t* ret_value);
+
+AOTI_TORCH_EXPORT AOTITorchError experimental_torch_module_forward_flattened(
+    ModuleHandle handle,
+    const TypedStableIValue* args,
```
albanD: It might be insteresting for this not to be const so the function can actually consume the arguments (and free memory faster when possible)

albanD (commented):
Looks pretty cool !

torch/csrc/executorch/shim/module_shim.h
JacobSzwejbka: My impression is adding new apis that rely on zero termination is not great from a security perspective. Even if we are incredibly unlikely to ever encounter a non zero terminated string in practice.

torch/csrc/executorch/shim/module_shim.h
JacobSzwejbka: Ah by consume you mean assume ownership of the lifespan? Yeah thats probably worth doing. There is the slight caveat that input mutation is expressible in export so I think its a bit strange if I pass in 

forward (x, y, z):
  return y.add_(1) + x + z 
  
  and then x and z are invalid after the function call but y is updated

torch/csrc/executorch/shim/module_shim.h
```diff
@@ -0,0 +1,72 @@
...
+extern "C" {
+#endif
+
+/**
+ * The type of a StableIValue.
+ */
+enum class StableIValueTag : uint32_t {
+  None = 0,
+  Tensor = 1,
+  Double = 2,
```
jackzhxng, line 24: no float?

JacobSzwejbka: These are raw inputs. so like 

forward(x : float)

which maps to double in cpp. You could still have float32 Tensor

torch/csrc/executorch/shim/module_shim.h
albanD: Ok! I feel like this always clash with the other principle of "only one source of truth" haha

torch/csrc/executorch/shim/module_shim.h
albanD: The way we do this in the stable layer is that the stack owns everything on it (both on the way in and on the way out).

For your example, it is the caller's responsibility to keep their own reference to x, y and/or z if they need them later.
In this case, I would expect something like `call(move(x), global_y, move(z))`. And that should do "the right thing" by default: as soon as the runner doesn't need x/z, they get de-allocated, while y will stay alive based on the outside reference.

torch/csrc/executorch/shim/module_shim.h
albanD: double and int64 only as raw dtypes haha
Tensor supports all

JacobSzwejbka: Ah yeah the concrete types are definitely double and int64. I guess my tag name here is a little weird because I have int and double. Should probably be Long and Double or Int and Float. Ill double check what we do today with IValue and EValue

JacobSzwejbka: Yeah Ivalue does int = int64 and double for float64

torch/csrc/executorch/shim/module_shim.h
```diff
@@ -0,0 +1,72 @@
...
+/**
+ * A wrapper containing StableIValue which is opaque and its actual type.
+ *
+ * StableIValue is typically used when interacting with the dispatcher.
+ * There the type is known ahead of time as the operators have schemas,
+ * and the value is guaranteed by the caller to be of the correct type.
+ *
+ * Here the value comes from user space. While we could technically
+ * infer the expected type from the module, it is not clear how to
+ * validate the user args which seems like a massive footgun. So for
```
janeyx99, line 38: Question out of curiosity: why is it important to validate the user args? Is it security? Or just for user experience?

torch/nativert/executorch/module_shim_impl.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include <torch/csrc/inductor/aoti_runtime/utils.h>
+#include <torch/csrc/inductor/aoti_torch/utils.h>
+#include <torch/csrc/stable/stableivalue_conversions.h>
+
+#include <torch/nativert/ModelRunner.h>
+
+using namespace ::torch::nativert;
+
+namespace {
+AOTITorchError to_ivalue(const TypedStableIValue& v, c10::IValue* ret_val) {
```
janeyx99, line 22: If these functions aren't exposed to the user and are mostly used internally, I don't think they need to return AOTITorchError + they don't need to be written in a C-error-return way

PR #162136, opened 2025-09-04
[CD] Build Mac wheels using `setup-python` action

malfet:
* __->__ #162136

Biggest difference between both conda and homebrew CPython builds and one from python.org, is that later are universal binaries and they are always trying to build universal extension...

Workaround lots of universal binary build attempts by explicitly specifying both `_PYTHON_PLATFORM` and `--plat-name` as well as `ARCH_FLAGS`

Suppressed actionlint warning on use of `freethreaded` flag which is document in https://github.com/actions/setup-python/tree/v5

TODO: Remove lots of temporary workarounds when `3.14` is out in October 2025


.github/templates/macos_binary_build_workflow.yml.j2
```diff
@@ -22,6 +22,14 @@ name: !{{ build_environment }}
           echo "MAC_PACKAGE_WORK_DIR=${RUNNER_TEMP}" >> "${GITHUB_ENV}"
 {%- endmacro %}
 
+{%- macro setup_python(py_ver) -%}
+      - name: Setup Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "!{{ py_ver.strip('t') }}.!{{0 if py_ver != '3.10' else 4}}!{{ '-rc.2' if '3.14' in py_ver else '' }}"
+          freethreaded: !{{ "true" if py_ver.endswith('t') else "false" }}
```
XuehaiPan, line 32: You will not need this if you provide `python-version: {X}.{Y}t`

.github/templates/macos_binary_build_workflow.yml.j2
```diff
@@ -22,6 +22,14 @@ name: !{{ build_environment }}
           echo "MAC_PACKAGE_WORK_DIR=${RUNNER_TEMP}" >> "${GITHUB_ENV}"
 {%- endmacro %}
 
+{%- macro setup_python(py_ver) -%}
+      - name: Setup Python
+        uses: actions/setup-python@v5
```
XuehaiPan: FYI, setup-python@v6 is out: https://github.com/actions/setup-python/releases/tag/v6.0.0

.github/templates/macos_binary_build_workflow.yml.j2
malfet: @XuehaiPan do you know how to do it for 3.14?

XuehaiPan: For pre-release versions (alpha, beta, rc), you can use: `allow-prereleases: true`.

```yaml
with:
  python-version: '3.12'

with:
  python-version: '3.13'

with:
  python-version: '3.13t'

with:
  python-version: '3.14'
  allow-prereleases: true  # use latest alpha/beta/rc if no release version

with:
  python-version: '3.14t'
  allow-prereleases: true  # use latest alpha/beta/rc if no release version
```

There is no harm in setting `allow-prereleases: true` for release versions (e.g., 3.13 in the example above). My practice is:

```yaml
with:
  python-version: ${{ PYTHON_VERSION }}
  allow-prereleases: true  # use rc version for 3.14 and release verison for pre-3.14
```

where `PYTHON_VERSION` can be: 3.12, 3.13, 3.13t, 3.14, 3.14t, etc.

---

If you want to pin to a specific version:

```yaml
with:
  python-version: '3.13.7'
  allow-prereleases: true  # no-op

with:
  python-version: '3.13.7t'
  allow-prereleases: true  # no-op

with:
  python-version: '3.14.0-rc.2'
  allow-prereleases: true

with:
  python-version: '3.14.0t-rc.2'
  allow-prereleases: true
```

malfet: I've tried it with `allow-prerelease` and it does not seem to work, see https://github.com/malfet/deleteme/blob/baad50f638bb4831232386f6629d673b04465c6a/.github/workflows/install_torch.yml#L25 and https://github.com/malfet/deleteme/actions/runs/17530584425/job/49786975781#step:2:12

XuehaiPan: > see [malfet/deleteme@`baad50f`/.github/workflows/install_torch.yml#L25](https://github.com/malfet/deleteme/blob/baad50f638bb4831232386f6629d673b04465c6a/.github/workflows/install_torch.yml#L25) and [malfet/deleteme/actions/runs/17530584425/job/49786975781#step:2:12](https://github.com/malfet/deleteme/actions/runs/17530584425/job/49786975781#step:2:12)

You need to remove the patch version, because `3.14.0t` has not been released.

```diff
  with:
-   python-version: '3.14.0t'
+   python-version: '3.14t'
    allow-prereleases: true
```

.ci/wheel/build_wheel.sh
```diff
@@ -200,11 +186,11 @@ export BUILD_TEST=OFF
 pushd "$pytorch_rootdir"
 echo "Calling setup.py bdist_wheel at $(date)"
 
-python setup.py bdist_wheel -d "$whl_tmp_dir" --plat-name ${mac_version}
+_PYTHON_HOST_PLATFORM=${mac_version} ARCHFLAGS="-arch arm64" python setup.py bdist_wheel -d "$whl_tmp_dir" --plat-name "${mac_version//[-.]/_}"
 
 echo "Finished setup.py bdist_wheel at $(date)"
 
-if [[ $package_type != 'libtorch' ]]; then
+if [[ $package_type == 'libooootorch' ]]; then
```
huydhn: lol, is this a typo?

.ci/wheel/build_wheel.sh
```diff
@@ -176,17 +167,12 @@ case $desired_python in
...
-conda create ${EXTRA_CONDA_INSTALL_FLAGS} -yn "$tmp_env_name" python="$desired_python" ${CONDA_ENV_CREATE_FLAGS}
-source activate "$tmp_env_name"
-
 PINNED_PACKAGES=(
     "numpy${NUMPY_PINNED_VERSION}"
 )
-retry pip install "${PINNED_PACKAGES[@]}" -r "${pytorch_rootdir}/requirements-build.txt"
-pip install requests ninja typing-extensions
-retry pip install -r "${pytorch_rootdir}/requirements.txt" || true
+python -mvenv ~/${desired_python}-build
```
huydhn, line 175: Just want to confirm that this is expected because the old code installs both `requirements-build.txt`, `requirements.txt`, and a couple of other packages, while the new code installs only `requirements.txt`

XuehaiPan: `requirements-build.txt` is part of `requirements.txt`.

https://github.com/pytorch/pytorch/blob/f216d64bfed2852dd1e9b7f75bcff5257f75ff8b/requirements.txt#L1-L4

atalman (approved):
Looks good, we only need to skip torch.compile test when doing 3.13t

.ci/wheel/build_wheel.sh
malfet: That was a way to make sure `delocate` is skipped

tools/linter/adapters/actionlint_linter.py
```diff
@@ -73,6 +73,8 @@ def check_file(
                 binary,
                 "-ignore",
                 '"runs-on" section must be sequence node but got mapping node with "!!map" tag',
+                "-ignore",
+                'input "freethreaded" is not defined in action "actions/setup-python@v',
```
XuehaiPan, line 77: Is this intentional?

malfet: Yes, actionlint does not seem to be maintained anymore, so they are not adding those options

huydhn (approved):

PR #162142, opened 2025-09-04
[SymmMEM] Allow to import _SymmetricMemory when NVSHMEM is not available

fegin:
* __->__ #162142

Summary:
As we have multiple backends, _SymmetricMemory should not be imported together with NVSHMEM related modules


dcci (approved):

kwen2501 (approved):

torch/distributed/_distributed_c10d.py
```diff
@@ -85,17 +85,23 @@
     from torch.distributed._C_stubs import HashStore
 
 # NVSHMEM/SymmetricMemory components
+
+# There are multiple backends for SymmetricMemory, as a result,
+# _SymmetricMemory should not be imported together with NVSHMEM related modules.
+try:
+    from torch._C._distributed_c10d import _SymmetricMemory
+except ImportError:
+    from torch.distributed._C_stubs import _SymmetricMemory
```
ngimel: should we always do this, to uncouple import from nvshmem even if it's available, and import nvshmem explicitly when needed?

Skylion007: Let's not forget nvshmem4py that exists now... sigh...

fegin:
Will land this PR first to fix the bug as the current code breaks async TP if people do not have NVSHMEM.

Issue #150852, opened 2025-04-08
NCCL init hits CUDA failure 'invalid argument' on 12.2 driver


kwen2501:
### 🐛 Describe the bug

Error seen with nightly build, e.g. torch==2.8.0.dev20250327+cu126
```
[2025-04-08 08:39:46] devgpu263:589012:591652 [0] transport/nvls.cc:254 NCCL WARN Cuda failure 1 'invalid argument'
devgpu263:589012:591652 [0] NCCL INFO transport/nvls.cc:409 -> 1
devgpu263:589012:591652 [0] NCCL INFO init.cc:1141 -> 1
devgpu263:589012:591652 [0] NCCL INFO init.cc:1409 -> 1
devgpu263:589012:591652 [0] NCCL INFO group.cc:75 -> 1 [Async thread]
devgpu263:589012:589012 [0] NCCL INFO group.cc:422 -> 1
devgpu263:589012:589012 [0] NCCL INFO group.cc:581 -> 1
devgpu263:589012:589012 [0] NCCL INFO init.cc:1836 -> 1
[rank0]: Traceback (most recent call last):
[rank0]:   File "/home/kw2501/local/driver_12.2/repro.py", line 12, in <module>
[rank0]:     dist.all_reduce(x)
[rank0]:   File "/home/kw2501/.conda/envs/titan/lib/python3.10/site-packages/torch/distributed/c10d_logger.py", line 81, in wrapper
[rank0]:     return func(*args, **kwargs)
[rank0]:   File "/home/kw2501/.conda/envs/titan/lib/python3.10/site-packages/torch/distributed/distributed_c10d.py", line 2868, in all_reduce
[rank0]:     work = group.allreduce([tensor], opts)
[rank0]: torch.distributed.DistBackendError: NCCL error in: /pytorch/torch/csrc/distributed/c10d/NCCLUtils.cpp:77, unhandled cuda error (run with NCCL_DEBUG=INFO for details), NCCL version 2.26.2
[rank0]: ncclUnhandledCudaError: Call to CUDA function failed.
[rank0]: Last error:
[rank0]: Cuda failure 1 'invalid argument'
```

Mini repro:
```
import os
import torch
import torch.distributed as dist


if __name__ == "__main__":
    rank = int(os.getenv("RANK"))
    world_size = int(os.getenv("WORLD_SIZE"))
    device = torch.device("cuda", rank)
    dist.init_process_group("nccl", rank=rank, world_size=world_size)
[...]

PR #162150, opened 2025-09-04
[SymmMem] Skip multicast init if any CUDA call fails

kwen2501:
* __->__ #162150

Sometimes [NV Fabric Manager can be corrupted](https://github.com/pytorch/pytorch/issues/150852#issuecomment-2795766064), causing cuMulticast* calls to error out with "invalid arguments."

This PR improves the skip of multicast binding when error occurs. Also added the skip path for `CUmemFabricHandle` case.

Differential Revision: [D81693280](https://our.internmc.facebook.com/intern/diff/D81693280)


c10/cuda/driver_api.h
```diff
@@ -20,6 +20,22 @@
     }                                                                      \
   } while (0)
 
+#define C10_CUDA_DRIVER_CHECK_GOTO(EXPR, NEXT)                             \
+  do {                                                                     \
+    CUresult __err = EXPR;                                                 \
+    if (__err != CUDA_SUCCESS) {                                           \
+      const char* err_str;                                                 \
+      CUresult get_error_str_err [[maybe_unused]] =                        \
```
Skylion007, line 28: Huh, why is MaybeUnused necessary? it's used on line 30 isn't it?

ngimel (commented):
Given that 
1) we don't run fabric tests in CI
2) this is a soft fail
you'd need to manually verify that multimem init works both with fabric and without

c10/cuda/driver_api.h
kwen2501: lol, good catch, it is a copy from `C10_CUDA_DRIVER_CHECK`

ngimel (approved):

kwen2501:
Merging via https://github.com/pytorch/pytorch/pull/168049

PR #84243, opened 2022-08-30
Add NCCL PreMul Sum to c10d `redce` ops

crcrpar:
This is based on #81272 but this conforms to TorchScript Compiler

cc @ptrblck @kwen2501 @aazzolini 
cc @zasdfgbnm for visibility to the TODO above

PR #162165, opened 2025-09-04
Add ReduceOp.unbox and fix use-after-free in ReduceOp pybind

ezyang:
* __->__ #162165

Analogous to Work.unbox and ProcessGroup.unbox.

Fixes https://github.com/pytorch/pytorch/issues/162668


torch/csrc/distributed/c10d/init.cpp
```diff
@@ -904,7 +904,12 @@ This class does not support ``__members__`` property.)");
             } else {
               return ::c10d::makeNCCLPreMulSum(t[1].cast<at::Tensor>());
             }
-          }));
+          }))
+      .def_static("unbox", [](py::object obj) {
+          auto typePtr = torch::getCustomClass("__torch__.torch.classes.c10d.ReduceOp");
+          auto ivalue = torch::jit::toIValue(std::move(obj), typePtr);
+          return ivalue.toCustomClass<::c10d::ReduceOp>();
```
Skylion007:
```suggestion
          return std::move(ivalue).toCustomClass<::c10d::ReduceOp>();

File: torch/csrc/distributed/c10d/init.cpp [hunk 2]
```diff
@@ -904,7 +904,12 @@ This class does not support ``__members__`` property.)");
             } else {
               return ::c10d::makeNCCLPreMulSum(t[1].cast<at::Tensor>());
             }
-          }));
+          }))
+      .def_static("unbox", [](py::object obj) {
+          auto typePtr = torch::getCustomClass("__torch__.torch.classes.c10d.ReduceOp");
+          auto ivalue = torch::jit::toIValue(std::move(obj), typePtr);
```
Skylion007:
```suggestion
          auto ivalue = torch::jit::toIValue(std::move(obj), std::move(typePtr));
```

Skylion007 (approved):

ezyang:
This appears to corrupt malloc on OSX

dcci (approved):
Thanks for fixing this one

cyyever (approved):

ezyang:
OK, so the test failures here show the rabbit hole goes even deeper. 

When you define a class with CustomClassHolder, you are implicitly also inheriting intrusive_ptr_target. This means that if you have a reference to this class, you can take out an owning reference to the struct. This means it is VERY UNSAFE to allocate a CustomClassHolder subclass by value (as opposed to holding an intrusive_ptr to it), because it makes it very easy to accidentally destruct the object too early.

ReduceOp is a CustomClassHolder. So you really shouldn't treat it as a value type. But actually, the code interacting with it REALLY REALLY REALLY wants to treat it as a value type, because it actually used to be an enum until this was changed in https://github.com/pytorch/pytorch/pull/84243 (cc @kwen2501 @crcrpar ). So it actually will be annoying to port all the code to be an intrusive ptr, because all sorts of things like switch statements no longer work when it's a pointer-y type. It is doable, but it's so unnatural that the LLM has trouble doing it. And I don't really feel like uglifying the code by hand.

I think my plan to unblock LocalTensor is to just add a method to ReduceOp ScriptObject that lets me get at the underlying enum. This is enough for what I need. unbox will be in limbo until then.

PR #162169, opened 2025-09-04
[easy] [precompile] Convert CompileArtifacts to callable

jamesjwu:
* #162171
* #162170
* __->__ #162169

The goal of this PR stack is to be able to implement `aot_compile_module`, which AOT precompiles a torch.nn.Module. 
Step 1 is a simple refactor to make CompileArtifacts itself the callable, which makes it easier to use directly. 


test/dynamo/test_aot_compile.py
```diff
@@ -49,6 +49,7 @@ def forward(self, x):
 
 
 @torch._dynamo.config.patch("enable_aot_compile", True)
+@torch._dynamo.config.patch("strict_precompile", True)
```
zhxchen17: I think aot precompile will directly error now on the main branch, so we don't need this enabled

torch/compiler/__init__.py
```diff
@@ -660,4 +660,4 @@ def load_compiled_function(file: io.IOBase):
     from torch._dynamo.aot_compile import CompileArtifacts
 
     data = file.read()
-    return CompileArtifacts.deserialize(data).compiled_function()
+    return CompileArtifacts.deserialize(data)
```
zhxchen17: Right now I consider CompileArtifacts an impl detail and not sure we will run into issue if user can access its data directly later. I'm fine making it callable but it seems now we expose strictly more information from load_compiled_function (and .aot_compile())

could we return a wrapper that is callable-only every time we pass compiled function to user space (also to match what dynamo does today)?

Another option is to type annotation the return type as `Callable` so user might have an idea it's callable-only (but I don't prefer this since no one reads type annotation in python :( )

jamesjwu: Hmm, what's the reasoning behind wanting to hide this? I don't think it's particularly bad for the user to have access to the CompileArtifacts. My only minor qualm is the naming of it, it probably should be named something like AOTCompiledFunction. 

I don't think the user will care what the type is underneath here, and unnecessarily hiding it under a callable (which still stores the artifact in its `serialize` function) seems kind of pointless. 

I guess you could have a class CompiledFunction that has a CompiledArtifacts inside of it as an implementation. But again, I feel like this is unnecessary wrapping.

jamesjwu: Eh, I don't mind putting it under a _artifacts field instead to signify that the object is opaque, will try that

zhxchen17 (approved):

ezyang:
Happy to defer to Zhengxu here.

PR #162171, opened 2025-09-04
[Precompile] [RFC] Implement aot_compile_module

jamesjwu:
* #162527
* __->__ #162171

This PR adds a new interface _aot_compile to `OptimizedModule`, so that the following is possible: 

```
mod = SimpleLinearModule()
inputs = [
            ModelInput(
                args=(torch.randn(3, 3),),
                kwargs={},
                contexts=[torch.no_grad(), eval_mode(model)],
            ),
            ModelInput(
                args=(torch.randn(3, 3),), kwargs={}, contexts=[train_mode(model)]
            ),
        ]
        assert isinstance(model, torch._dynamo.eval_frame.OptimizedModule)
        model._aot_compile(
            inputs,
        )
```

After this PR, you can AOT precompile NanoGPT and use it to train directly. I'll share my fork of the repo to make this work. 

## ModelInput
The `ModelInput` API is a work in progress; for now it represents a set of inputs and contexts to instruct the compiler to compile. Most commonly, this is "compile an eval mode with no grad, and  a training mode with grad", but also contains things like autocasting contexts, etc. 

## Dispatch 
Dispatching is super simple here, we just iterate through all the precompiled fullgraphs and check guards for each one until there's one htat passes. I'm a bit worried that having this in python code is going to be too expensive. The guard checks are happening in C++ anyway, though, so the only python bottlenecked step here is just the for loop, so perhaps the overhead will not be high. I'll work on measuring this, though. 

## TODOs

This PR does not support `mod.compile()`, only `torch.compile(mod)`. In order to support `mod.compile()`, we'll need to update torch.nn.Module with an updated implementation — I can add that frontend later. 


ezyang:
Adding @anijain2305 for guidance on whether or not precompile dispatch should try to use C++ guard trees. I guess it would be nice? But I also think most precompile users should be skipping MOST guards and only doing the minimum set of tests that are necessary to get you from one dispatch to another (and at best, guards are just around as an optional debug mode, where performance is not important.)

zhxchen17:
fwiw I think C++ guard trees are simpler to use anyway today...

@ezyang in aot precompile case, we already throw away a bunch of guards automatically. By default, we already throw away all guards on global variables and id matches, so the rest guards shouldn't be many imo.

ezyang:
> in aot precompile case, we already throw away a bunch of guards automatically. By default, we throw away all guards on global variables and id matches, so the rest of the guards shouldn't be many imo.

I still think that there should be a fast mode where there are NO guards except the bare minimum needed to distinguish branches

jamesjwu:
> Hi, are you ready for bikeshedding on this API?

@ezyang Yes (though I may fix it in a later PR rather than this particular one). One other bikeshed to be had is the API around
`mod.compile()`: should we add a method to `torch.nn.Module` like `mod.aot_compile()`?

ezyang:
@zhxchen17 if I have something that took a list as input and then I pass in a tuple, do you want that to error?

zhxchen17 (approved):
seems changable, let's figure out toplevel API later.

torch/_dynamo/aot_compile.py
```diff
@@ -254,44 +255,137 @@ def new_guard_filter_fn(
...
+    # Represents a single forward function of a model along with dispatch
+    # compiled_results is serializable. We require the model to deserialize again.
+    model: torch.nn.Module
+    compiled_results: list[AOTCompiledFunction]
+
+    def __call__(self, *args: Any, **kwargs: Any) -> Any:
+        for result in self.compiled_results:
+            if result.guard_check(self.model, *args, **kwargs):
+                return result(self.model, *args, **kwargs)
+        # All guards failed, just run one of them and throw the guard check error.
```
ezyang, line 350: NGL this is pretty strange

torch/_dynamo/aot_compile.py
```diff
@@ -254,44 +255,137 @@ def new_guard_filter_fn(
...
+    kwargs: dict[str, Any]
+    contexts: list[AbstractContextManager[Any]]
+
+
+@dataclass
+class AOTCompiledModel:
+    # Represents a single forward function of a model along with dispatch
+    # compiled_results is serializable. We require the model to deserialize again.
+    model: torch.nn.Module
+    compiled_results: list[AOTCompiledFunction]
```
ezyang, line 344: This compiled results structure means that if some of your functions share a prefix of guards that must evaluate the same way, you end up repeatedly evaluating those guards over and over

torch/_dynamo/aot_compile.py
jamesjwu: I think instead we should probably actually do the work of verbosely logging every guard failure reason yea

torch/_dynamo/aot_compile.py
jamesjwu: Hmm yeah... but I'm not sure how a "decision tree" of guards would work. I think in order for something like that to make sense, the user has to explicitly note the guards they care about and have us only evaluate those or something.

PR #162172, opened 2025-09-04
[export] Fix torch.export.load with storage offset

yiming0416:
Summary: As titled

Test Plan:
CI

Rollback Plan:

Differential Revision: D81687701


angelayi (approved):

angelayi (commented):
thanks for the quick fix! could you add a test?

PR #162177, opened 2025-09-04
Fix comment on broadcasting example to clarify dimension mismatch

### LINKED ISSUE #162116: Feedback about Broadcasting semantics
There is the following issue on this page: https://docs.pytorch.org/docs/stable/notes/broadcasting.html#broadcasting-semantics

The docs state that one precondition is, "Each tensor has at least one dimension."  Here's the example given:
```
>>> x=torch.empty((0,))
>>> y=torch.empty(2,2)
# x and y are not broadcastable, because x does not have at least 1 dimension
```

However, x does have one dimension.  That dimension is of size zero.  It's not failing the "at least one dimension" test; it's failing the "dimensions must be equal, 1, or do not exist" test.

Indeed, broadcasting does seem to work with tensors with no dimensions:

```
>>> x = torch.tensor((0,0,0,0,0))
>>> assert x.shape == (5,)
>>> y = torch.tensor(1)
>>> assert y.shape == ()
>>> torch.broadcast_tensors(x,y)
(tensor([0, 0, 0, 0, 0]), tensor([1, 1, 1, 1, 1]))

>>> x + y
tensor([1, 1, 1, 1, 1])
```

I would suggest that the bullet item about the precondition "Each tensor has at least one dimension." be removed, as it does not align with observable behavior.  The example can be removed, or its comments can be adjusted to note that the 0 dimension does not match the 2 dimension.

dsashidh:
Fixes #162116 

Updated the comment in the broadcasting example to clarify that tensors with mismatched dimension sizes (0 vs 2) are not broadcastable. Removed incorrect reference to missing dimensions.

### DIFF
--- a/docs/source/notes/broadcasting.rst
+++ b/docs/source/notes/broadcasting.rst
@@ -13,7 +13,6 @@ General semantics
 -----------------
 Two tensors are "broadcastable" if the following rules hold:
 
-- Each tensor has at least one dimension.
 - When iterating over the dimension sizes, starting at the trailing dimension,
   the dimension sizes must either be equal, one of them is 1, or one of them
   does not exist.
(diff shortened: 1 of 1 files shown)


soulitzer (approved):

dsashidh:
Hi @soulitzer . Thanks for rebasing earlier.
The PR is still blocked as CI workflows haven't run. There are no skip ci keywords, so I think it's a permissions issue. Can trigger the workflows please?

PR #162186, opened 2025-09-04
[CUDA][CUDAGraph] Reduce capture overhead in CUDA Graph memory reuse

eee4017:
Previous work #158352 delivered CUDAGraph memory footprint reduction with no replay-time impact, but capture time regressed (up to 20× slower) due to repeated full-graph traversals. See previous benchmark results [here](https://github.com/pytorch/pytorch/pull/158352#issuecomment-3215947565)

This PR removes capture/reply overhead while preserving the memory savings:

1. **Terminals as free markers**
   We stop inserting empty nodes and instead record the current stream terminals as free markers. This avoids mutating the user’s graph and keeps semantics unchanged.

2. **Incremental, cached reachability**
   We add a **per-graph reuse context** that caches reverse-traversal state:

   * `graph_reuse_context[graph].visited[stream]` tracks nodes already seen from that stream’s terminal frontier.
   * On each allocation during capture, we resume traversal from the latest terminals and only visit unseen nodes.
   * A block is freed when all its recorded markers are in the visited set of its allocation stream—i.e., all markers are proven predecessors of future work.

See [the performance results here](https://docs.google.com/spreadsheets/d/e/2PACX-1vRPvdd9Xa8W87ixbiA0da_qvOhrUAjUpFz0G-_j-MsDnoeRyhEa4_ut_W3rqcg1VVZVFJ-gucwov-3b/pubhtml?gid=1468302443&single=true), we sweep synthetic multi-stream CUDA Graphs built by `capture_benchmark.py` (same as before, we generate random interleaving of alloc/free/join with given probabilities, see [gist here](https://gist.github.com/eee4017/e2092d215b1d4bd46534148939af39e3)), and we compare median capture/replay times and memory. On an NVIDIA H100 PCIe across 24 configs, the optimization preserves reserved memory reduction at ~24–98%, leaves allocated memory unchanged, and brings capture time back to baseline (range 0.96–1.04× vs. baseline) with replay time unchanged (range 0.97–1.11×).

cc @ptrblck @msaroufim @eqy @jerryzh168 @mcarilli @ezyang @eellison @penguinwu @BoyuanFeng

### DIFF
--- a/c10/cuda/CUDACachingAllocator.cpp
+++ b/c10/cuda/CUDACachingAllocator.cpp
@@ -1189,9 +1189,7 @@ class DeviceCachingAllocator {
     ska::flat_hash_map<cudaStream_t, ska::flat_hash_set<cudaGraphNode_t>>
         visited;
   };
-  ska::flat_hash_map<cudaGraph_t, GraphReuseContext> graph_reuse_context;
-  // Each stream should only be in one graph
-  ska::flat_hash_map<cudaStream_t, cudaGraph_t> stream_to_graph;
+  ska::flat_hash_map<CaptureId_t, GraphReuseContext> graph_reuse_context;
 
   // outstanding cuda events
   ska::flat_hash_map<
@@ -1650,6 +1648,7 @@ class DeviceCachingAllocator {
 
   struct CaptureInfo {
     cudaGraph_t graph{};
+    CaptureId_t capture_id{0};
     const cudaGraphNode_t* terminals{nullptr};
     size_t num_terminals{0};
     cudaStreamCaptureStatus status{cudaStreamCaptureStatusNone};
@@ -1661,7 +1660,7 @@ class DeviceCachingAllocator {
     C10_CUDA_CHECK(cudaStreamGetCaptureInfo(
         stream,
         &info.status,
-        nullptr,
+        &info.capture_id,
         &info.graph,
         &info.terminals,
         nullptr,
@@ -1670,7 +1669,7 @@ class DeviceCachingAllocator {
     C10_CUDA_CHECK(cudaStreamGetCaptureInfo_v2(
         stream,
         &info.status,
-        nullptr,
+        &info.capture_id,
         &info.graph,
         &info.terminals,
         &info.num_terminals));
@@ -1808,15 +1807,7 @@ class DeviceCachingAllocator {
     if (info.status == cudaStreamCaptureStatusNone || info.num_terminals == 0) {
       return;
     }
-    auto& graph_context = graph_reuse_context[info.graph];
-    // If the stream is not in the graph context, we add it to the
-    // stream_to_graph map
-    if (graph_context.visited.count(stream) == 0) {
-      TORCH_INTERNAL_ASSERT(
-          stream_to_graph.count(stream) == 0,
-          "Each stream should only be involved in one graph");
-      stream_to_graph[stream] = info.graph;
-    }
+    auto& graph_context = graph_reuse_context[info.capture_id];
     auto& visited = graph_context.visited[stream];
     update_visited(info, visited);
 
@@ -2485,32 +2476,22 @@ class DeviceCachingAllocator {
   // Called by CUDAGraph::capture_end
   void endAllocateToPool(MempoolId_t mempool_id) {
     std::lock_guard<std::recursive_mutex> lock(mutex);
-    // Erase the streams and graphs from the stream_to_graph map
-    ska::flat_hash_set<cudaStream_t> streams_to_erase;
-    for (auto& [stream, graph] : stream_to_graph) {
-      if (streams_to_erase.count(stream) > 0) {
-        continue;
-      }
-      // The stream is not capturing, so we are going to erase the graph from
-      // the graph_reuse_context
-      if (stream_get_capture_info(stream).status ==
-          cudaStreamCaptureStatusNone) {
-        auto& graph_context = graph_reuse_context[graph];
-        // There might be other streams in the graph, so we need to erase them
-        // too
-        for (auto& [other_stream_in_graph, _] : graph_context.visited) {
+
+    if (CUDAAllocatorConfig::graph_capture_record_stream_reuse() &&
+        !graph_reuse_context.empty()) {
+      TORCH_INTERNAL_ASSERT(
+          graph_reuse_context.size() == 1, "Expected only one graph context");
+      for (auto& [capture_id, graph_context] : graph_reuse_context) {
+        for (auto& [stream, _] : graph_context.visited) {
           TORCH_INTERNAL_ASSERT(
-              stream_get_capture_info(other_stream_in_graph).status ==
+              stream_get_capture_info(stream).status ==
                   cudaStreamCaptureStatusNone,
               "This stream should not be capturing when the capture is ended");
-          streams_to_erase.insert(other_stream_in_graph);
         }
-        graph_reuse_context.erase(graph);
       }
+      graph_reuse_context.clear();
     }
-    for (auto stream : streams_to_erase) {
-      stream_to_graph.erase(stream);
-    }
+
     for (auto it = captures_underway.begin(); it != captures_underway.end();
          ++it) {
       if (it->first == mempool_id) {
--- a/c10/cuda/CUDACachingAllocator.cpp
+++ b/c10/cuda/CUDACachingAllocator.cpp
@@ -1653,10 +1653,6 @@ class DeviceCachingAllocator {
     cudaStreamCaptureStatus status{cudaStreamCaptureStatusNone};
   };
 
-  // Returns the current set of "terminal" nodes in the CUDA graph for a given
-  // stream. These represent the current endpoints of the stream, and may
-  // include additional nodes if the graph branches. Any new work captured will
-  // be attached after one or more of these terminals.
   inline CaptureInfo stream_get_capture_info(cudaStream_t stream) {
     CaptureInfo info{};
 #if (defined(CUDA_VERSION) && CUDA_VERSION >= 13000)
--- a/c10/cuda/CUDACachingAllocator.cpp
+++ b/c10/cuda/CUDACachingAllocator.cpp
@@ -1190,6 +1190,8 @@ class DeviceCachingAllocator {
         visited;
   };
   ska::flat_hash_map<cudaGraph_t, GraphReuseContext> graph_reuse_context;
+  // Each stream should only be in one graph
+  ska::flat_hash_map<cudaStream_t, cudaGraph_t> stream_to_graph;
 
   // outstanding cuda events
   ska::flat_hash_map<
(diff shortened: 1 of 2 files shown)


eqy (approved):

c10/cuda/CUDACachingAllocator.cpp
```diff
@@ -1820,36 +1787,33 @@ class DeviceCachingAllocator {
...
+      if (markers.empty() || block->stream != stream) {
         continue;
       }
 
       bool is_reusable = true;
-
-      for (const auto& node : inserted_empty_nodes) {
-        if (reusable_empty_nodes.find(node) == reusable_empty_nodes.end()) {
+      for (auto m : markers) {
+        if (!visited.count(m)) {
```
ngimel, line 1847: is it a sufficient condition? a node is marked as visited if it can be reached from just one terminal, not all of them?

c10/cuda/CUDACachingAllocator.cpp
```diff
@@ -1820,36 +1787,33 @@ class DeviceCachingAllocator {
...
-    auto reusable_empty_nodes = get_reusable_empty_nodes(stream);
+    auto info = stream_get_capture_info(stream);
 
     // If there are no reusable empty nodes (e.g., not currently capturing),
     // there is nothing to do.
-    if (reusable_empty_nodes.empty()) {
+    if (info.status == cudaStreamCaptureStatusNone || info.num_terminals == 0) {
       return;
     }
+    auto& graph_context = graph_reuse_context[info.graph];
```
ngimel: so entries in the `graph_reuse_context` are leaked and never deleted? Would it also prevent freeing resources associated with the graph when the user wants to destroy a graph?

eee4017: Hi, you’re right and I do want to remove the cached graph info once capturing ends. The most natural place seems to be [endAllocateToPool](https://github.com/pytorch/pytorch/blob/582d278983b28a91ac0cedd035183f2495bb6887/c10/cuda/CUDACachingAllocator.cpp#L2503). We could even move the graph-parsing logic into `at::cuda::cudaGraph` if correct graph-pool interaction API is defined. The difficulty is that `endAllocateToPool` doesn’t currently pass enough context, and extending that API would cascade into a lot of changes. Do you see a better approach to ensure cached graph info is released without overcomplicating the allocator API?

c10/cuda/CUDACachingAllocator.cpp
eee4017: From my understanding, a stream will only have a single terminal in most cases, so the condition holds. But when that’s not true, we do need to ensure that each marker is reachable from all terminals of the stream. My plan for multi-terminal stream is to cache only those nodes that can be reached by every terminal in the stream.

ngimel: image
In this situation, when you call update_visited all markers will be marked as visited? But the streams aren't joined, it's unsafe to reuse the memory? Or am I misunderstanding your logic

eee4017: Thanks for pointing this out.

The visited set is maintained per stream, not globally. So in your diagram, the usual case should be terminal0 belongs to stream0 and terminal1 belongs to stream1. Also, stream0 and stream1 are not joining. Then: On stream0, only marker0 would appear in its visited set of stream0, since marker1 is not reachable from stream0. That way, markers from unrelated streams aren’t accidentally marked as reusable.

Where it gets more subtle is when a **single stream itself has multiple terminals** (num_terminals > 1 [here](https://github.com/pytorch/pytorch/pull/162186/files#diff-acc6337586bf9cdcf0a684380779300ec171897d05b8569bf439820dc8c93bd5R1760)). This can sometimes happen, for example: 
1. When another stream is about to join into stream0 (see the [test program](https://gist.github.com/eee4017/06450105cf05c53fb2910319c6db6dab)). Immediately after the join, stream0 reports two dependencies. 
2. when we manually insert extra dependencies into the graph. 

In those situations, I’ll adjust it to only keep the intersection of ancestors of all the stream’s current terminals, and then revisit once the streams are fully joined. I’ll submit a commit with this fix shortly.

ngimel: For the situation 1 your algo should still be fine, as the streams are about to be joined, so you can update dependencies on them both?

eee4017: Yes, the current code should behave correctly in the about-to-join case. The challenge is that we can’t easily distinguish that case from rarer scenarios (like manually added deps), and there may be other ways for a single stream to end up with multiple terminals. So I think it’s safer to handle all multi-terminal cases conservatively.

ngimel: Manually added deps are still the same, no? They will be ordered before whatever comes after, so at this point it's safe to reuse the memory

eee4017: Hi, I’ve experimented with some single-stream-multiple-terminal scenarios and also didn’t find anything unsafe. When cudaStreamGetCaptureInfo reports num_dep > 1, it’s returning the current dependencies of the next node. As the docs say: “The next node to be captured in the stream will depend on this set of nodes”. That means it should always be safe at this point to free the previous marker. @glav do you see any counterexamples where this might not hold?

c10/cuda/CUDACachingAllocator.cpp
eee4017: I’ve added a map cleanup [here](https://github.com/pytorch/pytorch/pull/162186/files#diff-acc6337586bf9cdcf0a684380779300ec171897d05b8569bf439820dc8c93bd5R2489) to ensure the cached info is deleted once capture ends.

c10/cuda/CUDACachingAllocator.cpp
```diff
@@ -2511,6 +2485,32 @@ class DeviceCachingAllocator {
...
     std::lock_guard<std::recursive_mutex> lock(mutex);
+    // Erase the streams and graphs from the stream_to_graph map
+    ska::flat_hash_set<cudaStream_t> streams_to_erase;
+    for (auto& [stream, graph] : stream_to_graph) {
+      if (streams_to_erase.count(stream) > 0) {
+        continue;
+      }
+      // The stream is not capturing, so we are going to erase the graph from
+      // the graph_reuse_context
+      if (stream_get_capture_info(stream).status ==
```
ngimel: is stream capturing at this point an error?

c10/cuda/CUDACachingAllocator.cpp
```diff
@@ -1638,189 +1648,148 @@ class DeviceCachingAllocator {
...
+  inline CaptureInfo stream_get_capture_info(cudaStream_t stream) {
+    CaptureInfo info{};
 #if (defined(CUDA_VERSION) && CUDA_VERSION >= 13000)
-      C10_CUDA_CHECK(cudaStreamGetCaptureInfo(
-          stream, &status, nullptr, &graph, &deps, nullptr, &num_deps));
+    C10_CUDA_CHECK(cudaStreamGetCaptureInfo(
+        stream,
+        &info.status,
+        nullptr,
+        &info.graph,
```
ngimel, line 1666: Actually, I think you should switch from using `graph` as a map key to using unique id, `graph` is just a pointer, after hashing its not guaranteed to be unique, and what you want is a unique identifier for a graph, see e.g. how it's currently used in `capture_begin` https://github.com/pytorch/pytorch/blob/cf94cadbeee31a4d1d46a57f11bce7c9fd1cebc0/aten/src/ATen/cuda/CUDAGraph.cpp#L103

c10/cuda/CUDACachingAllocator.cpp
eee4017: I believe we only ever have a single active capture_id at a time. We are not allowed to do multi-threaded or nested capture for the same stream set, so concurrent multi-graph capture shouldn’t occur here. Given that, I’m assuming the active graph count is 1 and, at capture_end, every stream in that graph’s context should already be non-capturing. I added an assertion check and simplified cleanup under this assumption. If anyone knows a counter-example where multiple capture IDs can be live simultaneously in this codepath, please let me know.

c10/cuda/CUDACachingAllocator.cpp
eee4017: Fixed, thanks.

c10/cuda/CUDACachingAllocator.cpp
```diff
@@ -2511,6 +2476,22 @@ class DeviceCachingAllocator {
   // Called by CUDAGraph::capture_end
   void endAllocateToPool(MempoolId_t mempool_id) {
     std::lock_guard<std::recursive_mutex> lock(mutex);
+
+    if (CUDAAllocatorConfig::graph_capture_record_stream_reuse() &&
+        !graph_reuse_context.empty()) {
+      TORCH_INTERNAL_ASSERT(
+          graph_reuse_context.size() == 1, "Expected only one graph context");
```
ngimel: is that true? You could be in the process of capturing multiple graphs (e.g. each thread is capturing its own). I think you also need another map from mempool_id to capture_id, and then you won't need this assert.

eee4017: Fixed. Thanks.

c10/cuda/CUDACachingAllocator.cpp
```diff
@@ -2511,6 +2495,26 @@ class DeviceCachingAllocator {
   // Called by CUDAGraph::capture_end
   void endAllocateToPool(MempoolId_t mempool_id) {
     std::lock_guard<std::recursive_mutex> lock(mutex);
+
+    if (CUDAAllocatorConfig::graph_capture_record_stream_reuse() &&
+        !graph_reuse_context.empty()) {
+      std::vector<CaptureId_t> capture_ids_to_remove;
```
ngimel: why is it a vector? There couldn't be multiple capture_ids associated with this mempool_id?

eee4017: Yes, I think there should normally be only one CaptureId_t per endAllocateToPool. In the case of multi-threaded capture, is there any rare situation where two captures could use the same memory pool? I don’t think that’s allowed, though.

ngimel: It's not allowed https://github.com/pytorch/pytorch/blob/7a9c4d794c5048ee8ba4651f522292dee1116977/c10/cuda/CUDACachingAllocator.cpp#L2504

ngimel: So can you change it from vector to a value?

eee4017: I’ve updated the code, please take a look when you have a chance. Thanks!

ngimel (approved):

c10/cuda/CUDACachingAllocator.cpp
eqy: Leaving a link to this here as I was looking for it when thinking about this discussion https://github.com/pytorch/pytorch/blob/c1f40d33c89b361a1edad17aa25cfff1ab4014fd/torch/cuda/graphs.py#L226

PR #162193, opened 2025-09-04
[SymmMem] Feed tensor.data_ptr instead of handle.buffer_ptr into kernels

kwen2501:
* __->__ #162193

After MemPool support, `get_buffer_ptrs` points to base address of allocation segment.


torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
```diff
@@ -110,12 +110,13 @@ at::Tensor nvshmem_broadcast(at::Tensor& input, const int64_t root, const std::s
   auto input_hdl = c10d::symmetric_memory::rendezvous(input, group_name);
   int rank = input_hdl->get_rank();
   auto team = group_to_team(group_name, input_hdl->get_rank_to_global_rank());
-  void* buffer_ptr = input_hdl->get_buffer_ptrs()[rank];
+  void* buffer_ptr = input.data_ptr();
```
Skylion007: Shouldn't we use mutable_data_ptr instead of data_ptr?

kwen2501: Thanks a lot! Changed!

ngimel (approved):
we probably need to audit triton tests too

PR #162194, opened 2025-09-04
[dynamic shapes] DynamicInts prototype

pianpwk:
Initial prototype for dynamic int inputs, allows users to run with `torch.compile(f)(DynamicInt(4))`, compiling dynamically and using the underlying hint at runtime.

Current behavior:
- Also works in eager (mostly by subclassing int), as scalar input to torch functions, or numpy/math/etc. For example, `x = DynamicInt(3); torch.randn(x); torch.add(y, z, alpha=x); np.arange(x)` all act as if x = 3.
- Behavior for arithmetic ops is to return new DynamicInts rather than static ints; `DynamicInt(3) * 2 = DynamicInt(6)`. This is via SymNode magic methods, but coverage might not be 100% - for example, I had to explicitly override floordiv to avoid int casting. This is not necessarily the case for non-magic method ops (e.g. `math.cos(x)`). The alternative here is to int cast on all operations, but I opted for this for dynamism propagation in non-compiled regions.
- Doesn't ban fullgraph=False; DynamicInt objects might be leaked back to the user, but I guess this is fine, because they can be casted to ints when needed?
- Dynamo only allocates one symbol per DynamicInt; specifying the same DynamicInt for multiple inputs leads to input deduplication, and a guard installed.
- We don't raise on int specialization (in allowlist/maybe_mark_dynamic style) - but an easy change if needed.
- DynamicInts as nn.Module attributes are handled.
- We don't guard on the DynamicInt id, e.g. users can do the following without recompiling (maybe we should guard?)
```python
x = DynamicInt(4)
f(x)
f(1)
f(DynamicInt(3))  # same as f(3)
```

Follow-up work:
- Specifying shape constraints, either at the int-level, e.g.
```python
DynamicInt(64, name="s0", constraints=["s0 % 32 == 0", "s0 <= 1024"]
```
or at the compilation level, e.g. something like
```python
s0 = DynamicInt(64, name="s0")
s1 = DynamicInt(128, name="s1")
with some_compiler_config.dynamic_int_constraints(["s1 == 2*s0", "s0 % 32 == 0"]):
    f(s0, s1)
```
This should subsume the need for specifying derived SymInts?
- SymFloat support - currently it seems backed floats are specialized by the tensorify float pass, and there's no handling in inductor.
- Propagating dynamism in tensor constructors, e.g. `x = DynamicInt(4); torch.randn(x)` could annotate `_dynamo_dynamic_indices`.

Differential Revision: D81698719

cc @ezyang @SherlockNoMad @EikanWang @jgong5 @wenzhe-nrv @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela

### DIFF
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -1818,6 +1818,96 @@ class TestSymNumberMagicMethods(TestCase):
         self.assertTrue(isinstance(s3, int))
         self.assertTrue(str(s1.node.expr) != str(s2.node.expr))
 
+    @fresh_cache()
+    @torch._dynamo.config.patch("capture_scalar_outputs", True)
+    @parametrize("backend", ["inductor", "eager"])
+    def test_dynamic_int_basic_compile(self, backend):
+        from torch.fx.experimental.sym_node import DynamicInt
+
+        cnt = CompileCounterWithBackend(backend)
+
+        # test scalar inputs to function
+        def f(x, y, z):
+            out = torch.tensor([x + y + z])
+            out = out + torch.zeros(abs(x) + 2).sum()  # test out tensor construction
+            return out
+
+        fn = torch.compile(f, fullgraph=True, backend=cnt)
+        x = DynamicInt(1)
+        z = DynamicInt(3)
+        self.assertEqual(fn(x, x, z), f(1, 1, 3))  # guard: x == y
+        self.assertEqual(fn(2, 2, 0), f(2, 2, 0))
+        self.assertEqual(fn(-1, -1, 2), f(-1, -1, 2))
+        self.assertEqual(cnt.frame_count, 1)  # no recompiles
+
+        self.assertEqual(fn(3, 4, 5), f(3, 4, 5))  # now we recompile
+        self.assertEqual(cnt.frame_count, 2)
+
+        # test nn module property
+        class Foo(torch.nn.Module):
+            def __init__(self):
+                super().__init__()
+                self.i = DynamicInt(1)
+
+            def forward(self, x):
+                return torch.tensor([x + self.i])
+
+        cnt.clear()
+        m = Foo()
+        mc = torch.compile(m, backend=cnt, fullgraph=True)
+
+        self.assertEqual(mc(DynamicInt(0)), m(0))
+        mc.i = -2  # override attribute
+        self.assertEqual(mc(-1), m(-1))
+        self.assertEqual(cnt.frame_count, 1)
+
+    def test_dynamic_int_eager_usage(self):
+        from torch.fx.experimental.sym_node import DynamicInt
+
+        w = DynamicInt(-1)
+        x = DynamicInt(0)
+        y = DynamicInt(1)
+        z = DynamicInt(2)
+
+        def check(l, r):
+            self.assertTrue(isinstance(l, DynamicInt))
+            self.assertEqual(l, r)
+
+        # test arithmetic
+        check(2 * y + z, 4)
+        check((10 - z) // 2, 4)
+        check(1 // z, 0)
+        check(-w + w**2, 2)
+        check(x % z, 0)
+        check(1 << z, 4)
+        check(z | y, 3)
+        check(min(y, z), 1)
+        self.assertTrue(z > -2)
+        with self.assertRaises(ZeroDivisionError):
+            y % x
+
+        # math, numpy
+        self.assertEqual(math.cos(x), y)
+        self.assertEqual(math.prod([z, z], start=z), 8)
+        self.assertEqual(np.arange(z)[y], 1)
+        self.assertTrue(np.allclose(np.ones([y, z]).sum(axis=x), np.ones(z)))
+
+        # test conversions
+        self.assertTrue(isinstance(x + 2, int))
+        self.assertTrue(isinstance(x + 2, DynamicInt))
+        self.assertEqual(y / 2.0, 0.5)  # this could return DynamicFloat in future
+        self.assertEqual(float(z), 2.0)
+        self.assertFalse(bool(x))
+        self.assertEqual(DynamicInt(x).real, x.real)
+
+        # torch functions, scalar inputs
+        self.assertEqual(torch.arange(z)[:w][x], 0)
+        self.assertEqual(torch.add(torch.tensor(w), torch.tensor(w), alpha=z), -3)
+        self.assertEqual(
+            list(torch.nn.Linear(z, y)(torch.randn(z * 2, z)).shape), [4, 1]
+        )
+        self.assertEqual(z * torch.ones(z).sum(dim=x), 4)
+
 
 instantiate_parametrized_tests(TestSymNumberMagicMethods)
 
--- a/torch/_dynamo/source.py
+++ b/torch/_dynamo/source.py
@@ -527,6 +527,29 @@ class ConvertIntSource(ChainedSource):
 
 
 @dataclasses.dataclass(frozen=True)
+class DynamicScalarSource(ChainedSource):
+    is_int: bool
+
+    def __post_init__(self) -> None:
+        assert self.base is not None
+
+    def reconstruct(self, codegen: "PyCodegen") -> None:
+        # Integer casting at reconstruction helps reduce the amount of DynamicInts returned
+        # to the user, in favor of plain ints.
+        # For example, a compiled region that only does int arithmetic could return a
+        # DynamicInt without the casting here.
+        codegen.add_push_null(lambda: codegen.load_import_from("builtins", "int"))
+        codegen(self.base)
+        codegen.extend_output(create_call_function(1, False))
+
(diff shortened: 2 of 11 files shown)


torch/_dynamo/source.py
```diff
@@ -526,6 +526,31 @@ def name(self) -> str:
...
+        codegen(self.base)
+        codegen.extend_output(create_call_function(1, False))
+
+    def guard_source(self) -> GuardSource:
+        return self.base.guard_source()
+
+    def name(self) -> str:
+        return f"int({self.base.name()})"
+
+
```
pianpwk, line 551: I wasn't sure how to handle things here, but this does int() casting for input & guard handling, so both DynamicInts and ints are handled

torch/_dynamo/variables/builder.py
```diff
@@ -1097,6 +1102,51 @@ def build_key_value(i, k, v):
         ):
             self.install_guards(GuardBuilder.FUNCTION_MATCH)
             return ItertoolsVariable(value, source=self.source)
+        elif isinstance(value, DynamicScalar):
+            if not self.tx.error_on_graph_break:
+                raise AssertionError(
+                    "DynamicInts are only compatible with fullgraph compile, please use torch.compile(fullgraph=True)"
+                )
```
pianpwk: raising on fullgraph=False - I think what's actually needed right now, is to raise when this is passed into a non-compiled region?

torch/_dynamo/source.py
```diff
@@ -526,6 +526,25 @@ def name(self) -> str:
...
 
 
+@dataclasses.dataclass(frozen=True)
+class DynamicScalarSource(ChainedSource):
+    is_int: bool
+
+    def __post_init__(self) -> None:
+        assert self.base is not None
+
+    def reconstruct(self, codegen: "PyCodegen") -> None:
```
williamwen42, line 536: Are you allowing the compiled function to return a DynamicScalar normally? i.e. no graph breaks, but a normal `return` statement.

If so, then I don't think this function needs to be defined. `reconstruct` should only be called for values that are needed outside the compiled region - in both the case of graph break and return.

You might not even need the `self.tx.one_graph` check below then - you can just put an `unimplemented_v2` call here (this should cause dynamo to skip the frame), or assert failure.

williamwen42, line 536: Can you check if this implementation is necessary? You mentioned that this should not escape the compiled region, making me think that we should not ever reconstruct this, but maybe it's possible for this value to be an intermediate value during reconstruction that's not actually returned?

williamwen42: cc @StrongerXi the source of our discussion on reconstruct vs. what is allowed to escape the compiled region

ezyang:
> Doesn't work in eager yet; x = DynamicInt(3); torch.randn(x); x + 1 won't work

What makes this hard?

ezyang:
> Only supports fullgraph=True, as dynamo source handling is responsible for int casting, and we risk leaking a DynamicInt object back to the user as runtime output.

This, in conjunction with eager doesn't work, is confusing to me, actually. If you are forcing fullgraph=True, you don't need to annotate the argument at all; just pass in the dynamic shapes information indirectly to the compiler via the torch.compile (some sort of arg spec, similar to JAX style APIs). The point of an explicit DynamicInt struct is that it CAN propagate around and trigger dynamism at potentially unpredictable spots in the presence of graph breaks. And yes, running in eager should work; if you are unable to trigger `__int__` conversions then the user is obligated to rewrite their code for it!

bobrenjc93:
> And yes, running in eager should work;

+1 this should definitely go hand-in-hand with the compile implementation.

> The point of an explicit DynamicInt struct is that it CAN propagate around and trigger dynamism at potentially unpredictable spots in the presence of graph breaks.

While I think this would be great, I would contend that the biggest benefit of DynamicInt is it finally gives users an affordance to mark dynamism on ints and that it might make sense to take an incremental approach here (eg. doing dynamism propogation in another PR). For example @anijain2305 recently had a rough time marking dynamism of an int property of an nn_module https://github.com/huggingface/transformers/pull/40328/files. @Chillee has mentioned similar woes in the past in his support group. As MSL scales up, I'm sure they'll run into this wall and it'd be nice to have a go-to recommended solution.

> some sort of arg spec, similar to JAX style APIs

Perhaps there is some bikeshedding we need to do, but I personally think something like wrapping your ints in a DynamicInt is more ergonomic than torch.compile kwargs. It's a one line change and as a user, I don't need to learn about the custom DSL we decide on to map the arg spec syntax to the actual spot of dynamism (eg. we use source path for dynamic allowlist and that's not a trivial learning curve). Perhaps I'm misunderstanding the suggestion? If you have an example in mind for how to mark a deeply nested nn.module int property, I'd love to see it!

torch/_dynamo/source.py
pianpwk: Ah now `codegen(self.base)` works, but that might've also come from other changes around eager handling.

ezyang:
> It's a one line change and as a user, I don't need to learn about the custom DSL we decide on to map the arg spec syntax to the actual spot of dynamism (eg. we use source path for dynamic allowlist and that's not a trivial learning curve). Perhaps I'm misunderstanding the suggestion?

I think it's more tractable if the DSL is (1) solely for input arguments and (2) only works for "structural" types of some sort, so the user can express it in any valid Python syntax and we can map it to the correct Source. For example, if you could only do it for pytrees, I think this is tractable, esp. since that's how export specifies dynamism anyway.

I am not saying we shouldn't do DynamicInt, but we should be doing BOTH of these things, since DynamicInt simulating int is imperfect and there can be cases where it is inconvenient to ensure eager keeps working.

torch/fx/experimental/sym_node.py
```diff
@@ -625,6 +625,29 @@ def is_constant(self):
...
+    def __repr__(self):
+        return f"DynamicInt({self.real})"
+
+    def __floordiv__(self, other):  # // was casting to int without these overrides?
+        return DynamicInt(self.real // other)
+
+    def __rfloordiv__(self, other):
+        return DynamicInt(other // self.real)
+
+
```
pianpwk, line 661: moved this here from symbolic_shapes to avoid importing sympy by default, but maybe there's a better place to put it?

test/test_dynamic_shapes.py
```diff
@@ -1818,6 +1818,95 @@ def test_stride_symnode(self):
...
+        # test scalar inputs to function
+        @torch.compile(fullgraph=True, backend=cnt)
+        def fn(x, y, z):
+            out = torch.tensor([x + y + z])
+            out = out + torch.zeros(abs(x) + 2).sum()  # test out tensor construction
+            return out
+
+        x = DynamicInt(1)
+        z = DynamicInt(3)
+        self.assertEqual(fn(x, x, z), 5)  # guard: x == y
```
bobrenjc93: instead of hard coding the answer, can you compare with eager?

eg.

fn_c = torch.compile(fn)
self.assertEqual(fn(x, y, z), fn_c(x, y, z))

torch/fx/experimental/sym_node.py
```diff
@@ -1790,11 +1817,27 @@ def rbinary_magic_impl(self, other):
         ret = wrap_node(getattr(other_node, method_attr)(self.node))
         return get_constant(ret) if is_constant(ret) else ret
 
+    def setattrs(user_type, attr, impl):
```
bobrenjc93: What is this doing? can you write a comment?

torch/_dynamo/source.py
bobrenjc93: So is this necessary @pianpwk?

torch/_dynamo/source.py
bobrenjc93: I'm not sure what this comment is in reference to. Did you remove the code in question?

pianpwk: ah, this was just in reference to the entire implementation.

torch/_dynamo/source.py
pianpwk: I talked to @williamwen42 offline, and I don't think this isn't necessary, but pretty desirable to have, to leak less DynamicInts back to users (at the cost of slightly less dynamism propagation). Added a comment.

docs/source/conf.py
```diff
@@ -2727,6 +2727,7 @@
     # torch.fx.experimental.schema_type_annotation
     "AnnotateTypesWithSchema",
     # torch.fx.experimental.sym_node
+    "DynamicInt",
```
albanD: Adding new entries in this file is not allowed. Please make sure to properly document all new public APIs !

albanD (dismissed):

pianpwk:
ready for another pass

bobrenjc93 (approved):

PR #121225, opened 2024-03-05
Remove unnecessary const_casts

cyyever:
Fixes #ISSUE_NUMBER

cc @mrshenli @pritamdamania87 @zhaojuanmao @satgera @rohan-varma @gqchen @aazzolini @osalpekar @jiayisuse @H-Huang @kwen2501 @awgu @penguinwu @fegin @XilunWu @wanchaol @fduwjj @wz337 @tianyu-l @wconstab @yf225 @chauhang

PR #162195, opened 2025-09-04
Upgrades dlpack to v1.1 to include fp8/fp4

syed-ahmed:
* __->__ #162195


aten/src/ATen/dlpack.h
```diff
@@ -229,12 +253,12 @@ typedef struct {
   /*! \brief The data type of the pointer*/
   DLDataType dtype;
   /*! \brief The shape of the tensor */
-  const int64_t* shape;
+  int64_t* shape;
   /*!
    * \brief strides of the tensor (in number of elements, not bytes)
    *  can be NULL, indicating tensor is compact and row-majored.
    */
-  const int64_t* strides;
```
Skylion007, line 237: Why the const cast changes??

syed-ahmed: That's a valid question, but this was changed by dlpack in v1.1, and I left it unchanged in this PR. The question should be answered by dlpack (cc: @tqchen ). I'm not sure if we should diverge from dlpack header. That's why needed to make the necessary change (`const_cast<int64_t*>`) in pytorch to consume dlpack v1.1.

eqy (approved):

aten/src/ATen/dlpack.h
tqchen: from ABI pov either way is fine, but the intend was to not modify things

albanD (approved):
Sure.
FYI @rgommers

Skylion007 (approved):

aten/src/ATen/dlpack.h
rgommers: This was never changed in DLPack since its initial commit, see https://github.com/dmlc/dlpack/blame/addbc8b3d9449691d01827ac4a0e0d035cf8ea40/include/dlpack/dlpack.h#L258

I agree it doesn't matter, but I had a look at why it was changed in PyTorch. It was done last year in gh-121225. There are no review comments on that PR, but I think that PR missed that `dlpack.h` was a vendored header and modifying it for a trivial reason is a bad idea - it creates a diff on each upgrade, causing new PR authors and reviewers to wonder why this was changed in PyTorch.

This changes reverts that part of gh-121225, which looks like the right thing to do.

rgommers (approved):
LGTM after looking into the const cast change. Thanks @syed-ahmed, all!

aten/src/ATen/dlpack.h
syed-ahmed: @rgommers Thanks for pointing that out. I wasn't aware of it and was working under the assumption that we just copy dlpack header and make no modifications.

syed-ahmed:
Fixing CI error.

aten/src/ATen/dlpack.h
rgommers: > and was working under the assumption that we just copy dlpack header and make no modifications.

That is the right way of working indeed. And that's how it was until someone made a local change

syed-ahmed:
Needed to set lanes and bits for float4 and then also do some error handling for number of lanes.

PR #162202, opened 2025-09-04
[DCP][Quantization] Fix for FP8 multiplication during dequantization

saumishr:
Summary:
Weight vector needs to be upcasted since some FP8 formats (like Float8_e4m3fn) don't have CPU implementations in PyTorch. Reference: https://docs.pytorch.org/docs/stable/tensors.html#id13

We will use FP32 for the scale vector multiplication and convert to the target dtype. 

Upcasting helps with the following:

1.  **Full CPU support**: `float32` has complete CPU kernel implementations for all operations
2.  **Numerical stability**: `float32` provides more precision during intermediate calculations
3.  **Compatibility**: Works across all devices (CPU/GPU) and PyTorch versions

Test Plan:
UTs

Rollback Plan:

Differential Revision: D81711093

### DIFF
--- a/torch/distributed/checkpoint/quantized_hf_storage.py
+++ b/torch/distributed/checkpoint/quantized_hf_storage.py
@@ -123,6 +123,11 @@ class QuantizedHuggingFaceStorageReader(HuggingFaceStorageReader):
         Returns:
             Dequantized tensor
         """
+        # Convert to float32 for computation
+        # Certain quantized dtypes like Float8_e4m3fn
+        # don't support multiplication on CPU yet in PyTorch.
+        upcasted_weight = weight.to(torch.float32)
+
         # Get original dimensions
         orig_shape = weight.shape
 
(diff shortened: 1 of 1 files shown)


wwwjn (approved):

torch/distributed/checkpoint/quantized_hf_storage.py
```diff
@@ -131,7 +136,7 @@ def _dequantize_tensor(
         block_rows, block_cols = expected_scale_shape
 
         # Create output tensor in target dtype
-        dequantized = weight.detach().clone().to(dtype=self.target_dtype)
+        dequantized = upcasted_weight.detach().clone().to(dtype=self.target_dtype)
```
vadimkantorov: is it the same as `upcasted_weight.detach().to(dtype=self.target_dtype, copy=True)`?

saumishr: Yes. I believe so. In fact, I feel your suggestion may be more efficient. I will verify that and update the PR.

vadimkantorov: I also wonder if typically it would be faster to do `dequantized = weight.detach().to(dtype=self.target_dtype, copy=True)` - if `weight` is typically lower-width dtype than float32, it could spent less bandwidth/cycles reading input `weight` from VRAM (than reading `upcasted_weight` which float32)

saumishr: @vadimkantorov Great suggestion. We can probably work with the quantized weight here. For larger tensors, it may be more efficient. I will update the PR.

PR #162204, opened 2025-09-04
[dynamo] Add support for const prop on .item

angelayi:
Fixes some of the errors in https://fb.workplace.com/groups/1028545332188949/permalink/1303030824740397/

### DIFF
--- a/torch/_dynamo/variables/builder.py
+++ b/torch/_dynamo/variables/builder.py
@@ -2991,12 +2991,8 @@ def handle_traced_output(example_value, tx, proxy, options, subclass_type, targe
         return ConstantVariable.create(example_value, **options)
     elif isinstance(example_value, (int, float, bool)) and (
         proxy.node.target is call_torchbind
-    ):
-        set_example_value(proxy.node, example_value)
-        return ConstantVariable.create(example_value, **options)
-    elif (
-        isinstance(example_value, (int, float, bool))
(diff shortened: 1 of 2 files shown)


torch/_dynamo/variables/builder.py
```diff
@@ -2991,12 +2991,8 @@ def handle_traced_output(example_value, tx, proxy, options, subclass_type, targe
...
     elif isinstance(example_value, (int, float, bool)) and (
         proxy.node.target is call_torchbind
-    ):
-        set_example_value(proxy.node, example_value)
-        return ConstantVariable.create(example_value, **options)
-    elif (
-        isinstance(example_value, (int, float, bool))
-        and proxy.node.target is flat_apply
+        or proxy.node.target is flat_apply
+        or (proxy.node.op == "call_method" and proxy.node.target == "item")
```
williamwen42, line 2995: Is this the preferred way to check for `Tensor.item()`? I don't know if other cases could undesirably go through this code path.

angelayi: I think soo? like [this](https://github.com/pytorch/pytorch/blob/1ec2c15914da4ef7bd926ed9aebc8671c75fe965/torch/_dynamo/output_graph.py#L2340)

williamwen42 (approved):

PR #162207, opened 2025-09-04
[Graph Partition] interface for custom cg wrapper

BoyuanFeng:
This PR adds an interface to allow users to specify custom cudagraph wrapper. User example: [vllm](https://github.com/vllm-project/vllm/pull/24281)


torch/_inductor/config.py
```diff
@@ -460,6 +460,12 @@ def prologue_fusion_enabled() -> bool:
     == "1"
 )
 
+# A tuple of customized partition wrappers from users. Inductor mechanically wraps
+# each partition fn with the customized partition wrapper. When using for CUDAGraph wrapper,
+# users need to handle details such as static inputs, dynamic shapes, etc.
+customized_partition_wrappers: Optional[
+    tuple[Callable[..., Callable[..., Any]], ...]
+] = None
```
eellison: imo tuple is the wrong api. how do the users know how many to provide? Can we instead have an api that takes in a function and other metadata about the function - potentially partition index within function, graph index. 

It could be 

def wrap_cudagraph(fn, CudaGraphMeta)

so that we can add new fields without breaking bc.. 

Or at least mark this api as unstable.. 

Edit: misunderstood this as a tuple of callables, instead of a callable that takes in a tuple.. theres still some question of if we should pass over other metadata.

zou3519 (approved):

eellison (approved):
I think its worth passing metadata, or giving us an api that allow us to pass more information in the future without breaking bc

ProExpertProg:
I'm with @eellison on this one, I think it would be nice if the wrapper fn was a `lambda fn, metadata: ...` instead of a list of `lambda fn: ...`. If metadata included the index of the partition and total number of partitions, vLLM wouldn't have to guess the number of partitions ahead of time.

torch/_inductor/utils.py
```diff
@@ -3591,31 +3591,46 @@ def python_subprocess_env() -> dict[str, str]:
...
-    # metadata is unused for now
-    assert metadata is None
-    global _unstable_customized_partition_wrappers
 
-    _unstable_customized_partition_wrappers.clear()
-    _unstable_customized_partition_wrappers.extend(wrappers)
+    # The number of partitions that are cudagraphable.
+    num_partitions: int
+
+    # The number of partitions that are cudagraphable and have a fallback.
```
ProExpertProg: This comment seems wrong?

torch/_inductor/utils.py
```diff
@@ -3591,31 +3591,46 @@ def python_subprocess_env() -> dict[str, str]:
...
+
+PartitionFnType = Callable[..., Any]
+CUDAGraphWrapperType = Callable[
+    [PartitionFnType, CUDAGraphWrapperMetadata], PartitionFnType
+]
+
+
+# only incremented by user call of mark_step_begin
+class CUDAGraphWrapper:
+    wrapper: Optional[CUDAGraphWrapperType] = None
```
ProExpertProg, line 3618: Not sure why this is nested but I'm not opposed

ProExpertProg (approved):
This is nicer & should be easier to use :)

torch/_inductor/utils.py
BoyuanFeng: only for variable scope purpose

torch/_inductor/config.py
BoyuanFeng: Thanks for the great suggestion! I updated the interface to be `def wrapper(fn: PartitionFnType, metadata: CUDAGraphWrapperMetadata) -> PartitionFnType` which is more flexible.

PR #162209, opened 2025-09-04
MXFP8 grouped GEMM support for torch._scaled_grouped_mm + submodule bump

danielvegamyhre:
## Summary
- We just landed 2d-2d support for mxfp8 grouped gemm in FBGEMM: https://github.com/pytorch/FBGEMM/pull/4816 
- This is needed for backward pass of mxfp8 MoE training with grouped gemms
- Changes:
    - Add dispatching + input validation for mxfp8 grouped gemm in `torch._scaled_grouped_mm`
    - Add meta registration input validation for mxfp8 grouped gemm, for composability with compile
    - Add unit tests exercising torch._scaled_grouped_mm with mxfp8 inputs
    - Bump FBGEMM third party submodule to include:
          - https://github.com/pytorch/FBGEMM/pull/4816 
          - https://github.com/pytorch/FBGEMM/pull/4820
          - https://github.com/pytorch/FBGEMM/pull/4821
          - https://github.com/pytorch/FBGEMM/pull/4823

#### How fbgemm dependency was bumped
Documenting this since I haven't found it documented elsewhere:
- `cd ~/pytorch/third_party/fbgemm`
- `git fetch`
- `git checkout <hash>`
- `cd ~/pytorch`
- `git add third_party/fbgemm`

#### Test build 
```
USE_FBGEMM_GENAI=1 python -m pip install --no-build-isolation -v -e .
...
Successfully installed torch-2.9.0a0+gitf5070f3
```
[full build log](https://www.internalfb.com/phabricator/paste/view/P1933787581)

#### Unit tests
```
pytest test/test_matmul_cuda.py -k test_mxfp8_scaled_grouped_mm_
...

test/test_matmul_cuda.py .........                                                                                                                        [100%]

============================================================== 9 passed, 1668 deselected in 5.34s ===============================================================
```


aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1551,6 +1551,180 @@ _scaled_mm_out_cuda(const Tensor& mat1, const Tensor& mat2,
...
+        };
+
+        // For 2d-2d, groups are along K dim.
+        // Calculate total size of scale K dim after per group blocked format is applied
+        int64_t padded_per_group_scale_K = 0;
+        int64_t group_start_off = 0;
+        auto offs_accessor = offs.accessor<int64_t, 1>();
+
+        for (int i = 0; i < offs_accessor.size(0); ++i) {
+            int64_t group_end_off = offs_accessor[i];
```
ngimel: You can't check scales here, that would require d2h sync for offs. Like for inputs themselves, we cannot really check if scales tensor is fine, we can only make sure that it's not smaller than we expect in the setup_args kernel

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1551,6 +1551,180 @@ _scaled_mm_out_cuda(const Tensor& mat1, const Tensor& mat2,
...
+          return ((x + y - 1) / y) * y;
+        };
+
+        // For 2d-3d, groups are along M dim.
+        // Calculate total size of scale M dim after per group blocked format is applied
+        int64_t padded_per_group_scale_M = 0;
+        int64_t group_start_off = 0;
+        auto offs_accessor = offs.accessor<int64_t, 1>();
+
+        for (int i = 0; i < offs_accessor.size(0); ++i) {
```
ngimel: same comment as above, we have to bail on checking scales size

torch/_meta_registrations.py
```diff
@@ -7399,6 +7399,192 @@ def _create_grouped_mm_output_tensor(mat1, mat2, offs, out_dtype):
     return out
 
 
+def _meta_scaled_grouped_mm_fp8(
```
danielvegamyhre: Note for reviewers: I refactored _scaled_grouped_mm meta registrations to have separate branches for (1) fp8 rowwise, (2a) mxfp8 2d-2d, and (2b) mxfp8 2d-3d.

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1551,6 +1551,180 @@ _scaled_mm_out_cuda(const Tensor& mat1, const Tensor& mat2,
...
+        int64_t N = mat_b.size(-1);
+        int64_t padded_scale_N = round_up(N, (int64_t)128);
+        TORCH_CHECK(
+            scale_b.size(-2) == padded_scale_N && scale_b.size(-1) == padded_scale_K,
+            "For 2d-3d MXFP8 grouped gemm, for mat_b.shape=(", mat_b.size(0), ",", mat_b.size(1), ", ", mat_b.size(2),
+            "), expected scale_b to have shape (..., ", padded_scale_N, ", ",
+            padded_scale_K, "), but got (..., ", scale_b.size(-2), ", ", scale_b.size(-1), ").");
+    }
+
+  at::Tensor create_grouped_gemm_output_tensor(const Tensor& mat_a,
```
ngimel: how is this `create_grouped_gemm_output_tensor different from existing?

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1646,6 +1820,38 @@ bool use_fast_accum) {
     TORCH_CHECK(offs->dtype() == at::kInt, "Offsets have to be int32");
   }
 
+  Tensor out = create_grouped_gemm_output_tensor(mat_a, mat_b, offs, out_dtype);
+
+  #ifdef USE_FBGEMM_GENAI
+    // MXFP8 grouped GEMM dispatching
+    bool is_mx8m8bf16 = (
```
ngimel:
```suggestion
    bool is_mx8mx8bf16 = (
```

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1646,6 +1820,38 @@ bool use_fast_accum) {
...
+      out_dtype == at::kBFloat16
+    )
+
+    if (is_mx8m8bf16) {
+      bool b_is_3d = mat_b.dim() == 3;
+      bool is_2d_2d = a_is_2d && b_is_2d;
+      bool is_2d_3d = a_is_2d && b_is_3d;
+
+      if (is_2d_2d) {
+        check_scaled_grouped_mm_mxfp8_2d2d(mat_a, mat_b, scale_a, scale_b, offs);
```
ngimel: conceptually, what you are doing here is checking scale dimensions that you can. That's what's currently done in `check_scale` function, so you should move your checks there. Note also there's just a single `check_scale` function that hides the differences between 2d3d.

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1646,6 +1820,38 @@ bool use_fast_accum) {
     TORCH_CHECK(offs->dtype() == at::kInt, "Offsets have to be int32");
   }
 
+  Tensor out = create_grouped_gemm_output_tensor(mat_a, mat_b, offs, out_dtype);
```
ngimel: try creating output once all the checks have passed

torch/_meta_registrations.py
ngimel: the difference compared to existing scaled_grouped_mm is 
1) you expect different dtypes for inputs and scales
2) you expect different dimensions for scales (however you cannot test them beyond static dims and that dynamic dim is >= corresponding tensor dim)
So you should, both in c++ and here modify check_scale function to do what you need, but not do a lot of copy paste

aten/src/ATen/native/cuda/Blas.cpp
danielvegamyhre: Ah of course, thanks. Will remove.

aten/src/ATen/native/cuda/Blas.cpp
danielvegamyhre: This is still the same function, I didn't add a new one, not sure why diff is showing up this way

torch/_meta_registrations.py
danielvegamyhre: ah at compile time the offsets won't be known, so can't check in meta registration. and in C++ at dispatch time, it would require a d2h sync to check the offsets. how silly of me, updating it now.

aten/src/ATen/CMakeLists.txt
```diff
@@ -257,46 +257,75 @@ IF(USE_FBGEMM_GENAI)
...
+    # If you want to integrate a kernel from FBGEMM into torch, you have to add it here.
+    set(FBGEMM_CUTLASS_KERNELS_REGEX
+      ".*(f8f8bf16_rowwise).*"
+      ".*(mx8mx8bf16_grouped_mm).*"
+    )
+    file(GLOB_RECURSE fbgemm_genai_native_cuda_cu
+      "${FBGEMM_GENAI_SRCS}/cutlass_extensions/*.cu"
+      "${FBGEMM_GENAI_SRCS}/cutlass_extensions/**/*.cu")
+    list(FILTER fbgemm_genai_native_cuda_cu INCLUDE REGEX ${FBGEMM_CUTLASS_KERNELS_REGEX})
+    set(fbgemm_genai_native_cuda_cuh
```
drisspg: can we use the same regex?

aten/src/ATen/native/cuda/Blas.cpp
danielvegamyhre: Cleaned it up

aten/src/ATen/native/cuda/Blas.cpp
danielvegamyhre: Updated to simplify the checks and wrap them under `check_scale`, which was getting quite large so I extracted the groups of conditional checks into helpers.

torch/_meta_registrations.py
danielvegamyhre: Updated this accordingly. A couple things to clarify though:

> you expect different dtypes for inputs and scales

I believe input dtype should still be e4m3 for fp8 rowwise and mxfp8 - dtypes are only different for the scales (fp32 vs e8m0)

> you expect different dimensions for scales (however you cannot test them beyond static dims and that dynamic dim is >= corresponding tensor dim)

Is this always true? For 2d-3d the dynamic `total_m` dim will be >= M as every group will have the same rows or rounded up to next multiple of 128. However, for 2d-2d the dynamic `total_K` dim has scaling applied before rounding, so the scaled dynamic dim may be smaller. For example shape (128, 128) will have scales (128, 4) and no padding is needed during conversion to blocked format since cols is already a multiple of 4, so scale dynamic dim K is actually <= corresponding tensor dim. 

(Let me know if I've misunderstood either of your points)

Anyway we can check for the 2d-3d case above, but it would require a refactor given the current `check_scale` has no concept of what kind of grouped gemm (2d-2d, 2d-3d etc) the given scale is involved in.

As an aside, IMO as we add support for more dtypes, trying to capture scale checks in one generic function makes it a bit hard to express some of these things. I think splitting things out into explicit checks for each scaling type would be easier to understand and extend.

aten/src/ATen/CMakeLists.txt
danielvegamyhre: I think so? we'll see if it works once the WIP fix for the build issue in https://github.com/pytorch/FBGEMM/pull/4823 is landed

aten/src/ATen/CMakeLists.txt
```diff
@@ -257,46 +257,75 @@ IF(USE_FBGEMM_GENAI)
...
-
+  if(USE_CUDA)
+    add_library(fbgemm_genai OBJECT)
     set_target_properties(fbgemm_genai PROPERTIES POSITION_INDEPENDENT_CODE ON)
-    target_compile_definitions(fbgemm_genai PRIVATE FBGEMM_GENAI_NO_EXTENDED_SHAPES)
-
+    # To avoid increasing the build time/binary size unneccesarily, use an allow-list of kernels to build.
+    # If you want to integrate a kernel from FBGEMM into torch, you have to add it here.
+    set(FBGEMM_CUTLASS_KERNELS_REGEX
+      ".*(f8f8bf16_rowwise).*"
```
cthi: I think we don't want `f8f8bf16_rowwise` for now, only `mx8mx8bf16_grouped`?

danielvegamyhre: thanks, updated

torch/testing/_internal/common_cuda.py
```diff
@@ -129,9 +129,17 @@ def evaluate_platform_supports_mx_gemm():
             return SM100OrLater
     return False
 
+def evaluate_platform_supports_mxfp8_grouped_gemm():
+    if torch.cuda.is_available():
```
cthi: Should we also add a check for `if torch.version.hip:`, and return False? (similar to `evaluate_platform_supports_mx_gemm` above)

danielvegamyhre: good idea, done

torch/_meta_registrations.py
ngimel: Yeah you are correct for 2d-2d, you can check static dimension and you can check >= for dynamic_dim/16 or whatever (for rowwise scaling in 2d-2d case there was no dynamic dim). `check_scales` is still good imo, but internally it might have to dispatch to different helper fns? But it's not a very important decision.

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1584,10 +1584,65 @@ namespace {
           scale.size(1) == mat.size(1 + dim),
           "scale must have the same first dimension as mat for arg ",
           arg_idx);
+    } else {
```
ngimel: nit: you don't need this branch, matrices were checked earlier, but you can leave a comment that this function expects 2d or 3d matrix

ngimel (approved):
Nice, looks much better now!

test/test_matmul_cuda.py
```diff
@@ -1045,6 +1049,169 @@ def test_float8_scale(self, device) -> None:
...
+        # bf16 reference output
+        y_bf16 = torch._grouped_mm(
+            X, W.t(), offs=input_group_end_offsets, out_dtype=torch.bfloat16
+        )
+
+        # Assert no NaNs
+        assert not y_mxfp8.isnan().any(), "mxfp8 output contains NaN"
+
+        # Assert outputs are close
+        torch.testing.assert_close(y_mxfp8, y_bf16, atol=8.0e-2, rtol=8.0e-2)
```
eqy, line 1147: nit: `self.assertEqual`

torch/testing/_internal/common_cuda.py
```diff
@@ -129,9 +129,17 @@ def evaluate_platform_supports_mx_gemm():
...
+def evaluate_platform_supports_mxfp8_grouped_gemm():
+    if torch.cuda.is_available() and not torch.version.hip:
+        built_with_fbgemm_genai = "USE_FBGEMM_GENAI" in torch.__config__.show()
+        return built_with_fbgemm_genai and SM100OrLater
+    return False
+
 PLATFORM_SUPPORTS_MX_GEMM: bool = LazyVal(lambda: evaluate_platform_supports_mx_gemm())
 PLATFORM_SUPPORTS_FP8: bool = LazyVal(lambda: evaluate_platform_supports_fp8())
 PLATFORM_SUPPORTS_FP8_GROUPED_GEMM: bool = LazyVal(lambda: evaluate_platform_supports_fp8_grouped_gemm())
+PLATFORM_SUPPORTS_MX_GEMM: bool = LazyVal(lambda: TEST_CUDA and SM100OrLater)
```
eqy, line 142: Were SM110 (Thor) and SM120 (RTX Blackwell) also tested?

cthi: We've only been able to test the kernel on GB200 and B200. Im not familiar with 110 and 120 platforms, should we restrict only to 100 then?

test/test_matmul_cuda.py
danielvegamyhre: i actually found I cannot use this, most elements do not match within the threshold, I need to explicitly specify my own thresholds here.

torch/testing/_internal/common_cuda.py
danielvegamyhre: Updated to restrict to only sm100

aten/src/ATen/CMakeLists.txt
```diff
@@ -255,48 +255,82 @@ endif()
...
+    message(STATUS "fbgemm_genai_native_cuda_cpp: ${fbgemm_genai_native_cuda_cpp}")
+
+    # Combine all source files into a single list
+    list(APPEND fbgemm_genai_all_sources
+      ${fbgemm_genai_native_cuda_cu}
+      ${fbgemm_genai_native_cuda_cpp}
+    )
+
+    # Now, create the library and provide the sources at the same time
+    message(STATUS "FBGEMM_GENAI_SRCS directory: ${PROJECT_SOURCE_DIR}/third_party/fbgemm/fbgemm_gpu/experimental/gen_ai/src/quantize")
```
drisspg: remove these

danielvegamyhre: ah thanks, removed.

test/test_matmul_cuda.py
vkuzo: hmm, can we test vs mxfp8 `torch._scaled_mm` with bitwise equality for the individual gemm slices, instead of a less precise check vs bf16?  Can be a future PR.

torch/testing/_internal/common_quantized.py
```diff
@@ -479,3 +479,110 @@ def to_blocked(input_matrix) -> torch.Tensor:
     rearranged = blocks.reshape(-1, 4, 32, 4).transpose(1, 2).reshape(-1, 32, 16)
 
     return rearranged.flatten()
+
+# This function is extracted from https://github.com/pytorch/ao/blob/v0.12.0/torchao/prototype/mx_formats/mx_tensor.py#L142
+def to_mxfp8(
```
vkuzo, line 484: we already have this logic here (https://github.com/pytorch/pytorch/blob/081cab045472ce045634548cc6c14a4870641e23/test/test_matmul_cuda.py#L931), want to update that with RCEIL instead? Can be in a future PR.

drisspg (commented):
This won't be set in nightly builds, I think this okay just to land this but we will need another pr in order to actually build and ship these kernels

danielvegamyhre:
> This won't be set in nightly builds, I think this okay just to land this but we will need another pr in order to actually build and ship these kernels

What won't be set, TORCH_CUDA_ARCH_LIST?

PR #162212, opened 2025-09-04
remove gso from collapse_view_helper

laithsakka:
* __->__ #162212


torch/_prims/__init__.py
```diff
@@ -1406,50 +1411,68 @@ def _collapse_view_helper(
...
+    else:
+        if not guard_or_false(valid_op):
+            return None, None
+
+    # compute stride which is the min dimension in the collapsed range.
     for idx in range(end - 1, start - 1, -1):
-        if guard_size_oblivious(shape[idx] == 0) or guard_size_oblivious(
-            shape[idx + 1] == 0
-        ):
+        stride = min(stride, strides[idx])
```
aorenste: Isn't this just:
```
stride = min(strides[start:end+1])
```

torch/_prims/__init__.py
```diff
@@ -1406,50 +1411,68 @@ def _collapse_view_helper(
...
+
+    # compute length
+    for idx in range(end - 1, start - 1, -1):
+        # those are just show circuits, mm except
```
aorenste: huh?

torch/_prims/__init__.py
```diff
@@ -1406,50 +1411,68 @@ def _collapse_view_helper(
...
+        if guard_or_false(shape[idx] == 0) or guard_or_false(shape[idx + 1] == 0):
```
aorenste: The only reason for checking idx+1 is to catch [end], right? Maybe above initialize length via:
```
length = 0 if guard_or_false(shape[end] == 0) else shape[end]
```
and then this only has to check shape[idx]...

torch/_prims/__init__.py
```diff
@@ -1406,50 +1411,68 @@ def _collapse_view_helper(
...
-
-    if new_shape is None:
-        msg = "Attempting to view a collapsed tensor, but no such view exists!"
-        raise ValueError(msg)
-
+    new_shape, new_strides = _collapse_view_helper(
+        a, start, end, "Attempting to view a collapsed tensor, but no such view exists!"
+    )
     assert new_strides is not None
+    assert new_shape is not None
```
aorenste, line 1479: Do we need to worry about this being non-BC?

torch/_prims/__init__.py
```diff
@@ -1384,12 +1384,17 @@ def _collapsed_shape(shape: ShapeType, start: int, end: int) -> tuple[int, ...]:
     return shape[0:start] + (dim_length,) + shape[end + 1 :]
 
 
+# If must_be_valid is not None, do a torch._check that the collapse is valid else
+# return None, None.
+# When must_be_valid is False, this function returns None when validity is not decided.
+# When must_be_valid is True, this function generates a runtime assertion for validity
+# when it is not decided.
```
aorenste: This needs to be better - I had trouble trying to figure out why it was Optional[str] and there was no behavior change from False to None.
Consider something like:
```suggestion
# If the validity of the collapse cannot be determined (because of unbacked data) then `must_be_valid` determines the behavior:
#     None: return None, None.
#     str: Do a torch._check() to ensure the collapse is valid and if it isn't then fail with the provided string.
```

torch/_prims/__init__.py
```diff
@@ -1384,12 +1384,17 @@ def _collapsed_shape(shape: ShapeType, start: int, end: int) -> tuple[int, ...]:
...
 def _collapse_view_helper(
-    a: TensorLikeType, start: int, end: int
+    a: TensorLikeType, start: int, end: int, must_be_valid: Optional[str]
```
aorenste, line 1393: Is there a case where this function can return (None, None) that isn't one where must_be_valid would have triggered? i.e. Why add must_be_valid instead of just checking for None at the caller?

torch/_prims/__init__.py
```diff
@@ -1406,50 +1411,68 @@ def _collapse_view_helper(
...
+                valid_op,
+                sym_or(
+                    shape[idx] == 1,
+                    shape[idx + 1] == 1,
+                    strides[idx] == strides[idx + 1] * shape[idx + 1],
+                ),
+            )  # type: ignore[assignment]
+
+            # early exit if we already know its invalid.
+            if guard_or_false(valid_op is False):
```
aorenste, line 1430: nit?
```
guard_or_false(not valid_op)
```
(`is False` is a bit unusual)

torch/_prims/__init__.py
```diff
@@ -1406,50 +1411,68 @@ def _collapse_view_helper(
...
+            valid_op = sym_and(
+                valid_op,
+                sym_or(
+                    shape[idx] == 1,
+                    shape[idx + 1] == 1,
+                    strides[idx] == strides[idx + 1] * shape[idx + 1],
+                ),
+            )  # type: ignore[assignment]
+
+            # early exit if we already know its invalid.
```
aorenste, line 1429: super nit:
```suggestion
            # early exit if we already know it's invalid.
```

torch/_prims/__init__.py
laithsakka: right wonder if that preserves the symbolic representation of the output i can try it .

torch/_prims/__init__.py
laithsakka: haha no mm except

torch/_prims/__init__.py
laithsakka: this is existing code not sure what was its value.. i guess just short circuiting to exit the loop early but maybe we shall just remove it lol
how is this 
```
length = 0 if guard_or_false(shape[end] == 0) else shape[end]
```
different than just which we have now.
```
length = shape[end]
```

torch/_prims/__init__.py
laithsakka: you mean worry about raising RuntimeError instead of ValueError?

torch/_prims/__init__.py
laithsakka: reason i am doing this is this
1) if i do ==False I get linter error that i should use is False
2) if i use sym_not(valid_op) I get this issue https://github.com/pytorch/pytorch/issues/162287

torch/_prims/__init__.py
laithsakka: because we want the sympy expressions that we should convert to runtime assertions. 
the alternative to not passing must_be_valid is returning a 3rd output being valid_op variable when we return None, None,
make sense?

torch/_prims/__init__.py
aorenste: Yes

torch/_prims/__init__.py
aorenste: If you change the existing code to only check shape[idx] then if shape[end] is zero then it won't be set to the constant 0. So you need to initialize it properly.
Basically:
```
if guard_or_false(shape[idx] == 0) or guard_or_false(shape[idx + 1] == 0):
    length = 0
    break
```
is the same as:
```
if guard_or_false(shape[idx + 1] == 0):
    length = 0
    break
if guard_or_false(shape[idx] == 0):
    length = 0
    break
```
But that first test is being performed twice on every element (once for [idx] and once for [idx+1]) EXCEPT on the very first element (when idx+1 == end)

laithsakka: right but we already have 
length = shape[end]
maybe what you say is we shall do 
```
length = shape[end]
if guard_or_true(length!=0):
  if guard_or_false(shape[idx] == 0): 
       length = 0 
       break
    length = length * shape[idx]
 ```

File: torch/_prims/__init__.py (hunk 4 above)
Comment (laithsakka) [reply torch/_prims/__init__.py:1479]: I think its ok, 
i do not think usrs would have a failing case as expected. 
mm if we want to be safe we can do something like if 
```
// when it fails
if valid_op.has_hint():
  if not valid_op:
     throw value error
else:
      torch._check(valid_op,,,,)
 ```

laithsakka: but i feel its fine... IMO. 
what do you think?

torch/_prims/__init__.py
aorenste: What you're suggesting is changing the behavior (it guards length - maybe we want that - but what I'm suggesting is just a rewrite of the existing behavior).
I'm saying that this matches the existing behavior but is simpler (it skips the extra "`guard_or_false(shape[idx+1] == 0)`" check on each iteration.
```
    # compute length
    length = 0 if guard_or_false(shape[end] == 0) else shape[end]
    for idx in range(end - 1, start - 1, -1):
        # those are just show circuits, mm except
        if guard_or_false(shape[idx] == 0):
            length = 0
            break
```

torch/_prims/__init__.py
aorenste: If you're okay with it then I'm okay with it. I just wanted to point it out and make sure we're expecting the change.

torch/_prims/__init__.py
laithsakka: the guarding behavior is not intentional those checks were there to early exist as perf opt and its not ended to guard but that said guard_or_true(length!=0) would still guard on length value

aorenste (approved):

torch/_prims/__init__.py
```diff
@@ -1404,52 +1414,69 @@ def _collapse_view_helper(
...
-        if guard_size_oblivious(shape[idx] == 0) or guard_size_oblivious(
-            shape[idx + 1] == 0
-        ):
-            length = 0
-            stride = 0
-            break
-
-        if guard_size_oblivious(shape[idx] == 1):
-            continue
+        if shape[idx] != 1:
```
laithsakka, line 1445: I kept this throwing DDE on purpose if we ever hit i will priotrize fixing it

PR #162214, opened 2025-09-04
[DCP][Quantization] Fix the issue when scale vector is in a different SafeTensors file

saumishr:
Summary: The current dequantization implementation assumes that the weight and scale tenors are in the same SafeTensors files. This diff fixes the issue to support the case when these could be in different files.

Test Plan:
buck test fbcode//caffe2/test/distributed/checkpoint\:test_quantized_hf_storage

Buck UI: https://www.internalfb.com/buck2/532bf151-bb40-41fd-b080-ff898675afe2
Test UI: https://www.internalfb.com/intern/testinfra/testrun/15199648851011082

Rollback Plan:

Differential Revision: D81718598

### DIFF
--- a/test/distributed/checkpoint/test_quantized_hf_storage.py
+++ b/test/distributed/checkpoint/test_quantized_hf_storage.py
@@ -1,7 +1,7 @@
 # Owner(s): ["oncall: distributed checkpointing"]
 
 import tempfile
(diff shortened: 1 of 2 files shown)


wwwjn (approved):
Nice fix

PR #162216, opened 2025-09-04
support unbacked softmax / logsoftmax

ColinPeppler:
### DDE

```
GuardOnDataDependentSymNode: Could not guard on data-dependent expression Eq(3*u0, 0) (unhinted: Eq(3*u0, 0)).  (Size-like symbols: u0)

Caused by: (_decomp/decompositions.py:1185 in _softmax)
```

```
torch._dynamo.exc.UserError: Could not guard on data-dependent expression Eq(u0, 0) (unhinted: Eq(u0, 0)).  (Size-like symbols: u0)

Caused by: logsoft = torch.nn.functional.log_softmax(nz, dim=0)  # test/inductor/test_unbacked_symints.py:573 in fn (_decomp/decompositions.py:1212 in _log_softmax)
```

```
GuardOnDataDependentSymNode: Could not guard on data-dependent expression Ne(u0, 0) (unhinted: Ne(u0, 0)).  (Size-like symbols: u0)

Caused by: (_refs/__init__.py:2218 in _reduction)
```

### Cannot convert symbols to int
```
  File "torch/_inductor/lowering.py", line 7160, in prepare_softmax_online
    and V.graph.sizevars.size_hint(rnumel) >= config.unroll_reductions_threshold
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "orch/_inductor/sizevars.py", line 591, in size_hint
    return int(out)
           ^^^^^^^^
  File "sympy/core/expr.py", line 342, in __int__
    raise TypeError("Cannot convert symbols to int")
```

* __->__ #162216


torch/_refs/__init__.py
```diff
@@ -2227,11 +2227,21 @@ def _reduction(
         dims = (dims,)  # type: ignore[assignment]
     dims = utils.reduction_dims(a.shape, dims)
     if not has_identity:
-        valid_shape = a.ndim == 0 or builtins.all(a.shape[i] for i in dims)
+        from torch.fx.experimental.symbolic_shapes import guard_or_true
+
+        valid_shape = a.ndim == 0 or builtins.all(
```
laithsakka: here the right thing to do is to create a symbolic big expression and pass it to torch._check(..)
instead of suing guard_or_true. 
see valid_op in this PR for example
https://github.com/pytorch/pytorch/pull/162212

laithsakka (approved):

torch/_inductor/lowering.py
```diff
@@ -7197,7 +7197,8 @@ def prepare_softmax_online(x, dim):
 
     if (
         num_split == 1
-        and V.graph.sizevars.size_hint(rnumel) >= config.unroll_reductions_threshold
```
eellison, line 7200: should this be statically known true? cc @shunting314

eellison (approved):

torch/_inductor/lowering.py
ColinPeppler: actually, `statically_known_true(rnumel ≥ threshold)` might work better I think.

I do know that `statically_known_gte(rnumel, threshold)` will return false if `rnumel.upperbound ≥ threshold`, whereas, `size_hint(rnume) ≥ threshold` would pass.

shunting314: yea, skipping online softmax when rnumel is dynmaic shape looks better

PR #162219, opened 2025-09-04
Overload _get_operation_for_overload_or_packet & friends to accept ArrayRef

swolchok:
* #162220
* __->__ #162219
* #161692
* #161634
* #161633
* #161595
* #161591

Avoids requiring vector allocation to call this.


torch/csrc/jit/python/init.cpp
```diff
@@ -1735,7 +1735,7 @@ void initJITBindings(PyObject* module) {
                                        const py::kwargs& kwargs) {
                     ToIValueAllowNumbersAsTensors g(allow_numbers_as_tensors);
                     return _get_operation_for_overload_or_packet(
-                        {op}, symbol, args, kwargs, /*is_overload*/ true, dk_);
+                        op, symbol, args, kwargs, /*is_overload*/ true, dk_);
                   });
               return py::make_tuple(
                   func, func_dk, py::cast(op->getTags().vec()));
```
Skylion007, line 1741: These values should be std::move to prevent ref count increases

swolchok: I don't see anything movable here; if we move `op` from the lambda capture then the lambda won't work if called a second time.

Skylion007: This just moves the Python reference to the cpp_function pybind11 object. Isn't that reference created every time? 

This just avoid a pointer inc / decrease of the Python reference counter, it shouldn't move the function inside of the Python function since it's effectively a shared ptr, right?

Skylion007: To clarify, suggesting moving func and func_dk

Skylion007 (approved):

PR #162225, opened 2025-09-04
[ONNX] Hide draft export under a flag

justinchuby:
Use `TORCH_ONNX_ENABLE_DRAFT_EXPORT` to control whether draft_export should be used as a strategy in onnx export.

Follow up of https://github.com/pytorch/pytorch/pull/161454

cc @titaiwangms

### DIFF
--- a/torch/onnx/_flags.py
+++ b/torch/onnx/_flags.py
@@ -43,8 +43,8 @@ def _load_boolean_flag(
     return state
 
 
-PLACEHOLDER: bool = _load_boolean_flag(
-    "TORCH_ONNX_PLACEHOLDER",
-    this_will="do nothing",
-    default=True,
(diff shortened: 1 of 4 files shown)


xadupre (approved):

torch/onnx/_flags.py
```diff
@@ -43,8 +43,8 @@ def _load_boolean_flag(
     return state
 
 
-PLACEHOLDER: bool = _load_boolean_flag(
```
titaiwangms, line 46: What was this lol?

titaiwangms (approved):

torch/onnx/_flags.py
justinchuby: It was a place holder so we can quickly change it to a flag like this one. (previously all flags were removed)

PR #162243, opened 2025-09-05
[SymmMem] Tiled reduce

kwen2501:
* #164757
* __->__ #162243

Added op: `tile_reduce(Tensor input, Tensor(a!) out, int root, str group_name)`

For now supports only:
- NVSHMEM backed symmetric tensor;
- 2D tensor and tile;
- torch.float.

Testing on right-bottom quandrant:
```
rank 0:
tensor([[0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 0., 0., 0., 0.],
        [0., 0., 0., 0., 1., 1., 1., 1.],
        [0., 0., 0., 0., 1., 1., 1., 1.],
        [0., 0., 0., 0., 1., 1., 1., 1.],
        [0., 0., 0., 0., 1., 1., 1., 1.]], device='cuda:0')
PASSED                                                                                             
```


torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
```diff
@@ -842,6 +842,71 @@ void all_to_all_vdev_2d_offset(
...
+  c10::cuda::CUDAGuard guard(input.device());
+  auto hdl = c10d::symmetric_memory::rendezvous(input, group_name);
+  c10d::symmetric_memory::rendezvous(out, group_name);
+  nvshmem_team_t team = group_to_team(group_name, hdl->get_rank_to_global_rank());
+  TORCH_CHECK(root < nvshmem_team_n_pes(team), "root must be smaller than group size");
+  auto stream = at::cuda::getCurrentCUDAStream();
+
+  auto shape = nvshmemx::make_shape(input.sizes()[0], input.sizes()[1]);
+  auto stride = nvshmemx::make_stride(input.strides()[0], input.strides()[1]);
+  void* src_ptr = input.data_ptr();
```
Skylion007: should be const_data_ptr right? data_ptr is deprecated

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
```diff
@@ -842,6 +842,71 @@ void all_to_all_vdev_2d_offset(
...
+#ifndef _NVSHMEM_DEVICELIB_SUPPORTED
+  CUDA_KERNEL_ASSERT_MSG(false, "SM arch unsupported for NVSHMEM");
+#else
+  auto layout = nvshmemx::make_layout(shape, strides);
+  auto src_tensor = nvshmemx::Tensor(src_ptr, layout);
+  auto dst_tensor = nvshmemx::Tensor(dst_ptr, layout);
+  // src_tensor and dst_tensor are already the tiles to operate on, thus we set
+  // the start_coord to 0
+  auto start_coord = nvshmemx::make_shape(0, 0);
+  nvshmemx::tile_sum_reduce<decltype(src_tensor), decltype(dst_tensor), Shape2D, nvshmemx::tile_coll_algo_t::NVLS_ONE_SHOT_PULL_NBI>(
```
ngimel: One-shot algorithms are ok only for small sizes, for larger sizes they result in 4x more network traffic for 8 world size

ngimel (commented):
Can you give a tl;dr how you expect multiple teams to improve perf?

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
ngimel: The docs say `The users are expected to use tile_collective_wait routine to ensure completion of the non-blocking collectives.`, I don't see it here

kwen2501: That's the limitation of NVSHMEM today, only three algorithms are available:
```
tile_coll_algo_t::NVLS_ONE_SHOT_PUSH_NBI
tile_coll_algo_t::NVLS_ONE_SHOT_PULL_NBI
tile_coll_algo_t::NVLS_TWO_SHOT_PUSH_NBI
```
And I don't think TWO_SHOT would work for reduce. 

One-shot reduce (not all-reduce) will not create extra traffic, but it would indeed create a hot-spot at the root GPU.

So this relates to whether the collective has access to intermediate buffers or not:
- if not, reduce can do only one-shot, thus hot-spot and never bandwidth optimal;
- if yes, then algorithms like ring is possible, thus bandwidth optimal.

ngimel: if (for some not very small sizes) TWO_SHOT allreduce is faster than one-shot reduce we should be using it? At a higher level, what are you trying to achieve? Is it inter-node or intra-node?

kwen2501: TWO_SHOT allreduce will modify non-root ranks' buffer, so not so much a 1:1 in terms of semantics.

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
```diff
@@ -863,6 +864,128 @@ void all_to_all_vdev_2d_offset(
...
+    at::Tensor& in_tile,
+    at::Tensor& out_tile,
+    int64_t root,
+    std::string group_name,
+    std::string reduce_op) {
+  /* Perform a tile reduce operation on the input tensor, with the root rank
+   * receiving the reduced tensor. */
+  TORCH_CHECK(reduce_op == "sum", "tile_reduce: only sum is supported for now");
+  TORCH_CHECK(in_tile.dim() == 2 && out_tile.dim() == 2, "Only 2D tensors are supported");
+  TORCH_CHECK_EQ(in_tile.dtype(), out_tile.dtype());
```
ngimel, line 939: add user-friendly error messages here

ngimel (approved):

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
kwen2501: I had a look at the macro expansion of `TORCH_CHECK_EQ`, looks okay friendly? 
https://github.com/pytorch/pytorch/blob/6861a270624b44954826688f8dad668eb0154452/c10/util/logging_is_not_google_glog.h#L139-L141

PR #161768, opened 2025-08-29
[inductor][triton] support JITCallable._hash_lock

davidberard98:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #159158
* __->__ #161768

Fixes #161618

Triton # 7974 introduces a threading.RLock() in JITCallable, which is not pickle-able. This PR adds this field to the list of un-pickleable fields that need to be handled specially.

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @mlazos

PR #162244, opened 2025-09-05
[inductor][triton] more JITCallable._hash_lock support

davidberard98:
* #162278
* #162309
* __->__ #162244

Follow-up to #161768.

Context: ProcessPool pickles the outputs before sending them back to the main process. Triton kernels have some un-pickleable fields, so `prepare_for_pickle()` is used to strip out those fields. Previously, in the standard case (without triton_bundler.py), `prepare_for_pickle()` would strip out the un-pickleable fields and they would never be added back after unpickling, because the un-pickleable fields were not actually needed after compilation finished.

In #161768 updated `prepare_for_pickle` to also strip out the `fn._hash_lock` field, a newly added field in JITCallable instances which is a `threading.RLock()`, which is not pickleable.

It turns out that we do need to restore the `fn._hash_lock` field, even in the non-triton_bundler case - the MultiKernel case uses the hash lock.

To do this, we add `restore_after_unpickle()` which will restore fields (or if the old fields are not provided, initialize just the hash_lock)

Compile time benchmarks look good, maybe a very minor regression (see the comment below on the PR)


davidberard98:
# Perf dashboard results

https://hud.pytorch.org/benchmark/compilers?dashboard=torchinductor&startTime=Fri%2C%2029%20Aug%202025%2013%3A44%3A26%20GMT&stopTime=Fri%2C%2005%20Sep%202025%2013%3A44%3A26%20GMT&granularity=hour&mode=inference&dtype=bfloat16&deviceName=cuda%20(a100)&lBranch=gh/davidberard98/406/orig&lCommit=d9a0e30b883a149e846d74143c27033e6693401c&rBranch=main&rCommit=5c473e9f5ee0ef0fc38e6cf34a95b547f8cdc8d5

[WIP: analysis]

atalman (approved):

PR #162245, opened 2025-09-05
Improve device info with new flops and bandwidth formula based on hardware libraries

exclamaforte:
Previously, DeviceInfo provided theoretical hardware information based on a hardcoded list manually created from various datasheets. 

This update:
- Attempting to gather the information from a hardware library like `pynvml`, improving accuracy and expanding support to devices that don't have entries in the datasheet list.
- Adjusts flops and bw calculation based on these hardware values. For example, if the the memory or SMs are underclocked, it adjusts the theoretical max flops/bw accordingly.

### DIFF
--- a/torch/_inductor/analysis/device_info.py
+++ b/torch/_inductor/analysis/device_info.py
@@ -44,8 +44,6 @@ def _device_library_context(
     try:
         getattr(library, init_method)()
         yield library
-    except Exception:
-        raise
     finally:
         try:
             getattr(library, shutdown_method)()
@@ -132,7 +130,9 @@ class DeviceInfo:
         try:
             with _nvml_context() as pynvml:
                 handle = pynvml.nvmlDeviceGetHandleByIndex(0)
-                clock_mhz = pynvml.nvmlDeviceGetMaxClockInfo(handle, pynvml.NVML_CLOCK_SM)
+                clock_mhz = pynvml.nvmlDeviceGetMaxClockInfo(
+                    handle, pynvml.NVML_CLOCK_SM
+                )
                 return clock_mhz * 1e6
         except Exception:
             return None
@@ -177,7 +177,11 @@ class DeviceInfo:
                 mem_clock_info = amd_smi.amdsmi_get_clock_info(
                     device_handle, amd_smi.AmdSmiClkType.MEM
                 )
-                return mem_clock_info["max_clk"] * 1e6 if "max_clk" in mem_clock_info else None
+                return (
+                    mem_clock_info["max_clk"] * 1e6
+                    if "max_clk" in mem_clock_info
+                    else None
+                )
         except Exception as e:
             log.info("Failed to get AMD memory clock frequency: %s", e)
             return None
--- a/torch/_inductor/analysis/device_info.py
+++ b/torch/_inductor/analysis/device_info.py
@@ -10,53 +10,14 @@ import torch
 log = logging.getLogger(__name__)
 
 
-# Lazy import cache for pynvml
-_pynvml_cache: Optional[Any] = None
-_pynvml_initialized: bool = False
-
-# Lazy import cache for AMD ROCm SMI library
-_amd_smi_cache: Optional[Any] = None
-_amd_smi_initialized: bool = False
-
-
 def _get_pynvml() -> Optional[Any]:
-    """
-    Lazy import handler for pynvml.
-    Imports pynvml once on first call and caches the result.
-    Returns None if pynvml is not available.
-    """
-    global _pynvml_cache, _pynvml_initialized
-
-    if not _pynvml_initialized:
-        try:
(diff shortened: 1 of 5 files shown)


torch/_inductor/analysis/device_info.py
```diff
@@ -1,39 +1,374 @@
...
+    def lookup_memory_clock_hz(device_name: str) -> Optional[float]:
+        """Get the memory clock speed in Hz for the current device."""
+        result = DeviceInfo._generic_lookup(device_name, "memory_clock_hz")
+        return result if isinstance(result, (int, float)) or result is None else None
+
+    @staticmethod
+    def lookup_dram_gb(device_name: str) -> Optional[float]:
+        """Get the DRAM memory size in GB for the current device."""
+        result = DeviceInfo._generic_lookup(device_name, "dram_gb")
+        return result if isinstance(result, (int, float)) or result is None else None
```
shunting314: whey need this rather than 'return result' directly?  Same for multiple other similar methods.

torch/_inductor/analysis/device_info.py
```diff
@@ -1,39 +1,374 @@
...
+        if amd_smi is None:
+            return None
+
+        try:
+            amd_smi.amdsmi_init()
+            device_handle = amd_smi.amdsmi_get_processor_handles()[0]
+            mem_clock_info = amd_smi.amdsmi_get_clock_info(
+                device_handle, amd_smi.AmdSmiClkType.MEM
+            )
+            amd_smi.amdsmi_shut_down()
```
shunting314: Create a common function to lookup hw spec via amd smi.

Same for pynvml case.

torch/_inductor/analysis/device_info.py
```diff
@@ -1,39 +1,374 @@
...
+from typing import Any, Optional, Union
 
 import torch
 
 
 log = logging.getLogger(__name__)
 
 
+# Lazy import cache for pynvml
+_pynvml_cache: Optional[Any] = None
```
v0i0: torch._cuda._HAS_PYNVML etc seem to already do this for you, why not use that?

torch/_inductor/analysis/device_info.py
```diff
@@ -1,39 +1,374 @@
...
+        """Get the number of streaming multiprocessors from the hardware."""
+        try:
+            # rely on device_properties api
+            device_props = torch.cuda.get_device_properties(0)
+            return device_props.multi_processor_count
+        except Exception:
+            return None
+
+    @staticmethod
+    def _hardware_lookup_clock_hz() -> Optional[float]:
```
v0i0, line 124: why are we using nvml vs a device query for that? is the idea that this might capture current frequency (as opposed to the configured application clock)?

torch/_inductor/analysis/device_info.py
```diff
@@ -1,39 +1,374 @@
...
 # Indexing is based on `torch.cuda.get_device_name()`
 # TODO investigate profiler support for tf32 and allow device to report correct number when it's turned on.
-_device_mapping: dict[str, DeviceInfo] = {
+_device_mapping: dict[str, DeviceSpec] = {
     # Source:
     # @lint-ignore https://www.nvidia.com/en-us/data-center/h100/
-    "NVIDIA H100": DeviceInfo(
+    # These are from H100 SXM.
+    #
+    "NVIDIA H100": DeviceSpec(
```
v0i0, line 362: i wonder what it would take (in terms of making scaling these up top e.g. by SM in addition to frequency) to support the entirety of the Hopper device "zoo"

v0i0 (approved):
looks great, left some nits / questions

torch/_inductor/analysis/device_info.py
exclamaforte: I changed it to nvmlDeviceGetMaxClockInfo, I think that's the best bet

torch/_inductor/analysis/device_info.py
exclamaforte: it would be pretty easy refactor because you can query sm count, I'll do a followup PR.

shunting314 (approved):

eellison (commented):
deferring to @v0i0 and @shunting314 but great to see this change!

torch/_inductor/analysis/device_info.py
```diff
@@ -171,6 +526,8 @@ def datasheet_tops(dtype: torch.dtype, is_tf32: bool = False) -> Optional[float]
     Get the theoretical TFLOPS of the device for a given dtype. This can throw an exception if the device
     is not in the datasheet list above.
     """
+    if not torch.cuda.is_available():
```
xmfan: is this check needed? _device_mapping doesn't contain non nvidia/amd, and if it did shouldn't we still return their entry

PR #162247, opened 2025-09-05
[dynamo] Use relaxed CLOSURE_MATCH guard then ID_MATCH

anijain2305:
* __->__ #162247

I am unable to write a test that would fail here. The reason is that when we do _dynamo.disable(fn) in the compiled frame, the id of disabled function changes but currently we guard on the original function - `fn` whose id is not changing. This PR still guards on the `fn.__code__` just to be more precise.

Thanks to @thenumberouscode for pointing this out.

### DIFF
--- a/torch/_dynamo/variables/functions.py
+++ b/torch/_dynamo/variables/functions.py
@@ -1483,7 +1483,7 @@ class SkipFunctionVariable(VariableTracker):
                     guard_on_source, "_torchdynamo_orig_callable"
                 )
 
-            guard_on_source.make_guard(GuardBuilder.FUNCTION_MATCH)
+            guard_on_source.make_guard(GuardBuilder.CLOSURE_MATCH)
         elif not is_wrapper_or_member_descriptor(value):
             # These descriptors are not guaranteed to return the same object on
             # attribute lookup. They are unlikely to be changed, so we can skip


StrongerXi (approved):
> The reason is that when we do _dynamo.disable(fn) in the compiled frame, the id of disabled function changes but currently we guard on the original function - fn whose id is not changing. 

Doesn't this mean we can write a test that checks against recompilation, for the scenario you described here?

jansel (approved):

torch/_dynamo/variables/functions.py
```diff
@@ -1483,7 +1483,7 @@ def create_with_source(cls, value, source):
                     guard_on_source, "_torchdynamo_orig_callable"
                 )
 
-            guard_on_source.make_guard(GuardBuilder.FUNCTION_MATCH)
+            guard_on_source.make_guard(GuardBuilder.CLOSURE_MATCH)
```
thenumberouscode, line 1486: I noticed that line 1491 already has a CLOSURE_MATCH inserted. Should that be replaced with FUNCTION_MATCH instead?

anijain2305: Closure match is fine there. I don't see any reason where we would to guard on the id of the function and not the code of the function.

As a follow up, I will go and check if function match can be removed from the codebase.

PR #162253, opened 2025-09-05
[Build] CMake: a custom job pool for flash_attention

Aidyn-A:
The flash attention kernels are resource hungry on compilation, and because of that some users have to decrease `MAX_JOBS`, that causes other jobs be unnecessary slow. This PR introduces a job pool for for limiting number of processes for flash attention target only. The number of parallel jobs for flash attention is controlled via `FA_MAX_JOBS`, so one can do `MAX_JOBS=16 FA_MAX_JOBS=8 python3 setup.py ...` instead of the old way `MAX_JOBS=8 python3 setup.py ...`. This way, the build becomes significantly faster, allowing full parallelization for other targets.


aten/src/ATen/CMakeLists.txt
```diff
@@ -232,6 +232,14 @@ if(USE_CUDA AND (USE_FLASH_ATTENTION OR USE_MEM_EFF_ATTENTION))
   )
 
   set_target_properties(flash_attention PROPERTIES POSITION_INDEPENDENT_CODE ON)
+
+  if(DEFINED ENV{FA_MAX_JOBS} AND CMAKE_GENERATOR STREQUAL "Ninja")
```
Skylion007: Should we throw a warning here if FA_MAX_JOBS is set but generator is not Ninja? Seems like a footgun without some logging.

PR #162254, opened 2025-09-05
Fix: Remove incorrect non-negative validation for correction parameter in torch.var during export

### LINKED ISSUE #161083: torch.export fails for var when correction is negative
### 🐛 Describe the bug

```python
import torch
import torch.nn as nn

class VarModel(nn.Module):
    def __init__(self):
        super(VarModel, self).__init__()

    def forward(self, x):
        return torch.var(x, correction=-1)

model = VarModel()
x = torch.ones([3])

# this succeeds
model(x)

# this fails
# ValueError: correction argument should be non-negative
torch.export.export(model, (x,))
```

### Versions

nightly

parsshar-RH:
Fixes #161083

### Summary:
This PR fixes a bug where `torch.export.export `incorrectly validates that the `correction` parameter for `torch.var` should be non-negative, causing export to fail when using negative correction values that are perfectly valid and work correctly in eager PyTorch execution.

### Fix Impact:

- **Improved Consistency**: Aligns export behavior with eager PyTorch functionality
- **Enhanced Compatibility:** Allows more models to be exported successfully

cc @avikchaudhuri @gmagogsfm @zhxchen17 @tugsbayasgalan @angelayi @suo @ydwu4 @penguinwu

### DIFF
--- a/torch/_prims_common/__init__.py
+++ b/torch/_prims_common/__init__.py
@@ -1895,8 +1895,6 @@ def set_correction(
     # NB: we don't actually support symint here, but it's harmless to accept
     if not isinstance(correction, (IntLike, FloatLike)):
         raise ValueError("correction argument should be integer or float")
-    if correction < 0:
-        raise ValueError("correction argument should be non-negative")
     return sym_float(correction)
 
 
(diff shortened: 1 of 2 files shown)


parsshar-RH:
Looks like these tests are failing during gradients calculation due to new test cases with `dim=(1,) `

I think the straight forward solution should be to fix `dim=(1,)` in new test cases 

### Refernce from new test cases:
`yield SampleInput(tensor_nd(), dim=(1,), correction=-5)`

Please correct me if i am not correct or missing anything

isuruf (approved):

PR #162265, opened 2025-09-05
[CD][BE] Delete Python-3.9 case

malfet:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162136
* __->__ #162265
* #162297

And raise error when building for an unsupported version

PR #162266, opened 2025-09-05
[CD][BE] Get rid of SETUPTOOLS and PYYAML extra pins

malfet:
* #162136
* #162265
* #162297
* __->__ #162266

As those weren't really a pins to begin with, and requirments.txt
already has those


clee2000 (approved):

Skylion007 (approved):

ZainRizvi (approved):
Why did we even have these?

image

PR #157981, opened 2025-07-10
Fix DCE eliminating random operations by improving is_impure() (#151524)

soumith:
DCE was incorrectly eliminating unused random operations like torch.rand() that have global RNG side effects, causing inconsistent results between eager and compiled execution modes.

**Root cause**: Python random functions (torch.rand, torch.randn, etc.) don't have the _nondeterministic_seeded attribute, so node.is_impure() returns False, allowing DCE to eliminate them despite advancing global RNG state.

**Solution**: Enhanced is_impure() in torch/fx/node.py to recognize Python random functions and mark them as impure when they use global RNG, regardless of the impure_random parameter setting. This ensures consistency between eager and compiled execution even when config.fallback_random=False.

**Key features**:
- Handles comprehensive list of random functions: rand, randn, randint, randperm, rand_like, randn_like, randint_like, normal, poisson, bernoulli, multinomial
- Generator optimization: Only marks as impure when using global RNG (no generator or generator=None). Operations with explicit generators don't affect global state and can be optimized.
- Works with both impure_random=True and impure_random=False cases
- Cleaner architecture: addresses root cause rather than working around it

**Tests**: Enhanced test_impure_random to verify both FX tracing and AOT compilation codepaths, ensuring random operations are preserved and eager/compiled execution consistency is maintained.

🤖 Generated with [Claude Code](https://claude.ai/code)

Fixes https://github.com/pytorch/pytorch/issues/151524

cc @ezyang @SherlockNoMad @EikanWang @jgong5 @wenzhe-nrv

PR #162267, opened 2025-09-05
Fix DCE eliminating in-place operations by improving Node.is_impure()

Eldalie:
Change is_impure to check in-place operations on Node to prevent eliminate_dead_code from eliminating in-place operations.

### DIFF
--- a/test/fx/test_dce_pass.py
+++ b/test/fx/test_dce_pass.py
@@ -1,5 +1,6 @@
 # Owner(s): ["module: fx"]
 import copy
+import inspect
 import unittest
 from typing import Optional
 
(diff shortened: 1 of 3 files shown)


ezyang (dismissed):

ezyang:
It looks like the last time we extended this it was also ok so I guess it is not super risky to do this https://github.com/pytorch/pytorch/pull/157981

ezyang:
need lintfix

torch/fx/node.py
```diff
@@ -99,6 +99,18 @@
     _ops.profiler._record_function_exit,
     _ops.inductor.accumulate_grad_.default,
     operator.setitem,
+    operator.iadd,
+    operator.isub,
+    operator.imul,
+    operator.itruediv,
+    operator.ifloordiv,
+    operator.imod,
```
Skylion007: Why no operator.imatmul? https://docs.python.org/3/library/operator.html

torch/fx/node.py
```diff
@@ -809,6 +821,14 @@ def is_impure(self, impure_random: bool = True) -> bool:
             )
             return getattr(target_mod, "_is_impure", False)
 
+        if self.op == "call_method":
+            target_name = (
+                self.target if isinstance(self.target, str) else torch.typename(self.target)
+            )
+            # Check for functions with names ending in an underscore (e.g., 'add_') that are inplace in torch
+            if target_name.endswith("_"):
```
Skylion007, line 841: Yikes this is hacky and a performance footgun no way to use opinfo tags here?

torch/fx/node.py
Skylion007: May also be good to extract inplace operator set into its own subset / utility stub for reuse elsewhere.

test/fx/test_dce_pass.py
```diff
@@ -182,6 +182,22 @@ def forward(self, a: torch.Tensor) -> torch.Tensor:
             TestModule(), expect_dce_changes=False, modules_to_be_leafs={ReLUImpure}
         )
 
+    def test_keep_inplace_with_side_effects(self):
```
Skylion007, line 218: Shouldnt we test the newly inplace operator coverage somehow? Maybe dynamically create classes and call operator explicitly?

torch/fx/node.py
Eldalie: Good catch, I changed the PR to take all in-place operators from the documentation.

torch/fx/node.py
Eldalie: I know, but from the tests with Dynamo or with torch.fx.Tracer, target and op are only str, and no other data is present, so I don’t know how to do better.

Eldalie:
@ezyang i fix test with @torch.compiler.disable in _trace_with_dynamo

Eldalie:
@jerryzh168 I rebased the PR

PR #162268, opened 2025-09-05
[CUDA 13][cuDNN] Bump CUDA 13 to cuDNN 9.13.0

eqy:
Fixes some `d_qk` != `d_v` cases on Hopper that are broken by cuDNN 9.11-9.12


drisspg (approved):

drisspg:
Is the wheel already on pypi?

Skylion007 (approved):
Do we need to update CUDNN frontend too? What about CUDA 12 builds?

eqy:
> Is the wheel already on pypi?

Yes looks like it https://pypi.org/project/nvidia-cudnn-cu13/

atalman:
@nWEIdia correct we need to update these wheels, otherwise nightly build will fail

eqy:
@atalman it is a nice-to-have for Windows as we did not make cuDNN default for non-datacenter GPUs

atalman:
@eqy lets roll back window portion of this since I don't think we can do another AMI update before the branch cut

atalman:
Thank you @eqy Linux cudnn was uploaded to download.pytorch.org

PR #162272, opened 2025-09-05
Add support for NestedTensor share_memory_

### LINKED ISSUE #161915: `share_memory_` support for `NestedTensor`

call `share_memory_()` method of NestedTensor will get segmentation fault.

```
>>> a = torch.randn(3)
>>> b = torch.randn(5)
>>> nt = torch.nested.nested_tensor([a, b], layout=torch.jagged)
>>> nt
NestedTensor(size=(2, j1), offsets=tensor([0, 3, 8]), contiguous=True)
>>> nt
nt
>>> type(nt)
<class 'torch.nested._internal.nested_tensor.NestedTensor'>
>>> nt.share_memory_()
Segmentation fault (core dumped)
```

### Alternatives

_No response_

### Additional context

_No response_

cc @cpuhrsch @jbschlosser @bhosmer @drisspg @soulitzer @davidberard98 @YuqingJ

adabeyta:
Fixes: https://github.com/pytorch/pytorch/issues/161915

### Summary

Implements share_memory_() support for NestedTensor! 

### Changes

- Added share_memory_() method to NestedTensor class.
  - Shares storage for all NestedTensor components: _values, _offsets, _lengths, and cached seqlen tensors.
  - Guard for CUDA Tensors.

### Testing

Before Fix:

`pytest -q test/test_nestedtensor.py -k "test_share_memory" -v`

Output:

```
Running 1 items in this shard

test/test_nestedtensor.py Fatal Python error: Segmentation fault
```

After Fix:

`pytest -q test/test_nestedtensor.py -k "test_share_memory" -v`

Output:

```
Running 1 items in this shard

test/test_nestedtensor.py::TestNestedTensorDeviceTypeCPU::test_share_memory_cpu PASSED [0.0753s] 
```

### DIFF
--- a/test/test_nestedtensor.py
+++ b/test/test_nestedtensor.py
@@ -1230,6 +1230,24 @@ class TestNestedTensorDeviceType(NestedTensorTestCase):
         is_cuda = "cuda" in str(device)
         self.assertEqual(nt.is_cuda, is_cuda)
 
+    def test_share_memory(self, device):
+        a = torch.randn(3, 4, device=device)
+        b = torch.randn(5, 4, device=device)
+        nt = torch.nested.nested_tensor([a, b], layout=torch.jagged)
+
+        # Guard CUDA tensors
+        if device == "cuda":
+            result = nt.share_memory_()
+            self.assertIs(result, nt)
+            return
+
+        result = nt.share_memory_()
+        self.assertIs(result, nt)
+
+        # Verify values/offsets are in shared memory
+        self.assertTrue(nt._values.is_shared())
+        self.assertTrue(nt._offsets.is_shared())
+
     @dtypes(torch.float, torch.float16, torch.double)
     def test_nested_tensor_indexing(self, device, dtype):
         # edge case: empty nested tensor
--- a/test/test_nestedtensor.py
+++ b/test/test_nestedtensor.py
@@ -1230,6 +1230,7 @@ class TestNestedTensorDeviceType(NestedTensorTestCase):
         is_cuda = "cuda" in str(device)
         self.assertEqual(nt.is_cuda, is_cuda)
 
+    @skipIfTorchDynamo("Not a suitable test for TorchDynamo")
     def test_share_memory(self, device):
         a = torch.randn(3, 4, device=device)
         b = torch.randn(5, 4, device=device)
--- a/test/test_nestedtensor.py
+++ b/test/test_nestedtensor.py
@@ -1237,7 +1237,7 @@ class TestNestedTensorDeviceType(NestedTensorTestCase):
         nt = torch.nested.nested_tensor([a, b], layout=torch.jagged)
 
         # Guard CUDA tensors
-        if device == "cuda":
+        if "cuda" in device:
             result = nt.share_memory_()
             self.assertIs(result, nt)
             return
--- a/test/test_nestedtensor.py
+++ b/test/test_nestedtensor.py
@@ -1245,9 +1245,8 @@ class TestNestedTensorDeviceType(NestedTensorTestCase):
         result = nt.share_memory_()
         self.assertIs(result, nt)
 
-        # Verify values/offsets are in shared memory
-        self.assertTrue(nt._values.is_shared())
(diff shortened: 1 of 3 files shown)


torch/nested/_internal/nested_tensor.py
```diff
@@ -163,6 +163,25 @@ def offsets(self):
     def lengths(self):
         return self._lengths
 
+    def share_memory_(self):
+        # Guard CUDA tensors
+        if self._values.is_cuda:
```
jbschlosser: nit: perhaps `self.is_cuda` instead here?

torch/nested/_internal/nested_tensor.py
```diff
@@ -163,6 +163,25 @@ def offsets(self):
...
+        self._values.share_memory_()
+        self._offsets.share_memory_()
+
+        if self._lengths is not None:
+            self._lengths.share_memory_()
+
+        if self._min_seqlen_tensor is not None:
+            self._min_seqlen_tensor.share_memory_()
+        if self._max_seqlen_tensor is not None:
+            self._max_seqlen_tensor.share_memory_()
```
jbschlosser: nit: an alternative approach that is slightly more DRY would utilize `__tensor_flatten__()` to get the components on which to call `share_memory_()`

jbschlosser (commented):
Thanks for the contribution!

torch/nested/_internal/nested_tensor.py
adabeyta: Thank you for the suggestion, updated to reflect a more DRY components call to share_memory_()

test/test_nestedtensor.py
```diff
@@ -1230,6 +1230,25 @@ def test_device_checks(self, device):
...
         self.assertEqual(nt.is_cuda, is_cuda)
 
+    @skipIfTorchDynamo("Not a suitable test for TorchDynamo")
+    def test_share_memory(self, device):
+        a = torch.randn(3, 4, device=device)
+        b = torch.randn(5, 4, device=device)
+        nt = torch.nested.nested_tensor([a, b], layout=torch.jagged)
+
+        # Guard CUDA tensors
+        if device == "cuda":
```
jbschlosser:
```suggestion
        if "cuda" in device:
```

this is more accurate because in practice `device` will be something like `"cuda:0"`

torch/nested/_internal/nested_tensor.py
```diff
@@ -163,6 +163,19 @@ def offsets(self):
     def lengths(self):
         return self._lengths
 
+    def share_memory_(self):
```
jbschlosser: instead of implementing the method directly on the subclass, it's better to override it via NJT's `__torch_function__()` implementation here: https://github.com/pytorch/pytorch/blob/f638854e1da6b33e78dcc9f3e28c98c4cdce4e86/torch/nested/_internal/ops.py#L370

this should compose better with the rest of PyTorch.

jbschlosser (commented):
couple small things, nothing huge :) thanks again!

jbschlosser:
> Does the NestedTensor.is_shared() method return the correct result? Or we must use NestedTensor.values().is_shared() to check?

We definitely want `NestedTensor.is_shared()` to return the correct result. @adabeyta this will likely require an additional implementation of `is_shared()` in NJT's `torch_function` impl.

torch/nested/_internal/nested_tensor.py
adabeyta: Updated to override from NJT's __torch_function__(). Thank you for the tip @jbschlosser!

adabeyta:
> > Does the NestedTensor.is_shared() method return the correct result? Or we must use NestedTensor.values().is_shared() to check?
> 
> We definitely want `NestedTensor.is_shared()` to return the correct result. @adabeyta this will likely require an additional implementation of `is_shared()` in NJT's `torch_function` impl.

I went ahead and added NT.is_shared() to return the correct result and added it under the NJT impl as well. Thank you for catching that @Phimos

jbschlosser (approved):
Looks great, thanks for the updates!

PR #162281, opened 2025-09-05
Make stride argument to Tensor.set_ optional

swolchok:
* #161635
* __->__ #162281

Long-standing fixme. Seems to be blocking #161635.


aten/src/ATen/native/Resize.h
```diff
@@ -125,11 +125,10 @@ inline void checkInBoundsForStorage(
...
 template <typename T>
 inline void checkSetStorage(Tensor& result, Storage storage, T storage_offset,
-                                   ArrayRef<T> size, ArrayRef<T> stride, bool check_offset_in_bounds = true) {
-  // FIXME: stride should be optional
-  if (stride.data()) {
-    TORCH_CHECK(size.size() == stride.size(), "unequal size length (", size.size(),
-                                              ") and stride length (", stride.size(), ")");
+                                   ArrayRef<T> size, OptionalArrayRef<T> stride, bool check_offset_in_bounds = true) {
+  if (stride.has_value()) {
+    TORCH_CHECK(size.size() == stride.value().size(), "unequal size length (", size.size(),
```
Skylion007: Nit: should be TORCH_CHECK_VALUE

aten/src/ATen/native/mps/TensorFactory.cpp
```diff
@@ -140,12 +140,12 @@ Tensor& set_mps_(Tensor& result) {
...
-Tensor& set_storage_mps_(Tensor& result, Storage storage, int64_t storage_offset, IntArrayRef size, IntArrayRef stride) {
+Tensor& set_storage_mps_(Tensor& result, Storage storage, int64_t storage_offset, IntArrayRef size, OptionalIntArrayRef stride) {
   checkSetStorage(result, std::move(storage), storage_offset, size, stride);
-  //std::cout << "set storage_mps " << storage_offset << " stride " << stride << std::endl;
+  //std::cout << "set storage_mps " << storage_offset << " stride " << stride_opt << std::endl;
   result.unsafeGetTensorImpl()->set_storage_offset(storage_offset);
-  std::optional<IntArrayRef> stride_opt = stride.data() != nullptr ?
-                                          std::optional<IntArrayRef>(stride) : std::nullopt;
+  std::optional<IntArrayRef> stride_opt = stride.has_value() ?
+    std::optional<IntArrayRef>(stride.value()) : std::nullopt;
```
Skylion007, line 148: Wait, no need to unbox, use copyctor since stride is also the same optional type? Just keep it as taking in stride and use copy ctor, no need to call optional ctor right? The OptionalIntArrRef is same as std::optional<IntArrRed>, right?

swolchok: OptionalIntArrayRef is not identical with any kind of std::optional, sadly https://github.com/pytorch/pytorch/blob/065c446193e05a1c9012b35df977bd397a93acb1/c10/util/OptionalArrayRef.h#L24

swolchok:
not 100% sure it's a break, but i'm flagging topic: bc breaking to make sure this doesn't get overlooked

ezyang:
We have done this kind of break before. It is kinder to downstream if we can introduce a macro they can test against so they can do both signatures. You typically are expected to update XPU's code directly. I actually don't know how to do this though, typically I have to update XLA for signature changes like this.

PR #162284, opened 2025-09-05
Handle f([]) vs. f() in fake tensor caching

angelayi:
Fixes https://github.com/pytorch/pytorch/issues/162279

### DIFF
--- a/torch/_subclasses/fake_tensor.py
+++ b/torch/_subclasses/fake_tensor.py
@@ -1677,6 +1677,10 @@ class FakeTensorMode(TorchDispatchMode):
         )
         from torch._higher_order_ops.utils import FunctionalizeCtxWrapper
 
+        if isinstance(args, (list, tuple, dict)):
+            result.append(type(args))
+            result.append(f"length_{len(args)}")
+
         if isinstance(args, dict):
             self._prep_args_for_hash(result, args.keys(), state, id_hashed_objects)
             self._prep_args_for_hash(result, args.values(), state, id_hashed_objects)
--- a/test/test_fake_tensor.py
+++ b/test/test_fake_tensor.py
@@ -1943,6 +1943,16 @@ class FakeTensorDispatchCache(TestCase):
             self._test_cache_key(fm, 1.0, 1.0, 1)
             self._test_cache_key(fm, 0.0, 0.0, 0)
 
+    def test_empty_list(self):
+        with FakeTensorMode() as fm:
+            func = aten.any.dims
+            state = _CacheKeyState()
+            x = torch.ones((2, 3))
+            key_x = fm._cache_key(state, func, [x, []], {})
+            key_y = fm._cache_key(state, func, [x], {})
+
+        self.assertNotEqual(key_x, key_y)
+
     def assertHitsMisses(self, hits, misses):
         """
         Helper to assert on the number of recorded hits and misses.


manuelcandales (approved):

torch/_subclasses/fake_tensor.py
```diff
@@ -1699,6 +1699,9 @@ def _prep_args_for_hash(
             elif isinstance(arg, (SymBool, SymFloat)):
                 raise _BypassDispatchCache("symbolic shape")
             elif isinstance(arg, (list, tuple, dict)):
+                if len(arg) == 0:
```
aorenste: A previous comment pointed this out and I think it's correct - putting the check here will miss the check on the top-most list. Instead this should be a check on `args` outside the loop (and then skip the loop)

aorenste (changes requested):

torch/_subclasses/fake_tensor.py
manuelcandales: I wrote the comment and deleted it, because then I wasn't sure what would be implication for a function without inputs. This may cause problems, wouldn't it?

aorenste: Hm. That's an interesting point - but should be fine. It will cause a function without inputs to get an extra value (empty_tuple) as part of the key - but no harm - just an extra key value which will always match.

manuelcandales: Actually, it should be fine for a function without inputs, and the key would not be modified. Looking at the _cache_key function, I can see that _prep_args_for_hash is called on the args list, only if the args list is non-empty:
https://github.com/pytorch/pytorch/blob/main/torch/_subclasses/fake_tensor.py#L1579-L1580

So yes, the check should be moved outside the loop

angelayi: makes sense! updated!

torch/_subclasses/fake_tensor.py
```diff
@@ -1681,6 +1681,11 @@ def _prep_args_for_hash(
             self._prep_args_for_hash(result, args.keys(), state, id_hashed_objects)
             self._prep_args_for_hash(result, args.values(), state, id_hashed_objects)
             return
+                
+        if len(args) == 0:
```
manuelcandales: I think you want to move this before the `if isinstance(args, dict):` check above, otherwise you run into trouble with dicts.

manuelcandales: On this same `dict` topic, seeing the existing source code, I wonder, would the cache produce the same key for something like: `{1: 2}` vs `[1, 2]`?
Seems to me that the key would be `(int, 1, int 2)` for both

torch/_subclasses/fake_tensor.py
```diff
@@ -1681,6 +1681,11 @@ def _prep_args_for_hash(
...
+            result.append(type(args))
+            result.append(f"empty_{type(args)}")
+            return
```
manuelcandales: I think this should fix the lint errors and address the issue above, by differentiating dict vs list vs tuple

```suggestion

        if isinstance(args, (list, tuple, dict)):
            result.append(type(args))
            result.append(f"length_{len(args)}")
```

aorenste: I like this.

aorenste (approved):
Great - thanks!

PR #162285, opened 2025-09-05
[nativert] AOTI lowering and packaging as NativeRT delegate

yiming0416:
Summary:
A demo for creating AOTI delegate for NativeRT in OSS.

- It supports full graph lowering only.
- It leverages `executorch_call_delegate` HOP but doesn't rely on `executorch`.
- The delegate graph is obtained by tracing a `LoweredBackendModule` whose forward function calls `executorch_call_delegate`.
- The main difference between `executorch_call_delegate` and `aoti_call_delegate` is that the delegate graph from `executorch_call_delegate` doesn't have weights lifted as inputs.
- original_ep and delegate_ep are treated as flat EP dictionary and there is no nested structure.
- The naming contract is enforced by `model_name` and `backend_id`

Test Plan:
CI

Rollback Plan:

Differential Revision: D81641157

### DIFF
--- a/torch/_export/serde/serialize.py
+++ b/torch/_export/serde/serialize.py
@@ -982,6 +982,15 @@ class GraphModuleSerializer(metaclass=Final):
                     return Argument.create(
                         as_graph=GraphArgument(name=arg.target, graph=graph)
                     )
+                elif type(attr).__name__ == "LoweredBackendModule":
+                    # Special handling for executorch_call_delegate HOP
+                    # It's first argument is a LoweredBackendModule, for which we
+                    # serialize name and backend id of the lowered module
+                    module_name = getattr(attr, "module_name", None)
+                    backend_id = getattr(attr, "backend_id", None)
+                    assert module_name is not None, "module_name should not be None"
+                    assert backend_id is not None, "backend_id should not be None"
+                    return Argument.create(as_string=f"{module_name}-{backend_id}")
                 else:
                     raise SerializeError(
                         f"Unsupported getattr attribute {arg.target} with type: {type(attr)}"
--- a/torch/nativert/backends/lower_utils.py
+++ b/torch/nativert/backends/lower_utils.py
@@ -1,0 +1,50 @@
+import torch
+from torch.export import ExportedProgram
+from torch.export.pt2_archive._package import AOTI_FILES, package_pt2
+from torch.types import FileLike
+
+from .lowered_aoti_module import LoweredBackendModule
+
+
+def lower_exported_program(
+    exported_program: ExportedProgram, model_name: str, backend_id: str
+) -> tuple[ExportedProgram, AOTI_FILES]:
+    """
+    Lower an exported program to AOTInductor and return a delegate ExportedProgram
+    with the `executorch_call_delegate` HOP
+    """
+    args, kwargs = exported_program.example_inputs
+    aoti_files = torch._inductor.aot_compile(
+        exported_program.module(), args, kwargs, options={"aot_inductor.package": True}
+    )
+    assert isinstance(aoti_files, list)
+
+    lowered_aoti_module = LoweredBackendModule(
+        exported_program, backend_id, module_name=model_name
+    )
+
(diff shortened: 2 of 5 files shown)


torch/nativert/backends/lower_utils.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    Package a pt2 archive file that can be consumed by NativeRT with AOTI Delegate
+    """
+    package_pt2(
+        f,
+        exported_programs={
+            model_name: original_ep,
+            f"{model_name}-{backend_id}": delegate_ep,
+        },
+        aoti_files={f"{model_name}-{backend_id}": aoti_files},
```
dolpm: nit: i feel this can be called delegate_files or something since we already know which backend they are relevant to from the id.

dolpm (approved):
address lints

torch/_export/serde/serialize.py
```diff
@@ -982,6 +982,15 @@ def serialize_input(self, arg, arg_type: Optional[Any] = None) -> Argument:
                     return Argument.create(
                         as_graph=GraphArgument(name=arg.target, graph=graph)
                     )
+                elif type(attr).__name__ == "LoweredBackendModule":
```
angelayi, line 985: hmm I thought this would be a getattr node?

torch/nativert/backends/lowered_aoti_module.py
```diff
@@ -0,0 +1,30 @@
+import torch
+from torch.export import ExportedProgram
+
+
+class LoweredBackendModule(torch.nn.Module):
+    def __init__(
+        self,
+        module_name: str,
```
angelayi: what is module_name?

torch/nativert/backends/lowered_aoti_module.py
```diff
@@ -0,0 +1,30 @@
...
+        backend_id: str,
+        original_exported_program: ExportedProgram,
```
angelayi, line 10: maybe we can try to make the signature consistent with [executorch's LoweredBackendModule](https://www.internalfb.com/code/fbsource/fbcode/executorch/exir/lowered_backend_module.py?lines=54)?

torch/nativert/backends/lowered_aoti_module.py
yiming0416: @angelayi we use `{module_name}-{backend-id}` as the name of the delegate ep.
`module_name` is essentially the name of the model being lowered (given now we only support full graph lowering)

torch/nativert/backends/lowered_aoti_module.py
yiming0416: @angelayi Things like `CompileSpec` and `NamedDataStoreOutput` are exir-specific data structures.

The motivation of having this simplified `LoweredBackendModule` is that is doesn't depend on executorch and can be used standalone in `pytorch/pytorch`

torch/_export/serde/serialize.py
yiming0416: @angelayi  `executorch_call_delegate` hop only serves as an indicator for nativert to construct a AOTIDelegateExecutor in C++ runtime.

So all we need is a path in string format that can be looked up in the runtime for loading AOTI artifacts.

angelayi: sorry I think I misread this code

torch/nativert/backends/lowered_aoti_module.py
angelayi: yeah that makes sense, but since we're already using the executorch_call_delegate op, in the future maybe we also want to deduplicate the LoweredBackendModule, so maybe we can try to make the signature consistent 

it could look something like:
```python
    def __init__(
        self,
        original_exported_program: ExportedProgram,
        backend_id: str,
        *,
        module_name: Optional[str] = None,
        # ET could add these fields whenever they want to migrate
        processed_bytes,
        compile_specs: List[CompileSpec],
        named_data_store_output: Optional[NamedDataStoreOutput] = None,
    ) -> None:
```

PR #162312, opened 2025-09-05
[DTensor] DTensor.item() should Replicate() inputs

zou3519:
* __->__ #162312
* #162307
* #162117

Previously, DTensor.item() didn't replicate inputs. This caused silent
incorrectness in the case of DTensor.max().item() -- .max() returns
DTensors with Partial(max) placement and .item() just returned the local
values.

This PR changes it so that DTensor.item() replicates the inputs before
calling .item().

Test Plan:
- new tests

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci @gchanan

### DIFF
--- a/torch/distributed/tensor/_sharding_prop.py
+++ b/torch/distributed/tensor/_sharding_prop.py
@@ -25,6 +25,7 @@ from torch.distributed.tensor._utils import (
     compute_local_shape_and_global_offset,
     compute_local_stride,
 )
+from torch.distributed.tensor.placement_types import Replicate
 
 
 aten = torch.ops.aten
@@ -335,7 +336,25 @@ class ShardingPropagator:
         # special case op, we don't need to propagate for local
         # scalar. TODO: figure out a better way to handle this
         if op_schema.op is aten._local_scalar_dense.default:
-            return OutputSharding(None, op_schema)
+            # Need to Replicate() the input before calling .item() on the local tensor,
+            # to resolve any Partial(...) placements.
+            arg_spec = op_schema.args_spec[0]
+            placements = arg_spec.placements
+            return OutputSharding(
+                None,
+                OpSchema(
+                    op_schema.op,
+                    (
+                        DTensorSpec(
+                            arg_spec.mesh,
+                            tuple(Replicate() for _ in range(len(placements))),
+                            arg_spec.tensor_meta,
+                        ),
+                    ),
+                    {},
+                ),
+                needs_redistribute=True,
+            )
 
         out_tensor_meta = self._propagate_tensor_meta_non_cached(op_schema)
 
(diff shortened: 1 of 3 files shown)


ezyang:
@wanchaol I read over the linked PR and while I acknowledge that this behavior is intentional and it may now be BC breaking to change it, I don't understand the underlying justification. For example, its behavior seems inconsistent with another operator that returns a non-Tensor, aten.equal:

```
        if output_sharding.output_spec is None:
            if op_call == aten.equal.default:
                # For equal operator, The local results from all devices should be all-gathered
                # and a reduce op (AND) will be performed on the list of results to ensure SPMD
                # execution. We can extend this for more ops if necessary.
                obj_list = [None for _ in range(dist.get_world_size())]
                dist.all_gather_object(obj_list, local_results)  # type: ignore[possibly-undefined]
                obj_list = list(filter(lambda x: x is not None, obj_list))
                # perform reduce on the collection with AND op
                local_results = functools.reduce(operator.and_, obj_list, True)
```

Here, we implicitly trigger an allgather so that we return what aten.equal would have reported for the full tensor. Why should item() go the other way?

My best guess is that you are trying to avoid implicit communications for what should be a cheap operation (item()). But you can always do `dtensor.to_local().item()` for the no comms case. It seems weird to privilege one over the other.

wanchaol:
> Here, we implicitly trigger an allgather so that we return what aten.equal would have reported for the full tensor. Why should item() go the other way?
> 
> My best guess is that you are trying to avoid implicit communications for what should be a cheap operation (item()). But you can always do `dtensor.to_local().item()` for the no comms case. It seems weird to privilege one over the other.

@ezyang Yeah this is exactly what I was trying to avoid, `item` suppose to be a cheap call. 

But now I think about it again 🤔, I feel it make sense to force replicating the outputs. The reason is that `item()` call returns a python number instead of staying as a tensor object, so the moment user call `item()` the returned value does not being tracked by DTensor anymore compared to other ops, so we might want to make sure the returned value clears the pending reduction state to reduce user confusion, and forcing replicate in this case make sense to me!

torch/distributed/tensor/_sharding_prop.py
```diff
@@ -335,7 +336,25 @@ def propagate_op_sharding_non_cached(self, op_schema: OpSchema) -> OutputShardin
...
+                    (
+                        DTensorSpec(
+                            arg_spec.mesh,
+                            tuple(Replicate() for _ in range(len(placements))),
+                            arg_spec.tensor_meta,
+                        ),
+                    ),
+                    {},
+                ),
+                needs_redistribute=True,
```
wanchaol, line 360: nit: maybe only generate the op schema and set `needs_redistribute` when the placements is not all replicated

wanchaol (approved):
lgtm!

torch/distributed/tensor/_sharding_prop.py
zou3519: Fair enough

PR #162320, opened 2025-09-06
[SymmMem] Add team pool to hold duplicated teams for the same rank group

kwen2501:
* #162243
* __->__ #162320

When multiple threadblocks call device-side collectives concurrently, NVSHMEM requires each call being made on a separate team struct, see [Collective operations scopes and active sets](https://docs.nvidia.com/nvshmem/api/gen/api/collectives.html?highlight=nvshmem_barrier_all#collective-operations-scopes-and-active-sets).

This PR adds a util `get_n_teams` for creating duplicated nvshmem teams for the same rank group, i.e. team pool. So that we can use them on device side.


torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
```diff
@@ -0,0 +1,130 @@
...
+// #include <nvshmem_host.h>
+// It translates into the following two lines:
+#include <host/nvshmem_api.h>
+#include <host/nvshmemx_api.h>
+// For maximum compatibility, we use the "host/" style for now.
+
+namespace c10d::nvshmem_extension {
+
+// This corresponds to max nblocks
+constexpr int MAX_N_TEAMS = 32;
```
ngimel: In nvshmem looks like they create 128?

torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
```diff
@@ -0,0 +1,130 @@
...
+    bool pool_updated = false;
+    auto& team_pool =
+        group_to_team_pool(group_name, global_ranks, need_n, pool_updated);
+    // Check if the pool already exists in device memory
+    nvshmem_team_t* team_pool_dev = nullptr;
+    constexpr auto pool_bytes = sizeof(nvshmem_team_t) * MAX_N_TEAMS;
+    auto it = team_pool_devptrs_.find(group_name);
+    if (it == team_pool_devptrs_.end()) {
+      // If not, allocate a new pool in device memory
+      C10_CUDA_CHECK(cudaMalloc((void**)&team_pool_dev, pool_bytes));
```
ngimel: do you know how big is this allocation?

torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
```diff
@@ -0,0 +1,130 @@
...
+    auto pair = group_name_to_team_pool_.emplace(
+        group_name, nvshmemTeamPool_t(MAX_N_TEAMS, NVSHMEM_TEAM_INVALID));
+    // pair.first is an iterator to the inserted element or the existing element
+    auto& team_pool = pair.first->second;
+    // pair.second is true if the element is inserted, false if the element
+    // already exists
+    pool_updated = pair.second;
+
+    // Some checks before we create new teams
+    TORCH_CHECK(global_ranks.size() > 1);
```
ngimel: there's no point in checking global ranks here if we are not going to create new teams?

torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
kwen2501: Maybe bc they'd need to cover warp-scope calls as well? 
So far our ops do not need so small granularity yet, but I agree that we can bump this as needed.

torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
kwen2501: One `nvshmem_t` is just an `int`. So pretty small.

kwen2501: Okay, let me bump `MAX_N_TEAMS` to 128 bc of this.

torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
```diff
@@ -0,0 +1,130 @@
...
+    auto& team_pool =
+        group_to_team_pool(group_name, global_ranks, need_n, pool_updated);
+    // Check if the pool already exists in device memory
+    nvshmem_team_t* team_pool_dev = nullptr;
+    constexpr auto pool_bytes = sizeof(nvshmem_team_t) * MAX_N_TEAMS;
+    auto it = team_pool_devptrs_.find(group_name);
+    if (it == team_pool_devptrs_.end()) {
+      // If not, allocate a new pool in device memory
+      C10_CUDA_CHECK(cudaMalloc((void**)&team_pool_dev, pool_bytes));
+      team_pool_devptrs_[group_name] = team_pool_dev;
```
ngimel: you should think about structuring the code in such a way that when group is destroyed, it's team manager entries are also freed, right now you have cudaMalloc (and possibly other resources) that's leaking.

torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
kwen2501: Fixed, thanks!

torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
kwen2501: Added in the destructor now.

torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
```diff
@@ -0,0 +1,170 @@
...
+      // cudaFree generally implies a device synchronization, meaning it will
+      // block until all preceding CUDA operations on the device have completed
+      // before freeing the memory. Thus we don't need to worry about freeing
+      // the memory before CUDA kernels complete.
+      for (auto& [_, team_pool_dev] : team_pool_devptrs_) {
+        c10::cuda::CUDACachingAllocator::raw_delete(team_pool_dev);
+      }
+    } catch (...) {
+      // Ignore the error
+      std::cerr << "Failed to free the team pool in device memory, skipping\n";
```
Skylion007: Why not use our logging utility we already have in TORCH?

torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
Skylion007: Wouldn't it be cleaner to use std::unique_ptrs to accomplish this (with a custom destructor passed to the unique_ptr constructor)?

torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
```diff
@@ -0,0 +1,170 @@
...
+    }
+  }
```
kwen2501: @ngimel I am doing the free in best-effort manner due to undetermined destruction order of static objects and CUDA context. In my test runs, I never see the cerr message being printed tho, so I guess we are lucky enough. 

Groups are a bit static today too. So I guess it would make little difference if we were to implement a callback from group destructor.

torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
```diff
@@ -0,0 +1,170 @@
...
+          team_pool_dev,
+          team_pool.data(),
+          pool_bytes,
+          cudaMemcpyHostToDevice,
+          stream));
+    }
+    return std::make_pair(std::cref(team_pool), team_pool_dev);
+  }
+
+  ~TeamManager() {
```
Skylion007: Should be marked as noexcept, right?

torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
kwen2501: I was worried that those logging utility might have been destructed at this point. Can you shed some light on this?

torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
kwen2501: Normally yes (thanks for the suggestion)! But here I have specific comments + warning message, so I kind of prefer writing it out explicitly.

torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
kwen2501: added

torch/csrc/distributed/c10d/symm_mem/nvshmem_team_manager.hpp
Skylion007: @kwen2501 True, but comments has way more footguns than RAII

ngimel: Agree, using standard RAII techniques is preferable

ngimel (approved):
I take it, there will be tests for it some time down the stack?

kwen2501:
All the existing tests will exercise the host-side `get_team` API.
The tile_reduce feature on top will exercise the device-side `get_n_teams` API.

PR #162321, opened 2025-09-06
forward fix mypy type errors in torch when linalg is type checked

stmcgovern:
* (to be filled)

----

- Fix cholesky_ex tuple access in distributions/constraints.py
- Fix multi_dot argument type in nn/utils/spectral_norm.py
- Fix list to tuple conversions for linalg functions in _refs/__init__.py
- Fix ord parameter handling in masked/_ops.py
- Fix NumberType conversion in _refs/nn/functional/__init__.py
- Add TorchScript-compatible type ignores in functional.py

Resolves 14 MyPy type errors while maintaining TorchScript compatibility.

### DIFF
--- a/torch/_refs/__init__.py
+++ b/torch/_refs/__init__.py
@@ -3229,12 +3229,12 @@ def native_group_norm(
     computation_dtype = utils.get_computation_dtype(input.dtype)
     input_acc = _maybe_convert_to_dtype(input, computation_dtype)
     # num_channels / num_groups and flattened inner dimension are the reduction axes
-    reduction_dims = [2, 3]
+    reduction_dims_list = [2, 3]
     input_reshaped = torch.reshape(
         input_acc,
         [batch_size, num_groups, num_channels // num_groups, flattened_inner_size],
     )
-    reduction_dims = utils.canonicalize_dims(input_reshaped.ndim, reduction_dims)
+    reduction_dims = utils.canonicalize_dims(input_reshaped.ndim, reduction_dims_list)
     biased_var, mean = torch.var_mean(
         input_reshaped, dim=reduction_dims, unbiased=False, keepdim=True
     )
@@ -3394,10 +3394,10 @@ def renorm(
     acc_type = utils.get_computation_dtype(input.dtype)
     if acc_type != input.dtype:
         norm = torch.linalg.vector_norm(
-            input, p, reduce_dims, keepdim=True, dtype=acc_type
+            input, p, tuple(reduce_dims), keepdim=True, dtype=acc_type
         )
     else:
-        norm = torch.linalg.vector_norm(input, p, reduce_dims, keepdim=True)
+        norm = torch.linalg.vector_norm(input, p, tuple(reduce_dims), keepdim=True)
 
     eps = 1e-7
     norm_factor = torch.where(norm > maxnorm, maxnorm / (norm + eps), 1.0)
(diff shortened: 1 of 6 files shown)


torch/masked/_ops.py
```diff
@@ -1579,7 +1579,11 @@ def norm(
     if mask_input.layout == torch.strided:
         dim_ = _canonical_dim(dim, input.ndim)
         return torch.linalg.vector_norm(
-            mask_input, ord, dim_, bool(keepdim), dtype=dtype
+            mask_input,
+            ord if ord is not None else 2.0,
```
Skylion007, line 1583: This is a typing error of the definition (likely from gen_pyi.py) It clearly support None as input since it didn't error before.

torch/functional.py
```diff
@@ -1788,31 +1788,55 @@ def norm(  # noqa: F811
...
 
             # Here we either call the nuclear norm, or we call matrix_norm with some arguments
             # that will throw an error
             if _dim is None:
-                _dim = list(range(input.ndim))
+                _dim = [-2, -1]  # Default to last 2 dimensions for matrix norm
+
+            # Ensure we have at least 2 dimensions for matrix norm
+            if len(_dim) < 2:
+                raise RuntimeError(
```
Skylion007, line 1814: Should be ValueError

torch/distributions/constraints.py
```diff
@@ -634,7 +634,7 @@ def check(self, value):
         sym_check = super().check(value)
         if not sym_check.all():
             return sym_check
-        return torch.linalg.cholesky_ex(value).info.eq(0)
+        return torch.linalg.cholesky_ex(value)[1].eq(0)
```
Skylion007, line 637: Why the change here?

torch/_refs/nn/functional/__init__.py
```diff
@@ -1201,7 +1201,13 @@ def pairwise_distance(
     eps: NumberType = 1e-6,
     keepdim=False,
 ) -> TensorLikeType:
-    return torch.linalg.vector_norm(x1 - x2 + eps, ord=p, dim=-1, keepdim=keepdim)
+    if isinstance(p, bool):
```
Skylion007, line 1204: Do we actually want to support p as a bool and complex or is the typing of p just incorrect and should be narrow. Supporting Boolean and complex here feels weird.

torch/_refs/__init__.py
```diff
@@ -5865,9 +5865,13 @@ def norm(
         # that will throw an error
         if dim is None:
             dim = tuple(range(input.ndim))
-        return torch.linalg.matrix_norm(input, p, dim, keepdim, dtype=dtype)
+        # Extract exactly 2 dimensions for matrix norm
+        matrix_dims = (int(dim[-2]), int(dim[-1]))
```
Skylion007, line 5869: Do we not need a length check on dim? That would also fix the typing without manually extracting it, right?

torch/masked/_ops.py
stmcgovern: Thanks for the review! I'll be able to get back to this by the end of the week.

torch/_refs/nn/functional/__init__.py
stmcgovern: agreed. I think p can just be made a float.

torch/masked/_ops.py
stmcgovern: reverted. I'm looking into a change in gen_pyi.py

torch/functional.py
stmcgovern: done

torch/distributions/constraints.py
stmcgovern: I was seeing ` torch/distributions/constraints.py:637:16: error: "tuple[Tensor, Tensor]" has no attribute "info"  [attr-defined] `. Using this tuple access resolved the mypy errror, but the better solution is to keep `.info` and update `torch/_C/__init__.pyi.in`  to use `return_types_linalg_cholesk_ex` when applying the type checking.

torch/_refs/__init__.py
stmcgovern: yes, thanks. will replace with a check that dim has length 2.

PR #162328, opened 2025-09-06
`_compile_kernel()`: handle smem > 48kb

gau-nernst:
When shared memory is more than 48kb, we need to manually set `cudaFuncAttributeMaxDynamicSharedMemorySize` to the required shared memory size. This is documented in CUDA C++ API guide https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#shared-memory-7-x. And we can also see it in CUTLASS https://github.com/NVIDIA/cutlass/blob/76c96b0be35cb263debe3e3d8418b80911a544ab/include/cutlass/transform/device/transform_universal_adapter.hpp#L143-L148

When the requested shared memory is more than what is supported, `CUDA Invalid arguments` will be thrown. I decided not to check against `torch.cuda.get_device_properties().shared_memory_per_block_optin` at kernel invocation time to avoid overhead.

@msaroufim

### DIFF
--- a/test/test_cuda.py
+++ b/test/test_cuda.py
@@ -6879,7 +6879,7 @@ class TestCompileKernel(TestCase):
             grid=(blocks_per_grid, 1, 1),
             block=(threads_per_block, 1, 1),
             args=[a, b, c, N],
-            shared_mem=shared_mem
+            shared_mem=shared_mem,
         )
 
         # Verify results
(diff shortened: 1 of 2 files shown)


msaroufim (approved):

msaroufim:
pytorchbot merge -i

msaroufim (changes requested):
.

PR #162334, opened 2025-09-06
Add std::any_of to ConvParams struct

benjaminglass1:
* __->__ #162334

Removes some for-loops that didn't short-circuit in favor of std::any_of.

### DIFF
--- a/aten/src/ATen/native/Convolution.cpp
+++ b/aten/src/ATen/native/Convolution.cpp
@@ -14,6 +14,7 @@
 #include <c10/util/accumulate.h>
 #include <c10/util/irange.h>
 #include <c10/macros/Macros.h>
+#include <algorithm>
 #include <limits>
 #include <utility>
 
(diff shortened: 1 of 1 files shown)


aten/src/ATen/native/Convolution.cpp
```diff
@@ -300,67 +301,50 @@ struct ConvParams {
...
+      output_padding.cend(),
+      [](const T& p) { return p < 0; });
   }
 
   bool is_output_padding_big() const {
-    bool is_big = false;
+    // Revisit this with std::views::zip at C++20.
     for (auto i: c10::irange(output_padding.size())) {
-      is_big |= (output_padding[i] >= stride[i]);
+      if (output_padding[i] >= stride[i]) {
```
benjaminglass1, line 328: This changed to do short-circuit eval.

Skylion007 (approved):

aten/src/ATen/native/Convolution.cpp
```diff
@@ -300,67 +301,50 @@ struct ConvParams {
...
   }
 
   bool is_padded() const {
-    bool is_padded = false;
-    for (auto p : padding) {
-      is_padded |= (p != 0);
-    }
-    return is_padded;
+    return std::any_of(
+      padding.cbegin(), padding.cend(), [](const T& p) { return p != 0; });
```
Skylion007, line 315: Actually, slight nit here. should be 
```suggestion
      padding.cbegin(), padding.cend(), [](const T p) { return p != 0; });
```
no need to do const ref if the value itself is trivially copyable. It's int64 so it's only 1 unnecessary indirection. Really, really minor nit though.

benjaminglass1: @Skylion007 I can cancel the merge if that's important! Let me also check that this is guaranteed to be trivially copyable.

Skylion007: @benjaminglass1 If that is the case, we probably need to update this whole file anyway as a lot of the for each loops do the same thing. So this is actually optimal for consistency with the other methods in this file. That being said, we may want to revist the other methods in this file in a subsequent PR.

benjaminglass1: @Skylion007 So this class is instantiated in a few places with either `int64_t` or `c10::SymInt`. Looks like `c10::SymInt` is not trivially copyable, so I would prefer to keep this as a reference for now.

Skylion007: Ah, that's where it's coming form. The comment says it's always instantiated at int64_t or less, which is not true anymore I suppose. Yeah, looks good then.

PR #162340, opened 2025-09-06
[nn] Assert parsed iterable arguments are an appropriate length

benjaminglass1:
* __->__ #162340

Fixes #162327


cmake/Dependencies.cmake
```diff
@@ -112,6 +112,8 @@ if(USE_ASAN OR USE_TSAN)
   find_package(Sanitizer REQUIRED)
   if(USE_ASAN)
     if(TARGET Sanitizer::address)
+      target_compile_options(Sanitizer::address INTERFACE -mcmodel=medium)
+      target_link_options(Sanitizer::address INTERFACE -mcmodel=medium)
```
benjaminglass1: I'll split this out into another PR. This fixed an issue where the symbols section being linked in `libtorch_cpu` got too big for the default memory model when compiling with ASAN.

torch/nn/modules/utils.py
```diff
@@ -9,7 +9,8 @@
 
 def _ntuple(n, name="parse"):
     def parse(x):
-        if isinstance(x, collections.abc.Iterable):
+        if isinstance(x, Collection):
```
Skylion007: Technically doesn't need this, but does need to be:
```suggestion
        if isinstance(x, collections.abc.Iterable) and isinstance(x, collections.abc.Sized):
```

Skylion007 (approved):

torch/nn/modules/utils.py
benjaminglass1: @Skylion007 I thought about this over the weekend, and I think I should revert it to being an `Iterable` check, and then check the size of the result tuple before returning it. That way I don't have to add extra constraints.

test/nn/test_pooling.py
```diff
@@ -481,7 +481,7 @@ def test_max_unpool(self):
 
     def test_max_unpool3d_input_check(self):
         x = torch.ones(1, 3, 1, 1, 1)
-        with self.assertRaises(RuntimeError):
+        with self.assertRaises(AssertionError):
```
benjaminglass1, line 484: This changes because we intercept the error higher in the stack now.

test/test_nn.py
```diff
@@ -7466,14 +7466,8 @@ def test_padding_list(self):
...
-        arg_class = torch.nn.FractionalMaxPool2d(kernel_size=arg_1, output_ratio=arg_2,)
-        arg_3_0_tensor = torch.rand([20, 16, 50, 32], dtype=torch.float32)
-        arg_3_0 = arg_3_0_tensor.clone()
-        arg_3 = [arg_3_0,]
-
-        with self.assertRaisesRegex(ValueError,
-                                    "fractional_max_pool2d requires output_ratio to either be a single Int or tuple of Ints."):
-            res = arg_class(*arg_3)
+        with self.assertRaisesRegex(AssertionError, "Expected an iterable of length 2, but got length 3"):
+            arg_class = torch.nn.FractionalMaxPool2d(kernel_size=arg_1, output_ratio=arg_2,)
```
benjaminglass1, line 7491: Similarly, we intercept the error higher in the stack here.

test/test_mps.py
```diff
@@ -8924,7 +8924,7 @@ def helper(shape, padding, op, value=0):
         # pad dims == input dims
         helper((1, 3), (0, 2, 0, 1), nn.ConstantPad2d)
         # input.numel() == 0 but output.numel() > 0
-        helper((0, 3, 3), (1, 1, 1, 1, 1, 1), nn.ConstantPad2d)
```
benjaminglass1, line 8960: This was incorrect previously. `ConstantPad2d` already documents that it expects an `int` or a 4-tuple in this spot; we were just ignoring the extra values.

benjaminglass1:
I'm going to merge this tomorrow, if no one has any feedback before then (since I've already got one approval)!

huydhn:
> @huydhn Yeah, the test just needs to be updated. We've already documented that we expect an int or a 4-tuple in that argument, so this PR just enforces that, and one or two tests appear to break because of it.

If you have a PR on ExecuTorch, let me know, I can help coordinate the landing of these twos

PR #162342, opened 2025-09-06
fix-functorch-experimental-import-error

ghostspiders:
Fixes #162283
Maintain deprecation warnings and backward compatibility

### DIFF
--- a/functorch/experimental/__init__.py
+++ b/functorch/experimental/__init__.py
@@ -2,7 +2,8 @@
 try:
     from functorch import functionalize  # deprecation warnings
 except ImportError:
-    from torch._functorch.deprecated import functionalize  # Development environment 
+    from torch._functorch.deprecated import functionalize  # Development environment
+
 from torch._functorch.apis import chunk_vmap
 from torch._functorch.batch_norm_replacement import replace_all_batch_norm_modules_
 from torch._functorch.eager_transforms import hessian, jacfwd, jvp
--- a/functorch/experimental/__init__.py
+++ b/functorch/experimental/__init__.py
@@ -1,5 +1,8 @@
 # PyTorch forward-mode is not mature yet
-from functorch import functionalize
+try:
+    from functorch import functionalize  # deprecation warnings
+except ImportError:
+    from torch._functorch.deprecated import functionalize  # Development environment 
 from torch._functorch.apis import chunk_vmap
 from torch._functorch.batch_norm_replacement import replace_all_batch_norm_modules_
 from torch._functorch.eager_transforms import hessian, jacfwd, jvp


functorch/experimental/__init__.py
```diff
@@ -1,5 +1,9 @@
 # PyTorch forward-mode is not mature yet
-from functorch import functionalize
+try:
+    from functorch import functionalize  # deprecation warnings
+except ImportError:
+    from torch._functorch.deprecated import functionalize  # Development environment
+
```
zou3519, line 6: I don't understand what the problem is, can you elaborate please? Is it that `from functorch import functionalize` no longer works?

ghostspiders: After using "pip install -e . -v --no-build-isolation" to build torch. I got the following import error when running cmd "pytest test/functorch/test_control_flow.py"

ghostspiders: There is an issue importing the functorch module in development mode.

zou3519: so, after doing `pip install -e . -v --no-build-isolation`, `from functorch import functionalize` no longer works?

ghostspiders: To be precise:

1. `from functorch import functionalize` works normally.

2.  Running `pytest test/functorch/test_control_flow.py ` results in an import error;

3.  Whereas executing `python -m pytest test/functorch/test_control_flow.py` works normally.

zou3519: This doesn't tell me why it's failing, just that it is failing. I want to understand why it is failing -- did we regress `from functorch import functionalize`, and under what situations?

ghostspiders: error:
`ImportError while importing test module '/home/gaoyufeng/code/pytorch/test/functorch/test_control_flow.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
../../miniconda3/envs/mtorch/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
test/functorch/test_control_flow.py:8: in <module>
    from functorch.experimental import control_flow
functorch/experimental/__init__.py:2: in <module>
    from functorch import functionalize
E   ImportError: cannot import name 'functionalize' from 'functorch' (unknown location)`

ghostspiders: When running pytest directly, the test directory test/functorch is mistakenly identified as the functorch module, leading to incorrect import paths. This is a typical Python module search path priority issue.

zou3519: Why did it work before?

ghostspiders: I'm not sure how it was before, but now running `python test/run_test.py test/functorch/test_control_flow.py`is correct. When using pytest to execute directly, pytest adds the directory where the test file is located to the front of sys.path, causing it to erroneously recognize the test/functorch directory as the functorch module, resulting in an error.

ghostspiders: I think there are very few cases where we directly use pytest for testing. If there's no need to fix this issue, I will close this PR.

ghostspiders:
@ydwu4 @zou3519  Thanks for the review! Any other changes needed?

PR #149869, opened 2025-03-24
ProcessGroupGloo: support reduce_scatter + update support chart

d4l3k:
This adds a `reduce_scatter` implementation for ProcessGroupGloo. This is a pretty naive implementation as it does 1 allreduce per  rank but may be useful for testing in FSDP etc. There was an existing implementation of reduce_scatter_tensor/reduce_scatter_tensor_coalesed that has a very similar implementation but requires a fixed tensor size per rank.

If users find these functions to be too slow we can address them as issues arise.

Gloo now supports all major distributed operations. Quite a few of these were added by @rohan-varma and @yifuwang but they didn't update the support chart. We also have `CUDAWork` variants of most operations so those were also added to the chart.

Test plan:

```
pytest -v test/distributed/test_c10d_gloo.py -k reduce_scatter
```

cc @H-Huang @awgu @kwen2501 @wanchaol @fegin @fduwjj @wz337 @wconstab @c-p-i-o

PR #162344, opened 2025-09-07
Fix/fsdp gloo grad avg

### LINKED ISSUE #162338: Incorrect AVG reduction for GLOO
### 🐛 Describe the bug

```python
import os
import torch
import torch.nn as nn
from torch.distributed._composable.fsdp import fully_shard
from torch.distributed.device_mesh import init_device_mesh


def create_model(device):
    # Create Sequential model with 3 linear layers
    model = nn.Sequential(
        nn.Linear(4, 3, bias=False),  # Layer 1: 4 -> 3
        nn.ReLU(),
        nn.Linear(3, 3, bias=False),  # Layer 2: 3 -> 3
        nn.ReLU(),
        nn.Linear(3, 2, bias=False)   # Layer 3: 3 -> 2
    ).to(device)

    # Initialize with known weights for predictable gradients
    with torch.no_grad():
        # Layer 1 weights: 4x3
        model[0].weight.data = torch.tensor([[1.0, 2.0, 3.0, 4.0],
                                            [0.5, 1.5, 2.5, 3.5],
                                            [0.1, 0.2, 0.3, 0.4]])
        # Layer 2 weights: 3x3
        model[2].weight.data = torch.tensor([[1.0, 0.0, 0.0],
                                            [0.0, 1.0, 0.0],
                                            [0.0, 0.0, 1.0]])
        # Layer 3 weights: 3x2
        model[4].weight.data = torch.tensor([[1.0, 2.0, 3.0],
                                            [4.0, 5.0, 6.0]])
    model.to(device)
    return model


def main():
    use_gpu = int(os.environ.get('USE_GPU', '0'))
    # Initialize process group with gloo backend for CPU
    dist.init_process_group("gloo" if not use_gpu else "nccl")

    rank = dist.get_rank()
    world_size = dist.get_world_size()

    print(f"Starting rank {rank} of {world_size}")

    if use_gpu:
        torch.cuda.set_device(rank)
    device = torch.device("cpu" if not use_gpu else f"cuda:{rank}")

    input_data = torch.tensor([[1.0, 0.5, 0.2, 0.1]], device=device)  # Single sample for simplicity
    target = torch.tensor([[0.0, 0.0]], device=device)

    model_ref = create_model(device)
    (model_ref(input_data) - target).mean().backward()
    ref_grads = {name: param.grad.clone() for name, param in model_ref.named_parameters()}
    if rank == 0:
        print("Reference gradients:")
        for name, grad in ref_grads.items():
            print(f"  {name}: {grad}")

    # Create device mesh for FSDP2
    device_mesh = init_device_mesh("cpu" if not use_gpu else "cuda", (world_size,))
    # Create model
    model = create_model(device)
    # Apply FSDP2 to each linear layer separately
    fully_shard(model[0], mesh=device_mesh)  # First linear layer
    fully_shard(model[2], mesh=device_mesh)  # Second linear layer
    fully_shard(model[4], mesh=device_mesh)  # Third linear layer

    # Forward pass
    output = model(input_data)
    # Use different loss scales per rank for easier comparison
    loss_scale = int(rank + 1.0)
    print(f"Rank {rank}: Using loss scale = {loss_scale}")

    # MSE loss with different scales
    loss = loss_scale * (output - target).mean()
    print(f"Rank {rank}: Scaled loss: {loss.item()}")
    loss.backward()

    # Check gradients after FSDP reduction for each layer
    print(f"\nRank {rank}: Gradients after FSDP2 reduction:")

    layer_names = ['Layer 0 (4->3)', 'Layer 2 (3->3)', 'Layer 4 (3->2)']
    linear_layers = [model[0], model[2], model[4]]

    for i, (layer, name) in enumerate(zip(linear_layers, layer_names)):
        if hasattr(layer, 'weight') and layer.weight.grad is not None:
            grad = layer.weight.grad
            grad = grad.full_tensor()
            if rank == 0:
                print(f"  {name}: {grad}")
                print(f"    Shape: {grad.shape}")

    dist.barrier()
    dist.destroy_process_group()


if __name__ == "__main__":
    main()
```

The above script runs a FSDP model's backward pass and prints the reduced gradients. Run it with:
```
USE_GPU=0  torchrun --nproc_per_node 2 a.py
USE_GPU=1  torchrun --nproc_per_node 2 a.py
```

It can be observed that the gradients are 2x larger when running with USE_GPU=0:
```
  Layer 0 (4->3): tensor([[ 7.5000,  3.7500,  1.5000,  0.7500],
        [10.5000,  5.2500,  2.1000,  1.0500],
        [13.5000,  6.7500,  2.7000,  1.3500]])
    Shape: torch.Size([3, 4])
  Layer 2 (3->3): tensor([[22.5000, 15.7500,  2.2500],
        [31.5000, 22.0500,  3.1500],
        [40.5000, 28.3500,  4.0500]])
    Shape: torch.Size([3, 3])
  Layer 4 (3->2): tensor([[4.5000, 3.1500, 0.4500],
        [4.5000, 3.1500, 0.4500]])
    Shape: torch.Size([2, 3])
```
With USE_GPU=1:
```
  Layer 0 (4->3): tensor([[3.7500, 1.8750, 0.7500, 0.3750],
        [5.2500, 2.6250, 1.0500, 0.5250],
        [6.7500, 3.3750, 1.3500, 0.6750]], device='cuda:0')
    Shape: torch.Size([3, 4])
  Layer 2 (3->3): tensor([[11.2500,  7.8750,  1.1250],
        [15.7500, 11.0250,  1.5750],
        [20.2500, 14.1750,  2.0250]], device='cuda:0')
    Shape: torch.Size([3, 3])
  Layer 4 (3->2): tensor([[2.2500, 1.5750, 0.2250],
        [2.2500, 1.5750, 0.2250]], device='cuda:0')
    Shape: torch.Size([2, 3])
```

It appears that when using CPU, the gradients are summed rather than averaged.

### Versions

torch 2.8

cc @ezyang @gchanan @zou3519 @kadeng @msaroufim @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta

Vinayak-Pawar:
Fixes #162338 

This pull request addresses a bug in the composable FSDP implementation where gradients were being summed instead of averaged when using the gloo backend for CPU-based training. This led to incorrect gradient values that were scaled by the world size.
The root cause was an incorrect assumption that the gloo backend supported ReduceOp.AVG for reduce_scatter operations. The fix explicitly checks for the gloo backend and forces the use of ReduceOp.SUM followed by a manual division. This ensures that the gradient reduction is mathematically correct and consistent with the nccl backend's behavior.

Test Plan
A new unit test, test_fsdp_gradient_reduction_cpu, has been added to test_fsdp_core.py. This test verifies that the gradient averaging is performed correctly on the CPU by comparing the FSDP-computed gradients against a non-distributed reference model, ensuring the issue is resolved and preventing future regressions.


ppwwyyxx:
Thanks for the quick contribution!

If reduce-scatter-avg is not supported in GLOO, why was an error not thrown? It might be worth digging into why it did not fail more loudly as there is probably an assertion missing at some place.

Skylion007 (dismissed):

ppwwyyxx:
Did a quick test and it seems `dist.reduce_scatter_tensor(op=torch.distributed.ReduceOp.AVG)` silently does a sum for gloo backend.

That should be either fixed or raise an exception. cc @d4l3k from #149869

awgu:
@Skylion007 there seems to be some onnx changes in this PR that I am not sure if they are intended. Also, the unit test is using FSDP1, not FSDP2.

Skylion007:
@awgu I recognize those ONNX changes, I just merged it. The PR just needs to be rebased

Skylion007:
Good catch on the FSDP1/FSDP2 APIs though. ;-;

Skylion007 (dismissed):
codepath needs to test FSDP2, not FSDP1

Vinayak-Pawar:
Thanks for the feedback! I've updated the test to use the FSDP2 (fully_shard) API as requested. I ran into some issues with my local build environment when trying to run the test suite, but I believe the changes are correct. Please let me know if there are any other changes needed.

ppwwyyxx:
The PR content is still testing for FSDP1. Did you forget to include it? Also please verify that the new test would fail without your fix.

Vinayak-Pawar:
Hi @ppwwyyxx , thanks for the follow-up. I'm not sure why the FSDP2 changes weren't visible, as I can see them pushed on my end.

dcci:
+ Tristan to review

d4l3k:
@Vinayak-Pawar it doesn't seem like this PR has any changes? Did something get lost in rebasing/merges?

d4l3k (changes requested):
no changes?

PR #162349, opened 2025-09-07
[MPS] mps sparse mul op implementation

Isalia20:
Implements mps sparse mul operation as well as enables other operations such as:
1. copy_
2. div
3. sum
4. floor
5. power
6. sub
7. floor_divide

cc @alexsamardzic @nikitaved @pearu @cpuhrsch @amjames @bhosmer @jcaip @kulinseth @albanD @malfet @DenisVieriu97 @jhavukainen


aten/src/ATen/native/sparse/mps/SparseMPSTensorMath.mm
```diff
@@ -20,10 +21,268 @@
...
+  }
+
+  TORCH_CHECK(dense.sizes().equals(sparse.sizes()),
+              "mul(dense, sparse): sizes must match exactly (no broadcasting): ",
+              dense.sizes(), " vs ", sparse.sizes());
+
+  const int64_t nDim  = sparse.dim();
+  const int64_t nDimI = sparse.sparse_dim();
+  TORCH_CHECK(nDimI >= 0 && nDimI <= nDim,
+              "Invalid sparse_dim=", nDimI, " for dense tensor of dim ", nDim);
```
pearu: This does not make sense as `sparse.sparse_dim() <= sparse.dim()` is never false.
Did you mean
```
const int64_t nDim  = dense.dim();
```
?


Also, `sparse_dim()` is never negative (in fact, it can be zero only when `dim()` is zero and that is covered by the `scalar_like==true` path above). So, testing for `nDimO <= nDim` will be sufficient.

aten/src/ATen/native/sparse/mps/SparseMPSTensorMath.mm
```diff
@@ -20,10 +21,268 @@
...
+  auto rhs_keys = flatten_indices(rhs_indices, rhs.sizes());
+
+  // Intersect sorted keys (search the shorter in the longer)
+  const bool A_is_lhs = (lhs_nnz <= rhs_nnz);
+  const int64_t lenA = A_is_lhs ? lhs_nnz : rhs_nnz;
+  const int64_t lenB = A_is_lhs ? rhs_nnz : lhs_nnz;
+  auto A_keys = A_is_lhs ? lhs_keys : rhs_keys;
+  auto B_keys = A_is_lhs ? rhs_keys : lhs_keys;
+
+  const int64_t mmax = std::min(lhs_nnz, rhs_nnz);
```
pearu: Nit: notice that `mmax == lenA` always holds. So, you can eliminate `mmax`.

pearu (changes requested):
I have couple of nits and a typo report. Otherwise, LGTM! Thanks, @Isalia20!

aten/src/ATen/native/sparse/mps/SparseMPSTensorMath.mm
Isalia20: Updated with dense dim, thanks

aten/src/ATen/native/sparse/mps/SparseMPSTensorMath.mm
Isalia20: Nice catch, thank you. Updated

pearu (dismissed):
LGTM!

aten/src/ATen/native/sparse/mps/kernels/Mul.metal
```diff
@@ -0,0 +1,198 @@
...
+    long sz_d  = sizes[d];
+    key = key * sz_d + idx_d;
+  }
+
+  ulong dense_idx = (ulong)key * (ulong)view_cols + (ulong)col;
+  ulong val_idx = (ulong)i * (ulong)view_cols + (ulong)col;
+
+  float a = (float)values[val_idx];
+  float b = (float)dense[dense_idx];
+  out_values[val_idx] = (T)(a * b);
```
malfet: Please use C++ style casts, or function style ones
```suggestion
  out_values[val_idx] = static_cast<T>(a * b);
```

aten/src/ATen/native/sparse/mps/kernels/Mul.metal
```diff
@@ -0,0 +1,198 @@
...
+    long idx_d = indices[(ulong)d * (ulong)nnz + (ulong)i];
+    long sz_d  = sizes[d];
+    key = key * sz_d + idx_d;
+  }
+
+  ulong dense_idx = (ulong)key * (ulong)view_cols + (ulong)col;
+  ulong val_idx = (ulong)i * (ulong)view_cols + (ulong)col;
+
+  float a = (float)values[val_idx];
+  float b = (float)dense[dense_idx];
```
malfet:
```suggestion
  const auto a = static_cast<float>(values[val_idx]);
  const auto b = static_cast<float>(dense[dense_idx]);
```

aten/src/ATen/native/sparse/mps/kernels/Mul.metal
```diff
@@ -0,0 +1,198 @@
...
+    constant uint& view_cols      [[buffer(7)]],
+    uint3 gid                     [[thread_position_in_grid]])
+{
+  uint col = gid.x;
+  uint i = gid.z;
+
+  long key = 0;
+  for (uint d = 0; d < nDimI; ++d) {
+    long idx_d = indices[(ulong)d * (ulong)nnz + (ulong)i];
+    long sz_d  = sizes[d];
```
malfet:
```suggestion
    const auto sz_d  = sizes[d];
```

malfet (dismissed):
Please submit cleanup PR later, which is more consistent with the coding style

aten/src/ATen/native/sparse/mps/kernels/Mul.metal
```diff
@@ -0,0 +1,198 @@
...
+template <typename T>
+kernel void dense_sparse_mul_kernel(
+    device const T* dense         [[buffer(0)]],
+    device const T* values        [[buffer(1)]],
+    device T* out_values          [[buffer(2)]],
+    device const long* indices    [[buffer(3)]],
+    device const long* sizes      [[buffer(4)]],
+    constant uint& nnz            [[buffer(5)]],
+    constant uint& nDimI          [[buffer(6)]],
+    constant uint& view_cols      [[buffer(7)]],
```
malfet: We've talked in the high is the overhead of passing one scalar as argument buffer. Please merge those in followup PR
```suggestion
    constant uint3& nnz_dim_viewcols            [[buffer(5)]],
```

aten/src/ATen/native/sparse/mps/kernels/Mul.metal
```diff
@@ -0,0 +1,198 @@
...
+
+template <typename T>
+kernel void dense_sparse_mul_kernel(
+    device const T* dense         [[buffer(0)]],
+    device const T* values        [[buffer(1)]],
+    device T* out_values          [[buffer(2)]],
+    device const long* indices    [[buffer(3)]],
+    device const long* sizes      [[buffer(4)]],
+    constant uint& nnz            [[buffer(5)]],
+    constant uint& nDimI          [[buffer(6)]],
```
malfet: Stick to `snake_case` for variable names
```suggestion
    constant uint& ndim_i          [[buffer(6)]],
```

aten/src/ATen/native/sparse/mps/kernels/Mul.metal
```diff
@@ -0,0 +1,198 @@
...
+    constant uint& view_cols      [[buffer(7)]],
+    uint3 gid                     [[thread_position_in_grid]])
+{
+  uint col = gid.x;
+  uint i = gid.z;
+
+  long key = 0;
+  for (uint d = 0; d < nDimI; ++d) {
```
malfet:
```suggestion
  for (uint d = 0; d < ndim_i; ++d) {
```

test/test_sparse.py
```diff
@@ -1693,18 +1693,19 @@ def fn(S, D):
     @gradcheck_semantics()
     def test_sparse_mul(self, device, dtype, coalesced, gradcheck):
         # https://github.com/pytorch/pytorch/issues/79914
+        atol = 0.02 if device == "mps:0" else 1e-5  # higher atol for mps since we run gradcheck in float32
```
nikitaved: That is a lot. Would it be possible to change to float64?

Isalia20: Agree it's a lot, mps doesn't support float64 unfortunately, get similar error margin when running on CPU as well

malfet: But it still feels like a lot, are you sure the algorithm you are using yields correct results?

Isalia20: Yes here is the check where we can see that mps/cpu produce same error when running in float32

cpu
mps

nikitaved: Testing gradcheck with float32 does not look good to me. float64 is missing for MPS, I wonder? Alternatively, one can check the grads manually, maybe.

pearu: I can reproduce the exception above on a linux box for float32 tensors. So, this may not be MPS related.

Resolution: instead of lowering `atol`, use `eps=1e-4` as in
```
diff --git a/test/test_sparse.py b/test/test_sparse.py
index 6c74b73945e..b7b660ceaf7 100644
--- a/test/test_sparse.py
+++ b/test/test_sparse.py
@@ -1661,7 +1661,7 @@ class TestSparse(TestSparseBase):
         # https://github.com/pytorch/pytorch/issues/79914
         a = torch.tensor([[0., 1]], dtype=dtype, device=device).to_sparse().requires_grad_(True)
         b = torch.tensor([[0., 1]], dtype=dtype, device=device).to_sparse().requires_grad_(True)
-        gradcheck(lambda x, y: torch.sparse.sum(x * y).to_dense(masked_grad=gradcheck.masked), [a, b])
+        gradcheck(lambda x, y: torch.sparse.sum(x * y).to_dense(masked_grad=gradcheck.masked), [a, b], eps=1e-4)
```
that makes the tests pass for both float32 and float64 tensors.

Isalia20: Thank you, updated

pearu (approved):
LGTM! Thanks, @Isalia20!

malfet (approved):

PR #162354, opened 2025-09-07
are_strides_like_channels_last_or_false

laithsakka:
* __->__ #162354

Note this could change suggest_memory_format behaviour for unbacked

we used to return True for are_strides_like_channels_last sometimes even when results undecided
now when its not decided we return False.


torch/_prims_common/__init__.py
```diff
@@ -1990,28 +1996,32 @@ def are_strides_like_channels_last(
     else:
         return False
 
-    if guard_size_oblivious(strides[1] == 0):
+    # if a stride is 0, then only if the size is 0 its ok
+    if guard_or_true(strides[1] == 0):
```
laithsakka, line 1999: note that those behave different than previous guard_size_oblivious
but i think its better like this?

ezyang:
So are we committing to ALL the contiguity functions implicitly guard_or_false, or is it just this one? Can we document this somewhere?

laithsakka:
> So are we committing to ALL the contiguity functions implicitly guard_or_false, or is it just this one? Can we document this somewhere?

As for are_strides_like_channels_last, I changed it to are_strides_like_channels_last_or_false and added a comment. so its kind of explicit. Note: The C++ version of are_strides_like_channels_last would still throw DDE now, but it seems like we never had to handle it (yet). [is_channels_last_strides_3d and is_channels_last_strides_2d do not have 
guard_size_oblivious].


One alternative approach here is we could also decide to assume stride or size is not 0 at some checks and make is_strides_like_channels_last return true (like what guard_size_oblivious used to do). But in that case, it sounds like we SHOULD add runtime assertions that strides[1] != 0
and (shape[d] != 0) for each d.

I'm not sure what's better from a model writer's perspective.

With the current approach, authors can always opt into this by torch checking that tensors are never empty or strides are never 0 (note if strides are crafted carefully, we should see max(1, u0) and already know it's not 0 when it's always not 0).

Furthermore, maybe in the future, we can add APIs:
t.assume_never_empty()
t.assume_always_contiguous()
t.assume_always_channels_last()

for cases where modelers know for sure some properties always hold and there is a risk that unbacked semantics can make them false.

aorenste:
> So are we committing to ALL the contiguity functions implicitly guard_or_false, or is it just this one? Can we document this somewhere?

Right now we end up guarding on kind of a strict expression where we'll fail the guard even if the is_contiguous() of a tensor remains true/false (especially when broadcast or 0 is involved). Would it be beneficial to add some explicit contiguity guard types? So instead of guarding on some big expression with a bunch of "!= 1" checks we just guard on "is_contiguous()". I suspect it would make the guards a lot simpler and often would cover more things without a recompile.

@ezyang thoughts?

ezyang:
I am 100% in favor of adding sympy functions that black box things like IsContiguous, in the same way it was done for NonOverlappingInDense. I don't want a ton of API functions. I want to be able to say `guard_or_false(t.is_contiguous())` and the obvious thing happens.

laithsakka:
> I am 100% in favor of adding sympy functions that black box things like IsContiguous, in the same way it was done for NonOverlappingInDense. I don't want a ton of API functions. I want to be able to say `guard_or_false(t.is_contiguous())` and the obvious thing happens.

I like that idea also i think c++ contiguity checks are almost like that now, but maybe can do better refactor even 
but i still think even in that world we would need ways for user to 
tell us 
this inputs are contiguous, something like torch._checks but over those new functions/properties

laithsakka:
> > I like that idea also i think c++ contiguity checks are almost like that now, but maybe can do better refactor even but i still think even in that world we would need ways for user to tell us this inputs are contiguous, something like torch._checks but over those new functions/properties
> 
> Isn't that just `torch._check(t.is_contiguous())`?

mmm i mean is_contiguous() is expected to return bool that we cant change but could be 
torch._check(t.sym_is_contiguous()) sure.

aorenste:
> > > I like that idea also i think c++ contiguity checks are almost like that now, but maybe can do better refactor even but i still think even in that world we would need ways for user to tell us this inputs are contiguous, something like torch._checks but over those new functions/properties
> > 
> > 
> > Isn't that just `torch._check(t.is_contiguous())`?
> 
> mmm i mean is_contiguous() is expected to return bool that we cant change but could be torch._check(t.sym_is_contiguous()) sure.

We can't return SymBool when inputs are symbolic?

aorenste (approved):
Approving to unblock assuming we'll move toward special contiguity symbolics in the future...

PR #162030, opened 2025-09-03
[inductor] turn on loaf (for oss) by default

shunting314:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #162030

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela

PR #162355, opened 2025-09-08
[Inductor] do loop reordering in a separate final round

shunting314:
* #162102
* #162030
* __->__ #162355
* #162126
* #162101

Previous LOAF after fusion algorithm is not guaranteed to create more fusion opportunities even if loop reordering happens. I can not find an example that LOAF reduce the amount of fusion, but here is an example that reordering loops does not add more fusions:

https://github.com/pytorch/pytorch/blob/a1f7639922ee0470bd7109bab6fe62989cf5000d/test/inductor/test_loop_ordering.py#L612-L641

Move LOAF to a separate final round of fusion so that we are guaranteed to not reducing the amount of fusions. Hopefully this also helps compilation time since LOAF kicks in when there are less nodes.


eellison (commented):
What do you think about including reordering in the fusion score key, and not actually making the change to the loop body in the scoring round. 

This would allow other fusions to occur following reordering. We have other, similar loop body modifications we want to do (@BoyuanFeng's masking, and I have another use case here: https://github.com/pytorch/pytorch/pull/161404) and this could allow those use cases to compose.

shunting314:
@eellison 

> What do you think about including reordering in the fusion score key, and not actually making the change to the loop body in the scoring round.

I got what you mean. But the major blocker for turning LOAF on by default right now is the increase of compilation time and I'm worried this would further increase that.  There are a couple of ways to decide fusion-score-key without changing loop order
1. a most straightforward way is to make a clone of the SchedulerNode , reorder the loops, decide the fusion score and drop the clone. This would definitely increase compilation time further
2. a more sophisticated way is to pretend doing loop reordering without cloning the SchedulerNode. This would need us to update Scheduler.score_fusion_memory to decide the share of memory access for a specific loop order. But then we would need add caching to avoid doing these kind of analysis later again when we perform the fusion. Also the cache may be tricky to implement since a SchedulerNode may already get reordered for fusing another pair of nodes.

So I don't find a good way to do what you said without potentially further increasing compilation time.

> This would allow other fusions to occur following reordering. We have other, similar loop body modifications we want to do (@BoyuanFeng's masking, and I have another use case here: https://github.com/pytorch/pytorch/pull/161404) and this could allow those use cases to compose.

Can these fusion all be done in the final round and compose with LOAF? In high level the process of fusion would become
1. iteratively find regular fusions
2. a final round to apply other extra fusions with: LOAF, index-inversion (your PR) etc.

eellison (approved):
Sounds good - we can discuss other potential improvements in the future but I think this is a clear improvement so we should land.

jansel (approved):

PR #162360, opened 2025-09-08
[optim] avoid aliasing lr and initial_lr in `SequentialLR.__init__`

### LINKED ISSUE #162359: SequentialLR aliases lr Tensor with initial_lr, corrupting base_lrs for chained schedulers
### 🐛 Describe the bug

1. When we use an optimizer with a Tensor learning rate (as [recommended by the docs to avoid recompilation](https://docs.pytorch.org/tutorials/recipes/compiling_optimizer_lr_scheduler.html)), `SequentialLR.__init__` assigns `group["lr"] = group["initial_lr"]`, aliasing the tensors.
2. The initial step of the first scheduler writes the lr in-place, which also mutates the `initial_lr` and all chained schedulers' `base_lrs` that reference it.
3. Chained schedulers which rely on `base_lrs` therefore start from the wrong base (e.g. scaled by `start_factor` if we used a `LinearLR` first).

For example, if we use a Tensor learning rate and `SequentialLR` with `schedulers=[LinearLR, CosineAnnealingLR]`, the `base_lrs` of our chained schedulers (and thus our `CosineAnnealingLR` outputs) are scaled by the `LinearLR`'s `start_factor`:

Image

`LinearLR.get_lr` does not rely on `base_lrs`, hence the jump at the milestone. With the bug fixed, we obtain the intended schedule:

Image

Here's a reproduction:
```python
import torch
from torch import optim
import matplotlib.pyplot as plt

# 1. Use a tensor learning rate. The choice of optimizer doesn't matter.
lr = 1.0
x = torch.tensor(0.0, device='meta')
opt = optim.AdamW([x], lr=torch.tensor(lr))

# 2. Initialize our chained schedulers.
milestone, total_steps = 40, 100
start_factor, end_factor = 0.2, 1.0
warmup = optim.lr_scheduler.LinearLR(opt, start_factor, end_factor, total_iters=milestone)
decay = optim.lr_scheduler.CosineAnnealingLR(opt, T_max=total_steps-milestone)

# Note that each scheduler's base_lr aliases with the optimizer's initial_lr.
assert warmup.base_lrs[0].is_set_to(opt.param_groups[0]['initial_lr'])
assert decay.base_lrs[0].is_set_to(opt.param_groups[0]['initial_lr'])

# 3. Initialize our SequentialLR.
scheduler = optim.lr_scheduler.SequentialLR(opt, schedulers=[warmup, decay], milestones=[milestone])

# SequentialLR.__init__ aliases our optimizer's lr and initial_lr.
# for group in self.optimizer.param_groups:
#     group["lr"] = group["initial_lr"]
assert opt.param_groups[0]['lr'].is_set_to(opt.param_groups[0]['initial_lr'])

# Which means they're also aliased with the base_lrs of each scheduler!
assert (warmup.base_lrs[0].data_ptr() 
    == decay.base_lrs[0].data_ptr()
    == opt.param_groups[0]['initial_lr'].data_ptr() 
    == opt.param_groups[0]['lr'].data_ptr())

# It also calls _initial_step() on our LinearLR, setting all tensors to start_factor * lr.
assert (start_factor * lr
    == opt.param_groups[0]['lr']
    == opt.param_groups[0]['initial_lr']
    == warmup.base_lrs[0]
    == decay.base_lrs[0])

# 4. Now if we do something which breaks the aliases, e.g. save/load the optimizer's state_dict...
opt.load_state_dict(opt.state_dict())
assert not warmup.base_lrs[0].data_ptr() == opt.param_groups[0]['initial_lr'].data_ptr()

# Our base_lrs are incorrect. The LinearLR will work as intended (since get_lr updates based on the
# previous iteration, not base_lr), but the outputs of further schedulers are incorrectly scaled by 
# start_factor. In our case, these are the outputs of CosineAnnealingLR which appear after the milestone.
lrs = []
for step in range(total_steps):
    scheduler.step()
    lrs.append(scheduler.get_last_lr()[0].item())

plt.plot(lrs)
plt.show()
```

### Versions

PyTorch version: 2.8.0
Is debug build: False

OS: macOS 15.6.1 (arm64)
CMake version: version 4.1.1
Libc version: N/A

Is CUDA available: False
CUDA runtime version: No CUDA
cuDNN version: No CUDA
Is XPU available: False
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M3 Pro

[pip3] Could not collect
[conda] No relevant packages

cc @vincentqb @jbschlosser @albanD @janeyx99 @crcrpar

filipviz:
Prevents a SequentialLR edge case which corrupted base_lrs in chained schedulers when using an optimizer with a Tensor learning rate.

Fixes #162359

### DIFF
--- a/test/optim/test_lrscheduler.py
+++ b/test/optim/test_lrscheduler.py
@@ -812,6 +812,26 @@ class TestLRScheduler(TestCase):
         targets = [single_targets, [x * 10 for x in single_targets]]
         self._test_get_last_lr(scheduler, targets, epochs)
 
+    def test_sequentiallr_does_not_alias_lr_and_initial_lr(self):
+        # The TestLRScheduler object uses self.opt to avoid instantiating a new optimizer for each test.
+        # self.opt has a float lr, and we need to use a Tensor lr to ensure that a former SequentialLR bug is fixed.
+        # For more context, see https://github.com/pytorch/pytorch/issues/162359
+        old_opt = self.opt
+        lr = torch.as_tensor(2.0)
+        self.opt = SGD(self.net.parameters(), lr=lr)
+        milestone = 4
+        epochs = 8
+        start, end = 0.1, 0.8
+
+        schedulers = [None] * 2
+        schedulers[0] = LinearLR(self.opt, start, end, total_iters=milestone)
+        schedulers[1] = LinearLR(self.opt, end, start, total_iters=epochs - milestone)
+        targets = [[0.2, 0.55, 0.9, 1.25, 1.6, 1.25, 0.9, 0.55]]
+
+        scheduler = SequentialLR(self.opt, schedulers, milestones=[milestone])
+        self._test(scheduler, targets, epochs)
+        self.opt = old_opt
+
     def test_chained_lr2_get_last_lr_before_step(self):
         schedulers = [
             LinearLR(self.opt, start_factor=0.4, total_iters=3),
--- a/test/optim/test_lrscheduler.py
+++ b/test/optim/test_lrscheduler.py
@@ -817,15 +817,16 @@ class TestLRScheduler(TestCase):
         # self.opt has a float lr, and we need to use a Tensor lr to ensure that a former SequentialLR bug is fixed.
         # For more context, see https://github.com/pytorch/pytorch/issues/162359
         old_opt = self.opt
-        lr = torch.as_tensor(2.0)
+        lr = torch.tensor(2.0)
         self.opt = SGD(self.net.parameters(), lr=lr)
         milestone = 4
         epochs = 8
         start, end = 0.1, 0.8
 
-        schedulers = [None] * 2
-        schedulers[0] = LinearLR(self.opt, start, end, total_iters=milestone)
-        schedulers[1] = LinearLR(self.opt, end, start, total_iters=epochs - milestone)
+        schedulers = [
+            LinearLR(self.opt, start, end, total_iters=milestone),
+            LinearLR(self.opt, end, start, total_iters=epochs - milestone),
+        ]
         targets = [[0.2, 0.55, 0.9, 1.25, 1.6, 1.25, 0.9, 0.55]]
 
         scheduler = SequentialLR(self.opt, schedulers, milestones=[milestone])
--- a/torch/optim/lr_scheduler.py
+++ b/torch/optim/lr_scheduler.py
@@ -887,7 +887,10 @@ class SequentialLR(LRScheduler):
 
         # Reset learning rates back to initial values
         for group in self.optimizer.param_groups:
-            group["lr"] = group["initial_lr"]
+            if isinstance(group["lr"], Tensor):
+                group["lr"].fill_(_to_scalar(group["initial_lr"]))
+            else:
+                group["lr"] = group["initial_lr"]
 
         # "Undo" the step performed by other schedulers
         self.recursive_undo()


filipviz:
Note that the test I added isn't exhaustive. If there's interest, I can expand `test_lrscheduler.py` to test Tensor lrs more thoroughly. A quick skim suggests there may be similar aliasing issues in `ReduceLROnPlateau._reduce_lr`, `CosineAnnealingWarmRestarts.step`, and `OneCycleLR.__init__`.

Patterns like the one I used in this PR:
```python
if isinstance(group["lr"], Tensor):
    group["lr"].fill_(_to_scalar(group["initial_lr"]))
else:
    group["lr"] = group["initial_lr"]
```
Appear in `LRScheduler._update_lr` and `CyclicLR.__init__`. Maybe it'd be best to create a helper function and use it in all of these places.

torch/optim/lr_scheduler.py
```diff
@@ -887,7 +887,10 @@ def __init__(
 
         # Reset learning rates back to initial values
         for group in self.optimizer.param_groups:
-            group["lr"] = group["initial_lr"]
+            if isinstance(group["lr"], Tensor):
+                group["lr"].fill_(_to_scalar(group["initial_lr"]))
+            else:
+                group["lr"] = group["initial_lr"]
```
janeyx99, line 893:
```suggestion
            group["lr"].fill_(group["initial_lr"])
```
Does this just work? I suspect it may.

test/optim/test_lrscheduler.py
```diff
@@ -812,6 +812,26 @@ def test_get_last_lr_sequentiallr(self):
         targets = [single_targets, [x * 10 for x in single_targets]]
         self._test_get_last_lr(scheduler, targets, epochs)
 
+    def test_sequentiallr_does_not_alias_lr_and_initial_lr(self):
+        # The TestLRScheduler object uses self.opt to avoid instantiating a new optimizer for each test.
+        # self.opt has a float lr, and we need to use a Tensor lr to ensure that a former SequentialLR bug is fixed.
+        # For more context, see https://github.com/pytorch/pytorch/issues/162359
+        old_opt = self.opt
+        lr = torch.as_tensor(2.0)
```
janeyx99:
```suggestion
        lr = torch.tensor(2.0)
```
why not this?

test/optim/test_lrscheduler.py
```diff
@@ -812,6 +812,26 @@ def test_get_last_lr_sequentiallr(self):
...
+        self.opt = SGD(self.net.parameters(), lr=lr)
+        milestone = 4
+        epochs = 8
+        start, end = 0.1, 0.8
+
+        schedulers = [None] * 2
+        schedulers[0] = LinearLR(self.opt, start, end, total_iters=milestone)
+        schedulers[1] = LinearLR(self.opt, end, start, total_iters=epochs - milestone)
```
janeyx99:
```suggestion
        schedulers = [
            LinearLR(self.opt, start, end, total_iters=milestone),
            LinearLR(self.opt, end, start, total_iters=epochs - milestone)
        ]
```
nit

janeyx99 (commented):
Thanks for this; we would appreciate a more exhaustive test + fix across lr schedulers but that can also go in followup PRs!

test/optim/test_lrscheduler.py
filipviz: No reason! This (and the odd scheduler list initialization) were done for consistency with other lr_scheduler tests.

torch/optim/lr_scheduler.py
filipviz: It wouldn't, because:
- `group["lr"]` is often a float (so `.fill_` will raise an `AttributeError`).
- The `_to_scalar` is a bit paranoid, but an optimizer can technically have an lr with ndim > 1 - e.g. `torch.tensor([[[0.2]]])`. If it does, `.fill_` will raise a `RuntimeError`.

filipviz:
@janeyx99 thanks for your feedback (and such a fast review)! I've added both suggestions to the PR.

> Thanks for this; we would appreciate a more exhaustive test + fix across lr schedulers but that can also go in followup PRs!

Sounds great - I can put something together soon. A new PR makes sense to me. Would you like me to create a separate issue as well?

torch/optim/lr_scheduler.py
janeyx99: o right right

janeyx99:
> Sounds great - I can put something together soon. A new PR makes sense to me. Would you like me to create a separate issue as well?

sg! no need for new issue!

janeyx99 (approved):

filipviz:
Closing this in favor of #163098. I'll add several further PRs off of that one. Sorry for the delay @janeyx99!

Issue #114850, opened 2023-11-30
[RFC] Add Intel GPU support into PyTorch CI/CD 


chuanqi129:
## Motivation 
As the [[RFC] Intel GPU Upstreaming](https://github.com/pytorch/pytorch/issues/114723) mentioned, to integrate the new Intel GPU device and its associated features into PyTorch, we need to implement PyTorch CI/CD tests specifically designed for Intel GPU devices. These tests will ensure the quality of incoming pull requests and gate their acceptance accordingly. 

## Design Philosophy 
For the new part CI/CD enabling, we aim to leverage the existing PyTorch CI/CD infrastructure wherever possible. Intel GPU related test will be dispatched to Intel Develop Cloud (IDC) Instances, which provides Intel GPU hardware as self-hosted runners. Based on our understanding of current PyTorch CI/CD tests, we will divide Intel GPU related tests into several categories: pull test, inductor tests and other tests. For inductor tests, we will extend the existing inductor test workflow to accommodate Intel GPU inductor testing. And a new workflow will serve as the entry point for other tests for Intel GPU, mirroring the approach used for other devices. Overall, all Intel GPU tests follow the rules below. 
- Docker based builds & tests 
- Multiple steps both for inductor and other tests 
     + Base Docker image build on AWS instance runners which provides PyTorch build and tests environment. 
     + Wheel build on AWS instance runners directly. 
     + Tests can be sharded and dispatched on IDC instance runners. 
     
### Entrance of pull test 
For the basic build test for pull requests, we will add a new part for Intel GPU specific build in `.github/workflows/pull.yml` and triggered by each pull request.  

### Entrance of Inductor tests 
For Inductor tests, we will add a new part for Intel GPU specific tests in `.github/workflows/inductor.yml` and triggered by `ciflow/inductor`. To avoid breaking other inductor related PRs at the first stage, we plan to add a new ciflow label `ciflow/xpu` in `.github/pytorch-probot.yml`, and limit the Intel GPU inductor tests only for PR which has both `ciflow/inductor` and `ciflow/xpu`.  

### Entrance of other tests 
The Intel GPU device related remain tests will be triggered by PR with label `ciflow/xpu` or regular triggered by timer. To achieve it, we will add a new entrance workflow `.github/workflows/xpu.yml`, which like below content.
```yml
name: xpu

on:
  push:
    branches:
      - main
      - release/*
    tags:
      - ciflow/xpu/*
  workflow_dispatch:
  schedule:
[...]

PR #162370, opened 2025-09-08
[WIP][xpu][test] port test/test_nn.py to Intel GPU

daisyden:
For https://github.com/pytorch/pytorch/issues/114850, we will port aten unit tests to Intel GPU. This PR will work on test/test_nn.py. We could enable Intel GPU with following methods and try the best to keep the original code styles:

1. Replaced onlyCUDA with onlyOn(['cuda', 'xpu']) for supported tests
2. Added allow_xpu=True for supported test class in test parameterization.
3. Use torch.accelerator to extend cude specific test to XPU.
4. Added skipIfXPU decorator for cases with known issues on Intel GPU
5. Enabled 'xpu' for some test pathes
6. Make get_all_device_types general for accelerators.

### DIFF
--- a/test/test_namedtensor.py
+++ b/test/test_namedtensor.py
@@ -2,7 +2,7 @@
 # ruff: noqa: F841
 import unittest
 from torch.testing._internal.common_utils import TestCase, run_tests, TEST_NUMPY
-from torch.testing._internal.common_utils import skipIfTorchDynamo
+from torch.testing._internal.common_utils import skipIfTorchDynamo, skipIfXpu
 from torch.testing._internal.common_cuda import TEST_CUDA
 from torch.testing._internal.common_device_type import get_all_device_types
 from collections import namedtuple, OrderedDict
@@ -280,6 +280,7 @@ class TestNamedTensor(TestCase):
         self.assertEqual(named_tensor.diagonal(outdim='E', dim1='B', dim2='D').names,
                          ['A', 'C', 'E'])
 
+    @skipIfXpu   # https://github.com/intel/torch-xpu-ops/issues/2023
     def test_max_pooling(self):
         def check_tuple_return(op, inputs, expected_names):
             values, indices = op(*inputs)
--- a/test/test_nn.py
+++ b/test/test_nn.py
@@ -32,21 +32,21 @@ from torch.nn import Buffer, Parameter
 from torch.nn.parallel._functions import Broadcast
 from torch.testing._internal.common_dtype import integral_types, get_all_math_dtypes, floating_types
 from torch.testing._internal.common_utils import dtype_name, freeze_rng_state, run_tests, TestCase, \
-    skipIfNoLapack, skipIfRocm, \
-    TEST_NUMPY, TEST_SCIPY, TEST_WITH_CROSSREF, TEST_WITH_ROCM, \
+    skipIfNoLapack, skipIfRocm, skipIfXpu, \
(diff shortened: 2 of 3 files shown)


test/test_nn.py
```diff
@@ -60,9 +60,15 @@
 
 AMPERE_OR_ROCM = TEST_WITH_ROCM or torch.cuda.is_tf32_supported()
 
+
+TEST_GPU = torch.cuda.is_available() or torch.xpu.is_available()
```
guangyey: why does not use `torch.accelerator.is_available`?

test/test_nn.py
```diff
@@ -60,9 +60,15 @@
...
+TEST_MULTIGPU = TEST_GPU and torch.accelerator.device_count() >= 2
```
guangyey:
```suggestion
TEST_MULTIGPU = torch.accelerator.device_count() >= 2
```
This is enough~

test/test_nn.py
```diff
@@ -9217,6 +9230,7 @@ def test_ReflectionPad_empty(self, device, dtype):
     def test_ReflectionPad_fails(self, device):
         with self.assertRaisesRegex(RuntimeError, r'Padding size 2 is not supported for 4D input tensor'):
             mod = torch.nn.ReflectionPad1d(2)
+            print(f"#### {device}", flush=True)
```
guangyey: remove debug info.

guangyey (commented):
Please fix the CI failure.

test/test_nn.py
daisyden: Updated, thanks.

test/test_nn.py
daisyden: Updated, thanks!

test/test_namedtensor.py
EikanWang, line 1: The title of this PR indicates that the PR only changes test_nn.py. Why do we need to change this test file as well?

torch/testing/_internal/common_device_type.py
```diff
@@ -1395,13 +1395,13 @@ def efail_fn(slf, *args, **kwargs):
...
 class onlyOn:
-    def __init__(self, device_type):
+    def __init__(self, device_type: Union[str, list]):
         self.device_type = device_type
 
     def __call__(self, fn):
         @wraps(fn)
         def only_fn(slf, *args, **kwargs):
-            if self.device_type != slf.device_type:
+            if slf.device_type not in self.device_type:
```
EikanWang, line 1404: The `self.device_type` could be either a string literal or a list. If the `self.device_type` is a string, things get a little tricky. Because `not in` does not guarantee an exact match between two string literals.

EikanWang (commented):
Overall, the changes LGTM. But I'm considering splitting this PR into two parts, one is for generalization by replacing `cuda` with `torch.accelerator` or `device_type`

test/test_namedtensor.py
daisyden: Because I added xpu support in get_all_device_types() and the test_max_pooling() of test_namedtensor.py called this function and got a failure. So I submitted issue and skipped the test for Intel GPU.

torch/testing/_internal/common_device_type.py
daisyden: Eikan, "not in" is an exact match:
```
>>> my_list = ["apple", "banana", "cherry"]
>>> print("app" not in my_list)
True
>>> print("apple" not in my_list)
False
```

PR #162389, opened 2025-09-08
[precompile] Fix inlined source tracking with generators.

zhxchen17:
Summary:
When compiled code has generator, code.co_firstlineno will be inconsistent with the result from inspect.getsource, which returns the toplevel enclosing code source rather than the inner code location.

In this case, it seems simpler to just use the toplevel enclosing code location rather than the co_firstlineno field.

Test Plan:
test_package.py -k test_code_with_generator

Rollback Plan:

Differential Revision: D81929751

### DIFF
--- a/torch/_dynamo/package.py
+++ b/torch/_dynamo/package.py
@@ -521,15 +521,14 @@ class CompilePackage:
             module = inspect.getmodule(code)
             if module is None:
                 continue
-            source = inspect.getsource(code)
-            lastlineno = code.co_firstlineno + len(inspect.getsourcelines(code)[0])
-            assert source == "".join(
-                _get_sourcelines(module, code.co_firstlineno, lastlineno)
-            )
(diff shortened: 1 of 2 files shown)


torch/_dynamo/package.py
```diff
@@ -521,15 +521,14 @@ def add_inlined_source(self, sources: list[types.CodeType]) -> None:
             module = inspect.getmodule(code)
             if module is None:
                 continue
-            source = inspect.getsource(code)
-            lastlineno = code.co_firstlineno + len(inspect.getsourcelines(code)[0])
-            assert source == "".join(
-                _get_sourcelines(module, code.co_firstlineno, lastlineno)
-            )
+            sourcelines, firstlineno = inspect.getsourcelines(code)
```
dolpm, line 524: yeah, nice. this gets rid of a call to getsource too

dolpm (approved):

hrithickcodesai (approved):

PR #162393, opened 2025-09-08
In-tree registration for MTIA kl_div kernel

trirpi:
Summary: Updated kl_div to be registered correctly.

Differential Revision: D81819070

cc @egienvalue

### DIFF
--- a/aten/src/ATen/native/native_functions.yaml
+++ b/aten/src/ATen/native/native_functions.yaml
@@ -3275,6 +3275,9 @@
   manual_cpp_binding: True
 
 - func: kl_div(Tensor self, Tensor target, int reduction=Mean, *, bool log_target=False) -> Tensor
+  dispatch:
+    CompositeImplicitAutograd: kl_div
+    MTIA: kl_div_mtia
 
 - func: kron(Tensor self, Tensor other) -> Tensor
   variants: function, method


aten/src/ATen/native/native_functions.yaml
```diff
@@ -3275,6 +3275,9 @@
   manual_cpp_binding: True
 
 - func: kl_div(Tensor self, Tensor target, int reduction=Mean, *, bool log_target=False) -> Tensor
+  dispatch:
+    CompositeImplicitAutograd: kl_div
+    MTIA: kl_div_mtia
```
andyanwang, line 3280: Hi @albanD , @malfet , this should keep the dispatching behavior unchanged for non-MTIA devices, right?

malfet: I don't think you can mix and match dispatch keys like that. If it's composite explicit autograd, it must be like that for all dispatch keys. But I could be wrong

trirpi: @malfet Not sure how this is mixing/matching. All autograd here is implicit for all keys.

Note, the above change is similar to how MPS is dispatched for linear:
```
- func: linear(Tensor input, Tensor weight, Tensor? bias=None) -> Tensor
  python_module: nn
  dispatch:
    CompositeImplicitAutograd: linear
    NestedTensorCPU, NestedTensorHPU, NestedTensorCUDA: nested_linear
    MPS: _mps_linear
```
Could you elaborate?

andyanwang: @malfet IIUC CompositeImplicitAutograd works as the default for any device types that are not listed here. For the device types that are explicitly listed, like the `MTIA` here, or the `MPS` in the example of `func: linear`, pytorch will dispatch to its own function.

albanD: You can mix and match yes but why would you want to do that rather than implement the constituent ops.
As you've done it here, you won't be able to get any gradients through this op.

andyanwang: I see, yeah, personally I also prefer to go with the existing decomposition approach. Let me circle back with mtia-kernel team to check if there's other considerations why we might prefer the other way. Thanks @albanD!

srsuryadev: > You can mix and match yes but why would you want to do that rather than implement the constituent ops.
As you've done it here, you won't be able to get any gradients through this op.

@albanD  Mainly, we wanted to have support non-decomposed version as well for performance reasons.

albanD: If you want compile, autograd and other systems to work, you will need to do quite a bit more work. Note that because there are no OSS CI for this, we can't get signal here...

If you don't care about these, then sure.

andyanwang: Hi @albanD , confirmed with Kernel team that we do want non-decomposed version, not just for this specific operator, but may also for other operators in future.

And we can probably just go with the old out-of-tree registration approach for these operators for flexibility. But we found somehow the out-of-tree registration doesn't work in this case, i.e. the op still got dispatched to the Aten native implementation, instead of the out-of-tree implementation. 

By looking at torchgen, it seems the problem is that, pytorch prevents a dispatch table when an op has no any dispatch key listed in native_functions.yaml file. And it seems ignore all distpatch key(even the key is registered out-of-tree). @albanD is my understanding correct? 
https://github.com/pytorch/pytorch/blob/main/torchgen/model.py#L804-L806

PR #162394, opened 2025-09-08
[SymmMem] Use global pe for put and get

kwen2501:
* __->__ #162394
* #162320

NVSHMEM put/get APIs take global PE instead of local counterpart. So we'd need to do a translation within the kernel.

Also added a sub-group test for dispatch and combine mimic'ing the Expert Parallel cases.

### DIFF
--- a/torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
+++ b/torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
@@ -180,15 +180,19 @@ __device__ int64_t prefixSum(int64_t *odata, int64_t *idata, int n) {
 // - input splits (IN)
 // - output splits (OUT) and
 // - source offsets (OUT).
-__global__ void exchangeSplitAndOffset(int64_t* in_out_splits, int mype, int npes) {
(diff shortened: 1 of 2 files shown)


torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
```diff
@@ -276,14 +286,18 @@ at::Tensor all_to_all_vdev(
   void* output_ptr = out.mutable_data_ptr();
   int64_t* splits_ptr = (int64_t*)(in_out_splits.mutable_data_ptr());
 
-  auto stream = at::cuda::getCurrentCUDAStream(input.device().index());
+  TORCH_CHECK(input.device() == out.device());
```
Skylion007: Should be ValueError
```suggestion
  TORCH_CHECK_VALUE(input.device() == out.device());

kwen2501: Thanks, I changed it to `TORCH_CHECK_EQ`

ngimel (approved):

fegin (approved):

PR #162411, opened 2025-09-08
Give linear an explicit autograd formula always

ezyang:
* __->__ #162411

### DIFF
--- a/aten/src/ATen/functorch/BatchRulesLinearAlgebra.cpp
+++ b/aten/src/ATen/functorch/BatchRulesLinearAlgebra.cpp
@@ -109,6 +109,19 @@ static std::tuple<Tensor, std::optional<int64_t>> bmm_batch_rule(
   return std::make_tuple( at::matmul(self_, other_), 0 );
 }
 
+
+static std::tuple<Tensor, std::optional<int64_t>> linear_batch_rule(
+    const Tensor& self, std::optional<int64_t> self_bdim,
+    const Tensor& weight, std::optional<int64_t> weight_bdim,
+    const std::optional<Tensor>& bias, std::optional<int64_t> bias_bdim
+    ) {
+  // TODO: add error checking
+  auto self_ = moveBatchDimToFront(self, self_bdim);
+  auto weight_ = moveBatchDimToFront(weight, weight_bdim);
+  auto bias_ = bias.has_value() ? std::make_optional<Tensor>(moveBatchDimToFront(*bias, bias_bdim)) : std::nullopt;
+  return std::make_tuple( at::linear(self_, weight_, bias_), 0 );
+}
+
 // AFAICT, nothing here can be batched. So we decompose :)
 Tensor addmv_decomp(
   const Tensor& input, const Tensor& mat, const Tensor& vec, const Scalar& beta, const Scalar& alpha) {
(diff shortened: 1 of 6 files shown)


aten/src/ATen/native/Linear.cpp
```diff
@@ -893,4 +895,37 @@ Tensor &tensordot_out(const Tensor& input1, const Tensor& input2, IntArrayRef di
...
+
+  // dL/dx = dL/dy @ W
+  if (output_mask[0]) {
+    grad_input = at::matmul(grad_output, weight);            // [..., out] @ [out, in] -> [..., in]
+  }
+
+  // dL/dW = sum_batch ( (dL/dy)ᵀ @ x )
+  // Use einsum to contract over all leading dims without reshaping:
+  if (output_mask[1]) {
+    grad_weight = at::einsum("...o,...i->oi", {grad_output, self}); // [out, in]
```
tianyu-l, line 913: wouldn't this go into decomposition, since `linear_backward` is `CompositeImplicitAutograd` now?

tools/autograd/derivatives.yaml
```diff
@@ -2355,6 +2355,7 @@
 
 - name: linear(Tensor input, Tensor weight, Tensor? bias=None) -> Tensor
   input, weight, bias: "grad.defined() ? linear_backward(input, grad, weight, grad_input_mask) : std::tuple<Tensor, Tensor, Tensor>()"
+  result: auto_linear
```
tianyu-l, line 2358: curious what does this line do?

aten/src/ATen/native/Linear.cpp
ezyang: Yes. So another PR we have to do is make einsum not decompose, but THAT is likely to be a lot more controversial. Another reason why making views work is "better" (if you can shake it)

tools/autograd/derivatives.yaml
ezyang: It's for forward mode AD; it says that this function is linear and thus is forward ad formula is trivial (do the same function)

aten/src/ATen/functorch/BatchRulesLinearAlgebra.cpp
```diff
@@ -109,6 +109,19 @@ static std::tuple<Tensor, std::optional<int64_t>> bmm_batch_rule(
...
+static std::tuple<Tensor, std::optional<int64_t>> linear_batch_rule(
+    const Tensor& self, std::optional<int64_t> self_bdim,
+    const Tensor& weight, std::optional<int64_t> weight_bdim,
+    const std::optional<Tensor>& bias, std::optional<int64_t> bias_bdim
+    ) {
+  // TODO: add error checking
+  auto self_ = moveBatchDimToFront(self, self_bdim);
+  auto weight_ = moveBatchDimToFront(weight, weight_bdim);
+  auto bias_ = bias.has_value() ? std::make_optional<Tensor>(moveBatchDimToFront(*bias, bias_bdim)) : std::nullopt;
+  return std::make_tuple( at::linear(self_, weight_, bias_), 0 );
```
albanD, line 122: ho linear supports arbitrary batch dimensions on the weights?
Your backward formula seems to say no? :D

aten/src/ATen/native/Linear.cpp
albanD, line 913: What is the perf hit of this for a regular nn.Linear() layer?

albanD (commented):
Sounds ok as long as there is no perf hit, this is a pretty hot path.

aten/src/ATen/functorch/BatchRulesLinearAlgebra.cpp
ezyang: whoops :)

aten/src/ATen/native/Linear.cpp
ezyang: This is potentially actually pretty bad lol. And it doesn't even do what I want, because I want the einsum to also show up as its own operator LOL.

PR #161106, opened 2025-08-20
[DeviceMesh] Simplifying internal bookkeeping with CuTe layout

fduwjj:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #161106
* #161016
* #162690

We want to refactor the internal bookkeeping of DeviceMesh so that:
Simply the bookkeeping logics and make it generic enough so that it is easy to support new transformations like flatten noncontiguous dim, reshape and unflatten. (We leveraged the CuTe layout). This new layout also let us handle non-contiguous slicing, flatten, transpose possible.

Concretely, in this PR, we do the following:
1. Use the _Layout to handle all index operations rather use a map to record mesh dims.
2. Replaced`flatten_name_to_root_dims` with `flatten_name_to_root_layout`. Basically one (size, stride) pair maps to one PG. One mesh_dim_name can only map to only layout. (More than one mesh_dim_name can map to the same layout).
3. Replaced `_get_slice_mesh_dims` with `_get_slice_mesh_layout`.
4. Use a new function `_maybe_create_backend` to create PG based on mesh_dim_name and layout.
5. Use a new function `_from_layouts` to create a new DeviceMesh from a given layout list.

The PR looks big indeed but we don't change any existing behavior of DeviceMesh, so it is a pure refactor.

cc @H-Huang @awgu @wanchaol @fegin @wz337 @wconstab @d4l3k @pragupta

PR #161016, opened 2025-08-19
[DeviceMesh] Introduce CuTe layout into devicemesh code base for internal bookkeeping

fduwjj:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #161016
* #162690
* #162414
* #162534
* #162413

DeviceMesh essentially is a way to specify how devices interact with each other or device layout. They are all integers but because they can have various shapes and meshes, it make internal bookkeeping internally way more challenging. Currently our internal bookkeeing inside DeviceMesh is not scalable, so in order to support new functions like `_unflatten`, we need to introduce very complicated logics inside DeviceMesh as pointed out per comment (https://github.com/pytorch/pytorch/pull/159482/files#r2256025452). So thanks to @lw 's suggestion and PoC PR (https://github.com/pytorch/pytorch/pull/160429), we realize that by leveraging CuTe layout algebra([ref](https://docs.nvidia.com/cutlass/media/docs/cpp/cute/02_layout_algebra.html)) from Cutlass will greatly simply our internal mechanical bookkeeping for and make the abstraction ops way easier on top of it. So to make things go incrementally, we propose couple steps here https://github.com/pytorch/pytorch/issues/160337#issuecomment-3195106243.

On top of what we have been doing about PyCute we want to continue add methods into the wrapper class so that we can get rank indexes needed for ProcessGroup Creation with a layout object. We also added detailed explanations and comments (thanks to llm) and unit test to show case the code indeed is working as expected. 

More PRs are on the way.

cc @H-Huang @awgu @wanchaol @fegin @wz337 @wconstab @d4l3k @pragupta

PR #162413, opened 2025-09-08
[CuTe] Copy code from pycute for device mesh bookkeeping

fduwjj:
* #161106
* #161016
* #162690
* #162414
* #162534
* __->__ #162413

We copied the whole module and its unit test into pytorch codebase. (https://github.com/NVIDIA/cutlass/blob/main/python%2Fpycute%2Flayout.py).

We did change the indentation of code from 2 spaces to 4 spaces. And add lint suppressor to make mypy happy.

Also we need to make changes to unit test to include ownership and use `run_tests, TestCase` so that the test gets picked up by CI.


fduwjj:
@ezyang it is direct copy without any chance, the only change here is to make linter happy: add mypy suppression and formatting. I didn't copy all of them because we don't need. Also previous codes are in two different files, I just merge them into one.

ezyang:
Sorry where did you copy from? Here is what it looks like in CUTLASS: https://github.com/NVIDIA/cutlass/blob/main/python%2Fpycute%2Flayout.py

At minimum I was expecting to see 2 space indentation or you telling us you auto formatted it first. There's also three modules in the original.

ezyang:
We do need to reformat; I see now you have specified what changes you made, I am happy to trust you on those.

ezyang:
OK, so I see what you did is that you eliminated all code that wasn't being used by DeviceMesh. I guess this wasn't that difficult, so we can go this route.

ezyang:
(For future reference, something that I bet would have been quicker is to just import ALL the code without DCE'ing bits that aren't used; I would have also been OK with that.)

ezyang (approved):
You should also import in the tests (they're pretty short) and get them hooked up to CI. They will help us evaluate if we're doing refactors correctly.

torch/distributed/_mesh_layout.py
```diff
@@ -0,0 +1,269 @@
...
+Definition of CuTe Layouts and functions to manipulate them
+"""
+
+from functools import reduce
+from itertools import chain
+
+
+class Integer:
+    @classmethod
+    def __subclasshook__(cls, c):  # type: ignore[no-untyped-def]
```
ezyang: BTW, you can just put an ignore suppression at the top of the file, which would have worked too (and then you can have another intermediate PR that adds typing everywhere, for example. You could ask claude code to do it.)

torch/distributed/pycute/__init__.py
```diff
@@ -0,0 +1,38 @@
+# flake8: noqa
```
ezyang: _pycute to be private?

ezyang: it also might be better to not call it pycute as we're going to change the semantics to be lexicographic. I actually locally named my attempt at this "lcute".

fduwjj: sure.

ezyang:
Looks good. BTW we probably don't need swizzle (and it has no deps from the other modules).

test/distributed/_pycute/test_coalesce.py
```diff
@@ -0,0 +1,99 @@
+# ruff: noqa: PGH004, G004, F403
```
ezyang, line 1: Lol you should double check underscore in the test directory doesn't cause pytest to stop picking it up haha

fduwjj: wait do you mean _pytest won't get picked up? But I do run into CI failure before I use "from torch.testing._internal.common_utils import run_tests, TestCase"

Skylion007 (approved):

test/distributed/_pycute/test_coalesce.py
ezyang: yeah i just want you to be sure haha

PR #162414, opened 2025-09-08
[DeviceMesh] Make CuTe layout as mesh layout to be ready for using in DeviceMesh

fduwjj:
* #161106
* #161016
* #162690
* __->__ #162414
* #162534
* #162413

We create a wrapper class named "_MeshLayout" acting as a layout for device mesh so that we can add new methods more specific to DeviceMesh and keep the core logic of CuTe manipulation inside pycute module. This PR create the main body of the code and then next PR will come with actual implementation and unit test for device mesh layout. (Actual implementation can be found in https://github.com/pytorch/pytorch/pull/161016)


ezyang:
ngl, the size of this diff is NOT looking good. What's going on?

fduwjj:
@ezyang This one is still WIP and I will need to update its descriptions. Basically I want to do two things in this PR: To make API into layout class instead of using original pycute implementation. So it is purely cosmetic change and remove some part of functions which are not used by us, but still no change to the business logic.

fduwjj:
@ezyang yes they already have that Layout class but all methods are not members of that class unfortunately we need to move it into the class.

fduwjj:
As discussed offline, split the PR so that it's easier to review as well.

torch/distributed/_mesh_layout.py
```diff
@@ -0,0 +1,75 @@
...
+    def sizes(self) -> NestedIntTuple:
+        return self.shape
+
+    @property
+    def strides(self) -> NestedIntTuple:
+        return self.stride
+
+    @property
+    def sizes_and_strides(self) -> Iterator[tuple[int, int]]:
+        return zip(flatten(self.shape), flatten(self.stride))  # type: ignore[arg-type]
```
ezyang: OK, it's a lot clearer in this formulation of the PR. This method makes me nervous; it's not obvious to me that downstream people will use this correctly.

torch/distributed/_mesh_layout.py
```diff
@@ -0,0 +1,75 @@
...
+    # operator []    (get-i like tuples)
+    def __getitem__(self, i: int) -> "_Layout":
+        size = self.sizes[i]
+        stride = self.strides[i]
+        if is_tuple(size) and is_tuple(stride):
+            return _Layout(size, stride)  # type: ignore[arg-type]
+        elif isinstance(size, int) and isinstance(stride, int):
+            return _Layout((size,), (stride,))
+        else:
+            raise ValueError("size and stride must be either int or tuple")
```
ezyang: Why doesn't this directly use Layout.__getitem__

torch/distributed/_mesh_layout.py
```diff
@@ -0,0 +1,75 @@
...
+    is_tuple,
+    Layout,
+)
+
+
+NestedIntTuple: TypeAlias = tuple["int | NestedIntTuple", ...]
+
+
+@dataclass(frozen=True, init=True)
+class _Layout(Layout):
```
ezyang: So, I'm not entirely sure why you chose to do it this way, but I am ALSO ok with you directly patching the pycute Layout class to have these things. When you go to lexicographic, you're going to have to patch it anyway, so might as well avoid having another _Layout class running around...

torch/distributed/_mesh_layout.py
fduwjj: Because it will return a parent class and if we add some new functions in the child class, the code will crash.

torch/distributed/_mesh_layout.py
fduwjj: so ok that is something I want to discuss. The original pycute assume the size and stride can be just a int which makes their code very complicated so I really don't like it... Plus we will add more device-mesh bookkeeping related logics into this class as well, so that's why I want to have this wrapper class.

torch/distributed/_mesh_layout.py
fduwjj: ok to clarify, this is just a util for downstream use in UT and device mesh. Since we define it in the wrapper class so it will be less risky because we keep most of essential logic still in the parent class?

torch/distributed/_mesh_layout.py
```diff
@@ -0,0 +1,75 @@
...
+    shape: NestedIntTuple
+    stride: NestedIntTuple
+
+    def __post_init__(self) -> None:
+        if not is_tuple(self.shape):
+            raise ValueError(f"shape must be a tuple, got {type(self.shape)}")
```
Skylion007: These should raise TypeErrors, not ValueError, right?

torch/distributed/_mesh_layout.py
Skylion007: Should raise AssertionError

torch/distributed/_mesh_layout.py
Skylion007: If you add an assert that both types are not ints, MyPy won't complain here either.

torch/distributed/_mesh_layout.py
```diff
@@ -0,0 +1,75 @@
...
+    @property
+    def strides(self) -> NestedIntTuple:
+        return self.stride
+
+    @property
+    def sizes_and_strides(self) -> Iterator[tuple[int, int]]:
+        return zip(flatten(self.shape), flatten(self.stride))  # type: ignore[arg-type]
+
+    def numel(self) -> int:
+        return math.prod(flatten(self.shape))
```
Skylion007, line 54: Yeah surprised it's not complaining here

torch/distributed/_mesh_layout.py
ezyang: So specifically, if we do eventually want to support merging non-contiguous dimensions, it is important that device mesh not overuse this (because you won't get a list here that is the same as the mesh dimensionality). I agree it is possible to use this correctly, but it's also possible to use it incorrectly but in a way that will only show up when you flatten non-contiguous dims

ezyang (approved):
Nothing here seems objectionable

torch/distributed/_mesh_layout.py
fduwjj: After merging we will get a new layout for the device mesh so we don't have a problem with this property.

torch/distributed/_mesh_layout.py
fduwjj: Hmm I think since we explicitly check it, I think ValueError might be proper here?

torch/distributed/_mesh_layout.py
fduwjj: math.prod is available in python 3.9, so it should be fine.

fegin (approved):

PR #162418, opened 2025-09-08
[precompile] Fix issues with guard serialization on distributed types.

zhxchen17:
Summary: Add more support for torch internal distributed data structures.

Test Plan:
test_guard_serialization.py

Rollback Plan:

Differential Revision: D81927732

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela

### DIFF
--- a/test/dynamo/test_guard_serialization.py
+++ b/test/dynamo/test_guard_serialization.py
@@ -26,6 +26,7 @@ from torch._dynamo.symbolic_convert import (
 from torch._dynamo.utils import dynamo_timed, get_metrics_context
 from torch._guards import compile_context, CompileContext, tracing
 from torch.overrides import TorchFunctionMode
+from torch.testing._internal.common_utils import IS_MACOS
 from torch.testing._internal.inductor_utils import HAS_GPU
 from torch.utils import _pytree as pytree
 
@@ -52,6 +53,14 @@ class GlobalTorchFunctionMode(TorchFunctionMode):
         if kwargs is None:
             kwargs = {}
         return func(*args, **kwargs)
+
+
+class MyClass:
+    def __getstate__(self):
+        raise RuntimeError("Cannot pickle")
+
+    def add(self, x):
+        return x + 1
 
 
 class SubclassWithMeta(torch.Tensor):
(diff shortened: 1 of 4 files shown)


test/dynamo/test_guard_serialization.py
```diff
@@ -1346,6 +1355,49 @@ def forward(self, x):
...
+        FSDPTestMultiThread.setUp(self)
+
+    def test_guard_serialization_fsdp_module(self):
+        from torch.distributed._tensor import distribute_tensor, Replicate
+        from torch.distributed.device_mesh import init_device_mesh
+        from torch.distributed.fsdp import fully_shard
+
+        mesh = init_device_mesh(str(torch.get_default_device()), (1,))
+        m = SimpleModule()
+        m = fully_shard(m, mesh=mesh)
```
ezyang: I'm a bit confused about this. Are we actually compiling fullgraph through the fully sharded SimpleModule?

zhxchen17: @ezyang oh this is for caching precompile, not the aot/fullgraph case. James and I are working on an orthogonal use case (omnifm) and it uses fsdp on some module.

torch/_dynamo/guards.py
```diff
@@ -3222,6 +3252,9 @@ def reducer_override(
         elif isinstance(obj, types.MappingProxyType):
             return type(self)._unpickle_mapping_proxy, (obj.copy(),)
 
+        elif isinstance(obj, type({}.keys())):
```
dolpm: nit: use https://github.com/pytorch/pytorch/blob/main/torch/_dynamo/utils.py#L2567 ?

torch/_dynamo/guards.py
```diff
@@ -3234,14 +3267,27 @@ def reducer_override(
...
             return _Missing, ()
 
-        elif isinstance(obj, types.CodeType):
+        elif isinstance(
+            obj,
+            (
+                types.CodeType,
+                torch._C.Stream,
+                torch._C._distributed_c10d.ProcessGroup,
+                weakref.ReferenceType,
```
dolpm: where do we get this list?

dolpm (approved):

torch/_dynamo/guards.py
zhxchen17: we just ran precompile through an internal model with some complex data structure and saw these not needed but serialized.

PR #162423, opened 2025-09-08
Update docs for quantile to be clearer for nearest

janeyx99:
Correct the rounding scheme for nearest in quantile.

### DIFF
--- a/torch/_torch_docs.py
+++ b/torch/_torch_docs.py
@@ -7132,8 +7132,7 @@ indices ``i`` and ``j`` in the sorted order, result is computed according to the
 - ``linear``: ``a + (b - a) * fraction``, where ``fraction`` is the fractional part of the computed quantile index.
 - ``lower``: ``a``.
 - ``higher``: ``b``.
-- ``nearest``: ``a`` or ``b``, whichever's index is closer to the computed quantile index
-  (follows :func:`torch.round`).
+- ``nearest``: ``a`` or ``b``, whichever's index is closer to the computed quantile index (follows :func:`torch.round`).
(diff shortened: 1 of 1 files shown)


soulitzer (approved):

torch/_torch_docs.py
```diff
@@ -7132,7 +7132,8 @@ def merge_dicts(*dicts):
 - ``linear``: ``a + (b - a) * fraction``, where ``fraction`` is the fractional part of the computed quantile index.
 - ``lower``: ``a``.
 - ``higher``: ``b``.
-- ``nearest``: ``a`` or ``b``, whichever's index is closer to the computed quantile index (rounding down for .5 fractions).
+- ``nearest``: ``a`` or ``b``, whichever's index is closer to the computed quantile index
```
soulitzer: So the current docs are wrong? If we follow the "round to even" rules like torch.round, it would mean that 1.5 is rounded up and 2.5 is rounded down, and not always rounded down like the current docs suggest.

soulitzer: Oh you mention that in your description, but title suggests it just makes things clearer

trirpi: Yes, there's a discrepancy.

```
x = torch.tensor([0.0, 1.0, 2.0, 3.0, 4.0, 5.0, 6.0])
quantiles = torch.tensor([0.25])
result = torch.quantile(x, quantiles, interpolation="nearest")
print(result)
```

Rounds 1.5 up to 2 instead of 1

janeyx99: Yes! This was prompted from a pytorch q&a post

PR #162424, opened 2025-09-08
[torch][c10d] fix split_group in mixed backend case

suo:
Today we can initialize a mixed-backend process group (e.g. "cpu:gloo,cuda:nccl") but we can only pass one set of process group options.

However, when we call `split_group`, we retrieve that set of options from the parent PG and pass it to the ProcessGroup::groupSplit C++ API, which then attempts to propagate that set of options to all backends.

This leads to an assert on some user code, where ProcessGroupGloo::split is expecting gloo options but receives nccl options instead.

Arguably the APIs as currently designed are just broken; we should not ever expect a single set of backend options to apply across multiple backends. However, fixing this would require changing quite a few public APIs.

As a quick fix, since user-provided options really only exist for NCCL, just warn and fall-back to defaulted options for Gloo if non-gloo options are detected.

### DIFF
--- a/test/distributed/test_c10d_nccl.py
+++ b/test/distributed/test_c10d_nccl.py
@@ -1086,6 +1086,62 @@ class ProcessGroupNCCLGroupTest(MultiProcessTestCase):
         self.assertEqual(backend.comm_split_count(), 2)
 
         dist.destroy_process_group()
+        
+    @requires_nccl_version((2, 18), "Need NCCL 2.18+ for ncclCommSplit")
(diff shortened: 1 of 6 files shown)


d4l3k (approved):
LGTM

fduwjj (approved):

H-Huang (approved):
Thanks for fixing!

suo:
@d4l3k take a look at the latest commit, does this seem fine to you?

d4l3k (approved):
LGTM

PR #156674, opened 2025-06-24
simplify nvrtc discovery login in compile_kernel

msaroufim:
Followup from https://github.com/pytorch/pytorch/pull/156332

Tested a bunch while I was working on https://github.com/pytorch/pytorch/pull/156380

Works just fine on dev gpus

PR #162427, opened 2025-09-08
Revert "simplify nvrtc discovery login in compile_kernel (#156674)"

atalman:
This reverts commit 4e8dd11. Also removing cuda 11 logic.
Since we have cuda 12 and cuda 13 builds now. Lib name should be: ``lib_name = f"nvrtc64_{version}_0.dll"`` or ``f"lib64/libnvrtc.so.{version}"``

### DIFF
--- a/torch/cuda/_utils.py
+++ b/torch/cuda/_utils.py
@@ -10,25 +10,11 @@ from torch._utils import _get_device_index as _torch_get_device_index
 
 
 def _get_nvrtc_version(cuda_version: int) -> str:
-    # TODO: Expose this from native code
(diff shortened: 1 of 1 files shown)


malfet (changes requested):
What's your test plan?

msaroufim:
please don't bring back that code, it was a vibe coded mess. If there are any issues we should forward fix

msaroufim (changes requested):
please don't bring back that code, it was a vibe coded mess. If there are any issues we should forward fix. It caused other problems for instance, if you revert that PR it won't be possible to pass pytorch tests if someone installs with uv from source

PR #162434, opened 2025-09-08
[poc] Add config to allow gradient dtype mismatch

soulitzer:
* __->__ #162434
* #160883
* #160888


albanD (changes requested):
I'm very confused what problem it solves, why it solves it and why it is generally a valid change to make?

PR #162438, opened 2025-09-09
[pre_compile] Add check for cuda and hardware version

yushangdi:
if we detect compiled model is using cuda in meaningful way, we should store information about cuda + hardware
 
 Example: `SystemInfo(python_version='3.12.9', torch_version='2.9.0a0+gite02b0e6', cuda_version='12.6', triton_version=(3, 4), gpu_name='NVIDIA PG509-210')`

### DIFF
--- a/torch/_dynamo/aot_compile.py
+++ b/torch/_dynamo/aot_compile.py
@@ -1,5 +1,6 @@
 import abc
 import builtins
+import dataclasses
 import importlib
 import inspect
 import logging
@@ -10,7 +11,8 @@ from typing import Any, Callable, Optional
 
 import torch
 import torch.fx
-from torch._dynamo.precompile_context import PrecompileContext
+from torch._dynamo.graph_utils import _graph_uses_non_cpu
+from torch._dynamo.precompile_context import PrecompileContext, SystemInfo
 
 from . import convert_frame
 from .hooks import Hooks
@@ -50,6 +52,12 @@ class CompileArtifacts:
     compiled_fn: SerializableCallable
     original_code: types.CodeType
     closure: Optional[tuple[Any, ...]]
+    use_cuda: bool
+    system_info: SystemInfo = dataclasses.field(default_factory=SystemInfo.current)
+
+    def check_compatibility(self) -> None:
+        current_system = SystemInfo.current()
+        current_system.check_compatibility(self.system_info, self.use_cuda)
 
 
 @dataclass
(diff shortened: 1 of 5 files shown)


torch/_dynamo/package.py
```diff
@@ -275,13 +279,36 @@ def _find_code_source(obj: Any) -> Optional[str]:
 class _DynamoCacheEntry:
     codes: list[_DynamoCodeCacheEntry]
     inlined_sources: set[InlinedSource]
+    use_cuda: bool
     python_version: str = platform.python_version()
     torch_version: str = torch.__version__
+    cuda_version: Optional[str] = torch.version.cuda
+    triton_version: Optional[str] = triton.__version__ if has_triton() else ""
```
zhxchen17: @yushangdi I think it's better if we factor these versions out into a standalone type.

Then we can reuse the logic in aot precompile as well. (code pointer: https://fburl.com/797sx4uy)

torch/_dynamo/aot_compile.py
```diff
@@ -120,6 +124,9 @@ def deserialize(cls, data: bytes) -> "AOTCompiledFunction":
         state["compiled_fn"] = deserializer(compiled_fn_state)
         state["original_code"] = SerializedCode.to_code_object(state["original_code"])
 
+        current_system = SystemInfo.current()
+        current_system.check_compatibility(state["system_info"], state["use_cuda"])
+
```
zhxchen17: Would be better to move this to `__init__` of AOTCompiledFunction, such that every time it got constructed, the invariant that it's runnable will be preserved.

zhxchen17 (approved):

torch/_dynamo/aot_compile.py
yushangdi: @zhxchen17 I moved it to the __post_init__ of `CompileArtifacts` now. Since `AOTCompiledFunction` just takes in a `CompileArtifacts`, it should be equivalent (or even more comprehensive) of having it in the init of AOTCompiledFunction.

torch/_dynamo/aot_compile.py
```diff
@@ -50,6 +52,12 @@ class CompileArtifacts:
     compiled_fn: SerializableCallable
     original_code: types.CodeType
     closure: Optional[tuple[Any, ...]]
+    use_cuda: bool
+    system_info: SystemInfo = dataclasses.field(default_factory=SystemInfo.current)
+
+    def __post_init__(self) -> None:
+        current_system = SystemInfo.current()
+        current_system.check_compatibility(self.system_info, self.use_cuda)
```
zhxchen17: I mean we don't have to put this into CompileArtifacts, but the wrapper type AOTCompiledFunction.

putting it into post_init might be not ideal if we just want to create some dummy CompileArtifacts for testing/debugging

zhxchen17 (approved):
thanks

PR #162439, opened 2025-09-09
DeviceMesh: support _rank for use with non-global PGs

d4l3k:
Summary: This adds a `_rank` field to DeviceMesh init that allows for instantiating a DeviceMesh without depending on `dist.get_rank()` which requires a global PG to be instantiated.

Test Plan:
```
buck2 test mode/opt -c fbcode.enable_gpu_sections=true  //caffe2/test/distributed:device_mesh -- init_backend
```

Rollback Plan:

Differential Revision: D81981777

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @pragupta @ezyang @msaroufim @dcci

### DIFF
--- a/torch/distributed/device_mesh.py
+++ b/torch/distributed/device_mesh.py
@@ -396,6 +396,9 @@ if True:  # just to temporarily avoid reindentation
             device_type (str): The device type of the mesh. Currently supports: "cpu", "cuda/cuda-like".
             mesh (ndarray): A multi-dimensional array or an integer tensor describing the layout
                 of devices, where the IDs are global IDs of the default process group.
+            _rank (int): (experimental/internal)
+                The global rank of the current process. If not provided, it will
+                be inferred from the default process group.
 
         Returns:
             DeviceMesh: A :class:`DeviceMesh` object representing the device layout.
@@ -430,6 +433,7 @@ if True:  # just to temporarily avoid reindentation
                 tuple[tuple[Optional[str], Optional[C10dBackend.Options]], ...]
             ] = None,
             _init_backend: bool = True,
+            _rank: Optional[int] = None,
         ) -> None:
             self.device_type = device_type
             if isinstance(mesh, torch.Tensor) and mesh.device.type != "cpu":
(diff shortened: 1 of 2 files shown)


kwen2501 (approved):
Much needed, thanks!

ezyang:
@fduwjj you should take a look

torch/distributed/device_mesh.py
```diff
@@ -461,7 +468,7 @@ def __init__(
                     self._thread_id = threading.get_ident()
 
                 # calculate the coordinates of the current global rank on the mesh
-                rank_coords = (self.mesh == get_rank()).nonzero()
+                rank_coords = (self.mesh == _rank).nonzero()
```
ezyang, line 471: OK, this is relatively simple then

fduwjj:
I feel for short term this change is fine but for long term, we don't want to do this on the device mesh level? 

Do we think it makes more sense to let the `get_rank` API to get ranks inside the API without default PG instead? I mean IIUC we get local ranks from ENV variable for example, then why can't we do this in the get_rank API? Also to answer @kwen2501 's question, I really don't understand why device mesh wants to know the current rank.

fduwjj (approved):

torch/distributed/device_mesh.py
```diff
@@ -396,6 +396,9 @@ class DeviceMesh:
             device_type (str): The device type of the mesh. Currently supports: "cpu", "cuda/cuda-like".
             mesh (ndarray): A multi-dimensional array or an integer tensor describing the layout
                 of devices, where the IDs are global IDs of the default process group.
+            _rank (int): (experimental/internal)
+                The global rank of the current process. If not provided, it will
+                be inferred from the default process group.
```
kwen2501, line 401: Ohh, I totally misunderstood what this argument is for. 
I thought it was intended to describe "the rank of this process in the DeviceMesh under creation."
Like, if I'd like to create a DeviceMesh covering global ranks [100, 200), then rank 100 would set `_rank=0`, rank 101 would set `_rank=1`, etc.

PR #162296, opened 2025-09-05
Allow add_persistent_r_block to scale up rblock up to a limit

PaulZhang12:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162447
* #162446
* __->__ #162296

Screenshot 2025-09-18 at 4 22 53 PM

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben

PR #162446, opened 2025-09-09
Add num_store to inductor_meta and use it to scale persistent reduction x block

PaulZhang12:
* __->__ #162446
* #162296

Scale up XBLOCK for contiguous persistent reductions based on rnumel and number of loads + stores

Screenshot 2025-09-18 at 5 02 57 PM

Differential Revision: [](https://our.internmc.facebook.com/intern/diff/)

Differential Revision: [](https://our.internmc.facebook.com/intern/diff/)


torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -2904,8 +2907,26 @@ def _persistent_reduction_configs(
...
         pass
     # TODO(jansel): we should be able to improve these heuristics
-    elif reduction_hint == ReductionHint.INNER and rnumel >= 256:
-        configs = configs[:1]
+    elif reduction_hint == ReductionHint.INNER:
+        if rnumel > 1024:
+            configs = configs[:1]
+        else:
+            x_block = 8 if rnumel < 1024 else 4
+            if xnumel // x_block < 128 or (loads_and_stores >= 5 and rnumel >= 256):
```
eellison: > xnumel // x_block < 128

Should we set xblock so loads are vectorized correctly to a certain numel instead of setting xblock directly to 1 ?

eellison (commented):
Do you have any particular kernels or dashboard runs motivating this ?

PaulZhang12:
@eellison I have added the blackwell chart in the summary for quack comparison. This optimization is largely responsible for the smaller r shapes:
(32768, 256)
(32768, 512)
(32768, 1024)
(32768, 2048)

This also improves performance on the (32768, 256) shape for Hopper, a little over 10%

torch/_inductor/runtime/triton_heuristics.py
PaulZhang12: Does it matter what xblock is here? https://github.com/pytorch/pytorch/pull/162447 ensures that loads are vectorized properly through num_warps. Whether xblock increases or decreases shouldn't affect whether the loads are vectorized or not

torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -2904,8 +2907,26 @@ def _persistent_reduction_configs(
     if "y" in size_hints:
         pass
     # TODO(jansel): we should be able to improve these heuristics
-    elif reduction_hint == ReductionHint.INNER and rnumel >= 256:
-        configs = configs[:1]
+    elif reduction_hint == ReductionHint.INNER:
+        if rnumel > 1024:
+            configs = configs[:1]
+        else:
+            x_block = 8 if rnumel < 1024 else 4
```
v0i0: there seems to be a very narrow case here for exactly rnumel == 1024 (given the branch above and this)?

v0i0 (approved):

torch/_inductor/runtime/triton_heuristics.py
eellison: Right, but larger xblock just means more memory in flight for any sm no ?

eellison (approved):

torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -2904,8 +2907,26 @@ def _persistent_reduction_configs(
...
+            if xnumel // x_block < 128 or (loads_and_stores >= 5 and rnumel >= 256):
+                # If loads/stores greater than 5, a lot of register pressure
+                # rnumel < 256 means no vectorized loads if we split up r dim
```
shunting314: How do you come up with the threshold 256 here regarding vectorized load/store? Does it assume specific dtype and num_warps?

torch/_inductor/codegen/common.py
```diff
@@ -2701,6 +2702,7 @@ def store(
             self._update_store_cache(name, value)
         if name not in V.graph.removed_buffers:
             self.kernel.store(name, index, value, mode=mode)
+            self.kernel.num_store += 1
```
shunting314, line 2706: fwiw, I think the number may be bigger than reality since when fusion happens, some store may be removed later when we found it's not needed outside of the kernel.

shunting314 (approved):

torch/_inductor/runtime/triton_heuristics.py
PaulZhang12: I'm not sure what you mean by loads vectorized correctly to a certain numel. In the case of just setting it to 1, either xnumel is too small to which we need to balance the sms used, or we might have too high of register usage and cannot increase xblock

torch/_inductor/runtime/triton_heuristics.py
PaulZhang12: @v0i0 yeah I noticed this through some empirical observations, I guess rnumel == 1024 can be common given we round to power of 2? Happy to change this

torch/_inductor/runtime/triton_heuristics.py
PaulZhang12: Yes we assume dtype fp16/bf16. We also know that if rnumel < 256, we will not have full vectorization on the r dimension, so we want xblock > 1 to have full vectorization + memory coalescing for 1 whole warp at least

torch/_inductor/codegen/common.py
PaulZhang12: @shunting314 yes I saw that sometimes num_store = 0, is there a better way to store the actual number of stores (other than parsing through the source code I guess)? Or is this reasonable

shunting314: So the reason for the difference is because a store is written as a DeferredLine which can be canceled later (when we have more knowledge of the kernel). I think you can decrease V.kernel.num_store accordingly when a store is canceled during codegening the source code  of the kernel (no text parsing lol)

PaulZhang12:
@PaulZhang12 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/).

torch/_inductor/codegen/common.py
PaulZhang12: @shunting314 fixed the logic here, thanks!

PR #162469, opened 2025-09-09
[inductor][reordering] Use runtime estimations in reorder/sink passes

IvanKobzarev:
* __->__ #162469

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben


eellison (commented):
this is no longer needed, right ? close ? or mark as draft

torch/_inductor/comms.py
```diff
@@ -203,6 +206,103 @@ def _temp_group_visit_leaves(snode, fn):
         fn(snode)
 
 
+def wait_exposed_communication_time(
+    snodes_to_wait, runtimes
+) -> tuple[float, float, str]:
```
eellison, line 210: Woudl you mind adding type annotations, and a docstring to all of these functions ? it makes a lot easier to read/review

torch/_inductor/comms.py
```diff
@@ -258,6 +381,44 @@ def _initialize_double_linked_list(
...
+            return False
+
+    return True
+
+
+def contains_async_collective(snode):
+    return contains_collective(snode, is_async_collective)
+
+
 def _reorder_communication_preserving_peak_memory_internal(
```
eellison: i know a lot of this is existing, but I find the single, 515 line function, with various internal functions, very hard to follow... cc @wconstab for other opinions.

torch/_inductor/config.py
```diff
@@ -385,9 +385,9 @@ def prologue_fusion_enabled() -> bool:
         ],
     ]
 ] = [
-    "reorder_compute_for_overlap",
-    "sink_waits",
-    "raise_comms",
+    "reorder_communication_preserving_peak_memory",
+    "sink_waits_iterative",
+    "reorder_communication_preserving_peak_memory",
```
eellison: Do we have enough tests/ data to inform putting this on by default ?

torch/_inductor/comms.py
```diff
@@ -203,6 +206,126 @@ def _temp_group_visit_leaves(snode, fn):
...
+    for snode in snodes[1:]:
+        # We may have some ops without Wait,
+        # e.g. DTensor torch.ops._dtensor.shard_dim_alltoall
+        unmet_deps = OrderedSet(
+            d.name for d in snode.unmet_dependencies if not _is_fake_dep(d)
+        )
+
+        if unmet_deps & collective_outs:
+            overlap_info += f"->W[{snode.get_name()}]"
+            break
```
eellison, line 298: When would this happen ? isn't the first user necessarily the wait ?

torch/_inductor/comms.py
```diff
@@ -203,6 +206,126 @@ def _temp_group_visit_leaves(snode, fn):
...
+                continue
+        if contains_wait(snode):
+            has_wait_for_collectives_found = False
+            for coll in collectives_found:
+                if is_corresponding_collective_wait(collective_snode, snode):
+                    has_wait_for_collectives_found = True
+                    break
+            if has_wait_for_collectives_found:
+                # Any compute after not overlapping original Collective
+                break
```
eellison, line 314: this is dead code, since we break in the loop above

torch/_inductor/comms.py
```diff
@@ -203,6 +206,126 @@ def _temp_group_visit_leaves(snode, fn):
...
+
+        comp_time_before = comp_time
+
+        def accumulate_time(_snode: BaseSchedulerNode) -> None:
```
eellison, line 318: We don't model collective time here, intentional ?

torch/_inductor/config.py
IvanKobzarev: For llama3 1d and 2d and internal Shuai model it showed better result than "reorder" and ["reorder", "sink_waits"].
I would put reorder-sink-reorder as a recommended setting, but not turned on by default for inductor.

What about to add the comment and keep the passes list empty for now, as recommended reordering is fx passes?

torch/_inductor/comms.py
IvanKobzarev: The loop above breaks only its own loop `for coll in collectives_found:`, this break breaks the external loop `for snode in snodes[1:]:` to stop looking into next compute as we found a wait that is syncing the compute stream.

torch/_inductor/comms.py
IvanKobzarev: There is no multigroup modeling, but `runtimes[_snode]` should contain estimated time for all nodes - collectives, compute etc.

torch/_inductor/comms.py
IvanKobzarev: I think as we do not plan to do more changes here - we could keep it as it is to not invest time. 
But if we have to return or change the logic - we can do a refactoring.

eellison (approved):

PR #162470, opened 2025-09-09
[inductor][bucketing] Fx collectives bucketing of multiple dtypes

IvanKobzarev:
* __->__ #162470

Bucketing of multiple dtypes to be processed in one bucketed collective.

First target is to bucket bf16 and f32, but already can be used with other dtypes.

For now multidtype bucketing is only supported with "custom_ops" mode.
Non custom_ops needs additional work on inductor side.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @ezyang


torch/_inductor/fx_passes/bucketing.py
```diff
@@ -90,17 +92,17 @@ def is_wait_tensor_from_all_gather_into_tensor(node: torch.fx.Node) -> bool:
     return is_wait_tensor(node) and is_all_gather_into_tensor(node.args[0])  # type: ignore[arg-type]
 
 
-def collect_node_descendants(
+def collect_node_descendents(
```
eellison: sorry for the misspelling ! nit: for re-spellings like this, it can be nice to separate out real code changes from refactorings so it's easier to review..

eellison (commented):
now that the other prs have landed - mind rebasing ?

torch/_inductor/fx_passes/bucketing.py
IvanKobzarev: Yes, will clean this out this pr.

test/distributed/test_inductor_collectives.py
```diff
@@ -1754,6 +1754,115 @@ def func2(x, w, rs_0, rs_1, tag, ranks, group_size):
...
+        out = compiled(*inputs, **self.get_world_trs())
+        _, y_ag0, y_ag1 = out
+        assert y_ag0.dtype == ag_0.dtype
+        assert y_ag1.dtype == ag_1.dtype
+
+        assert same(out, correct), f"{out} va {correct}"
+
+    @unittest.skipIf(not HAS_GPU, "Inductor+gpu needs triton and recent GPU arch")
+    @unittest.skipIf(not SM80OrLater, "bfloat16")
+    @parametrize("bucket_mode", ["all_custom_ops_multidtype"])
```
eellison, line 1809: nit: can we update the config of `bucket_mode` : Union[literal[....]]]` so we know what the options are ?

torch/_inductor/fx_passes/bucketing.py
```diff
@@ -19,21 +19,45 @@
...
+    return (group_name, reduce_op)
+
+
+def pick_bucket_dtype(dtypes: list[torch.dtype]) -> torch.dtype:  # type: ignore[name-defined]
+    assert len(dtypes) > 0
+    s = OrderedSet(dtypes)
+    if len(s) == 1:
+        return dtypes[0]
+    if s == OrderedSet([torch.bfloat16, torch.float]):  # type: ignore[attr-defined]
+        return torch.bfloat16  # type: ignore[attr-defined]
```
eellison: reason for this special case ? we could just always choose the lowest itemsize dtype.

torch/_inductor/fx_passes/bucketing.py
```diff
@@ -226,6 +252,7 @@ def bucket_all_gather_by_mb(
     gm: torch.fx.GraphModule,
     bucket_cap_mb_by_bucket_idx: Callable[[int], float],
     filter_wait_node: Optional[Callable[[torch.fx.Node], bool]] = None,
+    mode: Optional[str] = None,
```
eellison: same question - add literal options ?

torch/_inductor/fx_passes/bucketing.py
```diff
@@ -258,6 +289,7 @@ def bucket_reduce_scatter_by_mb(
...
```
eellison: - here too ?

torch/_inductor/fx_passes/bucketing.py
```diff
@@ -275,11 +307,15 @@ def bucket_reduce_scatter_by_mb(
         list[list[torch.fx.Node]]: List of buckets, where each bucket is a list of reduce_scatter nodes.
     """
 
+    group_key_fn = (
+        _rs_group_key_multidtype if mode and "multidtype" in mode else _rs_group_key
```
eellison: I don't think we can do multidtype for reduce scatter, since nccl is actually doing a reduction.

torch/_inductor/fx_passes/bucketing.py
```diff
@@ -288,35 +324,52 @@ def bucket_reduce_scatter_by_mb(
 def _pre_bucket_reduce_scatter(
     rs_ins: list[torch.Tensor],
     group_size: int,
+    dtype: torch.dtype,  # type: ignore[name-defined]
+    numel_mults: list[int],
```
eellison: nit: not sure we need this if we always have the assumption we will view as lowest bitwidth dtype

eellison (commented):
Looks good ! 

Only blocking question is about reduce scatter multi dtype

torch/_inductor/fx_passes/bucketing.py
IvanKobzarev: Yes, we can pick lowert dtype.

torch/_inductor/fx_passes/bucketing.py
IvanKobzarev: Custom ops do not support List[dtype] to calculate it in the op. So I changed to passing multipliers as List[int] instead. We need to know how to split the result according to different dtypes.

test/distributed/test_inductor_collectives.py
IvanKobzarev: We can. It will be a bit long sequence, as we have 2**num_options (custom_ops +/ multidtype +/- all/fsdp). We should reduce it to 4 for now.

torch/_inductor/fx_passes/bucketing.py
IvanKobzarev: Yes, agree, we can do it only with casts and numerics changes. I will remove this option.
We can only do joint uppermost dtype, but that is not what we want :)

torch/_inductor/fx_passes/bucketing.py
```diff
@@ -17,23 +17,45 @@
...
     _, reduce_op, group_size, group_name = node.args
     dtype = node.meta["val"].dtype
     assert isinstance(group_name, str)
     assert isinstance(reduce_op, str)
     return (group_name, reduce_op, dtype)
 
 
+def pick_bucket_dtype(dtypes: list[torch.dtype]) -> torch.dtype:  # type: ignore[name-defined]
+    assert len(dtypes) > 0
+    lowest_dtype = dtypes[0]
```
eellison: nit: `return min(dtypes, key=operator.attrgetter("itemsize"))`

torch/_inductor/fx_passes/bucketing.py
```diff
@@ -391,15 +433,36 @@ def _pre_bucket_all_gather_fake(
 _pre_bucket_all_gather.register_fake(_pre_bucket_all_gather_fake)
 
 
+def _dtype_size_bytes(dtype: torch.dtype) -> int:  # type:  ignore[name-defined]
```
eellison: this is `dtype.itemsize`

torch/_inductor/fx_passes/bucketing.py
```diff
@@ -391,15 +433,36 @@ def _pre_bucket_all_gather_fake(
...
-    ag_ins: list[torch.Tensor],
+    _ag_ins: list[torch.Tensor],
     group_size: int,
     group_name: str,
     dtype: torch.dtype,  # type: ignore[name-defined]
+    out_dtypes: list[torch.dtype],  # type: ignore[name-defined]
     rank: int,
 ) -> list[torch.Tensor]:
+    ag_ins = [
+        torch._prims.convert_element_type(_ag_in, out_dtype)
```
eellison, line 513: We should only view dtype here... Could we make this only do collective merging of different dtypes if we can do it without increase the total bytes transmitted ? Potentially in the future we would want to upcast if the latency is longer than the cost of additional bytes.. leave for future change?

eellison (approved):
looks good, just one comment

torch/_inductor/fx_passes/bucketing.py
IvanKobzarev: Here we will have convert to out_dtypes only for fused-convert-dtype for all-gathers.

eellison: I'm not sure I follow. Why do we do this ? We shouldn't need extra casts

PR #162495, opened 2025-09-09
[ONNX] Expose the testing module

### LINKED ISSUE #162456: [ONNX] Expose `torch.onnx.testing`
Expose the onnx program testing utility

cc @titaiwangms

justinchuby:
* Created a new module `torch/onnx/testing.py` that exposes the `assert_onnx_program` function for testing exported ONNX models.
* Updated the ONNX documentation (`docs/source/onnx.md`) to include `onnx_testing` in the list of relevant modules.

cc @titaiwangms

### DIFF
--- a/docs/source/onnx.md
+++ b/docs/source/onnx.md
@@ -102,6 +102,7 @@ also be interested in reading our [development wiki](https://github.com/pytorch/
     onnx_export
     onnx_ops
     onnx_verification
+    onnx_testing
 ```
 
 ### Deprecated APIs
--- a/docs/source/onnx_testing.md
+++ b/docs/source/onnx_testing.md
@@ -1,0 +1,9 @@
+# torch.onnx.testing
+
+```{eval-rst}
+.. automodule:: torch.onnx.ops
+```
+
+```{eval-rst}
+.. autofunction:: torch.onnx.testing.assert_onnx_program
+```
--- a/docs/source/onnx_testing.md
+++ b/docs/source/onnx_testing.md
@@ -1,7 +1,7 @@
 # torch.onnx.testing
 
 ```{eval-rst}
-.. automodule:: torch.onnx.ops
+.. automodule:: torch.onnx.testing
 ```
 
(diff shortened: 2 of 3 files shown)


justinchuby:
I think it is becoming increasingly useful for users to validate their models when we are making dynamo=True as the default. As an example, we are using it in prototypes: https://github.com/jambayk/torch-onnx-models/blob/main/test/rms_norm_test.py

justinchuby:
In the example above, we can use this to compare the exported model with a reference ep: https://github.com/jambayk/torch-onnx-models/pull/1

> embed it into the exporter

I would prefer separating it, because this way it is more flexible (we can control and manipulate the onnx_program, for example by updating the onnx ir model first before asserting).

titaiwangms:
> In the example above, we can use this to compare the exported model with a reference ep: [jambayk/torch-onnx-models#1](https://github.com/jambayk/torch-onnx-models/pull/1)
> 
> > embed it into the exporter
> 
> I would prefer separating it, because this way it is more flexible (we can control and manipulate the onnx_program, for example by updating the onnx ir model first before asserting).

But this sounds like we don't guarantee the validity of the exporter model from our API.

docs/source/onnx.md
```diff
@@ -102,6 +102,7 @@ also be interested in reading our [development wiki](https://github.com/pytorch/
     onnx_export
     onnx_ops
     onnx_verification
+    onnx_testing
```
titaiwangms, line 105: Is there a place that we can notify the users where they might want to use this to validate their current onnx_program? right now, I am thinking probably it's for advanced users who do customized model surgery (although I feel like they won't do it in onnx program, they probably directly use onnx ir or model proto instead.)?

titaiwangms (approved):

xadupre (approved):

docs/source/onnx.md
justinchuby: I am not entirely sure what the alternative would be?

PR #162498, opened 2025-09-09
mul(sparse, dense): enable proper broadcasting

nikitaved:
* __->__ #162498
* #162490


aten/src/ATen/native/sparse/SparseTensorMath.cpp
```diff
@@ -855,13 +857,22 @@ static Tensor& intersection_binary_op_sparse_dense_out(
...
   }
 
-  const auto d_dim = d.dim();
-  const auto s_dim = s_.dim();
+  // Broadcast inputs {
+  const auto s = [&]() -> auto {
+    if (s_.sizes() != res_shape) {
+      // Always coalesce when sparse broadcasts over dense,
+      // because new sparse dimensions are created and
+      // repeated indices have to be eliminated because of that.
```
pearu, line 865: Perhaps OT to this PR:
Consider the case `s_.shape == res_shape[:s_.sparse_dim()]`, that is, broadcasting is effective only to dense dimensions of a COO tensor.
IIUC, we have multiple choices for broadcasted dimensions: (i) create new sparse dimensions, or (ii) create new dense dimensions, or (iii) create new sparse and dense dimensions. Why do we assume (i) and not (ii), for instance? Which is more efficient (say, memory consumption-wise) in the context of element-wise operations such as multiply?

nikitaved: We will broadcast the dense dims. It is better most of the time since it facilitates proper memory locality (no indexing, likely coalesced access) with a smaller memory footprint (no need to allocate new index).

PR #114068, opened 2023-11-19
Graph-Safe RNG State Exchange for Tensor Parallelism

eee4017:
See #113541

The PR allows for registering and controlling multiple RNG states using indices, ensuring cudagraph-safe operations, and includes both C++ and Python API changes to support this functionality.

cc  @eellison @anijain2305 @jansel @ezyang @ptrblck @csarofeen @mcarilli

Issue #113541, opened 2023-11-13
Graph-Safe RNG State Exchange for Tensor Parallelism


eee4017:
### Feature Overview

This proposal aims to introduce an improved system for managing Random Number Generator (RNG) states in PyTorch, particularly focusing on Tensor Parallelism (TP) and Pipeline Parallelism (PP) under CUDA Graph. The current approach, involving `set_rng_state` and `get_rng_state`, is not compatible with cudagraph-safe operations. To address this, we propose an index-based state management system that extends and refines the existing method of direct state manipulation (via `get_state`/`set_state` in [`CUDAGenerator`](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/cuda/CUDAGeneratorImpl.h)). This new system will use indices to register and control states, ensuring cudagraph-safe RNG state operations within the generator.

### Background

PyTorch's current mechanism for handling RNG state under CUDA Graph involves a specialized design (see [Note [CUDA Graph-safe RNG states]](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/cuda/CUDAGeneratorImpl.h)), designed in #47989 by @mcarilli  ). It uses the PhiloxCudaStates structure to record the intra-graph offset of each kernel and employs a pointer to a one-element, stream-local int64_t device tensor holding an initial offset value.

### Motivation

In TP and PP scenarios, it's common to exchange RNG states, involving setting and getting states to and from Python. This concept, originating from [MegatronLM](https://github.com/NVIDIA/Megatron-LM/blob/main/megatron/core/tensor_parallel/random.py#L129), usually involves a context manager like `TensorParallelRNGTracker` in [`torch/distributed/_tensor/random.py`](https://github.com/pytorch/pytorch/blob/main/torch/distributed/_tensor/random.py) that wraps `set_rng_state`/`get_rng_state`. However, this approach faces challenges with state changes in CUDA Graph, as PhiloxCudaStates are designed for one initial offset value per graph, but there may exist multiple offsets due to state operations in TP and PP.

```python
class TensorParallelRNGTracker:
  ...
  def _distribute_region(self, spec: DTensorSpec):
    ...
    self._device_handle.set_rng_state(self.rng_states["tensor-parallel-rng"])
    try:
      yield
    finally:
      self.rng_states["tensor-parallel-rng"] = self._device_handle.get_rng_state()
```

[...]

PR #162505, opened 2025-09-09
Make torch.cuda.rng_set_state() and torch.cuda.rng_get_state() work during stream capture.

### LINKED ISSUE #162504: torch.utils.checkpoint.checkpoint does not work under stream capture to a cuda graph
### 🐛 Describe the bug

The following does not work today:

```python
import torch

torch.cuda.manual_seed(42)

def fn(x):
    return x * torch.sigmoid(torch.randn(1, device="cuda"))

# initialize device state
fn(torch.ones(1, device="cuda"))

torch.cuda.manual_seed(42)
eager_in = torch.ones(1, device="cuda", requires_grad=True)
eager_out = torch.utils.checkpoint.checkpoint(
    fn, eager_in,
    use_reentrant=False,
    preserve_rng_state=True,
)
eager_in_grad,  = torch.autograd.grad(eager_out, eager_in)

g = torch.cuda.CUDAGraph()
with torch.cuda.graph(g):
    graph_in = torch.ones(1, device="cuda", requires_grad=True)
    graph_out = torch.utils.checkpoint.checkpoint(
        fn, graph_in,
        use_reentrant=False,
        preserve_rng_state=True,
    )
    graph_in_grad,  = torch.autograd.grad(graph_out, graph_in)

torch.cuda.manual_seed(42)
g.replay()
assert torch.allclose(eager_in_grad, graph_in_grad, rtol=0.0, atol=0.0), "Mismatch in gradient outputs"
print(eager_in_grad)
print(graph_in_grad)
```

It will fail with an assertion error from current_seed: https://github.com/pytorch/pytorch/blob/f03d635dc60068d63eda0d13e717036db60219cc/aten/src/ATen/cuda/CUDAGeneratorImpl.cpp#L304

However, that is just the proximal cause of the issue. The real problem is that torch.cuda.get_rng_state() and torch.cuda.set_rng_state() are banned during stream capture, due to a variety of other assertions.

Even if you were to comment out all of the failing assertions in CUDAGeneratorImpl.cpp, the assertion in my above test case would fail because the rng state isn't being tracked correctly.

This matters because people who want to run training at speed of light in models that depend upon activation recomputation cannot do it right now if the training is cpu-bound.

I already have a fix. I will make a PR shortly. Just making this issue for documentation purposes.

### Versions

This problem exists on top of tree.

cc @soulitzer @mcarilli @ezyang @eellison @penguinwu @BoyuanFeng

galv:
Note that this works only in a limited case, where you *don't* change the seed, but change only the offset of the philox generator. This captures the main use case we're interested in: Rewinding the RNG to a previous state. This is done by torch.utils.checkpoint.checkpoint in particular.

Calls to increase() change only the offset, not the seed. Thus, we allow for "no-op" calls to set_seed where the new seed is the same as the old seed. If a user does happen to try to change the seed during stream capture, they will receive an error.

Fixes #162504

cc @pbelevich @mcarilli @ezyang @eellison @penguinwu @BoyuanFeng

### DIFF
--- a/aten/src/ATen/cuda/CUDAGeneratorImpl.cpp
+++ b/aten/src/ATen/cuda/CUDAGeneratorImpl.cpp
@@ -266,11 +266,14 @@ CUDAGeneratorImpl::CUDAGeneratorImpl(
  * See Note [Acquire lock when using random generators]
  */
 void CUDAGeneratorImpl::set_current_seed(uint64_t seed) {
-  at::cuda::assertNotCapturing(
-      "Cannot call CUDAGeneratorImpl::set_current_seed");
-  state_->seed_ = seed;
-  state_->philox_offset_per_thread_ = 0;
-  no_reset_rnn_state_.clear();
+  if (at::cuda::currentStreamCaptureStatus() == at::cuda::CaptureStatus::None) {
+    state_->seed_ = seed;
+    state_->philox_offset_per_thread_ = 0;
+    no_reset_rnn_state_.clear();
+  } else {
+    TORCH_CHECK(state_->seed_ == seed, "CUDAGeneratorImpl::set_current_seed can be called during stream capture only if new seed is the same as the original seed.");
+    // no-op case
+  }
 }
 
 /**
@@ -299,9 +302,6 @@ uint64_t CUDAGeneratorImpl::get_offset() const {
  * Gets the current seed of CUDAGeneratorImpl.
  */
 uint64_t CUDAGeneratorImpl::current_seed() const {
-  // Debatable if current_seed() should be allowed in captured regions.
-  // Conservatively disallow it for now.
-  at::cuda::assertNotCapturing("Cannot call CUDAGeneratorImpl::current_seed");
   return state_->seed_;
 }
 
@@ -346,8 +346,6 @@ c10::intrusive_ptr<c10::TensorImpl> CUDAGeneratorImpl::get_state() const {
  * and size of the internal state.
  */
 void CUDAGeneratorImpl::set_state(const c10::TensorImpl& new_state) {
-  at::cuda::assertNotCapturing(
-      "Please ensure to utilize the CUDAGeneratorImpl::set_state_index method during capturing.");
   static const size_t seed_size = sizeof(uint64_t);
   static const size_t offset_size = sizeof(int64_t);
   static const size_t total_size = seed_size + offset_size;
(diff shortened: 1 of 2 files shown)


ngimel:
cc @drisspg, this also would have been useful for sdpa with dropout, but oh well

aten/src/ATen/cuda/CUDAGeneratorImpl.cpp
```diff
@@ -266,11 +266,14 @@ CUDAGeneratorImpl::CUDAGeneratorImpl(
  * See Note [Acquire lock when using random generators]
  */
 void CUDAGeneratorImpl::set_current_seed(uint64_t seed) {
-  at::cuda::assertNotCapturing(
-      "Cannot call CUDAGeneratorImpl::set_current_seed");
-  state_->seed_ = seed;
-  state_->philox_offset_per_thread_ = 0;
-  no_reset_rnn_state_.clear();
+  if (at::cuda::currentStreamCaptureStatus() == at::cuda::CaptureStatus::None) {
```
ngimel: C10_LIKELY

ngimel (approved):

eqy (approved):

eellison (approved):
Nice ! Would this work for the use case in https://github.com/pytorch/pytorch/pull/114068 as well ? cc @eee4017

eee4017 (approved):

aten/src/ATen/cuda/CUDAGeneratorImpl.cpp
```diff
@@ -346,8 +346,6 @@ c10::intrusive_ptr<c10::TensorImpl> CUDAGeneratorImpl::get_state() const {
  * and size of the internal state.
  */
 void CUDAGeneratorImpl::set_state(const c10::TensorImpl& new_state) {
-  at::cuda::assertNotCapturing(
```
eee4017, line 349: I think we only want to use this PR when we don’t want the offset to increase during the replay stage, right? #114068 was mainly to ensure that offset exchange remain safe during replay (See discussion here #113541). If the user isn’t aware that the offset is not updated in replay and calls `CUDAGeneratorImpl::set_state` during capture, they could end up with incorrect results.

galv: Yeah, the seeds could also be different among the states that get exchanged.

I initially considered using your API, BTW, but it was intrusive to integrate because I needed to detect whether stream capture was happening any time the user might want to call torch.cuda.get_rng_state. If stream capture was happening, I would call the graph safe getters and setters. So I could not make that work for arbitrary user code.

eee4017: 1. For the code here, I means that should you have any warning/reminder for the user if they use the set_state in the capture and get incorrect results unexpectedly, it should be hard for them to debug if there is no assertion.
2. Why do you need to check the capture status when using the graph safe set state? I think we can use it without the capturing.

cyyever (approved):

PR #162508, opened 2025-09-09
Fully native DTensor.__new__

swolchok:
* #163030
* #162968
* __->__ #162508
* #161695

Move the entirety of `__new__` into C++, saving a layer of disable_dynamo and making progress toward all-C++.


torch/csrc/autograd/python_variable.cpp
```diff
@@ -928,6 +864,109 @@ static bool intern_dtensor_strings() {
...
+  DispatchKeySet extra_dispatch_keys;
+  const auto tensor_keys = local_tensor.key_set();
+  if (tensor_keys.has(c10::DispatchKey::Conjugate)) {
+    extra_dispatch_keys = extra_dispatch_keys.add(c10::DispatchKey::Conjugate);
+  }
+  if (tensor_keys.has(c10::DispatchKey::Negative)) {
+    extra_dispatch_keys = extra_dispatch_keys.add(c10::DispatchKey::Negative);
+  }
+
+  py::handle spec = py::handle(r.pyobject(2));
```
albanD, line 948: Does .pyobject() return a new or borrowed reference?

torch/csrc/autograd/python_variable.cpp
```diff
@@ -928,6 +864,109 @@ static bool intern_dtensor_strings() {
...
+  const auto tensor_meta = spec.attr(dtensor_interned_strings.tensor_meta);
```
albanD, line 949: Some error checking from the old code is lost here when this is None?
Also what happens here when the attribute doesn't exist?

torch/csrc/autograd/python_variable.cpp
```diff
@@ -928,6 +864,109 @@ static bool intern_dtensor_strings() {
...
+  py::object py_tensor =
+      py::reinterpret_steal<py::object>(THPVariable_NewWithVar(
+          (PyTypeObject*)cls,
+          tensor,
+          // false is the default
+          /*allow_preexisting_pyobj=*/false,
+          // we know DTensor has __torch_dispatch__; avoid checking again.
+          /*has_torch_dispatch_if_known=*/true));
+  py_tensor.attr(dtensor_interned_strings._spec) = spec;
+  py_tensor.attr(dtensor_interned_strings._local_tensor) = local_tensor;
```
albanD, line 975: I've never seen this before. Is the refcounting acutally correct on this?!

albanD (commented):
Not really sure what the end goal here is and the level of optimization we're aiming for. But pybind is most likely costing you a lot, especially for things like setting a slot attribute?

torch/csrc/autograd/python_variable.cpp
swolchok: borrowed -- it [just vends from the input args array](https://github.com/pytorch/pytorch/blob/main/torch/csrc/utils/python_arg_parser.h#L1156), which is borrowed.

torch/csrc/autograd/python_variable.cpp
swolchok: > what happens here when the attribute doesn't exist?

PyObject_GetAttr() will set an error, and [pybind11 will throw](https://github.com/pybind/pybind11/blob/master/include/pybind11/pytypes.h#L903).

> Some error checking from the old code is lost here when this is None

you can't get an attribute from None, so this change is safe. it's harder to debug, so I suppose we can TORCH_CHECK here.

torch/csrc/autograd/python_variable.cpp
swolchok: why wouldn't it be? it would be an obvious bug in pybind11 if this didn't work

ezyang (approved):

PR #162509, opened 2025-09-09
[dynamo][guards] Prevent framelocals to dict conversion for not required LAMBDA_GUARD

anijain2305:
* #162525
* __->__ #162509

This is a smaller PR to reduce framelocals to dict conversion.


torch/csrc/dynamo/guards.cpp
```diff
@@ -1701,6 +1701,93 @@ class LAMBDA_GUARD : public LeafGuard {
...
+*/
+class LAMBDA_GUARD_NO_FRAMELOCALS : public LAMBDA_GUARD {
+ public:
+  LAMBDA_GUARD_NO_FRAMELOCALS(
+      RootGuardManager* root_guard_manager,
+      py::object guard_check_fn,
+      py::object verbose_code_parts)
+      : LAMBDA_GUARD(root_guard_manager, guard_check_fn, verbose_code_parts) {}
+
+  bool check_nopybind(PyObject* value) override { // borrowed ref
```
williamwen42, line 1777: How are we reaching here if the guard check always goes through https://github.com/pytorch/pytorch/blob/e4174b1fd798f977ba4dae04c01c2936373fd145/torch/csrc/dynamo/guards.cpp#L6526

anijain2305: I think this can happen in debug verbose mode, where we run guards during compilation for validation

anijain2305: Oh this guard is for a leaf guard. Like from_numpy, for which the the value is already accessed through a series of guard accessors.

williamwen42 (approved):

PR #151484, opened 2025-04-16
Add torch.cuda._compile_kernel()

msaroufim:
Followup work on top https://github.com/pytorch/pytorch/pull/149480

Wrapper on top of nvrtc inspired by https://gist.github.com/malfet/2c9a25976dd7396430c38af603f791da from @malfet

Compiling toy kernels with this setup takes 0.01s vs 90s using `load_inline()` on my local H100. This was primarily motivated by the timeouts I was seeing in the popcorn leaderboard but would also be useful to integrate into KernelBench

This PR is in the same spirit as https://github.com/pytorch/pytorch/pull/148972 which was a similar UX for Metal

For now we are planning on landing this as a private function because we expect to iterate both on the user facing API and the internals implementation, will open up a seperate issue to discuss the path towards making this work public and give a broader overview of the state of custom cuda kernel authoring in PyTorch

cc @ptrblck @eqy @jerryzh168

Future work, as a prereq to making the work public
* divup primitive
* support multiple kernels
* Expose _get_nvrtc_version from native code
* interop with torch.compile
* AMD support

PR #162510, opened 2025-09-09
[ROCm] Support torch.cuda._compile_kernel

jammm:
Supports `torch.cuda._compile_kernel` on ROCm. Related to https://github.com/pytorch/pytorch/pull/151484
Tested on Windows with gfx1201. Testing on Linux pending.

### DIFF
--- a/test/test_cuda.py
+++ b/test/test_cuda.py
@@ -6734,7 +6734,6 @@ class TestCudaAutocast(TestAutocast):
     os.environ.get("USE_LEGACY_DRIVER", None) == "1", "Doesn't work with older driver"
 )
 class TestCompileKernel(TestCase):
-    @unittest.skipIf(TEST_WITH_ROCM, "ROCM does not support nvrtc")
     @unittest.skipIf(not TEST_CUDA, "No CUDA")
     def test_compile_kernel(self):
         # Simple vector addition kernel
@@ -6844,7 +6843,6 @@ class TestCompileKernel(TestCase):
         with self.assertRaises(RuntimeError):
             _compile_kernel(invalid_kernel_source, "invalid_kernel")
 
-    @unittest.skipIf(TEST_WITH_ROCM, "ROCM does not support nvrtc")
     @unittest.skipIf(not TEST_CUDA, "No CUDA")
     def test_compile_kernel_large_shared_memory(self):
         kernel_source = """
@@ -6902,14 +6900,20 @@ class TestCompileKernel(TestCase):
         self.assertEqual(output_data.cpu(), expected)
 
         # Test error handling with more than supported shared memory size
-        max_smem = get_device_properties().shared_memory_per_block_optin
+        if torch.version.hip:
+            max_smem = (
+                65536
+                if get_device_properties().gcnArchName not in ["gfx950"]
+                else 160 * 1024
+            )
+        else:
+            max_smem = get_device_properties().shared_memory_per_block_optin
         excessive_shared_mem = max_smem * 2
 
         with self.assertRaises(RuntimeError):
             kernel.set_shared_memory_config(excessive_shared_mem)
 
     @tf32_on_and_off(0.005)
-    @unittest.skipIf(TEST_WITH_ROCM, "ROCM does not support nvrtc")
     @unittest.skipIf(not TEST_CUDA, "No CUDA")
     def test_compile_kernel_advanced(self):
         # Test matrix multiplication
@@ -6960,7 +6964,10 @@ class TestCompileKernel(TestCase):
 
         # Test with different compute capability if specified
         device_props = torch.cuda.get_device_properties(torch.cuda.current_device())
-        compute_cap = f"{device_props.major}{device_props.minor}"
+        if not torch.version.hip:
+            compute_cap = f"{device_props.major}{device_props.minor}"
+        else:
+            compute_cap = f"{device_props.gcnArchName}"
 
         # Recompile with explicit compute capability
         matmul_kernel_explicit = _compile_kernel(
@@ -6979,7 +6986,6 @@ class TestCompileKernel(TestCase):
         # Verify results
         self.assertEqual(C_explicit, expected)
 
-    @unittest.skipIf(TEST_WITH_ROCM, "ROCM does not support nvrtc")
     @unittest.skipIf(not TEST_CUDA, "No CUDA")
     def test_compile_kernel_as_custom_op(self):
         # Define a simple vector addition kernel
@@ -7039,7 +7045,6 @@ class TestCompileKernel(TestCase):
         expected = a + b
         torch.testing.assert_close(result, expected, rtol=1e-5, atol=1e-5)
 
-    @unittest.skipIf(TEST_WITH_ROCM, "ROCM does not support nvrtc")
     @unittest.skipIf(not TEST_CUDA, "No CUDA")
     def test_compile_kernel_custom_op_validation(self):
         kernel_source = """
@@ -7087,7 +7092,6 @@ class TestCompileKernel(TestCase):
         expected = input_data + scalar_val
         torch.testing.assert_close(result, expected, rtol=1e-5, atol=1e-5)
 
-    @unittest.skipIf(TEST_WITH_ROCM, "ROCM does not support nvrtc")
     @unittest.skipIf(not TEST_CUDA, "No CUDA")
     def test_compile_kernel_double_precision(self):
         """Test that Python floats are correctly handled as doubles in kernels."""
@@ -7122,12 +7126,13 @@ class TestCompileKernel(TestCase):
         )
         torch.testing.assert_close(output, expected, rtol=1e-14, atol=1e-14)
 
-    @unittest.skipIf(TEST_WITH_ROCM, "ROCM does not support nvrtc")
     @unittest.skipIf(not TEST_CUDA, "No CUDA")
     def test_compile_kernel_cuda_headers(self):
         """Test that kernels can include and use CUDA headers like cuda_fp16.h."""
         kernel_source = """
+        #ifndef __HIPCC__
         #include <cuda_fp16.h>
+        #endif
 
         extern "C"
         __global__ void half_precision_kernel(__half* output, double input_value, int n) {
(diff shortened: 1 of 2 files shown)


jammm:
cc @msaroufim (sorry for tagging you to the wrong PR earlier!)

msaroufim:
Oh damn very cool! Add some tests and lmk when you need a review

jammm:
> Oh damn very cool! Add (unskip) some tests and lmk when you need a review

Sure! I think simply removing the `@unittest.skipIf(TEST_WITH_ROCM, "ROCM does not support nvrtc")` line should work. How does one usually run those tests?

jammm:
@msaroufim should be ready for review

torch/cuda/_utils.py
```diff
@@ -305,7 +353,7 @@ def _cuda_load_module(
     libcuda = _get_cuda_library()
 
     # Convert PTX to bytes if it's a string
-    if isinstance(ptx, str):
+    if isinstance(ptx, str) and not torch.version.hip:
```
mycpuorg: curious, no need to properly encode here?

torch/cuda/_utils.py
```diff
@@ -159,7 +208,7 @@ def check_nvrtc(result: int) -> None:
     check_nvrtc(libnvrtc.nvrtcGetPTX(prog, ptx))
     libnvrtc.nvrtcDestroyProgram(ctypes.byref(prog))
 
-    return ptx.value
+    return ptx.value if not torch.version.hip else ptx.raw[:ptx_size.value]
```
mycpuorg: nit: this is a bit hard to read, prefer not sure if its a pytorch convention to keep it in cur form?
```python
return ptx.raw if hip else ptx.value
```

mycpuorg (approved):
a couple of nits but looks good.

torch/cuda/_utils.py
jammm: we don't have human readable test so thought it wasn't necessary. But can add it back if needed.

torch/cuda/_utils.py
jammm: Changed it to `return ptx.raw if torch.version.hip else ptx.value.` Would this be okay?

torch/cuda/_utils.py
mycpuorg: I see... let's test this and see what a sample string looks like.

jammm: s/test/text (from my previous comment)

jammm: seems to work fine after undoing this change. Pushed !

torch/cuda/_utils.py
```diff
@@ -159,7 +208,7 @@ def check_nvrtc(result: int) -> None:
     check_nvrtc(libnvrtc.nvrtcGetPTX(prog, ptx))
     libnvrtc.nvrtcDestroyProgram(ctypes.byref(prog))
 
-    return ptx.value
+    return ptx.raw if torch.version.hip else ptx.value
```
jammm, line 242: This was done because hipRTC generates raw CO binaries instead of PTX, and for some reason `.value` causes the string to be truncated, likely due to the presence of `\0` in the string. So we use `.raw` instead.

torch/cuda/_utils.py
```diff
@@ -30,13 +47,35 @@ def _check_cuda(result: int) -> None:
...
+    else:
+        lib = ctypes.CDLL("libhiprtc.so")
+
+    # Provide aliases for HIP RTC functions to match NVRTC API
+    lib.nvrtcGetErrorString = lib.hiprtcGetErrorString
+    lib.nvrtcCreateProgram = lib.hiprtcCreateProgram
+    lib.nvrtcDestroyProgram = lib.hiprtcDestroyProgram
+    lib.nvrtcCompileProgram = lib.hiprtcCompileProgram
+    lib.nvrtcGetPTXSize = lib.hiprtcGetCodeSize
+    lib.nvrtcGetPTX = lib.hiprtcGetCode
```
jammm: Note the alias to `GetCode/GetCodeSize` instead of `GetPTX/GetPTXSize` as HIP doesn't support PTX.

torch/cuda/_utils.py
msaroufim: Let's add this comment to code directly?

torch/cuda/_utils.py
```diff
@@ -98,11 +137,17 @@ def check_nvrtc(result: int) -> None:
     # Get compute capability if not provided
     if compute_capability is None:
         props = torch.cuda.get_device_properties(torch.cuda.current_device())
-        compute_capability = f"{props.major}{props.minor}"
+        if not torch.version.hip:
```
msaroufim: can we reverse the negation logic? `if torch.version.hip` - should be slightly easier to read. Same feedback for most of the negations

torch/cuda/_utils.py
```diff
@@ -10,11 +10,28 @@
 
 # Load CUDA driver and NVRTC
 def _get_cuda_library() -> ctypes.CDLL:
```
msaroufim: Let's refactor into a standalone `_get_gpu_runtime_library()` which calls into either hip or cuda

test/test_cuda.py
```diff
@@ -6896,7 +6895,10 @@ def test_compile_kernel_advanced(self):
 
         # Test with different compute capability if specified
         device_props = torch.cuda.get_device_properties(torch.cuda.current_device())
-        compute_cap = f"{device_props.major}{device_props.minor}"
+        if not torch.version.hip:
+            compute_cap = f"{device_props.major}{device_props.minor}"
+        else:
+            compute_cap = f"{device_props.gcnArchName}"
```
msaroufim, line 6970: is this always a string and do we want the whole one?


Claude says these are all valid

```
  It's a string that can have formats like:
  - Simple: "gfx942", "gfx950", "gfx1100"
  - Complex with features: "gfx90a:sramecc+:xnack-" (architecture with feature flags)
```

msaroufim (changes requested):
A few minor pieces of feedback in addition please look into 
1. Lint
2. Share some benchmark of difference in compilation time relative to load_inline on some toy problems

test/test_cuda.py
jammm: yea we should ideally have the whole one. I don't know if it'll include all the feature flags though.

jammm:
Test run:
```
(venv) PS D:\b\test> python .\test_cuda.py --use-pytest -k "TestCompileKernel" --verbose
================================================================================================================================== test session starts ===================================================================================================================================
platform win32 -- Python 3.12.10, pytest-8.4.1, pluggy-1.6.0 -- D:\jam\venv\Scripts\python.exe
cachedir: .pytest_cache
hypothesis profile 'dev' -> database=None, max_examples=10, suppress_health_check=[HealthCheck.too_slow]
rootdir: D:\b
configfile: pytest.ini
plugins: hypothesis-5.35.1, typeguard-4.3.0
collected 274 items / 270 deselected / 4 selected
Running 4 items in this shard

test_cuda.py::TestCompileKernel::test_compile_kernel PASSED [0.6563s]                                                                                                                                                                                                               [ 25%]
test_cuda.py::TestCompileKernel::test_compile_kernel_advanced PASSED [0.3140s]                                                                                                                                                                                                      [ 50%]
test_cuda.py::TestCompileKernel::test_compile_kernel_as_custom_op PASSED [0.1265s]                                                                                                                                                                                                  [ 75%]
test_cuda.py::TestCompileKernel::test_compile_kernel_custom_op_validation PASSED [0.7426s]                                                                                                                                                                                          [100%]

=========================================================================================================================== 4 passed, 270 deselected in 1.89s ============================================================================================================================
```

As for performance, I compared the `matrix_multiply` kernel with _compile_kernel vs. load_inline (which I had to fix for Windows https://github.com/pytorch/pytorch/pull/162577) and the result on my zen5 machine is ~17 seconds vs. 0.3 seconds on the initial compilation times.

test/test_cuda.py
```diff
@@ -7122,12 +7122,13 @@ def test_compile_kernel_double_precision(self):
         )
         torch.testing.assert_close(output, expected, rtol=1e-14, atol=1e-14)
 
-    @unittest.skipIf(TEST_WITH_ROCM, "ROCM does not support nvrtc")
     @unittest.skipIf(not TEST_CUDA, "No CUDA")
     def test_compile_kernel_cuda_headers(self):
         """Test that kernels can include and use CUDA headers like cuda_fp16.h."""
         kernel_source = """
+        #ifndef __HIPCC__
```
jammm, line 7133: @msaroufim I think we can add a HIPIFY pass before compilation (which is what `load_inline` does) to avoid changes like this, but that can be done as a followup for later, as I have ran out of bandwidth to focus on this at the moment.

msaroufim: It's not too bad, we can just keep this

torch/cuda/_utils.py
```diff
@@ -9,43 +9,76 @@
...
+        lib.nvrtcCreateProgram = lib.hiprtcCreateProgram
+        lib.nvrtcDestroyProgram = lib.hiprtcDestroyProgram
+        lib.nvrtcCompileProgram = lib.hiprtcCompileProgram
+        lib.nvrtcGetPTXSize = lib.hiprtcGetCodeSize
+        lib.nvrtcGetPTX = lib.hiprtcGetCode
+        lib.nvrtcGetProgramLogSize = lib.hiprtcGetProgramLogSize
+        lib.nvrtcGetProgramLog = lib.hiprtcGetProgramLog
+        return lib
 
+    else:
```
msaroufim, line 83: what i had in mind was more 2 seperate functions one for hip and and one for cuda and then

```python
def get_gpu_lib():
  if torch.version.hip:
	get_hip_gpu_lib()
  else:
    get_cuda_gpu_lib()
```

Cause getting the discovery right is kind of annoying and something I'm iterating on here https://github.com/pytorch/pytorch/pull/162691 so ideally want the vendor specific discovery in seperate functions cause I expect these when made correct will be quite long

msaroufim (approved):

Issue #161904, opened 2025-09-01
ZeroBubble and DualPipeV pipeline parallel schedules fail with torch.compiled model


awaelchli:
### 🐛 Describe the bug

The ZeroBubble and DualPipeV schedules don't seem to be compatible with models that are `torch.compiled`. Following the TorchTitan code base, I reduced the problem into a minimal repro example:

```py
"""
Run:

torchrun --nproc_per_node=2 repro.py
"""

import torch
from torch.distributed.device_mesh import init_device_mesh
from torch.distributed.pipelining import PipelineStage
from torch.distributed.pipelining.schedules import get_schedule_class
import torch.nn as nn
from torchtitan.distributed.pipeline_parallel import pipeline_module_split

PP_DEGREE = 2
SCHEDULE_NAME = "ZBVZeroBubble"  # or "DualPipeV"


class Transformer(nn.Module):
    def __init__(self):
        super().__init__()
        self.tok_embeddings = nn.Embedding(128, 32)
        self.layers = torch.nn.ModuleDict()
        for layer_id in range(4):
            self.layers[str(layer_id)] = nn.Linear(32, 32, bias=False)
        self.output = nn.Linear(32, 128, bias=False)

    def forward(self, x):
        x = self.tok_embeddings(x) if self.tok_embeddings else x
        for layer in self.layers.values():
            x = layer(x)
        return self.output(x) if self.output else x


def main() -> None:
    local_rank = torch.distributed.get_node_local_rank()
    device = torch.device("cuda", local_rank)
    torch.distributed.init_process_group("nccl")
    mesh = init_device_mesh("cuda", (2,), mesh_dim_names=("pp",))

    with torch.device("meta"):
        model = Transformer()

    pp_schedule, model_parts, has_first_stage, has_last_stage = pipeline_transformer(
        model,
        mesh,
        device,
        parallelize_transformer,
        loss_fn,
    )
    for m in model_parts:
        m.to_empty(device=device)
        m.train()

    input_ids = torch.randint(0, 128, (8, 4096), device=device)
    labels = input_ids.clone()

    losses: list[torch.Tensor] | None
    targets, losses = (labels, []) if has_last_stage else (None, None)
    if has_first_stage:
        pp_schedule.step(
            input_ids,
            target=targets,
            losses=losses,
        )
    else:
        pp_schedule.step(
            target=targets,
            losses=losses,
        )


def pipeline_transformer(
    model: nn.Module,
    world_mesh,
[...]

PR #162511, opened 2025-09-09
Add torch compile check for ZeroBubble

H-Huang:
* __->__ #162511

Fix https://github.com/pytorch/pytorch/issues/161904

cc @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci


torch/distributed/pipelining/schedules.py
```diff
@@ -3207,3 +3204,24 @@ def _dump_chrometrace(schedule, filename):
...
+
+    Args:
+        stages: List of pipeline stages to check
+        schedule_name: Name of the schedule for error message
+
+    Raises:
+        RuntimeError: If any stage uses torch.compile
+    """
+    for stage in stages:
+        if isinstance(stage.submod, OptimizedModule):
```
awaelchli: Should this check all submodules? Otherwise use cases like in TorchTitan where transformer blocks are compiled individually instead of the root module are not covered by this check?

H-Huang: Good point, thanks!

test/distributed/pipelining/test_schedule.py
fegin, line 282: Should we also verify model.compile(). I'm sure the module architecture is different. So it is better to verify it.

torch/distributed/pipelining/schedules.py
```diff
@@ -3207,3 +3204,27 @@ def _dump_chrometrace(schedule, filename):
...
+    Check if the schedule is compatible with torch.compile.
+
+    Args:
+        stages: List of pipeline stages to check
+        schedule_name: Name of the schedule for error message
+
+    Raises:
+        RuntimeError: If any stage uses torch.compile
+    """
+    for stage in stages:
```
fegin, line 3326: nit, maybe use `if not isinstance(...` to return earlier to avoid 5 nested indentions.

fegin (approved):
LGTM

PR #162514, opened 2025-09-09
[aoti] Add mtime to precompiled header hash

benjaminglass1:
* __->__ #162514

Fixes GCC errors when a header to be precompiled is touched, but the preprocessed content remains the same. This is apparently not allowed, so add an additional field to the deduplication hash with modification time.

### DIFF
--- a/torch/_inductor/codecache.py
+++ b/torch/_inductor/codecache.py
@@ -2196,14 +2196,14 @@ end
                 header_file = _get_cpp_wrapper_header(
                     device_type, aot_mode=graph.aot_mode
                 )
-                wrapper_build_options.precompiled_header = _precompile_header(
+                wrapper_build_options.precompiled_header = _precompile_torch_header(
                     header_file,
                     cpp_command,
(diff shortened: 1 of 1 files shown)


benjaminglass1:
@hameerabbasi This should fix the issue you sporadically run into.

desertfire (commented):
Is it possible to add a test for this?

benjaminglass1:
@desertfire With the way the function is currently structured, I think that would be a bad idea, since we would have to touch and precompile a header. I could possibly break some of this apart to make it more testable, but the core logic is not particularly testable within `torch` itself.

I will note that I am comforted that it took this long for an issue to crop up with this logic, which implies to me that this fix will likely finish the job in terms of invalidating existing precompilations.

desertfire:
> @desertfire With the way the function is currently structured, I think that would be a bad idea, since we would have to touch and precompile a header. 

By "bad idea", do you mean it will trigger a long recompilation? If compilation time is an issue, we can run it as a periodic test.

eellison (approved):
deferring to @desertfire on this one

PR #162529, opened 2025-09-09
[c10d] Add dist.init()

kwen2501:
* #162571
* #162549
* #163058
* #162770
* #162545
* __->__ #162529

Mainly to support a case where user don't want to call `init_process_group` which instantiates a real group that may not be used.

What `dist.init()` does:
```
    Initialize a distributed environment for current job. Once this API is
    called, `get_world_size()` is guaranteed to return the number of processes
    in current job, and `get_rank()` is guaranteed to return the rank of current
    process in the job.
```

With the proposed API, user can get group of interest via, e.g. DeviceMesh creation:
```
dist.init()
mesh = dist.init_device_mesh(...)
group = mesh.get_dim(...)
```

### DIFF
--- a/.lintrunner.toml
+++ b/.lintrunner.toml
@@ -140,6 +140,8 @@ include_patterns = [
 ]
 exclude_patterns = [
     '**/fb/**',
+    # TODO: Remove this once we clean up the code
+    'torch/distributed/distributed_c10d.py',
 ]
 command = [
     'python3',
--- a/test/distributed/fsdp/test_fsdp_sharded_grad_scaler.py
+++ b/test/distributed/fsdp/test_fsdp_sharded_grad_scaler.py
@@ -34,6 +34,7 @@ from torch.testing._internal.common_utils import (
     instantiate_parametrized_tests,
     parametrize,
     run_tests,
+    skip_but_pass_in_sandcastle,
     TEST_WITH_DEV_DBG_ASAN,
     TestCase,
 )
@@ -136,6 +137,7 @@ class TestShardGradScaler(TestCase):
         found_inf = scaler._unscale_grads_(opt, inv_scale, found_inf)[s1.device]
         self.assertEqual(found_inf, 1.0)
 
+    @skip_but_pass_in_sandcastle("DummyProcessGroup + collective is not supported now")
     @unittest.skipIf(
         amp_definitely_not_available(), "no supported device (cuda, xla) found"
     )
--- a/test/distributed/tensor/test_dtensor_compile.py
+++ b/test/distributed/tensor/test_dtensor_compile.py
@@ -45,6 +45,7 @@ from torch.testing._internal.common_utils import (
     instantiate_parametrized_tests,
     parametrize,
     run_tests,
+    skip_but_pass_in_sandcastle,
     skipIfHpu,
     skipIfTorchDynamo,
     TEST_CUDA,
@@ -627,6 +628,9 @@ def forward(self, b_parametrizations_buffer_original0, x):
             tmp_dt._local_tensor.stride(), tmp_dt_fake._local_tensor.stride()
         )
 
+    @skip_but_pass_in_sandcastle(
+        "InternalTorchDynamoError: `tensor` needs to be a `FakeTensor`wrapped by this instance of Dynamo"
+    )
     @unittest.skipIf(not HAS_GPU, "Inductor+gpu needs triton and recent GPU arch")
     def test_dtensor_contiguous_dtensor_noncontiguous_local_as_tangent(self):
         # Partial -> Shard on an unbalanced tensor results in:
--- a/torch/distributed/distributed_c10d.py
+++ b/torch/distributed/distributed_c10d.py
@@ -1755,6 +1755,12 @@ def init_process_group(
     # Initialize distributed environment if user didn't call it
     # (backward compatible purpose)
     if _World.maybe_get() is None:
+        # "fake" backend uses specific store
+        if backend == "fake":
+            from torch.testing._internal.distributed.fake_pg import FakeStore
+
+            store = store or FakeStore()
+
         init(world_size, rank, store, init_method=init_method)
 
     world = _World.get()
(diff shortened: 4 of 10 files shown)


torch/distributed/distributed_c10d.py
```diff
@@ -1526,6 +1551,83 @@ def _set_pg_timeout(timeout: timedelta, group: Optional[ProcessGroup] = None) ->
         backend._set_default_timeout(timeout)
 
 
+@_exception_logger
+@_time_logger
+def init(
```
d4l3k, line 1560: Should this be renamed `init_metadata` or something? Just `init()` implies it does everything rather than a restricted subset of initialization.

torch/distributed/distributed_c10d.py
```diff
@@ -1526,6 +1551,83 @@ def _set_pg_timeout(timeout: timedelta, group: Optional[ProcessGroup] = None) ->
...
+
+    if store is None:
+        # Create one from environment variables
+        rendezvous_iterator = rendezvous(
+            "env://",
+            rank,
+            world_size,
+            # TODO: should we support configurable timeout for rendezvous?
+        )
+        store, rank, world_size = next(rendezvous_iterator)
```
d4l3k, line 1647: this is interesting -- we may not actually want to create a global store for performance reasons. tcpstore is fairly cheap but still hmm

we don't actually need it for get_rank()/get_world_size()

torch/distributed/distributed_c10d.py
```diff
@@ -1526,6 +1551,83 @@ def _set_pg_timeout(timeout: timedelta, group: Optional[ProcessGroup] = None) ->
...
+    #
+    # Since this API must be called before all distributed code being compiled,
+    # clearing the cache here should be safe.
+    if "torch._dynamo" in sys.modules:
+        torch._dynamo.trace_rules.clear_lru_cache()
+
+    if store is None:
+        # Create one from environment variables
+        rendezvous_iterator = rendezvous(
+            "env://",
```
d4l3k: should we be consuming init_method to match init_process_group?

torch/distributed/distributed_c10d.py
fduwjj: Can we pass in the store so folks can use other stuff for example zues for rendezvous? Not sure...

.lintrunner.toml
```diff
@@ -140,6 +140,8 @@ include_patterns = [
 ]
 exclude_patterns = [
     '**/fb/**',
+    # TODO: Remove this once we clean up the code
+    'torch/distributed/distributed_c10d.py',
```
ezyang: I don't think this works, but I am also unsure why you changing this file triggered type errors. (To exclude type checking, put a global suppression on the file instead)

ezyang:
What's going on with the test failures?

torch/distributed/distributed_c10d.py
```diff
@@ -1526,6 +1551,83 @@ def _set_pg_timeout(timeout: timedelta, group: Optional[ProcessGroup] = None) ->
...
+def init(
+    world_size: int = -1,
+    rank: int = -1,
+    store: Optional[Store] = None,
+) -> None:
+    """
+    Initialize a distributed environment for current job. Once this API is
+    called, `get_world_size()` is guaranteed to return the number of processes
+    in current job, and `get_rank()` is guaranteed to return the rank of current
+    process in the job.
```
ezyang, line 1570: Can the doc explain why you want to do this? (Namely, to skip PG initialization). Can we describe exactly how many comms this initialization will take?

kwen2501:
> What's going on with the test failures?

@ezyang Mostly from (the need for) backward support of file-based rendezvous -- which I tend to think "nah.." when coding the PR. But should be easily fixable.

torch/distributed/distributed_c10d.py
kwen2501: Thanks will add.

.lintrunner.toml
kwen2501: Yeah, it comes from this change:
```
-  _world = _World()
+  _world: Optional[_World] = None
```
Thanks for suggesting the global suppression.

kwen2501:
@wconstab thanks for looking at it. The PR is just a sugarcoat of your "counter proposal".
Its little additional is that functions like `dist.get_rank()` lets user call it at any depth of the program stack.

wconstab:
Yea it's just a trade off, convenience vs "not having any global state". I'm ok either way

torch/distributed/distributed_c10d.py
kwen2501: The API accepts a store. @fduwjj

torch/distributed/distributed_c10d.py
kwen2501: Done, thanks!

torch/distributed/distributed_c10d.py
```diff
@@ -1,4 +1,8 @@
 # mypy: allow-untyped-defs
+# TODO: added `ignore-errors` to ignore type checking on _world. Clean them up
+# and remove.
+# mypy: ignore-errors
```
Skylion007, line 4: Use the not_none typing utility

torch/distributed/distributed_c10d.py
```diff
@@ -1526,6 +1555,105 @@ def _set_pg_timeout(timeout: timedelta, group: Optional[ProcessGroup] = None) ->
         backend._set_default_timeout(timeout)
 
 
+@_exception_logger
```
Skylion007, line 1558: I hope these decorators use TypeVar or ParamSpec to preserve typing info. Mypy would complain if they didn't, but we are ignoring errors on this file now so unsure

test/distributed/fsdp/test_fsdp_sharded_grad_scaler.py
```diff
@@ -136,6 +137,7 @@ def test_scaling_unscaling_sparse(self):
         found_inf = scaler._unscale_grads_(opt, inv_scale, found_inf)[s1.device]
         self.assertEqual(found_inf, 1.0)
 
+    @skip_but_pass_in_sandcastle("DummyProcessGroup + collective is not supported now")
```
fduwjj, line 140: hmmm is this test important? cc: @weifengpy

test/distributed/tensor/test_dtensor_compile.py
```diff
@@ -627,6 +628,9 @@ def fn(x):
             tmp_dt._local_tensor.stride(), tmp_dt_fake._local_tensor.stride()
         )
 
+    @skip_but_pass_in_sandcastle(
+        "InternalTorchDynamoError: `tensor` needs to be a `FakeTensor`wrapped by this instance of Dynamo"
+    )
     @unittest.skipIf(not HAS_GPU, "Inductor+gpu needs triton and recent GPU arch")
     def test_dtensor_contiguous_dtensor_noncontiguous_local_as_tangent(self):
```
fduwjj, line 635: cc: @XilunWu @zpcore

PR #162534, opened 2025-09-09
[CuTe] Add type for CuTe layout via claude

fduwjj:
* #161106
* #161016
* #162690
* #162414
* __->__ #162534
* #162413

This PR mostly is a cosmetic change using Claude to add types for copied PyCute code.
We removed all suppressions of linters and add type checker, type alias and mypy ignore(if needed) so that the pycute code will be checked by linter.

### DIFF
--- a/torch/distributed/_pycute/int_tuple.py
+++ b/torch/distributed/_pycute/int_tuple.py
@@ -36,13 +36,13 @@ Functions for manipulating IntTuples
 
 from functools import reduce
 from itertools import chain
-from typing import Optional, TypeGuard, Union
+from typing import Optional, TypeAlias, TypeGuard, Union
 
 from .typing import Integer
 
 
 # Type aliases for better readability
-IntTuple = Union[int, tuple["IntTuple", ...]]
+IntTuple: TypeAlias = Union[int, tuple["IntTuple", ...]]
 
 
 def is_int(x: object) -> TypeGuard[int]:
--- a/torch/distributed/_pycute/int_tuple.py
+++ b/torch/distributed/_pycute/int_tuple.py
@@ -76,11 +76,11 @@ def product(a: IntTuple) -> int:
 
 
 def inner_product(a: IntTuple, b: IntTuple) -> int:
-    if is_tuple(a):  # tuple tuple
-        assert len(a) == len(b)  # type: ignore[arg-type]
-        return sum(inner_product(x, y) for x, y in zip(a, b))  # type: ignore[arg-type,misc]
+    if is_tuple(a) and is_tuple(b):  # tuple tuple
+        assert len(a) == len(b)
+        return sum(inner_product(x, y) for x, y in zip(a, b))
     else:  # "int" "int"
-        assert not is_tuple(b)
+        assert not is_tuple(a) and not is_tuple(b)
         return a * b
 
 
@@ -153,17 +153,18 @@ def idx2crd(
         stride = prefix_product(shape)
 
     if is_tuple(idx):
-        if is_tuple(shape):  # tuple tuple tuple
-            assert len(idx) == len(shape) and len(stride) == len(shape)  # type: ignore[arg-type]
-            return tuple(idx2crd(i, s, d) for i, s, d in zip(idx, shape, stride))  # type: ignore[arg-type]
+        if is_tuple(shape) and is_tuple(stride):  # tuple tuple tuple
+            assert len(idx) == len(shape) and len(stride) == len(shape)
+            return tuple(idx2crd(i, s, d) for i, s, d in zip(idx, shape, stride))
         else:  # tuple "int" "int"
             raise AssertionError("Invalid combination: tuple with int stride")
     else:
-        if is_tuple(shape):  # "int" tuple tuple
-            assert len(shape) == len(stride)  # type: ignore[arg-type]
-            return tuple(idx2crd(idx, s, d) for s, d in zip(shape, stride))  # type: ignore[arg-type]
+        if is_tuple(shape) and is_tuple(stride):  # "int" tuple tuple
+            assert len(shape) == len(stride)
+            return tuple(idx2crd(idx, s, d) for s, d in zip(shape, stride))
         else:  # "int" "int" "int"
-            return (idx // stride) % shape  # type: ignore[operator,return-value]  # all are ints after type checks
+            assert not is_tuple(shape) and not is_tuple(stride)
+            return (idx // stride) % shape  # all are ints after type checks
 
 
 def crd2idx(
@@ -173,24 +174,25 @@ def crd2idx(
         stride = prefix_product(shape)
 
     if is_tuple(crd):
-        if is_tuple(shape):  # tuple tuple tuple
-            assert len(crd) == len(shape) and len(stride) == len(shape)  # type: ignore[arg-type]
-            return sum(crd2idx(c, s, d) for c, s, d in zip(crd, shape, stride))  # type: ignore[arg-type,misc]
+        if is_tuple(shape) and is_tuple(stride):  # tuple tuple tuple
+            assert len(crd) == len(shape) and len(stride) == len(shape)
+            return sum(crd2idx(c, s, d) for c, s, d in zip(crd, shape, stride))
         else:  # tuple "int" "int"
             raise AssertionError(f"Invalid combination: crd={crd}, shape={shape}")
     else:
         if crd is None:
             crd = 0
 
-        if is_tuple(shape):  # "int" tuple tuple
-            assert len(shape) == len(stride)  # type: ignore[arg-type]
+        if is_tuple(shape) and is_tuple(stride):  # "int" tuple tuple
+            assert len(shape) == len(stride)
             result = 0
             for i in range(len(shape) - 1):
-                result += crd2idx(crd % product(shape[i]), shape[i], stride[i])  # type: ignore[operator,index,arg-type]
-                crd = crd // product(shape[i])  # type: ignore[operator,index]
-            return result + crd2idx(crd, shape[-1], stride[-1])  # type: ignore[index,arg-type]
+                result += crd2idx(crd % product(shape[i]), shape[i], stride[i])
+                crd = crd // product(shape[i])
+            return result + crd2idx(crd, shape[-1], stride[-1])
         else:  # "int" "int" "int"
-            return crd * stride  # type: ignore[operator,return-value]  # all are ints after type checks
+            assert not is_tuple(shape) and not is_tuple(stride)
+            return crd * stride  # all are ints after type checks
 
 
 # Transform crd into the dst_shape's iteration space
--- a/torch/distributed/_pycute/int_tuple.py
+++ b/torch/distributed/_pycute/int_tuple.py
@@ -36,7 +36,8 @@ Functions for manipulating IntTuples
 
 from functools import reduce
 from itertools import chain
-from typing import Optional, TypeAlias, TypeGuard, Union
+from typing import Optional, Union
+from typing_extensions import TypeAlias, TypeGuard
 
 from .typing import Integer
 
--- a/torch/distributed/_pycute/int_tuple.py
+++ b/torch/distributed/_pycute/int_tuple.py
@@ -1,4 +1,3 @@
-# mypy: ignore-errors
 #################################################################################################
 #
 # Copyright (c) 2023 - 2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
@@ -37,31 +36,31 @@ Functions for manipulating IntTuples
 
 from functools import reduce
 from itertools import chain
-from typing import Any, Optional, Union
+from typing import Optional, TypeGuard, Union
 
 from .typing import Integer
 
 
 # Type aliases for better readability
-IntTuple = Union[int, tuple[Any, ...]]
+IntTuple = Union[int, tuple["IntTuple", ...]]
 
 
-def is_int(x: Any) -> bool:
+def is_int(x: object) -> TypeGuard[int]:
     return isinstance(x, Integer)
 
 
-def is_tuple(x: Any) -> bool:
+def is_tuple(x: object) -> TypeGuard[tuple]:
     return isinstance(x, tuple)
 
 
-def flatten(t: IntTuple) -> tuple[Any, ...]:
+def flatten(t: IntTuple) -> tuple[int, ...]:
     if is_tuple(t):
         if len(t) == 0:
             return ()
         else:
             return tuple(i for a in t for i in flatten(a))
     else:
-        return (t,)
+        return (t,)  # type: ignore[return-value]  # t is int, converted to tuple[int]
 
 
 def signum(a: int) -> int:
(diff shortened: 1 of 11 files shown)


torch/distributed/_pycute/int_tuple.py
```diff
@@ -39,20 +37,24 @@
...
 
 from .typing import Integer
 
 
-def is_int(x):
+# Type aliases for better readability
+IntTuple = Union[int, tuple[Any, ...]]
+
+
+def is_int(x: Any) -> bool:
```
ezyang: Any is bad, object is better. Additionally, you can use TypeGuard here to refine the type at call site.

ezyang: I won't repeat this for other Any instances

torch/distributed/_pycute/int_tuple.py
```diff
@@ -39,20 +37,24 @@
...
     return isinstance(x, Integer)
 
 
-def is_tuple(x):
+def is_tuple(x: Any) -> bool:
     return isinstance(x, tuple)
 
 
-def flatten(t):
+def flatten(t: IntTuple) -> tuple[Any, ...]:
```
ezyang: tuple[int, ...]

torch/distributed/_pycute/int_tuple.py
```diff
@@ -206,18 +214,21 @@ def crd2crd(crd, dst_shape, src_shape=None):
 
 
 # Filter trg according to crd: keep only elements of trg that are paired with None
-def slice_(crd: Union[None, tuple, int], trg: Union[tuple, int]):
+def slice_(crd: Union[None, tuple, int], trg: Union[tuple, int]) -> Union[tuple, int]:
```
ezyang, line 219: This type is kind of bad but the original type was also kind of bad so whatever

torch/distributed/_pycute/layout.py
```diff
@@ -38,39 +36,47 @@
...
-    def __len__(self):
+    def __len__(self) -> int:
         if is_tuple(self.shape):
             return len(self.shape)
         else:
             return 1
 
     # operator ()    (map coord to idx)
-    def __call__(self, *args):
+    def __call__(self, *args: Any) -> Union["Layout", int]:
```
ezyang: I think a better annot than Any is possible, it's like Optional[Sequence[...]], where ... is whatever slice_/crd2idx takes

torch/distributed/_pycute/layout.py
```diff
@@ -312,7 +320,9 @@ def left_inverse(layout):
 
 # Split a layout by the composition of B and the "rest"
 # Use tuples-of-layouts to perform this operation by-mode and None as no-op
-def logical_divide(layoutA, layoutB):
+def logical_divide(
+    layoutA: Layout, layoutB: Optional[Union[Layout, IntTuple, tuple[Any, ...]]]
```
ezyang: An alias for  Optional[Union[Layout, IntTuple, tuple[Any, ...]]] would be better; also the Any is too grabby

torch/distributed/_pycute/layout.py
```diff
@@ -385,28 +401,36 @@ def hier_unzip(splitter, layoutA, layoutB):
...
-def tiled_product(layoutA, layoutB):
+def tiled_product(
+    layoutA: Layout, layoutB: Optional[Union[Layout, IntTuple, tuple[Any, ...]]]
+) -> Layout:
     result = zipped_product(layoutA, layoutB)
     return make_layout([result[0]] + [result[1][i] for i in range(len(result[1]))])
 
 
-def slice_and_offset(crd: tuple, layout: Layout):
+def slice_and_offset(crd: tuple[Any, ...], layout: Layout) -> tuple[Layout, int]:
```
ezyang: Any here can be improved

ezyang:
You should be able to feed my CR comments directly to Claude. Please review the output yourself though.

torch/distributed/_pycute/int_tuple.py
```diff
@@ -39,20 +37,24 @@
...
 from itertools import chain
-from typing import Union
+from typing import Any, Optional, Union
 
 from .typing import Integer
 
 
-def is_int(x):
+# Type aliases for better readability
+IntTuple = Union[int, tuple[Any, ...]]
```
Skylion007: Make explicitly as TypeAlias from Typing_Extensions

torch/distributed/_pycute/int_tuple.py
Skylion007: Better than TypeGuard, feel free to use a typing_extension.TypeIs here. It fits better and gives more info the type checker

torch/distributed/_pycute/int_tuple.py
```diff
@@ -39,20 +37,24 @@
...
+
+
+def is_int(x: Any) -> bool:
     return isinstance(x, Integer)
 
 
-def is_tuple(x):
```
Skylion007, line 51: Same TypeIs is better here

torch/distributed/_pycute/int_tuple.py
Skylion007: Why not tuple[int, ...] as second argument in union

torch/distributed/_pycute/layout.py
```diff
@@ -38,39 +35,61 @@
...
-
-from .int_tuple import *
+from typing import Optional, TypeGuard, Union
+
+from .int_tuple import (
+    crd2idx,
+    flatten,
+    has_none,
+    IntTuple,
+    is_int,
```
Skylion007, line 46: Please explicitly type as TypeAlias. Removes ambiguity for type checker and IDEs.

torch/distributed/_pycute/layout.py
```diff
@@ -38,39 +35,61 @@
...
+    is_tuple,
+    prefix_product,
+    product,
+    slice_,
+)
+
+
+# Type aliases
```
Skylion007, line 54: TypeIs here too

torch/distributed/_pycute/int_tuple.py
Skylion007: TLDR TypeGuard tells the type checker that if it's true, it's definitely the object type. TypeIs also does narrowing on the if false branch and should be generally used when do you a return isinstance check as the only content of the function.

torch/distributed/_pycute/int_tuple.py
fduwjj: hmm we need to support python3.9 so we probably cannot use TypeIs here.

torch/distributed/_pycute/int_tuple.py
```diff
@@ -39,72 +36,76 @@
...
 
 from .typing import Integer
 
 
-def is_int(x):
+# Type aliases for better readability
+IntTuple = Union[int, tuple["IntTuple", ...]]
+
+
+def is_int(x: object) -> TypeGuard[int]:
```
Skylion007: Should actually be TypeIs not TypeGuard, see other comment

torch/distributed/_pycute/int_tuple.py
```diff
@@ -39,72 +36,76 @@
...
-def flatten(t):
+def flatten(t: IntTuple) -> tuple[int, ...]:
     if is_tuple(t):
         if len(t) == 0:
             return ()
         else:
             return tuple(i for a in t for i in flatten(a))
     else:
-        return (t,)
+        return (t,)  # type: ignore[return-value]  # t is int, converted to tuple[int]
```
Skylion007: Replacing the TypeGuard with TypeIs should let you remove this type ignore.

torch/distributed/_pycute/int_tuple.py
```diff
@@ -39,72 +36,76 @@
...
+def inner_product(a: IntTuple, b: IntTuple) -> int:
     if is_tuple(a):  # tuple tuple
-        assert len(a) == len(b)
-        return sum(inner_product(x, y) for x, y in zip(a, b))
+        assert len(a) == len(b)  # type: ignore[arg-type]
+        return sum(inner_product(x, y) for x, y in zip(a, b))  # type: ignore[arg-type,misc]
     else:  # "int" "int"
         assert not is_tuple(b)
-        return a * b
+        return a * b  # type: ignore[operator,return-value]  # both are ints after type checks
```
Skylion007: And this one etc...

torch/distributed/_pycute/int_tuple.py
fduwjj: I see we have CI to test python3.9 so TypeIs might be available in 3.9?

Skylion007: It's available via the typing_extensions backport and already used extensively throughout the codebase.

Skylion007: from typing_extensions import TypeIs

torch/distributed/_pycute/int_tuple.py
```diff
@@ -39,72 +36,76 @@
 
 from functools import reduce
 from itertools import chain
-from typing import Union
+from typing import Optional, TypeAlias, TypeGuard, Union
```
Skylion007: I think you might need to import TypeAlias and TypeGuard from typing_extensions too if you want to use it

torch/distributed/_pycute/int_tuple.py
Skylion007: https://typing-extensions.readthedocs.io/en/latest/

torch/distributed/_pycute/int_tuple.py
Skylion007: Yeah TypeGuard introduced in 3.10 so you need to use the typing_extensions backport

fduwjj: right this works for python 3.10+ (or 3.11+)

fduwjj: Done.

torch/distributed/_pycute/int_tuple.py
```diff
@@ -39,72 +36,77 @@
...
 
 from .typing import Integer
 
 
-def is_int(x):
+# Type aliases for better readability
+IntTuple: TypeAlias = Union[int, tuple["IntTuple", ...]]
+
+
+def is_int(x: object) -> TypeGuard[int]:
```
Skylion007:
```suggestion
def is_int(x: object) -> TypeIs[int]:

File: torch/distributed/_pycute/int_tuple.py [hunk 16]
```diff
@@ -39,72 +36,77 @@
 
 from functools import reduce
 from itertools import chain
-from typing import Union
+from typing import Optional, Union
+from typing_extensions import TypeAlias, TypeGuard
```
Skylion007:
```suggestion
from typing_extensions import TypeAlias, TypeIs

File: torch/distributed/_pycute/int_tuple.py [hunk 17]
```diff
@@ -39,72 +36,77 @@
...
-def flatten(t):
+def flatten(t: IntTuple) -> tuple[int, ...]:
     if is_tuple(t):
         if len(t) == 0:
             return ()
         else:
             return tuple(i for a in t for i in flatten(a))
     else:
-        return (t,)
+        return (t,)  # type: ignore[return-value]  # t is int, converted to tuple[int]
```
Skylion007:
```suggestion
        return (t,)
```
Wont be necessary with TypeIs

torch/distributed/_pycute/int_tuple.py
```diff
@@ -39,72 +36,77 @@
...
+# Type aliases for better readability
+IntTuple: TypeAlias = Union[int, tuple["IntTuple", ...]]
+
+
+def is_int(x: object) -> TypeGuard[int]:
     return isinstance(x, Integer)
 
 
-def is_tuple(x):
+def is_tuple(x: object) -> TypeGuard[tuple]:
```
Skylion007:
```suggestion
def is_tuple(x: object) -> TypeIs[tuple]:

File: torch/distributed/_pycute/int_tuple.py (hunk 15 above)
Comment (fduwjj) [reply torch/distributed/_pycute/int_tuple.py]: Hmm I think as I have explained, TypeIS is not available in python3.9 while we have CI for that. So we won't take your comment for now. And down the road, if we bump python version for PT, We will definitely use TypeIs. Does that sound good?

Comment (Skylion007) [reply torch/distributed/_pycute/int_tuple.py]: @fduwjj  Not, It is available through `typing_extensions`, not `typing` in Python 3.9. `typing_extensions` has backports for `TypeIs`. You are already using the TypeGuard backport from typing_extensions in this same PR. It's already extensively used throughout the codebase: https://github.com/search?q=repo%3Apytorch%2Fpytorch%20TypeIs&type=code TypeGuard is not available in typing until 3.10 and we are already using it in typing_extensions backport.

File: torch/distributed/_pycute/layout.py [hunk 19]
```diff
@@ -38,39 +35,62 @@
 """
 
 from itertools import chain
-
-from .int_tuple import *
+from typing import Optional, Union
+from typing_extensions import TypeAlias, TypeGuard
```
Skylion007:
```suggestion
from typing_extensions import TypeAlias, TypeIs
```

torch/distributed/_pycute/layout.py
```diff
@@ -38,39 +35,62 @@
...
+    Union[int, IntTuple, tuple[object, ...]]
+]  # Input for slice_ and crd2idx functions
 
 
 class LayoutBase:
     pass
 
 
-def is_layout(x):
+def is_layout(x: object) -> TypeGuard["Layout"]:
```
Skylion007:
```suggestion
def is_layout(x: object) -> TypeIs["Layout"]:
```

Skylion007 (commented):
Use TypeIs from here from the typing_extensions, which is available on Python 3.9 as it's a pip package which is a dependency of PyTorch and is not part of the standard library: `typing` which is bundled in 3.9

torch/distributed/_pycute/int_tuple.py
fduwjj: oh I see, if that is the case, I think I can use `TypeIs` for no problem.

ezyang (approved):
You should try to do Aaron's suggestions, but I will go ahead and preemptively approve here.

Skylion007 (approved):

torch/distributed/_pycute/int_tuple.py
```diff
@@ -62,40 +64,40 @@ def flatten(t):
...
     else:
         return a
 
 
-def inner_product(a, b):
+def inner_product(a: IntTuple, b: IntTuple) -> int:
     if is_tuple(a):  # tuple tuple
-        assert len(a) == len(b)
-        return sum(inner_product(x, y) for x, y in zip(a, b))
+        assert len(a) == len(b)  # type: ignore[arg-type]
```
Skylion007: You need to add an assert that b is also a tuple. Thats why it's complaining. If you do that, should be able to remove the type ignore

torch/distributed/_pycute/int_tuple.py
```diff
@@ -139,56 +141,62 @@ def prefix_product(a, init=1):
...
-def idx2crd(idx, shape, stride=None):
+def idx2crd(
+    idx: IntTuple, shape: IntTuple, stride: Optional[IntTuple] = None
+) -> IntTuple:
     if stride is None:
         stride = prefix_product(shape)
 
     if is_tuple(idx):
         if is_tuple(shape):  # tuple tuple tuple
-            assert len(idx) == len(shape) and len(idx) == len(stride)
```
Skylion007, line 153: Need to assert stride is tuple, should remove need for type ignore

torch/distributed/_pycute/int_tuple.py
```diff
@@ -139,56 +141,64 @@ def prefix_product(a, init=1):
...
+        if is_tuple(shape) and is_tuple(stride):  # "int" tuple tuple
             assert len(shape) == len(stride)
             result = 0
             for i in range(len(shape) - 1):
                 result += crd2idx(crd % product(shape[i]), shape[i], stride[i])
                 crd = crd // product(shape[i])
             return result + crd2idx(crd, shape[-1], stride[-1])
         else:  # "int" "int" "int"
-            return crd * stride
+            assert not is_tuple(shape) and not is_tuple(stride)
```
Skylion007, line 194: slight nit, better to split into different assert so it's easier to know which failed

torch/distributed/_pycute/layout.py
```diff
@@ -38,39 +35,62 @@
...
+    Union[int, IntTuple, tuple[object, ...]]
+]  # Input for slice_ and crd2idx functions
 
 
 class LayoutBase:
     pass
 
 
-def is_layout(x):
+def is_layout(x: object) -> TypeIs["Layout"]:
```
Skylion007, line 67: This isn't quite right because Layout is the subclass, but the LayoutBase is kind of useless here because it dosen't contain any abc methods etc so this is fine in practice

Skylion007: Keep in mind that if you intend to add other LayoutBase subclasses this may need to be updated

torch/distributed/_pycute/int_tuple.py
fduwjj: Agree, but in practice their types should always be same.. That's why I really don't like the original pycute implementation because it is too generic. I will keep this one as it is for now.

PR #162542, opened 2025-09-09
[CP]Introduce ContextParallal plan for parallelize_module()

fegin:
* #165039
* #163185
* #164500
* __->__ #162542

**Motivation**

Since FlexAttention and SDPA are both functions, not modules, we have tried numerous mechanisms to dispatch FlexAttention and SDPA to customized call paths so that we can inject the CP logic. Unfortunately, all of these approaches have their own composability issues with different techniques.

**Candidate Approaches**

1. Ask users to write a module to wrap FlexAttention/SDPA and use `parallelize_module` to install a forward hook.

   - Pros: This is similar to how we do TP.
   - Cons: 1) It is cumbersome for users as they need to create a new module. 2) We need two places to parallelize the CP, as a context_parallel context manager is still required for splitting the inputs.

2. Provide a function wrapper.

   - Pros: Users just need to replace their FlexAttention/SDPA calls with the wrapper.
   - Cons: It is not the same API, though we can maintain the API signatures to be the same as the core API.

**Summary**

~~This PR implements approach 2 and refactor the code in such a way that most code can be used by option approach 1, which will be introduced in another PR.~~

We changed this PR to implement option 1 as people like option 1 due to the consistency with the existing parallelisms. But this PR can also serve the foundation to implement option 2, which was the early version of this PR.

This PR also changes `create_cp_block_mask` logic since we now only focus on ModuleWrapper approach which doesn't require to hack the seq_len field in a BlockMask.

This PR also removes TorchFunctionMode dispatcher mode as it doesn't work well with SAC.

cc @H-Huang @awgu @wanchaol @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci


torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1178,6 +1132,138 @@ def _context_parallel_buffers(
...
+    def __init__(self, seq_dim: int, attention_type: AttentionType) -> None:
+        super().__init__()
+        self.seq_dim = seq_dim
+        self.attention_type = attention_type
+
+        # Used by FlexAttention
+        self._block_mask: Optional[BlockMask] = None
+        self._orig_seq_lengths: Optional[tuple[int, int]] = None
+
+    def _apply(self, module: nn.Module, mesh: DeviceMesh) -> nn.Module:
```
tianyu-l, line 1229: why not using the `distribute_module` API like other `ParallelStyle`?
https://github.com/pytorch/pytorch/blob/main/torch/distributed/tensor/parallel/style.py#L158

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1178,6 +1132,138 @@ def _context_parallel_buffers(
...
+        (e.g. DispatchKey::Autograd, DispatchKey::CompositeImplicitAutograd).  If your
+        operator is not differentiable, or to squash this warning and use the previous
+        behavior, please register torch::CppFunction::makeFallthrough() to DispatchKey::Autograd.
+        """
+        global_key = ft_c.all_gather_tensor_autograd(key, self.seq_dim, mesh)
+        global_value = ft_c.all_gather_tensor_autograd(value, self.seq_dim, mesh)
+        args_list[1] = global_key
+        args_list[2] = global_value
+
+        # shape rewrite: because torch.nn.flex_attention() checks
```
tianyu-l: hmm this sounds hacky.
@drisspg can we provide some other workaround?

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1178,6 +1132,138 @@ def _context_parallel_buffers(
...
+# Experimental APIs
+#####################
+
+
+class _ContextParallel(ParallelStyle):
+    class AttentionType(Enum):
+        FLEX = "flex_attention"
+        SDPA = "scaled_dot_product_attention"
+
+    def __init__(self, seq_dim: int, attention_type: AttentionType) -> None:
```
tianyu-l, line 1224: This may not be a priority, but I think now we can provide input/output layouts annotations, so that inputs/outputs can be plain Tensor or DTensors with placements not necessarily `Shard(seq_dim)`.
https://github.com/pytorch/pytorch/blob/main/torch/distributed/tensor/parallel/style.py#L85-L87
https://github.com/pytorch/pytorch/blob/main/torch/distributed/tensor/parallel/style.py#L371

tianyu-l (commented):
> We need two places to parallelize the CP, as a context_parallel context manager is still required for splitting the inputs.

I'm not sure how necessary this is. E.g. since we are already using `parallelize_module`, can we create another `ParallelStyle` that shards the input (with / without load balancing) and register it for the entire `Transformer`? (This is equivalent to calling a function before `model(inputs)` in `train.py`.)

torch/distributed/tensor/experimental/_attention.py
fegin: @XilunWu and I discussed this when we decided to do module hooks. That's part of the support of optional per-layer CP. However, while there were some discussions before, I am not sure if we have a real case in practice and the implication of performance is also unclear as we may not be able to hide the communication. But yes, that is the benefit of using module hook compared to TorchFunction.

torch/distributed/tensor/experimental/_attention.py
fegin: BlockMask rewrite is required as we have to convert a global BlockMask to local BlockMask. But I don't remember why we cannot just do it once in the beginning.  I saw the rewrite also happen in `create_cp_block_mask`. cc., @XilunWu

torch/distributed/tensor/experimental/_attention.py
fegin: Because we need `with_kwargs`, which `distribute_module` doesn't provide. Similar to `PrepareModuleInput` and `PrepareModuleOutput`, we just call register hook by ourselves.

torch/distributed/tensor/experimental/_attention.py
drisspg: This check is fair imo I think its really an ordering problem because these checks happend before the hop. An alternative would be to move these checks inside the actual HOP where dtensor would have a chance to convert to local 
https://github.com/pytorch/pytorch/blob/7a0f93344e2c851b9bcf2b9c3225a323d48fde26/torch/nn/attention/flex_attention.py#L1323-L1344

fegin: This code piece is not needed in the ModuleWrapper method. I temporality comment it out and will completely delete once we are sure that we are not going to support TorchFunctionMode anymore.

test/distributed/tensor/test_attention.py
```diff
@@ -86,12 +107,89 @@ def test_ring_attention_sdpa(self) -> None:
...
+        if dispatch_mode == _DispatchMode.MODULE_WRAPPER:
+            cp_plan = _ContextParallel(
+                seq_dim=seq_dim,
+                attention_type=_ContextParallel.AttentionType.SDPA,
+            )
+            attention = SDPAWrapper(compiled=compiled, backend=backend)
+            attention = parallelize_module(attention, mesh, cp_plan)
+        else:
+            # This won't work for MONKEY_PATCH.
+            attention = F.scaled_dot_product_attention
```
XilunWu: what does it mean by "This won't work"?

test/distributed/tensor/test_attention.py
```diff
@@ -86,12 +107,89 @@ def test_ring_attention_sdpa(self) -> None:
...
+
+        # Theoretically, context_parallel() should not be used to shard
+        # parameters because when require_grad is True, resize_ is not
+        # allowed. But requires_grad of cp_q, cp_k, and cp_v are False
+        # now. So we can just use context_parallel() to shard q, k, v.
+        # In reality, context_paralle() should be used to shard the input.
```
XilunWu: typo: context_parallel()

also "context_parallel() should be used to shard the input" is a bit confusing at first glance because the "input" to Attention function/module is the q,k,v tensor. I guess here it meant "model input" or "input embedding".

test/distributed/tensor/test_attention.py
```diff
@@ -86,12 +107,89 @@ def test_ring_attention_sdpa(self) -> None:
...
+            # If we use SDPAWrapper directly, the monkey patching dispatch mode
+            # does not function correctly. To ensure proper behavior,
+            # F.scaled_dot_product_attention must be referenced within the
+            # context_parallel() scope.
+            if dispatch_mode == _DispatchMode.MONKEY_PATCH:
+                attention = F.scaled_dot_product_attention
+                if compiled:
+                    attention = torch.compile(
+                        attention, fullgraph=True, backend="aot_eager"
+                    )
```
XilunWu, line 165: IIUC this is expected because Module wrapper approach applies to modules while the monkey patch is globally replacing SDPA (i.e. users won't be able to use the original SDPA function). Therefore these 2 things are conflicting. 

Likely we should deprecate the monkey patch approach but the deprecation time is TBD.

test/distributed/tensor/test_attention.py
```diff
@@ -354,46 +410,61 @@ def doc_mask_mod(b, h, q_idx, kv_idx):
     return doc_mask_mod
 
 
+class FlexAttentionWrapper(torch.nn.Module):
+    _flex_attn: ClassVar[Callable] = torch.compile(flex_attention)
```
XilunWu, line 438: probably add "fullgraph=True" to torch.compile

XilunWu (approved):
LGTM

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1190,6 +1072,191 @@ def _context_parallel_buffers(
...
+        new_args = tuple(all_args[0 : len(args)])
+        new_kwargs = dict(zip(kwargs.keys(), all_args[len(args) :]))
+        return new_args, new_kwargs
+
+    def sdpa_output_fn(
+        self, module: Optional[nn.Module], inputs: Any, outputs: Any, mesh: DeviceMesh
+    ) -> Any:
+        new_outputs = []
+        for output in [outputs] if isinstance(outputs, torch.Tensor) else outputs:
+            output = output.to_local() if isinstance(output, DTensor) else output
```
tianyu-l, line 1311: It seems the input can be DTensor but output has to be plain tensor. Shall we provide an option for users? E.g. this could be useful for the full DTensor mode I was talking about.

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1344,89 +1411,3 @@ def set_rotate_method(rotate_method: str) -> None:
             "Context Parallel does not support "
             f"using {rotate_method} for kv shards rotation"
         )
```
tianyu-l, line 1503: Shall we say somewhere that `set_rotate_method` is no op for FlexAttention CP?

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1190,6 +1072,191 @@ def _context_parallel_buffers(
...
+        module: Optional[nn.Module],
+        args: tuple[Any, ...],
+        kwargs: dict[str, Any],
+        mesh: DeviceMesh,
+    ) -> tuple[tuple[Any, ...], dict[str, Any]]:
+        placement = [Shard(self.seq_dim)]
+        all_args = []
+
+        for arg in itertools.chain(args, kwargs.values()):
+            if isinstance(arg, torch.Tensor) and not isinstance(arg, DTensor):
```
tianyu-l: If it is a DTensor we should assert the placement is correct.

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1004,124 +985,25 @@ def _enable_cp_dtensor_dispatcher() -> Generator[None, None, None]:
...
-                return func(*args, **kwargs)
-
-            if self._input_fn is not None:
-                args, kwargs = self._input_fn(self._device_mesh, *args, **kwargs)
-            output = func(*args, **kwargs)
-            if self._output_fn is not None:
-                output = self._output_fn(self._device_mesh, output)
-            return output
+    sdpa_cp = _ContextParallel(
+        seq_dim=seq_dim,
```
tianyu-l, line 997: It seems user can't set it, it has to be 2. Is this expected?
https://github.com/pytorch/pytorch/pull/162542/files#diff-60275c13ec2d1e34f06f4b2ece482548b1ba911c4e321ece76448461f0681d07R1336

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1190,6 +1072,191 @@ def _context_parallel_buffers(
...
+    H: int,
+    Q_LEN: int,
+    KV_LEN: int,
+    device_mesh: DeviceMesh,
+) -> BlockMask:
+    """
+    This API creates a special BlockMask for Context Parallel FlexAttention:
+    1. This BlockMask is masking on the attention of Q shard and KV global views, by
+    mapping the local q_idx to the global q_idx before sending to mask_mod.
+    2. The kv_seq_length (i.e. seq_lengths[1]) of this blockMask is tailored to match
```
tianyu-l: can this be removed?

test/distributed/tensor/test_attention.py
```diff
@@ -412,67 +483,60 @@ def _test_cp_flex_attention(
         # NOTE: we do not test load balance here
         _cp_options.enable_load_balance = False
```
tianyu-l: Do you think we should raise if this is True for FlexAttention? My worry is that the inputs are still permuted by block mask is not.

test/distributed/tensor/test_attention.py
fegin: > IIUC this is expected because Module wrapper approach applies to modules while the monkey patch is globally replacing SDPA (i.e. users won't be able to use the original SDPA function). Therefore these 2 things are conflicting.



I don't think it is expected. I would assume that the module that calls `F.scaled_dot_product_attention` should call into the patched one. But it is a very subtle issue that when `F.scaled_dot_product_attention` is referred and at that time if the patch is already applied. So I think this subtle timing issue can be very bad for debugging.

I agree with you that we should deprecate monkey patch as some loss NaN may be caused by the monkey patch.

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1190,6 +1072,191 @@ def _context_parallel_buffers(
...
+    from torch.nn.attention.flex_attention import _DEFAULT_SPARSE_BLOCK_SIZE
+
+    compiled_create_block_mask = torch.compile(
+        create_block_mask, dynamic=False, fullgraph=True
+    )
+
+    def _rewrite_mask_mod(
+        mask_mod: _mask_mod_signature,
+        rank: int,
+        world_size: int,
```
tianyu-l: `world_size` is not used

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1190,6 +1072,218 @@ def _context_parallel_buffers(
...
+    Raises:
+        NotImplementedError: If Q_LEN is not divisible by (CP world size * BLOCK_SIZE).
+
+    Warning:
+        Currently requires Q_LEN to be divisible by CP mesh world size * BLOCK_SIZE
+        (BLOCK_SIZE defaults to 128). This constraint exists because the BlockMask
+        must handle both padding and offsets correctly. For example, if Q_LEN is 384,
+        CP world size is 2, and BLOCK_SIZE is 128, the local Q_LEN would be 192. In
+        such cases, both rank0 and rank1 would have paddings in their local BlockMasks.
+        Support for padding in this scenario is planned for future work.
```
tianyu-l, line 1124: Two questions about the requirement
```
Q_LEN % (device_mesh.size() * _DEFAULT_SPARSE_BLOCK_SIZE) != 0
```
1.
> For example, if Q_LEN is 384, CP world size is 2, and BLOCK_SIZE is 128, the local Q_LEN would be 192. In such cases, both rank0 and rank1 would have paddings in their local BlockMasks.

Could you explain why this is problematic?

2. Would it be OK as long as the `local seq len` can divide `block_size`?

fegin: 1. Because we calculate the local block counts and then use it to infer the global block idx. It is not needed for now but will needed when load balancing logic is added, which is block based.

2. local_seq_len can be divided by block_size is equivalent to the current constraint.

torch/distributed/tensor/experimental/_attention.py
fegin: At this moment, this is hard coded, but we should change in the next few refactor PRs.

test/distributed/tensor/test_attention.py
```diff
@@ -392,15 +461,24 @@ def _test_cp_flex_attention(
...
         lb: Optional[_LoadBalancer] = None,
-        atol: float = 1e-4,
-        rtol: float = 1,
+        atol: float = 1e-6,
+        rtol: float = 1e-2,
     ) -> None:
+        # TODO: Reverify atol and rtol after
+        # https://github.com/pytorch/pytorch/pull/163185 is landed. The accuracy
+        # issue happens on the gradients.
+        torch.use_deterministic_algorithms(True)
```
jeffdaily, line 470: Please don't do things like this.  It might subtly break other tests since this globally changes the deterministic setting.  Please only set deterministic on the tests you need it for, and use a context manager so that it resets back to prior state.

PR #162545, opened 2025-09-09
[c10d] new_group works without world group created

kwen2501:
* #162571
* #162549
* #163058
* #162770
* __->__ #162545
* #162529

As titled.
Unblocking DeviceMesh use case where DeviceMesh may only want to create sub-groups along dimensions, without creating the world group.

### DIFF
--- a/torch/distributed/distributed_c10d.py
+++ b/torch/distributed/distributed_c10d.py
@@ -1755,6 +1755,12 @@ def init_process_group(
     # Initialize distributed environment if user didn't call it
     # (backward compatible purpose)
     if _World.maybe_get() is None:
+        # "fake" backend uses specific store
+        if backend == "fake":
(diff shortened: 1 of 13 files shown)


torch/distributed/distributed_c10d.py
```diff
@@ -5398,23 +5398,36 @@ def _new_group_with_tag(
...
     ``torch.distributed._functional_collectives`` for reference on how to use it.
     """
-    global _world
+    world = _World.get()
 
-    default_pg = _get_default_group()
+    try:
+        default_pg = _get_default_group()
+    except ValueError:
+        default_pg = None
```
ezyang, line 5409: It's too bad there isn't a version of the function that returns None instead

ezyang (approved):

PR #162550, opened 2025-09-10
[torchao][pt2e] Make prepare and convert faster by caching

navsud:
Summary: D79674759 tried to fix the expensive prepare and convert steps, as `assert_and_get_unique_device` was called multiple times. This change fixes that issue by using `functools.cache` decorator.

Test Plan:
Verified on llm export to QNN.
LLM Quantization prepare time of ~20min reduced to ~3min.

Rollback Plan:

Differential Revision: D82073679


andrewor14 (approved):
Looks great, thanks!

PR #162560, opened 2025-09-10
[DTensor] add op support for aten.unbind.int

tianyu-l:
* __->__ #162560

As titled.

It seems unbind returns views of the original tensor. E.g. see https://stackoverflow.com/questions/78910951/does-unbind-return-the-views-of-tensors-in-pytorch

So we error out when `shard_dim == unbind_dim`. This is similar to why we error out in view ops.
https://github.com/pytorch/pytorch/blob/main/torch/distributed/tensor/_ops/_view_ops.py#L544-L546

This PR also refactors some other tensor ops code, by creating two utils function `shift_shard_dims_after_insert`, `shift_shard_dims_after_remove`.

### DIFF
--- a/torch/distributed/tensor/_ops/utils.py
+++ b/torch/distributed/tensor/_ops/utils.py
@@ -370,3 +370,27 @@ def expand_to_full_mesh_op_strategy(
         )
         all_strategies.append(strategy)
     return OpStrategy(all_strategies)
+
+
+def shift_shard_dims_after_insert(
+    placements: Sequence[Placement], insert_dim: int = 0
+) -> Sequence[Placement]:
+    normalized_placements: list[Placement] = []
+    for placement in placements:
+        if isinstance(placement, Shard) and placement.dim >= insert_dim:
+            normalized_placements.append(Shard(placement.dim + 1))
+        else:
+            normalized_placements.append(placement)
+    return normalized_placements
+
+
+def shift_shard_dims_after_remove(
+    placements: Sequence[Placement], remove_dim: int = 0
+) -> Sequence[Placement]:
+    normalized_placements: list[Placement] = []
+    for placement in placements:
+        if isinstance(placement, Shard) and placement.dim > remove_dim:
+            normalized_placements.append(Shard(placement.dim - 1))
+        else:
+            normalized_placements.append(placement)
+    return normalized_placements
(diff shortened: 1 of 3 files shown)


torch/distributed/tensor/_ops/utils.py
```diff
@@ -370,3 +370,27 @@ def expand_to_full_mesh_op_strategy(
...
+
+def shift_shard_dims_after_remove(
+    placements: Sequence[Placement], remove_dim: int = 0
+) -> Sequence[Placement]:
+    normalized_placements: list[Placement] = []
+    for placement in placements:
+        if isinstance(placement, Shard) and placement.dim > remove_dim:
+            normalized_placements.append(Shard(placement.dim - 1))
+        else:
+            normalized_placements.append(placement)
```
zpcore, line 395: Shall we do here:
```
elif placement.dim == remove_dim:
    normalized_placements.append(Replicate())
```
Any cases that the removed dim don't need replicate?

zpcore (approved):
I was checking why we are still unable to remove xfail('unbind'), and notice the test fail:
```
$ PYTORCH_OPINFO_SAMPLE_INPUT_INDEX=0 python test/distributed/tensor/test_dtensor_ops.py TestDTensorOpsCPU.test_dtensor_op_db_unbind_cpu_float32
```
```
 RuntimeError: Output 0 of UnbindBackward0 is a view and its base or another view of its base has been modified inplace. This view is the output of a function that returns multiple views. Such functions do not allow the output views to be modified inplace. You should replace the inplace operation by an out-of-place one.
 ```
This seems to be a tricky issue. I think we can merge this first and look back later.

zpcore (approved):
I think we are unable to get rid of `xfail("unbind")` in test_dtensor_ops.py due to the view issue. You can see we still have `xfail("view")` there. I agree raising the assertion in the strategy is good enough. Thanks for the fixing!

torch/distributed/tensor/_ops/utils.py
tianyu-l: I think currently the usage this API assumes the `remove_dim` is not `Shard`, but I'm not sure which one makes more sense for new use cases
1. raise exception when `remove_dim` is sharded
2. always replicate if `remove_dim` is sharded 

Maybe let's leave this option for the future.

PR #156812, opened 2025-06-25
Introduce a new API torch.accelerator.get_memory_info

guangyey:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162564
* __->__ #156812

# Motivation
`torch.cuda.mem_get_info` and `torch.xpu.mem_get_info` are widely used in other popular repos, such as
- https://github.com/sgl-project/sglang/blob/076313bd099ac1ee484ee77009eaae864eacf396/python/sglang/srt/utils.py#L378，
- https://github.com/huggingface/accelerate/blob/7ecc2d7f394fc0686062a18d46128a8bd97c7dad/src/accelerate/utils/modeling.py#L822,
- https://github.com/vllm-project/vllm/blob/7ba34b1241ada58f8212f350a8b17382cb412cf2/vllm/worker/worker.py#L150.
- 
This PR introduces a unified API `torch.accelerator.get_memory_info` to cover this scenario.

cc @albanD @EikanWang

PR #162564, opened 2025-09-10
[xpu][feature] Add XPU support on torch.accelerator.get_memory_info

guangyey:
* __->__ #162564
* #156812

# Motivation
Support XPU for `torch.accelerator.get_memory_info`.


guangyey:
@albanD May I know if any comments on this PR

test/test_accelerator.py
```diff
@@ -234,6 +234,12 @@ def test_memory_stats(self):
         self.assertEqual(torch.accelerator.max_memory_allocated(), prev_max_allocated)
         self.assertEqual(torch.accelerator.max_memory_reserved(), prev_max_reserved)
 
+    @unittest.skipIf(TEST_MPS, "MPS doesn't support torch.accelerator memory API!")
```
albanD, line 242: Could you please open an issue on github with the mps label for this one so the relevant people can know

torch/csrc/xpu/Module.cpp
```diff
@@ -386,23 +386,8 @@ static void bindGetDeviceProperties(PyObject* module) {
...
-        "You can file an issue at https://github.com/pytorch/pytorch/issues ",
-        "to help us prioritize its implementation.");
-    auto free = device.get_info<sycl::ext::intel::info::device::free_memory>();
-    return std::make_tuple(free, total);
-#else
-  TORCH_CHECK_NOT_IMPLEMENTED(
-      false,
-      "torch.xpu.mem_get_info requires PyTorch to be built with SYCL compiler version 2025.0.0 or newer.");
-#endif
+    py::gil_scoped_release no_gil;
```
albanD, line 389: Why release the GIL here, is getting that information expensive?
If so, we should build a caching system for it.

Releasing the GIL is not free, so for something cheap, this will actually be slower than keeping it.

albanD (approved):
SGTM

torch/csrc/xpu/Module.cpp
guangyey: Yes, retrieving free memory information is quite expensive. I tested it on a CPU with 112 physical ovres and a Data Center GPU, and it took over 140us. This value is real-time and cannot be cached. It must be queried from the runtime each time.

test/test_accelerator.py
guangyey: I open an issue here https://github.com/pytorch/pytorch/issues/167447

PR #162569, opened 2025-09-10
[FlexAttention][Easy] turn off TMA when cannot use it

BoyuanFeng:

### DIFF
--- a/torch/_inductor/kernel/flex/flex_attention.py
+++ b/torch/_inductor/kernel/flex/flex_attention.py
@@ -10,7 +10,6 @@ from typing import Any, Optional, Union
 import sympy
 
 import torch
-from torch._inductor.utils import can_use_tma
 from torch._inductor.virtualized import V
 
 from ...ir import ComputedBuffer, ExternKernel, FixedLayout, TensorBox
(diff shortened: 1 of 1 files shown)


torch/_inductor/kernel/flex/flex_attention.py
```diff
@@ -317,8 +317,8 @@ def flex_attention(
         # USE TMA = false by default
         cur_kernel_options.setdefault("USE_TMA", False)
 
-        if cur_kernel_options["USE_TMA"] and can_use_tma(query, key, value):
-            cur_kernel_options["USE_TMA"] = True
```
BoyuanFeng, line 321: this is a noop

drisspg: The original plan was to actually enable this flag by default which is why I think it was implemented as is. I think that @NikhilAPatel was planning to enable this work, but AFAIK found that it was not universally better and hard to come up w/ the correct herusitic for usage. SO we can remove

BoyuanFeng: yes removed

drisspg (approved):

PR #162549, opened 2025-09-09
[c10d] Remove init_process_group from DeviceMesh

kwen2501:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.12.0) (oldest at bottom):
* #162571
* __->__ #162549
* #163058
* #162770
* #162545
* #162529

As titled.
So that `DeviceMesh` creation will no longer implicitly create a global process group, which is time consuming and most-of-time not used.
`init_process_group` is replaced with `dist.init()` -- just for `get_rank()` and `get_world_size()` to work.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci

PR #162571, opened 2025-09-10
[c10d] Differentiate participate requirements between init_device_mesh and direct construction

kwen2501:
* __->__ #162571
* #162549
* #163058
* #162770
* #162545
* #162529

1. `init_device_mesh` is a collective call across World.
Reason: it supports implicit `dist.init()` which is a world-wide call.
```
# Valid
dist.init()
init_device_mesh(...)

# Valid
init_device_mesh(...)
```

2. `DeviceMesh` direct construction is more flexible, can be called on a subset of ranks. But, prior to `DeviceMesh` construction, *the* distributed environment must have been initialized. We don't want to take the risk to perform `dist.init()` implicitly here due to the flexibility of the constructor API. 
```
# Valid
dist.init()
dm = DeviceMesh(...)

# Invalid, will throw environment uninitialized error
dm = DeviceMesh(...)
```

### DIFF
--- a/torch/distributed/device_mesh.py
+++ b/torch/distributed/device_mesh.py
@@ -447,19 +447,18 @@ if True:  # just to temporarily avoid reindentation
             self._flatten_mesh_list = tuple(self.mesh.flatten().tolist())
             self._thread_id = None
 
-            # Optionally init distributed environment if user has not done so.
+            # Direct DeviceMesh construction requires the distributed
+            # environment to be initialized first because the mesh size may
+            # differ from the world size -- no implicit initialization here.
             try:
-                get_world_size()
+                world_size = get_world_size()
             except RuntimeError:
-                logger.debug(
-                    "Distributed environment is not initialized, initializing "
-                    "it during DeviceMesh creation with an assumption that "
-                    "the DeviceMesh has the same scope as the world."
+                warnings.warn(
+                    "DeviceMesh construction requires the distributed environment to be initialized first. "
                 )
-                torch.distributed.init()
+                raise
 
             # Check if the requested mesh is valid
-            world_size = get_world_size()
             if self.mesh.numel() > world_size:
                 raise RuntimeError(
                     f"Mesh should not be bigger than world size {world_size}, but found {self.mesh.numel()} ranks!"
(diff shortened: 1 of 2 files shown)


torch/distributed/device_mesh.py
```diff
@@ -447,19 +447,18 @@ def __init__(
...
             try:
-                get_world_size()
+                world_size = get_world_size()
             except RuntimeError:
-                logger.debug(
-                    "Distributed environment is not initialized, initializing "
-                    "it during DeviceMesh creation with an assumption that "
-                    "the DeviceMesh has the same scope as the world."
+                warnings.warn(
+                    "DeviceMesh construction requires the distributed environment to be initialized first. "
```
Skylion007: Please include stack level kwargs

torch/distributed/device_mesh.py
```diff
@@ -1124,10 +1123,32 @@ def init_device_mesh(
                 "If you maintained a 'torch.device' object, it's recommended to pass in 'device.type'.",
             )
 
+        # Optionally init distributed environment if user has not done so.
+        try:
+            get_world_size()
+        except RuntimeError:
+            logger.debug(
+                "Distributed environment is not initialized, initializing "
+                "it during DeviceMesh creation with an assumption that "
```
Skylion007: Add exc_info=True to the debug call

torch/distributed/device_mesh.py
```diff
@@ -447,19 +447,18 @@ def __init__(
...
             try:
-                get_world_size()
+                world_size = get_world_size()
             except RuntimeError:
-                logger.debug(
-                    "Distributed environment is not initialized, initializing "
-                    "it during DeviceMesh creation with an assumption that "
-                    "the DeviceMesh has the same scope as the world."
+                logger.error(
+                    "DeviceMesh construction requires the distributed environment to be initialized first. "
```
ezyang: We can, like, directly tell the user what they're supposed to call right? Maybe it's good to explain why there might be some ambiguity so they can decide if there's a different problem?

kwen2501: Fixed.

Issue #153265, opened 2025-05-09
[RFC][API-Unstable] Support 3rd party SYCL kernels with CPP Extension API


dvrogozh:
### Background

This RFC is submitted to formally describe and track enabling of SYCL support in CPP Ectension API. The initial enabling of this feature was done in response to the https://github.com/pytorch/pytorch/issues/132944. This RFC tries to follow a guideline process suggested in https://github.com/pytorch/pytorch/issues/152134 for Pytorch 2.8.

### Motivation

Pytorch defines a range of standard operators to support variety of deep learning models. However, with the rapid development of the technology new operators and optimization techniques emerge. Their adoption requires implementations to mature before being accepted in a standard pytorch distribution. To both equip developers with the way to implement new operators capable to work within Pytorch ecosystem and mitigate a delay to finalize the implementation and get it accepted in a Pytorch code base, Pytorch provides CPP Extension API. At the moment Pytorch defines that API to work with C++, CUDA and Metal sources.

This RFC suggests to extend Pytorch CPP Extension API to allow building new operators for Intel GPU platforms by supporting SYCL standard and compilers. [SYCL](https://www.khronos.org/sycl/) is an open standard developed by the Khronos Group that allows developers to program heterogeneous architectures in standard C++. Intel GPU software stack supports this standard and provides DPC++ compiler to build SYCL code.

### Plan

Here is a check list of items to complete to add SYCL support to Pytorch CPP Extension API for Intel GPUs:

#### Support for Linux:

[API-Unstable]
- [x] Add SYCL support to `torch.utils.cpp_extension.load()` API on Linux
  * https://github.com/pytorch/pytorch/pull/132945
- [x] Add SYCL support to `torch.utils.cpp_extension.load_inline()` API on Linux
  * https://github.com/pytorch/pytorch/pull/132945
- [x] Support new `class torch.utils.cpp_extension.SyclExtension` API on Linux
  * https://github.com/pytorch/pytorch/pull/132945
- [x] Docstring documentation
  * https://github.com/pytorch/pytorch/pull/132945
- [x] Add CI tests for SYCL support in CPP Extensions
  * https://github.com/pytorch/pytorch/pull/132945
 
[API-Stable]
- [ ] Create (or extend existing) tutorial
  * https://github.com/pytorch/tutorials/pull/3391
  * https://github.com/pytorch/extension-cpp/pull/114
[...]

PR #162579, opened 2025-09-10
[xpu][feature] enable Sycl CPP extension on Windows

tbohutyn:
PR for enabling #153265 on Windows

### DIFF
--- a/test/test_cpp_extensions_jit.py
+++ b/test/test_cpp_extensions_jit.py
@@ -141,6 +141,7 @@ class TestCppExtensionJIT(common.TestCase):
                 sources=[sycl_file],
                 extra_sycl_cflags=extra_sycl_cflags,
                 verbose=True,
+                build_directory=temp_dir,
             )
 
             x = torch.zeros(100, device="xpu", dtype=torch.float32)
@@ -153,7 +154,12 @@ class TestCppExtensionJIT(common.TestCase):
             # 2 * sigmoid(0) = 2 * 0.5 = 1
             self.assertEqual(z, torch.ones_like(z))
         finally:
-            shutil.rmtree(temp_dir)
+            if IS_WINDOWS:
+                # rmtree returns permission error: [WinError 5] Access is denied
+                # on Windows, this is a workaround
+                subprocess.run(["rm", "-rf", temp_dir], stdout=subprocess.PIPE)
+            else:
+                shutil.rmtree(temp_dir)
 
     @unittest.skipIf(not (TEST_XPU), "XPU not found")
     def test_jit_xpu_extension(self):
--- a/test/test_cpp_extensions_jit.py
+++ b/test/test_cpp_extensions_jit.py
@@ -155,7 +155,9 @@ class TestCppExtensionJIT(common.TestCase):
             # 2 * sigmoid(0) = 2 * 0.5 = 1
             self.assertEqual(z, torch.ones_like(z))
         finally:
-            shutil.rmtree(temp_dir)
+            # Workaround on Windows: can't remove dll while it's loaded
+            if not IS_WINDOWS:
+                shutil.rmtree(temp_dir)
 
     @unittest.skipIf(not (TEST_XPU), "XPU not found")
     def test_jit_xpu_extension(self):
--- a/test/test_cpp_extensions_jit.py
+++ b/test/test_cpp_extensions_jit.py
@@ -157,7 +157,7 @@ class TestCppExtensionJIT(common.TestCase):
             if IS_WINDOWS:
                 # rmtree returns permission error: [WinError 5] Access is denied
                 # on Windows, this is a workaround
-                subprocess.run(["rm", "-rf", temp_dir], stdout=subprocess.PIPE)
+                subprocess.run(["rd", "/s", "/q", temp_dir], stdout=subprocess.PIPE)
             else:
                 shutil.rmtree(temp_dir)
 
--- a/test/test_cpp_extensions_jit.py
+++ b/test/test_cpp_extensions_jit.py
@@ -141,8 +141,6 @@ class TestCppExtensionJIT(common.TestCase):
                 sources=[sycl_file],
                 extra_sycl_cflags=extra_sycl_cflags,
                 verbose=True,
-                keep_intermediates=True,
-                build_directory=temp_dir,
             )
 
             x = torch.zeros(100, device="xpu", dtype=torch.float32)
@@ -155,9 +153,7 @@ class TestCppExtensionJIT(common.TestCase):
             # 2 * sigmoid(0) = 2 * 0.5 = 1
             self.assertEqual(z, torch.ones_like(z))
         finally:
-            # Workaround on Windows: can't remove dll while it's loaded
-            if not IS_WINDOWS:
-                shutil.rmtree(temp_dir)
+            shutil.rmtree(temp_dir)
 
     @unittest.skipIf(not (TEST_XPU), "XPU not found")
     def test_jit_xpu_extension(self):
--- a/torch/utils/cpp_extension.py
+++ b/torch/utils/cpp_extension.py
@@ -577,10 +577,17 @@ def _append_sycl_std_if_no_std_present(cflags) -> None:
 
 def _wrap_sycl_host_flags(cflags):
     host_cxx = get_cxx_compiler()
-    host_cflags = [
-        f'-fsycl-host-compiler={host_cxx}',
-        shlex.quote(f'-fsycl-host-compiler-options={cflags}'),
-    ]
+    if IS_WINDOWS:
+        host_cflags = [
+            "-fsycl",
+            f"-fsycl-host-compiler={host_cxx}",
+            f'-fsycl-host-compiler-options="/std:c++17 {cflags}"',
+        ]
+    else:
+        host_cflags = [
+            f"-fsycl-host-compiler={host_cxx}",
+            shlex.quote(f"-fsycl-host-compiler-options={cflags}"),
+        ]
     return host_cflags
 
 
@@ -1015,6 +1022,7 @@ class BuildExtension(build_ext):
             else:
                 common_cflags.extend(COMMON_MSVC_FLAGS)
             with_cuda = any(map(_is_cuda_file, sources))
+            with_sycl = any(map(_is_sycl_file, sources))
 
             # extra_postargs can be either:
             # - a dict mapping cxx/nvcc to extra flags
@@ -1058,6 +1066,40 @@ class BuildExtension(build_ext):
             else:
                 cuda_dlink_post_cflags = None
 
+            sycl_cflags = None
+            sycl_post_cflags = None
+            sycl_dlink_post_cflags = None
+            if with_sycl:
+                sycl_cflags = common_cflags + pp_opts + _COMMON_SYCL_FLAGS
+                if isinstance(extra_postargs, dict):
+                    sycl_post_cflags = extra_postargs['sycl']
+                else:
+                    sycl_post_cflags = list(extra_postargs)
+                _append_sycl_targets_if_missing(sycl_post_cflags)
+                append_std17_if_no_std_present(sycl_cflags)
+                _append_sycl_std_if_no_std_present(sycl_cflags)
+                host_cflags = common_cflags + pp_opts + post_cflags
+                append_std17_if_no_std_present(host_cflags)
+
+                sycl_cflags = _nt_quote_args(sycl_cflags)
+                host_cflags = _nt_quote_args(host_cflags)
+
+                for i, flag in enumerate(host_cflags):
+                    if "-I" in flag:
+                        flag = flag.replace("\\", "\\\\").replace("-I", "/I")
+                    else:
+                        flag = flag.replace("-D", "/D")
+                    flag = flag.replace('"', '\\"')
+
+                    host_cflags[i] = flag
+
+                host_cflags = ' '.join(host_cflags)
+                sycl_cflags += _wrap_sycl_host_flags(host_cflags)
+                sycl_dlink_post_cflags = _SYCL_DLINK_FLAGS.copy()
+                sycl_dlink_post_cflags += _get_sycl_device_flags(sycl_post_cflags)
+                sycl_post_cflags = _nt_quote_args(sycl_post_cflags)
+
+
             _write_ninja_file_and_compile_objects(
                 sources=sources,
                 objects=objects,
@@ -1066,13 +1108,13 @@ class BuildExtension(build_ext):
                 cuda_cflags=cuda_cflags,
                 cuda_post_cflags=cuda_post_cflags,
                 cuda_dlink_post_cflags=cuda_dlink_post_cflags,
-                sycl_cflags=None,
-                sycl_post_cflags=None,
-                sycl_dlink_post_cflags=None,
+                sycl_cflags=sycl_cflags,
+                sycl_post_cflags=sycl_post_cflags,
+                sycl_dlink_post_cflags=sycl_dlink_post_cflags,
                 build_directory=output_dir,
                 verbose=True,
                 with_cuda=with_cuda,
-                with_sycl=False)
+                with_sycl=with_sycl)
 
             # Return *all* object filenames, not just the ones we just built.
             return objects
@@ -1492,6 +1534,7 @@ def SyclExtension(name, sources, *args, **kwargs):
     libraries.append("c10_xpu")
     libraries.append("torch")
     libraries.append("torch_cpu")
+    libraries.append("sycl")
     if not kwargs.get('py_limited_api', False):
         # torch_python uses more than the python limited api
         libraries.append("torch_python")
@@ -2273,6 +2316,7 @@ def _write_ninja_file_and_build_library(
     extra_ldflags = _prepare_ldflags(
         extra_ldflags or [],
         with_cuda,
+        with_sycl,
         verbose,
         is_standalone)
     build_file_path = os.path.join(build_directory, 'build.ninja')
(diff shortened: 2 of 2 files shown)


torch/utils/cpp_extension.py
```diff
@@ -2721,8 +2771,17 @@ def _write_ninja_file_to_build_library(path,
         host_cflags = cflags
         # escaping quoted arguments to pass them thru SYCL compiler
         icpx_version = _get_icpx_version()
-        if int(icpx_version) < 20250200:
+        if IS_WINDOWS:
```
dvrogozh: Which oneAPI version have you used on Windows?

test/test_cpp_extensions_jit.py
```diff
@@ -155,7 +155,9 @@ def _test_jit_xpu_extension(self, extra_sycl_cflags):
             # 2 * sigmoid(0) = 2 * 0.5 = 1
             self.assertEqual(z, torch.ones_like(z))
         finally:
-            shutil.rmtree(temp_dir)
+            # Workaround on Windows: can't remove dll while it's loaded
+            if not IS_WINDOWS:
+                shutil.rmtree(temp_dir)
```
EikanWang: As you mentioned, this is a workaround. You can run the script through python subprocess and then try to remove the temporary directory.

torch/utils/cpp_extension.py
```diff
@@ -1005,6 +1012,7 @@ def win_wrap_ninja_compile(sources,
             else:
                 common_cflags.extend(COMMON_MSVC_FLAGS)
             with_cuda = any(map(_is_cuda_file, sources))
+            with_sycl = any(map(_is_sycl_file, sources))
```
EikanWang, line 1043:
```suggestion
            with_sycl = any(map(_is_sycl_file, sources))
            assert not (with_sycl and with_cuda)
```

torch/utils/cpp_extension.py
```diff
@@ -1048,6 +1056,40 @@ def win_wrap_ninja_compile(sources,
...
+                    flag = flag.replace('"', '\\"')
+
+                    host_cflags[i] = flag
+
+                host_cflags = ' '.join(host_cflags)
+                sycl_cflags += _wrap_sycl_host_flags(host_cflags)
+                sycl_dlink_post_cflags = _SYCL_DLINK_FLAGS.copy()
+                sycl_dlink_post_cflags += _get_sycl_device_flags(sycl_post_cflags)
+                sycl_post_cflags = _nt_quote_args(sycl_post_cflags)
+
```
EikanWang, line 1110: A dedicated function is preferred.

torch/utils/cpp_extension.py
```diff
@@ -573,10 +573,17 @@ def _append_sycl_std_if_no_std_present(cflags):
 
 def _wrap_sycl_host_flags(cflags):
     host_cxx = get_cxx_compiler()
-    host_cflags = [
-        f'-fsycl-host-compiler={host_cxx}',
-        shlex.quote(f'-fsycl-host-compiler-options={cflags}'),
-    ]
+    if IS_WINDOWS:
+        host_cflags = [
+            "-fsycl",
```
EikanWang: I'm curious why Windows build requires `-fsycl`, while Linux does not.

torch/utils/cpp_extension.py
```diff
@@ -1048,6 +1056,40 @@ def win_wrap_ninja_compile(sources,
...
+                append_std17_if_no_std_present(sycl_cflags)
+                _append_sycl_std_if_no_std_present(sycl_cflags)
+                host_cflags = common_cflags + pp_opts + post_cflags
+                append_std17_if_no_std_present(host_cflags)
+
+                sycl_cflags = _nt_quote_args(sycl_cflags)
+                host_cflags = _nt_quote_args(host_cflags)
+
+                for i, flag in enumerate(host_cflags):
+                    if "-I" in flag:
```
guangyey:
```suggestion
                    if flag.startswith("-I"):
```
Is this safer.

torch/utils/cpp_extension.py
```diff
@@ -2721,8 +2771,17 @@ def _write_ninja_file_to_build_library(path,
         host_cflags = cflags
         # escaping quoted arguments to pass them thru SYCL compiler
         icpx_version = _get_icpx_version()
-        if int(icpx_version) < 20250200:
+        if IS_WINDOWS:
+            for i, flag in enumerate(host_cflags):
```
guangyey: Maybe we could define a utility function to make it reusable.

torch/utils/cpp_extension.py
```diff
@@ -2305,7 +2349,7 @@ def verify_ninja_availability():
         raise RuntimeError("Ninja is required to load C++ extensions (pip install ninja to get it)")
 
 
-def _prepare_ldflags(extra_ldflags, with_cuda, verbose, is_standalone):
+def _prepare_ldflags(extra_ldflags, with_cuda, with_sycl, verbose, is_standalone):
```
EikanWang, line 2384: Could you help generalize it a little bit by handling `with_cuda`  and `with_sycl` with different functions accordingly?

torch/utils/cpp_extension.py
tbohutyn: I tested on both 2025.1 and 2025.2

torch/utils/cpp_extension.py
tbohutyn: moved to _wrap_sycl_host_flags

torch/utils/cpp_extension.py
tbohutyn: I followed how cuda is implemented and how it is in unix_wrap_ninja_compile

torch/utils/cpp_extension.py
tbohutyn: removed

test/test_cpp_extensions_jit.py
dvrogozh: @tbohutyn : please, align with what's already being done for CUDA if possible:
https://github.com/pytorch/pytorch/blob/167ad09be5af5c52666759412a3804068c6955d1/test/test_cpp_extensions_jit.py#L295-L300

torch/utils/cpp_extension.py
```diff
@@ -571,13 +571,31 @@ def _append_sycl_std_if_no_std_present(cflags):
...
+            else:
+                flag = flag.replace("-D", "/D")
+            flag = flag.replace('"', '\\"')
+            cflags.append(flag)
+        cflags = ' '.join(cflags)
+
+        external_include = _join_sycl_home("include").replace("\\", "\\\\")
+        wrapped_host_cflags = [
+            f"-fsycl-host-compiler={host_cxx}",
+            f'-fsycl-host-compiler-options="/std:c++17 \\"/external:I{external_include}\\" /external:W0 {cflags}"',
```
dvrogozh: > `\"/external:I{external_include}\\" /external:W0`

These flags require explanation. Please, add the one in the in-source comment. I believe both of these flags are needed to suppress deprecated warnings coming from SYCL header files. That's a DPC++ compiler level issue and we expect the fix in the future version. When the fix will be available, this place here will need to be updated and these flags dropped conditionally on the compiler version. Please, explain this in the comment. Also, why `/external:W0` is needed? I thought `/external:I` should be enough.... I would suggest something like:
```
# Some versions of DPC++ compiler pass paths to SYCL headers as user include paths (`-I`) rather
# than system paths (`-isystem`). This makes host compiler to report warnings encountered in the
# SYCL headers, such as deprecated warnings, even if warmed API is not actually used in the program.
# We expect that this issue will be addressed in the later version of DPC++ compiler. To workaround the
# issue now we wrap paths to SYCL headers in `/external:I`. Warning free compilation is especially important
# for Windows build as `/sdl` compilation flag assumes that and we will fail compilation otherwise.
```

torch/utils/cpp_extension.py
```diff
@@ -571,13 +571,31 @@ def _append_sycl_std_if_no_std_present(cflags):
...
+                flag = flag.replace("\\", "\\\\").replace("-I", "/I")
+            else:
+                flag = flag.replace("-D", "/D")
+            flag = flag.replace('"', '\\"')
+            cflags.append(flag)
+        cflags = ' '.join(cflags)
+
+        external_include = _join_sycl_home("include").replace("\\", "\\\\")
+        wrapped_host_cflags = [
+            f"-fsycl-host-compiler={host_cxx}",
```
dvrogozh, line 600: The `/std:c++17` which you add on the next line is actually handled by `append_std17_if_no_std_present(host_cflags)` which you also call. Can you drop explicit `/std:c++17` from the line below?

torch/utils/cpp_extension.py
```diff
@@ -571,13 +571,31 @@ def _append_sycl_std_if_no_std_present(cflags):
         cflags.append('-sycl-std=2020')
 
 
-def _wrap_sycl_host_flags(cflags):
+def _wrap_sycl_host_flags(host_cflags):
```
dvrogozh: To be honest I don't much care, but what was the reason to rename `cflags` to `host_cflags`? To me that's a user preference naming change and I would stick with the existing naming to minimize the code difference and make it easier to revie.

torch/utils/cpp_extension.py
```diff
@@ -2742,7 +2797,7 @@ def _write_ninja_file_to_build_library(path,
         icpx_version = _get_icpx_version()
         if int(icpx_version) < 20250200:
             host_cflags = [item.replace('\\"', '\\\\"') for item in host_cflags]
-        host_cflags = ' '.join(host_cflags)
+
```
dvrogozh, line 2824: This got moved to `_wrap_sycl_host_flags()`. Ok...

test/test_cpp_extensions_jit.py
```diff
@@ -141,8 +141,6 @@ def _test_jit_xpu_extension(self, extra_sycl_cflags):
                 sources=[sycl_file],
                 extra_sycl_cflags=extra_sycl_cflags,
                 verbose=True,
-                keep_intermediates=True,
-                build_directory=temp_dir,
```
tbohutyn, line 145: removed keep_intermediates and build_directory so it's handled by pytest tearDown:
https://github.com/pytorch/pytorch/blob/2226143ebd1316541d2136bfa324be93d1729a7a/test/test_cpp_extensions_jit.py#L65-L67

dvrogozh: I think `keep_intermediates=True` is a debug leftover, it should be fine to change it to `=False` - that's we value used across other cuda/xpu tests.

However, I have a concern to drop `build_directory=temp_dir`. In one of the tests, this helper function is called in a loop to give different cases a try. Working in separate `build_directory` assures that no object will accidentally overlap from the previous test case. @tbohutyn : are you dropping `build_directory=temp_dir` because otherwise SYCL test on Windows fails or you drop that to propose test simplification? In case of the latter I suggest to avoid this change as it will make test less robust. CUDA test is doing the same by the way:
https://github.com/pytorch/pytorch/blob/c332d5818417c315f304aad9d8dd4e0d38371b59/test/test_cpp_extensions_jit.py#L276

tbohutyn: If we remove temp_dir, the built module will still remain, just in the default directory which will be cleaned up by the pytest tear down. If you prefer I can change it back, just removed duplicating the logic of cleaning up after tests

dvrogozh: I would prefer if you change it back. Let's avoid unnecessary changes and target PR for the specific goal - enable sycl extensions on Windows. Other than that PR looks good to me.

dvrogozh (approved):
LGTM

EikanWang (approved):

test/test_cpp_extensions_jit.py
```diff
@@ -155,7 +154,12 @@ def _test_jit_xpu_extension(self, extra_sycl_cflags):
             # 2 * sigmoid(0) = 2 * 0.5 = 1
             self.assertEqual(z, torch.ones_like(z))
         finally:
-            shutil.rmtree(temp_dir)
+            if IS_WINDOWS:
+                # rmtree returns permission error: [WinError 5] Access is denied
+                # on Windows, this is a workaround
+                subprocess.run(["rm", "-rf", temp_dir], stdout=subprocess.PIPE)
```
etaf: There is no "rm" command in Windows native cmd.exe, have you tested this code?

tbohutyn: It is the same as what is done on test teardown https://github.com/pytorch/pytorch/blob/main/torch/testing/_internal/common_utils.py#L5765
and in other tests:
https://github.com/pytorch/pytorch/blob/main/test/test_cpp_extensions_jit.py#L304

dvrogozh: This might be common issue across few tests. Can you, please, check whether that's getting called, is run and if it's executed where `rm` is coming from?

etaf: > It is the same as what is done on test teardown https://github.com/pytorch/pytorch/blob/main/torch/testing/_internal/common_utils.py#L5765 and in other tests: https://github.com/pytorch/pytorch/blob/main/test/test_cpp_extensions_jit.py#L304

These parts are not working on Windows. I had fixed one of them earlier, but it looks like not all of them were properly addressed.

tbohutyn: How did you address it?

etaf: Replace  "rm -rf"  with "rd /s /q".

torch/utils/cpp_extension.py
```diff
@@ -2383,6 +2439,12 @@ def _prepare_ldflags(extra_ldflags, with_cuda, verbose, is_standalone):
             else:
                 extra_ldflags.append(f'-L{_join_rocm_home("lib")}')
                 extra_ldflags.append('-lamdhip64')
+    if with_sycl:
+        if IS_WINDOWS:
```
guangyey, line 2445: @dvrogozh I know this is not directly related to this PR, but I’m curious — why isn’t there a Linux branch here?

guangyey (approved):

torch/utils/cpp_extension.py
```diff
@@ -2383,6 +2439,12 @@ def _prepare_ldflags(extra_ldflags, with_cuda, verbose, is_standalone):
...
+            extra_ldflags.append('c10_xpu.lib')
+            extra_ldflags.append('torch_xpu.lib')
```
guangyey, line 2447: Actually, we also have an additional `torch-xpu-ops.lib` that may need to be linked.

dvrogozh: I think we need a use case example to add this library here. Example will help to create a test to check the feature. I think that's a good consideration, but out of scope of this PR. I have submitted a https://github.com/intel/torch-xpu-ops/issues/2215 for a follow up.

torch/utils/cpp_extension.py
dvrogozh: I looked into this. These libraries are needed for both Windows and Linux, otherwise there will be unresolved symbols. The reason why it works on Linux without adding these libraries on the command line is 1) lazy linkage on the build stage, 2) global symbol resolution at runtime (i.e. required libraries are already loaded by `import torch` by the time extension is being loaded and symbols got resolved).

I think it makes sense to align the code between xpu and cuda/hip and add these libs on Linux as well. @tbohutyn, I have submitted a PR against your fork: https://github.com/tbohutyn/pytorch/pull/1. It adds libs on Linux and adds a test to formally test this out. Please, take a look on PR and merge to your fork if you are ok with it. Once done your pytorch PR will be automatically updated.

guangyey: Thanks @dvrogozh

dvrogozh: @tbohutyn, can you, please, take a look at https://github.com/tbohutyn/pytorch/pull/1 PR and let me know if you plan to merge it in or you would like me to follow up on 162579 once it's merged?

guangyey: I think it would be better to handle this as a follow-up since it targets Linux, while this PR focuses on Windows.

dvrogozh: I am fine with that. I'll post the Linux PR as soon as this one will get merged as they partially overlap and there will be conflicts to resolve. @guangyey, please, help to merge Windows PR.

dvrogozh:
@ezyang , @guangyey , can this PR, please, be merged before any merge conflict has appeared? PR gates further testing and preparation for 2.10 release.

test/test_cpp_extensions_jit.py
```diff
@@ -155,7 +154,12 @@ def _test_jit_xpu_extension(self, extra_sycl_cflags):
             # 2 * sigmoid(0) = 2 * 0.5 = 1
             self.assertEqual(z, torch.ones_like(z))
         finally:
-            shutil.rmtree(temp_dir)
+            if IS_WINDOWS:
+                # rmtree returns permission error: [WinError 5] Access is denied
+                # on Windows, this is a workaround
```
albanD, line 159: Why not set ignore_error=True in rmtree?

albanD (approved):

test/test_cpp_extensions_jit.py
dvrogozh: @tbohutyn : can you, please, check if that's something which can be quick addressed? If not, please, post a note here and let's do that as a follow up.

astachowiczhabana:
Hi @EikanWang 
I'm taking over this PR after @tbohutyn (he's no longer available). The failing test (xpu / linux-noble-xpu-n-py3.10 / test ) is unrelated to the changes in this PR. It’s caused by a timeout issue that has been observed in other PRs as well [HUD](https://hud.pytorch.org/failure?name=xpu%20%2F%20linux-noble-xpu-n-py3.10%20%2F%20test%20(default%2C%207%2C%2012%2C%20linux.idc.xpu)&jobName=linux-noble-xpu-n-py3.10%20%2F%20test%20(default%2C%207%2C%2012%2C%20linux.idc.xpu)&failureCaptures=%27Test%27%20).

torch/utils/cpp_extension.py
dvrogozh: Filed https://github.com/pytorch/pytorch/pull/169322.

PR #162580, opened 2025-09-10
WIP Profiler Performance Visualization

exclamaforte:
This PR creates a compact visualization of all the kernels in a model trace, with bandwidth, flops, and roofline calculation (mem bound vs comp bound) +  call-stack for each kernel. An analogy here would be, if the profile is an audio waveform (samples), then this visualization is a Spectrogram (frequency domain.) You can provide many different traces, and they will be smashed together into the same visualization. The kernels are color coded to match which profile it belongs to. This allows easy comparison of the performance between eager vs compile, or between hardware vendors. The coloring of the kernel nodes provides quick identification of the important kernels by different stats (runtime by default.)

Note: this doesn't work on cudagraphed traces because they lack the cpu events through which to build the graph!

## `--visualize <trace_name_1> <trace_name_2> <dtype> <output>`

Examples:
```
# Single trace visualization 
python profile_analysis.py --visualize trace.json float32 output.svg 
```
a2a84271-test_output2

^ click to open in a new window

```
# Multi-trace comparison 
python profile_analysis.py --visualize baseline.json current.json float32 comparison.svg --compact False
```
945b0061-test_output2

^ click to open in a new window

## --color
Controls how kernel nodes are colored to highlight different performance characteristics. 

Examples:
```
# Color by kernel runtime (default) - darker = longer runtime
python profile_analysis.py --visualize trace.json float32 output.png --color time

# Highlight memory bandwidth utilization = utilization %
python profile_analysis.py --visualize trace.json float32 output.png --color mem-utilization

# Show compute utilization = FLOPS utilization %
python profile_analysis.py --visualize trace.json float32 output.png --color compute-utilization

# Roofline percentage: % away from mem optimal if mem bound, % away from compute optimal if compute bound
python profile_analysis.py --visualize trace.json float32 output.png --color roofline
```
TODO Layer stuff

## --height
Limits visualization complexity by showing only the specified number of operation levels above kernel nodes.

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @mlazos @chenyang78

### DIFF
--- a/torch/_inductor/analysis/dag_nodes.py
+++ b/torch/_inductor/analysis/dag_nodes.py
@@ -1,0 +1,592 @@
+"""
+DAG node classes for representing trace execution graphs.
+"""
+
+from typing import Dict, List, Optional, Tuple, TYPE_CHECKING
+
+from torch.utils._ordered_set import OrderedSet
+
+if TYPE_CHECKING:
+    from .json_profile import JsonProfile
+
+
+try:
+    import graphviz
+except ImportError:
+    pass
+
+
+class TraceDAGNode:
+    """Represents a node in the DAG - either an operation or a kernel."""
+
+    def __init__(self, name: str, node_type: str):
+        self.name = name
+        self.node_type = node_type  # 'op' or 'kernel'
+        self.kernel_instances: List[
+            Tuple[float, int]
+        ] = []  # List of (duration_us, thread_id) for kernels
+        self.instance_count: int = (
+            0  # Number of times this operation appears in the trace
+        )
+        # Performance statistics for kernels
+        self.achieved_flops_list: List[
+            float
+        ] = []  # List of achieved FLOPS % for each instance
+        self.achieved_bandwidth_list: List[
+            float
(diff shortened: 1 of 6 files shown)


ezyang:
Another question is, are you willing to make the output format stable? Because if so, it may be better for the post-processing script to live in a dedicated repo (this seems to be a theme with you, c.f. torchdiode!) so that it can be iterated on more quickly without having to wait for dedicated PyTorch release.

exclamaforte:
@ezyang 
> Is this Inductor only? I see the only files changed are in torch/_inductor
No it's generic. I think another repo makes sense; analyzing the trace is fairly compute intensive given that a normal run could be in the GB, especially if you turn on memory tracing. Maybe take inspiration from `tlparse` with a Rust version if you're happy with that decision?

PR #162595, opened 2025-09-10
Don't unconditionally import torch._dynamo, it's slow

ezyang:
* __->__ #162595

A trivial test on OS X.

Before:

```
real	0m6.550s
user	0m2.532s
sys	0m3.359s
```

After:

```
real	0m2.607s
user	0m1.898s
sys	0m3.344s
```

### DIFF
--- a/torch/testing/_internal/common_utils.py
+++ b/torch/testing/_internal/common_utils.py
@@ -68,7 +68,6 @@ import torch.backends.xnnpack
 import torch.cuda
 from torch import Tensor
 from torch._C import ScriptDict, ScriptList  # type: ignore[attr-defined]
-from torch._dynamo.trace_rules import _as_posix_path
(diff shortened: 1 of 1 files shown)


torch/testing/_internal/common_utils.py
```diff
@@ -5606,6 +5605,8 @@ class LazyVal:
 
 
 def munge_exc(e, *, suppress_suffix=True, suppress_prefix=True, file=None, skip=0):
+    from torch._dynamo.trace_rules import _as_posix_path
```
ezyang, line 5608: if this utility really is useful it might be better to move it somewhere else...

albanD (approved):
Thanks!
As a follow-up, we can extend the test that dynamo is not eagerly imported with torch to also check the testing tools

PR #162615, opened 2025-09-10
[standalone_compile] fix multiprocess compile

zou3519:
* __->__ #162615
* #162432

If multiple processes standalone_compile the same exact thing, this
resulted in two issues:
1) We were always deleting the directory. This leads to one process
   trying to read from something that is no longer there
2) there is a line that iterates through all inductor cache artifacts.
   Another process might have put a temporary file in this directory,
   and we don't want the current process to iterate through temporary
   files, just the new file.

For (2), we can't change `write_atomic` to use a temporary directory.
If the temporary directory is still a subdirectory of the current
directory, then we have the same problem. If we put the temporary
directory elsewhere (like /tmp), we run into issues: if the Inductor
cache directory is not on the same filesystem as the /tmp directory,
then we cannot move files between them without a copy.

Test Plan:
```
CUDA_VISIBLE_DEVICES=2 vllm serve --gpu-memory-utilization 0.1 &
CUDA_VISIBLE_DEVICES=3 vllm serve --gpu-memory-utilization 0.1 &
CUDA_VISIBLE_DEVICES=4 vllm serve --gpu-memory-utilization 0.1 &
CUDA_VISIBLE_DEVICES=5 vllm serve --gpu-memory-utilization 0.1 &
CUDA_VISIBLE_DEVICES=6 vllm serve --gpu-memory-utilization 0.1 &
CUDA_VISIBLE_DEVICES=7 vllm serve --gpu-memory-utilization 0.1 &
```


oulgen (changes requested):
discussed offline, this feels very brittle

PR #162620, opened 2025-09-10
bf16 support for fused_moving_avg_obs_fake_quant() op

liangel-02:
enabling bf16 support for `torch.fused_moving_avg_obs_fake_quant()` op on cuda

**testing**
`python test/quantization/pt2e/test_quantize_pt2e.py`

### DIFF
--- a/test/quantization/core/test_workflow_module.py
+++ b/test/quantization/core/test_workflow_module.py
@@ -908,6 +908,19 @@ class TestFakeQuantize(TestCase):
         self.assertEqual(fq_module.activation_post_process.quant_min, 0)
         self.assertEqual(fq_module.activation_post_process.quant_max, 127)
 
+    @given(device=st.sampled_from(['cpu', 'cuda'] if torch.cuda.is_available() else ['cpu']))
+    def test_fused_moving_avg_obs_fake_quant(self, device):
+        try:
+            sampled_dtype = st.sampled_from(["bf16", "fp32"]) if device == "cuda" else "fp32"
+            dtype = torch.bfloat16 if sampled_dtype == "bf16" else torch.float32
+            torch.set_default_dtype(dtype)
+
+            with torch.device(device):
+                fake_quantize = FusedMovingAvgObsFakeQuantize()
+                fake_quantize.forward(torch.rand((256, 512)))
+        finally:
+            torch.set_default_dtype(torch.float32)
+
 def _get_buffer_ids(module):
     """
     Object addresses stay constant if and only if all modifications are in-place
--- a/test/quantization/core/test_workflow_ops.py
+++ b/test/quantization/core/test_workflow_ops.py
@@ -3,6 +3,7 @@
 
 import torch
 import math
+from typing import Union
 from torch.ao.quantization import (
     FakeQuantize,
     MovingAverageMinMaxObserver,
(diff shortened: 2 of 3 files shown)


test/quantization/pt2e/test_quantize_pt2e.py
```diff
@@ -55,10 +55,20 @@
     TEST_CUDA,
     TEST_HPU,
 )
+from torchao.quantization.pt2e import FusedMovingAvgObsFakeQuantize
 
 
 @skipIfNoQNNPACK
 class TestQuantizePT2E(PT2EQuantizationTestCase):
+    @parametrize("dtype", (torch.float32, torch.bfloat16))
+    def test_fused_moving_avg_obs_fake_quant(self, dtype):
```
jerryzh168: module test should be in https://github.com/pytorch/pytorch/blob/7e2e83cdbe532b230dee40cfe0454116c9b64710/test/quantization/core/test_workflow_module.py#L856 I think

also add an op test in https://github.com/pytorch/pytorch/blob/7e2e83cdbe532b230dee40cfe0454116c9b64710/test/quantization/core/test_workflow_ops.py#L255?

I think we also need to test the output against reference implementations

andrewor14 (approved):
Looks great other than Jerry's test comment. Thanks!

jerryzh168 (approved):
LG, thanks!

test/quantization/core/test_workflow_module.py
```diff
@@ -908,6 +908,18 @@ def test_quant_min_max_override(self):
         self.assertEqual(fq_module.activation_post_process.quant_min, 0)
         self.assertEqual(fq_module.activation_post_process.quant_max, 127)
 
+    @given(device=st.sampled_from(['cpu', 'cuda'] if torch.cuda.is_available() else ['cpu']))
+    def test_fused_moving_avg_obs_fake_quant(self, device):
+        try:
+            dtype = torch.bfloat16 if device == 'cuda' else torch.float32
```
andrewor14: should we also test cuda with torch.float32? If I understand correctly this test doesn't do that right now

test/quantization/core/test_workflow_ops.py
```diff
@@ -1064,10 +1072,13 @@ def test_fused_obs_fake_quant_moving_avg(self, device, symmetric_quant, use_bool
         Tests the case where we call the fused_obs_fake_quant op multiple times
         and update the running_min and max of the activation tensors.
         """
-        in_running_min_ref = out_running_min_ref = float("inf")
-        in_running_min_op = torch.tensor(float("inf"), device=device)
-        in_running_max_ref = out_running_max_ref = float("-inf")
-        in_running_max_op = torch.tensor(float("-inf"), device=device)
+        sampled_dtype = st.sampled_from(["bf16", "fp32"]) if device == "cuda" else "fp32"
```
andrewor14, line 1075: I think we should avoid using `sampled_from` where possible to reduce potential non-determinism. Can we parametrize on `[bf16, fp32]` instead?

andrewor14 (approved):

test/quantization/core/test_workflow_ops.py
liangel-02: decided offline to keep this format to maintain consistency with the rest of the tests in this file

PR #162636, opened 2025-09-10
[FSDP][Replicate] tests replicate casting module after init

anshul-si:
**Summary:** In order to ensure that replicate acts as intended (a specialized version of hsdp) we need to make sure that it can pass the same tests that fully_shard can for training. This test is important as it verifies we can cast a replicated module to a different type after initialization, and import feature for enabling mixed precision,

**Test Cases**
1. pytest test/distributed/_composable/test_replicate_training.py -k test_to_float64_after_init

* #162861
* #162855
* #162853
* #162851
* #162839
* #162836
* #162830
* #162785
* #162658
* #162656
* #162654
* #162650
* __->__ #162636
* #162631


mori360 (approved):
Thanks for the test

PR #162644, opened 2025-09-10
Fix protobuf test comparison by parsing proto instead of raw strings

0xjeffro:
The tests were comparing raw exported strings for protobuf comparison, which is not backward/forward compatible with different versions of protobuf.

This PR parses the strings into protobuf and compares the protobufs directly, similar to what we did in assertImageProto.

Our test failed because we used a different version of protobuf, which output 44100.0 instead of 44100, which resulted in an error. However, they are equal, but only different in the exported strings.

cc @haifeng-jin

### DIFF
--- a/test/test_tensorboard.py
+++ b/test/test_tensorboard.py
@@ -82,13 +82,14 @@ class BaseTestCase(TestCase):
             if os.path.exists(temp_dir):
                 shutil.rmtree(temp_dir)
 
-    def assertProto(self, str_to_compare):
+    def assertProto(self, actual_proto):
         if expecttest.ACCEPT:
-            write_proto(str_to_compare, self)
+            write_proto(actual_proto, self)
             return True
-        expected = read_expected_content(self)
-        str_to_compare = str(str_to_compare)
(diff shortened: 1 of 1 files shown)


test/test_tensorboard.py
```diff
@@ -82,13 +82,14 @@ def tearDown(self):
             if os.path.exists(temp_dir):
                 shutil.rmtree(temp_dir)
 
-    def assertProto(self, str_to_compare):
+    def assertProto(self, actual_proto):
```
justinchuby, line 85: Do you need to update the call sites for them to pass in proto objects?

haifeng-jin: Thanks for the comments!
No, we do not need to because they are passing the actaul proto already. ([link](https://github.com/pytorch/pytorch/blob/9f4989e205f05e7a75438c10e7e9e6f87566fc57/test/test_tensorboard.py#L495))
It is the naming of `str_to_compare` that is confusing. It is a protobuf not string. They had to use `str(str_to_compare)` to convert the protobuf to string on line 90.

justinchuby (approved):

Skylion007 (approved):

PR #162645, opened 2025-09-10
[ONNX] Implement while_loop

### LINKED ISSUE #146674: [ONNX] Support while and scan HOPs

justinchuby:
Fixes #146674 

cc @titaiwangms

### DIFF
--- a/torch/onnx/_internal/exporter/_torchlib/ops/hop.py
+++ b/torch/onnx/_internal/exporter/_torchlib/ops/hop.py
@@ -264,9 +264,7 @@ def higher_order_while_loop(
     )
 
     if len(cond_func.outputs) != 1:
-        raise AssertionError(
-            "Condition function must return a single boolean value."
-        )
+        raise AssertionError("Condition function must return a single boolean value.")
 
     # ONNX Runtime complains about duplicate output names if we don't rename them
     for func_out, out in zip(body_func.outputs, body_node.outputs):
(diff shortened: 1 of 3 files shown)


xadupre (approved):

titaiwangms (commented):
It looks like it's not passing its own test.

torch/onnx/_internal/exporter/_torchlib/ops/hop.py
```diff
@@ -155,3 +155,140 @@ def higher_order_scan(
         scan_input_directions=[(1 if reverse else 0) for _ in scan_inputs],
         scan_output_directions=[(1 if reverse else 0) for _ in range(n_outputs)],
     )
+
+
+@onnx_impl(torch.ops.higher_order.while_loop, no_compile=True)
```
titaiwangms, line 161: When do we compile the function? Or this is an outdated parameter?

justinchuby: Pretty much never (we don’t even make it a traced function, just leave it as is). We can remove it

justinchuby: Actually sorry - onnx_impl is used for other ops too. So we still need it for now.

titaiwangms: I don't see any op uses it. Did you mean we should have another PR to deprecate it?

PR #162646, opened 2025-09-10
Smoke tests don't run nvshmem on Windows

atalman:
Only available for linux x86 and aarch64 :
https://pypi.org/project/nvidia-nvshmem-cu13/#files

nvshmem is available only on linux:
``
"nvidia-nvshmem-cu12==3.3.24; platform_system == 'Linux' and platform_machine == 'x86_64' | "
``
https://github.com/pytorch/pytorch/blob/main/.github/scripts/generate_binary_build_matrix.py#L57

### DIFF
--- a/.ci/pytorch/smoke_test/smoke_test.py
+++ b/.ci/pytorch/smoke_test/smoke_test.py
@@ -396,7 +396,9 @@ def smoke_test_nvshmem() -> None:
     except ImportError:
         # Not built with NVSHMEM support.
         # torch is not compiled with NVSHMEM prior to 2.9
-        if torch.__version__ < "2.9.*":
+        maj = int(torch.__version__.split(".")[0])
+        min = int(torch.__version__.split(".")[1])
+        if (maj, min) < (2,9):
             return
         else:
             # After 2.9: NVSHMEM is expected to be compiled in current build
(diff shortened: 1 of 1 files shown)


malfet (changes requested):
This should have been the case already, shouldn't it? I.e. what changed?

kwen2501 (approved):

.ci/pytorch/smoke_test/smoke_test.py
```diff
@@ -396,7 +396,7 @@ def smoke_test_nvshmem() -> None:
     except ImportError:
         # Not built with NVSHMEM support.
         # torch is not compiled with NVSHMEM prior to 2.9
-        if torch.__version__ < "2.9":
```
atalman, line 399: We are trying to compare different types of string:
```
>>> "2.10.0.dev20250910+cu128" < "2.9"
True
>>> "2.10.0.dev20250910+cu128" < "2.9.*"
True
>>> 
``

Comment (kwen2501) [reply .ci/pytorch/smoke_test/smoke_test.py:399]: I believe python has a way to convert string "2.9" into a version object?

Comment (atalman) [reply .ci/pytorch/smoke_test/smoke_test.py:399]: Yes using packaging, however this is not a built in module, don't want to add another dependency to smoke test

File: .ci/pytorch/smoke_test/smoke_test.py [hunk 2]
```diff
@@ -396,7 +396,9 @@ def smoke_test_nvshmem() -> None:
...
+        maj = int(torch.__version__.split(".")[0])
```
malfet: Please don't do that (or at least use `maj, min = [int(x) for x in torch.__version__.split(".")[:2]]` ) but rely on TorchVersion

PR #162647, opened 2025-09-10
compile_kernel large shared memory fix

msaroufim:
Alternate solution to https://github.com/pytorch/pytorch/pull/162328

cc @gau-nernst

### DIFF
--- a/torch/cuda/_utils.py
+++ b/torch/cuda/_utils.py
@@ -214,7 +214,7 @@ class _CudaKernel:
     def __init__(self, func: ctypes.c_void_p, module: ctypes.c_void_p) -> None:
         self.func = func
         self.module = module
-        self._max_shared_mem_bytes = None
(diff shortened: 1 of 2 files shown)


torch/cuda/_utils.py
```diff
@@ -282,6 +314,40 @@ def __call__(
...
 
+    def set_shared_memory_config(self, shared_mem_bytes: int) -> None:
+        if shared_mem_bytes < 48 * 1024:
+            # No configuration needed for <= 48KB, just update the value
+            self._max_shared_mem_bytes = shared_mem_bytes
+            return
+
+        import torch
+
+        libcuda = torch.cuda._utils._get_cuda_library()
```
gau-nernst: can't you just call `_get_cuda_library()` here? without importing torch

eqy (approved):

PR #162648, opened 2025-09-10
[ROCm] Build FBGEMM_GENAI for gfx942 only

jithunnair-amd:
Fixes build timeouts >4h on libtorch build jobs: https://hud.pytorch.org/hud/pytorch/pytorch/75e7f49f9c70116d7c4f8f86c3d0688ade306284/1?per_page=50&name_filter=inux-binary-libtorch%20%2F%20libtorch-rocm&mergeEphemeralLF=true

Brings back code to narrow down CK compilation targets from https://github.com/pytorch/pytorch/commit/69a25f68884a168550695fdb1a7c310c54d29536#diff-ce80f3115ab2f6be5142f0678a1fc92c6b2d7727766ce44f48726c99e720f777

gfx942 supports fp8

Don't enable gfx950 for now, until more optimizations are in place as per https://github.com/pytorch/pytorch/pull/162648/files#r2369588738

Validation:
[rocm6.4](https://github.com/pytorch/pytorch/actions/runs/17944766350/job/51028483128) and [rocm6.3](https://github.com/pytorch/pytorch/actions/runs/17944766350/job/51028483093) libtorch builds finished within 3.9h.

cc @jeffdaily @sunway513 @pruthvistony @ROCmSupport @dllehr-amd @jataylo @hongxiayang @naromero77amd @danielvegamyhre (since their [change](https://github.com/pytorch/pytorch/commit/b6d0a9ea9056ede4f7024dbf3bd6c43be3aff49c) had removed this snippet, causing ROCm builds to increase >4h)


torch/testing/_internal/common_cuda.py
```diff
@@ -131,15 +131,14 @@ def evaluate_platform_supports_mx_gemm():
     return False
 
 def evaluate_platform_supports_mxfp8_grouped_gemm():
-    if torch.cuda.is_available() and not torch.version.hip:
+    if torch.cuda.is_available():
         built_with_fbgemm_genai = "USE_FBGEMM_GENAI" in torch.__config__.show()
-        return built_with_fbgemm_genai and IS_SM100
+        return built_with_fbgemm_genai and evaluate_platform_supports_mx_gemm
```
danielvegamyhre: `evaluate_platform_supports_mx_gemm` returns true for sm100 or later but i believe the cutlass mxfp8 grouped gemm has only been built and tested on sm100a, not sm120 etc.

danielvegamyhre:
>@danielvegamyhre (since their [change](https://github.com/pytorch/pytorch/commit/b6d0a9ea9056ede4f7024dbf3bd6c43be3aff49c) had removed this snippet, causing ROCm builds to increase >4h)

Can you clarify what code exactly was removed in my PR that cause build time to increase for AMD...? It's not clear to me and I'd like to understand, thanks

jeffdaily:
> > @danielvegamyhre (since their [change](https://github.com/pytorch/pytorch/commit/b6d0a9ea9056ede4f7024dbf3bd6c43be3aff49c) had removed this snippet, causing ROCm builds to increase >4h)
> 
> Can you clarify what code exactly was removed in my PR that cause build time to increase for ROCM...? It's not clear to me and I'd like to understand, thanks

https://github.com/pytorch/pytorch/commit/b6d0a9ea9056ede4f7024dbf3bd6c43be3aff49c#diff-ce80f3115ab2f6be5142f0678a1fc92c6b2d7727766ce44f48726c99e720f777L277-L282

torch/testing/_internal/common_cuda.py
cthi: For the MXFP8 grouped support via fbgemm that @danielvegamyhre  added support for, it would only support nvidia right now. So I think we should revert this change.

jithunnair-amd: Okay, no issues, I can revert this specific change to limit it to SM100. I was trying to set the stage for MI350 that *does* support MXFP8, but we can always add it later after ensuring it supports grouped GEMM. The main intent of this PR is to reduce the build time.

jeffdaily: Will revert this change for now until MI350 CI is in place.  Note that the mxfp8 scaled_mm operator is supported on MI350, tested locally.

jeffdaily (approved):

CMakeLists.txt
```diff
@@ -885,7 +885,7 @@ cmake_dependent_option(
   "USE_ROCM"
   OFF)
 
-IF(USE_FBGEMM_GENAI AND USE_ROCM AND NOT "gfx942" IN_LIST PYTORCH_ROCM_ARCH)
+IF(USE_FBGEMM_GENAI AND USE_ROCM AND NOT ("gfx942" IN_LIST PYTORCH_ROCM_ARCH OR "gfx950" IN_LIST PYTORCH_ROCM_ARCH))
```
cthi: We should only turn on fbgemm for 942, so 950 related things could be removed from here as well.

aten/src/ATen/CMakeLists.txt
```diff
@@ -317,10 +317,23 @@ IF(USE_FBGEMM_GENAI)
...
         -fhip-new-launch-api)
 
+      # Only compile for gfx942 for now.
+      # This is rather hacky, I could not figure out a clean solution :(
+      set(HIP_CLANG_FLAGS_ORIGINAL ${HIP_CLANG_FLAGS})
+      string(REGEX REPLACE "--offload-arch=[^ ]*" "" FILTERED_HIP_CLANG_FLAGS "${HIP_CLANG_FLAGS}")
+      if("gfx942" IN_LIST PYTORCH_ROCM_ARCH)
+        list(APPEND FILTERED_HIP_CLANG_FLAGS --offload-arch=gfx942;)
+      endif()
+      if("gfx950" IN_LIST PYTORCH_ROCM_ARCH)
```
cthi: Same as my comment above.

CMakeLists.txt
jithunnair-amd: @cthi Just to be clear, this is because we don't have any data on how FBGEMM GENAI does on gfx950, right? Would you like ROCm team to own the testing and enablement part of FBGEMM GENAI on gfx950? Are the workloads/benchmarks to be used public, or will we need Meta internal team to help us validate func and perf?

Cc @jeffdaily @pruthvistony

cthi: @jithunnair-amd  Sorta - let me go in more detail about this.

Currently for ROCm we've integrated from fbgemm the grouped gemm FP8 rowwise (CK based) that works on gfx942. It would work on gfx950 as well (but needs some minor changes around the FP8 dtype since MI350 would not use fp8 FNUZ but instead OCP I believe) and that also likely need more fixes in Blas.cpp maybe), but the performance would not be a big improvement.

We plan to integrate CK tile based FP8 group gemm this year, which would let us get full potential on MI350 and improvements on MI300.

jithunnair-amd: Removing gfx950 from list as per @cthi's request

aten/src/ATen/CMakeLists.txt
jithunnair-amd: Removed gfx950

CMakeLists.txt
jithunnair-amd: cc @jagadish-amd for awareness

PR #161007, opened 2025-08-19
[rfc] add hint_override kwarg to mark_dynamic

bobrenjc93:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #161007

The motivation for this change can be seen through the following example:

```
import torch

GPU_TYPE = "cuda"

@torch.compile
def no_override(x):
    return x.sum(dim=0)

@torch.compile
def override(x):
    return x.sum(dim=0)

x_small = torch.randn(4096, 512, device=GPU_TYPE)
no_override(x_small)
torch._dynamo.decorators.mark_dynamic(x_small, 0, hint_override=4096 * 1000)
override(x_small)
```

Previously, when reductions were split, codegen relied only on the first observed shape. With a small input, this resulted in a small split size:

```
def triton_red_fused_sum_0(in_ptr0, out_ptr0, ks0, xnumel, r0_numel, XBLOCK : tl.constexpr, R0_BLOCK : tl.constexpr):
    xnumel = 16384
    rnumel = r0_numel
```

With the new scheme, inductor honors hint_override during codegen, producing larger and more appropriate split sizes:

```
def triton_red_fused_sum_0(in_ptr0, out_ptr0, ks0, xnumel, r0_numel, XBLOCK : tl.constexpr, R0_BLOCK : tl.constexpr):
    xnumel = 1024000
    rnumel = r0_numel
```

This addresses a broader problem with dynamism: performance and numerics previously depended on whichever shape was seen first. For example:

```
f(s0) -> f(s2)
f(s1) -> f(s2)
```

could generate different kernels. With the new approach, an explicit override pins the chosen configuration:

```
f(s0, hint_override=s0) -> f(s2)
f(s1, hint_override=s0) -> f(s2)
```

ensuring consistent kernel generation regardless of input order.

cc @ezyang @SherlockNoMad @EikanWang @jgong5 @wenzhe-nrv @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela

PR #162652, opened 2025-09-10
add support for hint_override in mark_unbacked

bobrenjc93:
* __->__ #162652

Very similar to https://github.com/pytorch/pytorch/pull/161007 except now for mark_unbacked.

cc @ezyang @SherlockNoMad @EikanWang @jgong5 @wenzhe-nrv @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela

### DIFF
--- a/test/inductor/test_torchinductor.py
+++ b/test/inductor/test_torchinductor.py
@@ -10676,9 +10676,9 @@ def forward(self, arg0_1: "Sym(s77)", arg1_1: "Sym(s27)", arg2_1: "Sym(s53)", ar
 
     @requires_gpu()
     @skip_if_not_triton
-    # @unittest.skipIf(
-    #     not IS_BIG_GPU, "Skipping triton backend only since not big GPU (not enough SM)"
-    # )
+    @unittest.skipIf(
+        not IS_BIG_GPU, "Skipping triton backend only since not big GPU (not enough SM)"
+    )
     @config.patch({"force_disable_caches": True})
     def test_mark_unbacked_with_hint_override(self):
         @torch.compile
@@ -10688,6 +10688,12 @@ def forward(self, arg0_1: "Sym(s77)", arg1_1: "Sym(s27)", arg2_1: "Sym(s53)", ar
         @torch.compile
         def override(x):
             return x.sum(dim=0)
+
+        @torch.compile(fullgraph=True)
+        def branching(x):
+            if x.shape[0] > 4096:
+                return 1
+            return 2
 
         x_small = torch.randn(4096, 512, device=GPU_TYPE)
         torch._dynamo.decorators.mark_unbacked(x_small, 0)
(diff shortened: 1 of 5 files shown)


test/inductor/test_torchinductor.py
```diff
@@ -10674,6 +10674,34 @@ def override(x):
...
+    # )
+    @config.patch({"force_disable_caches": True})
+    def test_mark_unbacked_with_hint_override(self):
+        @torch.compile
+        def no_override(x):
+            return x.sum(dim=0)
+
+        @torch.compile
+        def override(x):
+            return x.sum(dim=0)
```
laithsakka, line 10690: can we ensure that we not use the hint to decide branching
maybe ensure we throw DDE still if we branch on unbacked?

i feel we need parallel API to access the hint or flag to size_hint that allow reading it for unbacked.

test/inductor/test_torchinductor.py
```diff
@@ -10674,6 +10674,34 @@ def override(x):
...
+
+        x_small = torch.randn(4096, 512, device=GPU_TYPE)
+        torch._dynamo.decorators.mark_unbacked(x_small, 0)
+        code1 = run_and_get_triton_code(no_override, x_small)
+
+        torch._dynamo.reset_code_caches()
+
+        torch._dynamo.decorators.mark_unbacked(x_small, 0, hint_override=4096 * 10)
+        code2 = run_and_get_triton_code(override, x_small)
+        breakpoint()
```
laithsakka: not needed

test/inductor/test_torchinductor.py
bobrenjc93: added test

torch/_inductor/sizevars.py
```diff
@@ -557,7 +558,7 @@ def symbolic_hint(
             return hint_override
 
         expr = self.remove_precomputed_replacements(expr)
-        return sympy_subs(expr, self.var_to_val)
+        return sympy_subs(sympy_subs(expr, self.var_to_hint_override), self.var_to_val)
```
laithsakka: risky but ok

laithsakka (approved):

ezyang (commented):
Help me understand why this prevents people from guarding/adding runtime asserts based on the hints. It seems unsafe to allow arbitrary queries to symbolic_hint to rely on this. You need to make some assertion at the use site that you promise to ONLY use this in a heuristic nature, and that your output is sound even if it varies the full range.

ezyang:
Thanks, and hope you are feeling better. My puzzlement now is, why is this flag not implied by the fallback argument at all these call sites?

bobrenjc93:
> why is this flag not implied by the fallback argument at all these call sites?

My intuition is against muddying the waters since `fallback` is a global fallback (all unbacked symints get the same value) whereas the `use_user_provided_hint_override` flag opts into a unique user provided value on a per-unbacked basis. 

Are you implying deriving `use_user_provided_hint_override` via `use_user_provided_hint_override=fallback is not None`? I just worry that can be a confusing API since a user might pass in `fallback=0` and be surprised when autotuning happens with a different value. Admittedly this is still the case with `fallback=0 && use_user_provided_hint_override=True` but, at least in mind, it's much easier to reason through since the flags are separate.

Or perhaps I'm misunderstanding? If you have a cleaner API proposal, I'm all ears!

bobrenjc93:
Synced with @ezyang offline and collectively decided to consolidate on the `fallback` kwarg.

PR #162655, opened 2025-09-10
[aoti] add config for libtorch free so

yushangdi:
Users can specify the following to get a libtorch_free `.so`. 

"aot_inductor.use_libtorch": False, 

The following config is only used for torchnative (see https://github.com/meta-pytorch/torchnative/pull/110). It's not intended to be used by executorch. The reason we need it for torchnative is because a lot of the symbol definitions in torchnative repo is only in header files. 

"aot_inductor.libtorch_free_header": "/data/users/shangdiy/torchnative/standalone,/data/users/shangdiy/torchnative/" (or their custom headers)

The main motivating use case is for executorch to produce a libtorch free `.so`.

TODO for follow-up PR: this flag should be consolidated with the `compile_standalone` flag. 
  

### DIFF
--- a/torch/_inductor/config.py
+++ b/torch/_inductor/config.py
@@ -1580,6 +1580,17 @@ class aot_inductor:
     # Whether to enable link-time-optimization
     enable_lto = os.environ.get("AOT_INDUCTOR_ENABLE_LTO", "0") == "1"
 
+    # Whether the compiled .so should link to libtorch
+    # TODO: should consolidate this flag with compile_standalone
+    link_libtorch: bool = True
+
+    # If None, the default torch headers such as torch/include
+    # will be used. Otherwise, the provided path will be used instead.
+    # This is needed for torchnative to load libtorch-free .so.
+    # Such as [f"{torchnative_dir}/standalone",f"{torchnative_dir}/",].
+    # TODO: should consolidate this flag with compile_standalone
+    libtorch_free_headers: Optional[list[str]] = None
+
 
 class cuda:
     """Settings for cuda backend, today this consists of cutlass"""
--- a/torch/_inductor/cpp_builder.py
+++ b/torch/_inductor/cpp_builder.py
@@ -1086,13 +1086,17 @@ def _get_torch_related_args(
 ) -> tuple[list[str], list[str], list[str]]:
     from torch.utils.cpp_extension import include_paths, TORCH_LIB_PATH
 
-    include_dirs = include_paths()
-    libraries_dirs = [TORCH_LIB_PATH]
     libraries = []
-    if sys.platform != "darwin" and not config.is_fbcode():
-        libraries = ["torch", "torch_cpu"]
-        if not aot_mode:
-            libraries.append("torch_python")
+    include_dirs = config.aot_inductor.libtorch_free_headers or include_paths()
+
+    if config.aot_inductor.link_libtorch:
+        libraries_dirs = [TORCH_LIB_PATH]
+        if sys.platform != "darwin" and not config.is_fbcode():
+            libraries.extend(["torch", "torch_cpu"])
+            if not aot_mode:
+                libraries.append("torch_python")
+    else:
+        libraries_dirs = []
 
     if _IS_WINDOWS:
         libraries.append("sleef")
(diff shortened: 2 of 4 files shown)


torch/_inductor/cpp_builder.py
```diff
@@ -1086,13 +1086,22 @@ def _get_torch_related_args(
...
+        include_dirs = include_paths()
+        libraries_dirs = [TORCH_LIB_PATH]
+        if sys.platform != "darwin" and not config.is_fbcode():
+            libraries = ["torch", "torch_cpu"]
+            if not aot_mode:
+                libraries.append("torch_python")
+    else:
+        if config.aot_inductor.libtorch_free_header is None:
+            raise RuntimeError(
+                "Must provide ot_inductor.libtorch_free_header config when aot_inductor.use_libtorch=False"
```
angelayi:
```suggestion
                "Must provide aot_inductor.libtorch_free_header config when aot_inductor.use_libtorch=False"
```

torch/_inductor/config.py
```diff
@@ -1580,6 +1580,14 @@ class aot_inductor:
     # Whether to enable link-time-optimization
     enable_lto = os.environ.get("AOT_INDUCTOR_ENABLE_LTO", "0") == "1"
 
+    # Whether to compile the model with libtorch.
+    # If doesn't want to compile with libtorch, an address to header
+    # must be provided for the compilation.
```
angelayi: can you explain more ab the requirements for the header file that needs to be provided

torch/utils/cpp_extension.py
```diff
@@ -1505,12 +1505,14 @@ def include_paths(device_type: str = "cpu") -> list[str]:
...
     """
+    paths = []
     lib_include = os.path.join(_TORCH_PATH, 'include')
-    paths = [
-        lib_include,
-        # Remove this once torch/torch.h is officially no longer supported for C++ extensions.
-        os.path.join(lib_include, 'torch', 'csrc', 'api', 'include'),
-    ]
+    if torch_include_dirs:
+        paths = [
```
angelayi:
```suggestion
        paths.extend([
```

torch/utils/cpp_extension.py
```diff
@@ -1543,8 +1545,12 @@ def library_paths(device_type: str = "cpu") -> list[str]:
...
         A list of library path strings.
     """
-    # We need to link against libtorch.so
-    paths = [TORCH_LIB_PATH]
+
+    paths = []
+
+    if torch_include_dirs:
+        # We need to link against libtorch.so
+        paths = [TORCH_LIB_PATH]
```
angelayi:
```suggestion
        paths.append(TORCH_LIB_PATH)
```

angelayi (approved):

torch/_inductor/config.py
Gasoonjia: same question. Mind provide the structure / format for the provide header?

yushangdi: I added an example. After the unified runtime discussion, I think this config should only be used for torchnative. Execturoch should not use this config. Executorch should only need the `link_libtorch=False` config.

torch/_inductor/config.py
```diff
@@ -1580,6 +1580,18 @@ class aot_inductor:
...
+    # TODO: should consolidate this flag with compile_standalone
+    link_libtorch: bool = True
+
+    # If None, the default torch headers such as torch/include
+    # will be used. Otherwise, the provided path will be used instead.
+    # If there're multiple directories to include, separate by ",".
+    # This is needed for torchnative to load libtorch-free .so.
+    # Such as f"{torchnative_dir}/standalone,{torchnative_dir}/".
+    # TODO: should consolidate this flag with compile_standalone
+    libtorch_free_header: Optional[str] = None
```
angelayi: why cant we just take a list?

torch/_inductor/cpp_builder.py
```diff
@@ -1086,13 +1086,21 @@ def _get_torch_related_args(
...
     libraries = []
-    if sys.platform != "darwin" and not config.is_fbcode():
-        libraries = ["torch", "torch_cpu"]
-        if not aot_mode:
-            libraries.append("torch_python")
+    include_dirs = []
+    if config.aot_inductor.libtorch_free_header is None:
+        include_dirs.extend(include_paths())
+    else:
+        include_dirs.extend(config.aot_inductor.libtorch_free_header.split(","))
```
angelayi: `include_dirs = config.aot_inductor.libtorch_free_headers or include_paths()` ? 😛

torch/_inductor/config.py
yushangdi: good point, changed to optional list now

angelayi (approved):

PR #161182, opened 2025-08-21
Move non inductor workflows to Python 3.9 -> 3.10

atalman:
Related to: https://github.com/pytorch/pytorch/issues/161167

PR #162657, opened 2025-09-10
[ez][CI] Fix docs push in nightly workflow

clee2000:
HUD metrics page says docs push hasn't happened in 21 days
image

I guess main branch docs just haven't been updated?  Did anyone notice?  Do we care?

Either way I think this should fix it

Likely started after https://github.com/pytorch/pytorch/pull/161182


huydhn (approved):
Oh, good catch!

PR #162662, opened 2025-09-10
Back DTensor with Symmetric Memory ops

kwen2501:
* (to be filled)

If user allocate inner of DTensor with Symmetric Memory, DTensor's `redistribute` op gets accelerated by SymmMem. For example:
```
local_tensor = symm_mem.empty((32, 32), device=self.device)
dtensor = DTensor.from_local(
    local_tensor, device_mesh, [Partial()],
)
dtensor.redistribute(device_mesh, [Replicate()])  # Two-shot all-reduce will be triggered
```

- Intra-node support only.
- Needs explicit enablement.

TODO:
- Use MemPool for creating output tensors

### DIFF
--- a/torch/csrc/distributed/c10d/Functional.cpp
+++ b/torch/csrc/distributed/c10d/Functional.cpp
@@ -6,7 +6,10 @@
 #include <torch/csrc/distributed/c10d/Functional.hpp>
 #include <torch/csrc/distributed/c10d/GroupRegistry.hpp>
 #include <torch/csrc/distributed/c10d/ProcessGroup.hpp>
+#include <torch/csrc/distributed/c10d/symm_mem/SymmetricMemory.hpp>
 #include <utility>
+
+using namespace c10d::symmetric_memory;
 
 namespace {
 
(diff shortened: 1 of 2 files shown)


torch/csrc/distributed/c10d/Functional.cpp
```diff
@@ -96,10 +121,22 @@ at::Tensor all_reduce(
...
-  auto output = input_real.clone(at::MemoryFormat::Contiguous);
-  auto output_ret =
-      all_reduce_(output, std::move(reduce_op), std::move(group_name));
-  return input.is_complex() ? at::view_as_complex(output_ret) : output_ret;
+
+  if (is_symm_mem_enabled_for_funcoll()) {
+    auto output_ret = at::empty_like(input_real, at::MemoryFormat::Contiguous);
+    static auto op =
+        c10::Dispatcher::singleton()
+            .findSchemaOrThrow("symm_mem::two_shot_all_reduce_out", "")
```
ezyang, line 129: Do we consider it desirable for the symm_mem ops to actually show up in traces? I don't think so. For one, these operators are not functional anymore (since you explicitly allocate output_ret and seemingly write into it). That will make it harder for the compiler passes to propagate them

kwen2501: I agree there may be two design choices here:
1. Use `if-else` to branch normal case and symm-mem case, as did in this PR.
2. Add a dispatch mode based on whether the tensor is a symmetric tensor -- this requires a tensor subclass, if I am not mistaken? What would be the longer-term cost if we choose to do that?

For option 1, I can move the allocator into `two_shot_all_reduce` if that's desired.

Btw, the current implementation of funcoll is not functional at all -- if you look at the normal path, it is also pre-alloc->call in-place op.

torch/csrc/distributed/c10d/Functional.cpp
```diff
@@ -159,10 +196,21 @@ at::Tensor all_gather_into_tensor(
...
+    auto output =
+        allocate_all_gather_output(real_input, group_size, group_name);
+    static auto op =
+        c10::Dispatcher::singleton()
+            .findSchemaOrThrow("symm_mem::multimem_all_gather_out", "")
+            .typed<at::Tensor(const at::Tensor&, std::string, at::Tensor)>();
+    op.call(real_input, group_name, output);
+    return input.is_complex() ? at::view_as_complex(output) : output;
+  } else {
+    std::vector<at::Tensor> inputs{real_input};
```
Skylion007, line 209:
```suggestion
    std::vector<at::Tensor> inputs{std::move(real_input)};
```

torch/csrc/distributed/c10d/Functional.cpp
kwen2501: Discussed offline, since the funcoll itself is an op registered with dispatcher, the graph will just see the funcoll instead of the symm-mem op. And internally it'd call symmetric memory. Thus option 1 is okay.

PR #162663, opened 2025-09-10
[CUDAGraph][UX] error instead of warn for excessive re-record

BoyuanFeng:
Excessive re-recording CUDAGraphs lead to bad performance. We previously warns if this happens.

However, the limit (=50) is too high and users may just observe bad performance before actually seeing the warning message. Even worse, users may not see the warning message when there are many other logs. @anijain2305 reported that he never saw this warning message when using transformer library, but he DOES observe slowdown due to cudagraph re-recording & needs to turn off cudagraph.

In practice, we usually observe either tons of re-recordings, or rare re-recording. This PR reduces the limit to 8 and hard error instead of warn. Also provides more error messages.

Fixes #162299

cc @mcarilli @ezyang @eellison @penguinwu @voznesenskym @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela @mlazos


eellison (commented):
+1 - this is BC breaking, and running a large number of batch sizes in inference is a legitimate use case.

I'd prefer to change this to more aggressively warning on a smaller number of re-records but I'm open to conversation.

PR #162665, opened 2025-09-10
[DTensor] Introduce DebugMode

SherlockNoMad:
Introduce a lightweight TorchDispatchMode for understanding the magic behind DTensor. 

- Tracks redistribution, see `redistribute_input(input_idx, from_placement, to_placement)` 
- Optionally tracks torch-level functions, via `__torch_function__`
- Optionally tracks FakeTensor operations, which was needed for propagating tensor meta as a step of sharding propagation
- Optionally tracks real tensor operations, including functional c10d op, and regular ops
- Calls are shown in the hierarchical structure! 
- shorthand representation
  - dt: DTesnor, ft: FakeTensor, t: Tensor
  - DM(2, 2) == DeviceMesh(shape = [2, 2])
  - [R, P, S(0)] == Placement[Replicate, Partial, Shard(0)]
  - f32[8,8] == float32 with shape[8, 8]

```
  debug_mode = DTensorDebugMode(record_faketensor=False, record_realtensor=True)
  with debug_mode:
      torch.mm(x_dtensor, y_dtensor)
  print(debug_mode.debug_string())
```
produces: 
```
  torch.mm(dt: f32[8, 8][S(0)], dt: f32[8, 32][S(0)])
    aten::mm(dt: f32[8, 8][S(0)], dt: f32[8, 32][S(0)])
      redistribute_input(1, [S(0)], [R])
        _c10d_functional::all_gather_into_tensor(t: f32[1, 32], 8, 0)
        _c10d_functional::wait_tensor(t: f32[8, 32])
      aten::mm(t: f32[1, 8], t: f32[8, 32])
```

Another example, for torch.einsum
```
  torch.functional.einsum(bld,dnh->blnh, dt: f32[16, 6, 8][P, R], dt: f32[8, 4, 4][R, P])
    aten::unsqueeze(dt: f32[16, 6, 8][P, R], 3)
      aten::unsqueeze(t: f32[16, 6, 8], 3)
    aten::unsqueeze(dt: f32[16, 6, 8, 1][P, R], 4)
      aten::unsqueeze(t: f32[16, 6, 8, 1], 4)
    aten::permute(dt: f32[16, 6, 8, 1, 1][P, R], [0, 1, 3, 4, 2])
      aten::permute(t: f32[16, 6, 8, 1, 1], [0, 1, 3, 4, 2])
    aten::unsqueeze(dt: f32[8, 4, 4][R, P], 3)
      aten::unsqueeze(t: f32[8, 4, 4], 3)
    aten::unsqueeze(dt: f32[8, 4, 4, 1][R, P], 4)
      aten::unsqueeze(t: f32[8, 4, 4, 1], 4)
    aten::permute(dt: f32[8, 4, 4, 1, 1][R, P], [3, 4, 1, 2, 0])
      aten::permute(t: f32[8, 4, 4, 1, 1], [3, 4, 1, 2, 0])
    aten::permute(dt: f32[16, 6, 1, 1, 8][P, R], [0, 1, 4, 2, 3])
      aten::permute(t: f32[16, 6, 1, 1, 8], [0, 1, 4, 2, 3])
    aten::view(dt: f32[16, 6, 8, 1, 1][P, R], [1, 96, 8])
      aten::view(t: f32[16, 6, 8, 1, 1], [1, 96, 8])
    aten::permute(dt: f32[1, 1, 4, 4, 8][R, P], [4, 2, 3, 0, 1])
      aten::permute(t: f32[1, 1, 4, 4, 8], [4, 2, 3, 0, 1])
    aten::view(dt: f32[8, 4, 4, 1, 1][R, P], [1, 8, 16])
      aten::view(t: f32[8, 4, 4, 1, 1], [1, 8, 16])
    aten::bmm(dt: f32[1, 96, 8][P, R], dt: f32[1, 8, 16][R, P])
      redistribute_input(0, [P, R], [S(2), S(2)])
        aten::chunk(t: f32[1, 96, 8], 4, 2)
        aten::cat(['t: f32[1, 96, 2]', 't: f32[1, 96, 2]', 't: f32[1, 96, 2]', 't: f32[1, 96, 2]'])
        _c10d_functional::reduce_scatter_tensor(t: f32[4, 96, 2], sum, 4, 2)
        aten::clone(t: f32[1, 96, 1])
      redistribute_input(1, [R, P], [S(1), S(1)])
        aten::chunk(t: f32[1, 8, 16], 4, 1)
        aten::clone(t: f32[1, 2, 16])
        aten::chunk(t: f32[1, 2, 16], 2, 1)
        aten::cat(['t: f32[1, 1, 16]', 't: f32[1, 1, 16]'])
        _c10d_functional::reduce_scatter_tensor(t: f32[2, 1, 16], sum, 2, 3)
        _c10d_functional::wait_tensor(t: f32[1, 1, 16])
      aten::bmm(t: f32[1, 96, 1], t: f32[1, 1, 16])
    aten::view(dt: f32[1, 96, 16][P, P], [16, 6, 1, 4, 4])
      aten::view(t: f32[1, 96, 16], [16, 6, 1, 4, 4])
    aten::permute(dt: f32[16, 6, 1, 4, 4][P, P], [0, 1, 3, 4, 2])
      aten::permute(t: f32[16, 6, 1, 4, 4], [0, 1, 3, 4, 2])
    aten::view(dt: f32[16, 6, 4, 4, 1][P, P], [16, 6, 4, 4])
      aten::view(t: f32[16, 6, 4, 4, 1], [16, 6, 4, 4])
```

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -3180,6 +3180,8 @@ coverage_ignore_classes = [
     "WeakIdKeyDictionary",
     "WeakIdRef",
     "WeakTensorKeyDictionary",
+    # torch.utils.debug_mode
+    "DebugMode",
 ]
 
 # The suffix(es) of source filenames.
--- a/test/distributed/tensor/debug/test_debug_mode.py
+++ b/test/distributed/tensor/debug/test_debug_mode.py
@@ -3,12 +3,12 @@
 import torch
 import torch.distributed as dist
 from torch.distributed.tensor import DeviceMesh, DTensor, Partial, Replicate, Shard
-from torch.distributed.tensor.debug import DTensorDebugMode
 from torch.testing._internal.common_utils import run_tests, TestCase
 from torch.testing._internal.distributed.fake_pg import FakeStore
+from torch.utils.debug_mode import DebugMode
 
 
-class TestCommMode(TestCase):
+class TestDTensorDebugMode(TestCase):
     def tearDown(self):
         super().tearDown()
         dist.destroy_process_group()
@@ -26,23 +26,23 @@ class TestCommMode(TestCase):
     def test_debug_mode_mm(self):
         mesh = DeviceMesh(self.device_type, list(range(self.world_size)))
 
-        debug_mode = DTensorDebugMode()
         x = torch.randn(1, 8, requires_grad=True)
         y = torch.randn(1, 32, requires_grad=True)
         x_dtensor = DTensor.from_local(x, mesh, [Shard(0)], run_check=False)
         y_dtensor = DTensor.from_local(y, mesh, [Shard(0)], run_check=False)
 
-        with debug_mode:
+        with DebugMode() as debug_mode:
             torch.mm(x_dtensor, y_dtensor)
 
         self.assertExpectedInline(
             debug_mode.debug_string(),
             """\
-  aten::mm(dt: f32[8, 8]DM(8)[S(0)], dt: f32[8, 32]DM(8)[S(0)])
-    redistribute_input(1, [S(0)], [R])
-      _c10d_functional::all_gather_into_tensor(t: f32[1, 32], 8, 0)
-      _c10d_functional::wait_tensor(t: f32[8, 32])
-    aten::mm(t: f32[1, 8], t: f32[8, 32])""",
+  torch.mm(dt: f32[8, 8][S(0)], dt: f32[8, 32][S(0)])
+    aten::mm(dt: f32[8, 8][S(0)], dt: f32[8, 32][S(0)])
+      redistribute_input(1, [S(0)], [R])
+        _c10d_functional::all_gather_into_tensor(t: f32[1, 32], 8, 0)
+        _c10d_functional::wait_tensor(t: f32[8, 32])
+      aten::mm(t: f32[1, 8], t: f32[8, 32])""",
         )
 
     def test_debug_mode_einsum(self):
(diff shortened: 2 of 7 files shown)


torch/distributed/tensor/_dispatch.py
```diff
@@ -336,6 +339,12 @@ def redistribute_local_args(
         suggested_input_schema: OpSchema,
         use_val_from_redistribute_schema: bool,
     ) -> None:
+        from torch.distributed.tensor.debug import DTensorDebugMode
+
+        debug_mode = _get_current_dispatch_mode()
+        in_debug_mode = isinstance(debug_mode, DTensorDebugMode)
+        redistribute_context = contextlib.nullcontext()
```
tianyu-l: maybe don't need to define it here

tianyu-l (commented):
Thanks! Looks reasonable to me.

Do you think there is a way to associate aten ops with python level code, i.e. the torch ops? For `torch.einsum` it's a bit difficult to associate all the squeeze / unsqueeze / permute / view / bmm with it.

Idk, is it possible combine `__torch_function__` and `__torch_dispatch__`?

test/distributed/tensor/debug/test_debug_mode.py
```diff
@@ -0,0 +1,104 @@
...
+        a_dt = DTensor.from_local(a, mesh, [Partial(), Replicate()], run_check=False)
+        b_dt = DTensor.from_local(b, mesh, [Replicate(), Partial()], run_check=False)
+
+        # Capture the operator decomposition
+        with DTensorDebugMode() as debug_mode:
+            torch.einsum("bld,dnh->blnh", a_dt, b_dt)
+
+        self.assertExpectedInline(
+          debug_mode.debug_string(), """\
+  aten::unsqueeze(dt: f32[16, 6, 8]DM(4, 2)[P, R], 3)
```
ezyang: In many cases the DTensor's device mesh is unambiguous (because there's only one) so I kind of wonder if we can elide it in that case; e.g., keep track of how many device meshes showed up and only start printing something about them the moment we see two distinct ones. Also cc @fduwjj

test/distributed/tensor/debug/test_debug_mode.py
```diff
@@ -0,0 +1,104 @@
...
+    aten::view(t: f32[16, 6, 8, 1, 1], [1, 96, 8])
+  aten::permute(dt: f32[1, 1, 4, 4, 8]DM(4, 2)[R, P], [4, 2, 3, 0, 1])
+    aten::permute(t: f32[1, 1, 4, 4, 8], [4, 2, 3, 0, 1])
+  aten::view(dt: f32[8, 4, 4, 1, 1]DM(4, 2)[R, P], [1, 8, 16])
+    aten::view(t: f32[8, 4, 4, 1, 1], [1, 8, 16])
+  aten::bmm(dt: f32[1, 96, 8]DM(4, 2)[P, R], dt: f32[1, 8, 16]DM(4, 2)[R, P])
+    redistribute_input(0, [P, R], [S(2), S(2)])
+      aten::chunk(t: f32[1, 96, 8], 4, 2)
+      aten::cat(['t: f32[1, 96, 2]', 't: f32[1, 96, 2]', 't: f32[1, 96, 2]', 't: f32[1, 96, 2]'])
+      _c10d_functional::reduce_scatter_tensor(t: f32[4, 96, 2], sum, 4, 2)
```
wconstab: Hm no wait tensor here..?

test/distributed/tensor/debug/test_debug_mode.py
```diff
@@ -0,0 +1,104 @@
...
+  aten::permute(dt: f32[16, 6, 1, 1, 8]DM(4, 2)[P, R], [0, 1, 4, 2, 3])
+    aten::permute(t: f32[16, 6, 1, 1, 8], [0, 1, 4, 2, 3])
+  aten::view(dt: f32[16, 6, 8, 1, 1]DM(4, 2)[P, R], [1, 96, 8])
+    aten::view(t: f32[16, 6, 8, 1, 1], [1, 96, 8])
+  aten::permute(dt: f32[1, 1, 4, 4, 8]DM(4, 2)[R, P], [4, 2, 3, 0, 1])
+    aten::permute(t: f32[1, 1, 4, 4, 8], [4, 2, 3, 0, 1])
+  aten::view(dt: f32[8, 4, 4, 1, 1]DM(4, 2)[R, P], [1, 8, 16])
+    aten::view(t: f32[8, 4, 4, 1, 1], [1, 8, 16])
+  aten::bmm(dt: f32[1, 96, 8]DM(4, 2)[P, R], dt: f32[1, 8, 16]DM(4, 2)[R, P])
+    redistribute_input(0, [P, R], [S(2), S(2)])
```
wconstab: Maybe an arrow -> instead of a comma to separate src,dst placements?

torch/distributed/tensor/debug/_comm_mode.py
```diff
@@ -733,3 +737,112 @@ def __torch_dispatch__(self, func, types, args=(), kwargs=None):
         ].append(operation_dict)
 
         return out
+
+
+def stringify_shape(shape) -> str:
+    return f"[{', '.join([str(x) for x in shape])}]"
+
+
+def stringify_device_mesh(mesh) -> str:
```
wconstab: Tricky to know how much to abbreviate here. Just printing the shape is definitely not enough to be thorough in an adversarial case but obviously it's nice to keep things as brief as possible.

Another way is to print DM=0 (0 being a tag you use to look up in the table) in the per-dtensor strings and then in a section below or above the dispatch trace print all the device meshes in a table with more detail

torch/distributed/tensor/_dispatch.py
SherlockNoMad: it's latter needed at line 367

test/distributed/tensor/debug/test_debug_mode.py
SherlockNoMad: Omitted DeviceMesh.

torch/distributed/tensor/_dispatch.py
```diff
@@ -336,6 +339,12 @@ def redistribute_local_args(
         suggested_input_schema: OpSchema,
         use_val_from_redistribute_schema: bool,
     ) -> None:
+        from torch.utils.debug_mode import DebugMode
+
+        debug_mode = _get_current_dispatch_mode()
```
wconstab, line 341: I'm a bit worried the overheads of all these checks and function calls will be significant in the case the debug mode is disabled. Any suggestions @swolchok

torch/utils/debug_mode.py
```diff
@@ -0,0 +1,160 @@
...
+            return NotImplemented
+        elif any(t == FakeTensor for t in types) or isinstance(
+            _get_current_dispatch_mode(), FakeTensorMode
+        ):
+            if self.record_faketensor:
+                if func not in {torch.ops.prim.device.default}:
+                    self.operators.append((func, args, kwargs, self.call_depth + 1))
+        elif len(types) == 0:
+            if self.record_realtensor:
+                self.operators.append((func, args, kwargs, self.call_depth + 1))
```
wconstab, line 125: Is it impossible to redispatch to a second torch dispatch? If not then how is call depth accumulated between redispatches?

test/distributed/tensor/debug/test_debug_mode.py
SherlockNoMad: right... 
the wait tensor only show up in the second reduce scatter. 

Could this be a bug?

torch/distributed/tensor/_dispatch.py
tianyu-l: oh I see, maybe nit: I think it's better to put this `nullcontext()` init closer to 367 if possible, e.g. in the `else` branch of `if in_debug_mode`

tianyu-l (commented):
> > torch_function

> great idea. I tried, and it works!
See updated summary. Now it can also record torch level functions.

Nice! Could you also show an example of backward?

torch/utils/debug_mode.py
```diff
@@ -0,0 +1,160 @@
...
+            return func(*args, **kwargs)
+        finally:
+            self.call_depth -= 1
+
+    def __torch_dispatch__(self, func, types, args=(), kwargs=None):
+        if kwargs is None:
+            kwargs = {}
+
+        # Record the operation with its call depth
+        if any(t == DTensor for t in types):
```
swolchok: isn't this `if DTensor in types:`

torch/utils/debug_mode.py
```diff
@@ -0,0 +1,160 @@
...
+            self.operators.append((func, args, kwargs, self.call_depth))
+            return NotImplemented
+        elif any(t == FakeTensor for t in types) or isinstance(
```
swolchok: ditto `in`

test/distributed/tensor/debug/test_debug_mode.py
wconstab: cc @zpcore can you take a look? I'm not sure if there is a valid reason for us to not be waiting on this one but I'd also be surprised if we were getting by without seeing correctness issues if we were missing a wait here

torch/distributed/tensor/_dispatch.py
swolchok: 1) can we really not do the import at top-level?
2) don't love the need for an isinstance check
3) since dispatch modes already involve a native function, we can accelerate this if needed with a specific get_debug_mode_if_active call that returns DebugMode or None
4) this will stop working if and when we get DTensor to stop using `__torch_dispatch__`, won't it?

test/distributed/tensor/debug/test_debug_mode.py
zpcore: Oh interesting, this maybe relevant to an issue I noticed recently that when I call `loss.backward()`, the code will complain 
```
Warning: At the time of process termination, there are still 20 unwaited collective calls. Please review your program to ensure that:
1. c10d_functional.wait_tensor() is invoked on all tensors returned from c10d_functional collective,
2. c10d_functional.wait_tensor() is invoked on all output tensors of async_op=True torch.distributed collective called under `with allow_inflight_collective_as_graph_input_ctx():`
```
Let me take a look.

test/distributed/tensor/debug/test_debug_mode.py
SherlockNoMad: done. 

```
      redistribute_input(0, [P, R] -> [S(2), S(2)])
```

torch/utils/debug_mode.py
SherlockNoMad: fixed.

test/distributed/tensor/debug/test_debug_mode.py
zpcore: Made a fix in https://github.com/pytorch/pytorch/pull/162749. I can see the wait() appear in the debug with that fix.

SherlockNoMad: glad that DebugMode help catching bugs. 

Just another idea, do you think it would be useful to test all redistributions under DebugMode?

torch/distributed/tensor/_dispatch.py
SherlockNoMad: 1. fixed.
2,3. Yes. If overhead is a concern, we can write a native function for it. The bigger questions is that, shall we make it an infra mode? 
4. yea. I am not sure about the plan for DTensor refactoring, would love to learn more.

torch/utils/debug_mode.py
SherlockNoMad: I tried tracking the redispatch, but there are some issues which I don't understand yet. 

So call depth accumulated doesn't work for redispatch.

test/distributed/tensor/debug/test_debug_mode.py
zpcore: I don't oppose if we enforce the wait() call for all tests under `test/distributed/tensor/test_redistribute.py` :)

zpcore: I feel I really need this PR to help with the debugging. Even though my PR above fixed the missing wait(), but in a different test I still see unmatched wait() warning when I call `loss.backward()` on DTensor module. Once the PR is merge, I can do further debugging.

torch/distributed/tensor/_dispatch.py
tianyu-l: > can we really not do the import at top-level?

I wonder how this was addressed? It seems `DebugMode` and `_get_current_dispatch_mode` are still imported at top level.

torch/utils/debug_mode.py
```diff
@@ -0,0 +1,168 @@
...
+    def __torch_dispatch__(self, func, types, args=(), kwargs=None):
+        if kwargs is None:
+            kwargs = {}
+
+        # Record the operation with its call depth
+        if dt.DTensor in types:
+            self.operators.append((func, args, kwargs, self.call_depth))
+            return NotImplemented
+        elif FakeTensor in types or isinstance(
+            _get_current_dispatch_mode(), FakeTensorMode
```
tianyu-l, line 118: Could you elaborate a bit more on the restrictions of orders of `FakeTensorMode` and `DebugMode`, and maybe possibly other modes?

E.g. is the code here assuming `FakeTensorMode` is applied inside `DebugMode`? Or in other words, is `_get_current_dispatch_mode` obtained from a stack or a queue?

test/distributed/tensor/debug/test_debug_mode.py
```diff
@@ -0,0 +1,173 @@
...
+  <method 'sum' of 'torch._C.TensorBase' objects>(t: f32[8, 8, 8])
+      aten::sum(t: f32[8, 8, 8])""",
+        )
+
+    def test_fake_tensor(self):
+        with FakeTensorMode():
+            x = torch.randn(8, 8)
+            y = torch.randn(8, 8, 8)
+
+        with DebugMode(record_faketensor=True) as debug_mode:
```
tianyu-l, line 183: Could you add cases where the two modes are nested? E.g. we can explicit error out if some nesting is not supported.

tianyu-l (commented):
No big concern from me. Had a few more questions regarding interaction with `FakeTensorMode`.

Will let @ezyang stamp.

ezyang (approved):

test/distributed/tensor/debug/test_debug_mode.py
SherlockNoMad: added a test case for Nested TorchDispatchMode. AFAIK, it compose well with other dispatch mode, no error are thrown.

torch/utils/debug_mode.py
SherlockNoMad: I don't think there is a restriction here. 
FakeTensorMode is a special "infra mode", which always run below user-defined TorchDispatchMode and User-defined TensorSubclass (Dtensor in this case) . See https://www.internalfb.com/excalidraw/EX225856 for detailed dispatching order

This grantees that DTensor's torch_dispatch is always run first, and the FakeTensorMode needed under DTensor sharding propagation is triggered latter, and finally real tensor is ran.

torch/distributed/tensor/_dispatch.py
SherlockNoMad: IIUC, the original concern here was doing lazy import has perf impact, So I have move the import to the beginning. 
In util/debug_mode.py, I avoided import DTensor, instead I write torch.distributed.tensor.DTensor. 

This is a just a trick to avoid circular import.

wanchaol (commented):
Do you have plans to migrate CommDebugMode to use this later?

docs/source/conf.py
```diff
@@ -3180,6 +3180,8 @@
     "WeakIdKeyDictionary",
     "WeakIdRef",
     "WeakTensorKeyDictionary",
+    # torch.utils.debug_mode
+    "DebugMode",
```
albanD, line 3184: @SherlockNoMad no new entry should EVER be added to this file.
If it is here, it means that it is not actually showing up on the website.

PR #162666, opened 2025-09-10
Cache most used constants in Dynamo

guilhermeleobas:
* __->__ #162666

~This PR adds a cache for some constants like small integers, latin-1
one-char strings, True, False, None, NotImplemented. CPython caches a
bit more constants than this, which we might want to add in the future.
This list can also be extended to constants that are more common in ML
workloads.~

Caches some of the most used constants:
* `True`
* `False`
* `None`
* `NotImplemented`

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @kadeng @chauhang @amjames @Lucaskabela @jataylo


guilhermeleobas:
This is now 0.3 seconds faster on [auroflow](https://github.com/pytorch/pytorch/pull/155866)

guilhermeleobas:
This reduces the runtime of `test_itertools::TestBasicOps.test_ziplongest` from 101s -> 95s.

test/dynamo/cpython/3_13/test_sort.py
```diff
@@ -183,6 +183,7 @@ def bad_key(x):
             x = [e for e, i in augmented] # a stable sort of s
             check("stability", x, s)
 
+    @skipIfTorchDynamo("slow test")
```
Lucaskabela, line 186: Does this decorator cause the expected fail?

torch/_dynamo/variables/dicts.py
```diff
@@ -1394,8 +1394,8 @@ def reconstruct(self, codegen: "PyCodegen"):
 
     def call_obj_hasattr(self, tx, name):
         if name in self.python_type().__dict__:
```
Lucaskabela: Can we fold this into one line similar to the other bool expressions?

Lucaskabela (approved):
Minor nit but generally LGTM! I wonder if we can get a cprof of a simple program before/after this change to validate the speedup more granularly

test/dynamo/cpython/3_13/test_sort.py
guilhermeleobas: Yes! I can re-enable the test if you want. It is just that this test is kinda slow to run. I can also reduce the inputs.

torch/_dynamo/variables/user_defined.py
```diff
@@ -1989,7 +1989,7 @@ def call_method(self, tx: "InstructionTranslator", method_name, args, kwargs):
             if self.idx != self.REMOVED:
                 tx.output.side_effects.remove_hook(self.idx)
                 self.idx = self.REMOVED
-            return variables.ConstantVariable.create(None)
+            return variables.constant_none
```
zou3519, line 1992: Here's what i want to test for. If someone forgets to use variables.constant_none, and instead uses ConstantVariable.create(None), then I want this to not cause any problems. Is it possible to write a test for it?

Alternatively you can write a lint rule that says "you're not allowed to have ConstantVariable.create(None)" in PyTorch code, but I don't know how easy this is.

guilhermeleobas: Maybe it doesn't worth the hassle for such a small gain

PR #162673, opened 2025-09-11
[Triton] [Inductor] Pruned failed compilations from Autotuning candidates

njriasan:
Summary:
When exahaustively autotuning a new template you may hit situations that lead to compilation failures. This template will still attempt to autotune because nothing was marking this as failed and in my experiments lead to a crash/segfault if I didn't set `TORCHINDUCTOR_AUTOTUNE_IN_SUBPROC=1`.

To help eliminate this issue this PR marks any template that fails to compile as "failed" and then removes all of the failed templates from the choice candidates. In the case where it would have just failed to compile twice, this should at least reduce compilation time.

Test Plan:
Tested locally when experminenting with the new blackwell templates and a Triton version that contains a bug related to `num_warps < 4`.

Rollback Plan:

Differential Revision: D82172207

### DIFF
--- a/torch/_inductor/select_algorithm.py
+++ b/torch/_inductor/select_algorithm.py
@@ -2421,16 +2421,19 @@ class AlgorithmSelectorCache(PersistentCache):
             N = input_nodes[-1].get_size()[-1]
             append_to_log(mm_file_name, {"invoke": str((M, K, N))})
 
-        if len(choices) == 0:
+        def create_no_valid_choices() -> NoValidChoicesError:
             backend_config = (
                 "max_autotune_gemm_backends"
                 if name != "convolution"
                 else "max_autotune_conv_backends"
             )
-            raise NoValidChoicesError(
+            return NoValidChoicesError(
                 f"No choices to select, please consider adding ATEN into {backend_config} "
                 "config (defined in torch/_inductor/config.py) to allow at least one choice. "
             )
+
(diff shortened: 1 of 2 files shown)


PaulZhang12 (approved):
This is an awesome fix! Thank you, cc @eellison for any thoughts

torch/_inductor/select_algorithm.py
```diff
@@ -2421,16 +2421,19 @@ def __call__(
...
 
-        if len(choices) == 0:
+        def create_no_valid_choices() -> NoValidChoicesError:
             backend_config = (
                 "max_autotune_gemm_backends"
                 if name != "convolution"
                 else "max_autotune_conv_backends"
             )
-            raise NoValidChoicesError(
+            return NoValidChoicesError(
```
mlazos, line 2430: It might be good to modify this to indicate the reason for no valid choices, I think this could be a good idea (e.g. no compileable choices vs no choices at the beginning)

I had some plans to update this since this is the most common error I've seen with users by far. They usually end up adding aten, but it would be useful to know why.

mlazos (approved):

torch/_inductor/select_algorithm.py
njriasan: Happy to submit a followup. Thanks for the suggestion.

PR #162680, opened 2025-09-11
[SymmMem] Fix memory allocation hold-up

kwen2501:
* #162681
* __->__ #162680
* #163298

Problem:
Without MemPool it looks like nvshmem backend never deallocates memory.

Cause:
Handles in `symm_mems_` (a map) keeps reference to memory allocations.

Solution:
- Remove reference to allocation from handles -- the reference is never used anyway.
- Use `unique_ptr` instead of `shared_ptr` to wrap allocation to ensure single ownership.

### DIFF
--- a/torch/csrc/distributed/c10d/symm_mem/NVSHMEMSymmetricMemory.cu
+++ b/torch/csrc/distributed/c10d/symm_mem/NVSHMEMSymmetricMemory.cu
@@ -51,7 +51,7 @@ struct NVSHMEMAllocation {
 class NVSHMEMPeerAllocInfo : public c10::intrusive_ptr_target {
  public:
   NVSHMEMPeerAllocInfo(
-      std::shared_ptr<NVSHMEMAllocation> allocation,
+      const std::unique_ptr<NVSHMEMAllocation>& allocation,
       const std::string& group_name)
       : base_ptr_(allocation->ptr),
         buffer_size_(allocation->buffer_size) {
@@ -144,10 +144,9 @@ class NVSHMEMPeerAllocInfo : public c10::intrusive_ptr_target {
 class NVSHMEMSymmetricMemory : public SymmetricMemory {
  public:
   NVSHMEMSymmetricMemory(
-      std::shared_ptr<NVSHMEMAllocation> allocation,
+      const std::unique_ptr<NVSHMEMAllocation>& allocation,
       const std::string& group_name)
-      : allocation_(allocation),
-        device_idx_(allocation->device_idx),
+      : device_idx_(allocation->device_idx),
         group_name_(group_name) {
     // A handle stores two types of info:
     // (i) allocation's base ptrs and base signal pads, ours and peers'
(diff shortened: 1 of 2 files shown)


torch/csrc/distributed/c10d/symm_mem/NVSHMEMSymmetricMemory.cu
```diff
@@ -51,7 +51,7 @@ struct NVSHMEMAllocation {
 class NVSHMEMPeerAllocInfo : public c10::intrusive_ptr_target {
  public:
   NVSHMEMPeerAllocInfo(
-      std::shared_ptr<NVSHMEMAllocation> allocation,
+      const std::unique_ptr<NVSHMEMAllocation>& allocation,
```
ezyang: This makes me nervous. Previously the code was owning, now it is non-owning. How are you reasoning about the lifetimes, aka why is use after free not possible? In particular, you DO record `base_ptr_` derived off of allocation. Additionally, if you need non-owning pointer to NVSHMEMAllocation, it's more classic to pass a `const NVSHMEMAllocation*` here (there is no reason to pass the unique pointer by reference as you aren't doing something like moving it or similar.)

kwen2501: I want to enforce that the Allocation is owned by a single place instead of sharable. Thus unique_ptr. 
And yeah, I don't need transfer, agree that `NVSHMEMAllocation*` works better here.

kwen2501: Fixed

torch/csrc/distributed/c10d/symm_mem/NVSHMEMSymmetricMemory.cu
```diff
@@ -51,7 +51,7 @@ struct NVSHMEMAllocation {
 class NVSHMEMPeerAllocInfo : public c10::intrusive_ptr_target {
  public:
   NVSHMEMPeerAllocInfo(
-      std::shared_ptr<NVSHMEMAllocation> allocation,
+      NVSHMEMAllocation* allocation,
```
Skylion007, line 54: Why use bare pointers here? We should use unique_ptr here and just use `std::move` with the member intiializers

kwen2501: I will let you battle with @ezyang lol, see discussion above.
tldr:
- there is no ownership move here.
- Passing a bare pointer just for reading some values in the object.

torch/csrc/distributed/c10d/symm_mem/NVSHMEMSymmetricMemory.cu
ngimel: I don't think lifetime story is completely clean yet, but we'll address that in a future refactor

torch/csrc/distributed/c10d/symm_mem/NVSHMEMSymmetricMemory.cu
ezyang: no ownership move means don't std::move

ezyang (approved):
suuure

PR #162681, opened 2025-09-11
[SymmMem] Add `num_active_allocations` and dealloc checks

kwen2501:
* __->__ #162681
* #162680
* #163298

As titled.
`del tensor` should cause `num_active_allocations` to go down to 0.

### DIFF
--- a/test/distributed/test_nvshmem.py
+++ b/test/distributed/test_nvshmem.py
@@ -7,6 +7,7 @@
 import torch
 import torch.distributed as dist
 import torch.distributed._symmetric_memory as symm_mem
+from torch._C._distributed_c10d import _SymmetricMemory
 from torch.distributed.device_mesh import init_device_mesh
 from torch.testing._internal.common_distributed import (
     MultiProcContinuousTest,
@@ -66,6 +67,29 @@ class NVSHMEMSymmetricMemoryTest(MultiProcContinuousTest):
 
         out = symm_mem.empty(numel, dtype=dtype, device=self.device)
         symm_mem.rendezvous(out, group=group_name)
+
+    @skipIfRocm
+    def test_dealloc(self) -> None:
+        self._init_device()
+        group_name = dist.group.WORLD.group_name
+        symm_mem.enable_symm_mem_for_group(group_name)
+
+        dtype = torch.float
+        numel = 1024
+
+        def foo():
+            inp = symm_mem.empty(numel, dtype=dtype, device=self.device)
+            symm_mem.rendezvous(inp, group=group_name)
+
+        foo()
+        self.assertEqual(_SymmetricMemory.num_active_allocations(self.device), 0)
+
+        t0 = symm_mem.empty(numel, dtype=dtype, device=self.device)
+        symm_mem.rendezvous(t0, group=group_name)
+        t1 = symm_mem.empty(numel, dtype=dtype, device=self.device)
+        symm_mem.rendezvous(t1, group=group_name)
+        del t0, t1
+        self.assertEqual(_SymmetricMemory.num_active_allocations(self.device), 0)
 
     @skipIfRocm
     def test_alloc_without_device_context(self) -> None:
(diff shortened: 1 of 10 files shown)


test/distributed/test_nvshmem.py
```diff
@@ -67,6 +68,29 @@ def foo():
         out = symm_mem.empty(numel, dtype=dtype, device=self.device)
         symm_mem.rendezvous(out, group=group_name)
 
+    @skipIfRocm
+    def test_dealloc(self) -> None:
```
ngimel: this test should test all symm mem backends

test/distributed/test_nvshmem.py
```diff
@@ -67,6 +68,29 @@ def foo():
...
+            inp = symm_mem.empty(numel, dtype=dtype, device=self.device)
+            symm_mem.rendezvous(inp, group=group_name)
+
+        foo()
+        self.assertEqual(_SymmetricMemory.num_active_allocations(self.device), 0)
+
+        t0 = symm_mem.empty(numel, dtype=dtype, device=self.device)
+        symm_mem.rendezvous(t0, group=group_name)
+        t1 = symm_mem.empty(numel, dtype=dtype, device=self.device)
+        symm_mem.rendezvous(t1, group=group_name)
```
ngimel, line 73: do you also want to test num_active_allocations here? (should be nonzero)

test/distributed/test_nvshmem.py
kwen2501: Added. 
Eventually we should put all common tests in `test_symmetric_memory.py` and all backends should be able to pass those tests. I am working on it (via a separate PR).

test/distributed/test_nvshmem.py
kwen2501: Added.

test/distributed/test_symmetric_memory.py
```diff
@@ -87,6 +87,33 @@ def test_cuda_nvlink_connectivity_detection(self) -> None:
...
+            self.assertGreater(_SymmetricMemory.num_active_allocations(self.device), 0)
+
+        foo()
+        self.assertEqual(_SymmetricMemory.num_active_allocations(self.device), 0)
+
+        t0 = symm_mem.empty(numel, dtype=dtype, device=self.device)
+        symm_mem.rendezvous(t0, group=group_name)
+        t1 = symm_mem.empty(numel, dtype=dtype, device=self.device)
+        symm_mem.rendezvous(t1, group=group_name)
+        self.assertGreater(_SymmetricMemory.num_active_allocations(self.device), 0)
```
fduwjj, line 112: why don't we check the actual number I mean it is not 2 here?

test/distributed/test_nvshmem.py
```diff
@@ -61,11 +62,19 @@ def test_alloc(self) -> None:
...
         foo()
+        self.assertEqual(_SymmetricMemory.num_active_allocations(self.device), 0)
 
-        out = symm_mem.empty(numel, dtype=dtype, device=self.device)
-        symm_mem.rendezvous(out, group=group_name)
+        t0 = symm_mem.empty(numel, dtype=dtype, device=self.device)
+        symm_mem.rendezvous(t0, group=group_name)
+        t1 = symm_mem.empty(numel, dtype=dtype, device=self.device)
+        symm_mem.rendezvous(t1, group=group_name)
+        self.assertGreater(_SymmetricMemory.num_active_allocations(self.device), 0)
```
fduwjj, line 74: Same as my question below, what is the actual number, why only check larger than 0?

fduwjj (approved):
I have some questions for test and unblock for now

test/distributed/test_nvshmem.py
```diff
@@ -61,11 +62,19 @@ def test_alloc(self) -> None:
...
+
+        del t0, t1
```
ngimel, line 76: if we are implicitly using mempool, `del t0, t1` should not actually delete allocations?

PR #162690, opened 2025-09-11
[CuTe] Change the logic of pycute manipulation ops like coalesce, complement from co-lex to lex

fduwjj:
* #161106
* #161016
* __->__ #162690
* #162414
* #162534
* #162413

PyTorch tensor iteration (.view, contiguous, broadcasting) and NumPy array indexing all follow lexicographic (row-major) order. In Lexicographic (lex) on (i0, i1, …, i{k-1}): the leftmost index(stride is larger) changes fastest and the rightmost index changes slowest and usually last dim is contiguous.

However original pycute is all based on co-lex, after porting their code into pytorch and some cosmetic change, we now make it lex so that we can use it for use cases like device mesh internal bookkeeping and other stuff as well.

Changes included in this PR:
1. We changes all API ported in, included prefix_product(stride inferring and rename it to suffix_product), idx2crd, crd2idx, coalesce, composition, complement, right_inverse and left_inverse to make sure they are working in the lex way.
2. Added more unit test cases for some API mentioned above since existing unit tests do not have full coverage.
3. One bug fix inside composition, which will lead to infinite recursive call.

### DIFF
--- a/test/distributed/_pycute/test_coalesce.py
+++ b/test/distributed/_pycute/test_coalesce.py
@@ -47,11 +47,14 @@ _LOGGER = logging.getLogger(__name__)
 
 
 class TestCoalesce(TestCase):
-    def helper_test_coalesce(self, layout):
+    def helper_test_coalesce(self, layout, coalesced_layout=None):
         layoutR = coalesce(layout)
 
         _LOGGER.debug(f"{layout}  =>  {layoutR}")
 
+        if coalesced_layout:
+            self.assertEqual(coalesced_layout.shape, layoutR.shape)
+            self.assertEqual(coalesced_layout.stride, layoutR.stride)
         self.assertEqual(size(layoutR), size(layout))
 
         for i in range(size(layout)):
@@ -82,17 +85,27 @@ class TestCoalesce(TestCase):
         layout = Layout((2, (4, 6)))
         self.helper_test_coalesce(layout)
 
+        layout = Layout((1, 2), (8, 1))
+        coalesced_layout = Layout(2, 1)
+        self.helper_test_coalesce(layout, coalesced_layout)
+
         layout = Layout((2, 4), (4, 1))
-        self.helper_test_coalesce(layout)
+        coalesced_layout = Layout(8, 1)
+        self.helper_test_coalesce(layout, coalesced_layout)
 
         layout = Layout((2, 4, 6), (24, 6, 1))
-        self.helper_test_coalesce(layout)
+        coalesced_layout = Layout(48, 1)
+        self.helper_test_coalesce(layout, coalesced_layout)
 
         layout = Layout((2, 1, 3), (2, 4, 4))
         self.helper_test_coalesce(layout)
 
         layout = Layout(((2, 2), (2, 2)), ((1, 4), (8, 32)))
         self.helper_test_coalesce(layout)
+
+        layout = Layout(((2, 2), (2, 2)), ((32, 8), (4, 1)))
+        coalesced_layout = Layout((2, 4, 2), (32, 4, 1))
+        self.helper_test_coalesce(layout, coalesced_layout)
 
 
 if __name__ == "__main__":
--- a/test/distributed/_pycute/test_composition.py
+++ b/test/distributed/_pycute/test_composition.py
@@ -51,11 +51,13 @@ class TestComposition(TestCase):
         layoutR = composition(layoutA, layoutB)
 
         _LOGGER.debug(f"{layoutA} o {layoutB}  =>  {layoutR}")
+        print(layoutR.shape, layoutR.stride, size(layoutR))
 
         # True post-condition: Every coordinate c of layoutB with L1D(c) < size(layoutR) is a coordinate of layoutR.
 
         # Test that R(c) = A(B(c)) for all coordinates c in layoutR
         for i in range(size(layoutR)):
+            print(i, layoutR(i), layoutB(i), layoutA(layoutB(i)))
             self.assertEqual(layoutR(i), layoutA(layoutB(i)))
 
     def test_composition(self):
@@ -208,9 +210,24 @@ class TestComposition(TestCase):
         layoutB = Layout((6), (1))
         self.helper_test_composition(layoutA, layoutB)
 
+        # Pre-coalesced RHS
+        layoutA = Layout((8, 6, 4), (7, 4, 1))
+        layoutB = Layout((6), (1))
+        self.helper_test_composition(layoutA, layoutB)
+
+        # Case when not meet stride divisibility condition
+        with self.assertRaises(AssertionError):
+            layoutA = Layout((4, 6, 8, 10), (2, 3, 5, 7))
+            layoutB = Layout(6, 12)
+            self.helper_test_composition(layoutA, layoutB)
+
         # Mid-layout truncation
-        layoutA = Layout((4, 6, 8, 10), (2, 3, 5, 7))
+        layoutA = Layout((10, 8, 6, 4), (7, 5, 3, 2))
         layoutB = Layout(6, 12)
+        self.helper_test_composition(layoutA, layoutB)
+
+        layoutA = Layout((4,), (3,))
+        layoutB = Layout((6,), (2,))
         self.helper_test_composition(layoutA, layoutB)
 
 
--- a/test/distributed/_pycute/test_composition.py
+++ b/test/distributed/_pycute/test_composition.py
@@ -51,13 +51,11 @@ class TestComposition(TestCase):
         layoutR = composition(layoutA, layoutB)
 
         _LOGGER.debug(f"{layoutA} o {layoutB}  =>  {layoutR}")
-        print(layoutR.shape, layoutR.stride, size(layoutR))
 
         # True post-condition: Every coordinate c of layoutB with L1D(c) < size(layoutR) is a coordinate of layoutR.
 
         # Test that R(c) = A(B(c)) for all coordinates c in layoutR
         for i in range(size(layoutR)):
-            print(i, layoutR(i), layoutB(i), layoutA(layoutB(i)))
             self.assertEqual(layoutR(i), layoutA(layoutB(i)))
 
     def test_composition(self):
--- a/test/distributed/_pycute/test_int_tuple.py
+++ b/test/distributed/_pycute/test_int_tuple.py
@@ -70,16 +70,155 @@ class TestIntTuple(TestCase):
     def test_prefix_product(self):
         self.assertEqual(prefix_product(2), 1)
 
-        self.assertEqual(prefix_product((3, 2)), (1, 3))
+        self.assertEqual(prefix_product((3, 2)), (2, 1))
 
-        self.assertEqual(prefix_product((3, 2, 4)), (1, 3, 6))
+        self.assertEqual(prefix_product((3, 2, 4)), (8, 4, 1))
 
-        self.assertEqual(prefix_product(((2, 3), 4)), ((1, 2), 6))
+        self.assertEqual(prefix_product(((2, 3), 4)), ((12, 4), 1))
 
         self.assertEqual(
             prefix_product(((2, 3), (2, 1, 2), (5, 2, 1))),
-            ((1, 2), (6, 12, 12), (24, 120, 240)),
+            ((120, 40), (20, 20, 10), (2, 1, 1)),
         )
+
+    def test_crd2idx_basic(self):
+        # Test basic int/int case
+        self.assertEqual(crd2idx(2, 5, 1), 2)
+        self.assertEqual(crd2idx(0, 5, 1), 0)
+        self.assertEqual(crd2idx(4, 5, 1), 4)
+
+        # Test with custom stride
+        self.assertEqual(crd2idx(2, 5, 3), 6)
+        self.assertEqual(crd2idx(1, 5, 3), 3)
+
+    def test_crd2idx_tuple(self):
+        # Test tuple coordinates with default stride
+        self.assertEqual(crd2idx((1, 2), (3, 4)), 6)  # 1*4 + 2*1 = 6
+        self.assertEqual(crd2idx((0, 0), (3, 4)), 0)
+        self.assertEqual(crd2idx((2, 3), (3, 4)), 11)  # 2*4 + 3*1 = 11
+
+        # Test with custom stride
+        self.assertEqual(crd2idx((1, 2), (3, 4), (8, 2)), 12)  # 1*8 + 2*2 = 12
+
+        # Test 3D case
+        self.assertEqual(crd2idx((1, 0, 2), (2, 3, 4)), 14)  # 1*12 + 0*4 + 2*1 = 14
+
+    def test_crd2idx_none(self):
+        # Test None coordinate (should default to 0)
+        self.assertEqual(crd2idx(None, 5), 0)
+        self.assertEqual(crd2idx(None, (3, 4)), 0)
+
+    def test_crd2idx_int_with_tuple_shape(self):
+        # Test single integer coordinate with multi-dimensional shape and stride
+        # When crd is int and shape is tuple, it converts the int to multi-dim coordinate first
+        self.assertEqual(crd2idx(0, (2, 2), (2, 1)), 0)  # 0 -> (0,0) -> 0*2 + 0*1 = 0
+        self.assertEqual(crd2idx(1, (2, 2), (2, 1)), 1)  # 1 -> (0,1) -> 0*2 + 1*1 = 1
+        self.assertEqual(crd2idx(2, (2, 2), (2, 1)), 2)  # 2 -> (1,0) -> 1*2 + 0*1 = 2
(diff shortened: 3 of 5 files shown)


test/distributed/_pycute/test_coalesce.py
```diff
@@ -47,11 +47,14 @@
 
 
 class TestCoalesce(TestCase):
-    def helper_test_coalesce(self, layout):
+    def helper_test_coalesce(self, layout, coalesced_layout=None):
         layoutR = coalesce(layout)
 
         _LOGGER.debug(f"{layout}  =>  {layoutR}")
 
+        if coalesced_layout:
```
fduwjj, line 55: previously the unit test only test both coalesced layout and original one will have the same coordinate and index mapping. We added the actual shape and stride check here for some coalesced case.

test/distributed/_pycute/test_composition.py
```diff
@@ -208,11 +210,26 @@ def test_composition(self):
...
+            layoutB = Layout(6, 12)
+            self.helper_test_composition(layoutA, layoutB)
+
         # Mid-layout truncation
-        layoutA = Layout((4, 6, 8, 10), (2, 3, 5, 7))
+        layoutA = Layout((10, 8, 6, 4), (7, 5, 3, 2))
         layoutB = Layout(6, 12)
         self.helper_test_composition(layoutA, layoutB)
 
+        layoutA = Layout((4,), (3,))
```
fduwjj, line 227: this case previous code will run into indefinitely recursion, lol.

test/distributed/_pycute/test_composition.py
```diff
@@ -208,11 +210,26 @@ def test_composition(self):
...
+        with self.assertRaises(AssertionError):
+            layoutA = Layout((4, 6, 8, 10), (2, 3, 5, 7))
+            layoutB = Layout(6, 12)
+            self.helper_test_composition(layoutA, layoutB)
+
         # Mid-layout truncation
-        layoutA = Layout((4, 6, 8, 10), (2, 3, 5, 7))
+        layoutA = Layout((10, 8, 6, 4), (7, 5, 3, 2))
         layoutB = Layout(6, 12)
         self.helper_test_composition(layoutA, layoutB)
```
fduwjj, line 225: This is the only case we need to change unit test because after changing to lexicographic order, the devisability condition now is not valid

test/distributed/_pycute/test_int_tuple.py
```diff
@@ -70,17 +70,156 @@ def test_shape_div(self):
...
         self.assertEqual(prefix_product(2), 1)
 
-        self.assertEqual(prefix_product((3, 2)), (1, 3))
+        self.assertEqual(prefix_product((3, 2)), (2, 1))
 
-        self.assertEqual(prefix_product((3, 2, 4)), (1, 3, 6))
+        self.assertEqual(prefix_product((3, 2, 4)), (8, 4, 1))
 
-        self.assertEqual(prefix_product(((2, 3), 4)), ((1, 2), 6))
+        self.assertEqual(prefix_product(((2, 3), 4)), ((12, 4), 1))
```
fduwjj: this is expected for co-lexicographic to lexicographic.

test/distributed/_pycute/test_int_tuple.py
```diff
@@ -70,17 +70,156 @@ def test_shape_div(self):
...
 
         self.assertEqual(
             prefix_product(((2, 3), (2, 1, 2), (5, 2, 1))),
-            ((1, 2), (6, 12, 12), (24, 120, 240)),
+            ((120, 40), (20, 20, 10), (2, 1, 1)),
         )
 
+    def test_crd2idx_basic(self):
```
fduwjj, line 84: previously there are no tests for crd2idx and idx2crd. Let's add them and they are in lexicographic order.

torch/distributed/_pycute/layout.py
```diff
@@ -230,35 +237,49 @@ def composition(layoutA: Layout, layoutB: LayoutInput) -> Layout:
                 (layoutA[i] for i in range(len(layoutB), len(layoutA))),
             )
         )
-    elif is_tuple(layoutB.shape):  # type: ignore[union-attr]
+    elif is_tuple(layoutB.shape) and len(layoutB.shape) > 1:  # type: ignore[union-attr]
```
fduwjj: when length == 1 original logic will run into infinite recursive loop, lol.

torch/distributed/_pycute/layout.py
```diff
@@ -290,6 +306,10 @@ def complement(layout: LayoutOrIntTuple, max_idx: int = 1) -> Layout:
 
     result_shape.append((max_idx + current_idx - 1) // current_idx)  # ceil_div
     result_stride.append(current_idx)
+    # This is different from original pycute implementation, because we want to follow the lexicographic order here
+    # where the right-most dimension is the innermost dimension (smallest stride).
+    result_shape.reverse()
+    result_stride.reverse()
```
fduwjj, line 319: this kind of reverse is to make review easy. More efficiently algorithm wise is to first allocate a list with a fixed length first and then O(1) access. But overall it is still O(n) anyway, so I don't do it that way.

torch/distributed/_pycute/int_tuple.py
```diff
@@ -126,18 +126,24 @@ def shape_div(a: IntTuple, b: IntTuple) -> IntTuple:
             return (a + b - 1) // b
 
 
-# Exclusive prefix product with output congruent to input a
+# Exclusive prefix product with output congruent to input a (lexicographic)
```
ezyang: It's not a prefix product anymore right? It's a suffix_product now

ezyang: It's interesting that you decided to reverse the output. So it's also like a reversed suffix product

torch/distributed/_pycute/int_tuple.py
```diff
@@ -161,7 +167,15 @@ def idx2crd(
     else:
         if is_tuple(shape) and is_tuple(stride):  # "int" tuple tuple
             assert len(shape) == len(stride)
-            return tuple(idx2crd(idx, s, d) for s, d in zip(shape, stride))
+            # Process from left to right for lexicographic ordering (opposite of crd2idx)
```
ezyang: hmm? The old code processed left to right too??

ezyang: I think you meant to say "right to left" here.

ezyang: But you don't actually process right-to-left. It's probably dramatically simpler to go right-to-left

torch/distributed/_pycute/layout.py
```diff
@@ -166,25 +167,27 @@ def coalesce(layout: Layout, profile: LayoutProfile = None) -> Layout:
...
+    result_shape: list[int] = []
+    result_stride: list[int] = []
     for shape, stride in zip(flatten(layout.shape), flatten(layout.stride)):
         # skip their shape-1s
         if shape == 1:
             continue
-        # replace our shape-1 with anything
-        elif result_shape[-1] == 1:
-            result_shape[-1] = shape
-            result_stride[-1] = stride
```
ezyang: Why did we lose this case?

torch/distributed/_pycute/layout.py
```diff
@@ -166,25 +167,27 @@ def coalesce(layout: Layout, profile: LayoutProfile = None) -> Layout:
...
         # merge modes if the shape*stride match
-        elif result_shape[-1] * result_stride[-1] == stride:
-            result_shape[-1] = result_shape[-1] * shape
+        elif result_stride and result_stride[-1] == shape * stride:
+            result_stride.pop()
+            prev_shape = result_shape.pop()
+            result_shape.append(shape * prev_shape)
+            result_stride.append(stride)
```
ezyang: Yeah, it doesn't feel like it should look like this. Instead, it feels like we should have processed shape, stride in reversed order at the loop above

torch/distributed/_pycute/layout.py
```diff
@@ -230,35 +233,56 @@ def composition(layoutA: Layout, layoutB: LayoutInput) -> Layout:
                 (layoutA[i] for i in range(len(layoutB), len(layoutA))),
             )
         )
-    elif is_tuple(layoutB.shape):
+    elif is_tuple(layoutB.shape) and len(layoutB.shape) > 1:
```
ezyang: This extra condition doesn't seem warranted by the swap

torch/distributed/_pycute/layout.py
```diff
@@ -230,35 +233,56 @@ def composition(layoutA: Layout, layoutB: LayoutInput) -> Layout:
...
 
     if layoutB.stride == 0:
         return Layout(layoutB.shape, 0)
     else:
         result_shape = []
         result_stride = []
-        rest_shape = layoutB.shape
-        rest_stride = layoutB.stride
+        rest_shape = layoutB.shape if is_int(layoutB.shape) else layoutB.shape[0]  # type: ignore[union-attr,index]
+        rest_stride = layoutB.stride if is_int(layoutB.stride) else layoutB.stride[0]  # type: ignore[union-attr,index]
```
ezyang: This also doesn't seem warranted

torch/distributed/_pycute/layout.py
```diff
@@ -327,7 +357,7 @@ def left_inverse(layout: Optional[LayoutOrIntTuple]) -> Optional[Layout]:
         return None
     elif is_int(layout):
         return Layout(layout)
-    return right_inverse(make_layout(layout, complement(layout)))  # type: ignore[arg-type]
+    return right_inverse(make_layout(complement(layout), layout))  # type: ignore[arg-type]
```
ezyang, line 358: What's going on here?

torch/distributed/_pycute/layout.py
```diff
@@ -318,6 +346,8 @@ def right_inverse(layout: Optional[LayoutOrIntTuple]) -> Optional[Layout]:
         result_stride.append(rstride)
         current_idx = shape * stride
 
+    result_shape.reverse()
+    result_stride.reverse()
```
ezyang, line 348: This is also surprising

torch/distributed/_pycute/layout.py
```diff
@@ -230,35 +233,56 @@ def composition(layoutA: Layout, layoutB: LayoutInput) -> Layout:
...
+        # for integral s and d means that we want:
+        # (1) “remove” the first d elements from left, starting from rightmost. (This will increase the stride.)
+        # (2) “keep” the first s of those strided elements. (This does not affect the stride.)
+        # For example, if self = (6,2):(2,1), layout = (3:2)
+        # Step 1: remove the first 2 elements from self with stride increase, i.e., (6,2):(2,1) -> (6,1):(2,2)
+        # Step 2: keep the first 3 of those strided elements, i.e., (6,1):(2,2) -> (3,1):(2,2)
+        # Because we are going lexicographically, we go through left layout from right to left.
+        for i in range(len(flat_A_shapes) - 1, 0, -1):
+            curr_shape = flat_A_shapes[i]
+            curr_stride = flat_A_strides[i]
```
ezyang: This would be more symmetric with the original as `for curr_shape, curr_stride in reversed(zip(flatten(flat_A.shape[1:]), flatten(flat_A.stride[1:]))` which I think is equivalent?

torch/distributed/_pycute/layout.py
```diff
@@ -230,35 +233,56 @@ def composition(layoutA: Layout, layoutB: LayoutInput) -> Layout:
...
+        # For example, if self = (6:2), layout = (3:2), the result is (3:(2*2)) = (3:4).
         if rest_shape != 1 or len(result_shape) == 0:
-            result_shape.append(rest_shape)
-            result_stride.append(rest_stride * flatten(flat_A.stride)[-1])
+            result_shape.append(rest_shape)  # Append to end, will reverse later
+            result_stride.append(rest_stride * flat_A_strides[0])
+
+        # Reverse the lists because we build lists in reverse order (append to end), this way it is more efficient.
+        result_shape.reverse()
+        result_stride.reverse()
```
ezyang, line 283: Everything else here looks plausible but I am going to have to sweat the math lol

torch/distributed/_pycute/layout.py
ezyang: Yup, this is good, I understand complement and this is correct. The important fact for other readers:

image

torch/distributed/_pycute/layout.py
ezyang: TBF I guess we don't care that much about these, as we have no use of them right now

torch/distributed/_pycute/layout.py
ezyang: So it looks like the way the old code worked was by "absorbing" new shape/stride into shape-1 as we iterate over the shape/stride. Because you didn't change the iteration order above (you are still going left-to-right), this strategy no longer works and you had to invert everything. It does seem plausible you made enough changes to make it work, but there's probably a simpler way to do it.

ezyang (approved):
While I think some of the changes here can be done more simply, it does look plausible. Not sure if you think the test coverage is good enough, since you certainly had to add more tests!

torch/distributed/_pycute/layout.py
fduwjj: yes you are right but this change will make CI happy (passing UT) in the lexicographic case.

ezyang: Just need to make sure the UTs aren't actually explicitly checking for colex.

fduwjj: OK one more point here is that the definition of right_inverse is that one index will be itself after layout(inv_layout(i)) and the UT is pretty thorough on this case. So I think with this change it will work in the lexico case. Why? Because when you do map from crd to idx it is using lexico order now.

torch/distributed/_pycute/layout.py
fduwjj: Same answer to right_inverse, I found that as long as we swap the order, i == inv_layout(layout(i)) in the case of lexico. You might ask "can you give me proof using math". I don't have any tbh, and since we don't use them, so I decide to cut corners here. Pycute has two UT: test/distributed/_pycute/test_left_inverse.py and test/distributed/_pycute/test_right_inverse.py which tests the logic pretty thoroughly IIUC.

torch/distributed/_pycute/layout.py
fduwjj: sure, basically we just do it the reverse order as described here: https://docs.nvidia.com/cutlass/media/docs/cpp/cute/02_layout_algebra.html..

And I tried to read the proof in https://leimao.github.io/article/CuTe-Layout-Algebra/, for example, Definition 2.13 Composition - Restricted Case from right to left. I mean at the end of the day the divisor is the same it's the division order?

torch/distributed/_pycute/int_tuple.py
fduwjj: agree we should name it suffix_product.. no it is not a reversed suffix. I just don't want to do insert(0, suffix_product(a[i], current_init))

torch/distributed/_pycute/int_tuple.py
fduwjj: you are right... We don't even need to change this line.. As long as it is lexico, from left to right or from right to left does not matter that much..

torch/distributed/_pycute/layout.py
fduwjj: so if we don't check the length, we will run into infinite recursive call of composition..

torch/distributed/_pycute/layout.py
fduwjj: sure, let me see if I can change this and coalesce both.

torch/distributed/_pycute/int_tuple.py
```diff
@@ -126,18 +126,24 @@ def shape_div(a: IntTuple, b: IntTuple) -> IntTuple:
             return (a + b - 1) // b
 
 
-# Exclusive prefix product with output congruent to input a
+# Exclusive prefix product with output congruent to input a (lexicographic)
 def prefix_product(a: IntTuple, init: IntTuple = 1) -> IntTuple:
     if is_tuple(a):
         if is_tuple(init):  # tuple tuple
             assert len(a) == len(init)
```
Skylion007, line 134: With all these length asserts, may want to create a zip_strict wrapper. I think we already have one in the codebase somewhere

fduwjj: I can add a TODO here so we can do the cosmetic change in another PR.

torch/distributed/_pycute/layout.py
fduwjj: ok when the code hit this line, it is already a tuple of length of 1 or int. The reason why _Layout is working is that inside `__getitem__` we will eventually flatten the size and stride into ints while in the devicemesh layout, we will need to keep the tuple structure always. Hmmm maybe we should allow layout to contain int only as well. Why not.

torch/distributed/_pycute/layout.py
fduwjj: NVM I think I will revert this part.

Issue #162367, opened 2025-09-08
libnvrtc.so failure in TestCompileKernel.test_compile_kernel


tinglvv:
### 🐛 Describe the bug

Error seen from both x86 and sbsa CUDA wheels from our internal nightly testing:
```
root@e8faae565f19:/# python a.py
/usr/local/lib/python3.12/site-packages/torch/_subclasses/functional_tensor.py:279: UserWarning: Failed to initialize NumPy: No module named 'numpy' (Triggered internally at /pytorch/torch/csrc/utils/tensor_numpy.cpp:84.)
  cpu = _conversion_method_template(device=torch.device("cpu"))
Traceback (most recent call last):
  File "//a.py", line 17, in <module>
    add_kernel = _compile_kernel(kernel_source, "add_tensors")
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/torch/cuda/__init__.py", line 1780, in _compile_kernel
    ptx = _nvrtc_compile(
          ^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/torch/cuda/_utils.py", line 68, in _nvrtc_compile
    libnvrtc = _get_nvrtc_library()
               ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/torch/cuda/_utils.py", line 38, in _get_nvrtc_library
    return ctypes.CDLL("libnvrtc.so")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/ctypes/__init__.py", line 379, in __init__
    self._handle = _dlopen(self._name, mode)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: libnvrtc.so: cannot open shared object file: No such file or directory
```
https://github.com/pytorch/pytorch/pull/156674 seems to introduce this error, we should discuss if we would revert this PR, or search for other fixes. 
Tried symlink with https://github.com/pytorch/pytorch/pull/162252, but this won't work with the existing small wheels that are installed through pip install. The symlinks only would work with system libraries. The test would fail in a container without local CTK.

To reproduce 
```python
"""Reproduce NVRTC error from TestCompileKernel.test_compile_kernel"""

import torch
from torch.cuda import _compile_kernel

# Exact kernel from the failing test
kernel_source = """
[...]

PR #162691, opened 2025-09-11
compile_kernel: fix nvrtc discovery and force cuda init

msaroufim:
Untested since I don't have a docker available on dev gpu to test this https://github.com/pytorch/pytorch/issues/162367 - there's 2 additions
1. I borrow the nvrtc discovery logic from cuda-python https://github.com/NVIDIA/cuda-python/tree/main/cuda_pathfinder/cuda/pathfinder/_dynamic_libs
2. I was more explicit about initializing cuda

EDIT: So I talked with @tinglvv offline and it seems like the issue is her NVRTC passes are

```
/usr/local/lib/python3.12/site-packages/nvidia/cu13/lib/libnvrtc-builtins.so.13.0
/usr/local/lib/python3.12/site-packages/nvidia/cu13/lib/libnvrtc.so.13
```

So they're not global but instead in site-packages, this has 2 workarounds
1. I added an explicit nvrtc_path argument to compile_kernel that way people can pass in whatever they want
2. It's possible to fix discovery by using symlinks `sudo ln -s /usr/local/lib/python3.12/site-packages/nvidia/cuda_nvrtc/lib/libnvrtc.so.12 /usr/local/lib/libnvrtc.so.12`

Both approaches above are fairly non intrusive

But if we really want discovery to work reliably it's kinda nasty and we could follow allmost exactly what cuda-python is doing below, testing this seems nasty but if CI passes and @tinglvv code passes then this might work. My vote is against this unless we expect site-packages to be one of. the main places users will have their cuda libraries

```python
def _get_nvrtc_library(nvrtc_path: Optional[str] = None) -> ctypes.CDLL:
    """Find and load NVRTC library following cuda-python's approach.
    
    Based on:
    https://github.com/NVIDIA/cuda-python/blob/main/cuda_bindings/cuda/bindings/_path_finder/find_nvidia_dynamic_library.py
    https://github.com/NVIDIA/cuda-python/blob/main/cuda_bindings/cuda/bindings/_path_finder/supported_libs.py
    """
    import os
    import glob
    import site
    
    if nvrtc_path is not None:
        try:
            return ctypes.CDLL(nvrtc_path)
        except OSError as e:
            raise OSError(f"Failed to load NVRTC library from provided path: {nvrtc_path}") from e
    
    if sys.platform == "win32":
        nvrtc_lib_names = [
            "nvrtc64_130_0.dll",
            "nvrtc64_120_0.dll",
            "nvrtc64_112_0.dll",
            "nvrtc64_111_0.dll",
            "nvrtc64_110_0.dll",
        ]
    else:
        nvrtc_lib_names = [
            "libnvrtc.so.13",
            "libnvrtc.so.12",
            "libnvrtc.so.11.2",
            "libnvrtc.so.11.1",
            "libnvrtc.so.11.0",
            "libnvrtc.so",
        ]
    
    def _find_sub_dirs_all_sitepackages(sub_dirs):
        results = []
        site_packages_dirs = [site.getusersitepackages()] + site.getsitepackages()
        
        for base in site_packages_dirs:
            if not os.path.isdir(base):
                continue
                
            current_paths = [base]
            for sub_dir in sub_dirs:
                next_paths = []
                for current_path in current_paths:
                    if sub_dir == "*":
                        try:
                            for entry in os.listdir(current_path):
                                entry_path = os.path.join(current_path, entry)
                                if os.path.isdir(entry_path):
                                    next_paths.append(entry_path)
                        except OSError:
                            continue
                    else:
                        next_path = os.path.join(current_path, sub_dir)
                        if os.path.isdir(next_path):
                            next_paths.append(next_path)
                current_paths = next_paths
            
            results.extend(current_paths)
        return results
    
    def _try_load_from_paths(lib_names, search_paths):
        for lib_dir in search_paths:
            if not os.path.isdir(lib_dir):
                continue
                
            for lib_name in lib_names:
                lib_path = os.path.join(lib_dir, lib_name)
                if os.path.isfile(lib_path):
                    try:
                        return ctypes.CDLL(lib_path)
                    except OSError:
                        continue
                        
                if not lib_name.endswith("*"):
                    pattern = lib_name + "*"
                    for match in sorted(glob.glob(os.path.join(lib_dir, pattern)), reverse=True):
                        if os.path.isfile(match):
                            try:
                                return ctypes.CDLL(match)
                            except OSError:
                                continue
        return None
    
    for lib_name in nvrtc_lib_names:
        try:
            return ctypes.CDLL(lib_name)
        except OSError:
            continue
    
    if sys.platform == "win32":
        nvidia_sub_dirs = ("nvidia", "*", "bin")
    else:
        nvidia_sub_dirs = ("nvidia", "*", "lib")
    
    site_packages_paths = _find_sub_dirs_all_sitepackages(nvidia_sub_dirs)
    lib = _try_load_from_paths(nvrtc_lib_names, site_packages_paths)
    if lib is not None:
        return lib
    
    cuda_home = os.environ.get("CUDA_HOME") or os.environ.get("CUDA_PATH")
    if cuda_home:
        if sys.platform == "win32":
            cuda_subdirs = ["bin"]
        else:
            cuda_subdirs = ["lib64", "lib"]
            
        cuda_paths = [os.path.join(cuda_home, subdir) for subdir in cuda_subdirs]
        lib = _try_load_from_paths(nvrtc_lib_names, cuda_paths)
        if lib is not None:
            return lib
    
    conda_prefix = os.environ.get("CONDA_PREFIX")
    if conda_prefix:
        if sys.platform == "win32":
            conda_paths = [os.path.join(conda_prefix, "Library", "bin")]
        else:
            conda_paths = [os.path.join(conda_prefix, "lib"), os.path.join(conda_prefix, "lib64")]
            
        lib = _try_load_from_paths(nvrtc_lib_names, conda_paths)
        if lib is not None:
            return lib
    
    if sys.platform != "win32":
        system_paths = [
            "/usr/local/cuda/lib64",
            "/usr/local/cuda/lib",
            "/usr/lib/x86_64-linux-gnu",
            "/usr/lib64",
            "/usr/lib",
            "/opt/cuda/lib64",
            "/opt/cuda/lib",
        ]
        lib = _try_load_from_paths(nvrtc_lib_names, system_paths)
        if lib is not None:
            return lib
    
    raise OSError("Could not find any NVRTC library")
```


tinglvv:
Thanks Mark for the fix, talked offline but let me also add comment here, the site-package path is coming from my testing environment of python3.12 container and could change depending on the test setting. The main concern here is that we would lose the symlinks when we pull the package from pypi dynamically in our wheels.

torch/cuda/_utils.py
```diff
@@ -29,13 +29,48 @@ def _check_cuda(result: int) -> None:
...
+    # Try to load versioned NVRTC libraries in order from newest to oldest
+    # This matches how cuda-python does it
+    # https://github.com/NVIDIA/cuda-python/tree/90096d270d795b9295545e25dc84e1bcd0ca2765/cuda_pathfinder/cuda/pathfinder/_dynamic_libs
     if sys.platform == "win32":
-        return ctypes.CDLL("nvrtc64_120_0.dll")
+        nvrtc_libs = [
+            "nvrtc64_130_0.dll",
+            "nvrtc64_120_0.dll",
+            "nvrtc64_112_0.dll",
+            "nvrtc64_111_0.dll",
```
atalman: We currently don't support cuda 11.0 - should we only keep 12.0 and 13.0 ?

PR #162696, opened 2025-09-11
[CUDAGraph][UX] warn many times for rerecording from dynamic shapes

BoyuanFeng:
Excessive re-recording CUDAGraphs lead to bad performance. We previously warns once if this happens.

However, the limit (=50) is too high and users may just observe bad performance before actually seeing the warning message. Even worse, users may not see the warning message when there are many other logs. @anijain2305 reported that he never saw this warning message when using transformer library, but he DOES observe slowdown due to cudagraph re-recording & needs to turn off cudagraph.

#162663 attempts to hard error when re-recording too many times due to dynamic shapes. But it is a bc-breaking change. Actually, hf-t5-generate model in torchbench failed due to 256 re-recordings.

This PR a) reduces to smaller limit (=8); and b) makes the warning more spam, i.e., warn once for every distinct shapes once the limit is reached.

Fixes #162299

cc @mcarilli @ezyang @eellison @penguinwu @voznesenskym @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @mlazos

### DIFF
--- a/torch/_inductor/cudagraph_trees.py
+++ b/torch/_inductor/cudagraph_trees.py
@@ -371,13 +371,9 @@ def cudagraphify_impl(
     int_key = [i for i, v in enumerate(inputs) if isinstance(v, int)]
     get_ints: Any = operator.itemgetter(*int_key) if int_key else lambda _: None
 
-    has_warn = False
-
     del inputs
 
     def deferred_cudagraphify(inputs: list[InputType]) -> OutputType:
-        nonlocal has_warn
-
         int_key = get_ints(inputs)
 
         if not is_cudagraph_capture_sizes(int_key):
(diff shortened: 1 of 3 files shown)


mlazos (approved):

torch/_inductor/cudagraph_trees.py
```diff
@@ -392,8 +388,7 @@ def deferred_cudagraphify(inputs: list[InputType]) -> OutputType:
         else:
             log.info("recording cudagraph tree for symint key %s", int_key)
 
-        if not has_warn:
-            has_warn = maybe_warning_due_to_dynamic_shape(fn_cache, int_key)
+        maybe_warning_due_to_dynamic_shape(fn_cache, int_key)
```
anijain2305: Not sure if spamming is the right UX. I am ok with one warning as well, reducing 8 is a start (given we dont want to fail hard).

eellison: +1

PR #162702, opened 2025-09-11
[dtensor] do not mutate specs when doing sharding prop

xmfan:
* #163133
* __->__ #162702
* #161458

Because these specs are cached by reference. So by reusing them and mutating them, we're overwriting the cached specs of another op. I'm just fixing these 2, there are more instances, we'll need to do an audit separately.

This fixes a few opinfo tests, but side note that `PYTORCH_OPINFO_SAMPLE_INPUT_INDEX=0 python test/distributed/tensor/test_dtensor_ops.py TestDTensorOpsCPU.test_dtensor_op_db_nn_functional_multi_head_attention_forward_cpu_float32` fails for me locally even on the base commit, but it is not marked as xfail

NOTE: I am renaming `_wrap_output_spec_tensor_meta` so that external libraries will loudly fail. You should migrate to the functional `_create_output_spec_with_new_tensor_meta` or create your own mutation wrapper and take responsibility for the cache! This should be improved in https://github.com/pytorch/pytorch/issues/162731


ezyang (approved):

torch/distributed/tensor/_sharding_prop.py
```diff
@@ -276,7 +277,16 @@ def _wrap_output_spec_tensor_meta(
...
-                    spec.tensor_meta = output_tensor_meta_i
+                    new_specs.append(
+                        spec.shallow_copy_with_tensor_meta(output_tensor_meta_i)
+                    )
+                else:
+                    new_specs.append(spec)
+
+            return tuple(new_specs)
+        else:
+            # I think it should always be None here
```
Skylion007: Add an assert then and mypy can confirm lol

Skylion007 (approved):
Nit on the comment: otherwise looks good

torch/distributed/tensor/_sharding_prop.py
```diff
@@ -223,7 +223,7 @@ def _wrap_output_spec_tensor_meta(
         op: OpOverload,
```
Skylion007, line 223: If you want you can add typing overload signatures to further refine the typing so you know the output given various combinations of input typing

torch/distributed/tensor/_sharding_prop.py
dcci: +1

dcci (approved):

Skylion007 (approved):

Issue #162179, opened 2025-09-04
test_fully_shard_logging.py is not being run


Flamefire:
### 🐛 Describe the bug

In `test_fully_shard_logging.py` the test class is decorated with `skip_if_lt_x_gpu` causing the class to become a function which then causes unittest to not consider it as a test.

Hence it is skipped and no tests are executed:

```
$ python distributed/_composable/fsdp/test_fully_shard_logging.py 

----------------------------------------------------------------------
Ran 0 tests in 0.000s

NO TESTS RAN
```

The issue with the decorator will be fixed by https://github.com/pytorch/pytorch/pull/158846

The test itself fails however. I assume it got broken which went unnoticed as it wasn't run

### Versions

At least PyTorch 2.7.1 up to current master is affected

PR #158846, opened 2025-07-22
Fix decorators skipping NCCL tests 

Flamefire:
Avoid failures caused by tests exiting via sys.exit instead of `unittest.skip`

In particular it will not try to start the test (causing forks into subprocess) just to stop them (killing the subprocess) which is done in the test setup

Using `unittest.skip` decorators avoids the starting of the test in the first place.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci

PR #162706, opened 2025-09-11
Fix accelerator-related skip-decorators for tests

Flamefire:
Avoid failures caused by tests exiting via sys.exit instead of `unittest.skip`

In particular it will not try to start the test (causing forks into subprocess) just to stop them (killing the subprocess) which is done in the test setup

Using `unittest.skip` decorators avoids the starting of the test in the first place.

Reland of https://github.com/pytorch/pytorch/pull/158846 fixing a missing argument

Part of fix for #162179

It changes some decorators, like `require_n_gpus_for_nccl_backend` to check the device count before starting the subprocesses. 
The issue was that the initialization done before checking the GPU count and exit, might use a GPU/NCC based barrier which requires "n GPUs" to be present or will hang forever.

However this evaluates `torch.cuda.device_count()` on importing the test instead of on starting, which may be an issue for tests forking.
Similarly where checking the device count on execution is required due to WORLD_SIZE being used the decorator checks if any GPU is available at all and doesn't start the test otherwise, but then checks and exits based on actual device count

cc @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @aditvenk @weifengpy @kapilsh @H-Huang @ezyang

### DIFF
--- a/torch/testing/_internal/common_distributed.py
+++ b/torch/testing/_internal/common_distributed.py
@@ -224,6 +224,20 @@ def exit_if_lt_x_accelerators(x: int):
         sys.exit(TEST_SKIPS[f"multi-gpu-{x}"].exit_code)
 
 
+def require_accelerator_for_each_rank(func):
+    """Check that at least `self.world_size` accelerators are available.
+    Skip early if none are available."""
+
+    @wraps(func)
+    def wrapper(self, *args, **kwargs):
+        exit_if_lt_x_accelerators(self.world_size)
+        return func(self, *args, **kwargs)
+
+    return unittest.skipUnless(
+        torch.accelerator.current_accelerator is not None, TEST_SKIPS["no_cuda"].message
+    )(wrapper)
+
+
 def skip_if_no_gpu(func):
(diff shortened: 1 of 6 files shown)


Skylion007 (approved):

torch/testing/_internal/common_distributed.py
```diff
@@ -122,14 +122,26 @@ def requires_ddp_rank(device):
     return device in DDP_RANK_DEVICES
 
 
+def exit_if_lt_x_cuda_devs(x):
```
Skylion007: Good to keep track arg should be an int with typing.

Skylion007 (approved):

torch/testing/_internal/common_distributed.py
Flamefire: Makes sense, did the same to similar functions as noticed during doing this.

While doing that I noticed `require_n_gpus_for_nccl_backend` is the same as  `nccl_skip_if_lt_x_gpu` so I change the latter to use the former to prefer the `@require_x` decorator style

PR #162713, opened 2025-09-11
[FIX] [AOTI] improve `load_constants` bug fix

shaoyuyoung:
This is a supplement to #161887, which only fixes one place, but in fact, it may also cause problems during `MPS compilation`.
Please correct me if I am wrong.


angelayi (commented):
this makes sense to me! thanks! could you please add a test?

PR #162715, opened 2025-09-11
[FIX] add out-of-bound array check for `foreachTensorInplaceWithFlag`

shaoyuyoung:
Originally, we set three assert rules, but we have
https://github.com/pytorch/pytorch/blob/afdd4247a2251b3f4c2f4b402cb625f46d6784ba/aten/src/ATen/functorch/DynamicLayer.cpp#L303
The max size of `use_flag_relative` is 64.
AS a result, `end - begin` should be `<=64`

One more word: `end - begin` means the `number of return values`. Theoretically speaking, it seems unlikely that there will be more than 64 return values, so in general, this assertion error is not likely to be triggered to cause a crash.


zou3519 (commented):
I think the question is: can you write a test case that triggers this issue? This might already error out nicely, or there may already be some system invariant that disallows this issue.

I don't remember what I was thinking when I wrote this code.

Also, if you have a lot of these changes in similar files, please do try to bundle the changes together.

shaoyuyoung:
Sorry to bother, how should I add a test for CPP filse in `functorch`? 
Previously, I only contributed PyTorch Python part. I'm a newcomer to CPP part. 
If u can tell me where I should add the test and give some examples, it will be nice :)

PR #162716, opened 2025-09-11
[FIX] Add negative number checking to `ADInterpreters.cpp`, aoviding out-of-bound array

shaoyuyoung:
If `front<0`, `front + arg_idx` may also be less than 0.


aten/src/ATen/functorch/ADInterpreters.cpp
```diff
@@ -131,6 +131,7 @@ static void autogradBasedTransformSendToNext(
   auto args_size = op.schema().arguments().size();
   const auto ret_size = op.schema().returns().size();
   // Step 1
+  TORCH_INTERNAL_ASSERT(stack->size() >= args_size); // prevent `front` from being negative
```
zou3519, line 134: some invariant of the system might already prevent this from happening..

PR #162724, opened 2025-09-11
Used a fixed-length array to represent the interpreter stack

guilhermeleobas:
* __->__ #162724

CPython represents the stack of a frame as a fixed length C array + a stack pointer. The pointer points to the next free slot. When the interpreter executes a PUSH or a POP operation, it increments or decrements the pointer, without having to resize the underlying array. This allows fast O(1) append/pop operations.


torch/_dynamo/symbolic_convert.py
```diff
@@ -1048,6 +1048,73 @@ def _missing(opname: str, *args: Any) -> None:
...
+    https://github.com/python/cpython/blob/32e1e0699ffda8ec1dd5a0eb178b052352ab7d31/Objects/frameobject.c#L2122-L2139
+    """
+
+    def __init__(self, capacity: int) -> None:
+        super().__init__([None] * capacity)
+        self.stack_pointer = 0  # points to the next free slot
+
+    # @torch._dynamo.utils.measure_time
+    def append(self, value: VariableTracker) -> None:
+        if self.stack_pointer >= super().__len__():
```
Skylion007: Why not add an assert failure if it's never supposed to happen?

torch/_dynamo/symbolic_convert.py
```diff
@@ -1048,6 +1048,73 @@ def _missing(opname: str, *args: Any) -> None:
         cls.dispatch_table = [dispatch_table.get(i) for i in range(2**8)]
 
 
+class Stack(list[Optional[VariableTracker]]):
```
Skylion007: Subclasses list is going to add a bunch of methods that could break things here, we should explicitly add method stubs? Should we just implement a mutable sequence abc or a deque abc and manage the list with a wrapper class instead of a subclass inheritance strategy? Seems cleaner.

torch/_dynamo/symbolic_convert.py
```diff
@@ -1048,6 +1048,73 @@ def _missing(opname: str, *args: Any) -> None:
...
+    def __iter__(self):
+        for i in range(self.stack_pointer):
+            yield super().__getitem__(i)
+
+    def __str__(self) -> str:
+        return super().__getitem__(slice(0, self.stack_pointer)).__str__()
+
+    def __repr__(self) -> str:
+        return super().__getitem__(slice(0, self.stack_pointer)).__repr__()
+
```
Skylion007: Still a ton of inherited methods that could break things, like clear() etc

torch/_dynamo/symbolic_convert.py
```diff
@@ -1048,6 +1048,72 @@ def _missing(opname: str, *args: Any) -> None:
...
+        self._data[self.stack_pointer] = value
+        self.stack_pointer += 1
+
+    def pop(self) -> VariableTracker:
+        if self.stack_pointer == 0:
+            raise IndexError("pop from empty stack")
+        self.stack_pointer -= 1
+        value = self._data[self.stack_pointer]
+        self._data[self.stack_pointer] = None
+        return value
```
Skylion007: need not_none typing utility a lot here

torch/_dynamo/symbolic_convert.py
```diff
@@ -1048,6 +1048,72 @@ def _missing(opname: str, *args: Any) -> None:
         cls.dispatch_table = [dispatch_table.get(i) for i in range(2**8)]
 
 
+class Stack:
```
Skylion007: This is really faster than collections.Deque? Sure collections Deque has O(n) get / set access, but is that really necessary here I guess?

Skylion007: Also probably should subclass some abstract classes like Sized, MutableSequence, etc

Skylion007 (commented):
See nits with Stack class

torch/_dynamo/symbolic_convert.py
guilhermeleobas: Hey, thanks for the feedback. I'm trying some stuff before marking this PR as ready for review. Subclassing list or having this implemented in Python was actually a little bit slower. So, I'm trying to implement it in C++ to see if we can get any speed up.

PR #160601, opened 2025-08-14
[dynamo, nested graph breaks] move cell codegen before side effects codegen

williamwen42:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #160611
* #163335
* #163503
* #162737
* __->__ #160601

This is needed because if we codegen cells for nested frames AFTER side effects, then reconstruction could get messed up. From below:

>The added test case demonstrates the reconstruction failure if we kept cell codegen at the original place (only happens with nested graph breaks since we reconstruct nested frame cells from VariableTracker rather than directly using LOAD_CLOSURE).

>At a high level, what happened before this change was that side_effects was pruning the cells (I don't recall exactly why this happens), and because cells were codegen'd after the side effects were applied, we were unable to properly reconstruct the cell. The error I was seeing was a list/tuple IndexError.

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela

PR #162737, opened 2025-09-11
[dynamo, nested graph breaks] fix nested step graph break related issues

williamwen42:
* #160611
* #163335
* #163503
* __->__ #162737
* #160601

Turns out codegen'ing a nested step graph break is significantly more complicated than first thought. The optimized function should actually do:
- call graph/load values/do side effects etc.
- call into the leaf's resume function, but skipped (this essentially step graph break function for just the leaf function)
- call into all the other resume functions, traced.

This PR also adds `torch._dynamo.step_unsupported()`, which can be used for internal testing purposes to better test step graph break handling.


Lucaskabela (approved):
Generally LGTM, the comments in symbolic_convert made it easier to reason about

PR #162747, opened 2025-09-11
bring back the old vllm's use_existing_torch.py

yangw-dev:
vllm's pr will override our dependencies for torch.

quick fix to add the use_existing_torch.py. syncing with vllm now regarding the uv approach they have


huydhn (approved):
LGTM!

PR #162749, opened 2025-09-11
fix wait() missing in redistribute tensor

zpcore:
We notice that the wait() op is missing after collective op call: https://github.com/pytorch/pytorch/pull/162665#discussion_r2338460562.

The issue is that `_maybe_warp_tensor` calls AsyncCollectiveTensor in https://github.com/pytorch/pytorch/blob/3ad3bfe11df7bf542885a6dd2dc2ada4ab940e53/torch/distributed/_functional_collectives.py#L829 We need to check whether the wait() is required after collective op call.


ezyang (approved):

SherlockNoMad (approved):

wconstab (approved):
Thanks for fixing this!

PR #162754, opened 2025-09-11
[scan] create fw and bw graphs via partitioning

ydwu4:
* __->__ #162754
* #161732
* #162025
* #161808
* #161664
* #161557

cc @ezyang @EikanWang @jgong5 @wenzhe-nrv @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela @SherlockNoMad


torch/_functorch/partitioners.py
```diff
@@ -1551,6 +1551,8 @@ def cleanup_recompute_tags(joint_module: fx.GraphModule) -> fx.GraphModule:
             for user in node.users:
                 if (
                     must_recompute(user)
+                    and "ac_graph_id" in user.meta
+                    and "ac_graph_id" in node.meta
```
zou3519, line 1565: what's this for?

torch/_higher_order_ops/partitioner.py
```diff
@@ -0,0 +1,362 @@
+import logging
+from typing import Any, Callable, Union
+
```
zou3519, line 3: Do you have any tests? Are there no expecttests for scan forward+backward that need to be updated as a result of this change? if not we should add a test that demonstrates the partitioning is actually happening

torch/_higher_order_ops/scan.py
```diff
@@ -559,300 +449,361 @@ def combine_fn_bw_grad_accumulation(*args):
...
+                    0
+                ]
+            ),
+        )
+
+        flat_grads = scan_op(
+            bw_single_step_gm,
+            pytree.tree_flatten(bw_init)[0],
+            # TODO: torch.flip copies the tensor, we should optimize it away
+            [torch.flip(x, (0,)) for x in pytree.tree_flatten(bw_xs)[0]],
```
zou3519, line 771: how are you thinking of optimizing it away?

zou3519 (approved):

torch/_functorch/partitioners.py
ydwu4: We didn't go through aot_autograd so these two metas are not set. The prior logic only works when this meta exists.

torch/_higher_order_ops/partitioner.py
ydwu4: yeah, I can add some tests guarding this.

torch/_higher_order_ops/scan.py
ydwu4: We could pass the reverse keyword all the way to inductor and do a reversed traverse there.

PR #162757, opened 2025-09-11
[inductor] partially type torch/_inductor/codegen/multi_kernel.py

pianpwk:
Test Plan:
None

Rollback Plan:

Differential Revision: D82253817

### DIFF
--- a/torch/_inductor/codegen/multi_kernel.py
+++ b/torch/_inductor/codegen/multi_kernel.py
@@ -3,6 +3,7 @@ import functools
 import logging
 import os
 import pathlib
+from typing import Any, Optional, Union
 
 from torch._inductor.ir import MultiTemplateBuffer
 from torch._inductor.metrics import get_metric_table, is_metric_table_enabled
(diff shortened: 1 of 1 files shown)


torch/_inductor/codegen/multi_kernel.py
```diff
@@ -49,7 +51,7 @@ def define_kernel(self, kernels):
         # Prevent circular import
         from ..select_algorithm import TritonTemplateKernel
 
-        kernel_names = tuple(k.kernel_name for k in kernels)
+        kernel_names: tuple[str] = tuple(k.kernel_name for k in kernels)
```
Skylion007, line 54: This should be tuple[str, ...]

torch/_inductor/codegen/multi_kernel.py
```diff
@@ -364,7 +374,7 @@ def inner():
             for index, kernel in enumerate(self.kernels)
         ]
 
-    def _get_filtered_args(self, args, index):
+    def _get_filtered_args(self, args: tuple[Any, ...], index: int) -> Union[list[Any], tuple[Any, ...]]:
```
Skylion007, line 377: It could be a tuple of a type var matching the inputs, right?

```suggestion
    def _get_filtered_args(self, args: tuple[_T, ...], index: int) -> Union[list[Any], tuple[_T, ...]]:

PR #162763, opened 2025-09-11
fix high=0 bug in nll_loss test

haifeng-jin:
Minor bug fix for the `nll_loss` test.
Before this PR, it runs `torch.randint(high=0)`, which will fail because it would try to generate a number that >= low and < high, i.e. x>=0 and x<0.

The test did not fail because that line is not run when testing on CPU because it failed earlier because of a unsupported dtype.
However, as we support TPUs at Google, this line is reached first before the dtype check, which triggers the bug.

To my understanding, these OpInfo should be general enough to support different hardware.
Fixing this obvious bug would make it more general cross different hardware.

### DIFF
--- a/torch/testing/_internal/common_methods_invocations.py
+++ b/torch/testing/_internal/common_methods_invocations.py
@@ -8468,7 +8468,7 @@ def sample_inputs_nll_loss(op_info, device, dtype, requires_grad, **kwargs):
             yield make_input(s), make_target(s), dict(reduction=reduction)
             yield make_input(s), make_target(s), dict(weight=make_weight(), reduction=reduction)
             yield make_input(s), make_target(s), dict(weight=make_weight(low=0), reduction=reduction)
-            yield make_input(s), make_target(s), dict(weight=make_weight(high=0), reduction=reduction)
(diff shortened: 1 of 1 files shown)


torch/testing/_internal/common_methods_invocations.py
```diff
@@ -8468,7 +8468,7 @@ def gen_shape_kwargs():
             yield make_input(s), make_target(s), dict(reduction=reduction)
             yield make_input(s), make_target(s), dict(weight=make_weight(), reduction=reduction)
             yield make_input(s), make_target(s), dict(weight=make_weight(low=0), reduction=reduction)
-            yield make_input(s), make_target(s), dict(weight=make_weight(high=0), reduction=reduction)
+            yield make_input(s), make_target(s), dict(weight=make_weight(high=1), reduction=reduction)
```
soulitzer, line 8471: since low is only default 0 if dtype is integral, and -9 otherwise
how about doing something like:
```
if dtype.is_floating_point or dtype.is_complex:
    yield make_input(s), make_target(s), dict(weight=make_weight(high=0), reduction=reduction)
```

soulitzer (approved):
Thanks!

PR #162767, opened 2025-09-11
[flatbuffer] Fix compile error due to discarded result

pulsejet:
Summary:
One of our builds fails because the return value of fread is discarded. Explicit cast to void fixes the build.

```log
In file included from fbcode/caffe2/torch/csrc/jit/mobile/import.cpp:15:
fbcode/caffe2/torch/csrc/jit/mobile/file_format.h:156:3: error: ignoring return value of function declared with 'warn_unused_result' attribute [-Werror,-Wunused-result]
  156 |   fread(data.get(), size, 1, f);
      |   ^~~~~ ~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
...
BUILD FAILED
Failed to build 'fbcode//caffe2:libtorch (cfg:opt-linux-x86_64-clang19-no-san-opt-by-default#fef256f7ee896871)'
```

Test Plan:
No runtime behavior change. CI.

Rollback Plan:

Differential Revision: D82265002

### DIFF
--- a/torch/csrc/jit/mobile/file_format.h
+++ b/torch/csrc/jit/mobile/file_format.h
@@ -153,7 +153,8 @@ static inline std::tuple<std::shared_ptr<char>, size_t> get_file_content(
   size_t buffer_size = (size / kMaxAlignment + 1) * kMaxAlignment;
   std::shared_ptr<char> data(
       static_cast<char*>(c10::alloc_cpu(buffer_size)), c10::free_cpu);
-  fread(data.get(), size, 1, f);
+  auto nread = fread(data.get(), size, 1, f);
+  TORCH_CHECK(nread == 1, "Failed to read file: ", filename);
   fclose(f);
 #endif
   return std::make_tuple(data, size);


torch/csrc/jit/mobile/file_format.h
```diff
@@ -153,7 +153,7 @@ static inline std::tuple<std::shared_ptr<char>, size_t> get_file_content(
   size_t buffer_size = (size / kMaxAlignment + 1) * kMaxAlignment;
   std::shared_ptr<char> data(
       static_cast<char*>(c10::alloc_cpu(buffer_size)), c10::free_cpu);
-  fread(data.get(), size, 1, f);
+  (void)fread(data.get(), size, 1, f);
```
Skylion007: Use maybe_unused C++17 feature instead of the void trick

pulsejet: @Skylion007 TIL, thank you!

torch/csrc/jit/mobile/file_format.h
```diff
@@ -153,7 +153,7 @@ static inline std::tuple<std::shared_ptr<char>, size_t> get_file_content(
   size_t buffer_size = (size / kMaxAlignment + 1) * kMaxAlignment;
   std::shared_ptr<char> data(
       static_cast<char*>(c10::alloc_cpu(buffer_size)), c10::free_cpu);
-  fread(data.get(), size, 1, f);
+  size_t nread [[maybe_unused]] = fread(data.get(), size, 1, f);
```
malfet: nit. But IMO it would be good to check that `nread == size`, because if it's not then one is returning bogus/incomplete results
```suggestion
  auto nread [[maybe_unused]] = fread(data.get(), size, 1, f);
```

pulsejet: Added a check instead. Note that fread returns the number of chunks read, not the byte count

torch/csrc/jit/mobile/file_format.h
```diff
@@ -153,7 +153,8 @@ static inline std::tuple<std::shared_ptr<char>, size_t> get_file_content(
   size_t buffer_size = (size / kMaxAlignment + 1) * kMaxAlignment;
   std::shared_ptr<char> data(
       static_cast<char*>(c10::alloc_cpu(buffer_size)), c10::free_cpu);
-  fread(data.get(), size, 1, f);
+  auto nread = fread(data.get(), size, 1, f);
+  TORCH_CHECK(nread == 1, "Failed to read file: ", filename);
   fclose(f);
 #endif
   return std::make_tuple(data, size);
```
Skylion007, line 160: Slight other nit: data should have std::move called on it to avoid atomic lock.

Skylion007 (approved):

torch/csrc/jit/mobile/file_format.h
pulsejet: Have to defer, I don't have any background on this (just trying to fix a downstream build)

PR #162768, opened 2025-09-11
[dynamo] Implement iter with a polyfill

rtimpe:
* #166350
* __->__ #162768
* #166349

Currently most variable trackers implement `iter` via `_call_iter_tuple_list`.
This makes it difficult to customize the behavior of `iter` for different
variable types.  Instead, implement `iter` via a polyfill, which will delegate
to the appropriate `__iter__` method.

While this method is more flexible, it increases the overhead of dynamo tracing.
For example, `iter(x)` will generate 9x more instructions than the current
implementation for common iterable types.  Microbenchmarking shows a ~6x
slowdown for this operation.  I suspect this would be much less for realistic
workloads, but more work would be needed to get specific numbers.  If the
performance is a concern we could also consider adding a fast path for types
that are known to correctly implement `__iter__`.


torch/_dynamo/variables/builtin.py
```diff
@@ -1817,31 +1817,23 @@ def _call_tuple_list(self, tx, obj=None, *args, **kwargs):
         else:
             return self._call_iter_tuple_list(tx, obj, *args, **kwargs)
 
+    # TODO use substitute_in_graph on polyfill instead
     def call_iter(self, tx: "InstructionTranslator", obj, *args, **kwargs):
-        if isinstance(obj, variables.IteratorVariable):
-            ret = obj
-        elif isinstance(obj, variables.RangeVariable):
-            ret = obj.call_method(tx, "__iter__", [], {})
-        else:
```
guilhermeleobas, line 1837: I think we need to keep some of these to reduce tracing time. If we know `obj` is already an iterator, we don't need to pay the cost of tracing of calling the `iter_` polyfill.

torch/_dynamo/variables/constant.py
```diff
@@ -152,6 +152,14 @@ def call_method(
...
             except NotImplementedError:
                 return super().call_method(tx, name, args, kwargs)
+        elif name == "__iter__" and istype(self.value, str):
+            # this could be some generic iterator to avoid the circular import,
+            # but ListIterator does what we want
+            from .lists import ListIteratorVariable
+
+            return ListIteratorVariable(
+                self.unpack_var_sequence(tx), mutation_type=ValueMutationNew()
+            )
```
guilhermeleobas, line 178: Nice addition! Move this one to a separate PR and we can land it.

torch/_dynamo/variables/dicts.py
```diff
@@ -728,6 +729,12 @@ def call_method(
         elif name == "__ior__":
             self.call_method(tx, "update", args, kwargs)
             return self
+        elif name == "__iter__":
+            if self.source and not is_constant_source(self.source):
+                tx.output.guard_on_key_order.add(self.source)
+            return ListIteratorVariable(
```
guilhermeleobas, line 786: Same here

torch/_dynamo/variables/dicts.py
```diff
@@ -742,6 +749,11 @@ def call_obj_hasattr(self, tx, name):
             if name in self.user_cls.__dict__:
                 return ConstantVariable.create(True)
             return ConstantVariable.create(False)
+        elif (
+            self.user_cls in (collections.OrderedDict, collections.defaultdict)
+            and name == "__iter__"
+        ):
```
guilhermeleobas: Maybe just augment the if above: `if self.user_cls in (dict, collections.OrderedDict, collections.defaultdict)`

torch/_dynamo/variables/lists.py
```diff
@@ -1403,15 +1414,21 @@ def has_unpack_var_sequence(self, tx):
         return True
 
     def unpack_var_sequence(self, tx):
+        if self.is_exhausted:
+            return []
         r = list(self.items[self.index :])
         self.index = len(self.items)
+        self.is_exhausted = True
```
guilhermeleobas: Move the `self.index` pointer to the end of the list, instead. I think that is enough and you can avoid the extra conditionals for `self.is_exhausted`.

guilhermeleobas: Actually, the index is already moved to the end. Is `self.exhausted` needed?

torch/_dynamo/variables/tensor.py
```diff
@@ -453,6 +453,11 @@ def var_getattr(self, tx: "InstructionTranslator", name):
 
         if name == "__class__":
             return UserDefinedClassVariable(self.python_type())
+        elif name == "__iter__":
+            it = ListIteratorVariable(
+                list(self.unpack_var_sequence(tx)), mutation_type=ValueMutationNew()
```
guilhermeleobas: I think `self.unpack_var_sequence(tx)` already returns a list. Is the extra list call needed?

guilhermeleobas (changes requested):
Code looks good to me @rtimpe. We might just have to stick with parts of the old implementation for a bit more to avoid increasing the tracing time.

torch/_dynamo/variables/lists.py
rtimpe: `self.exhausted` is needed because `self.items` is now shared with the original list.  So if you call `unpack_var_sequence`, then add an item to the list, then call `unpack_var_sequence` again `self.index` will no longer point to the end of the list.

torch/_dynamo/variables/tensor.py
rtimpe: I think I just copied it, shouldn't be needed.  I'll remove it (and the numpy one as well)

torch/_dynamo/variables/builtin.py
rtimpe: I think I got all of the VT types we care about, but let me know if I missed any

torch/_dynamo/variables/dicts.py
rtimpe: Done, although it makes the attribute lookup a bit more complicated.  I'm using getattr, which I think is safe, but please take a look

torch/_dynamo/variables/builtin.py
```diff
@@ -1817,15 +1817,21 @@ def _call_tuple_list(self, tx, obj=None, *args, **kwargs):
...
-        elif isinstance(obj, variables.RangeVariable):
-            ret = obj.call_method(tx, "__iter__", [], {})
+        # avoid the overhead of tracing the polyfill if we already know the class implemented __iter__
+        if isinstance(
+            obj,
+            (
+                variables.ListVariable,
+                variables.IteratorVariable,
+                variables.ConstDictVariable,
+                variables.NNModuleVariable,
```
guilhermeleobas, line 1839: Can you add `variables.RangeVariable` to this list?

torch/_dynamo/variables/dicts.py
```diff
@@ -738,12 +745,16 @@ def unpack_var_sequence(self, tx):
...
-        # dict not allow setting arbitrary attributes. To check for hasattr, we can just check the __dict__ of the dict.
-        # OrderedDict though requires side effects tracking because it supports arbitrary setattr.
-        if self.user_cls is dict:
-            if name in self.user_cls.__dict__:
+        # dict not allow setting arbitrary attributes.  OrderedDict and
+        # defaultdict allow arbitrary setattr, but not deletion of default attrs
+        if any(
+            self.user_cls is t
+            for t in (dict, collections.OrderedDict, collections.defaultdict)
+        ):
```
guilhermeleobas, line 802: Is this the same as `if self.user_cls in (dict, collections.OrderedDict, collections.defaultdict)`?

guilhermeleobas: Also, maybe move all the `call_obj_hasattr` changes to a separated PR.

torch/_dynamo/variables/lists.py
```diff
@@ -1461,14 +1470,16 @@ def __init__(self, items, index: int = 0, **kwargs) -> None:
         # assert all(isinstance(x, VariableTracker) for x in items)
         self.items = items
         self.index = index
+        self.is_exhausted = False
```
guilhermeleobas: Move the changes related to `is_exhausted` to a separated PR and I'll approve it. Maybe just mention why it is needed + a test case.

guilhermeleobas (changes requested):

torch/_dynamo/variables/dicts.py
rtimpe: It's not quite the same.  Checking `self.user_class in ...` will compare `self.user_class` using `==` as well as `is`, which can execute user code if `self.user_class` has a custom metaclass.  So probably not much difference in practice, but this version is slightly safer.

torch/_dynamo/variables/builtin.py
```diff
@@ -1828,17 +1828,22 @@ def _call_tuple_list(self, tx, obj=None, *args, **kwargs):
...
+                variables.RangeVariable,
+                variables.IteratorVariable,
+                variables.ConstDictVariable,
+                variables.NNModuleVariable,
+            ),
+        ):
+            return obj.call_method(tx, "__iter__", [], {})
+        elif isinstance(obj, variables.TensorVariable):
+            m = obj.var_getattr(tx, "__iter__")
+            return m.call_function(tx, [], {})
```
guilhermeleobas: Can you just merge this if with the previous one? Just add `variables.TensorVariable` to the previous `if`.

guilhermeleobas (changes requested):

torch/_dynamo/variables/builtin.py
rtimpe: Done, though this required some changes to TensorVariable in the previous pr

guilhermeleobas (approved):

PR #162770, opened 2025-09-11
[c10d] destroy_process_group() works without default group created

kwen2501:
* #162571
* #162549
* #163058
* __->__ #162770
* #162545
* #162529

### DIFF
--- a/test/distributed/test_c10d_nccl.py
+++ b/test/distributed/test_c10d_nccl.py
@@ -831,6 +831,22 @@ class ProcessGroupNCCLGroupTest(MultiProcessTestCase):
         dist.destroy_process_group()
 
     @requires_nccl()
+    @skip_but_pass_in_sandcastle_if(not TEST_MULTIGPU, "NCCL test requires 2+ GPUs")
+    def test_destroy_without_default(self):
+        """Test if destroy works without default world group created"""
+        store = c10d.FileStore(self.file_name, self.world_size)
+        device = torch.device(f"cuda:{self.rank % torch.cuda.device_count()}")
+        dist.init(self.world_size, self.rank, store)
+        subgroup_ranks = [0, 1]
+        subgroup = dist.new_group(
+            ranks=subgroup_ranks,
+            backend="nccl",
+        )
+        tensor = torch.empty(1024, device=device)
+        dist.all_reduce(tensor, group=subgroup)
+        dist.destroy_process_group()
+
+    @requires_nccl()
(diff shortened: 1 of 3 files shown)


test/distributed/test_c10d_nccl.py
```diff
@@ -830,6 +830,22 @@ def test_abort_in_destroy_mixed_empty_pgs(self):
...
+        device = torch.device(f"cuda:{self.rank % torch.cuda.device_count()}")
+        dist.init(self.world_size, self.rank, store)
+        subgroup_ranks = [0, 1]
+        subgroup = dist.new_group(
+            ranks=subgroup_ranks,
+            backend="nccl",
+        )
+        tensor = torch.empty(1024, device=device)
+        dist.all_reduce(tensor, group=subgroup)
+        dist.destroy_process_group()
```
ezyang, line 847: Hmm, so the problem is I don't know if we will reliably not have a default group when running unit tests together in one process, it is something where it is easy to forget to do in tearDown and now you have a flaky test. Should you assert that there is no default PG at the start of the test, maybe?

torch/distributed/distributed_c10d.py
```diff
@@ -2350,13 +2353,16 @@ def _abort_process_group(group: Optional[ProcessGroup] = None):
...
+    # destruction of default group below.
+    # Otherwise, some checking:
+    if pg is not None:
+        if _world.pg_map.get(pg, None) is None:
+            raise ValueError("Invalid process group specified or has been destroyed.")
 
     try:
         backend = pg._get_backend(torch.device("cuda"))
-    except RuntimeError:
+    except Exception:
```
ezyang: How come? Suppressing the exception unconditionally already is pretty skeevy, and expanding the exception catch is worse.

test/distributed/test_c10d_nccl.py
kwen2501: We have `destroy_process_group` at the TearDown of the test template:
https://github.com/pytorch/pytorch/blob/main/torch/testing/_internal/common_distributed.py#L894

So I believe with traditional `MultiProcessTestCase` (the case here, not the continuous TestCase), the default PG is killed after each test method.

torch/distributed/distributed_c10d.py
kwen2501: Thanks, fixed.

ezyang:
deferring review to @fduwjj

PR #162772, opened 2025-09-11
Fix boxcox to return same result for same input in one batch

Yuhta:
Summary:
The SIMD path is using SLEEF version of `pow` which is slightly different from `std::pow`.  The fix is to use the same vectorized code (with partial load and store) for the trailing data as well to ensure consistency between results.

Rollback Plan:

Differential Revision: D82265247


swolchok (changes requested):

swolchok (approved):

PR #162774, opened 2025-09-11
Fix cpp build

svekars:
Fixes #ISSUE_NUMBER

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -41,7 +41,7 @@ jobs:
     uses: ./.github/workflows/_linux-build.yml
     needs: get-label-type
     with:
-      runner: "${{ needs.get-label-type.outputs.label-type }}linux.8xlarge.memory"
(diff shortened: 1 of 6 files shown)


huydhn:
Adding `ciflow/nightly` to build C++ docs

.github/workflows/nightly.yml
```diff
@@ -41,7 +41,7 @@ jobs:
     uses: ./.github/workflows/_linux-build.yml
     needs: get-label-type
     with:
-      runner: "${{ needs.get-label-type.outputs.label-type }}linux.2xlarge"
+      runner: "${{ needs.get-label-type.outputs.label-type }}linux.8xlarge.memory"
```
clee2000: It's is poorly named, this job is actually the job that builds the pytorch binary, not the one that builds the docs.  The job that builds the docs should be in _docs.yml

malfet (approved):
Looks good to me, though it feels like a better fix to move exhale dependency forward, as 0.2.3 was release in 2019

atalman (approved):

Issue #161905, opened 2025-09-01
[MPS] `torch.compile` ResNet-18 model fails during `loss.backward()` on MPS backend, but works on CPU


Zi-Gao:
#### 🐛 Bug

When using `torch.compile` on a model that has been moved to the `mps` device on macOS, calling `loss.backward()` results in a runtime error. The forward pass completes without issues.

Notably, the exact same code runs successfully without any errors when the model and data are on the `cpu` backend. This suggests a potential bug or unimplemented feature in the `torch.compile` support for the MPS backend, specifically concerning the backward pass (gradient computation).

#### To Reproduce

Here is a minimal, self-contained code snippet to reproduce the error:

```python
import torch
import torch.nn as nn
import torch.optim as optim
from torch import Tensor
from typing import Type, Any, Callable, Union, List, Optional
from torchvision.models.resnet import resnet18


BATCH_SIZE = 4
NUM_CLASSES = 10
LEARNING_RATE = 0.01
device='mps'

model = resnet18(num_classes=NUM_CLASSES)

criterion = nn.CrossEntropyLoss()
optimizer = optim.SGD(model.parameters(), lr=LEARNING_RATE)

model=model.to(device)
model.train() 

@torch.compile
def train(images,labels):
    images=images.to(device)
    labels=labels.to(device)
    
    optimizer.zero_grad()

    outputs = model(images)
    loss = criterion(outputs, labels)
    loss.backward()
    optimizer.step()

images = torch.randn(BATCH_SIZE, 3, 224, 224)
labels = torch.randint(0, NUM_CLASSES, (BATCH_SIZE,))

train(images,labels)
```

#### Expected behavior

The code should execute without any errors, regardless of whether the `DEVICE` is set to `'mps'` or `'cpu'`. The `loss.backward()` call should complete successfully, and the gradients for the model's parameters should be computed.

#### Actual behavior

When `DEVICE = 'mps'`, the code fails at `loss.backward()` with a runtime error. The forward pass and loss calculation are successful.

### Error logs

```
W0901 19:47:33.366000 11122 site-packages/torch/_inductor/utils.py:2241] [0/0_1] DeviceCopy in input program
W0901 19:47:33.367000 11122 site-packages/torch/_inductor/utils.py:2241] [0/0_1] DeviceCopy in input program
W0901 19:47:34.821000 11122 site-packages/torch/_inductor/utils.py:1554] [1/0_1] Not enough SMs to use max_autotune_gemm mode
[...]

PR #162776, opened 2025-09-12
[MPS] Fix conv layout handling

malfet:
* __->__ #162776

What started as simple fix for `mps_convolution_backward_input` resulted in a pretty significant refactor/fixes:
- Updated `mps_conv_use_channels_last` to return channels last output if either input or weights are channels last
- Use the same primitive throughout `Convolution.mm` to determine wether output should be allocated in channels last format or not

But doing only those two, resulted in crash in `test_memory_format_nn_Conv2d_mps_float32`, when weights were backward, and bias is present:
```
% python -c "import torch;print(torch.nn.functional.conv2d(torch.rand(2, 4, 3, 4,device='mps'), torch.rand(5, 4, 3, 3,device='mps').to(memory_format=torch.channels_last), torch.rand(5,device='mps')))"
/AppleInternal/Library/BuildRoots/4~B5E4ugDCh2RsPWAjMEoPu8LC5w1yXEwd7XweDhg/Library/Caches/com.apple.xbs/Sources/MetalPerformanceShadersGraph/mpsgraph/MetalPerformanceShadersGraph/Core/Files/MPSGraphExecutable.mm:3619: failed assertion `Error: MLIR pass manager failed'
zsh: abort      python -c 
``` 

Which requires a more thorough redesign/cleanup, namely:
- Do not alter the layout based on MacOS version, but rather do additional copies on MacOS-14 if inputs/output or weight are in channels last format ( done by defining `std::optional<Tensor> output_c;` that contains a contiguous copy of the output tensor
- Introduced `input_suggested_layout` which is set to ChannelsLast if and only if input is channels last and is running on MacOS-15+
- Delete unused `memory_layout` and `group` arguments from `fill_depthwise_conv_desc`
- Fix bias broadcasting logic for channels last

As result, in addition to adding one more regression test this change removes `expectedFailures` from:
- `TestModule.test_memory_format` for `Conv2d`, `ConvTranspose2d`, `LazyConv1d`, `LazyConvTranspose1d`
- `test_require_stride_expanded_dynamic_shapes`
-  `test_mutable_custom_op_fixed_layout2` for MacOS-14

Fixes https://github.com/pytorch/pytorch/issues/161905


aten/src/ATen/native/mps/operations/Convolution.mm
```diff
@@ -400,17 +403,7 @@ static Tensor mps_convolution_backward_input(IntArrayRef input_size,
...
-      case at::MemoryFormat::ChannelsLast:
-        mem_format_key = "ChannelsLast";
-        break;
-      default:
-        assert(0 && "Check should have been done earlier\n");
-    }
+    const auto mem_format_key = is_channels_last ? "ChannelsLast" : "Contiguous";
 
     MPSShape* mps_input_shape = getMPSShape(input_size);
     std::string key;
```
Skylion007: Key generation would have been cleaner and more efficient with fmtlib. Probably better for another PR

malfet: The entire Convolution.mm would benefit from some BE work.. Will probably submit a small PRs soon

Skylion007 (approved):

PR #162777, opened 2025-09-12
[pcache] Cache and AsyncCache implementations

nmacchioni:
Summary:
Implemented caching abstractions: `Cache` and `AsyncCache`.

`Cache` provides an abstraction for defining simple key -> value stores with get and put functionality. We propose using `Cache` for implementations with very low (microseconds) overhead, for example an in-memory cache.

`AsyncCache` provides an abstraction for defining simple key -> value stores with asynchronous get and put functionality. We propose using `AsyncCache` for implementations with medium to high (> millisecond) overhead, for example an on-disk cache.

We provide an initial extension of `Cache` in the form of `InMemoryCache`. `InMemoryCache` provides fast, in-memory caching that can be later used to memoize more expensive cache accesses. `InMemoryCache` also provides a custom constructor `InMemoryCache.from_env_var` that can be used to pre-populate the in-memory cache, which will be helpful for enabling determinism in the future.

We also provides extensions of `AsyncCache`. `OnDiskCache` subclasses `AsyncCache` and serves as a generic on-disk caching implementation with atomic, write-once guarantees. `OnDiskCache` is semi-generic, allowing subclassing to alter the output directory. `InductorOnDiskCache` subclasses `OnDiskCache` to create an Inductor-specific on-disk cache that outputs to Inductor's default caching directory.

Test Plan:
`Cache` Tests:
1. Get -> Set -> Get
- Checks that `get(key)` returns `None` when `key` is not cached, and that after calling `put(key, value)` subsequent `get(key)` calls return `value`
2. Set -> Set
- Checks that with duplicated `set(key, value)` calls only the initial call is successful
3. From env var
- Checks that constructing an `InMemoryCache` from an environment variable works.

`AsyncCache` Tests:
1. Get -> Set -> Get
- Same as `Cache` test, but checks both with synchronous and asynchronous execution
2. Set -> Set
- Same as `Cache` test, but checks both with synchronous and asynchronous execution
3. Set -> Set Concurrent
- Checks that of two concurrent `set(key, value)` operations, only one passes

```
cd ~/fbsource/fbcode && buck test mode/opt //caffe2/test/inductor:pcache
```

 {F1981926248}

Rollback Plan:

Differential Revision: D82269762

### DIFF
--- a/torch/_inductor/pcache.py
+++ b/torch/_inductor/pcache.py
@@ -1,0 +1,330 @@
+from __future__ import annotations
+
+from functools import cached_property
+from os import getenv
+from pathlib import Path
+from typing import Generic, TYPE_CHECKING, TypeVar
+from typing_extensions import override, Self
+
+from torch.utils._filelock import FileLock
+
+
+if TYPE_CHECKING:
+    from concurrent.futures import Future, ThreadPoolExecutor
+
+
+Key = TypeVar("Key")
+Value = TypeVar("Value")
+
+
+class Cache(Generic[Key, Value]):
+    """
+    Abstract base class for cache implementations.
+
+    Provides the interface for basic synchronous get and insert methods for storing and retrieving data.
+    Subclasses must implement both methods.
+
+    Note:
+        - Not guaranteed to be thread-safe.
+        - For asynchronous and thread-safe cache, see `AsyncCache`.
+
+    Methods:
+        get(key): Retrieve a value by key.
+        insert(key, value): Insert a value if the key does not already exist.
+    """
+
+    def get(self: Self, key: Key) -> Value | None:
+        """
+        Retrieve the value associated with the given key from the cache.
+
+        Args:
+            key (Key): The key used to query the cache.
+
+        Returns:
+            Value | None: The value associated with the key, or None if not found.
+
+        Raises:
+            NotImplementedError: If not implemented by a subclass.
+        """
+        raise NotImplementedError
+
+    def insert(self: Self, key: Key, value: Value) -> bool:
+        """
+        Store the given value in the cache with the associated key if the key does
+        not already exist in the cache, otherwise do nothing.
+
+        Args:
+            key (Key): The key to associate with the value.
+            value (Value): The value to be stored in the cache.
+
+        Returns:
+            bool: True if the value was stored successfully, False if the key already exists.
+
+        Raises:
+            NotImplementedError: If not implemented by a subclass.
+        """
+        raise NotImplementedError
+
+
+class InMemoryCache(Cache[str, bytes]):
+    """
+    In-memory cache implementation.
+
+    Stores cache data in a dictionary for fast lookups and insertions.
+    Not thread-safe.
+    """
+
+    def __init__(self: Self) -> None:
+        """
+        Initialize the in-memory cache.
+        """
+        self._cache: dict[str, bytes] = {}
+
+    @override
+    def get(self: Self, key: str) -> bytes | None:
+        """
+        Retrieve the value associated with the given key from the cache.
+
+        Args:
+            key (str): The key used to query the cache.
+
+        Returns:
+            bytes | None: The value associated with the key, or None if not found.
+        """
+        return self._cache.get(key)
+
+    @override
+    def insert(self: Self, key: str, value: bytes) -> bool:
+        """
+        Store the given value in the cache with the associated key if the key does
+        not already exist in the cache, otherwise do nothing.
+
+        Args:
+            key (str): The key to associate with the value.
+            value (bytes): The value to be stored in the cache.
+
+        Returns:
+            bool: True if the value was stored successfully, False if the key already exists.
+        """
+        if key in self._cache:
+            return False
+        self._cache[key] = value
+        return True
+
+    @classmethod
+    def from_env_var(cls, env_var: str) -> Self:
+        """
+        Create a new in-memory cache instance from an environment variable.
+
+        The environment variable should contain key-value pairs separated by ';',
+        with each pair formatted as 'key,value'. The value should be a string
+        representation of bytes (e.g., b'...').
+
+        Args:
+            env_var (str): The environment variable containing cache data.
+
+        Returns:
+            InMemoryCache: A new in-memory cache instance populated with data from the environment variable.
+
+        Raises:
+            ValueError: If a key is associated with two distinct values, or if the environment variable
+                is malformed (e.g., missing comma, value not a bytes string).
+        """
+        cache: Self = cls()
+        env_val: str | None = getenv(env_var, None)
+
+        if env_val is not None:
+            for kv_pair in env_val.split(";"):
+                if not kv_pair:
+                    # can happen if env_val is an empty string, or ends with ;
+                    continue
+                try:
+                    key, raw_value = kv_pair.split(",", 1)
+                except ValueError as err:
+                    raise ValueError(
+                        f"Malformed kv_pair {kv_pair!r} in env_var {env_var!r}, missing comma separator!"
+                    ) from err
+                # check that raw_value is a str repr of bytes
+                if (not raw_value.startswith("b'")) or (not raw_value.endswith("'")):
+                    raise ValueError(
+                        f"Malformed value {raw_value!r} in kv_pair {kv_pair!r}, expected b'...' format!"
+                    )
+                # remove b' prefix and ' suffix
+                str_value = raw_value[2:-1]
+                try:
+                    # make sure the value is legitimately encoded
+                    value = bytes([ord(char) for char in str_value])
+                except ValueError as err:
+                    raise ValueError(
+                        f"Malformed value {raw_value!r} in kv_pair {kv_pair!r}!"
+                    ) from err
+                # duplicates are ok, so long as the key does not point to two distinct values
+                if (not cache.insert(key, value)) and (cache.get(key) != value):
+                    raise ValueError(
+                        f"Duplicated values for key {key!r}, got {cache.get(key)!r} and {value!r}!"
+                    )
+
+        return cache
+
+
+class AsyncCache(Cache[Key, Value]):
+    """
+    Abstract base class for asynchronous, thread-safe cache implementations.
+
+    Provides synchronous get/insert methods and additional asynchronous (_async) methods
+    for concurrent access using a ThreadPoolExecutor. All methods are thread-safe.
+
+    Note:
+        - Use this class or its subclasses when thread safety or async access is required.
+        - The _async methods return concurrent.futures.Future objects.
+
+    Methods:
+        get(key): Retrieve a value by key.
+        get_async(key, executor): Asynchronously retrieve a value by key.
+        insert(key, value): Insert a value.
+        insert_async(key, value, executor): Asynchronously insert a value.
+    """
+
+    def get_async(
+        self: Self, key: Key, executor: ThreadPoolExecutor
+    ) -> Future[Value | None]:
+        """
+        Retrieve the value associated with the given key from the cache asynchronously.
+
+        Args:
+            key (Key): The key used to query the cache.
+            executor (ThreadPoolExecutor): The executor to use for asynchronous execution.
+
+        Returns:
+            Future[Value | None]: A Future representing the result of the asynchronous operation.
+        """
+        return executor.submit(self.get, key)
+
+    def insert_async(
+        self: Self, key: Key, value: Value, executor: ThreadPoolExecutor
+    ) -> Future[bool]:
+        """
+        Store the given value in the cache with the associated key if the key does
+        not already exist in the cache, otherwise do nothing, asynchronously.
+
+        Args:
+            key (Key): The key to associate with the value.
+            value (Value): The value to be stored in the cache.
+            executor (ThreadPoolExecutor): The executor to use for asynchronous execution.
+
+        Returns:
+            Future[bool]: A Future representing the result of the asynchronous operation.
+        """
+        return executor.submit(self.insert, key, value)
+
+
+class OnDiskCache(AsyncCache[str, bytes]):
+    """
+    Abstract base class for on-disk cache implementations.
+
+    Provides synchronous and asynchronous get/insert methods for storing and retrieving data on disk.
+    All methods are thread-safe.
+
+    Methods:
+        get(key): Retrieve a value by key from disk.
+        get_async(key, executor): Asynchronously retrieve a value by key from disk.
+        insert(key, value): Insert a value on disk.
+        insert_async(key, value, executor): Asynchronously insert a value on disk.
+    """
+
+    @property
+    def base_dir(self: Self) -> Path:
+        """
+        Get the base directory for the on-disk cache.
+
+        Returns:
+            Path: The base directory for the on-disk cache.
+
+        Raises:
+            NotImplementedError: If not implemented by a subclass.
+        """
+        raise NotImplementedError
+
+    def _fpath_from_key(self: Self, key: str) -> Path:
+        """
+        Get the file path associated with the given key.
+
+        Args:
+            key (str): The key used to query the cache.
+
+        Returns:
+            Path: The file path associated with the key.
+        """
+        return self.base_dir / key
+
+    def _flock_from_fpath(self: Self, fpath: Path) -> FileLock:
+        """
+        Get the file lock associated with the given file path.
+
+        Args:
+            fpath (Path): The file path to lock.
+
+        Returns:
+            FileLock: The file lock associated with the file path.
+        """
+        return FileLock(str(fpath) + ".lock")
+
+    @override
+    def get(self: Self, key: str) -> bytes | None:
+        """
+        Retrieve the value associated with the given key from the cache on disk.
+
+        Args:
+            key (str): The key used to query the cache.
+
+        Returns:
+            bytes | None: The value associated with the key, or None if not found.
+        """
+        fpath = self._fpath_from_key(key)
+        flock = self._flock_from_fpath(fpath)
+        with flock:
+            return fpath.read_bytes() if fpath.is_file() else None
+
+    @override
+    def insert(self: Self, key: str, value: bytes) -> bool:
+        """
+        Store the given value in the cache with the associated key on disk.
+
+        Args:
+            key (str): The key to associate with the value.
+            value (bytes): The value to be stored in the cache.
+
+        Returns:
+            bool: True if the value was stored successfully, False if the key already exists.
+        """
+        fpath = self._fpath_from_key(key)
+        flock = self._flock_from_fpath(fpath)
+        fpath.parent.mkdir(parents=True, exist_ok=True)
+        try:
+            # "x" mode is exclusive creation, meaning the file will be created
+            # iff the file does not already exist (atomic w/o overwrite)
+            with flock as _, open(fpath, "xb") as fp:
+                fp.write(value)
+        except FileExistsError:
+            return False
+        return True
+
+
+class InductorOnDiskCache(OnDiskCache):
+    """
+    On-disk cache implementation for Inductor.
+
+    Uses the default cache directory provided by Inductor.
+    """
+
+    @cached_property
+    def base_dir(self: Self) -> Path:
+        """
+        Get the base directory for the on-disk cache.
+
+        Returns:
+            Path: The base directory for the on-disk cache.
+        """
+        from torch._inductor.runtime.runtime_utils import default_cache_dir
+
+        return Path(default_cache_dir(), "pcache")
--- a/test/inductor/test_pcache.py
+++ b/test/inductor/test_pcache.py
@@ -1,0 +1,297 @@
+# Owner(s): ["module: inductor"]
+from __future__ import annotations
+
+from concurrent.futures import Future, ThreadPoolExecutor
+from os import environ
+from random import randint
+from typing import TYPE_CHECKING
+from typing_extensions import Self
+
+from torch._inductor import pcache
+from torch._inductor.test_case import run_tests, TestCase
+from torch.testing._internal.common_utils import (
+    instantiate_parametrized_tests,
+    parametrize,
+)
+
+
+if TYPE_CHECKING:
+    from collections.abc import Generator
+
+
+str_key_gen: Generator[str, None, None] = (
+    f"dummy_key_{randint(0, 100000)}" for _ in iter(int, 1)
+)
+bytes_value_gen: Generator[bytes, None, None] = (
+    f"dummy_value_{randint(0, 100000)}".encode() for _ in iter(int, 1)
+)
+
+
+Caches: list[type[pcache.Cache]] = [pcache.InMemoryCache]
+AsyncCaches: list[type[pcache.AsyncCache]] = [pcache.InductorOnDiskCache]
+
+
+@instantiate_parametrized_tests
+class CacheTest(TestCase):
+    @parametrize("Cache", Caches)
+    def test_get_insert_get(self: Self, Cache: type[pcache.Cache]) -> None:
+        key: str = next(str_key_gen)
+        value: bytes = next(bytes_value_gen)
+
+        cache: pcache.Cache = Cache()
+
+        # make sure our key is fresh
+        while cache.get(key) is not None:
+            key = next(str_key_gen)
+
+        # first get should return None, no hit
+        self.assertIsNone(cache.get(key))
+        # insert should return True, having set key -> value
+        self.assertTrue(cache.insert(key, value))
+        # second get should return value, hit
+        self.assertEqual(cache.get(key), value)
+
+    @parametrize("Cache", Caches)
+    def test_insert_insert(self: Self, Cache: type[pcache.Cache]) -> None:
+        key: str = next(str_key_gen)
+        value: bytes = next(bytes_value_gen)
+
+        cache: pcache.Cache = Cache()
+
+        if cache.get(key) is None:
+            # if key isn't already cached, cache it
+            self.assertTrue(cache.insert(key, value))
+
+        # second insert should not update the value
+        self.assertFalse(cache.insert(key, value))
+
+    def test_in_memory_cache_from_env_var(
+        self: Self, Cache: type[pcache.InMemoryCache] = pcache.InMemoryCache
+    ) -> None:
+        key_1: str = next(str_key_gen)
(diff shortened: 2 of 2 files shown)


torch/_inductor/pcache.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        Retrieve the value associated with the given key from the cache.
+        Args:
+            key: The key used to query the cache.
+        Returns:
+            The value associated with the key, or None if not found.
+        """
+        raise NotImplementedError
+
+    def put(self: Self, key: Key, value: Value) -> bool:
```
masnesral: Would failing a put be an exceptional case? Maybe throw an exception on failure instead of requiring users to check the return value?

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        Store the given value in the cache with the associated key.
+        Args:
+            key: The key to associate with the value.
+            value: The value to be stored in the cache.
+        Returns:
+            True if the value was stored successfully, False otherwise.
+        """
+        raise NotImplementedError
+
```
masnesral, line 67: What about a method to clear the cache?

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,247 @@
...
+        Raises:
+            ValueError: If a key is associated with two distinct values.
+        """
+        cache: Self = cls()
+        env_val: str | None = getenv(env_var, None)
+
+        if env_val is not None:
+            for kv_pair in env_val.split(";"):
+                key, value = kv_pair.split(",")
+                # value is a str repr of bytes, remove b' prefix and ' suffix
```
masnesral: It seems weird to me that you need to manipulate the string like this. Is that really needed? For example, what if we convert from bytes to string first before the splitting? Does that remove the need for stripping off the prefix and suffix? Also, it seems like a little more error checking would be prudent here, e.g., if each element isn't exactly a tuple separated by comma?

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,247 @@
...
+    Stores cache data in a dictionary for fast lookups and insertions.
+    """
+
+    def __init__(self: Self) -> None:
+        """
+        Initialize the in-memory cache.
+        """
+        self._cache: dict[Key, Value] = {}
+
+    def get(self: Self, key: Key) -> Value | None:
```
masnesral: Is this `| None` syntax compatible with all python versions we have to support? I don't know when it got introduced, but it seems like most of the codebase uses `Optional`

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,247 @@
...
+        Retrieve the value associated with the given key from the cache synchronously.
+        Args:
+            key: The key used to query the cache.
+        Returns:
+            The value associated with the key, or None if not found.
+        """
+        raise NotImplementedError
+
+    def get(
+        self: Self, key: Key, executor: ThreadPoolExecutor | None = None
```
masnesral: What's the thinking behind passing the executor? I would have expected that to be purely an implementation detail of the concrete implementation that implements this interface. What if it uses some other kind of executor? Also, maybe a dumb question, but could possibly use actual `asyncio` Python rather than executors and Futures?

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,247 @@
...
+    ) -> Future[Value | None] | Value | None:
```
masnesral: What do you think about always returning a Future instead? An interface like this is gonna make it so that callers are checking the type of return values.

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,247 @@
...
+        """
+        Retrieve the value associated with the given key from the cache.
+        Args:
+            key: The key used to query the cache.
+            executor: The executor to use for asynchronous execution, or None for synchronous execution.
+        Returns:
+            The value associated with the key, or None if not found. If an executor is provided,
+            returns a Future representing the result of the asynchronous operation.
+        """
+        if executor is None:
```
masnesral: I'm not seeing the future use case for sync puts and gets when using an AsyncCache. If the call-site must do that, they can just wait on the future? (or the analogous call if we were to use actual asyncio)

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,247 @@
...
+            value: The value to be stored in the cache.
+        Returns:
+            True if the value was stored successfully, False otherwise.
+        """
+        fpath = self._fpath_from_key(key)
+        fpath.parent.mkdir(parents=True, exist_ok=True)
+        try:
+            # "x" mode is exclusive creation, meaning the file will be created
+            # iff the file does not already exist (atomic w/o overwrite)
+            with open(fpath, "xb") as fp:
```
masnesral: But is this robust if multiple processes are racing for the same file? What happens if one tries to read while the other is writing? Don't we need to do the rename trick or take some FileLock?

torch/_inductor/pcache.py
aorenste: It's introduced in py3.10 which we'll be officially switching to as our minimum version later this year (but not quite yet) OR requires a "`from __future__ import annotations`" for py3.9.
(FYI at some point after the upgrade I plan on codemodding the codebase to the new syntax)

torch/_inductor/pcache.py
nmacchioni: Not very exceptional; once we implement remote caching it will be probable that training jobs with >1 rank will have two or more ranks attempting to put two different values for the same key at the same time (for example, with padmm if padded vs. unpadded is nearly identical so due to noise one rank picks pad while the other picks not pad). We'll actually need this feature later on when we implement intra-job determinism, as whoever's put succeeds can continue along but whoever's put fails needs to re-query the cache

torch/_inductor/pcache.py
nmacchioni: Yes, some caches will define a cache clear method but that is part of a later commit; we won't define a clearing method in any base class however as not all caches have the ability to be cleared (remote cache is one)

torch/_inductor/pcache.py
nmacchioni: > For example, what if we convert from bytes to string first before the splitting?

Before splitting it is already a string

> Also, it seems like a little more error checking would be prudent here, e.g., if each element isn't exactly a tuple separated by comma?

Do you have any examples of parsing/sanitizing env vals in Inductor? I would have thought it would be okay to crash if the user supplies a malformed env val. What do you think about simply wrapping the whole thing in a try except block so we can give a better error message if the env val is malformed?

torch/_inductor/pcache.py
nmacchioni: I believe that "|" is the newer format, but it does seem supported https://github.com/pytorch/pytorch/blob/main/torch/_inductor/codecache.py#L667

torch/_inductor/pcache.py
nmacchioni: > What's the thinking behind passing the executor? I would have expected that to be purely an implementation detail of the concrete implementation that implements this interface.

I decided to pass in the executor to 1) keep the executor construction logic localized to the Inductor specific code (Inductor decides if we can run asynchronously) so that if/when I want to bring this to Triton the base AsyncCache classes don't contain Inductor code and 2) executors need to be explicitly shutdown to prevent leaking resources, and I thought having a single, shared executor would be easier to track (also faster, creating a new executor per-get/put would cause overhead)

> What if it uses some other kind of executor?

I chose to only support ThreadPoolExecutors at this time, as that is what is currently used elsewhere in Inductor

> Also, maybe a dumb question, but could possibly use actual asyncio Python rather than executors and Futures?

We could use asyncio, but I don't think it has been done yet in Inductor so I wanted to keep with the norm for now

torch/_inductor/pcache.py
nmacchioni: I don't think the callee will have to check the return type, since the callee will know whether or not they have supplied an executor which will determine the return type

torch/_inductor/pcache.py
nmacchioni: Sometimes creating a ThreadPoolExecutor is not possible, for example if `TORCHINDUCTOR_COMPILE_THREADS=1`, hence we have to fallback to synchronous gets/puts

torch/_inductor/pcache.py
masnesral: Oh, so does failure here mean the case where the key already exists and not, e.g., some infra failure (like disk full or remote service down)? If so, then I would have expected that to be captured in the API. Maybe either an insert vs. put method, or some additional parameter like `allow_overwrite=True|False`

torch/_inductor/pcache.py
nmacchioni: I do believe "x" is atomic: https://docs.python.org/3/library/functions.html#open

The rename trick doesn't work in this case, as rename overwrites existing files (which kills determinism). A lock would work, but I think "x" already handles atomicity for us

nmacchioni: But just to be sure, I added a couple more tests cases for exactly that

torch/_inductor/pcache.py
nmacchioni: I made the parsing much more robust, and added a couple more test cases to check the robustness

masnesral: Oh sry; I didn't read the doc closely enough. I thought the string manipulation was because getenv was returning bytes or something. I didn't see that you expect users to supply `b'something'`. In that case it seems like you should validate that the value portion actually adheres to that format before stripping off the prefix and suffix? And yes, better error message sounds like a good way to go

torch/_inductor/pcache.py
masnesral: Maybe it's time to start modernizing and using asyncio ;)
I actually haven't used it that much myself so I guess I don't know what can of worms that risks opening up.

torch/_inductor/pcache.py
masnesral: Maybe different methods would be appropriate then? A sync vs. non-sync version with different return types? Maybe it's just me and I'm too old school; it always feels like a bad idea to return different kinds of things. It muddies the API and (I'd assume) weakens the typechecker. @aorenste is also here; maybe we can get another opinion.

torch/_inductor/pcache.py
masnesral: Sorry, maybe I'm still not getting the broader objective. I was thinking this change was gonna be the basis for a generic cache impl that could potentially be leveraged anywhere we need some in-memory, on-disk, or remote cache. You mention `TORCHINDUCTOR_COMPILE_THREADS` -- Is this new cache somehow more tightly tied to something around async_compile? I don't think the intent of that config option is meant to indicate "don't use threads anywhere" -- we have plenty of other places in the code that launch thread helpers that have nothing to do with the code in async_compile.py and don't honor that setting. 

But now that you mention `TORCHINDUCTOR_COMPILE_THREADS=1` (which is used by async_compile.triton): That's a great example of an API that's a pain. Callers have to somehow check if the result will be a future or a compile result and operate accordingly. I think it would be better if the setting for `TORCHINDUCTOR_COMPILE_THREADS` was honored appropriately by that module, but callers didn't have to know anything about it.

torch/_inductor/pcache.py
masnesral: Ok, but I was also wondering about the case where one process is writing while another is trying to read. I don't think 'x' would protect against reading a partially-written file? And what happens when there's a race and one of the clients fails to open? Are clients gonna have to wait and retry? I wonder if blocking on a FileLock would be easier? At least, I assumed that was part of the reason I see file locks in codecache.py?

torch/_inductor/pcache.py
nmacchioni: Ah yes you're totally right, now that I think about it the accepted practice is generally that insert/add don't override but put does. I'll go back and put can become insert. For now I won't be implementing any overrides as I won't have a use for it

torch/_inductor/pcache.py
nmacchioni: I added validation and better error messages, closing this out

torch/_inductor/pcache.py
nmacchioni: I'd rather go forward with ThreadPoolExecutors at this time as to not hold up the other work to figure our asyncio

torch/_inductor/pcache.py
nmacchioni: I'm not too stuck on it being a single method, I'd be happy to just change it to get_sync and get_async

torch/_inductor/pcache.py
nmacchioni: Ah ok, so at a high level the code here (Cache and AsyncCache) are just basic building blocks which will be wrapped in a future PR into something called MultiCache. MultiCache will basically just handle get/put calls with multiple backing cache implementations. Another PR will wrap MultiCache into something I'm thinking of calling PCache or maybe just InductorCache. The end goal is that InductorCache provides a function wrapper @InductorCache.record(...) that will be used to replace the existing intra-function recording in pad mm for example. So ultimately I don't expect devs to be interacting with Cache or AsyncCache directly, only through the InductorCache interface

nmacchioni: Let me send you the design doc, it is a touch outdated but it has the general idea of things captured

torch/_inductor/pcache.py
nmacchioni: I thought "x" being atomic would ensure that you can't have a partially-written file, but adding a FileLock is so easy I'm just going to take your suggestion there

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,317 @@
...
+from torch.utils._filelock import FileLock
+
+
+# Key represents data used to query from cache instances
+Key = str
+# Value represents data stored in/retrieved from cache instances
+Value = bytes
+
+
+class Cache:
```
Skylion007: Shouldn't this subclass Generic with defaults of the Key and Value TypeAliases? Would make it more flexible if we need to change it later

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,317 @@
...
+from functools import cached_property
+from os import getenv
+from pathlib import Path
+from typing_extensions import Self
+
+from torch.utils._filelock import FileLock
+
+
+# Key represents data used to query from cache instances
+Key = str
```
Skylion007: Please type these as typing_extensions.TypeAlias to make it easier for static analysis

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,317 @@
...
+            bool: True if the value was stored successfully, False if the key already exists.
+        """
+        fpath = self._fpath_from_key(key)
+        flock = self._flock_from_fpath(fpath)
+        fpath.parent.mkdir(parents=True, exist_ok=True)
+        try:
+            with flock:
+                # "x" mode is exclusive creation, meaning the file will be created
+                # iff the file does not already exist (atomic w/o overwrite)
+                with open(fpath, "xb") as fp:
```
Skylion007: Have both with contexts on one line to save indentation and improve readability that show they are the same scope

torch/_inductor/pcache.py
nmacchioni: Done; I haven't used Generic previously so please double check my understanding. I ended up using TypeVar as Generic does not take TypeAlias

torch/_inductor/pcache.py
nmacchioni: Done

nmacchioni:
@Skylion007 @masnesral do you mind taking a final look? cc @eellison

masnesral (approved):
LG; thanks.

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,315 @@
...
+    Note:
+        - Not guaranteed to be thread-safe.
+        - For asynchronous and thread-safe cache, see `AsyncCache`.
+
+    Methods:
+        get(key): Retrieve a value by key.
+        insert(key, value): Insert a value if the key does not already exist.
+    """
+
+    def get(self: Self, key: Key) -> Value | None:
```
aorenste, line 36: Currently our minimum python version is 3.9 (although we're phasing that out because the next release is 3.10+). We should probably check on that before committing this.

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,315 @@
...
+    Not thread-safe.
+    """
+
+    def __init__(self: Self) -> None:
+        """
+        Initialize the in-memory cache.
+        """
+        self._cache: dict[str, bytes] = {}
+
+    def get(self: Self, key: str) -> bytes | None:
```
aorenste, line 84: consider `@override` (here and in the later overrides)

torch/_inductor/pcache.py
aorenste: Also - by defining this as `Value | None` it means that you can't have a cache that stores a `None` value... maybe that's fine...

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,315 @@
...
+        """
+        Retrieve the value associated with the given key from the cache.
+
+        Args:
+            key (str): The key used to query the cache.
+
+        Returns:
+            bytes | None: The value associated with the key, or None if not found.
+        """
+        return self._cache.get(key, None)
```
aorenste: nit: `None` is the default

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,315 @@
...
+        not already exist in the cache, otherwise do nothing.
+
+        Args:
+            key (str): The key to associate with the value.
+            value (bytes): The value to be stored in the cache.
+
+        Returns:
+            bool: True if the value was stored successfully, False if the key already exists.
+        """
+        if key in self._cache:
```
aorenste, line 109:
```suggestion
        if key not in self._cache:
```

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,315 @@
...
+        return True
+
+    @classmethod
+    def from_env_var(cls, env_var: str) -> Self:
+        """
+        Create a new in-memory cache instance from an environment variable.
+
+        The environment variable should contain key-value pairs separated by ';',
+        with each pair formatted as 'key,value'. The value should be a string
+        representation of bytes (e.g., b'...').
```
aorenste, line 121: So the values must be literally in the form `b'abc'`, no double-quotes, no pure ascii? That seems kind of restrictive?

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,315 @@
...
+        Raises:
+            ValueError: If a key is associated with two distinct values, or if the environment variable
+                is malformed (e.g., missing comma, value not a bytes string).
+        """
+        cache: Self = cls()
+        env_val: str | None = getenv(env_var, None)
+
+        if env_val is not None:
+            for kv_pair in env_val.split(";"):
+                if kv_pair == "":
```
aorenste: nit:
```suggestion
                if not kv_pair:
```

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,315 @@
...
+                    # can happen if env_val is an empty string, or ends with ;
+                    continue
+                try:
+                    key, str_value = kv_pair.split(",")
```
aorenste:
```suggestion
                    key, str_value = kv_pair.split(",", 1)
```

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,315 @@
...
+                    ) from err
+                # check that str_value is a str repr of bytes
+                if (not str_value.startswith("b'")) or (not str_value.endswith("'")):
+                    raise ValueError(
+                        f"Malformed value {str_value} in kv_pair {kv_pair}!"
+                    )
+                # remove b' prefix and ' suffix
+                str_value = str_value[2:-1]
+                value = str_value.encode()
+                # duplicatess are ok, so long as the key does not point to two distinct values
```
aorenste: nit, spelling:
```suggestion
                # duplicates are ok, so long as the key does not point to two distinct values
```

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,315 @@
...
+        if env_val is not None:
+            for kv_pair in env_val.split(";"):
+                if kv_pair == "":
+                    # can happen if env_val is an empty string, or ends with ;
+                    continue
+                try:
+                    key, str_value = kv_pair.split(",")
+                except ValueError as err:
+                    raise ValueError(
+                        f"Malformed kv_pair {kv_pair} in env_var {env_var}!"
```
aorenste: nit
I would report the string as repr and also indicate a little more "why":
```suggestion
                        f"Malformed kv_pair {kv_pair!r} in env_var {env_var} (missing comma)!"
```

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,315 @@
...
+                    ) from err
+                # check that str_value is a str repr of bytes
+                if (not str_value.startswith("b'")) or (not str_value.endswith("'")):
+                    raise ValueError(
+                        f"Malformed value {str_value} in kv_pair {kv_pair}!"
```
aorenste: nit (same as above)
```suggestion
                        f"Malformed value {str_value!r} in kv_pair {kv_pair} (missing byte string notation)!"
```

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,315 @@
...
+                    )
+                # remove b' prefix and ' suffix
+                str_value = str_value[2:-1]
+                value = str_value.encode()
```
aorenste: This whole byte encoding thing seems weird - like how would I even put in a non-ascii-7 value? If I want something like char 200 what would I put? It's not at all portable and this encode won't do what you expect. Maybe the input should just be ascii only? Otherwise you really need to handle escape sequences.

torch/_inductor/pcache.py
nmacchioni: @aorenste we don't overwrite on insert, so key in self._cache is correct here

aorenste: Oops - my mistake.

aorenste (approved):
The rest of my comments were basically nits (although you should consider the ascii/bin thing) so approving.

torch/_inductor/pcache.py
nmacchioni: > Currently our minimum python version is 3.9 (although we're phasing that out because the next release is 3.10+). We should probably check on that before committing this.

I double checked, | is already used elsewhere https://github.com/pytorch/pytorch/blob/main/torch/_inductor/codecache.py#L667

> Also - by defining this as Value | None it means that you can't have a cache that stores a None value... maybe that's fine...

None should be cacheable regardless; in our case the plan is to use Value = bytes so we can just pickle None, if someone wants to use another type for Value they'll have to figure a way to encode None

torch/_inductor/pcache.py
nmacchioni: Done, thanks for the suggestion

torch/_inductor/pcache.py
nmacchioni: Fixed

torch/_inductor/pcache.py
aorenste: That other code has:
```
from __future__ import annotations
```
which enables it in py3.9.

torch/_inductor/pcache.py
nmacchioni: insert doesn't overwrite prior values, so "if key in self._cache: return False" is correct here

torch/_inductor/pcache.py
nmacchioni: Fixed

torch/_inductor/pcache.py
nmacchioni: Fixed

torch/_inductor/pcache.py
nmacchioni: Fixed, thanks haha that was metamate

nmacchioni: Thanks! Fixed

torch/_inductor/pcache.py
nmacchioni: Gotcha, fixed this and the one below

torch/_inductor/pcache.py
nmacchioni: Ah good catch, fixing

torch/_inductor/pcache.py
```diff
@@ -0,0 +1,319 @@
+from concurrent.futures import Future, ThreadPoolExecutor
+from functools import cached_property
+from os import getenv
+from pathlib import Path
+from typing import Generic, override, TypeVar
```
aorenste: override is py3.12 - so you need to get it from typing_extensions

torch/_inductor/pcache.py
nmacchioni: The values must be pure ascii, b'\xff' is more of the type that we'd expect

torch/_inductor/pcache.py
nmacchioni: Good catch, .encode() here is not correct since str_value should already be escaped. I think bytes([ord(char) for char in str_value]) is more appropriate here, as long as we have a good error message to go with it. Making that change + adding new test

masnesral: What is the use case for passing via env var anyway? Without knowing how you'll use it, I would have expected it would be more ergonomic to initialize from the contents of a file. Env vars probably have a max length. Would that be a problem?

aorenste: Instead of stripping the b and quotes manually you could instead use `ast.literal_eval`:
```
>>> ast.literal_eval("b'abc\\x41'")
b'abcA'
```

nmacchioni: > What is the use case for passing via env var anyway? Without knowing how you'll use it, I would have expected it would be more ergonomic to initialize from the contents of a file. Env vars probably have a max length. Would that be a problem?

End goal is to enable determinism in the compiler, for example we'll have the inductor caching mechanism spit out a history of key -> value mappings from one run which you can plug into the next. It hasn't been fully scoped out yet, but I do think having it point to a file has its own consequences (PGs have to add a file to their resources, to get it included with the binary)

nmacchioni: > Instead of stripping the b and quotes manually you could instead use `ast.literal_eval`:
> 
> ```
> >>> ast.literal_eval("b'abc\\x41'")
> b'abcA'
> ```

I think literal_eval has its own problems, for one we'd still have to check that there is a b' prefix and ' suffix otherwise we'd have to check the return type to make sure it is not evaluating another type

PR #162781, opened 2025-09-12
[inductor][heuristics] add kernel template params

coconutruben:
* #163206
* #163209
* __->__ #162781

# why

- enable a clear interface for kernel templates to declare all their
  instantiation parameters and any potential defaults
- simplify KernelTemplateChoice to just have a single params, and not kwargs and extra_kwargs

# what

- KernelTemplateParams interface
- placeholder implementation where we just pass through a dict

# testing

- existing ci tests


torch/_inductor/template_heuristics/params.py
```diff
@@ -0,0 +1,42 @@
...
+    def to_serializeable_dict(self) -> dict[str, Any]:
+        """Convert params to serializable dict for storage/caching"""
+
+    @classmethod
+    @abstractmethod
+    def from_dict(cls, data: dict[str, Any]) -> KernelTemplateParams:
+        """Create params instance from dict"""
+
+
+class DictKernelTemplateParams(KernelTemplateParams):
```
jansel, line 24: When will e need something other than a dict here?  Is the abstract base used for anything?

torch/_inductor/template_heuristics/params.py
```diff
@@ -0,0 +1,42 @@
...
+    # has time to define their own params class, with meaningful
+    # defaults etc.
+
+    def __init__(self, kwargs: dict[str, Any]):
+        self.kwargs = kwargs
+
+    def to_kwargs(self) -> dict[str, Any]:
+        return self.kwargs.copy()
+
+    def to_serializeable_dict(self) -> dict[str, Any]:
```
jansel, line 37: When will this be different from the above?

jansel (changes requested):
When will this be something other than a dict?  Not totally clear why we need to wrap the dict in a class...

torch/_inductor/template_heuristics/params.py
coconutruben: I want this to be a dataclass for the existing templates, so that we can explicitly declare the full set of parameters and any defaults in them
The dict implementation is to allow for templates to be onboarded while the interface is already in place

torch/_inductor/template_heuristics/params.py
coconutruben: https://github.com/pytorch/pytorch/blob/main/torch/_inductor/codegen/cuda/serialization.py

when we onboard cutlass, the kwargs just return a cutlass `GemmOperation` object, whereas this would return the serialization of that

jansel: When would `to_kwargs` and `to_serializeable_dict` be different?

coconutruben: when we onboard cutlass
- `to_kwargs` will be a dict of the keys `op: GemmOperation` and `swizzle: int`. 
- `to_serializeable_dict` will be the output of the serialization of `op` to a dict https://github.com/pytorch/pytorch/blob/main/torch/_inductor/codegen/cuda/serialization.py#L60

jansel (approved):

PR #162784, opened 2025-09-12
Port shared_ptr optimization in std::shared_ptr to intrusive_ptr

mcfi:
Summary:
Please see D21021645 for details about the optimization and why it's beneficial.

A similar change has been added to libstdc++ as well, see https://github.com/gcc-mirror/gcc/commit/dbf8bd3c2f2cd2d27ca4f0fe379bd9490273c6d7

Rollback Plan:

Reviewed By: yfeldblum

Differential Revision: D81960754

### DIFF
--- a/c10/util/intrusive_ptr.h
+++ b/c10/util/intrusive_ptr.h
@@ -283,23 +283,55 @@ class intrusive_ptr final {
   }
 
   void reset_() noexcept {
-    if (target_ != NullType::singleton() &&
-        detail::atomic_refcount_decrement(target_->refcount_) == 0) {
-      // See comment above about weakcount. As long as refcount>0,
-      // weakcount is one larger than the actual number of weak references.
-      // So we need to decrement it here.
-      bool should_delete =
-          target_->weakcount_.load(std::memory_order_acquire) == 1;
-      if (!should_delete) {
-        // justification for const_cast: release_resources is basically a
-        // destructor and a destructor always mutates the object, even for const
-        // objects. NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
-        const_cast<std::remove_const_t<TTarget>*>(target_)->release_resources();
-        should_delete =
-            detail::atomic_weakcount_decrement(target_->weakcount_) == 0;
+    if (target_ != NullType::singleton()) {
+#if defined(__linux__) && (defined(__aarch64__) || defined(__x86_64__))
+      if constexpr (
+          std::atomic<uint64_t>::is_always_lock_free &&
+          std::atomic<uint32_t>::is_always_lock_free &&
+          sizeof(std::atomic<uint64_t>) == 8 &&
+          sizeof(std::atomic<uint32_t>) == 4) {
+        auto both_counts_ =
+            reinterpret_cast<std::atomic<uint64_t>*>(&target_->refcount_);
+        TORCH_INTERNAL_ASSERT_DEBUG_ONLY(
+            (reinterpret_cast<std::uintptr_t>(both_counts_) %
+             sizeof(std::atomic<uint64_t>)) == 0 &&
+            (reinterpret_cast<std::uintptr_t>(&target_->weakcount_) -
+             reinterpret_cast<std::uintptr_t>(both_counts_)) ==
+                sizeof(std::atomic<uint32_t>));
+        // 0x100000001ULL is a 64-bit number combination of both the refcount_
+        // and weakcount_ being 1.
+        constexpr uint64_t unique_ref_ = 0x100000001ULL;
+        if (both_counts_->load(std::memory_order_acquire) == unique_ref_) {
+          // Both counts are 1, so there are no weak references and
+          // we are releasing the last strong reference. No other
+          // threads can observe the effects of this target_ deletion
+          // call (e.g. calling use_count()) without a data race.
+          target_->refcount_.store(0, std::memory_order_relaxed);
+          delete target_;
+          return;
+        }
       }
-      if (should_delete) {
-        delete target_;
+#endif
+
+      if (detail::atomic_refcount_decrement(target_->refcount_) == 0) {
+        // See comment above about weakcount. As long as refcount>0,
+        // weakcount is one larger than the actual number of weak references.
+        // So we need to decrement it here.
+        bool should_delete =
+            target_->weakcount_.load(std::memory_order_acquire) == 1;
+        if (!should_delete) {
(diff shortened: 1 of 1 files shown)


c10/util/intrusive_ptr.h
```diff
@@ -283,23 +283,55 @@ class intrusive_ptr final {
...
-            detail::atomic_weakcount_decrement(target_->weakcount_) == 0;
+    if (target_ != NullType::singleton()) {
+#if defined(__linux__) && (defined(__aarch64__) || defined(__x86_64__))
+      if constexpr (
+          std::atomic<uint64_t>::is_always_lock_free &&
+          std::atomic<uint32_t>::is_always_lock_free &&
+          sizeof(std::atomic<uint64_t>) == 8 &&
+          sizeof(std::atomic<uint32_t>) == 4) {
+        auto both_counts_ =
+            reinterpret_cast<std::atomic<uint64_t>*>(&target_->refcount_);
```
swolchok, line 294: dereferencing the result of this has undefined behavior.

mcfi: In general, yes, but this change is limited to linux on arm64 and x64, on which the undefined behavior is "defined" and it's safe to do it. This optimization has been adopted in libstdc++ and I'm simply trying to port the change there to intrusive_ptr.

swolchok: > on which the undefined behavior is "defined" and it's safe to do it

Do you have a reference for this? I understand that the instructions will work if issued; it's the strict aliasing violation I'm worried about.

mcfi: Yes, please see https://github.com/gcc-mirror/gcc/commit/dbf8bd3c2f2cd2d27ca4f0fe379bd9490273c6d7

swolchok: "libstdc++ does it" does not necessarily imply "gcc agreed to keep the behavior defined" and doesn't seem to imply "clang/llvm agreed to keep the behavior defined". You can't do reinterpret_casts that aren't on the list of allowable reinterpret_casts and then dereference the resulting pointer.

yfeldblum: Per the language spec, this is indeed UB. However, it is a measurable and significant improvement (internal experiments show this). Unfortunately, we sometimes have to sacrifice strict adherence in those specific cases where the code is *in practice* correct according to the *in practice* compiler and architecture behaviors and where the wins are significant.

I think the object code which gets generated is defined behavior according to the hardware architecture memory models.

swolchok: I guess it's hard to argue that it's good enough for libstdc++ but not good enough for pytorch. I also don't think we ever write to the refcounts through a pointer which is where strict aliasing would really start to bite us, and since we need load-acquire we also can't do something like try to generate the load with a memcpy + compiler optimizations. I'll unblock this.

swolchok (approved):

yfeldblum:
@swolchok I like it. That would remove the need for undefined behavior. They might not be able to do it in libstdc++ because it's not abi-compatible and that's something they specifically care about, but it should be fine in pytorch.

mcfi:
> @mcfi it does occur to me that the existing code assumes refcounts never overflow a 32-bit int anyway, so we might be able to do this with well-defined behavior by fusing the refcounts into a single std::atomic<uint64_t> for real. The code change is a bit more invasive (you have to define an interface to the fused thing that wraps a std::atomic<uint64_t> on the platforms you want to affect in this diff and wraps the existing pair of 32-bit ints otherwise), but should be well-defined. (since neither count overflows, we can just increment/decrement weakcount by adding/subtracting 0x100000000UL.) What do you think?

> @swolchok I like it. That would remove the need for undefined behavior. They might not be able to do it in libstdc++ because it's not abi-compatible and that's something they specifically care about, but it should be fine in pytorch.

It's indeed a good idea that can remove undefined behavior and simplify the code. However, my concern is that adding/subtracting 0x100000000UL cannot be done in one instruction, so performance of weak pointers may downgrade. Moreover, if the two counters are merged in one, contention might be higher than before, which downgrades performance as well. Is this task something good for the open-source community to pick up? :)

c10/util/intrusive_ptr.h
Skylion007: Hmm c10::bitcast won't make this defined either, right?

Skylion007:
> > @mcfi it does occur to me that the existing code assumes refcounts never overflow a 32-bit int anyway, so we might be able to do this with well-defined behavior by fusing the refcounts into a single std::atomic<uint64_t> for real. The code change is a bit more invasive (you have to define an interface to the fused thing that wraps a std::atomic<uint64_t> on the platforms you want to affect in this diff and wraps the existing pair of 32-bit ints otherwise), but should be well-defined. (since neither count overflows, we can just increment/decrement weakcount by adding/subtracting 0x100000000UL.) What do you think?
> 
> 
> 
> > @swolchok I like it. That would remove the need for undefined behavior. They might not be able to do it in libstdc++ because it's not abi-compatible and that's something they specifically care about, but it should be fine in pytorch.
> 
> 
> 
> It's indeed a good idea that can remove undefined behavior and simplify the code. However, my concern is that adding/subtracting 0x100000000UL cannot be done in one instruction, so performance of weak pointers may downgrade. Moreover, if the two counters are merged in one, contention might be higher than before, which downgrades performance as well. Is this task something good for the open-source community to pick up? :)

Does godbolt show that compilers cannot fuse this into one instruction?

c10/util/intrusive_ptr.h
Skylion007: My concern here is that this is valid UB behavior in stdlib because stdlib controls the other stdlib objects, PyTorch does not though? What if we are using the clang stdlib on Linux for example?

mcfi: The optimization is only enabled if all static conditions hold and the runtime assertions can also help catch issues. Plus, the code is limited to linux on x64 and arm64 of which we know the ABI and it's safe to use.

mcfi:
> @mcfi The cost is likely negligible. Worth double checking, of course. But the costs we care about here are primarily interlocked instructions, barrier instructions, seq-cst, and unpredictable branch conditions or jump targets. Changing the representation would affect none of these.
> 
> Incrementing by 1ull<<32 would probably just be a larger-encoded instructions (x86) or an extra math instruction (arm). Would probably be lost in the pipeline. (Still worth checking. Because avoiding UB is a worthy goal.)

Sounds good, I’ll explore that next week.

PR #162794, opened 2025-09-12
[async_tp] Support mm+rs with scatter_dim matmul K by sharding B

IvanKobzarev:
* #163069
* #163068
* __->__ #162794

Current state: Shape mismatch failure when mm+rs on the last mm scatter dim.

Adding separate path to handle lastdim for aten.mm, scaled_mm should be handled similarly, but needs additional PR.
So disabling scaled_mm case with filter matmul function.

Adding inductor.config for this change that is True by default for fast debuggability of new path.


fegin (approved):
LGTM, but I'm wondering do we actually want a config or always enable this. Async TP ops already have logic to dispatch to a cublas version if kernels are small.

fegin:
The failing test is not related as async TP doesn't use NVSHMEM.

IvanKobzarev:
> LGTM, but I'm wondering do we actually want a config or always enable this. Async TP ops already have logic to dispatch to a cublas version if kernels are small.

Agree, I think I should remove config.

PR #162799, opened 2025-09-12
[AOTI] raise PyTorchStreamWriter open failed error code on windows

xuhancn:
When I debug AOTI UT: `TestAOTInductorPackage_cpu::test_add`.  I found it didn't output the verbose error code, when PyTorchStreamWriter open failed.

This PR add the verbose error code output for debug. Local test shows as below:
image

The error code is 32, we can check the Windows error code 32 at https://learn.microsoft.com/en-us/windows/win32/debug/system-error-codes--0-499- 
```
ERROR_SHARING_VIOLATION
32 (0x20)
The process cannot access the file because it is being used by another process.
```

This issue is caused by the file is opened by another process. I fixed same issue in zip open as PR: https://github.com/pytorch/pytorch/pull/162617 But still no idea how to open file with shared access in `std::ofstream`. I will continue to researching it. 

cc @peterjc123 @mszhanyi @skyline75489 @nbcsm @iremyux @Blackhex @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10

### DIFF
--- a/caffe2/serialize/inline_container.cc
+++ b/caffe2/serialize/inline_container.cc
@@ -27,6 +27,10 @@
 #include "caffe2/serialize/versions.h"
 #include "miniz.h"
 
+#ifdef _WIN32
+#include <Windows.h>
+#endif // _WIN32
+
 namespace caffe2 {
 namespace serialize {
 constexpr std::string_view kDebugPklSuffix(".debug_pkl");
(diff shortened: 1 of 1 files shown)


caffe2/serialize/inline_container.cc
```diff
@@ -711,21 +715,35 @@ void PyTorchStreamWriter::setup(const string& file_name) {
   if (archive_name_.size() == 0) {
     CAFFE_THROW("invalid file name: ", file_name);
   }
+
+  const std::string dir_name = parentdir(file_name);
+  if (!dir_name.empty()) {
+    struct stat st;
+    bool dir_exists =
+        (stat(dir_name.c_str(), &st) == 0 && (st.st_mode & S_IFDIR));
+    TORCH_CHECK(
```
Skylion007, line 724: We have to have a TORCH_CHECK_IO wrapper or wrapper for file not found error. Should be pretty easy to bind to exception type too

xuhancn: Hi @Skylion007 these code is existing code, I just move its place to pass the failed UTs: https://github.com/pytorch/pytorch/actions/runs/17673500150/job/50232301673

Skylion007: Fair, we should fix in subsequent PR

jansel (approved):

PR #162807, opened 2025-09-12
Expunge use of filesystem header from distributed

ezyang:
* __->__ #162807

The fbsource//arvr/mode/embedded/linux/clang-aarch64-debug internal build config is on too old a version of Clang to have full filesystem support. So until it is upgraded, lets have our own portability wrapper here.

Although the APIs mostly line up, I didn't bother implementing the entire path abstraction. So the use sites do need to change slightly.

Authored with claude code.


c10/util/FileSystem.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+#ifndef _WIN32
+#include <unistd.h>
+#else
+#include <direct.h>
+#include <io.h>
+#endif
+
+namespace c10 {
+namespace filesystem {
```
malfet, line 14: We are in C++17 world now, aren't we?
```suggestion
namespace c10::filesystem {
```

c10/util/FileSystem.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+#include <sys/stat.h>
+#include <cerrno>
+
+#ifndef _WIN32
+#include <unistd.h>
+#else
+#include <direct.h>
+#include <io.h>
+#endif
```
malfet, line 11:
```suggestion
#include <filesystem>
```

malfet (commented):
If you need all of that to just pass the compiler sanity, why not do something like
```cpp
static inline bool exists(const std::string& path ) {
#ifdef OVRSOURCE
  throw std::runtime_error("Not implemented");
#else
   return std::filesystem::exists(path);
#endif
}
```
And be done with it?

c10/util/FileSystem.cpp
Skylion007, line 14: Why not just use the "experimental" clang shim if it's too old?

Skylion007: std::experimental::filesystem

Skylion007: We use std::experimental for older versions of clang transparently in pybind11

ezyang:
It seems to work. It's currently living in https://www.internalfb.com/diff/D82283623 but I will probably split it out so it lands quicker.

PR #118579, opened 2024-01-29
Introduce size oblivious guards

ezyang:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #117356
* __->__ #118579

Fixes https://github.com/pytorch/pytorch/issues/117361

The implementation here slightly diverges from what was proposed in the issue, so I will recap what this PR is doing here. Today, when doing computations involving size-like unbacked SymInts, we assume for all operations that the compile time range of the integer is `[2, inf]`, even though at runtime we also accept zero and one.

This PR removes the carte blanche assumption, and instead does the analysis in a much more limited and controlled fashion: only for guards which we have designated as "size oblivious" are we willing to do the analysis under the assumption that the range of all size-like unbacked SymInts is `[2, inf]`; otherwise, we will faithfully only do analysis with `[0, inf]` (or whatever the user provided) bounds.

The infra pieces of this PR are:

* Remove runtime_var_to_range from torch/fx/experimental/symbolic_shapes.py; modify `_constrain_range_for_size` to refine the range without clamping min to 2, and instead add the symbol to a `size_like` set in the ShapeEnv
* When evaluating an expression, if the expression is requested to be evaluated in a `size_oblivious` way, we attempt to statically compute the value of the expression with the assumption that all symbols in `size_like` are updated to assume that they are `>= 2`.
* Add Python and C++ APIs for guarding on a SymBool in a size-oblivious way. In C++, I also need to add some helpers for performing symbolic comparisons, since the stock comparisons immediately specialize in the "normal" way.

The rest of the changes of the PR are marking various spots in PyTorch framework code as size oblivious, based on what our current test suite exercises.

As you review the places where we have marked things as size oblivious, it may become clear why I ended up not opting for the "designate a branch as the default branch when it's not statically obvious which way to go": for some of the conditions, this answer is rather non-obvious. I think potentially there is another refinement on top of this PR, which is something like "I don't care if you can't figure it out with ValueRange analysis, go down this path anyway if there are unbacked sizes involved." But even if we add this API, I think we are obligated to attempt the ValueRange analysis first, since it can lead to better outcomes sometimes (e.g., we are able to figure out that something is contiguous no matter what the unbacked size is.)

[...]

PR #162809, opened 2025-09-12
[FakeTensor] Make view_copy meta copy-if-needed to match eager

YufengShi-dudu:
Eager `aten.view_copy` flattens broadcasted tensors by materializing when a view is not possible (e.g., expand -> flatten), returning a non-aliasing result. The FakeTensor meta kernel delegated to `_reshape_view_helper` with `allow_copy=False`, so the same call raised:

    x = torch.arange(7)
    y = x.expand(12, 7)  # strides (0, 1)
    aten.view_copy(y, [84])  # eager: OK, fake: ValueError

This change plumbs `allow_copy=True` for `aten.view_copy` in the FakeTensor impl while keeping `view` / `_unsafe_view` unchanged. This aligns FakeTensor with eager semantics.

Fixes #162817

### DIFF
--- a/torch/_subclasses/fake_impls.py
+++ b/torch/_subclasses/fake_impls.py
@@ -728,7 +728,7 @@ def _view_unbacked_meta(
     # as "materialize": clone the input to break aliasing, then reshape.
     if allow_copy:
         strides = make_contiguous_strides_for(shape)
-        return a.clone().as_strided(shape, strides)
+        return a.clone(memory_format=torch.contiguous_format).as_strided(shape, strides)
 
     msg = f"Cannot view a tensor with shape {a.shape} and strides {a.stride()} as a tensor with shape {shape}!"
     raise ValueError(msg)
--- a/torch/_subclasses/fake_impls.py
+++ b/torch/_subclasses/fake_impls.py
@@ -656,6 +656,7 @@ def _view_unbacked_meta(
     a: torch.Tensor,
     shape: ShapeType | tuple[ShapeType],
     size_oblivious_enabled: bool = True,
+    allow_copy: bool = False,
 ) -> torch.Tensor:
     from torch._prims import view_of
     from torch.fx.experimental.symbolic_shapes import guard_or_false, sym_eq
@@ -719,7 +720,15 @@ def _view_unbacked_meta(
         torch.fx.experimental._config.backed_size_oblivious
         or _view_has_unbacked_input(a, shape)
     ):
-        return _view_unbacked_meta(a, shape, size_oblivious_enabled=False)
+        return _view_unbacked_meta(
+            a, shape, size_oblivious_enabled=False, allow_copy=allow_copy
+        )
+
+    # When allow_copy=True (i.e., view_copy), define unbacked semantics
+    # as "materialize": clone the input to break aliasing, then reshape.
+    if allow_copy:
+        strides = make_contiguous_strides_for(shape)
+        return a.clone().as_strided(shape, strides)
 
     msg = f"Cannot view a tensor with shape {a.shape} and strides {a.stride()} as a tensor with shape {shape}!"
     raise ValueError(msg)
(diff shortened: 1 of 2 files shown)


laithsakka (dismissed):

laithsakka:
can you add a unit test before landing it? should be east!

laithsakka (changes requested):
actually we probably should also add pass allow_copy to _view_unbacked_meta and address it there propagate it there also. 


basically this part
```
   if size_oblivious_enabled and (
        torch.fx.experimental._config.backed_size_oblivious
        or _view_has_unbacked_input(a, shape)
    ):
        return _view_unbacked_meta(a, shape, size_oblivious_enabled=False)
```

we can define unbacked semantics for view_copy as the following: clone the input and instead of throwing DDE, similar to a reshape.
we prob also need to add a unit test

torch/_subclasses/fake_impls.py
```diff
@@ -586,26 +586,37 @@ def _view_unbacked_meta(a, shape, size_oblivious_enabled=True):
...
-        return _view_unbacked_meta(a, shape, size_oblivious_enabled=False)
+        return _view_unbacked_meta(
+            a, shape, size_oblivious_enabled=False, allow_copy=allow_copy
+        )
+
+    # When allow_copy=True (i.e., view_copy), define unbacked semantics
+    # as "materialize": clone the input to break aliasing, then reshape.
+    if allow_copy:
+        a_clone = a.clone()
+        return torch._refs._reshape_view_helper(a_clone, *shape, allow_copy=True)
```
laithsakka: here you want probably to call 
return _view_unbacked_meta(a_clone, *shape, size_oblivious_enabled=True, allow_copy=False)

why?
1) _reshape_view_helper can still throw DDE even when a_clone is contig
2) _view_unbacked_meta guaranteed to handle contig tensors no need for  another allow_copy

YufengShi-dudu: Thanks for your suggestion, but I 'm not quite sure how we should set the `size_oblivious_enabled` here.
It seems that when we reach the `if allow_copy` branch, the `size_oblivious_enabled` should already be `False` here. Do we need to reset it to `True` here or set it as `size_oblivious_enabled=size_oblivious_enabled`?

Could you also explain a bit more about the `size_oblivious_enabled` for knowledge sharing? I think it is consumed by `_compute_stride` but not sure how it works behind.

YufengShi-dudu: Find this one, might be helpful: https://github.com/pytorch/pytorch/pull/118579

laithsakka: ok, let me explain https://github.com/pytorch/pytorch/pull/118579 is not relevant here.
We actually moved away from guard_size_oblivious to guard_or_true/guard_or_false and other explicit unbacked semantics.

What is unbacked semantics? It refers to how the program should behave when we do not know some property that we are checking due to unbacked symbols (unbacked symbols are dynamic shapes that we are not allowed or can't guard on).
Ex: if (u==0) // shall I take this branch or not?
This would fail with Data-dependent error because we are not allowed to guard on u0 and actually do not know the answer anyway.But if we use if guard_or_false(u==0) then this would just return false instead of DDE.

Ok, back to this function.
When we have a view op, we run a simpler version (with less check but possible different output than eager) with size_oblivious_enabled being True first. The effect of size_oblivious_enabled is to turn on guard_or_true, guard_or_false, that called within this function.

Ex: if is_contiguous_or_false(a) if size_oblivious_enabled else is_contiguous(a).
Now if anyway we fail to determine the view op validity even with size_oblivious_enabled=True, we run it with size_oblivious_enabled=False to throw a DDE.

Now with allow_clone=True,
we probably should just return
```
strides = make_contiguous_strides_for(shape)
return a.clone().as_strided(shape, strides)
```
No? We do not even need to do recursive call we know clone will give contiguous output.

laithsakka:
If you just want to fix eager i can follow up with unbacked stuff my self no worries. lmk.

torch/_subclasses/fake_impls.py
YufengShi-dudu: Thanks for your detailed explanations! Now I have a much clearer understanding on how stride and view op work in pytorch

YufengShi-dudu: But I still have a concern of using `make_contiguous_strides_for` here. I think the clone will try to preserve the original memory format for both contiguous and non-contiguous cases. 
Is it ok for us to always compute strides in a contiguous way? Or this kind of "reshape" is expected?

laithsakka: you are right we needs to do  clone(memory_format=torch.contiguous_format)

YufengShi-dudu: Fixed in a follow-up commit.

laithsakka (dismissed):

YufengShi-dudu:
Hi @laithsakka, shall we get this PR merged? Thanks

YufengShi-dudu:
[@YufengShi-dudu can you rebase?
](https://github.com/pytorch/pytorch/pull/162809#issuecomment-3751814835)
Done

laithsakka (approved):

PR #162815, opened 2025-09-12
Support setting grad_dtype on leaf tensors

soulitzer:
* #164407
* __->__ #162815

`grad_dtype` is a new attribute on Tensor to control gradient dtype:
- Access/setting is leaf-only.
- grad_dtype is respected when (1) when assigning to .grad, and (2) in the engine after the previous node produces incoming gradients for AccumulateGrad. (See table below for details)
- Not setting grad_dtype preserves the current behavior. Accessing it returns `t.dtype`
- `grad_dtype` cannot be set when there is already a `.grad` present and the dtypes conflict.

| `grad_dtype` setting | Setting `.grad` manually | Incoming gradient from autograd engine |
|-----------------------|--------------------------|-----------------------------------------|
| **Default (tensor’s dtype)** | `.grad` must match tensor’s dtype | Engine casts incoming grad to tensor’s dtype |
| **Set to specific dtype** | `.grad` must match that dtype | Engine casts incoming grad to the specified dtype |
| **Set to `None`** | `.grad` may be any dtype | Engine does not cast; accepts incoming grad dtype as-is |

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela @xmfan


soulitzer:
Discussed with @albanD:
-  We will go with `grad_dtype_=None` means `.grad `is allowed to be any dtype.
- Although dtype on tensor is mutable, it is only possible through the means of .data

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1481,6 +1491,105 @@ static PyObject* THPVariable_get_nbytes(THPVariable* self, void* unused) {
...
+    allow_mismatch = true;
+  } else {
+    auto* dtype = reinterpret_cast<THPDtype*>(obj);
+    new_dtype = dtype->scalar_type;
+    allow_mismatch = false;
+  }
+  // There are two grad_acc cases to handle:
+  // 1) If there's already a grad_acc, update its InputMetadata now.
+  if (auto grad_acc = torch::autograd::impl::try_get_grad_accumulator(var)) {
+    grad_acc->mutable_input_metadata(0).set_grad_dtype(new_dtype);
```
soulitzer: How concerned should we be about thread safety?

How reads and writes happen:
- Write:
  - Python attribute setting, guarded by GIL
- Read:
  - The engine, not guarded by lock (grad_fn is only looked when running ::apply and ::release_saved)
  - In python via attribute access, guarded by GIL
  
There can be a read-write race if one thread writes via python attribute setting, and the engine reads this information from another thread.

Possible fixes:
- Make this field atomic
- Lock the read and write locations
- more?

torch/csrc/autograd/engine.cpp
```diff
@@ -949,15 +949,23 @@ static void validate_outputs_impl(
...
-      grad = grad.to(c10::typeMetaToScalarType(metadata.options().dtype()));
-    }
-    if (grad.dtype() != metadata.dtype()) {
-      std::stringstream ss;
-      ss << "invalid gradient at index " << i << " - expected dtype ";
-      ss << metadata.dtype() << " but got " << grad.dtype();
-      TORCH_CHECK(false, format_error(ss.str()));
+
+    // grad_dtype and allow_grad_dtype_mismatch can only be customized
+    // on leaf tensors, defaulting to nullopt and false otherwise.
```
albanD: Not sure what this comment means.

torch/csrc/autograd/input_metadata.h
```diff
@@ -109,5 +125,10 @@ struct TORCH_API InputMetadata {
   bool is_tensor_subclass_ = false;
   bool is_nested_ = false;
   bool was_default_constructed_ = true;
+  // When nullopt, defer to the tensor's dtype; otherwise, use this dtype.
+  std::optional<at::ScalarType> grad_dtype_;
```
albanD, line 130: Is it going to add crazy padding to put this in between all the boolean fields?

torch/csrc/autograd/python_variable.cpp
albanD: The GIL doesn't actually help as call into the python api might release it.

torch/csrc/autograd/variable.h
```diff
@@ -293,6 +300,22 @@ struct TORCH_API AutogradMeta : public c10::AutogradMetaInterface {
       uint64_t level,
       bool is_inplace_op) override;
 
+  const std::optional<at::ScalarType>& grad_dtype() const {
```
albanD: How does this all work at the C++ level?
Wouldn't the logic be significantly simpler if you had a single C++ field analogous to your python Tensor.grad_dtype and only bind this one?

soulitzer: Having a single C++ field was what I had in an earlier version of this PR, but not sure it is possible due to the possibility of Tensor's dtype being mutated.

Suppose I have only a grad_dtype_: Optional<ScalarType> field. If nullopt on the field means that I am allowed to have any dtype, then its ambiguous whether non-nullopt means that the dtype should match upon mutation or stay the same because e.g. if the user requested a concrete dtype. It's also annoying to make sure the dtypes are consistent. I also have to make sure to handle the cases where the classes are default constructed.

torch/csrc/autograd/engine.cpp
soulitzer: took me a second to parse that as well haha, what I mean is:
"defaulting to nullopt and false, respectively, otherwise"

torch/csrc/autograd/python_variable.cpp
soulitzer: From offline discussion: The correctness of mutating such attributes are the responsibility of the user.

torch/csrc/autograd/engine.cpp
```diff
@@ -949,15 +949,17 @@ static void validate_outputs_impl(
...
-      std::stringstream ss;
-      ss << "invalid gradient at index " << i << " - expected dtype ";
-      ss << metadata.dtype() << " but got " << grad.dtype();
-      TORCH_CHECK(false, format_error(ss.str()));
+
+    if (metadata.grad_dtype().has_value()) {
+      if (grad.scalar_type() != metadata.grad_dtype().value()) {
+        grad = grad.to(metadata.grad_dtype().value());
+      }
+      if (grad.dtype() != metadata.grad_dtype().value()) {
```
albanD: The updated conditionals look wrong. One is checking dtype and the other scalar type against the same thing

torch/csrc/autograd/input_metadata.cpp
```diff
@@ -44,7 +46,8 @@ InputMetadata::InputMetadata(const at::Tensor& t)
           t.options(),
           compute_variant_shape(t),
           is_python_dispatch(t),
-          t.is_nested()) {}
+          t.is_nested(),
+          t.grad_dtype()) {}
```
albanD, line 50: Please document that the value of this is frozen during the forward.

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1481,6 +1489,63 @@ static PyObject* THPVariable_get_nbytes(THPVariable* self, void* unused) {
...
+static int THPVariable_set_grad_dtype(
+    THPVariable* self,
+    PyObject* obj,
+    void* unused) {
+  HANDLE_TH_ERRORS
+  if (check_has_torch_function((PyObject*)self)) {
+    return handle_torch_function_setter(self, "grad_dtype", obj);
+  }
+  const auto& var = THPVariable_Unpack(self);
+  TORCH_CHECK(!var.grad_fn(), "grad_dtype can only be set on leaf tensors.");
```
albanD: Shouldn't this check happen lower as this is true as well for the C++ API.

torch/csrc/autograd/input_metadata.cpp
albanD: Ho actually you do update it later? Hopefully that is tested!
Either way, we should document it!

torch/csrc/autograd/variable.h
```diff
@@ -293,6 +300,22 @@ struct TORCH_API AutogradMeta : public c10::AutogradMetaInterface {
...
+    grad_dtype_ = grad_dtype;
+  }
+
+  bool allow_grad_dtype_mismatch() const {
+    return allow_grad_dtype_mismatch_;
+  }
+
+  void set_allow_grad_dtype_mismatch(bool allow_mismatch) {
+    allow_grad_dtype_mismatch_ = allow_mismatch;
+  }
```
albanD: I feel like you can push the API unification all the way here and move the logic in Tensor.cpp to here.
That way we only have 2 APIs on the VariableHooks instead of 4 !

torch/csrc/autograd/engine.cpp
soulitzer: Good catch

torch/csrc/autograd/input_metadata.cpp
soulitzer: Beefed up the documentation on the field!

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1481,6 +1489,62 @@ static PyObject* THPVariable_get_nbytes(THPVariable* self, void* unused) {
...
+        var.grad().dtype(),
+        "'. Please clear the gradient (.grad = None) before changing grad_dtype, "
+        "or ensure the new grad_dtype matches the existing gradient's dtype.");
+  }
+  std::optional<at::ScalarType> new_dtype;
+  if (obj != Py_None) {
+    auto* dtype = reinterpret_cast<THPDtype*>(obj);
+    new_dtype = dtype->scalar_type;
+  }
+  // 1) If there's already a grad_acc, update its InputMetadata now.
```
albanD: This logic should be on the TensorBase version otherwise C++ calls to set_grad_dtype will have a different behavior

albanD (approved):
The existing node change should be moved. But once that is done sounds good to me.

PR #162818, opened 2025-09-12
compile_kernel: POC cpp outer template support

msaroufim:
Users would write templated code

```cpp
template<typename T>
__global__ void add_kernel(T* a, T* b, T* c, int n) {
    int i = blockIdx.x * blockDim.x + threadIdx.x;
    if (i < n) c[i] = a[i] + b[i];
}
```

And we explicitly instantiate it

```cpp
template __global__ void add_kernel<float>(float* a, float* b, float* c, int n);

extern "C" {
__global__ void add_kernel_wrapper(float* a, float* b, float* c, int n) {
    add_kernel<float>(a, b, c, n);
}
}
```

NVRTC is happy 

It's not totally free since we expect users to change how they call compile_kernel

```python
kernel = torch.cuda._compile_kernel(
    template_code, "add_kernel",
    is_template=True,
    template_types=["float"], 
    wrapper_signature="float* a, float* b, float* c, int n",
    wrapper_body="add_kernel<float>(a, b, c, n);"
)
```

Inner templates that would show up in header libraries like CUB already works here https://github.com/pytorch/pytorch/pull/156380 so feels like both PRs are useful altho still not 100% sure


msaroufim:
Closed in favor of https://github.com/pytorch/pytorch/pull/162875

PR #162823, opened 2025-09-12
Fix pipeline parallelism not correctly initializing backwards stages when evaluating before training.

### LINKED ISSUE #162822: Pipeline parallelism does not correctly initialize backwards stages when evaluating before training.
### 🐛 Describe the bug

When running evaluation with [pipeline parallelism schedules](https://github.com/pytorch/pytorch/blob/release/2.9/torch/distributed/pipelining/schedules.py) before training, the following happens:

- Calling [eval()](https://github.com/pytorch/pytorch/blob/152383b74540785da418611a8ae8a7c5028a0cdc/torch/distributed/pipelining/schedules.py#L385) on any of the schedules, it will temporary set _has_backwards to False, before performing a step().
- The step() will then initialize the [stage](https://github.com/pytorch/pytorch/blob/152383b74540785da418611a8ae8a7c5028a0cdc/torch/distributed/pipelining/schedules.py#L573C9-L573C26) or [stages](https://github.com/pytorch/pytorch/blob/152383b74540785da418611a8ae8a7c5028a0cdc/torch/distributed/pipelining/schedules.py#L1454).
- Since _has_backwards is False, stage._prepare_backward_infra() will not be called.
- When running the first training step, initialization will not be run again (e.g. [here](https://github.com/pytorch/pytorch/blob/152383b74540785da418611a8ae8a7c5028a0cdc/torch/distributed/pipelining/schedules.py#L1578)).
- Thus, an error occurs when the backwards step is reached.

### Versions

PyTorch 2.9 Nightly

BlueCrescent:
Previously, an eval() call before a training step() would not correctly initialize the backward pass of the pipeline stages, leading to errors during the subsequent training step. This PR ensures that the backward stages can still be initialized after an eval() call.

Fixes #162822

### DIFF
--- a/torch/distributed/pipelining/schedules.py
+++ b/torch/distributed/pipelining/schedules.py
@@ -558,7 +558,8 @@ class PipelineScheduleSingle(_PipelineSchedule):
         # Self attributes
         self._stage = stage
         self._num_stages = stage.num_stages
-        self._stage_initialized = False
+        self._forward_stage_initialized = False
+        self._backward_stage_initialized = False
 
         if n_microbatches < self._num_stages:
             raise ValueError(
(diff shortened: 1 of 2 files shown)


dcci (changes requested):
Thanks for your contribution. can you add a test?

dcci (approved):
LGTM, cc: @ezyang in case he might want to take a look

wconstab:
@H-Huang can you take a look?

torch/distributed/pipelining/schedules.py
```diff
@@ -558,7 +558,8 @@ def __init__(
         # Self attributes
         self._stage = stage
         self._num_stages = stage.num_stages
-        self._stage_initialized = False
+        self._forward_stage_initialized = False
```
H-Huang: Would prefer the naming to be `_stage_forward_initialized` and `_stage_backward_initialized`.

H-Huang (commented):
Thanks for adding, overall LGTM, just one comment.

torch/distributed/pipelining/schedules.py
BlueCrescent: Renamed it (also in the multi stage schedules).

H-Huang (approved):
Thanks for fixing!

BlueCrescent:
I rebased to main. Sadly, that seems to have requested a review from everyone.

PR #162825, opened 2025-09-12
[nativert] fix rocm usage for cmake build

dolpm:

### DIFF
--- a/test/cpp/nativert/test_aoti_model_container_registration.cpp
+++ b/test/cpp/nativert/test_aoti_model_container_registration.cpp
@@ -12,5 +12,5 @@ TEST(AOTIModelContainerRegistrationTests, TestRegister) {
   EXPECT_TRUE(AOTIModelContainerRunnerRegistry()->Has(at::kCUDA));
 #else
   EXPECT_FALSE(AOTIModelContainerRunnerRegistry()->Has(at::kCUDA));
-#endif // USE_CUDA
+#endif // defined(USE_CUDA) || defined(USE_ROCM)
 }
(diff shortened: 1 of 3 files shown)


test/cpp/nativert/test_aoti_model_container_registration.cpp
```diff
@@ -8,9 +8,16 @@ using namespace torch::nativert;
 TEST(AOTIModelContainerRegistrationTests, TestRegister) {
   EXPECT_TRUE(AOTIModelContainerRunnerRegistry()->Has(at::kCPU));
 
+#ifdef USE_ROCM
+  EXPECT_TRUE(AOTIModelContainerRunnerRegistry()->Has(at::kHIP));
```
jeffdaily: This would fail for ROCm builds since kHIP isn't used for ROCm.  Instead we reuse aka masquerade as kCUDA.

dolpm: IC - for the triton test in here it's a little different since we set the device manually based on the binary suffix we are trying to load (in that case, .hsaco --> check registry for at::kHIP) so this will still hold. i've updated the aoti test to always assume registered at kCUDA. i'd still like to land this for the TORCH_CHECK as well.

test/cpp/nativert/test_triton_kernel_manager_registration.cpp
```diff
@@ -8,18 +8,16 @@ using namespace torch::nativert;
 TEST(TritonKernelManagerRegistrationTests, TestRegister) {
   EXPECT_TRUE(TritonKernelManagerRegistry()->Has(at::kCPU));
 
-#ifdef USE_CUDA
```
dolpm, line 11: for internal builds we have both USE_CUDA and USE_ROCM set, but for cmake it is either/or -- so this needs to be updated to check for rocm before cuda, as opposed to the other way around

jeffdaily: I don't understand how you can set both.  There are quite a few places in torch sources that assume it is either/or.

dolpm: for our internal builds it's set this way (where it was originally tested), but this is incorrect for cmake hence the change here.

PR #162827, opened 2025-09-12
[ROCm] enable aoti tests, forward fix 162353

jeffdaily:
Forward fix for tests added by #162353.  Enables aoti tests on rocm.

cc @sunway513 @jithunnair-amd @pruthvistony @ROCmSupport @dllehr-amd @jataylo @hongxiayang @naromero77amd

### DIFF
--- a/test/cpp/nativert/test_aoti_model_container_registration.cpp
+++ b/test/cpp/nativert/test_aoti_model_container_registration.cpp
@@ -8,7 +8,7 @@ using namespace torch::nativert;
 TEST(AOTIModelContainerRegistrationTests, TestRegister) {
   EXPECT_TRUE(AOTIModelContainerRunnerRegistry()->Has(at::kCPU));
 
-#ifdef USE_CUDA
+#if defined(USE_CUDA) || defined(USE_ROCM)
   EXPECT_TRUE(AOTIModelContainerRunnerRegistry()->Has(at::kCUDA));
 #else
   EXPECT_FALSE(AOTIModelContainerRunnerRegistry()->Has(at::kCUDA));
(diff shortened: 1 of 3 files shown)


test/cpp/nativert/test_aoti_model_container_registration.cpp
```diff
@@ -8,7 +8,7 @@ using namespace torch::nativert;
 TEST(AOTIModelContainerRegistrationTests, TestRegister) {
   EXPECT_TRUE(AOTIModelContainerRunnerRegistry()->Has(at::kCPU));
 
-#ifdef USE_CUDA
+#if defined(USE_CUDA) || defined(USE_ROCM)
```
dolpm, line 11: i'm actually very surprised that this works, since iiuc, the device type should be at::kHIP at https://github.com/pytorch/pytorch/blob/b5f4a7dc141471f6b6c3c012b4a6f468f91834f4/torch/nativert/executor/AOTInductorDelegateExecutor.cpp#L115 and this is registered at kCUDA

dolpm: i think this, since i've uncovered some other gaps for the oss build, is to merge the cmake changes here with https://github.com/pytorch/pytorch/pull/162825

jeffdaily: ROCm builds masquerade as the kCUDA device.  the kHIP device is a thin shell that does nothing.

jeffdaily: All we need is this current PR, truly.  Trust me.  :-)

dolpm (approved):

huydhn (approved):
Ty for the fix!

PR #162828, opened 2025-09-12
[DCP] Add timeout for checkpoint background process join

kevinmtang:
Summary:
Cleaning up checkpoint background process can currently block trainer thread indefinitely if the process is hanging (notably due to Gloo pg init timeout).

This diff adds a 5s grace period for normal termination and sends SIGTERM if unable to shut down in that period.

Rollback Plan:

Differential Revision: D82268979

### DIFF
--- a/torch/distributed/checkpoint/_async_process_executor.py
+++ b/torch/distributed/checkpoint/_async_process_executor.py
@@ -113,9 +113,16 @@ class _AsyncCheckpointProcess:
 
     def __del__(self) -> None:
         if self._save_process.is_alive():
-            logger.info("Terminating the checkpoint background process...")
-            self._send(_CheckpointSaveProcessControlOpts.TERMINATE)
-            self._save_process.join()
+            try:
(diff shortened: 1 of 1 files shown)


meetv18 (approved):

torch/distributed/checkpoint/_async_process_executor.py
```diff
@@ -113,9 +113,16 @@ def __init__(
 
     def __del__(self) -> None:
         if self._save_process.is_alive():
-            logger.info("Terminating the checkpoint background process...")
-            self._send(_CheckpointSaveProcessControlOpts.TERMINATE)
-            self._save_process.join()
+            try:
+                logger.info("Terminating the checkpoint background process.")
+                self._send(_CheckpointSaveProcessControlOpts.TERMINATE)
+                self._save_process.join(timeout=5)
```
Skylion007, line 119: Should this be configurable via an ENV var?

PR #162834, opened 2025-09-12
[BE] Add Documentation for Device APIs

jiannanWang:
Added documentation for torch.cuda APIs. 
Fixed docstring for xpu and mtia is_bf16_supported API.

### DIFF
--- a/torch/cuda/__init__.py
+++ b/torch/cuda/__init__.py
@@ -500,7 +500,7 @@ class CudaError(RuntimeError):
 
 
 def check_error(res: int) -> None:
-    r"""Raise an error if the result of a CUDA runtime API call is not sucess."""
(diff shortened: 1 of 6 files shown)


torch/cuda/__init__.py
```diff
@@ -500,6 +500,7 @@ def __init__(self, code: int) -> None:
 
 
 def check_error(res: int) -> None:
+    r"""Raise an error if the result of a CUDA runtime API call is not sucess."""
```
janeyx99:
```suggestion
    r"""Raise an error if the result of a CUDA runtime API call is not success."""
```

janeyx99 (commented):
Nice, I'm waiting for the docs build to check that everything looks ok before approving.

janeyx99 (approved):

PR #162840, opened 2025-09-12
[BE] Remove named tensor

PaliC:
cc @EikanWang @jgong5 @wenzhe-nrv @sanchitintel @mingfeima @XiaobingSuper @ashokei @jingxu10 @jerryzh168 @aditew01


vadimkantorov:
Ha-ha, maybe instead of named dims, it would be better to add some support for named tensors - for debugging / vis / profiling :)
- https://github.com/pytorch/pytorch/issues/104247

(at least, tensor.name attribute without auto-propagation support)

PR #162844, opened 2025-09-12
[lint][CI] Don't checkout submodules for lintrunner-noclang

clee2000:
Shouldn't be needed?


huydhn (approved):
LGTM! It makes sense that this is not needed for noclang lint jobs

PR #162850, opened 2025-09-12
Clean up 'torch.onnx' entries from public API allowlist

justinchuby:
Clean up entries related to 'torch.onnx' from the allowlist as the apis in onnx are properly configured.


albanD (approved):
Thanks!

PR #162310, opened 2025-09-05
[BE] Update Python min version to 3.10

malfet:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #162310

PR #162862, opened 2025-09-12
[CI] Move Windows build/tests to Python-3.10

malfet:
* #162310
* __->__ #162862
* #163341
* #163339

What supposed to be a very simple change end up being quite involved, as current Windows CI framework is quite inflexible, i.e. it takes a lots of argument, but later on ignores them, namely:
 - `PYTHON_VERSION` used to be a no-op that is simply ignored by the scripts
 - With this change, `setup-win` action will create an environment called `py_tmp` with specific python version + intel-openmp (that is hard runtime requirement, but for some reason not packaged into the wheel nor marked as such)
 - Copied test type dependencies from https://github.com/pytorch/test-infra/blob/be01a40157c36cd5a48391fdf44a7bc3ebd4c7e3/aws/ami/windows/scripts/Installers/Install-Pip-Dependencies.ps1#L16 into `win-test.sh`, but made some adjustments to be compatible with 3.10 runtime (scipy version update) and just make rerun-tests compatible with the rest of the deps

I think in the long run, one needs to update https://github.com/pytorch/test-infra/blob/4432e2cacd8a5bb7a46158e71d08c937c502e35a/aws/ami/windows/scripts/Installers/Install-Miniconda3.ps1 that currently pins Miniconda python to 3.9, but also figure out how CI can still create a new environment without having to download all the dependencies all the time


wdvr (approved):

huydhn (approved):
OK as a workaround I guess, but maybe we need to remember to revert this one the AMI is updated

malfet:
> OK as a workaround I guess, but maybe we need to remember to revert this one the AMI is updated

I'll wait for CI to be green, and then push a change with the TODO comment...

malfet:
CI is green, but all those settings are lie, not sure why it happening, but Python version are still 3.9...

PR #162863, opened 2025-09-12
[Inductor-FX] Support IndexPutFallback

blaine-rister:
# Feature

This PR supports lowering `IndexPutFallback` through Inductor's FX converter. The approach is very similar to the one taken in https://github.com/pytorch/pytorch/pull/162686.

Compared to `ScatterFallback`, this required one additional change: the value of `self.op_overload` for `IndexPutFallback` was inaccurate. Previously, it used `aten.index_put`, which would result in unsound FX IR. The existing Python/C++ codegen use `aten.index_put_`, since the fallback mutates its input. This PR changes `self.op_overload` to match that.

# Test plan
Added a CI test lowering deterministic index put via the FX converter.


torch/_inductor/codegen/wrapper_fxir.py
```diff
@@ -654,6 +655,42 @@ def _generate_multi_output(self, line: WrapperLine) -> None:
         node.name = line.result_name
         self.buffer_to_node[line.result_name] = node
 
+    def _generate_fallback_call(
```
blaine-rister, line 658: A small refactor to reduce code duplication between `ScatterFallback` and `IndexPutFallback`.

blaine-rister:
Some CI tests started failing after merging with main. Investigating...

torch/_inductor/lowering.py
```diff
@@ -3717,8 +3717,8 @@ def index_put_as_masked_fill(self, indices, value, accumulate):
 
 
 def index_put_fallback(self, indices, values, accumulate):
-    assert isinstance(V.graph.current_node.target, torch._ops.OpOverload)
-    ir.IndexPutFallback(V.graph.current_node.target, self, indices, values, accumulate)
+    op_overload = getattr(aten.index_put_, V.graph.current_node.target._overloadname)  # type: ignore[union-attr]
```
angelayi, line 3720: what is V.graph.current_node.target?

blaine-rister: Good question. I think this is the target of the FX node which we're currently lowering, so in this case it could be `aten.index_put` or `aten.index_put_`. However, the semantics of `IndexPutFallback` follow the latter op, which mutates its input. Hence if `V.graph.current_node.target == aten.index_put`, we need to change the target to `aten.index_put_`.

I copied this trick from the existing scatter fallback: https://github.com/pytorch/pytorch/blob/main/torch/_inductor/lowering.py#L3967

angelayi: thanks for the explanation!

angelayi (approved):

PR #162867, opened 2025-09-13
Do not use // but use CleanDiv or FloorDiv instead

laithsakka:
* __->__ #162867

When operating rewriting sympy expressions in the compiler code base we want to generate
FloorDiv(a, b) CleanDiv(a, b) directly and not a//b. since the later become floor(a*pow(b, -1))

For symnodes we automatically handle that conversions in the symnode op dispatch. 

I will follow up with an issue to track all other usages of //.

Block internal Model. 


laithsakka:
switch to internal first PR https://github.com/pytorch/pytorch/pull/162869

PR #162873, opened 2025-09-13
[Graph Partition] allow sharing default device context

BoyuanFeng:
Entering a device context takes 30 us and exiting a device context takes 11 us. If all graph partitions and cudagraph-unsafe ops happen on the same device, we can share the device context.

## Trace

Use vLLM as an example. The first trace shows dynamo graph partition.
image

The second trace shows inductor graph partition prior to this PR. 
image

Comparing with fx graph partition, we can see inductor graph partition shows extra overhead from enter/exit device contexts (13+6 us -> 30+11 us), but smaller runtime overhead (13 us -> 7 us). This motivates the PR to share default device context. 

The third trace shows Inductor graph partition after this PR. We observe that the extra overhead from enter/exit device contexts have been fixed. At the same time, we observe the smaller runtime overhead.
image

cc @mcarilli @ezyang @eellison @penguinwu @voznesenskym @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @mlazos


torch/_inductor/codegen/wrapper.py
```diff
@@ -1333,7 +1334,7 @@ def codegen_input_size_and_nan_asserts(self) -> None:
     # that stream caching happens per graph instance. this
     # is important for nested subgraph codegening.
     def write_get_raw_stream(self, device_idx: int, graph_name: str) -> str:
-        self.write_get_raw_stream_header_once()
+        self.write_get_raw_stream_header()
```
BoyuanFeng, line 1337: standalone_compile may need additional imports of `get_raw_stream`

shunting314: Why additional imports are needed? Can those be added as cache key if repeated imports are needed

shunting314 (commented):
Nice finding!

Is it possible to improve device context manager instead? i.e., if we are already on the mentioned device, make it cheaper to enter/exit

Also any number for the overall improvement for inference?

torch/_inductor/codegen/wrapper.py
BoyuanFeng: We need imports of `get_raw_stream` for both subgraphs and the main graphs, including the final output code and autotune_at_compile_time blocks. 

Currently, `write_get_raw_stream_header_once` relies on `cache_on_self` to import once. When generating subgraph, it calls `SubgraphPythonWrapperCodegen.write_get_raw_stream_header_once()` (cache_on_self) which calls `PythonWrapperCodegen.write_get_raw_stream_header_once()` (cache_on_self).

https://github.com/pytorch/pytorch/blob/8590c3a66b4adb77b59645dd12bf44695ac3b994/torch/_inductor/codegen/wrapper.py#L3652-L3660

https://github.com/pytorch/pytorch/blob/8590c3a66b4adb77b59645dd12bf44695ac3b994/torch/_inductor/codegen/wrapper.py#L1191-L1193

As a result, the main graph will skip `PythonWrapperCodegen.write_get_raw_stream_header_once()` (cache_on_self) since it has just been called. However, this leads to errors since the main graph code also needs this import.

So I call `write_get_raw_stream_header` here whenever we need it, and only write to `self.imports` when `self.imports` does not contain `import_get_raw_stream_str`.

BoyuanFeng:
> Also any number for the overall improvement for inference?

For vllm Qwen/Qwen3-0.6B, this PR reduces latency from 0.34 seconds to 0.328 seconds, around 3.5% speedup.

Command to repro:
`vllm bench latency -O.splitting_ops=[] -O.cudagraph_mode=PIECEWISE -O.use_inductor_graph_partition=true`

We can easily see the speedup from trace. `torch/_inductor/output_code.py __call__` reduces from 5.5 ms to 5 ms.

[trace before the pr](https://interncache-all.fbcdn.net/manifold/perfetto-artifacts/tree/ui/index.html#!/?url=https://interncache-all.fbcdn.net/manifold/perfetto_internal_traces%2Ftree%2Fshared_trace%2Fboyuan_fe258c1b-a085-4671-8b2d-4e93fbc40d55_qwen3-0.6b-before-162873) vs [trace after the pr](https://interncache-all.fbcdn.net/manifold/perfetto-artifacts/tree/ui/index.html#!/?url=https://interncache-all.fbcdn.net/manifold/perfetto_internal_traces%2Ftree%2Fshared_trace%2Fboyuan_8b409d44-6971-4605-8bbd-f5a0cf9a1dc8_qwen3-0.6b-after-162873)

torch/_inductor/codegen/wrapper.py
shunting314: > Currently, write_get_raw_stream_header_once relies on cache_on_self to import once. When generating subgraph, it calls SubgraphPythonWrapperCodegen.write_get_raw_stream_header_once() (cache_on_self) which calls PythonWrapperCodegen.write_get_raw_stream_header_once() (cache_on_self).

Is the import for the subgraph wrapper not added to the global scope?

shunting314:
> For vllm Qwen/Qwen3-0.6B, this PR reduces latency from 0.34 seconds to 0.328 seconds, around 3.5% speedup.

Want to dive a bit more on the benchmarking. How many subgraphs do we generate for Qwen/Qwen3-0.6B?  Let's say we save X device.__enter__/__exit__ each time we call the wrapper. Assuming a pair of device.__enter__/__exit__ cost Y us (~40us according to your measurement). Then the total saving for a wrapper call is (X * Y) ms.  (X * Y) / 3.5% should roughly match the latency of one inference  call.

Also what blocks us from going one step further to call device.__enter__/__exit__ only once and do multiple wrapper calls.

torch/_inductor/codegen/wrapper.py
BoyuanFeng: No. A new wrapper code is generated for each subgraph.
https://github.com/pytorch/pytorch/blob/8590c3a66b4adb77b59645dd12bf44695ac3b994/torch/_inductor/scheduler.py#L5155-L5173

BoyuanFeng:
> what blocks us from going one step further to call device.enter/exit only once and do multiple wrapper calls.

Yes we currently only call device.enter/exit once with multiple wrapper calls. [tlparse](https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/.tmpjvLR9c/rank_0/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000) and [output code](https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/.tmpjvLR9c/rank_0/-_0_0_0/inductor_output_code_cyj6ffto2az4zbaqd4u2o5vk4w4rkfb57wmi4q4bagtexpgngbs7_5.txt?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000). There are  29 partitions and 28 `torch.ops.vllm.unified_attention_with_output.default`s but only 1 `with torch.cuda._DeviceGuard(0):` in `def call()`.

BoyuanFeng:
For Qwen3-0.6B.

Before this PR:
image

After this PR:
image

Overall, `output_code call` reduces from 5.5 ms to 5 ms. Since there are 29 subgraphs and 28 attention ops, we expect each subgraph/attention ops save 500/28=18 us. From the trace, we see around (10+27)-(4+9) = 24 us saving. Note that the saving changes a bit across attention&partition calls.

This 10% saving is larger than the end-to-end 3.5% saving. Because the model also have several other components such as update_states, prepare_inputs, etc.

torch/_inductor/codegen/wrapper.py
shunting314: > No. A new wrapper code is generated for each subgraph.

But all wrapper code still dump to the same .py file right?

shunting314:
ok. didn't realize that there are so many partitions due to not capture attention kernels in cudagraphs

shunting314 (approved):

torch/_inductor/codegen/wrapper.py
BoyuanFeng: yes for the output_code. But no for `config.triton.autotune_at_compile_time`, which creates a separate code to run IIUC.

PR #162874, opened 2025-09-13
[MPS] Do not explicit wait for sync

oraluben:
Fixes #162872

The deleted line was waiting for this notify:

https://github.com/pytorch/pytorch/blob/aa41d3e49cbfef8117693ab80ec1ad57accfcb41/aten/src/ATen/mps/MPSEvent.mm#L28

But when we explicitly sync stream as in #162872, this will trigger a dead lock as no other threads will notify.

I think users are responsible for syncing, not here. Or we can add a full sync here, not just a wait.

### DIFF
--- a/.ci/docker/requirements-ci.txt
+++ b/.ci/docker/requirements-ci.txt
@@ -382,7 +382,7 @@ cmake==4.0.0
 tlparse==0.4.0
 #Description: required for log parsing
 
-cuda-bindings>=12.0,<13.0 ; platform_machine != "s390x"
+cuda-bindings>=12.0,<13.0 ; platform_machine != "s390x" and sys_platform != "darwin"
 #Description: required for testing CUDAGraph::raw_cuda_graph(). See https://nvidia.github.io/cuda-python/cuda-bindings/latest/support.html for how this version was chosen. Note "Any fix in the latest bindings would be backported to the prior major version" means that only the newest version of cuda-bindings will get fixes. Depending on the latest version of 12.x is okay because all 12.y versions will be supported via "CUDA minor version compatibility". Pytorch builds against 13.z versions of cuda toolkit work with 12.x versions of cuda-bindings as well because newer drivers work with old toolkits.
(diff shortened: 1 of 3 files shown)


malfet:
Can you add a regression test?

oraluben:
> Can you add a regression test?

done. without this pr, the new test will get stuck. I didn't find a way to let it gracefully fail.

aten/src/ATen/mps/MPSEvent.mm
```diff
@@ -226,8 +226,7 @@
   std::lock_guard<std::recursive_mutex> lock(m_mutex);
   MPSEvent* start_event = getInUseEvent(start_event_id, false);
   MPSEvent* end_event = getInUseEvent(end_event_id, false);
-  // the notify is called on a separate thread, so this waits for that
-  end_event->waitForCpuSync();
```
can-gaa-hou, line 230: Feels like better to have a time-out check rather than just deleting it.

.ci/docker/requirements-ci.txt
```diff
@@ -382,7 +382,7 @@ cmake==4.0.0
 tlparse==0.4.0
 #Description: required for log parsing
 
-cuda-bindings>=12.0,<13.0 ; platform_machine != "s390x"
+cuda-bindings>=12.0,<13.0 ; platform_machine != "s390x" and sys_platform != "darwin"
```
can-gaa-hou, line 385: Is it better to be fixed in another PR?

aten/src/ATen/mps/MPSEvent.mm
oraluben: I don't think that's how it works.

.ci/docker/requirements-ci.txt
oraluben: Running mps tests need to install test dependencies, which will be blocked by the previous version of this file. This is related. Would you stop commenting on changes you're not familiar with?

PR #162875, opened 2025-09-13
[NVRTC] Enable compiling templated kernels

gau-nernst:
Per NVRTC doc - https://docs.nvidia.com/cuda/nvrtc/index.html#accessing-lowered-names, we can compile a templated kernel (e.g. `kernel<float>`) with the following steps

NVRTC side
- (new) `nvrtcAddNameExpression` -> C++ template e.g. `f<float>`
- `nvrtcCompileProgram`
- (new) `nvrtcGetLoweredName` -> get mangled name. need to do a copy since later this string is freed after NVRTC program is destroyed
- `nvrtcDestroyProgram`

CUDA side
- use mangled name instead of normal name -> profit
- `extern "C"` is not even needed

cc @msaroufim

### DIFF
--- a/test/test_cuda.py
+++ b/test/test_cuda.py
@@ -7160,6 +7160,67 @@ class TestCompileKernel(TestCase):
         expected = torch.full((n,), test_value, device="cuda", dtype=torch.float16)
         torch.testing.assert_close(output, expected, rtol=1e-3, atol=1e-3)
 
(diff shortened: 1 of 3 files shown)


msaroufim (approved):

tohskai:
thanks a lot for this!

PR #162877, opened 2025-09-13
[CI] Move libtorch-cpu-shared-with-deps-release-build to python 3.10

atalman:
Related to https://github.com/pytorch/pytorch/pull/162862


malfet (approved):
Thank you, I've incorporated the fix in my other PR, but good not having to worry about it

PR #162879, opened 2025-09-13
Enable computing Hessian matrices of larger functions by chunking in torch.func.hessian via jacrev

jmaczan:
In func.hessian I added passing `chunk_size` to `jacrev` to enable chunking when computing Hessian, so we can use bigger functions as input


jmaczan:
@zou3519 @mikaylagawarecki review ping

zou3519 (changes requested):
I don't think we want to take one-line changes to the API. Plus also the question is: why chunk jacrev and not jacfwd?

PR #162889, opened 2025-09-13
[functional] Avoid duplicate custom get_device call in constructor

anijain2305:
* #163019
* #162987
* __->__ #162889

Trying to reduce the number of `__torch_dispatch__` calls of FakeTensorMode in the AOT metadata collection pass.


Lucaskabela (approved):
Change lgtm - for reference, do we have a benchmark/result for how this improves compile time?

anijain2305:
> Change lgtm - for reference, do we have a benchmark/result for how this improves compile time?

It does not haha. I was hoping it would, but nope. Same with the next one in the stack.

zou3519 (approved):

PR #162899, opened 2025-09-14
[user-cuda-streams] Pass streams/events to the graph via lookup table

mlazos:
Stores streams in a global object look table that maps a dynamo selected index to objects. This index is generated during tracing, and at runtime, a helper function is called from the bytecode to populate this map. 

This differs from the previous implementation that simply mapped IDs to the associated objects. This required specialization on the IDs of the specific objects, while this new approach does not. 

* #162905
* #164523
* #165356
* #165212
* #165211
* #164819
* #164522
* #164304
* #162901
* #164507
* #164344
* #164343
* #162903
* #162900
* #163028
* __->__ #162899
* #163027


torch/_dynamo/utils.py
```diff
@@ -4687,7 +4687,50 @@ def _extract_tensor_dict(t: torch.Tensor) -> dict[str, Any]:
...
+index_to_source: dict[int, Source] = {}
 
+index_to_user_object_weakref: dict[int, weakref.ReferenceType[Any]] = {}
+
+
+def has_user_objects() -> bool:
+    return bool(index_to_source)
+
+
+def get_user_object_by_index(index: int) -> Any:
```
mlazos: Add notes for where each of these are used

torch/_dynamo/utils.py
```diff
@@ -4687,7 +4687,50 @@ def _extract_tensor_dict(t: torch.Tensor) -> dict[str, Any]:
...
 user_obj_id_to_weakref: dict[int, weakref.ReferenceType[object]] = {}
 
+# Similar to above, except instead of ID, we use a dynamo-generated index
+# this allows us to register objects externally in pre-graph bytecode that we want
+# to pass to the graph, but not support their types as graph inputs
+index_to_source: dict[int, Source] = {}
 
+index_to_user_object_weakref: dict[int, weakref.ReferenceType[Any]] = {}
+
+
```
mlazos: See if there's a way to make this stateless: think mutliple artifacts reading from this/ also precompile

mlazos: Maybe we could add this as an attribute on the graph

torch/_dynamo/output_graph.py
```diff
@@ -1413,6 +1415,21 @@ def compile_subgraph(
 
         from .decorators import disable
 
+        if has_user_objects():
```
anijain2305, line 1516: Add some comments on the index_to_user_object_weakref being a sync point between the pre-graph bytecode and compiled object.

torch/_dynamo/utils.py
```diff
@@ -4687,7 +4687,50 @@ def _extract_tensor_dict(t: torch.Tensor) -> dict[str, Any]:
 # e.g. torch.cuda.Event is a prime example.
 user_obj_id_to_weakref: dict[int, weakref.ReferenceType[object]] = {}
 
+# Similar to above, except instead of ID, we use a dynamo-generated index
```
anijain2305: Consider moving all of this to a separate file, where we can add these sync points

torch/_dynamo/variables/streams.py
```diff
@@ -70,11 +70,19 @@ def call_method(
                 ),
             )
         elif name in cmp_name_to_op_mapping and len(args) == 1 and not kwargs:
+            from ..guards import GuardBuilder, install_guard
+
+            if self.source:
+                install_guard(self.source.make_guard(GuardBuilder.EQUALS_MATCH))
```
anijain2305, line 76: Does EQUAL_MATCH really work? Do streams have __eq__ defined?

anijain2305 (approved):
Requesting a few changes. Pre-approving, but please consider those requests.

torch/_dynamo/variables/streams.py
mlazos: Yes

torch/_dynamo/graph_bytecode_inputs.py
```diff
@@ -0,0 +1,62 @@
+import weakref
+from typing import Any
+
+from torch._dynamo.source import Source
+
+
+# This file is to handle types that we don't want to support
+# as explicit FX graph inputs. This uses a sidetable which
+# we populate in bytecode and is loaded during graph execution
```
StrongerXi, line 9: qq: Looks like `register_user_object` only got invoked for objects created _outside_ the compile region. What about creating streams/events inside a compile region, what would happen and is that something we want to support?

mlazos: In the case of streams, we actually proxy the cosntructor into the graph, so we're okay for this. Although this doesn't allow us to output objects. I implement this in https://github.com/pytorch/pytorch/pull/164819 by constructing local streams outside the graph, passing them to the graph through the table, and if we need to return them return the object from the table that we passed to the graph.

PR #162900, opened 2025-09-14
[user-cuda-streams] Add fork/join custom ops

mlazos:
Creates the fork/join stream ops. These ops are passthrough ops which mutate all of their args (without actually performing any computation on them) so that during functionalization, implicit dependencies are added on all of their args. This allows us to prevent reordering during our pre/post grad graph passes. 

* #162905
* #164523
* #165356
* #165212
* #165211
* #164819
* #164522
* #164304
* #162901
* #164507
* #164344
* #164343
* #162903
* __->__ #162900
* #163028
* #162899
* #163027

Make custom ops inplace


torch/_dynamo/variables/streams.py
```diff
@@ -14,6 +14,35 @@
...
     from ..codegen import PyCodegen
 
+from torch._library.custom_ops import custom_op
+
+
+Tensor = torch.Tensor
+
+
+@custom_op("streams::fork", mutates_args={"args"})
+def fork_stream_(
```
mlazos: Make sure there are not extra copies around this op

torch/_dynamo/variables/streams.py
```diff
@@ -14,6 +14,35 @@
...
+    pass
+
+
+@fork_stream_.register_fake
+def _(index: int, device: torch.device, device_index: int, args: list[Tensor]) -> None:
+    pass
+
+
+@custom_op("streams::join", mutates_args={"args"})
+def join_stream_(
```
mlazos: We could add a custom functionalization

anijain2305 (approved):
You have recorded my concerns. Preapproving but please ensure that the functionalization does not insert extra copy.

PR #162906, opened 2025-09-14
Assoc scan mem eff

bohnstingl:
This PR replaces the matrix-based gradient computation of associative_scan with a backward associative_scan.
This enhances readability, maintainability and memory footprint. However, runtime needs to be checked.

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @ydwu4


ydwu4:
Maybe we can bring back this? @bohnstingl

bohnstingl:
Yes, happy to revive this effort

bohnstingl:
@yoyolicoris Thanks for reopening, I'll be looking into this.

yoyolicoris:
I just noticed that this version is only for pointwise combine_fn, right? Would be nice to make it general.

bohnstingl:
@yoyolicoris No, this is not quite right. 
The backward is not restricted to pointwise combine_fns. In the backward path, the associative_scan uses the combine_fn:

```
def g_ys_combine_fn_rev(bw, gl, bw_next, gl_next):
    return bw * bw_next, gl_next + bw_next * gl
```

This is unconditionally pointwise — only element-wise `*` and `+` — regardless of the user's forward `combine_fn`. It is true that the `bwys` values (instantaneous Jacobians `∂ys_{t+1}/∂ys_t`) depend on whatever the user's `combine_fn` is, i.e., it can be pointwise or not. However, the `g_ys_combine_fn_rev` only combines them element-wise. So the backward scan is always a pointwise operation, irrespective of whether the forward associative_scan uses combine_mode="generic".

bohnstingl:
@claude Please review these changes and provide your details in a GH collapsable section.

yoyolicoris:
> @yoyolicoris No, this is not quite right. The backward is not restricted to pointwise combine_fns. In the backward path, the associative_scan uses the combine_fn:
> 
> ```
> def g_ys_combine_fn_rev(bw, gl, bw_next, gl_next):
>     return bw * bw_next, gl_next + bw_next * gl
> ```
> 
> This is unconditionally pointwise — only element-wise `*` and `+` — regardless of the user's forward `combine_fn`. It is true that the `bwys` values (instantaneous Jacobians `∂ys_{t+1}/∂ys_t`) depend on whatever the user's `combine_fn` is, i.e., it can be pointwise or not. However, the `g_ys_combine_fn_rev` only combines them element-wise. So the backward scan is always a pointwise operation, irrespective of whether the forward associative_scan uses combine_mode="generic".

@bohnstingl thanks for the detailed explanation!
After a closer inspection, I found that `combine_mode="generic"` take a different forward function so its autograd path is different. Since this new backward only affect `combine_mode="pointwise"`, my concern is not an issue. The later added test cases is thus unnecessary imh. 

nit suggestion: you can use `torch.addcmul` to fuse some computation inside `g_ys_combine_fn_flat`

bohnstingl:
@claude I made several commits to integrate the review from https://github.com/pytorch/pytorch/pull/162906#issuecomment-5162781044. Please check whether all concerns are properly addressed

bohnstingl:
@claude please do a fresh review on this PR

bohnstingl:
I made another round of integrating code review. 

In addition, I also created a benchmarking script, see https://gist.github.com/bohnstingl/7ff9c0b19896b0cbe452eca4ccfb9bdb. 
One can clearly see that the peak memory behaves as expected. However, it reveals two performance bottlenecks:

1. **Working with many leaves**: As mentioned in the reviews and discussion above, there is a loop iterating over every leaf node to compute the gradients. This can be seen in the last table, that as the scan length increases and there are more leaves, we are approaching the limit imposed by iterating over the leaves in sequence. This is the primary performance bottleneck and will be reduced with vmap support for `associative_scan`. I am working on this and have a PR up soon.
2. **No SplitScan**: Although we use `associative_scan_op` in the backward as well, we currently cannot use the SplitScan that would help for long L, by splitting the scan into parts. This issue though is much less pressing than the first one.

```
python bench_assoc_scan_backward.py --compile
device=NVIDIA GeForce RTX 3090 iters=15 warmup=3

====================================================================
mode=compiled feature_dim=16 dtype=torch.float32
====================================================================

total fwd+bwd time (ms)
  leaves/L        1024        4096       16384       65536      262144
         1       0.295       0.312       0.280       0.536       1.311
         2       0.314       0.326       0.509       1.568       5.715
         4       0.407       0.432       0.838       2.616       9.652
         8       0.490       0.633       1.394       4.451      17.137

backward-only time (ms)
  leaves/L        1024        4096       16384       65536      262144
         1       0.173       0.188       0.157       0.348       0.877
         2       0.199       0.206       0.233       0.512       1.688
         4       0.276       0.271       0.371       0.967       3.277
         8       0.326       0.391       0.639       1.697       6.271

backward peak memory (MB)
  leaves/L        1024        4096       16384       65536      262144
         1        0.31        1.25        5.00       20.00       80.00
         2        0.56        2.25        9.00       36.00      144.00
         4        1.06        4.25       17.00       68.00      272.00
         8        2.07        8.25       33.00      132.00      528.00

--- per-leaf scaling of backward (fixed L) ---
         L        bwd n=1        bwd n=8   measured x    ideal x    ratio
      1024          0.173          0.326        1.88x       8.0x     0.24
      4096          0.188          0.391        2.08x       8.0x     0.26
     16384          0.157          0.639        4.08x       8.0x     0.51
     65536          0.348          1.697        4.87x       8.0x     0.61
    262144          0.877          6.271        7.15x       8.0x     0.89
```

bohnstingl:
@ydwu4 @yoyolicoris I think this PR should be ready for your review.

As identified, the major performance bottleneck now is the per-leaf loop. vmap for associative_scan would help there and I have already a draft PR locally for this.

bohnstingl:
I've raised a [PR](https://github.com/pytorch/pytorch/pull/192131) for the associative_scan vmap, which (in a follow-up PR) can alleviate the sequential loop over the scan leaves.

torch/_higher_order_ops/associative_scan.py
```diff
@@ -759,77 +752,57 @@ def backward(ctx, *gl_ys):
...
+            # 5.2) Flip, scan left-to-right, flip back, and drop the last (padding)
+            # element to get g_ys. We call the raw associative_scan_op HOP (not
+            # generic_associative_scan) so this scan is Triton-lowerable under compiled
+            # autograd.
+            result_rev = associative_scan_op(
+                g_ys_combine_fn_flat,
+                [bwys_pinit.flip([0]), gl_ys_pinit.flip([0])],
+                (),
+            )
+            g_ys = result_rev[1].flip([0])[:-1]
```
yoyolicoris: Suggest the following, which only requires `T` steps instead of `T+1` steps, and it looks cleaner.

```suggestion
            bwys_aligned = torch.cat(
                [bwys[1:], torch.ones_like(bwys[0:1])], dim=0
            )
            def g_ys_combine_fn_flat(bw, gl, bw_next, gl_next):
                return bw * bw_next, torch.addcmul(gl_next, tensor1=bw_next, tensor2=gl)

            # 5.2) Flip, scan left-to-right, flip back, and drop the last (padding)
            # element to get g_ys. We call the raw associative_scan_op HOP (not
            # generic_associative_scan) so this scan is Triton-lowerable under compiled
            # autograd.
            result_rev = associative_scan_op(
                g_ys_combine_fn_flat,
                [bwys_aligned.flip([0]), gl_ys.flip([0])],
                (),
            )
            g_ys = result_rev[1].flip([0])
```

test/functorch/test_control_flow.py
```diff
@@ -6447,6 +6447,72 @@ def test_associative_scan_non_pointwise(self):
...
+
+        x = torch.eye(2).unsqueeze(0).repeat(4, 1, 1).requires_grad_(True)
+        result = associative_scan(matmul_combine, x, dim=0, combine_mode="generic")
+        result_ref = _fake_associative_scan(matmul_combine, x, dim=0)
+
+        self.assertEqual(result, result_ref)
+
+        grads = torch.autograd.grad(result.sum(), x)
+        grads_ref = torch.autograd.grad(result_ref.sum(), x)
+        self.assertEqual(grads, grads_ref)
```
yoyolicoris: This should be removed, as the "generic" path never reaches the proposed backward path.
My question about whether this PR supports generic associative scan was a misunderstanding and can be neglected.
```suggestion
```

test/inductor/test_control_flow.py
```diff
@@ -1797,6 +1800,54 @@ def fct(x: torch.Tensor, y: torch.Tensor):
...
+
+        result = compiled_scan(matmul_combine, x, 0, combine_mode="generic")
+        result_ref = _fake_associative_scan(matmul_combine, x, 0)
+
+        self.assertEqual(result, result_ref)
+
+        if autograd:
+            grads = torch.autograd.grad(result.sum(), x)
+            grads_ref = torch.autograd.grad(result_ref.sum(), x)
+            self.assertEqual(grads, grads_ref)
```
yoyolicoris: Same reason as above.

```suggestion
```

test/functorch/test_control_flow.py
```diff
@@ -6447,6 +6447,72 @@ def test_associative_scan_non_pointwise(self):
...
+            return x + y + H1 + H2
+
+        xs = torch.randn(4, 2, device=device, requires_grad=True)
+        result = associative_scan(combine_fn, xs, dim=0, combine_mode="pointwise")
+        result_ref = _fake_associative_scan(combine_fn, xs, dim=0)
+        self.assertEqual(result, result_ref)
+
+        grads = torch.autograd.grad(result.sum(), xs)
+        grads_ref = torch.autograd.grad(result_ref.sum(), xs)
+        self.assertEqual(grads, grads_ref)
```
yoyolicoris, line 6467: I don't fully understand the necessity of this test ... The only difference between this and the above `test_associative_scan_pointwise_additional_input_autograd` is the number of no-grad additional tensors, and I didn't see any feature in this PR related to it.

yoyolicoris (changes requested):
I left a comment on the core implementation and some questions on the tests.

In addition, could you clarify the actual blocker for differentiable `additional_inputs`? L740 purposely drops them. Is it because the per-step Jacobian size `[T, P]` would be too large to keep, or something different?  I would like to know the motivation behind. Some context would be helpful.

torch/_higher_order_ops/associative_scan.py
bohnstingl: Good point, I followed your suggestion and adopted your suggested changes

test/functorch/test_control_flow.py
bohnstingl: I removed

test/inductor/test_control_flow.py
bohnstingl: I removed

test/functorch/test_control_flow.py
bohnstingl: Good point, I removed the almost duplicate test.

bohnstingl:
@yoyolicoris I have integrated your review, but moved the PR into the a different `ghstack` with vmap for the backward, see https://github.com/pytorch/pytorch/pull/162906.

Thus, closing this PR in favor of https://github.com/pytorch/pytorch/pull/162906

PR #162915, opened 2025-09-14
20% CPU and 15% CUDA speedup of func.hessian for scalar functions (opt-in is_scalar argument), default behavior untouched

jmaczan:
When using a new opt-in `is_scalar` argument for `hessian`, there's a consistent 15-20% speedup for all dimensions, with some high variance on CPU in range 500-700 dims (see a sortable benchmark table in Google Sheet below)

Currently, we use `jacrev` for all types of functions. I checked that `grad` works better for scalar functions. Ideally, we'd have an auto-tuner inside `hessian` to pick which version it should run, but I didn't worked it out yet (since it's not obvious if `func` is a scalar function or not)

Benchmark results: https://docs.google.com/spreadsheets/d/1Ie2Nudasz0yewBMrEyKGtMyJdM9tSAcUh_EeMQ3sHnU/edit?usp=sharing

I benchmarked all tested combinations (device (cpu/cuda), function (quadratic/rosenbrock), dimension(2,5,10,50,100,200,500,700,1000), version (base/optimized)) for 1000 times on a low noise machine (Ubuntu 24.04.2, AMD Ryzen 7 9800X3D, 64 GB RAM, no background activity except Terminal with benchmark script)

Benchmark script: https://github.com/jmaczan/pytorch-benchmarks/blob/main/hessian/benchmark.py

I added unit tests to test against regression

Sum of CPU+CUDA runtimes
Average CPU+CUDA runtimes
Median CPU+CUDA runtimes


jmaczan:
Oh btw I wonder if we can add `.venv/` to .gitignore? It's my default to store venv files locally. If that's a problem, then I'll revert the change from .gitignore

jmaczan:
@zou3519 review ping

zou3519 (commented):
I don't really want to complicate this API for scalar arguments. Maybe the question is: why is jacrev slower for scalar arguments?

PR #162925, opened 2025-09-14
[BE] Use `fmt::format` to define Conv key

malfet:
* __->__ #162925
* #162921

Also use `getArrayRefString` instead of having separate cases for 2D and 3D Conv

### DIFF
--- a/aten/src/ATen/native/mps/operations/Convolution.mm
+++ b/aten/src/ATen/native/mps/operations/Convolution.mm
@@ -6,6 +6,7 @@
 #include <ATen/ops/_mps_convolution_transpose_native.h>
 #include <ATen/ops/mps_convolution_backward_native.h>
 #include <ATen/ops/mps_convolution_transpose_backward_native.h>
+#include <fmt/format.h>
 
 namespace at::native {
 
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):

aten/src/ATen/native/mps/operations/Convolution.mm
```diff
@@ -172,39 +173,23 @@ static Tensor _mps_convolution_impl(const Tensor& input_t_,
...
-          std::to_string(padding[1]) + ":" + std::to_string(groups) + ":" + mem_format_key +
-          mps::getTensorsStringKey({input_t, weight_t}) + ":" + std::to_string(bias_defined) + ":" + bias_shape_key;
-    }
+    std::string key = fmt::format("mps_{}convolution:{}:{}:{}:{}:{}:{}:{}:{}",
+                                  is3DConv ? "3d_" : "",
+                                  getArrayRefString(stride),
+                                  getArrayRefString(dilation),
+                                  getArrayRefString(padding),
+                                  groups,
+                                  is_channels_last,
```
Skylion007, line 189: Not exactly the same as the original key, this is a boolean right?

Skylion007 (approved):
not sure about ths is_chanels_last as that will cause the key to change. Otherwise, looks good.

aten/src/ATen/native/mps/operations/Convolution.mm
malfet: Keys are not persistent, so this is fine, as long as they are unique

aten/src/ATen/native/mps/operations/Convolution.mm
```diff
@@ -172,39 +173,23 @@ static Tensor _mps_convolution_impl(const Tensor& input_t_,
...
+                                  mps::getTensorsStringKey({input_t, weight_t}),
+                                  bias_defined,
+                                  bias_shape_key);
```
Skylion007, line 192: Slight nit I should have noticed, bias shape key should have probably been std::move here

PR #162926, opened 2025-09-14
[MTIA] [PyTorch] [Export] Allow setting names for lifted constants before tracing.

patrick-toulme:
Summary:
The lift_constants_pass will set the name of lifted to constants to module_path_{lifted_tensor_counter}. This makes it very difficult for a compiler or runtime to map certain constants to what the placeholder value is in the IR. 

This PR adds a new API to Torch Export allowing users to set the names of certain tensors. The lift_constants_pass then respects this name. 

This is not supported for strict=True or for naming FakeTensors.

```
        tensor = torch.tensor([1, 2, 3])
        result = torch.export._name_tensor_constants.set_tensor_name(tensor, "my_custom_tensor")
        self.assertIs(result, tensor)  # Should return the same tensor
```

Before:
```
 %model_layers_0_self_attn_attn_lifted_tensor_0 : bf16[2, 346732, 16, 8, 64][num_users=2] = placeholder[target='model_layers_0_self_attn_attn_lifted_tensor_0']
```

```
 kv_cache = torch.export._name_tensor_constants.set_tensor_name(kv_cache, "kv_cache")
```

After:

```
%model_layers_0_self_attn_attn_kv_cache : bf16[2, 346544, 16, 8, 64][num_users=2] = placeholder[target='model_layers_0_self_attn_attn_kv_cache']
```

Test Plan:
Added unit tests

Rollback Plan:

Differential Revision: D82387095

### DIFF
--- a/torch/_export/passes/lift_constants_pass.py
+++ b/torch/_export/passes/lift_constants_pass.py
@@ -102,6 +102,114 @@ def get_constant_fqn(node: torch.fx.Node, constant_name: str) -> str:
         return constant_name
 
 
+def _get_constant_name_and_fqn(
+    node: torch.fx.Node,
+    constant_val: _ConstantAttributeType,
+    constant_attrs: ConstantAttrMap,
+    used_target_names: set[str],
+    num_tensor_constants: int,
+    num_custom_obj: int,
+) -> tuple[str, str, int, int]:
+    """
+    Single helper to get the name and FQN for all constants, handling custom names,
+    module attribute names, and default names with uniqueness checking.
+    """
+    if isinstance(constant_val, torch.Tensor):
+        # Check for custom name first
+        custom_name = getattr(constant_val, "_export_name", None)
+        if custom_name is not None:
+            return _get_unique_name(
+                node,
+                custom_name,
+                1,
+                used_target_names,
+                num_tensor_constants,
+                num_custom_obj,
+                is_custom_name=True,
+            )
+
+        # Fall back to module attribute name if available
+        constant_fqn = _get_first_fqn(constant_attrs, constant_val)
+        if constant_fqn is not None:
+            constant_name = constant_fqn.replace(".", "_")
+            return constant_name, constant_fqn, num_tensor_constants, num_custom_obj
+
+        # Use default tensor naming
+        return _get_unique_name(
+            node,
+            "",
+            num_tensor_constants,
+            used_target_names,
+            num_tensor_constants,
+            num_custom_obj,
+            is_custom_name=False,
+            prefix="lifted_tensor",
+        )
+
+    elif isinstance(constant_val, (torch.ScriptObject, FakeScriptObject)):
+        # Fall back to module attribute name if available
+        constant_fqn = _get_first_fqn(constant_attrs, constant_val)
+        if constant_fqn is not None:
+            constant_name = constant_fqn.replace(".", "_")
+            return constant_name, constant_fqn, num_tensor_constants, num_custom_obj
+
+        # Use default custom object naming
+        return _get_unique_name(
+            node,
+            "",
+            num_custom_obj,
+            used_target_names,
+            num_tensor_constants,
+            num_custom_obj,
+            is_custom_name=False,
+            prefix="lifted_custom",
+        )
+
+    else:
+        raise SpecViolationError(
+            f"Unsupported constant type {type(constant_val)} in _get_constant_name_and_fqn"
+        )
+
+
+def _get_unique_name(
+    node: torch.fx.Node,
+    base_name: str,
+    counter: int,
+    used_target_names: set[str],
+    num_tensor_constants: int,
+    num_custom_obj: int,
+    is_custom_name: bool = False,
+    prefix: str = "",
+) -> tuple[str, str, int, int]:
+    """Get a unique name, handling both custom names and default names with counter increments."""
+    if is_custom_name:
+        # For custom names, start with the base name and add suffixes if needed
+        constant_name = base_name
+        counter = 1
+    else:
+        # For default names, start with prefix_counter format
+        constant_name = f"{prefix}_{counter}"
+
+    constant_fqn = get_constant_fqn(node, constant_name)
+
+    while constant_fqn in used_target_names:
+        counter += 1
+        if is_custom_name:
+            constant_name = f"{base_name}_{counter}"
+        else:
+            constant_name = f"{prefix}_{counter}"
+        constant_fqn = get_constant_fqn(node, constant_name)
+
+    # Return updated counters
+    if is_custom_name:
+        # Custom names don't affect the default counters
+        return constant_name, constant_fqn, num_tensor_constants, num_custom_obj
+    elif prefix == "lifted_tensor":
+        return constant_name, constant_fqn, counter + 1, num_custom_obj
+    else:  # prefix == "lifted_custom"
+        return constant_name, constant_fqn, num_tensor_constants, counter + 1
+
+
 def _get_first_fqn(
     const_attrs: ConstantAttrMap,
     key: _ConstantAttributeType,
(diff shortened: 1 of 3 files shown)


jansel (changes requested):
Failing tests?

angelayi (commented):
I believe we only name things "...lifted_tensor" when there is a `torch.tensor(..)` call in the graph, not when it is set as an attribute on init, so it's not clear to me what issue you're running into. Could you check your test cases?

This example should have a constant called `c_constant_tensor`. 
```
        class SimpleModule(torch.nn.Module):
            def __init__(self):
                super().__init__()
                # Create a named tensor that will be lifted
                self.constant_tensor = torch.tensor([10, 20, 30])

            def forward(self, x):
                return x + self.constant_tensor
```

Whereas this example should have a constant called `c_lifted_tensor_0`:
```
        class SimpleModule(torch.nn.Module):
            def __init__(self):
                super().__init__()

            def forward(self, x):
                return x + torch.tensor(1)
```

patrick-toulme:
@angelayi  I will send you my workload internally to show you why I need this

test/export/test_tensor_naming.py
```diff
@@ -0,0 +1,180 @@
...
+
+        # Get original output
+        original_output = module(input_tensor)
+
+        # Export and run
+        exported_program = torch.export.export(module, (input_tensor,))
+        exported_output = exported_program.module()(input_tensor)
+
+        # Results should be identical
+        torch.testing.assert_close(original_output, exported_output)
```
angelayi: These test cases don't feel representative of the issue you were running into, which is where the kv cache tensor is being created outside of the exported function. In these test cases, the tensors are being registered as buffers already so should already have a good name. Can you update the test cases to match this?

Additionally, lifted tensors show up when a user calls torch.tensor(1) within the model code. Can you write a test case that uses the new naming API to set those names and verify that this works?

And finally can you add this test to test_export.py and use `export()` instead of `torch.export.export` -- this will test multiple flavors of export, strict/non-strict, serialization/deserialization. I suspect your test will fail in `export(..., strict=True)`

docs/source/torch.rst
```diff
@@ -768,6 +768,16 @@ Export Path
...
 
+Export Functions
+~~~~~~~~~~~~~~~~
+.. autosummary::
+    :toctree: generated
+    :nosignatures:
+
+    export.get_tensor_name
+    export.name
+    export.set_tensor_name
```
angelayi: let's keep this as a private api

torch/export/__init__.py
```diff
@@ -53,6 +56,76 @@
 
 PassType = Callable[[torch.fx.GraphModule], Optional[PassResult]]
 
+
+# Tensor naming utilities for torch export
+def set_tensor_name(tensor: torch.Tensor, name: str) -> torch.Tensor:
```
angelayi: can you make these apis private for now? Maybe put them under `torch/export/_name_tensor_constants.py`

torch/export/__init__.py
```diff
@@ -53,6 +56,76 @@
...
+            f"Name '{name}' contains invalid characters. Only alphanumeric, underscore, and dot are allowed."
+        )
+
+    # Store the name as a custom attribute on the tensor itself
+    # This survives tensor transformations better than WeakKeyDictionary
+    tensor._export_name = name  # type: ignore[attr-defined]
+    return tensor
+
+
+def name(tensor: torch.Tensor, custom_name: str) -> torch.Tensor:
```
angelayi: Let's just keep the `set_tensor_name` api and not have this alias

angelayi (dismissed):
I'm not opposed to theoretically adding an api to hint to export the names for these tensor constants, but I don't think this works for export(..., strict=True). An easier fix would be to rewrite the original code to register the constant as a buffer.

torch/export/__init__.py
patrick-toulme: will fix

torch/export/__init__.py
patrick-toulme: will fix this

docs/source/torch.rst
patrick-toulme: will fix

test/export/test_tensor_naming.py
patrick-toulme: Will make better test cases and use torch.export.export

patrick-toulme:
> I'm not opposed to theoretically adding an api to hint to export the names for these tensor constants, but I don't think this works for export(..., strict=True). An easier fix would be to rewrite the original code to register the constant as a buffer.

as we discussed, I am not allowed to rewrite vLLM code and it would be a major rewrite of vLLM. I will make the other changes you requested. Thank you!

zhxchen17 (dismissed):
Not sure about adding a private interface that will only benefit 1 group of torch.export user, especially when node renaming is doable out of export.

This can get complicated for maintenance in the long term, when tensor subclasses are involved (which decomposes a tensor and break name convention a user would like to establish here). This is also fragile if we re-export the original module in processing pipeline which can easily screw up the node names.

A better approach is to add separate metadata to placeholder nodes indicating the "sources" of input tensors (can be dynamo sources or whatever) regardless of annotating tensors. This is more likely to scale to different use cases and useful for debugging purposes anyway.

patrick-toulme:
@zhxchen17  Let's discuss internally. This is blocking any further development of vLLM for MTIA graph mode.

patrick-toulme:
> Not sure about adding a private interface that will only benefit 1 group of torch.export user, especially when node renaming is doable out of export.
> 
> This can get complicated for maintenance in the long term, when tensor subclasses are involved (which decomposes a tensor and break name convention a user would like to establish here). This is also fragile if we re-export the original module in processing pipeline which can easily screw up the node names.
> 
> A better approach is to add separate metadata to placeholder nodes indicating the "sources" of input tensors (can be dynamo sources or whatever) regardless of annotating tensors. This is more likely to scale to different use cases and useful for debugging purposes anyway.

Let me add more context to this PR, 

1. MTIA is not using dynamo, we are using export(strict=False) in our compiler. 

We are in a situation where vLLM creates raw tensors for a KV cache here 

https://github.com/vllm-project/vllm/blob/e3db5ebb66590031ecfd3338de41b6f1ee95bf2a/vllm/v1/worker/gpu_model_runner.py#L3581

This is a torch.zeros() call outside of our jitted graph at model init time. 


We then jit the entire forward of the model to make one graph for the entire model. 

The KV Cache is of shape [2, 346732, 16, 8, 64]. 


vLLM calls .unbind() in the forward graph

https://github.com/vllm-project/vllm/blob/e3db5ebb66590031ecfd3338de41b6f1ee95bf2a/vllm/v1/attention/backends/xformers.py#L356

This decomposes into 
```
kv_cache = lifted_constant(kv_cache)
key = slice(kv_cache)
reshaped_key = reshape(slice)
```

Moving on to MTIA runtime / compiler:

1. We support in-place writes to graph inputs (kv_cache is a graph input here). Our compiler internally will transform our lower end internal dialect to write in-place.
2. The user has to tell us the targets of placeholders that the user would like us to write in-place too
3. In our runtime, the user has to provide us a raw map of placeholder target to raw tensor object. 

In the current form of lifted_tensor_# name, the user has no way to pass to MTIA Runtime/Compiler the map of placeholder targets to raw tensors. 


This PR is quite small and maintains existing behavior of PyTorch. If we cannot land, this I really need PyTorch to provide some solution to my problem above. 

Also, we are not allowed to modify vLLM code to use register_buffer() etc. That would be a massive refactor of vLLM.

torch/_export/passes/lift_constants_pass.py
```diff
@@ -274,17 +275,35 @@ def lift_constants_pass(
...
+                    else:
+                        # Use default naming as last resort
                         constant_name = f"lifted_tensor_{num_tensor_constants}"
                         constant_fqn = get_constant_fqn(node, constant_name)
-                    num_tensor_constants += 1
+                        while constant_fqn in used_target_names:
+                            num_tensor_constants += 1
+                            constant_name = f"lifted_tensor_{num_tensor_constants}"
+                            constant_fqn = get_constant_fqn(node, constant_name)
+                        num_tensor_constants += 1
```
jansel: Duplicate code with the above?

I'd suggest having a single helper to get the name, rather than both `get_tensor_name` and `get_constant_fqn`.   And remove the `if custom_name is not None` branch from this function.

jansel (changes requested):

jansel (approved):

zhxchen17 (changes requested):
sorry, my concerns not addressed

angelayi (changes requested):
+1 my concerns were also not addressed :) also I believe we've found a better solution in https://github.com/pytorch/pytorch/pull/163236

PR #162927, opened 2025-09-14
Fix: ShapeEnv not propagated properly to inductor SizeVars

laithsakka:
Summary:
I am really skeptical about inductor sizevars creating an empty shape env when not provided with one
i think we should fail there if the graph has dynamic shapes and no shape env is provided. 

however i wonder if there are actually use cases that depends on the shape env not being there?
Reasoning APIs depends on facts in the shape env. and assumes some stuff exists for specific symbols.

Test Plan:
Fix the bug reported in creating simple e2e unit test is not trivial
https://www.internalfb.com/diff/D82337184

Rollback Plan:

Differential Revision: D82412384

### DIFF
--- a/torch/_inductor/compile_fx.py
+++ b/torch/_inductor/compile_fx.py
@@ -1227,7 +1227,9 @@ class _InProcessFxCompile(FxCompile):
             # structured logs...
             # trace_structured("inductor_input_graph", payload_fn=lambda: gm.print_readable(print_output=False))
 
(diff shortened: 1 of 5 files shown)


ezyang (approved):
I'm OK to unblock, I am kind of meh groveling in sizes/strides of tensors (it should be a lot easier to get it), test case would really help.

ezyang (approved):

eellison (approved):

jansel (approved):

PR #162941, opened 2025-09-15
[BE] refine import sympy

thenumberouscode:
I've consolidated all the scattered import sympy statements in sym_node.py into a single class. This makes the code cleaner and ensures sympy is only imported once, improving efficiency.

### DIFF
--- a/torch/fx/experimental/sym_node.py
+++ b/torch/fx/experimental/sym_node.py
@@ -16,6 +16,7 @@ to avoid having to load SymPy at import time, as doing so is *very* slow.
 
 import builtins
 import functools
+import importlib
 import inspect
 import itertools
 import logging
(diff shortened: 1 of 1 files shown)


Skylion007:
In Python modules are only imported the first time they are referenced in the execution path and then cached...

Skylion007 (changes requested):
Pretty sure Python already does this...

thenumberouscode:
> Pretty sure Python already does this...

Yes, I've also tested it in my local dev, you are right.  so how should I change my code? I'm a bit stuck since python already done the cache.

thenumberouscode:
@Skylion007 Could you please review the PR if you have an moment? thanks

torch/fx/experimental/sym_node.py
```diff
@@ -65,6 +67,18 @@ def _to_symtype(t):
     return t
 
 
+class LazySymPy:
```
Skylion007: This will erase all the typehints from the Sympy via the dynamic __getattr__, is that worth the marginal performance improvement from this check?

I mean you might be able to get around with that a static if else TYPE_CHECKING block.

thenumberouscode: Awesome! Let's get started on this.

cyyever (commented):
Why introduce `_LazySymPy`?

cyyever:
@thenumberouscode I still can't understand its purpose. Directly `import sympy` on the top is more clear..

cyyever (approved):

cyyever:
I am fine, but you need others' approval to merge

PR #162942, opened 2025-09-15
[torch][cuda][device_limits] Library for querying device hardware limits for flops and bandwidth

valentinandrei:
In various benchmarks scattered across the repo, the limits for flops/second and memory bandwidth are usually hardcoded for a single device. This utility could help in providing a more structured way to query the device capabilities. If this is approved, we can use it when reporting flops efficiency and bandwidth relative to peak in the benchmarks and tests. The intent is to add more devices, more parameters (e.g. L2 cache bandwidth, NVLink, etc.) for both CPUs and accelerators.

Testing:

```
import torch

if torch.cuda.is_available():
    device = torch.cuda.current_device()
    mod = torch.get_device_module('cuda')
    hw = mod._device_limits.GPULimits(device)

    print(hw.get_tflops_per_second(torch.float16))
    print(hw.get_tflops_per_second(torch.float32))
    print(hw.get_tflops_per_second(torch.float64))
    print(hw.get_tflops_per_second(torch.bfloat16))
    print(hw.get_tflops_per_second(torch.int8))
    print(hw.get_memory_bandwidth_Bps() / 1e9)
    print(hw.get_shared_memory_bandwidth_Bps() / 1e9)

# Output on an H100 GPU
1070.53056
535.26528
66.90816
1070.53056
2141.06112
4893.696
33454.08
```


ngimel (approved):
tbh I don't think we need both GBps and GiBPs versions, but otherwise looks good

torch/accelerator/_device_limits/_hardware_peak.py
```diff
@@ -0,0 +1,63 @@
...
+
+
+def get_tflops_per_second_per_dtype(
+    target_device: torch.device,
+    data_type: dtype,
+) -> float:
+    if torch.cuda.is_available():
+        return nv.get_tflops_per_second(target_device, data_type)
+    else:
+        warnings.warn(
```
albanD: We should raise an error when we don't know unless we have a good reason why this default value is actually useful in many cases.
Same for the others below.

eellison:
cc @exclamaforte @v0i0

torch/accelerator/_device_limits/_hardware_peak.py
```diff
@@ -0,0 +1,63 @@
+import warnings
+
+import torch
+import torch.accelerator._device_limits._nvidia_gpu as nv
```
albanD: Ho sorry I think my earlier statement about the move was a bit confusing.
You can see the pattern in  https://github.com/pytorch/pytorch/blob/9b7a8c4d0551b105f5ad7cebf7b8db408be82581/torch/accelerator/__init__.py#L46-L62

For this PR, that would means:
- Move _nvidia_gpu.py into torch.cuda._device_limits.
- Make this code indirect for cuda-specific pieces using get_device_module
- Keep the shared logic here as it is.

torch/accelerator/_device_limits/_nvidia_gpu.py
```diff
@@ -0,0 +1,148 @@
...
+        dict_key = "unknown"
+
+    if dict_key not in hardcoded_device_values.keys():
+        # We return 1 because typically the peak FLOPs/sec value is used to compute
+        # MFU. We want to avoid a division by zero.
+        warnings.warn(
+            f"No data for sm_{compute_capability} and {data_type}. Returning default value of 1.",
+            UserWarning,
+        )
+        return 1
```
albanD: Definitely an error on this one!

torch/accelerator/_device_limits/_hardware_peak.py
```diff
@@ -0,0 +1,63 @@
...
+        return bytes_per_second / 1e9
+    else:
+        warnings.warn(
+            "Memory bandwidth is only supported for CUDA devices. Returning default value of 1.0.",
+            UserWarning,
+        )
+    return 1.0
+
+
+def get_memory_bandwidth_GiBps(target_device: torch.device) -> float:
```
albanD: Curious why you have Gi only for this one and not the others?

albanD (commented):
I'm curious where you plan to use these APIs?
If you plan on using them for anything important, we should aim to make them public ASAP once we have feedback about the shared API for other hw.

torch/accelerator/_device_limits/_hardware_peak.py
valentinandrei: I removed this one per @ngimel 's suggestion.

torch/accelerator/_device_limits/_nvidia_gpu.py
```diff
@@ -0,0 +1,148 @@
...
+        "fp16_80": 1024,
+        "fp32_80": 512,
+        "fp64_80": 64,
+        # Hopper Architecture
+        "int8_90": 4096,
+        "fp8_90": 4096,
+        "fp16_90": 2048,
+        "fp32_90": 1024,
+        "fp64_90": 128,
+        # Blackwell Architecture
```
v0i0: shouldnt blackwell be 2x hopper? how are we testing / deriving this right now?

torch/accelerator/_device_limits/_nvidia_gpu.py
```diff
@@ -0,0 +1,148 @@
...
+# https://resources.nvidia.com/en-us-hopper-architecture/nvidia-h100-tensor-c
+
+
+# Returns the number of FMA instructions retired per cycle per SM for a given
+# data type, when tensor cores are NOT used
+def get_fma_per_cycle_per_sm_cuda_cores(
+    compute_capability: int, data_type: dtype
+) -> int:
+    hardcoded_device_values = {
+        # Ampere Architecture
```
v0i0: there aren't that many CCs, might not be hard to complete these: https://developer.nvidia.com/cuda-gpus

valentinandrei: Yep, that's the plan. I'll probably add those after some checks.

torch/accelerator/_device_limits/_nvidia_gpu.py
valentinandrei: My understanding is that Blackwell Tensor Cores' generation 5 added FP4 at double the throughput vs FP8, but in the rest of the data types have the same FMA rate per cycle as Hopper. The 2X at GPU level might come from increased SM count and higher frequency compared to Hopper.

I'll double check with a GEMM that is not frequency throttled before landing.

valentinandrei: @v0i0 I checked with a CUTLASS sample and ran it at stable frequency and yes, the FMA rate of the tensor cores doubled from Hopper to Blackwell. I updated the values. Thanks for catching this.

eqy:
Note that this is using APIs removed in CUDA 13 and breaks CUDA 13 builds: https://docs.nvidia.com/cuda/archive/13.0.0/cuda-toolkit-release-notes/index.html#id6

These are also already deprecated in CUDA 12 IIUC

valentinandrei:
@huydhn @albanD @eqy  I changed the code to ensure forward compatibility per @eqy 's suggestion. The failing tests seem to be unrelated.

- The slow-gradcheck tests fail on trunk currently
- The legacy_nvidia_driver tests are marked as flaky
- The inductor tests that fail numerics seem to do so because they run on M60 GPUs on bfloat16 tensors
- The s390 tests fail with various reasons (one is a rust build failure).

Can you please review and check if the changes make sense and should be a safe land?

albanD (approved):
Thanks!

torch/csrc/cuda/Module.cpp
```diff
@@ -1053,6 +1053,25 @@ static void registerCudaDeviceProperties(PyObject* module) {
       .def_readonly("warp_size", &cudaDeviceProp::warpSize)
 #ifndef USE_ROCM
       // NVIDIA-only properties
+      .def_property_readonly(
+          "clock_rate",
+          [](const cudaDeviceProp&) {
```
ngimel, line 1058: should be able to avoid passing unused arg

PR #162959, opened 2025-09-15
[export] handling NamedTuple inputs

### LINKED ISSUE #160547: [export] Non-strict export fails when tracing with NamedTuple input
### 🐛 Describe the bug

```
    def test_namedtuple(self):
        from collections import namedtuple
        Point = namedtuple('Point', 'x y')
        
        class M(torch.nn.Module):
            def forward(self, x, y):
                return x + y 
        
        inp = Point(torch.ones(3), torch.ones(3))
        print(M()(*inp))
        
        # errors
        ep = torch.export.export(M(), inp, strict=False)
        print(ep)

        # succeeds
        ep = torch.export.export(M(), inp, strict=True)
        print(ep)

        # workaround could be to convert namedtuple to a kwarg
        inp_kwargs =  {field: getattr(inp, field) for field in inp._fields}
        ep = torch.export.export(M(), (), inp_kwargs)
        print(ep)
```

### Versions

main

Raman-RH:
Fixes #160547
### Summary:
bug
```
    def test_namedtuple(self):
        from collections import namedtuple
        Point = namedtuple('Point', 'x y')
        
        class M(torch.nn.Module):
            def forward(self, x, y):
                return x + y 
        
        inp = Point(torch.ones(3), torch.ones(3))
        print(M()(*inp))
        
        # errors
        ep = torch.export.export(M(), inp, strict=False)
        print(ep)

        # succeeds
        ep = torch.export.export(M(), inp, strict=True)
        print(ep)

        # workaround could be to convert namedtuple to a kwarg
        inp_kwargs =  {field: getattr(inp, field) for field in inp._fields}
        ep = torch.export.export(M(), (), inp_kwargs)
        print(ep)
```
FIx :
namedtuple is subclass of tuple 
but namedtuple is not expected 
So, this change handles named tuple case

I have added 🧪 test case for this as well

### DIFF
--- a/test/export/test_export.py
+++ b/test/export/test_export.py
@@ -16017,6 +16017,26 @@ def forward(self, q, k, v):
         ):
             export(Foo(), (torch.randn(1, 33, 256, 128), k, v))
 
+    def test_namedtuple_input_export(self):
+        # test for NamedTuple inputs with both strict and non-strict export modes
+        from collections import namedtuple
+
+        PointNT = namedtuple("PointNT", ["x", "y"])
+
+        class M(torch.nn.Module):
+            def forward(self, x, y):
+                return x + y
+
+        inp = PointNT(torch.ones(3), torch.ones(3))
+
+        ep_non_strict = export(M(), inp, strict=False)
+        result_non_strict = ep_non_strict.module()(*inp)
+
+        ep_strict = export(M(), inp, strict=True)
+        result_strict = ep_strict.module()(*inp)
+
+        self.assertEqual(result_non_strict, result_strict)
+
 
 @unittest.skipIf(not torchdynamo.is_dynamo_supported(), "dynamo isn't support")
 class TestOneOffModelExportResult(TestCase):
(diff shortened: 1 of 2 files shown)


Raman-RH:
@angelayi @yushangdi @yiming0416 @desertfire @eellison @davidberard98 @ydwu4 @pianpwk 
let me know if any additional change required

test/export/test_export.py
```diff
@@ -16017,6 +16017,26 @@ def forward(self, q, k, v):
...
+
+        PointNT = namedtuple("PointNT", ["x", "y"])
+
+        class M(torch.nn.Module):
+            def forward(self, x, y):
+                return x + y
+
+        inp = PointNT(torch.ones(3), torch.ones(3))
+
+        ep_non_strict = export(M(), inp, strict=False)
```
angelayi: Just called export() should be enough! We have some tests that override `export` with different flags, like strict=True/False, so it should cover it!

```suggestion
        ep_non_strict = export(M(), inp)
```

angelayi (approved):
thanks for the fix!

test/export/test_export.py
Raman-RH: Yes, export(M(), inp) without explicitly passing strict=False should already exercise the non-strict path, since the default is strict=False

Issue #162978, opened 2025-09-15
[CI] run_test.py mutes test errors


kwen2501:
### 🐛 Describe the bug

Without run_test.py:
```
$ python test/distributed/test_nvshmem_triton.py
... some failures

$ echo $?
1
```

With run_test.py:
```
$ python test/run_test.py --include distributed/test_nvshmem_triton.py
... some failures

$ echo $?
0
```

Example:
https://github.com/pytorch/pytorch/actions/runs/17716027762/job/50342092319?pr=162529

Not sure if the muting of error is related to the "RERUN" feature in run_test.py.

### Versions

main

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci @seemethere @malfet @pytorch/pytorch-dev-infra

clee2000:
It doesn't surface because of
https://github.com/pytorch/pytorch/blob/fa919feab6a55ea9104e4ce61d38c3725f5728e6/torch/testing/_internal/common_distributed.py#L1775

The test fails so it gets queued up for a rerun, which then gets skipped because of the above, so everything looks ok

kwen2501:
@clee2000 thanks for looking into this. 
I do want to skip the rest of sub-tests once one failed, which means something like `self.__class__.poison_pill` is somewhat necessary. 
It would leave us only option I guess?
- Disable rerun
cc @malfet

malfet:
@kwen2501 this issues gives me an eerie sense of de ja vu, i.e. distributed team wanted to have a bespoke testrun solution, but then also wanted to be able to disable test thru issues, auto-detect flaky test and benefit from automated sharding 

If none of those are important right now, feel free to use `pytest`, but than I guess one should also understand that you are on your own with monitoring, reporting, dashboarding etc

malfet:
> It doesn't surface because of
> 
> [pytorch/torch/testing/_internal/common_distributed.py](https://github.com/pytorch/pytorch/blob/fa919feab6a55ea9104e4ce61d38c3725f5728e6/torch/testing/_internal/common_distributed.py#L1775)
> 
> Line 1775 in [fa919fe](/pytorch/pytorch/commit/fa919feab6a55ea9104e4ce61d38c3725f5728e6)
> 
>  if self.__class__.poison_pill: 
[...]

PR #162981, opened 2025-09-15
[SymmMem] Not use run_test.py in CI

kwen2501:
* __->__ #162981

run_test.py can mute errors in test (mechanism unknown, issue filed #162978).
Replacing it with bare pytest in SymmMem CI commands.

### DIFF
--- a/.ci/pytorch/test.sh
+++ b/.ci/pytorch/test.sh
@@ -339,10 +339,10 @@ test_h100_distributed() {
 
 test_h100_symm_mem() {
   # symmetric memory test
-  time python test/run_test.py --include distributed/test_symmetric_memory.py  $PYTHON_TEST_EXTRA_OPTION --upload-artifacts-while-running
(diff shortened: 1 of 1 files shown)


.ci/pytorch/test.sh
```diff
@@ -339,10 +339,10 @@ test_h100_distributed() {
 
 test_h100_symm_mem() {
```
nWEIdia, line 340: Does this have to be `test_h100_symm_mem`? 
I mean could this support test_b200_symm_mem without code changes? 
There is not much that is specific to h100, right? 
Background: https://github.com/pytorch/pytorch/pull/162988

malfet (changes requested):
What's your test plan? Lots of those tests might not have a main routine (distributed tests are notorius for doing that), so invoking `distribtued/test_nccl.py` can be just a noop

Also, doing things like that would ignore all tests disabled by issue, will not re-run failed tests and might not report the test stats correctly

kwen2501:
Close in favor of https://github.com/pytorch/pytorch/pull/163025

PR #162983, opened 2025-09-15
[precompile] Store traced file information with CompileArtifacts.

zhxchen17:
Summary:
Add some metadata to CompileArtifacts, so that it contains the source code information about the original code while they are being traced.

For now, we will not provide a verification method to end user and instead we just provide which files are inlined. It's up to user to verify the content from these files are not changed (because it's optional for many users to validate source code changes anyway in aot precompile)

Test Plan:
buck run @mode/opt test/dynamo:test_dynamo -- -k test_file_change
buck run @mode/opt test/dynamo:test_dynamo -- -k test_aot_compile_source_info

Fixes #ISSUE_NUMBER

### DIFF
--- a/torch/_dynamo/output_graph.py
+++ b/torch/_dynamo/output_graph.py
@@ -506,6 +506,7 @@ class OutputGraph(OutputGraphGuardsState):
             )
         self.tracing_context: TracingContext = TracingContext(fake_mode)
         self.tracing_context.traced_code.append(f_code)
+        self.traced_code = self.tracing_context.traced_code
         self.dynamo_compile_id: Optional[CompileId] = (
             CompileContext.current_compile_id()
         )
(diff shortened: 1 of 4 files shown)


torch/_dynamo/output_graph.py
```diff
@@ -506,6 +506,7 @@ def __init__(
             )
         self.tracing_context: TracingContext = TracingContext(fake_mode)
         self.tracing_context.traced_code.append(f_code)
+        self.traced_code = self.tracing_context.traced_code
```
yushangdi, line 509: why do we need to add a `traced_code` field here? We can just retrieve it like `self.tracing_context.traced_code`?

zhxchen17: self.tracing_context is cleaned up after dynamo finishes (see `def cleanup` in output_graph), so we want to still make it alive after that.

yushangdi (approved):

PR #162984, opened 2025-09-15
[BE] Adding aliases for CUDA and XPU API documentation

jiannanWang:
This PR reorganizes CUDA and XPU API documentation with additional aliases pages. Multiple entries of APIs under torch.cuda are thus removed.


janeyx99 (approved):
Thanks for the change. I've looked through and then spotchecked `get_rng_state` and it looks good.

PR #162990, opened 2025-09-15
DTensor: C++ compute_global_tensor_info

swolchok:
* #163667
* __->__ #162990

compute_global_tensor_info is on the hot path for DTensor.{from,to}_local. More incremental progress toward C++.


torch/csrc/autograd/python_variable.cpp
```diff
@@ -1122,6 +1135,124 @@ static PyObject* DTensor_OpSchema_post_init(PyObject* mod, PyObject* self) {
...
+  py::object mesh_size;
+  for (const auto& placement : placements) {
+    // TODO: C++ify Placement and DeviceMesh somehow; profiling seems
+    // to say that nearly all our remaining time spent is spent
+    // calling back into Python.
+    if (checked_is_true(
+            placement.attr(dtensor_interned_strings.is_shard)().ptr())) {
+      const auto shard_dim =
+          py::cast<int64_t>(placement.attr(dtensor_interned_strings.dim));
+      TORCH_CHECK(
```
Skylion007, line 1173: Don't we have a TORCH_CHECK_ASSERT to raise an Assertion Error?

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1122,6 +1135,124 @@ static PyObject* DTensor_OpSchema_post_init(PyObject* mod, PyObject* self) {
...
+    }
+    idx++;
+  }
+  return py::make_tuple(
+             int_array_to_list(tensor_shape), int_array_to_list(tensor_strides))
+      .release()
+      .ptr();
+}
+
+static const char compute_global_tensor_info_doc[] =
```
Skylion007: Make it constexpr too

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1122,6 +1135,124 @@ static PyObject* DTensor_OpSchema_post_init(PyObject* mod, PyObject* self) {
...
+    "    tensor_shape: A List of int which specifies the size of DTensor which build\n"
+    "        on top of the local tensor.\n"
+    "    tensor_stride: A List of int which specifies the stride of DTensor.\n";
+
+static PyObject* DTensor_compute_global_tensor_info(
+    PyObject* self,
+    PyObject* const* args,
+    Py_ssize_t nargs) {
+  HANDLE_TH_ERRORS
+  TORCH_CHECK(
```
Skylion007: TORCH_CHECK_VALUE

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1122,6 +1135,124 @@ static PyObject* DTensor_OpSchema_post_init(PyObject* mod, PyObject* self) {
...
+        }
+      }
+    } else if (
+        checked_not(
+            placement.attr(dtensor_interned_strings.is_replicate)().ptr()) &&
+        checked_not(
+            placement.attr(dtensor_interned_strings.is_partial)().ptr())) {
+      const auto placement_type_name =
+          py::str(py::handle(PyType_GetName(Py_TYPE(placement.ptr()))));
+      return PyErr_Format(
```
Skylion007, line 1209: You want to return the Error instead of throwing it?

swolchok: This does raise an exception:
- PyErr_Format flags an error
- PyErr_Format always returns nullptr, which is what you return while raising an exception
- the previous PR did add a test for this

torch/csrc/autograd/python_variable.cpp
swolchok: OK, that seems perfectly reasonable. note that it won't do anything in particular. constexpr is a "may", not a "must", and the `static const` part will already permit this to be initialized at compile time.

torch/csrc/autograd/python_variable.cpp
swolchok: not that I found. TORCH_INTERNAL_ASSERT exists but doesn't seem to do anything differently w.r.t. the exception it raises.

swolchok:
> will try TORCH_LIBRARY

Hmm, making this a custom op seems bad, at least right now. The function signature includes DeviceMesh (not currently IValue-able) and a list of Placements (also not currently IValue-able). I guess maybe I'm supposed to 1) add the Placement stuff from https://github.com/pytorch/pytorch/pull/163030 to IValue 2) have a C++ DeviceMesh representation 3) add that DeviceMesh representation to IValue 4) teach the custom operator / native function schema stuff about all that and then 5) sequence this PR after all those things?

On the other hand, torch.compiler.allow_in_graph might make this easy. I'll try that first.

swolchok:
> On the other hand, torch.compiler.allow_in_graph might make this easy. I'll try that first.

Simply adding
```
        import torch._C
        torch.compiler.allow_in_graph(torch._C._DTensor_compute_global_tensor_info)
```

to the start of test_dtensor_dynamic does not cause it to pass. @ezyang does it sound right that C++ifying utility functions that happen to work with Tensors has to be done through custom operators, which necessitates making all their argument types (i.e., DeviceMesh, Placement) go through IValue?

ezyang:
They look ignorable. Is there a reason you put periodic label on it?

test/inductor/test_distributed_patterns.py
```diff
@@ -3,6 +3,7 @@
 import functools
 
 import torch
+import torch.distributed.fsdp
```
ezyang: weird

torch/csrc/utils/python_compat.h
```diff
@@ -9,11 +9,11 @@ extern "C" {
...
-#define IS_PYTHON_3_11_PLUS PY_VERSION_HEX >= 0x030B00C1
-#define IS_PYTHON_3_12_PLUS PY_VERSION_HEX >= 0x030C0000
-#define IS_PYTHON_3_13_PLUS PY_VERSION_HEX >= 0x030D0000
-#define IS_PYTHON_3_14_PLUS PY_VERSION_HEX >= 0x030E0000
-#define IS_PYTHON_3_15_PLUS PY_VERSION_HEX >= 0x030F0000
+#define IS_PYTHON_3_11_PLUS (PY_VERSION_HEX >= 0x030B00C1)
+#define IS_PYTHON_3_12_PLUS (PY_VERSION_HEX >= 0x030C0000)
+#define IS_PYTHON_3_13_PLUS (PY_VERSION_HEX >= 0x030D0000)
+#define IS_PYTHON_3_14_PLUS (PY_VERSION_HEX >= 0x030E0000)
+#define IS_PYTHON_3_15_PLUS (PY_VERSION_HEX >= 0x030F0000)
```
ezyang, line 16: 👍

torch/csrc/utils/pycfunction_helpers.h
```diff
@@ -11,3 +12,15 @@ inline PyCFunction castPyCFunctionWithKeywords(PyCFunctionWithKeywords func) {
   C10_DIAGNOSTIC_POP()
   C10_DIAGNOSTIC_POP()
 }
+
+#if !IS_PYTHON_3_13_PLUS
+using PyCFunctionFast = _PyCFunctionFast;
```
ezyang, line 17: cc @williamwen42

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1130,6 +1140,130 @@ static PyObject* DTensor_OpSchema_post_init(PyObject* mod, PyObject* self) {
...
+    "        Object which describes the mesh topology\n"
+    "        of devices for the DTensor.\n"
+    "    placements (Sequence[:class:`Placement`]]):\n"
+    "        The attribute of the DTensor that describes its layout\n"
+    "        on the mesh topology.\n"
+    "\n"
+    "Return:\n"
+    "    tensor_shape: A List of int which specifies the size of DTensor which build\n"
+    "        on top of the local tensor.\n"
+    "    tensor_stride: A List of int which specifies the stride of DTensor.\n";
```
ezyang, line 1247: How do you feel about unit test coverage on this?

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1130,6 +1140,130 @@ static PyObject* DTensor_OpSchema_post_init(PyObject* mod, PyObject* self) {
...
+    Py_ssize_t nargs) {
+  HANDLE_TH_ERRORS
+  TORCH_CHECK_VALUE(
+      nargs == 3,
+      "compute_global_tensor_info expects 3 arguments, got ",
+      nargs);
+  TORCH_CHECK_TYPE(
+      THPVariable_Check(args[0]),
+      "compute_global_tensor_info 1st argument must be Tensor!");
+  const auto& tensor = THPVariable_Unpack(args[0]);
```
ezyang, line 1261: ah the joy of const reference lifetime extension

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1130,6 +1140,130 @@ static PyObject* DTensor_OpSchema_post_init(PyObject* mod, PyObject* self) {
...
+  // NOTE: if this is a py::handle then this code stops working;
+  // apparently we can't rely on the bound method to stick around.
+  py::object mesh_size;
+  for (const auto& placement : placements) {
+    // TODO: C++ify DeviceMesh somehow; profiling seems
+    // to say that nearly all our remaining time spent is spent
+    // calling back into Python.
+    const auto& cpp_placement = placement.cast<const distributed::Placement&>();
+    if (const auto* cpp_shard =
+            dynamic_cast<const distributed::Shard*>(&cpp_placement)) {
```
ezyang, line 1171: I love the smell of RTTI for breakfast (do we really not have a type tag on these yet?)

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1130,6 +1140,130 @@ static PyObject* DTensor_OpSchema_post_init(PyObject* mod, PyObject* self) {
...
+      }
+      const auto mesh_dim_size = py::cast<int64_t>(mesh_size(idx));
+      tensor_shape[shard_dim] *= mesh_dim_size;
+      for (const auto i : c10::irange(tensor_strides.size())) {
+        if (static_cast<int64_t>(i) != shard_dim &&
+            tensor_strides[i] >= tensor_strides[shard_dim]) {
+          tensor_strides[i] *= mesh_dim_size;
+        }
+      }
+    } else if (!cpp_placement.is_replicate() && !cpp_placement.is_partial()) {
```
ezyang, line 1200: FWIW, I kind of suspect the logic is wrong for testing this way, and this should just be an else, but that'd be semantics changing so best to do it separately.

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1130,6 +1140,130 @@ static PyObject* DTensor_OpSchema_post_init(PyObject* mod, PyObject* self) {
...
+             symint_array_to_list(tensor_shape),
+             symint_array_to_list(tensor_strides))
+      .release()
+      .ptr();
+}
+
+// NOLINTNEXTLINE(modernize-avoid-c-arrays,cppcoreguidelines-avoid-c-arrays)
+static constexpr const char compute_global_tensor_info_doc[] =
+    "Compute the global size and stride of a DTensor from the given local tensor.\n"
+    "The local size is multiplied by `world_size` per Sharding dim.\n"
```
ezyang, line 1226: thanks for fixing the typo lol

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1130,6 +1140,130 @@ static PyObject* DTensor_OpSchema_post_init(PyObject* mod, PyObject* self) {
...
+  Py_ssize_t idx = 0;
+  c10::SymDimVector tensor_shape(
+      tensor.sym_sizes().begin(), tensor.sym_sizes().end());
+  c10::SymDimVector tensor_strides(
+      tensor.sym_strides().begin(), tensor.sym_strides().end());
+  // NOTE: if this is a py::handle then this code stops working;
+  // apparently we can't rely on the bound method to stick around.
+  py::object mesh_size;
+  for (const auto& placement : placements) {
+    // TODO: C++ify DeviceMesh somehow; profiling seems
```
ezyang, line 1166: cc @fduwjj @lw 

In this case, you don't actually need the full mesh: you only need the size which is basically an IntArrayRef.

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1130,6 +1140,130 @@ static PyObject* DTensor_OpSchema_post_init(PyObject* mod, PyObject* self) {
...
+    // to say that nearly all our remaining time spent is spent
+    // calling back into Python.
+    const auto& cpp_placement = placement.cast<const distributed::Placement&>();
+    if (const auto* cpp_shard =
+            dynamic_cast<const distributed::Shard*>(&cpp_placement)) {
+      const auto shard_dim = cpp_shard->dim;
+      TORCH_CHECK(
+          shard_dim >= 0,
+          "Shard placements should have negative dims normalized in the user-facing APIs: ",
+          py::cast<std::string>(py::str(placement)));
```
ezyang, line 1176: this is annoyingly wordy, printing placement directly ought to work

ezyang: I guess the problem is it's a Python object so you can't cuz it would implicitly have to take GIL :(

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1130,6 +1140,130 @@ static PyObject* DTensor_OpSchema_post_init(PyObject* mod, PyObject* self) {
...
+          tensor_ndim,
+          " for placement number ",
+          idx);
+
+      if (!mesh_size) {
+        mesh_size = mesh.attr(dtensor_interned_strings.size);
+      }
+      const auto mesh_dim_size = py::cast<int64_t>(mesh_size(idx));
+      tensor_shape[shard_dim] *= mesh_dim_size;
+      for (const auto i : c10::irange(tensor_strides.size())) {
```
ezyang, line 1194: Can we preserve the comments from the old code? thx

ezyang (approved):

torch/csrc/autograd/python_variable.cpp
swolchok: it's not lifetime extension; unpack returns a reference

test/inductor/test_distributed_patterns.py
swolchok: this test uses ops that that module defines

torch/csrc/autograd/python_variable.cpp
swolchok: Hmm. It's been long enough that I'm not 100% sure what I was thinking, but.I now note that if this was just an  `is_shard()` call and someone has written a Python class that is a subclass of Placement but not the C++ shard class and yet is_shard returns true for it, sadness would ensue

torch/csrc/autograd/python_variable.cpp
swolchok: there appear to be tests in test/distributed/tensor/test_utils.py

swolchok:
failures this time around:
-  [linux-jammy-cuda12.8-py3.10-gcc11 / test (default, 4, 5, linux.g6.4xlarge.experimental.nvidia.gpu)](https://github.com/pytorch/pytorch/actions/runs/18918352725/job/54010750231?pr=162990#logs) looks like it was also segfaulting on main around the time of the base commit: https://github.com/pytorch/pytorch/actions/runs/18862532009/job/53825651259
- the executorch job is marked unstable and seems quite unlikely to be related


as best I can tell we're good to merge with -i.

jithunnair-amd:
@swolchok You might want to use `merge -f` if the ROCm distributed jobs are the only ones pending, since the ROCm distributed jobs running in the latest round are looking for the label `linux.rocm.gpu.mi250.4`, and will probably be queued for a long time.

PR #162764, opened 2025-09-11
fix cpp extension distributed warning spew

msaroufim:
With the new change we only log the warning if we're running non distributed code or if we're in rank 0. Unit testing that certain messages get printed on certain ranks only feels kinda jank so test plan is below instead

Test plan

```python
# torchrun --nproc_per_node=2 demo_fix.py 

import os
import logging

logging.getLogger('torch.utils.cpp_extension').setLevel(logging.DEBUG)

import torch
if 'RANK' in os.environ:
    torch.distributed.init_process_group('nccl')

from torch.utils.cpp_extension import _get_cuda_arch_flags
_get_cuda_arch_flags()

print(f"Rank {os.environ.get('RANK', '0')} done")
```

Logs showing how how `TORCH_CUDA_ARCH_LIST`only shows up once if we explicitly set the the logging level to `logging.DEBUG`. It also improves the debug message to explain what the actual behavior will be

```
(source) [marksaroufim@devgpu005]~% torchrun --nproc_per_node=2 demo_fix.py

W0911 18:30:16.594000 1315439 /home/marksaroufim/pytorch/torch/distributed/run.py:814] 
W0911 18:30:16.594000 1315439 /home/marksaroufim/pytorch/torch/distributed/run.py:814] *****************************************
W0911 18:30:16.594000 1315439 /home/marksaroufim/pytorch/torch/distributed/run.py:814] Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed. 
W0911 18:30:16.594000 1315439 /home/marksaroufim/pytorch/torch/distributed/run.py:814] *****************************************
[rank0]:V0911 18:30:18.921000 1316753 pytorch/torch/utils/cpp_extension.py:2444] TORCH_CUDA_ARCH_LIST is not set, using TORCH_CUDA_ARCH_LIST='10.0+PTX' for visible GPU architectures. Set os.environ['TORCH_CUDA_ARCH_LIST'] to override.
Rank 0 done
Rank 1 done
```

But if we just use the default and comment out `logging.getLogger('torch.utils.cpp_extension').setLevel(logging.DEBUG)`

Then we get

```
(source) [marksaroufim@devgpu005]~% torchrun --nproc_per_node=2 demo_fix.py
W0911 18:14:33.926000 690759 /home/marksaroufim/pytorch/torch/distributed/run.py:814] 
[...]

PR #162999, opened 2025-09-15
[RFC]: No Distributed Log Spew

msaroufim:
## Background
This PR is meant to be an RFC to solve the logging spew problem that often occurs in PyTorch distributed where the same log statement occurs on each rank. Right now the main way we detect these problems is we rely on @stas00 to yell at us and when he does we'll make a patch like so https://github.com/pytorch/pytorch/pull/162764

Not really looking for feedback on code style or linting, will be up front and say Claude did a lot of the heavy lifting here. I'm looking for feedback on whether we think monkey patching logging, warning and print statements is a terrible idea or not. And if it's not a terrible idea should this be on by default or not

The core idea of that patch is to do something along the lines of

```python
if not torch.distributed.is_available() or not torch.distributed.is_initialized() or torch.distributed.get_rank() == 0:
	logger.log_stuff()
```

That way we're guaranteed to log only once on rank 0.

But why do we keep spewing logs by accident? Well because often we can detect log spew within a rank fairly easily because we just run `pytest test.py` and if we get overwhelmed with log statements some good BE citizen will fix it.

However for the most part we don't run all pytorch tests using `torchrun` and so log spew problems that affect distributed often won't get caught since only PTD devs and people training large models will see it

So the solution so this exists in various forms in the code base there are statements such
1. warn_once
2. if conditions on rank 0

But ideally "all" logging statements should be distributed aware and not log spew and historically the best techniques we have to resolve this problem is to rely on code review and/or code linters

 ## This PR

This PR has a different more intrusive take which is that we want all logging statements to be distributed aware, we don't want to use linters and we don't want to send a large volume of small PRs to audit and fix the distributed log spew problem.

In particular when we know that a user is doing distributed stuff i.e when they run `init_process_group` we patch the pytorch logger, python warnings module and optionally print statements.

The core idea is simple and just relies on monkey patching

```python
    if _is_non_rank_zero():
        return
    return orig_func(msg, *args, **kwargs)
return distributed_logging_func
```

If a logging statement is already distributed aware by printing only on rank 0, the above simply no-ops

## Other options

This solves the problem but feels "magical" so some other option is 
1. We don't do it in init_process group but instead do it globally when users enable it and we mention it every time a user complains about log spew
2. We create an authoritative distributed aware logger, send PRs for all the existing log and warn statements in pytorch and force their usage using a new linter in lintrunner

## Test plan
1. I have a unit test but idk how useful it is since testing for low spew across ranks is a bit janky using pytest and torchrun, I can brinig it back
3. demo_no_spew.py is nice to look at and simple
4. I reverted the changes I made for cpp extension log spew to prove the point
5. This PR might have some nasty side effects like changing user loggers, we shouldn't do that for a real solution

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @dcci


stas00:
First, I wanted to say I super appreciate that I'm being listened to, Mark and the pytorch team. And thank you, Mark for working on improving the user experience.

Tristan, would you consider reframing the intermediate option from the perspective of a user?

May I dare to suggest that for each pytorch dev there is 1-10k users? In which case the onus of enabling additional debug noise should be on developers, e.g. set `PYTORCH_INFO_MODE=TRUE` in your `~/.bashrc` and forget it and you will be always notified about everything (I of course invented this env var).

So indeed, there should be an easy way for a developer to activate whatever debug levels they want, but it shouldn't be on by default.

Users are users, they should get minimal noise from the system, IMHO. The current state of software (not just pytorch) is such that most users ignore everything the program spews, because there is so much of the noise, rendering warnings completely useless, even when they are actually important.

Because of that in my experience there should be just 2 operating user-facing modes: 
1. everything is fine -> be quiet
2. something isn't fine -> throw an exception and if a user doesn't want to fix the issue give them a way to override it

everything else is different debug/info/warning messages that 99.9% of the time are only useful for developers.

For example, if a user files an Issue and there are no logs, tell them to enable debug logging to whatever subsystem is malfunctioning and then they will have logs to share.

msaroufim:
> Maybe an intermediate option is to include some special log key for redundant logs so users can opt to filter those if they don't find them useful.

@d4l3k Apologies if I'm not following but the crux of the issue is PyTorch developers might not know a priori what's a redundant log or not. Take the example in the PR of cpp_extensions, I doubt it's something a PTD developer would have caught because the code was not really PTD code

However at least reading @stas00's message it seems the concern isn't too much about whether there exists a mechanism to silence logs but moreso whether this behavior should be on by default or not so one compromise we can explore is having a silencer similar to what's been implemented above but make it opt-in, if enough people enable it after a release or two then make it the default?

wconstab:
Could we solve this another way- with some sort of log aggregation built into torchrun so that we print "ranks 1:8 said foo, rank 0 said bar"? 

I am wary of trying to codify the practice of developers not logging stuff from nonzero ranks (not sure that's what's being suggested exactly), though, we do already support suppressing logs form specific ranks from the torchrun side (see how torchtitan configures its logging in train.sh - only rank 0 logs are printed by default)

msaroufim:
I think moving the responsibility to torchrun would work. I like your proposal on collapsing identical logs in a single print statement. That said we'd need to encourage people not using torchrun to implement something similar. I believe Stas uses Deepspeed so this wouldn't totally solve his specific problem

stas00:
I use both `torchrun` and `deepspeed` launchers.

As long as non-`torchrun` launchers can do the same as `torchrun` it can be implemented and is a viable solution.

ezyang:
Some things I believe.

1. A benign torch warning showing up world size times on the default aggregated log view of a distributed run is clearly bad experience. It is bad for a single node log, and it is extra bad if you're aggregating across all nodes. At the point when you are actually doing a distributed run, you probably don't care about deprecation warnings anymore (they should have been fixed when you were doing single-node or ideally single process debugging of your model.)
2. Dumping the exceptions to log for all ranks that failed is generally bad experience, since it's typical that every rank failed and the resulting jumble is unreadable. However, this information is *critical* for diagnosing non-SPMD problems, as a difference in stack trace in one node from others is going to be typically your first clue that ranks diverged.
3. Many users separately have a cloud logging service to ingest logs from their training logs. Typical affordances for inspecting logs there is ability to filter logs to a single rank and do searches. It is also common to have a "response file" for a rank to report what its final result is. You can build additional tooling on top to show just the backtrace at the thing that failed, or to analyze logs across ranks and give you information about which one diverged. Doing these things by default is an important part of UX.
4. In a "torchrun" style modality, we only have control over the CLI output. Furthermore, in the default usage of torchrun (without any arguments) you are generally expected to give all the information and it would be BC breaking to change this.

I tweeted about this PR and someone brought up the interesting point that, maybe the problem is that most people *shouldn't* be using raw torchrun. Instead, there is a missing niche for a blessed, higher level launcher (perhaps using Monarch) which has the features that you normally see in (3) in a nice way, by default. WDYT?

ppwwyyxx:
Detectron2's logger configuration: https://github.com/facebookresearch/detectron2/blob/a1ce2f956a1d2212ad672e3c47d53405c2fe4312/detectron2/utils/logger.py#L82-L111

* All ranks log everything to help debugging - but to individual log files
* Only rank0 logs to the screen to improve day to day experience
  * In an improved private version we let all ranks log to the screen for level>ERROR logs.

Only the end application shall decide where logs go, not any individual library. So it might makes sense for torchrun (not torch itself) to support something along this line. Non-torchrun cases would have to configure this on their own, but it's just a few lines of code anyway, so we're actually happy with the status quo (we don't use torchrun either). Not able to redirect C++ logs / python intepreter stack traces is the only thing I don't like about the above solution.

Builtin aggregation is an improvement over repeating, but in my experience it may not be a huge improvement. Logs across ranks can be technically different, hence cannot be aggregated, but semantically equal. For example during training every rank would print lines of ` iter: 1  loss: 0.355`, or 'time spent on XX: 5 seconds` that are slightly different per rank, still creating a log spew.

ngimel:
fwiw ray ignores numbers when deduplication logs https://docs.ray.io/en/latest/ray-observability/user-guides/configure-logging.html#log-deduplication, that would take care of deduping loss value etc.

PR #163001, opened 2025-09-15
[CUDA13][CUDA Graphs] Fix `test_graph_external_wait_and_record` to account for minimum supported compute-capability

eqy:
CUDA 13 requires SM 7.5

cc @ptrblck @msaroufim @jerryzh168 @tinglvv @nWEIdia @mcarilli @ezyang @eellison @penguinwu @BoyuanFeng

### DIFF
--- a/aten/src/ATen/native/cuda/Reduce.cuh
+++ b/aten/src/ATen/native/cuda/Reduce.cuh
@@ -416,6 +416,7 @@ struct ReduceOp {
     if (config.should_block_y_reduce()) {
       value = block_y_reduce<output_vec_size>(value, shared_memory);
     }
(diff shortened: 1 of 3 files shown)


Skylion007 (approved):

aten/src/ATen/native/cuda/Reduce.cuh
```diff
@@ -416,6 +416,7 @@ struct ReduceOp {
     if (config.should_block_y_reduce()) {
       value = block_y_reduce<output_vec_size>(value, shared_memory);
     }
+    __syncthreads();
```
Skylion007: Why this?

PR #163014, opened 2025-09-15
[ROCm][CI] update _rocm-test.yml based on _linux-test.yml

jeffdaily:
Fixes missing huggingface secrets and aligns _rocm-test.yml with other updates from _linux-test.yml that it was initially based on.

cc @sunway513 @jithunnair-amd @pruthvistony @ROCmSupport @dllehr-amd @jataylo @hongxiayang @naromero77amd @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela

### DIFF
--- a/.github/workflows/_rocm-test.yml
+++ b/.github/workflows/_rocm-test.yml
@@ -62,6 +62,15 @@ on:
         required: false
         type: number
         default: 1
+    secrets:
+      HUGGING_FACE_HUB_TOKEN:
+        required: false
+        description: |
+          HF Auth token to avoid rate limits when downloading models or datasets from hub
+      VLLM_TEST_HUGGING_FACE_TOKEN:
+        required: false
+        description: |
+          HF Auth token to test vllm
 env:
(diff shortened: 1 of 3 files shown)


.github/workflows/_rocm-test.yml
```diff
@@ -62,6 +62,15 @@ on:
         required: false
         type: number
         default: 1
+    secrets:
+      HUGGING_FACE_HUB_TOKEN:
+        required: false
+        description: |
+          HF Auth token to avoid rate limits when downloading models or datasets from hub
+      VLLM_TEST_HUGGING_FACE_TOKEN:
```
huydhn: You don't need `VLLM_TEST_HUGGING_FACE_TOKEN` here, maybe this can be needed later once we run vLLM tests on ROCm runners

huydhn (approved):
LGTM! Just need to remove `VLLM_TEST_HUGGING_FACE_TOKEN` before landing as we don't need it here atm

PR #160532, opened 2025-08-13
Return NoOpDeviceGuardImpl in replace of CudaDeviceGuard when device is not available, or cpu-only build

SherlockNoMad:
Summary: 

To support exporting a cuda model on a CPU-only machine under fake tensor mode. 
User commonly need to move sample inputs to the cuda device with .to("cuda:0") or .to("cuda") call. 
This diff supports this.
I expect the following pattern to work
```
with FakeTensorMode(allow_non_fake_inputs=True):
    cuda_module = module.to("cuda:0")
    cuda_sample_inputs = tuple([x.to("cuda:0") for x in sample_inputs])
    with torch.no_grad():
        ep = torch.export.export(cuda_module, cuda_sample_inputs)
```

Test Plan:
CI

Rollback Plan:

PR #163016, opened 2025-09-15
[Reland] Return NoOpDeviceGuardImpl in replace of CudaDeviceGuard when device is not available, or cpu-only build

SherlockNoMad:
Reland of #160532

Summary: 

To support exporting a cuda model on a CPU-only machine under fake tensor mode. 
User commonly need to move sample inputs to the cuda device with .to("cuda:0") or .to("cuda") call. 
This diff supports this.
I expect the following pattern to work
```
with FakeTensorMode(allow_non_fake_inputs=True):
    cuda_module = module.to("cuda:0")
    cuda_sample_inputs = tuple([x.to("cuda:0") for x in sample_inputs])
    with torch.no_grad():
        ep = torch.export.export(cuda_module, cuda_sample_inputs)
```


huydhn (approved):
Stamped to unblock!  Let's see if the issue with distributed CPU job is showing up on PR

PR #163017, opened 2025-09-15
[Bugfix] Match eager stride semantics for cloned tensors with preserve_format in compile

Lucaskabela:
Fixes #161010 by making `clone_meta` match the semantics of strides for eager mode.  

This is: 
  * Case 1: Tensor is_non_overlapping_and_dense; in this case, stride should match input tensor stride
  * Case 2: Otherwise, stride should be contiguous computed from input tensor using `compute_elementwise_output_strides`

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames

### DIFF
--- a/torch/_prims/__init__.py
+++ b/torch/_prims/__init__.py
@@ -692,16 +692,22 @@ def _clone_meta(
             device=input.device,
             memory_format=memory_format,
         )
+    else:
+        # Match eager behavior by preserving strides for non_overlapping_and_dense tensors
+        # If not, eager clone creates contiguous strides
+        computed_stride = None
+        if utils.is_non_overlapping_and_dense(input):
+            computed_stride = input.stride()
+        else:
+            computed_stride = utils.compute_elementwise_output_strides(input)
 
-    # memory_format == torch.preserve_format
-    strides = utils.compute_elementwise_output_strides(input)
-    return torch.empty_strided(
-        input.shape,
-        strides,
-        dtype=input.dtype,
-        layout=input.layout,
-        device=input.device,
-    )
+        return torch.empty_strided(
+            input.shape,
+            computed_stride,
+            dtype=input.dtype,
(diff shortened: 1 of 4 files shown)


williamwen42 (approved):

xmfan (approved):

torch/_prims/__init__.py
```diff
@@ -692,16 +692,22 @@ def _clone_meta(
...
             memory_format=memory_format,
         )
+    else:
+        # Match eager behavior by preserving strides for non_overlapping_and_dense tensors
+        # If not, eager clone creates contiguous strides
+        computed_stride = None
+        if utils.is_non_overlapping_and_dense(input):
+            computed_stride = input.stride()
+        else:
+            computed_stride = utils.compute_elementwise_output_strides(input)
```
zou3519, line 702: I think compute_elementwise_output_strides is wrong in general. We should update it to do: 
```
if utils.is_non_overlapping_and_dense(input):
   computed_stride = input.stride()
 else:
    something()           
```
So while you should feel free to ship this PR, there are still problems lurking in whatever relies on `compute_elementwise_output_strides` (e.g. the other unary pointwise ops).

In the other PR we did try to update compute_elementwise_output_strides to do the above and it led to some issues that were not yet debugged.

Lucaskabela: cc @morrison-turnansky ah let's follow up with this - these test cases in this PR pass, and I am not aware of others we are missing, but let's think on this and if we find anything we can work together to put out a follow up PR fixing any issues in compute_elementwise_output_strides

morrison-turnansky: @Lucaskabela sounds like a plan

PR #163025, opened 2025-09-16
[CI] disable rerun of distributed tests

kwen2501:
* __->__ #163025

#162978 identified an issue that distributed test failures were wrongly muted. 
Per discussion with @malfet, one solution is to disable rerun of distributed tests in `run_test.py`.
The PR makes use of the `is_distributed_test` flag to identify those tests.


malfet (approved):
Sure, so there will be no more flaky failures detection, but I guess distributed team will just live with that

PR #163030, opened 2025-09-16
C++-accessible Placements via pybind11

swolchok:
* #163667
* #162990
* #164414
* __->__ #163030

This makes Placement data representation available in C++ via pybind11.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @kadeng @chauhang @amjames @Lucaskabela @jataylo @ezyang @chenyang78


torch/csrc/autograd/python_variable.cpp
```diff
@@ -2886,6 +2888,119 @@ static void initTensorImplConversion(PyObject* module) {
 }
 } // namespace torch::autograd
 
+// XXX: Move this somewhere
```
albanD: You can create a torch/csrc/distributed folder where we can host all the distributed related objects you're moving to c++.
The convention here is foo.cpp contains the c++ only code and python_foo.cpp contain the python-aware parts.

torch/csrc/autograd/python_variable.cpp
```diff
@@ -2886,6 +2888,119 @@ static void initTensorImplConversion(PyObject* module) {
...
+static const char placement_class_docstring[] =
```
Skylion007: Should also be constexpr

torch/csrc/autograd/python_variable.cpp
```diff
@@ -2886,6 +2888,119 @@ static void initTensorImplConversion(PyObject* module) {
...
+  }
+
+  bool operator!=(const Shard& rhs) const {
+    return !operator==(rhs);
+  }
+};
+
+class Replicate : public Placement {
+ public:
+  ~Replicate() {}
```
Skylion007: Should be equals default, right?

torch/csrc/autograd/python_variable.cpp
```diff
@@ -2886,6 +2888,119 @@ static void initTensorImplConversion(PyObject* module) {
...
+class Partial : public Placement {
+ public:
+  std::string reduce_op;
+
+  Partial() : Partial("sum") {}
+
+  explicit Partial(std::string reduce_op_) : reduce_op(std::move(reduce_op_)) {}
+
+  ~Partial() {}
+
```
Skylion007: Likewise here

torch/csrc/autograd/python_variable.cpp
```diff
@@ -2886,6 +2888,119 @@ static void initTensorImplConversion(PyObject* module) {
...
+  virtual bool is_replicate() const {
+    return false;
+  }
+
+  virtual bool is_partial(std::optional<std::string_view> reduce_op) const {
+    return false;
+  }
+};
+
+Placement::~Placement() {}
```
Skylion007: Shouldn't this be equals default too? Why is it out of line?

swolchok: > Why is it out of line?

It's pure virtual, but still needs an implementation. Can't do both inline.

torch/distributed/tensor/placement_types.py
```diff
@@ -20,29 +21,11 @@
 __all__ = ["Placement", "Shard", "Replicate", "Partial"]
```
ezyang, line 21: BTW there's another placement MaskPartial which should be moved over here

torch/_dynamo/variables/distributed.py
```diff
@@ -142,7 +142,7 @@ def is_placement_type(value):
 
         from torch.distributed.tensor.placement_types import Placement
 
-        return type(value) is type and issubclass(value, Placement)
+        return isinstance(value, type) and issubclass(value, Placement)
```
ezyang, line 145: lol

torch/csrc/distributed/python_placement.cpp
```diff
@@ -0,0 +1,89 @@
...
+      .def(py::init<int64_t>(), py::arg("dim"))
+      .def_readonly("dim", &Shard::dim)
+      .def("is_shard", &Shard::is_shard, py::arg("dim") = py::none())
+      .def(
+          "__eq__",
+          [](const Shard& lhs, const Shard& rhs) { return lhs == rhs; },
+          py::is_operator())
+      .def(py::pickle(
+          [](const Shard& shard) { return shard.dim; },
+          [](std::int64_t dim) { return Shard(dim); }));
```
ezyang: Ah, this one is potentially trouble. Are you changing the Pickle format here?

ezyang (approved):

torch/_dynamo/variables/distributed.py
albanD: This is potentially significantly slower. You're ok with that?

swolchok: `type` is the class of Python classes (i.e., their metaclass). the old check will not work because pybind11 classes have their own metaclass.

torch/csrc/distributed/python_placement.cpp
swolchok: yes. since you're asking, I guess that's not allowed? I only had to touch it because copy.deepcopy pickles

torch/distributed/tensor/placement_types.py
swolchok: yeah, I suffered a lot of pain discovering that it existed thanks to unit test failures :(

torch/csrc/distributed/python_placement.cpp
swolchok: IIUC the format for dataclasses is just "a dict with keys matching the class's field names", so I can change to that without much trouble

torch/csrc/distributed/python_placement.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+
+#include <Python.h>
```
malfet: Do you mind just forward declaring the struct here?

torch/csrc/distributed/python_placement.cpp
```diff
@@ -0,0 +1,98 @@
...
+#include <pybind11/pybind11.h>
+#include <torch/csrc/distributed/Placement.h>
+#include <torch/csrc/utils/pybind.h>
+
+using namespace pybind11::literals;
+
+namespace torch::distributed {
+namespace {
+// NOLINTNEXTLINE(modernize-avoid-c-arrays,cppcoreguidelines-avoid-c-arrays)
+const char placement_class_docstring[] =
```
malfet: Aren't that the usual way of defining string literals (or may be even do `const auto` and also gets rid of the suppression
```suggestion
const char *placement_class_docstring =
```

torch/csrc/distributed/python_placement.cpp
```diff
@@ -0,0 +1,98 @@
...
+    "The base class for the Placement type, where it describes how a DTensor is placed onto the\n"
+    "``DeviceMesh``. ``Placement`` and ``DeviceMesh`` together could describe the DTensor Layout.\n"
+    "It is the base class of the three main DTensor Placement types: ``Shard``, ``Replicate``,\n"
+    "and ``Partial``.\n"
+    "\n"
+    "This class is not meant to be used directly, mainly served as a typing stub.\n";
```
malfet: Multi-line string linterals are supported since C++11 
```suggestion
    R""""The base class for the Placement type, where it describes how a DTensor is placed onto the
   ``DeviceMesh``. ``Placement`` and ``DeviceMesh`` together could describe the DTensor Layout.
   It is the base class of the three main DTensor Placement types: ``Shard``, ``Replicate``,
    and ``Partial``.
    \n
    "This class is not meant to be used directly, mainly served as a typing stub.
""";
```

torch/csrc/distributed/python_placement.cpp
swolchok: I was going to say "this definition gets you an extra pointer; the array definition leaves that out", but TIL if you make the pointer const that's not true. https://godbolt.org/z/n4a493xvP (`const auto p = "str"` seems to mean `const char* const p = "str", FWIW)

torch/csrc/distributed/python_placement.h
swolchok: I can try `typedef struct _object PyObject;` and see if that passes CI. wasn't clear to me how to forward-declare it before

swolchok: done -- we have a header that takes care of this

PR #163031, opened 2025-09-16
Use C++-accessible Placement in compute_global_tensor_info

swolchok:
* __->__ #163031
* #162990
* #163030
* #162968
* #162508
* #161695

This seems to speed things up, though there is more overhead than I'd like to unwrap a Placement via pybind11. Might be time to introduce nanobind; I'm not sure that we can use it in general, but the Placement type system is (currently?) isolated. (Alternatively, I could look into whether we can cut this overhead.)


swolchok:
folding into previous PR https://github.com/pytorch/pytorch/pull/162990

PR #163034, opened 2025-09-16
Add device argument to torch.random.get_rng_state

### LINKED ISSUE #162812: Make torch.random.get_rng_state take a device argument
### 🐛 Describe the bug

The device argument would accept any device type and dispatch to the correct backend get_rng_state.

This API was hallucinated by ChatGPT:

Image

It seems very reasonable to me. Thoughts?

### Versions

main

cc @albanD

evan-conway:
Fixes #162812

Adds support for either passing a device directly into get_rng_state, or passing in a string or int (which is then wrapped into a device inside, as in torch.cuda.get_rng_state).

I wasn't exactly sure where tests for this should go, please let me know. I used this script for testing:
```python
import torch

# note: when running with CUDA GPU, first three tests will give the same result,
# as will the last two

# test with no device specified
print(torch.get_rng_state())

# test with CPU
cpu_device = torch.device("cpu")
print(torch.get_rng_state(cpu_device))

# test with direct name
print(torch.get_rng_state("cpu"))

# test with CUDA
cuda_device = torch.device("cuda:0")
print(torch.get_rng_state(cuda_device))

# test with integer
print(torch.get_rng_state(0))
```

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -561,6 +561,43 @@ class TestCheckpoint(TestCase):
         self.assertTrue("first device type: cuda" in warning_msg)
 
 
+class TestRNGState(TestCase):
+    def test_get_rng_state_cpu(self):
+        # get state from get_rng_state with no arguments
+        state_no_args = torch.get_rng_state()
+
+        # get state from get_rng_state with device passed in
+        device = torch.device("cpu")
+        state_device_arg = torch.get_rng_state(device)
+
+        # get state from get_rng_state with string passed in
+        state_string_arg = torch.get_rng_state("cpu")
+
+        # assert that all three are equal
+        self.assertEqual(state_no_args, state_device_arg)
+        self.assertEqual(state_no_args, state_string_arg)
+
+    @unittest.skipIf(not HAS_CUDA, "No CUDA")
+    def test_get_rng_state_cuda(self):
+        # get state from cuda's get_rng_state
+        state_no_args = torch.cuda.get_rng_state()
+
+        # get state from get_rng_state with device passed in
+        device = torch.device("cuda")
+        state_device_arg = torch.get_rng_state(device)
+
+        # get state from get_rng_state with string passed in
+        state_string_arg = torch.get_rng_state("cuda")
+
+        # get state from get_rng_state with integer passed in
+        state_int_arg = torch.get_rng_state(0)
+
+        # assert that all four are equal
+        self.assertEqual(state_no_args, state_device_arg)
+        self.assertEqual(state_no_args, state_string_arg)
+        self.assertEqual(state_no_args, state_int_arg)
+
+
 class TestDataLoaderUtils(TestCase):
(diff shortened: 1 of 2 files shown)


ezyang:
Looks good. Put the tests in test_utils.py I guess.

Shoumik-Gandre:
Hey, may I have edit access to this PR?
I would like to add these tests!

test/test_utils.py
```diff
@@ -313,7 +313,7 @@ def test_checkpoint_rng_cuda(self):
             def run_fn(input):
                 return phase2(input)
 
-            state = torch.cuda.get_rng_state()
```
hyunchel, line 315: wouldn't this simply remove a part of test for `torch.cuda.get_rng_state` function?

ezyang: Yes, a dedicated test would be better

evan-conway: I wasn't sure if it was worth setting up a dedicated test since this pull request is fairly minor. I'll add a dedicated test sometime soon.

evan-conway: Should be resolved now, let me know if anything needs to be changed.

ezyang (dismissed):

ezyang:
test failures real

torch/random.py
```diff
@@ -19,14 +20,25 @@ def set_rng_state(new_state: torch.Tensor) -> None:
     default_generator.set_state(new_state)
 
 
-def get_rng_state() -> torch.Tensor:
+def get_rng_state(device: Union[int, str, torch.device] = "cpu") -> torch.Tensor:
```
cyyever: `Union[int, str, torch.device]` can be replaced by `torch.Device`, hope that this change can make CI green.

cyyever (dismissed):

cyyever:
@evan-conway See my comment.

torch/random.py
evan-conway: Sure, will do.

evan-conway:
Okay, as per @cyyever's suggestion I removed support for just passing in a string or int, so the Union isn't necessary. Hopefully that fixes the issue.

torch/random.py
```diff
@@ -19,14 +19,20 @@ def set_rng_state(new_state: torch.Tensor) -> None:
...
-    .. note:: The returned state is for the default generator on CPU only.
+    Args:
+        device (torch.device, optional): The device to return the RNG state of.
+            Default: ``torch.device('cpu')``, the current CPU device.
 
     See also: :func:`torch.random.fork_rng`.
     """
-    return default_generator.get_state()
+
+    if device.type == "cpu":
```
cyyever, line 32: The passed device can be `str` or `int`, therefore this line would crash.

cyyever (changes requested):

cyyever (dismissed):

ezyang (approved):

PR #163036, opened 2025-09-16
Fix invalid indices bug for max_unpool2d/3d on MPS

### LINKED ISSUE #163035: [MPS] max_unpool2d/max_unpool3d did not raise an error for invaid indices
### 🐛 Describe the bug

```python
import torch

input = torch.randn(1, 1, 2, 2)
indices = torch.tensor([[[[10000, 10], [0, 2]]]], dtype=torch.int64)
output_size = (2, 2)

try:
    out_cpu = F.max_unpool2d(input, indices, output_size)
    print("CPU:", out_cpu)
except Exception as e:
    print("CPU:", e)  # RuntimeError: Found an invalid max index: 10000 (output volumes are of size 4x4

# MPS
input = input.to("mps")
indices = indices.to("mps")
try:
    out_mps = F.max_unpool2d(input, indices, output_size)
    print("MPS:", out_mps)  # return a tensor here
except Exception as e:
    print("MPS:", e)


input = torch.randn(1, 1, 2, 2, 2)

indices = torch.tensor(
    [[[
        [[10000, 10], [0, 2]],
        [[1, 3], [4, 5]]
    ]]], dtype=torch.int64
)
output_size = (4, 4, 4)

# CPU
try:
    out_cpu = F.max_unpool3d(input, indices, output_size)
    print("CPU:", out_cpu)  # RuntimeError: Found an invalid max index: 10000 (output volumes are of size 4x4x4
except Exception as e:
    print("CPU:", e)

# MPS
if torch.backends.mps.is_available():
    input_mps = input.to("mps")
    indices_mps = indices.to("mps")
    try:
        out_mps = F.max_unpool3d(input_mps, indices_mps, output_size)
        print("MPS:", out_mps)  # return a tensor here
    except Exception as e:
        print("MPS:", e)
```

* Invalid `idx` could be either larger than output size or negative;
* The invalid `idx` would raise an error on CPU and CUDA, but not on MPS.

### Versions

nightly built

can-gaa-hou:
Fixes #163035

### DIFF
--- a/aten/src/ATen/native/mps/operations/Pooling.mm
+++ b/aten/src/ATen/native/mps/operations/Pooling.mm
@@ -535,37 +535,14 @@ static void max_unpool_out_mps_template(const Tensor& input,
   output.fill_(0);
 
   if (indices.defined() && indices.numel() > 0) {
-    int64_t channels = 0;
-    [[maybe_unused]] int64_t output_depth = 1;
-    [[maybe_unused]] int64_t output_height = 1;
-    [[maybe_unused]] int64_t output_width = 1;
-
-    if (pooling_dims == 3) {
-      channels = input.ndimension() == 4 ? input.size(0) : input.size(0) * input.size(1);
-      output_depth = output.size(-3);
-      output_height = output.size(-2);
-      output_width = output.size(-1);
-    } else {
-      channels = input.ndimension() == 3 ? input.size(0) : input.size(0) * input.size(1);
-      output_height = output.size(-2);
-      output_width = output.size(-1);
-    }
-
-    int64_t output_image_size = 1;
-    for (int dim = leading_dims; dim < dims; ++dim) {
-      output_image_size *= output.size(dim);
-    }
+    auto output_image_size = c10::multiply_integers(output_size_);
 
     int64_t min_idx = indices.min().item<int64_t>();
     int64_t max_idx = indices.max().item<int64_t>();
 
     if (min_idx < 0 || max_idx >= output_image_size) {
       int64_t error_idx = (min_idx < 0) ? min_idx : max_idx;
-        TORCH_CHECK(false,
-                    "Found an invalid max index: ",
-                    error_idx,
-                    "for output tensor of shape ", output.shape());
-      }
+      TORCH_CHECK(false, "Found an invalid max index: ", error_idx, " for output tensor of shape ", output_size_);
     }
   }
 
--- a/aten/src/ATen/native/mps/operations/Pooling.mm
+++ b/aten/src/ATen/native/mps/operations/Pooling.mm
@@ -534,6 +534,57 @@ static void max_unpool_out_mps_template(const Tensor& input,
   output.resize_(output_size, memory_format);
   output.fill_(0);
 
+  if (indices.defined() && indices.numel() > 0) {
+    int64_t channels = 0;
+    [[maybe_unused]] int64_t output_depth = 1;
+    [[maybe_unused]] int64_t output_height = 1;
+    [[maybe_unused]] int64_t output_width = 1;
+
+    if (pooling_dims == 3) {
+      channels = input.ndimension() == 4 ? input.size(0) : input.size(0) * input.size(1);
+      output_depth = output.size(-3);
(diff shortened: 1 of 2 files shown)


kulinseth (approved):

aten/src/ATen/native/mps/operations/Pooling.mm
```diff
@@ -534,6 +534,57 @@ static void max_unpool_out_mps_template(const Tensor& input,
   output.resize_(output_size, memory_format);
   output.fill_(0);
 
+  if (indices.defined() && indices.numel() > 0) {
+    int64_t channels = 0;
+    [[maybe_unused]] int64_t output_depth = 1;
```
Skylion007: Why all maybe unused? Because of TORCH_CHECK?

test/nn/test_pooling.py
```diff
@@ -1386,6 +1387,34 @@ def test_max_pool2d_with_indices_backward_fails(self, device):
                 indices,
             )
 
+    @onlyMPS
```
malfet: Why onlyMPS? Same checks should hold true for CPU and CUDA, wouldn't they?

aten/src/ATen/native/mps/operations/Pooling.mm
```diff
@@ -534,6 +534,57 @@ static void max_unpool_out_mps_template(const Tensor& input,
...
+    } else {
+      channels = input.ndimension() == 3 ? input.size(0) : input.size(0) * input.size(1);
+      output_height = output.size(-2);
+      output_width = output.size(-1);
+    }
+
+    int64_t output_image_size = 1;
+    for (int dim = leading_dims; dim < dims; ++dim) {
+      output_image_size *= output.size(dim);
+    }
```
malfet: Use `c10::multiply_integers` primitive, which also throws an exception on overflow

aten/src/ATen/native/mps/operations/Pooling.mm
```diff
@@ -534,6 +534,57 @@ static void max_unpool_out_mps_template(const Tensor& input,
...
+      } else {
+        TORCH_CHECK(false,
+                    "Found an invalid max index: ",
+                    error_idx,
+                    " (output volumes are of size ",
+                    output_height,
+                    "x",
+                    output_width,
+                    ")");
+      }
```
malfet: Simplest solution are usually the best: what's wrong with printing the entire shape?
```suggestion
        TORCH_CHECK(false,
                    "Found an invalid max index: ",
                    error_idx,
                    "for output tensor of shape ", output.shape());
      }
```

malfet (changes requested):
Thank you for working on the change, but I'm a bit suspicious why this is tested only for MPS?
Also, use simpler error checking message and print entire shape rather than having separate output for 2d and 3d cases, which would also eliminate the need for may-be unsued

aten/src/ATen/native/mps/operations/Pooling.mm
can-gaa-hou: I get your point here :) I was trying to keep the same error log as the CPU one. It would be much easier in this way. Thanks for the review.

test/nn/test_pooling.py
can-gaa-hou: Ah hah, I thought it would break the entire test process on CUDA since it checks invalid indices in its kernel. But in fact, `self.assertRaisesRegex` would generate a subprocess, and only the context in the subprocess would be destroyed, which would not break the entire testcase. It should be tested on all of these backends.

aten/src/ATen/native/mps/operations/Pooling.mm
can-gaa-hou: Already removed :)

malfet (approved):
Looks good to me, code looks much neater now

can-gaa-hou:
Sorry about that. The check introduced a sync point on MPS, and the test only validated CPU behavior. I’ll be more careful about host-side reads/reductions on MPS paths and make sure tests cover the intended device next time. Thanks for catching and fixing this.

cc @malfet

PR #163042, opened 2025-09-16
[docs] Add cross-reference to torch.nn.ReLU from Softplus (and LeakyReLU “See also”)

Username46786:
### What
- Convert plain "ReLU" in the Softplus docstring to a Sphinx class reference so it links to `torch.nn.ReLU`.
- Add a short "See also" line in LeakyReLU pointing to `torch.nn.ReLU`.

### Why
Improves navigation in the activation docs by making related functions clickable and easier to discover.

### How
- Edited `torch/nn/modules/activation.py`:
  - Softplus docstring: use `:class:`~torch.nn.ReLU``.
  - LeakyReLU docstring: add "See also :class:`~torch.nn.ReLU`".

### Notes
Docs-only change. Happy to tweak wording/placement if preferred.

### DIFF
--- a/torch/nn/modules/activation.py
+++ b/torch/nn/modules/activation.py
@@ -885,7 +885,10 @@ class LeakyReLU(Module):
         \begin{cases}
         x, & \text{ if } x \geq 0 \\
         \text{negative\_slope} \times x, & \text{ otherwise }
-        \end{cases}
+        \end{cases} 
+
+        See also :class:`~torch.nn.ReLU` for the non-leaky variant.
+
 
     Args:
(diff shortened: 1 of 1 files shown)


torch/nn/modules/activation.py
```diff
@@ -961,8 +964,8 @@ class Softplus(Module):
     .. math::
         \text{Softplus}(x) = \frac{1}{\beta} * \log(1 + \exp(\beta * x))
 
-    SoftPlus is a smooth approximation to the ReLU function and can be used
-    to constrain the output of a machine to always be positive.
```
albanD, line 965: You mistakenly deleted this line I think

PR #163044, opened 2025-09-16
[AOTI-FX] Solve for undefined symbols in dynamic input shapes

blaine-rister:
# Problem
When dynamic shapes are passed to AOTInductor, they usually have a very basic form like `(s0, 5, 27)`. In these cases it's straightforward to generate code defining the symbol `s0` as a specific dimension of the input tensor. However, AOTI can handle slightly more generic expressions than this, such as `(2 * s0, 5, 27)`. In these cases, we don't immediately know the value of `s0`, but we need to solve for it, since it may be referenced in other parts of the program such as kernel call arguments, launch grids, etc.

# Feature
This PR adds support for more generic dynamic input expressions in the FX backend, following the implementation already present in AOTI's C++ backend:
 1. Check if the expression contains *one* undefined symbol, as multiple variables would make the equation underdetermined. Let's call this `s0`. (We could potentially generalize this, but this PR focuses on cases AOTI can already handle.)
 2. Generate a new symbol for the relevant size or stride of the input tensor. Let's call this `size`. This is computed with FX nodes just as a normal symbol would be.
 3. Use sympy to solve for `s0` in terms of `size`. Let's call the resulting expression `solution`.
 4. Since we know `s0` is an integer, `solution == floor(solution)`. Take the floor and then convert division to `FloorDiv`. This is required to trace through the expression, since the return value of regular division is not guaranteed to be an integer.
 5. Generate FX for the modified `solution`, which defines the value `s0`.
 6. Override the relevant method of `PythonWrapperCodegen` to a no-op, since the FX converter handles the above on its own.

# Test plan
In addition to the existing dynamic shapes tests, this PR adds new test cases where the input shape contains a non-trivial expression. This dynamic input dimension is then multiplied by other dimensions to form the argument to a `reshape`.

Here's an example graph from one of the CI tests. In this case, the input expression was `2*x + 1`, and the solution is `x = (sym_size_int - 1) / 2`:
```
graph():
    %arg0_1 : [num_users=2] = placeholder[target=arg0_1]
    %sym_size_int : [num_users=1] = call_function[target=torch.ops.aten.sym_size.int](args = (%arg0_1, 0), kwargs = {})
    %sym_sum : [num_users=1] = call_function[target=torch.sym_sum](args = ([-1, %sym_size_int],), kwargs = {})
    %floordiv : [num_users=1] = call_function[target=operator.floordiv](args = (%sym_sum, 2), kwargs = {})
    %mul : [num_users=2] = call_function[target=operator.mul](args = (8, %floordiv), kwargs = {})
    %sym_sum_1 : [num_users=2] = call_function[target=torch.sym_sum](args = ([4, %mul],), kwargs = {})
    %buf0 : [num_users=2] = call_function[target=torch.empty_strided](args = ([%sym_sum_1], [1]), kwargs = {dtype: torch.float32, device: cuda:0})
    %sym_sum_2 : [num_users=1] = call_function[target=torch.sym_sum](args = ([35, %mul],), kwargs = {})
    %floordiv_1 : [num_users=1] = call_function[target=operator.floordiv](args = (%sym_sum_2, 32), kwargs = {})
    %triton_kernel_wrapper_mutation : [num_users=0] = call_function[target=torch.ops.higher_order.triton_kernel_wrapper_mutation](args = (), kwargs = {kernel_idx: 0, constant_args_idx: 0, grid: [(%floordiv_1, 1, 1)], tma_descriptor_metadata: {}, kwargs: {in_ptr0: %arg0_1, out_ptr0: %buf0, xnumel: %sym_sum_1, XBLOCK: 32}})
    return buf0
```

The `sym_size_int` node returns the first dimension of the input tensor. Next, `floordiv` computes the input symbol in terms of the input size. Then, the launch grid is computed by `floordiv_1`, the kernel argument by `sym_sum_1`.

### DIFF
--- a/test/inductor/test_fxir_backend.py
+++ b/test/inductor/test_fxir_backend.py
@@ -727,6 +727,7 @@ class FxirTestCase(InductorTestCase):
         self._compile_and_check(foo, args, expected_num_triton_kernels=0)
 
 
+@instantiate_parametrized_tests
 class AOTFxirTestCase(InductorTestCase):
     device = GPU_TYPE
 
@@ -866,7 +867,14 @@ class AOTFxirTestCase(InductorTestCase):
         # Now the backend should have been called.
         self.assertTrue(called)
 
-    def test_dynamic_input_expr(self):
+    @parametrize(
+        "expr",
+        [
+            (2 * Dim("x") + 1),
+            (Dim("x", min=3) - 3),
+        ],
+    )
+    def test_dynamic_input_expr(self, expr: sympy.Expr):
         """
         Test dynamic shapes with a nontrivial input expression.
         """
@@ -875,7 +883,7 @@ class AOTFxirTestCase(InductorTestCase):
             def forward(self, x):
                 return x.reshape(x.shape[0] * x.shape[1]) + x.shape[1]
 
-        dynamic_shapes = {"x": {0: 2 * Dim("x", min=1) + 1}}
+        dynamic_shapes = {"x": {0: expr}}
         inp = (torch.randn((5, 4), device=self.device),)
         self.check(M().to(self.device), inp, dynamic_shapes=dynamic_shapes)
 
--- a/test/inductor/test_fxir_backend.py
+++ b/test/inductor/test_fxir_backend.py
@@ -866,6 +866,19 @@ class AOTFxirTestCase(InductorTestCase):
         # Now the backend should have been called.
         self.assertTrue(called)
 
+    def test_dynamic_input_expr(self):
+        """
+        Test dynamic shapes with a nontrivial input expression.
+        """
+
+        class M(torch.nn.Module):
+            def forward(self, x):
+                return x.reshape(x.shape[0] * x.shape[1]) + x.shape[1]
+
+        dynamic_shapes = {"x": {0: 2 * Dim("x", min=1) + 1}}
+        inp = (torch.randn((5, 4), device=self.device),)
+        self.check(M().to(self.device), inp, dynamic_shapes=dynamic_shapes)
+
 
 if __name__ == "__main__":
     from torch._inductor.test_case import run_tests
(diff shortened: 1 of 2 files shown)


torch/_inductor/codegen/wrapper_fxir.py
```diff
@@ -141,6 +146,12 @@ class WrapperFxCodegen(PythonWrapperCodegen):
 
     supports_caching = False
 
+    def codegen_inputs(self) -> None:
+        """
+        This would generate code for symbolic input shapes, strides, etc.
+        Since the FX converter handles this, do nothing here.
+        """
```
angelayi, line 153:
```suggestion
        """
        pass
```

test/inductor/test_fxir_backend.py
```diff
@@ -866,6 +866,19 @@ def forward(self, x):
...
+    def test_dynamic_input_expr(self):
+        """
+        Test dynamic shapes with a nontrivial input expression.
+        """
+
+        class M(torch.nn.Module):
+            def forward(self, x):
+                return x.reshape(x.shape[0] * x.shape[1]) + x.shape[1]
+
+        dynamic_shapes = {"x": {0: 2 * Dim("x", min=1) + 1}}
```
blaine-rister: TODO: I should write a negative test with 2 undefined symbols, to make sure it errors gracefully.

torch/_inductor/codegen/wrapper_fxir.py
blaine-rister: Believe it or not, I did this originally, but the linter insisted on removing the `pass` 😄

jansel (approved):

test/inductor/test_fxir_backend.py
blaine-rister: I had trouble constructing a test case which would get past torch.export, so I gave up on this for now. It seems like the current logic closely mirrors which dynamic shapes torch.export will accept.

PR #163053, opened 2025-09-16
[ContextParallel][example] FLOPS measure

XilunWu:
* #165222
* #164919
* #163806
* __->__ #163053
* #163617


torch/distributed/tensor/examples/flex_perf.py
```diff
@@ -0,0 +1,1487 @@
+import csv
```
drisspg, line 1: probably not check this in right?

PR #163058, opened 2025-09-16
Refactor test_dtensor_compile

kwen2501:
* #162571
* #162549
* __->__ #163058
* #162770
* #162545
* #162529

I don't know how the following worked previously:
```
store = FakeStore()
init_process_group("fake", rank=0, world_size=2, store=store)
init_device_mesh("cpu", ...)
```
But it is not supported now.
Updating tests accordingly.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela

### DIFF
--- a/test/distributed/tensor/test_dtensor_compile.py
+++ b/test/distributed/tensor/test_dtensor_compile.py
@@ -5,6 +5,7 @@ import contextlib
 import copy
 import functools
 import unittest
+from typing import Optional
 from unittest.mock import patch
 
 import torch
@@ -39,7 +40,10 @@ from torch.distributed.tensor.parallel import (
     RowwiseParallel,
 )
 from torch.distributed.tensor.placement_types import _StridedShard
-from torch.testing._internal.common_distributed import skip_if_lt_x_gpu
+from torch.testing._internal.common_distributed import (
+    MultiProcContinuousTest,
+    skip_if_lt_x_gpu,
+)
 from torch.testing._internal.common_fsdp import get_devtype
 from torch.testing._internal.common_utils import (
     instantiate_parametrized_tests,
@@ -49,7 +53,6 @@ from torch.testing._internal.common_utils import (
     skipIfHpu,
     skipIfTorchDynamo,
     TEST_CUDA,
-    TEST_HPU,
 )
 from torch.testing._internal.distributed._tensor.common_dtensor import (
     DTensorTestBase,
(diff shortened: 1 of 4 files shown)


test/distributed/tensor/test_dtensor_compile.py
```diff
@@ -124,32 +127,33 @@ def shard_module_params(name, module, device_mesh):
...
 
-    def tearDown(self):
-        super(
-            type(self), self
-        ).tearDown()  # use explicit params for compiled autograd test wrapping
-        dist.destroy_process_group()
+    return wrapper
+
+
+class TestDTensorCompile(MultiProcContinuousTest):
```
ezyang, line 139: thanks

test/dynamo/test_fake_distributed.py
```diff
@@ -30,6 +34,7 @@ def setUp(self):
 
     def tearDown(self):
         dist.destroy_process_group()
+        dist.distributed_c10d._world = None
```
ezyang, line 37: :0 how come. I have lots of code that's only doing the first line, do I have to do this one too?

test/dynamo/test_fake_distributed.py
```diff
@@ -116,10 +121,22 @@ def forward(self, primals_1: "Sym(u0)", primals_2: "Sym(u1)", primals_3: "Sym(u2
...
+class TestDeviceMesh(DynamoTestCase):
+    def tearDown(self):
+        dist.destroy_process_group()
+        dist.distributed_c10d._world = None
+
     def test_device_mesh_get_local_rank(self):
+        os.environ["RANK"] = "0"
+        os.environ["WORLD_SIZE"] = "1"
+        os.environ["MASTER_ADDR"] = "localhost"
+        os.environ["MASTER_PORT"] = "12666"
```
ezyang, line 136: What's going on here?

ezyang (approved):

test/dynamo/test_fake_distributed.py
kwen2501: `dist.init_process_group(backend="fake")` gives `_world` a `FakeStore`, which cannot be reused for other real inits, say `init_device_mesh("cpu")` in another test method. So I have to destroy the `_world` here. 

In normal use cases, `_world` would have a usable store, which would remain valid no matter how many device groups or meshes the user has created or destroyed.

test/dynamo/test_fake_distributed.py
kwen2501: For env based `dist.init()`

test/dynamo/test_fake_distributed.py
ezyang: Would it be terrible if destroying the default PG also implicitly destroyed the world? Now that I put it this way it sounds a little naughty, but where I am getting stuck is that typically when I initialize the default process group I am always also constructing a store, and now you are saying actually the store sticks around even when I destroy it.

PR #163067, opened 2025-09-16
Add DTensor/CP support for Fused SDPA overrideable

sspintel:
In summary, this PR aims to extend DTensor and Context Parallel support for the aten._scaled_dot_product_fused_attention_overrideable ops so that vendors of devices who override this operator (in this case, HPU) can benefit from torch's DTensor framework and CP support for the attention operator.

- Add DTensor strategies for SDPA overrideable forward and backward
- Add overrideable in Ring Attention list of supported ops

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci @drisspg

### DIFF
--- a/test/distributed/tensor/test_attention.py
+++ b/test/distributed/tensor/test_attention.py
@@ -37,7 +37,11 @@ from torch.testing._internal.common_cuda import (
     PLATFORM_SUPPORTS_MEM_EFF_ATTENTION,
 )
 from torch.testing._internal.common_distributed import skip_if_lt_x_gpu
-from torch.testing._internal.common_utils import run_tests, skipIfRocm
+from torch.testing._internal.common_utils import (
+    run_tests, 
+    skipIfRocm, 
+    TEST_HPU
+)
 from torch.testing._internal.distributed._tensor.common_dtensor import (
     DTensorTestBase,
     with_comms,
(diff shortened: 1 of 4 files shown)


sspintel:
@ezyang I have extended the existing SDPA DTensor/CP tests so that it can work with the overrideable SDPA variant. I've followed the CUDA way of checking if the torch build has an expected device support (currently only HPU), and only then it will run the tests with it. HPU tests are out-of-tree and so these are expected to be skipped in torch CI.

sspintel:
Hi @ezyang Sorry for not being clear before, I've addressed it below:

Almost all of the test cases pass on the HPU with the SDPA overrideable op backend except the case for context parallel training (i.e, test_forward_only=False) because we don't support it yet. I've made some changes to skip it for now on HPU.

I'm attaching two log files of cmdline stdout that show the DTensor and CP tests I have modified for the overrideable backend passing on a 2x HPU setup. (I made the CP pytest parametrized so we can see individual test cases passing)

[test_attention.txt](https://github.com/user-attachments/files/22470017/test_attention.txt)
[test_matrix_ops.txt](https://github.com/user-attachments/files/22470020/test_matrix_ops.txt)

ezyang:
Looking over this, I would still like tests that run on our regular CI. Here is how to do the test. First, register CPU/CUDA implementation of overrideable that calls into the non-overrideable variant. Then, add an OpInfo test that exercises the overrideable signature (similar to the original one). Then show that the OpInfo test passes.

PR #163068, opened 2025-09-16
[async_tp] Support ag+mm with gather_dim lastdim of mat_A

IvanKobzarev:
* __->__ #163068

Adding ag+mm support for the case, when gather_dim is last dim of matmul (reduction dim).

When we decompose matmul by reduction dimension we result in partials that needs additional reduction,
we allocate memory for accumulator.

Decomposition should not produce small (thin) mms that can not efficiently load the GPU. Limiting for minimal size of the shard 1024 (found empirically by testing in torchtitan).

scaled_mm is not supported yet for this case.


torch/distributed/_symmetric_memory/__init__.py
```diff
@@ -638,6 +788,15 @@ def _fused_all_gather_matmul_fallback(
         A_shard.contiguous(), group_size, group_name
     )
     A = torch.ops._c10d_functional.wait_tensor(A)
+    if gather_dim == A.ndim - 1 or gather_dim == -1:
```
ngimel, line 788: Why do you need this special case? The logic below already covers all gaher_dims

torch/distributed/_symmetric_memory/__init__.py
```diff
@@ -624,6 +637,143 @@ def default_consumer(shard: torch.Tensor, rank: int) -> None:
...
+            shard.dtype,
+        )
+        return buf
+
+    local_p2p_buf = get_p2p_buf(rank)
+
+    # shards[i] => shard from rank i
+    shards = []
+    for i, c in enumerate(ag_out.chunk(group_size)):
+        shards.append(c)
```
ngimel:
```suggestion
    shards = ag_out.chunk(group_size)
```

torch/distributed/_symmetric_memory/__init__.py
```diff
@@ -624,6 +637,143 @@ def default_consumer(shard: torch.Tensor, rank: int) -> None:
...
+            stream = torch.cuda.current_stream()
+        else:
+            stream = backend_stream
+        remote_rank = (step + rank) % group_size
+        remote_p2p_buf = get_p2p_buf(remote_rank)
+        with stream:
+            copy_shard(dst=shards[remote_rank], src=remote_p2p_buf)
+            shard_consumer(shards[remote_rank], remote_rank)
+
+    if ag_out_needed:
```
ngimel, line 692: why are you doing a separate copy if you already copied everyone's shards to their correct destinations in the loop above?

ngimel (approved):

torch/_inductor/fx_passes/micro_pipeline_tp.py
```diff
@@ -645,9 +649,17 @@ def fuse_all_gather_matmul(all_gather: _AllGatherMatch) -> None:
...
         return
 
-    if gather_dim >= len(_get_tensor(shard_node).shape) - 1:
-        # Decomposing the matmul on the K dimension is not supported
-        return
+    filter_matmul = None
+    if _is_last_dim(_get_tensor(shard_node), gather_dim):
+        # Decomposed mms should not be too small
+        if _get_tensor(shard_node).shape[-1] < 1024:
+            return
```
fegin, line 656: What if we enable multimem ag + mm? Is the performance still bad?

torch/_inductor/fx_passes/micro_pipeline_tp.py
```diff
@@ -881,7 +904,7 @@ def fuse_matmul_reduce_scatter(reduce_scatter: _ReduceScatterMatch) -> None:
         return
 
     filter_matmul = None
-    if orig_scatter_dim == _get_tensor(input_node).ndim - 1:
+    if _is_last_dim(_get_tensor(input_node), orig_scatter_dim):
```
fegin, line 907: Do you see the same small last dim len issue (< 1024) for fuse matmul reduce scatter?

torch/distributed/_symmetric_memory/__init__.py
fegin: This copy is for `self` rank to copy its input shard to the output buffer.

torch/_inductor/fx_passes/micro_pipeline_tp.py
```diff
@@ -663,18 +675,29 @@ def fuse_all_gather_matmul(all_gather: _AllGatherMatch) -> None:
     if len(matmuls) == 0 or len(OrderedSet(map(type, matmuls))) != 1:
         return
 
+    if _is_last_dim(_get_tensor(shard_node), gather_dim) and len(
+        all_gather.res_node.users
```
fegin, line 679: But we always set `ag_out_needed` to True if there are users of the ag. Why is `_is_last_dim` so special that having users of all_gather causes problems? Can it be that 1024 len restriction is not enough?

torch/_inductor/fx_passes/micro_pipeline_tp.py
IvanKobzarev: Yes, for autoparallel torchtitan I observed that with TP dim = 8 and sequence length = 4096, the decomposition was slower. With sequence length = 8192 (chunked size = 1024) the AsyncTp decomposed was +4-5%.

torch/_inductor/fx_passes/micro_pipeline_tp.py
IvanKobzarev: I have not tried multimem. I will try it for small sequence size (4096) if it is better than this.

torch/distributed/_symmetric_memory/__init__.py
IvanKobzarev: Hm. The logic below for last dim fails with shape mismatch, maybe it can be generalized:
```
G = group.size()
S = full last dim, 
s = S // G (local last dim)
B = (S, N)

A_shard.shape == (B, s)
all_gather: A.shape (B * G, s)
.view(group_size, *A_shard.shape): (G, B, s)
.move_dim(2, 1): (G, s, B)
.flatten(0, 1): (G*s, B) == (S, B)

@ B (S, N) fails.
```
I did not find general formula for that, except adding .movedim(0, gather_dim) before matmul for lastdim case and not to do movedim after and decided to put a separate branch for easier understanding.

torch/distributed/_symmetric_memory/__init__.py
IvanKobzarev: This is local copy of local all gather input `shard` into allocated Tensor `ag_out` (shards). 

Before we only copied remote ranks and called shard_consumer directly from the local input `shard_consumer(shard, rank)`. 

Potentially we can do transform to make this shard_input to be a slice of pre-allocated all-gather-output, that producer of the shard_input will be writing in slice and this copy will not be needed.

torch/distributed/_symmetric_memory/__init__.py
ngimel:
```
import torch
allgathered = torch.empty(8, 18, 20)
group_size = 4
A_shard = torch.empty(2, 18, 20)
gather_dim = 2
A = allgathered
A = A.view(group_size, *A_shard.shape).movedim(gather_dim + 1, 1).flatten(0, 1)
print(A.movedim(0, gather_dim).shape) # generic path
A = allgathered
A_splits = A.chunk(group_size)
A_mm = torch.cat(A_splits, dim=-1)
print(A_mm.shape) # this path
```
prints the same shapes for both paths

IvanKobzarev: But the strides will be different:
```
torch.Size([2, 18, 80])
(18, 1, 36)
torch.Size([2, 18, 80])
(1440, 80, 1)
```

I will remove split/cat with direct usage of allocated out of all_gather with matching strides and verify perf. In this case strideness will match and no split-cat

IvanKobzarev: Tried matching strides removing separate path, but converting A_out to be contiguous for chunks on last dim to avoid split-cat makes perf significantly worse 

https://gist.github.com/IvanKobzarev/2724036c54814c06321e0c3860f7f9c7

Non contiguous copy will be significantly slower.

PR #163073, opened 2025-09-16
[inductor] in emulate_precision_casts, disable fma fusion in triton

v0i0:
* __->__ #163073

### DIFF
--- a/test/inductor/test_torchinductor.py
+++ b/test/inductor/test_torchinductor.py
@@ -14044,6 +14044,20 @@ def forward(self, arg0_1: "Sym(s77)", arg1_1: "Sym(s27)", arg2_1: "Sym(s53)", ar
         with self.assertRaises(RuntimeError):
             compiled = torch.compile(fn, backend="inductor")(a, b)
 
+    @requires_cuda_and_triton
+    @config.patch(emulate_precision_casts=True)
+    def test_emulate_precision_triton_fp_fusion(self):
+        def fn(a, b):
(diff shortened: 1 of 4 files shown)


eellison (commented):
Would you mind adding/extending existing tests to check that we do not add this flag to both:

- triton templates 
- user triton kernels

eellison (approved):
looks great, thanks!

v0i0:
@StrongerXi I tracked the referenced issue by finding a roll-out where the delta is significant, then tracked the delta to the rope procedure, and added eager casts to match inductor behavior - that way i know this rounding is critical. Then i tried to get the same behavior from inductor emulate precision casts, and it didn't do it, even tho triton had the right casts in it. So i looked at the ptx, which contained fma. I knew that in nvcc `-fmad=false` disables implicit fma generation. Grepping through triton code then finds that flag in .../nvidia/.../compiler.py, mediated by `enable_fp_fusion`. So i added this change to thread it through, and the logprob error disappeared.

@eellison I added tests that mm triton templates & user triton code do not have that flag set, and a test that the fusion is correctly being avoided.

eellison:
cc @v0i0 - this gives emulate_precision_casts its first real potential slowdown. is it worth it to config this  - potentially on by default ? thoughts ?

eellison (commented):
cc @jansel, @shunting314 for thoughts

jansel (approved):
Seems fine to me.

PR #163079, opened 2025-09-16
[ROCm][tunableop] Improvements to tunableop Numerical Check

sarthaktandon9amd:
Modified the flag PYTORCH_TUNABLEOP_NUMERICAL_CHECK, so that it accepts the numerical tolerances in the format atol_rtol as compared to the previous 0 and 1. Retains previous functionality with default values as well. 


aten/src/ATen/cuda/tunable/Tunable.cpp
```diff
@@ -521,12 +521,51 @@ void TuningContext::EnableNumericsCheck(bool value) {
...
+
+  // Default: OFF
+  if (!env_opt.has_value()) {
+    return NumericalCheckConfig(false, 1e-5, 1e-5);
+  }
+
+  const std::string& env = env_opt.value();
+
+  // Explicit OFF
+  if (env == "0" || env == "OFF" || env == "off") {
```
naromero77amd: Just check for env == "0" not variants of "off" most due to consistency with other environment variables.

aten/src/ATen/cuda/tunable/Tunable.cpp
```diff
@@ -521,12 +521,51 @@ void TuningContext::EnableNumericsCheck(bool value) {
...
+    return NumericalCheckConfig(false, 1e-5, 1e-5);
+  }
+
+  // Legacy ON with default tolerances
```
naromero77amd: Let's get rid of legacy code path.

aten/src/ATen/cuda/tunable/Tunable.cpp
```diff
@@ -521,12 +521,51 @@ void TuningContext::EnableNumericsCheck(bool value) {
...
+  const size_t underscore = env.find('_');
+  if (underscore == std::string::npos) {
+    TORCH_WARN("Invalid PYTORCH_TUNABLEOP_NUMERICAL_CHECK format. Expected 'atol_rtol', got: ", env);
+    return NumericalCheckConfig(false, 1e-5, 1e-5);
+  }
+
+  try {
+    const double atol = std::stod(env.substr(0, underscore));
+    const double rtol = std::stod(env.substr(underscore + 1));
+    if (atol <= 0.0 || rtol <= 0.0) {
```
naromero77amd: @jeffdaily Do we want to have a hard terminate if somebody puts a 0.0 or a negative number?

aten/src/ATen/cuda/tunable/Tunable.cpp
```diff
@@ -521,12 +521,51 @@ void TuningContext::EnableNumericsCheck(bool value) {
...
+      TORCH_WARN("Tolerance values must be positive. atol=", atol, ", rtol=", rtol);
+      return NumericalCheckConfig(false, 1e-5, 1e-5);
+    }
+    return NumericalCheckConfig(true, atol, rtol);
+  } catch (const std::exception& e) {
+    TORCH_WARN("Failed to parse PYTORCH_TUNABLEOP_NUMERICAL_CHECK: ", e.what());
+    return NumericalCheckConfig(false, 1e-5, 1e-5);
```
naromero77amd: Same question, I think we want to terminate.

naromero77amd (changes requested):
Good job. Some things that we can use @jeffdaily feedback on.

There are a couple of more things:
- Unit test
- Placement of numerical check in FindFastest routine after the 2nd skip
- Update the README here: https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/cuda/tunable/README.md

test/test_linalg.py
```diff
@@ -5980,6 +5980,50 @@ def test_call_count_tunableop(self, device, dtype):
...
+            with self.assertRaisesRegex(RuntimeError, r"positive"):
+                torch.cuda.tunable.set_numerical_check_tolerances(True, 1e-5, -1e5)
+            with self.assertRaisesRegex(RuntimeError, r"positive"):
+                torch.cuda.tunable.set_numerical_check_tolerances(True, -1e-5, -1e5)
+
+    @onlyCUDA
+    @skipCUDAIfNotRocm
+    @dtypes(torch.float16, torch.float32)
+    def test_numerical_check_accuracy_tunableop(self, device, dtype):
+        import os
```
naromero77amd: Do we need import os here?

test/test_linalg.py
```diff
@@ -5980,6 +5980,50 @@ def test_call_count_tunableop(self, device, dtype):
...
+        shapes = [(127,193,61), (251,317,73), (89,149,41)]
+        atol, rtol = 1e-2, 1e-1
+
+        for (m, k, n) in shapes:
+            a = torch.randn(m, k, device='cuda')
+            b = torch.randn(k, n, device='cuda')
+            with self._tunableop_ctx():
+                torch.cuda.tunable.enable(False)
+                torch.cuda.tunable.set_numerical_check_tolerances(False, 1e-5, 1e-5)
+                C_baseline = a @ b
```
naromero77amd: I don't think the baseline needs to be in a with self._tunableop_ctx().

test/test_linalg.py
```diff
@@ -185,6 +184,7 @@ def _set_tunableop_defaults(self):
         torch.cuda.tunable.set_max_tuning_duration(30)
         torch.cuda.tunable.set_max_tuning_iterations(100)
         torch.cuda.tunable.set_rotating_buffer_size(-1)
+        torch.cuda.tunable.set_numerical_check_tolerances(False, 1e-5, 1e-5)
```
naromero77amd: Question: If the first argument is False, I am assuming 1e-5 and 1e-5 can be ignored.

torch/csrc/cuda/Module.cpp
naromero77amd, line 1: You committed a hipified version of this file. We need to only commit the new stuff:
Lines:1880-1913, 2196-2199

naromero77amd (changes requested):
Great progress.

Could you please update the README.md file as well? Thanks
https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/cuda/tunable/README.md

test/test_linalg.py
sarthaktandon9amd: Yes, that would be correct. It would make those redundant.

torch/csrc/cuda/Module.cpp
```diff
@@ -1274,14 +1293,16 @@ static void registerCudaPluggableAllocator(PyObject* module) {
             self.set_release_pool(func);
           });
   m.def("_cuda_customAllocator", [](uint64_t malloc_ptr, uint64_t free_ptr) {
-    using namespace torch::cuda::CUDAPluggableAllocator;
+    using MallocFuncType = void*(size_t, int, cudaStream_t);
+    using FreeFuncType = void(void*, size_t, int, cudaStream_t);
```
naromero77amd: These lines of code should not have changed. Please revert.

torch/csrc/cuda/Module.cpp
```diff
@@ -1274,14 +1293,16 @@ static void registerCudaPluggableAllocator(PyObject* module) {
...
     std::function<MallocFuncType> malloc_fn =
         // NOLINTNEXTLINE(performance-no-int-to-ptr)
         reinterpret_cast<MallocFuncType*>(malloc_ptr);
     std::function<FreeFuncType> free_fn =
         // NOLINTNEXTLINE(performance-no-int-to-ptr)
         reinterpret_cast<FreeFuncType*>(free_ptr);
-    return createCustomAllocator(malloc_fn, free_fn);
+    return torch::cuda::CUDAPluggableAllocator::createCustomAllocator(
```
naromero77amd: These lines of code should not have changed. Please revert.

naromero77amd (changes requested):
There are still some snippets of hipified code that need to be fixed.

naromero77amd (approved):

naromero77amd:
Verifying that the new UT ran in this shard `rocm / linux-jammy-rocm-py3.10 / test (default, 2, 6, linux.rocm.gpu.2)`:
```
test/test_linalg.py::TestLinalgCUDA::test_numerical_check_accuracy_tunableop_cuda_float16, test/test_linalg.py::TestLinalgCUDA::test_numerical_check_accuracy_tunableop_cuda_float32, 
``

File: aten/src/ATen/cuda/tunable/TunableOp.h [hunk 10]
```diff
@@ -310,6 +293,22 @@ class TunableOp {
           continue;
         }
 
+        if (do_numerics_check) {
+          ParamsT* numerical_params = params->DeepCopy(false);
+          auto status = candidate->Call(numerical_params);
+          if (status != OK) {
```
naromero77amd, line 299: @jeffdaily Do we still need this `if (status != OK)` block on line 299 if now we always do this check early on line 271?

naromero77amd (commented):
@jeffdaily Can you please take a look at the one commend in the FindFastest routine? Thanks

aten/src/ATen/cuda/tunable/TunableOp.h
jeffdaily: Yes I would like to keep it.  Defensive programming.

test/test_linalg.py
jeffdaily: Could you update the python API set_numerical_check_tolerances with default values of 1e-5 for both atol and rtol?

aten/src/ATen/cuda/tunable/README.md
```diff
@@ -145,7 +145,7 @@ programmatically since the settings become fixed. Use the C++ or Python APIs ins
 | PYTORCH_TUNABLEOP_VERBOSE | Default is 0. Set to 1 to enable basic logging. 2 for basic tuning status. 3 for full trace. |
 | PYTORCH_TUNABLEOP_VERBOSE_FILENAME | Default is "err" for stderr. Set to "out" for stdout or a filename for capturing verbose logging. |
 | PYTORCH_TUNABLEOP_FILENAME | Default is 'tunableop_results.csv'. |
-| PYTORCH_TUNABLEOP_NUMERICAL_CHECK | Default is 0. Set to 1 to enable. |
+| PYTORCH_TUNABLEOP_NUMERICAL_CHECK | Default is off. Set 'atol_rtol' to enable|
```
jeffdaily: Please add example to this.
```suggestion
| PYTORCH_TUNABLEOP_NUMERICAL_CHECK | Default is off. Set "atol_rtol" to enable,  example "1e-5_1e-5". |
```

jeffdaily (approved):
A few nits, but otherwise LGTM.

aten/src/ATen/cuda/tunable/README.md
sarthaktandon9amd: I have updated the API to have the defaults as 1e-5 for both atol and rtol. I've also updated the subsequent documentation.

PR #163083, opened 2025-09-16
[sdpa] make sure to recompile if alignment is different than before

ColinPeppler:
## Context
An example from Qwen2-7B
- This come from running torch.compile with a sequence length that is
divisible by 8 (no padding needed). Call this `Run1`.
- If we then run the compiled model with a difference length that isn't
divisible by 8 (requires padding). Call this `Run2`.
- Then we'll see this error.
```
File "/var/tmp/torchinductor_nobody/2w/c2wby7ilxbna45xrtrrfjqpeutwouruviu2742ockunnd2bleeiz.py", line 1963, in call
    buf24 = torch.ops.aten._scaled_dot_product_efficient_attention_backward.default(reinterpret_tensor(buf18, (s85, 3584 // s19, s48, 512 // (512 // s19)), (s48*(512 // (512 // s19))*(3584 // s19), 512 // (512 // s19), (512 // (512 // s19))*(3584 // s19), 1), 0), buf20, buf21, buf22, buf23, getitem, getitem_1, getitem_2, getitem_3, 0.0, [True, True, True, False], scale=0.08838834764831845)
File "torch/_ops.py", line 841, in __call__
RuntimeError: attn_bias is not correctly aligned (strideM). attn_bias.stride(2) = 6102, and should be a multiple of 4.
```
- We only see the error because we did not recompile on `Run2`. Instead we ran the inputs on the same graph as `Run1`.

### A bit more on why.
Here we check whether to realize the unpadded buffer (unwrapped slice) which we want for `Run1` but not for `Run2`.
https://github.com/pytorch/pytorch/blob/0897affcd5e0e5281db7202408ca637bbf2f7abe/torch/_inductor/lowering.py#L2687-L2694

## Fix
Size hint doesn't guard, so the fix is to use `guard_or*` to guard. 

* __->__ #163083

### DIFF
--- a/test/inductor/test_cuda_repro.py
+++ b/test/inductor/test_cuda_repro.py
@@ -1348,7 +1348,7 @@ class CudaReproTests(TestCase):
             out2 = model(input2)
             out3 = model(input3)
 
-        self.assertEqual(cnts.frame_count, 1)
+        self.assertEqual(cnts.frame_count, 2)
 
     @config.patch({"triton.cudagraphs": True})
     def test_index_put_no_fallback_cudagraph(self):
--- a/test/inductor/test_cuda_repro.py
+++ b/test/inductor/test_cuda_repro.py
@@ -2215,6 +2215,109 @@ def triton_poi_fused_add_reflection_pad2d_0(in_ptr0, in_ptr1, out_ptr0, xnumel, 
                     f"Max diff: {torch.max(torch.abs(eager_output - compiled_output)):.6f}",
                 )
 
+    def test_qwen2_7b_sdpa_input_alignment_requires_recompile(self):
+        # SDPA constraints ensures inputs have alignment (8).
+        device = "cuda"
+
+        def forward(q_proj, k_proj, attn_mask):
+            scale = 0.08838834764831845  # 1/sqrt(128)
+
+            B = attn_mask.size(0)
+            S = attn_mask.size(3)
+            D = 128
+            d_model = q_proj.size(1)
+
+            query_states = q_proj.view(B, S, -1, D).transpose(1, 2)  # [B, Hq, S, D]
+            q = query_states.contiguous()
+
+            Hkv = k_proj.size(1) // D
+            Hq = query_states.size(1)
+
+            nrepeats = Hq // Hkv
+            key_states = k_proj.view(B, S, -1, D).transpose(1, 2)  # [B, Hkv, S, D]
+            kv_repeated = key_states[:, :, None, :].expand(B, Hkv, nrepeats, S, D)
+            kv_repeated = kv_repeated.contiguous()
+            k = kv_repeated.reshape(B, Hq, S, D)
+            v = k.clone()  # value tensor
+
+            inf = torch.scalar_tensor(
+                float("-inf"), dtype=torch.bfloat16, device=device
+            )
+            zero = torch.scalar_tensor(0.0, dtype=torch.bfloat16, device=device)
+            where = torch.where(condition=attn_mask, input=zero, other=inf)
+            pad_amount = 8 - (S % 8)
+            padded = torch.nn.functional.pad(
+                where, (0, pad_amount), value=0.0
+            )  # pad last-dim
+            sliced = padded[..., :S]  # back to [B,1,S,S]
+            attn_bias = sliced.expand(B, Hq, S, S)
+
+            sdpa_out, logsumexp, seed, offset = (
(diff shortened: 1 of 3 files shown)


torch/_inductor/ir.py
```diff
@@ -446,14 +446,13 @@ def is_aligned_realized_tensor_hint(
         return False
 
     aligned_strides = all(
-        (V.graph.sizevars.size_hint_or_throw(x.get_stride()[i]) % alignment) == 0
+        (V.graph.sizevars.guard_or_false(sympy.Eq(x.get_stride()[i] % alignment, 0)))
```
ColinPeppler: @laithsakka a bit concerned that we might be overguarding with this

Would a better approach be to keep the sizehint condition? So, if the sizehint condition is True, then we can guard. If the sizehint conidition is false we don't need to guard.

eellison: The other alternative is 

if statically_known_true:
- fast path
else:
- copy

I think this might be better. Wdyt @laithsakka ?

laithsakka: we can use statically known true, 
or you can create a single big sympy expressioons that represent the condition and call guard_or_false once this way you wont be over guarding. @ColinPeppler 
```
cond = sympyTrue
for i in range(len(x.get_stride()) - 1):
  cond = sympy.And(cond, sympy.Eq(x.get_stride()[i] % alignment, 0))
```
  then do 
  if V.graph.sizevars.guard_or_false(cond)
  
  Note also please instead of using % use   Mod class in 
  /home/lsakka/pytorch/torch/utils/_sympy/functions.py

ColinPeppler: my only complaint with the big sympy expression is that it isn't natural to write inductor code that way haha

ColinPeppler: okay the long sympy expr isn't too bad

```
    // A. check then guard
    aligned_strides = all(
        V.graph.sizevars.statically_known_multiple_of(x.get_stride()[i], alignment)
        for i in range(len(x.get_stride()) - 1)
    )
    aligned_last_dim = (
        V.graph.sizevars.statically_known_equals(x.get_size()[-1], 1)
        or V.graph.sizevars.statically_known_leq(x.get_size()[-1], 1)
    )
    if aligned_strides and aligned_last_dim:
        # make sure to guard
        guard_aligned_strides = sympy.And(
            *(sympy.Eq(Mod(s, alignment), 0) 
            for s in x.get_stride()[:-1])
        )
        guard_aligned_last_dim = sympy.Or(
            sympy.Eq(x.get_stride()[-1], 1),
            sympy.Le(x.get_size()[-1], 1)
        )
        V.graph.sizevars.expect_true(guard_aligned_strides)
        V.graph.sizevars.expect_true(guard_aligned_last_dim)
        return True
    return False
```
```
    // B. long sympy expr
    aligned_strides = sympy.And(
        *(sympy.Eq(Mod(s, alignment), 0) 
        for s in x.get_stride()[:-1])
    )
    aligned_last_dim = sympy.Or(
        sympy.Eq(x.get_stride()[-1], 1),
        sympy.Le(x.get_size()[-1], 1)
    )
    is_aligned = sympy.And(aligned_strides, aligned_last_dim)
    return V.graph.sizevars.guard_or_false(is_aligned)
```

eellison (approved):
Looks good, thanks ! Nice test case as well

test/inductor/test_cuda_repro.py
```diff
@@ -1348,7 +1348,7 @@ def forward(self, x):
             out2 = model(input2)
             out3 = model(input3)
 
-        self.assertEqual(cnts.frame_count, 1)
+        self.assertEqual(cnts.frame_count, 2)
```
ColinPeppler, line 1351: @eellison looks like this one didn't need a recompile on S=513 for flash attention, 

not sure if I should have separate set of guards for flash attn vs. memory efficient attn vs. cudnn...

https://github.com/pytorch/pytorch/blob/de001728d7c30a59ed985c6b14c0b6a84e65e157/test/inductor/test_cuda_repro.py#L1337-L1345

EDIT: recompile reason
`  "0/0: (x.size()[1] % 8) == 0  # (_inductor/ir.py:468 in is_aligned_realized_tensor)"`

ColinPeppler: Okay, I don't think this is a trivial re-write to avoid the alignment guard. 

If anyone thinks this PR causes too much recompiling, then feel free to revert.

PR #163089, opened 2025-09-16
Fix ONNX dynamic_axes segfault by adding validation

ghostspiders:
Fixes #163033

cc @justinchuby @titaiwangms

### DIFF
--- a/test/onnx/test_utility_funs.py
+++ b/test/onnx/test_utility_funs.py
@@ -137,6 +137,73 @@ class TestUtilityFuns(_BaseTestCase):
         )
         self.assertEqual(len(messages), 2)
 
+    def test_dynamic_axes_invalid_keys_segfault_fix(self):
+        """Test fix for segfault when dynamic_axes keys don't match input/output names."""
+        import os
+        import tempfile
+
+        import torch.nn as nn
+
+        class BaseModule(nn.Module):
+            def __init__(self):
+                super().__init__()
+                self.linear = nn.Linear(3, 1)
+
+            def forward(self, x):
+                return self.linear(x)
+
+        class MyModel(BaseModule):
+            def __init__(self):
+                super().__init__()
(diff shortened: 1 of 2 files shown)


titaiwangms (changes requested):
This is the wrong place. Legacy exporter is under torch/onnx/_internal/torchscript_exporter: specifically, I think this is the function that we could potentially improve to solve the issue - https://github.com/pytorch/pytorch/blob/c230ac7300a9c447a51b99fc7dc534c34a9d23d4/torch/onnx/_internal/torchscript_exporter/utils.py#L1883

Please also add a test with your changes.

titaiwangms (commented):
Check https://github.com/pytorch/pytorch/wiki/lintrunner for lintrunner Ci failures.

torch/onnx/_internal/torchscript_exporter/utils.py
```diff
@@ -1900,8 +1900,8 @@ def _validate_dynamic_axes(dynamic_axes, model, input_names, output_names):
     # provided dynamic axes of specified input/output
     for key, value in dynamic_axes.items():
         if key not in valid_names:
-            warnings.warn(
-                f"Provided key {key} for dynamic axes is not a valid input/output name"
+            raise ValueError(
```
titaiwangms: @justinchuby Do you think this would be a bc breaking change?

justinchuby: It could. Would it be helpful to focus on catching the segfault and error out explicitly in c++?

justinchuby: @ghostspiders

justinchuby (changes requested):
Blocking to wait for the new changes. Thanks.

ghostspiders:
@titaiwangms @justinchuby  update torch/csrc/jit/frontend/tracer.cpp

PR #156628, opened 2025-06-23
multi-kernel matmuls based on varying hint sizes

bobrenjc93:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #156628

The core idea is to generate multiple matmul kernels using different hints for symbolic variables, then select the most appropriate one at runtime for each unique shape we encounter. You can find some early experimentation details in these posts:

https://fb.workplace.com/groups/8940092306109185/posts/9803850776399996/
https://fb.workplace.com/groups/8940092306109185/posts/9695805170537891/
https://fb.workplace.com/groups/257735836456307/posts/906589324904285/

Here’s a graph illustrating the empirically observed worst-case performance if an oracle always selected the least optimal hint for a given runtime size:

image

This graph illustrates the performance of a hint size of 64 relative to the worst case. Notice that as the runtime sizes increase, the performance gradually approaches the worst case:

image

This graph shows the performance of a hint size of 4096 — very poor for small sizes, and also suboptimal for some mid-sized shapes:

image

Finally, here’s the graph that motivated this PR. It illustrates the performance when selecting the best of three kernels generated with three different hints — 64, 256, and 4096:

image

## How to review this PR

At a high level, this extends @shunting314's multi-kernel abstraction to support varying GEMM choices driven by different hints. A few key points:

1. Unlike reduction kernels, triton template matmuls pass their grid as arguments to the kernel. This PR updates `MultiKernelCall` to support kernels with varying arguments.
2. The `V.graph.sizevars.size_hints` API is extended to accept a `hint_override`, allowing us to substitute the example input’s size hint with a custom value when generating multiple kernels.
3. The choice generation and benchmarking logic is updated to support multiple hint values. One kernel is generated per value in `torch._inductor.config.multi_kernel_hints`, and at runtime, we select the most suitable kernel for the current shape.
4. This PR does not add support for cpp wrapper codegen to keep it scoped. That will be added in the next PR.

## Results 

The following is a basic test that shows our basic multi kernel working where we no longer show significant variance based on the original hint size: https://gist.github.com/bobrenjc93/ba711d529e65fd65839b34799f6323ec

Before
```
Hint\Runtime |     64     |    256     |    4096   
---------------------------------------------------
     64      |   0.0948   |   0.3124   |   4.9477  
    256      |   0.2243   |   0.2256   |   3.3880  
    4096     |   0.3384   |   0.3404   |   3.3010  
```

After
```
Hint\Runtime |     64     |    256     |    4096   
[...]

Issue #158274, opened 2025-07-14
DISABLED test_triton_relu_fused_gemm (__main__.MultiKernelTest)


huydhn:
Platforms: rocm

This test was disabled because it is failing on main branch ([recent examples](https://torch-ci.com/failure?failureCaptures=%5B%22inductor%2Ftest_multi_kernel.py%3A%3AMultiKernelTest%3A%3Atest_triton_relu_fused_gemm%22%5D)).

I think this is related to https://github.com/pytorch/pytorch/pull/156628 as the failure starts showing up when it was landed in trunk https://hud.pytorch.org/pytorch/pytorch/commit/52214485747ceb0ea47ce5a09babb0b82b3282b8.  As ROCm job has been unstable in the past few day (https://github.com/pytorch/pytorch/issues/156098, https://github.com/pytorch/pytorch/issues/158182), I try to disable the test there instead

cc @bobrenjc93

nWEIdia:
@xwang233 
@pianpwk mentioned  the forward fix PR: https://github.com/pytorch/pytorch/pull/164207

PR #163090, opened 2025-09-16
[multi-kernel] shape-similarity kernel selection

pianpwk:
Introduces a variant of size-hint multi-kernel, where for novel runtime shapes, instead of performing full benchmarking to determine the optimal kernel, selects one of many kernels pre-generated from multi-kernel hints, based off similarity b/w hint / runtime input & output shapes (L1 distance in log2 space).

Some caveats/changes:
- Size-hint multi-kernel now only kicks in if the kernel has dynamic shapes
- Pre-generation still only does 1-d search over specified hints, e.g. `matmul([s0, s1], [s1, s2])` with size-hints `[64, 256]` only generates 2 kernels - based on tuning shapes ([64, 64], [64, 64]) and ([256, 256], [256, 256]). Extending this to reasonable n-d search (via user API?) is an extension

Benchmarking results, compared to multi-kernel w/ full benchmarking (hints 64, 4096), and compiling with the ground truth hint:
550541081_1088709150049684_6528797079439730237_n

Full benchmarking doing worse is extremely weird, but we did see similar spikes in #156628


torch/_inductor/codegen/multi_kernel.py
```diff
@@ -460,6 +464,77 @@ def run(self, *args, **kwargs):
...
+        max_kernels = 4
+        assert len(timings) <= max_kernels
+        for i in range(max_kernels):
+            if i < len(self.kernels):
+                row[f"kernel{i}_path"] = get_kernel_path(self.kernels[i])
+                row[f"kernel{i}_latency"] = timings[i]
+            else:
+                row[f"kernel{i}_path"] = ""
+                row[f"kernel{i}_latency"] = ""
+        return row
```
pianpwk, line 485: just code movement

torch/_inductor/codegen/simd.py
```diff
@@ -1743,13 +1798,18 @@ def codegen_template(
                         prologue_nodes,
                         only_gen_src_code=False,
                     )
-                    kernels.append(kernel)
+                    shape_cache_key = (
+                        None
```
pianpwk, line 1799: this should probably be the concrete input values instead of None, if we wanna cache this. But wasn't sure what to fill in in the unbacked case, so didn't attempt it.

torch/_inductor/codegen/simd.py
```diff
@@ -1692,6 +1692,51 @@ def _codegen_single_template(
...
+        shapes = self._get_multikernel_shapes(node)
+        return tuple(
+            tuple(
+                hint
+                if isinstance(s, sympy.Expr) and not isinstance(s, sympy.Integer)
+                else s
+                for s in shape
+            )
+            for shape in shapes
+        )
```
pianpwk, line 1738: I'm not sure what a good approach is, should we be substituting in these hints for symbols, or replacing the whole expression?

test/inductor/test_multi_kernel.py
```diff
@@ -112,6 +122,7 @@ def fn(x, y):
                 "max_autotune": True,
                 "max_autotune_gemm_backends": "TRITON",
             },
+            dynamic=True,
```
bobrenjc93: remove since it's not representative of real world workloads?

torch/_inductor/codegen/multi_kernel.py
```diff
@@ -85,21 +97,34 @@ def define_kernel(self, kernels):
...
         buf.writeline("}")
-        buf.writeline(
-            f"{multi_kernel_name} = async_compile.multi_kernel({multi_kernel_name!r}, ["
-        )
-        with buf.indent():
-            for name in kernel_names:
-                buf.writeline(f"{name},")
-        buf.writeline(f"], arg_index=arg_index, shape_specialize={shape_specialize})")
+
+        if not shape_specialize:  # no size hint keys, just call with list of kernels
```
bobrenjc93: let's just remove the shape_specialize flag altogether? it's bad since we do syncs at runtime

torch/_inductor/codegen/multi_kernel.py
```diff
@@ -478,36 +553,64 @@ def _get_cached_shape_choice(self, cache_key):
 
     def _cache_shape_choice(self, cache_key, kernel_idx):
         """
-        Cache kernel choice for a specific shape
+        Cache kernel choice for a specific shape.
         """
         self._shape_cache[cache_key] = kernel_idx
 
+    def _l1_dist(self, k1, k2):
```
bobrenjc93: this name seems misleading? this is more like a custom log-domain heuristic?

in that vein, maybe you can also introduce some other heuristics like true euclidean distance l1 and allow users to override the heuristic strategy?

test/inductor/test_multi_kernel.py
pianpwk: sounds good, but note I added mark_dynamic instead - size-hint multi kernel now doesn't turn on if there's no dynamic shapes.

torch/_inductor/codegen/multi_kernel.py
pianpwk: renamed, but did you have an API in mind for users to specify custom heuristics?

For now just kept the one heuristic.

torch/_inductor/codegen/multi_kernel.py
pianpwk: removed the benchmark option

bobrenjc93 (approved):

nWEIdia:
Could you please double check: https://github.com/pytorch/pytorch/issues/158274#issuecomment-3349801664 
We bisected to this commit. 

python3 test/inductor/test_multi_kernel.py MultiKernelTest.test_triton_relu_fused_gemm 

compute-sanitizer shows: 

========= Invalid __global__ read of size 16 bytes                                                                                                                                 
=========     at triton_+0x840 in czzd6dgadvdl6ackf477sb5kvrkydfdtcvb6reshkdzxhnkpavje.py:83                                                                                       
=========     by thread (29,0,0) in block (148,0,0)                                      
=========     Access to 0x7f016300c050 is out of bounds                                                                                                                            
=========     and is 49233 bytes after the nearest allocation at 0x7f0162e00000 of size 2097152 bytes                                                                              
=========     Saved host backtrace up to driver entry point at kernel launch time

pianpwk:
@nWEIdia Thanks, I think https://github.com/pytorch/pytorch/pull/164207 should forward fix it if I'm not missing anything

PR #163102, opened 2025-09-16
Eliminate use of deprecated 'device' argument in Tensor.pin_memory()

drivanov:
A crash was observed due to the usage of the deprecated `device` parameter. 
```
../opt/pyg/pytorch_geometric/test/test_edge_index.py::test_data_loader[True-0-int64] DeprecationWarning: The argument 'device' of Tensor.pin_memory() is deprecated. Please do not pass this argument. (Triggered internally at /opt/pytorch/pytorch/aten/src/ATen/native/Memory.cpp:46.)
DeprecationWarning: The argument 'device' of Tensor.is_pinned() is deprecated. Please do not pass this argument. (Triggered internally at /opt/pytorch/pytorch/aten/src/ATen/native/Memory.cpp:31.)
  File "/usr/local/lib/python3.12/dist-packages/torch/utils/data/_utils/pin_memory.py", line 91, in pin_memory
    clone[i] = pin_memory(item, device)
               ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch/utils/data/_utils/pin_memory.py", line 57, in pin_memory
    return data.pin_memory(device)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/dist-packages/torch_geometric/edge_index.py", line 1211, in __torch_dispatch__
    return HANDLED_FUNCTIONS[func](*args, **(kwargs or {}))
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: _pin_memory() takes 1 positional argument but 2 were given
```

This PR removes its use to ensure compatibility with current PyTorch versions


eqy:
Would it be possible to add a test for this?

torch/utils/data/dataloader.py
```diff
@@ -680,7 +680,7 @@ def __init__(self, loader: DataLoader) -> None:
         self._pin_memory = loader.pin_memory and torch.accelerator.is_available()
 
         # Set pin memory device based on the current accelerator.
-        self._pin_memory_device = (
+        pin_memory_device = (
```
divyanshk, line 689: Do we need pin_memory_device anymore if we are not using it ?

divyanshk (changes requested):
@drivanov Seconding @eqy  here. lets add it to test_dataloader.py it already has other unit tests on pin memory. We can regroup tests here later.

Also, might have to check if we should update the docstring as well

torch/utils/data/dataloader.py
drivanov: We still use `pin_memory_device` is used in [line #L694](https://github.com/pytorch/pytorch/blob/1671acb933b43f7454e7c35d24772e958f96fe06/torch/utils/data/dataloader.py#L694) of [torch/utils/data/dataloader.py](https://github.com/pytorch/pytorch/pull/163102/files/1671acb933b43f7454e7c35d24772e958f96fe06#diff-6a03229dd777afbb9bff7c87a0018357a61afc22fa1d6083e759768c58639f38).
Since the `pin_memory device` is only used locally, I removed `self._pin_memory_device`.

drivanov:
@eqy : I moved the 6 `pin_memory` tests to `test_dataloader.py`

eqy (approved):

test/test_dataloader.py
```diff
@@ -3673,6 +3673,63 @@ def test_out_of_order_iterable_ds(self):
...
 
 
+import torch.utils.data._utils.pin_memory as pm
+
+
+class TestPinMemory(TestCase):
+    def test_pin_memory_on_tensor(self):
+        x = torch.randn(2, 3)
+        y = pm.pin_memory(x)
+        assert isinstance(y, torch.Tensor)
```
ngimel: please don't use raw asserts in tests

test/test_dataloader.py
```diff
@@ -3673,6 +3673,63 @@ def test_out_of_order_iterable_ds(self):
...
+
+    def test_dataloader_pin_memory_disabled(self):
+        ds = TensorDataset(torch.randn(8, 4))
+        dl = DataLoader(ds, batch_size=2, pin_memory=False, num_workers=0)
+
+        for (batch,) in dl:
+            assert not batch.is_pinned()
+
+    def test_pin_memory_does_not_accept_device_arg(self):
+        import pytest
```
ngimel: please don't import pytest in the testcase itself

test/test_dataloader.py
```diff
@@ -3673,6 +3673,63 @@ def test_out_of_order_iterable_ds(self):
...
+
+        # The deprecated device arg should now raise
+        x = torch.randn(2, 2)
+        with pytest.raises(TypeError):
```
ngimel: use TestCase utilities to do this

torch/utils/data/dataloader.py
ngimel: move it closer to the use then

ngimel (approved):
For the future, AI assistance on PRs has to be disclosed

torch/utils/data/_utils/pin_memory.py
```diff
@@ -15,7 +15,7 @@
 from . import MP_STATUS_CHECK_INTERVAL
 
 
-def _pin_memory_loop(in_queue, out_queue, device_id, done_event, device) -> None:
+def _pin_memory_loop(in_queue, out_queue, device_id, done_event):
```
pstjohn: Even though these are "internal" APIs, it seems they're used by torchdata: https://github.com/meta-pytorch/data/blob/92950795e0790eb74df995daf40b658e85fd2c9f/torchdata/stateful_dataloader/stateful_dataloader.py#L1005-L1011

This change breaks StatefulDataLoader until the corresponding `_pin_memory_loop` call can be updated. I'll need to test it, but I think we might be able to just add a deprecation warning here in the meantime: 

```suggestion
def _pin_memory_loop(in_queue, out_queue, device_id, done_event, device=None):

    if device is not None:
        warnings.warn(
            "The device argument to _pin_memory_loop has been deprecated and will be removed in a "
            "future version of PyTorch."
            DeprecationWarning,
            stacklevel=2,
        )
```

drivanov: Done

ngimel:
superceded by #174584

PR #163119, opened 2025-09-16
[CUDA] hack for host-memory stats test

eqy:
Basically this test appears to fail when run with other tests that manipulate pinned memory, even with the `@serialTest` decorator

An ugly workaround here is to ensure it runs first due to alphabetical ordering...

cc @ptrblck @msaroufim @jerryzh168 @tinglvv @nWEIdia

### DIFF
--- a/test/test_cuda.py
+++ b/test/test_cuda.py
@@ -175,7 +175,8 @@ class TestCuda(TestCase):
             thread.join()
 
     @serialTest()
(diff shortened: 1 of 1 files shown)


ngimel:
which of the stats is failing? It's doing empty_cache so most should be fine, and we can relax checks on those that are failing rather than resort to ugly workarounds

PR #163098, opened 2025-09-16
[optim] prevent problematic tensor aliasing in lr_scheduler

filipviz:
Prevents edge cases in SequentialLR and ReduceLROnPlateau which could corrupt learning rates or trigger recompilation.

Supersedes #162360
Fixes #162359
Fixes #163093

While putting #162360 together, I noticed the class of issue I was fixing (i.e. unintended aliasing in lr_schedulers when using Tensor lrs) appeared in several other places. @janeyx99 suggested I put together a follow-up PR.

There are several bugs resembling the one fixed in #162360. I added a helper to fix these:
```python
def _update_param_group_val(param_group: dict[str, Any], key: str, val: float | Tensor):
    """Set param_group[key] to val without aliasing or assignment when they're both tensors.
    Raises a KeyError if param_group[key] does not exist.
    """
    if isinstance(param_group[key], Tensor):
        param_group[key].fill_(_to_scalar(val))
    else:
        param_group[key] = val
```

And applied it to fix bugs in `SequentialLR.__init__` and `LRScheduler._update_lr`. I also added it to `CyclicLR.__init__` which was using an equivalent pattern, and `CosineAnnealingWarmRestarts.step` which *should* have had a similar issue:
```python
for param_group, lr in zip(self.optimizer.param_groups, self.get_lr()):
    param_group["lr"] = lr
```

But did not, because `get_lr()` actually returns tensors when using a tensor lr (despite its `list[float]` return type annotation). Relying on this propagation seems fragile, so I conservatively added the method here as well. I'll be fixing the type annotations and several related issues in followup PRs built off of this one.

Issue #125531, opened 2024-05-04
`LRScheduler` Refactoring


GdoongMathew:
## Feature
Proposal to solve issue mainly raised from issue #67760, and some other issues related to `lr_scheduler` module.

* Refactor `lr_scheduler`  module
  * make `lr_scheduler` class to support all hyperparameter settings in the given `optimizer`
  * refactor `SequentialLR` & `ChainedScheduler` to support `ReduceLROnPlateau` with arbitrary step kwargs inputs.
    * #110761 
    * #68978
* Deprecate `step(epoch)` warnings and remove supports of `epoch`, and *probably change to closed-form implementation* as well.
  * conflicts: 
    * `Secheduelr` composed by a `SequentialLR`, will raise #119168. Closed-form implementation will cause every .
    * Currently, `CyclicLR` is missing closed-form implementation, as well as `ReduceLROnPlateau`.
    * Unittests for `SequentialLR` expect all schedulers to use close-form implementation, whereas the `ChainedScheduler` is not.
  * #55135
  * #67586
  * #116776
  * #67958
* Allow `param_groups` to be passed to `lr_scheduler` class
  * #72146
  * probable issues:
    * `param_groups` must be an instance of `dict` right from the beginning, or else any additional keys added by the `optimizer` will not be accessible by the `lr_scheduler` class.
    * Currently, there's no way to validate whether the `param_groups` passed to the `lr_scheduler` class are all in the same `optimizer` instance or not.
* Deprecate `print_lr`
  * #99270

## Proposed Changes
1. A new abstract class `_SchedulerBase` will be created.
```python
class _SchedulerBase(abc.ABC):
    optimizer: Optimizer

    @abc.abstractmethod
    def step(self, **kwargs):
        raise NotImplementedError

    @abc.abstractmethod
    def state_dict(self): ...

    @abc.abstractmethod
    def load_state_dict(self): ...
```
2. Reneme `lr_scheduler` to `Scheduler` and refactor it to support all hyperparameter settings in the given `optimizer`.
```python
class Scheduler(_SchedulerBase):
    @property
    def targets(self) -> Sequence[str]:
        """The targets that the schedulers can update."""
        ...

    @property
    def last_targets(self) -> Sequence[Dict[str, Any]]:
        """The targets at the last step for each parameter group."""
        ...

    def update_targets(self, **kwargs) -> None: 
        """Update the targets (lr, momentum, ...) for each parameter group."""
        ...
    
[...]

Issue #67760, opened 2021-11-03
Add an LRScheduler interface for torch schedulers.


OliverFM:
## 🚀 Feature
Add an `LRScheduler` interface to all schedulers featuring, at a minimum, an abstract property `optimizer` and an abstract method `step`.

## Motivation

I have three reasons for opening this issue:

1. The type system is not clear -- it would be much easier to do static type checking and work with torch if we can abstract away the details of scheduler implementation when we don't need them.
2. When working on fixing #67696 I found that the testing logic is tricky to implement in a clean an reusable way as one must can't verify that the optimizer you think a scheduler is using is actualyl being used. This feature would improve the quality of the tests that we can write.
2. When using the OneCycle and CyclicLR schedulers at work, I ran into linting and typing issues with mypy (unfortunately I don't have a reproducible example at hand) as we could not specify that we wanted to use anything other than an`_LRScheduler` which is a private class.

## Pitch

Add an interface with some basic methods and properties we expect to have in a scheduler.

e.g.
```
import abc

class LRScheduler(abc.ABC):
    @abc.abstractmethod
    @property
    def optimizer(self):
        pass


    @abc.abstractmethod
    def step(self, *args, **kwargs):
        pass
```

one example where this could not be implemented in `_LRScheduler` is in `ChainedScheduler` -- this could be implemented as follows:
```
@property
def optimizer(self):
    # the validity check that is in __init__ can be moved here instead
    return self._schedulers[0].optimizer
```
[source](https://pytorch.org/docs/stable/_modules/torch/optim/lr_scheduler.html#ChainedScheduler)

Doing this ought to make working with torch easier -- when docs talk about using a scheduler, they have a clear interface to point to. This will make type checking easier too!

## Alternatives

The only alternative I have considered is making a different typing hierarchy.
It is not clear, to me at least, what the scope of the optimizer interface should be, but I think the proposal covers a good minimum.

## Additional context

I would be happy to implement this feature, if it is approved.

OliverFM:
One more point -- it is possible that the lack of the call to `super().__init__` in `ChainedScheduler` is a bug, in which case, fixing that would unblock the testing, as it looks like every scheduler has `self.optimizer` as an instance variable.

vadimkantorov:
Maybe actually a contradicting usecase: https://github.com/pytorch/pytorch/issues/67761, but IMO schedulers need a redesign :(

OliverFM:
[...]

PR #163120, opened 2025-09-16
[optim] prevent unintended aliasing in lr_scheduler; update type annotations/docs

### LINKED ISSUE #163103: [lr_scheduler] Multiple LR schedulers incorrectly alias `self._last_lr`/`get_last_lr(...)` with `group["lr"]` when using tensor lr
### 🐛 Describe the bug

`LRScheduler._update_lr`, `ChainedScheduler.__init__`, `ChainedScheduler.step`, `ReduceLROnPlateau.__init__`, `ReduceLROnPlateau.step`, and `CosineAnnealingWarmRestarts.step` alias `self._last_lr`/`get_last_lr(...)` with `group["lr"]` when `group["lr"]` is a tensor with patterns like this:
```python
self._last_lr: list[float] = [group["lr"] for group in self.optimizer.param_groups]
```

This can lead to unintended lr corruption and violates type annotations (which currently mark `_last_lr` and `get_last_lr(...)` as `list[float]`.

### Versions

PyTorch version: 2.10.0.dev20250916
Is debug build: False

OS: macOS 15.6.1 (arm64)
CMake version: version 3.25.0
Libc version: N/A

Is CUDA available: False
CUDA runtime version: No CUDA
cuDNN version: No CUDA
Is XPU available: False
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M3 Pro

[pip3] flake8==7.3.0
[pip3] flake8-pyi==25.5.0
[pip3] mypy==1.16.0
[pip3] mypy-extensions==1.0.0
[pip3] numpy==2.1.0
[pip3] optree==0.13.0
[conda] Could not collect

filipviz:
1. Prevents unintended aliasing of `self._last_lr`/`get_last_lr(...)` with `group["lr"]` when `group["lr"]` is a tensor.
2. Prevents unintended aliasing of `LRScheduler.base_lrs` with the `group["initial_lr"]`s.
3. Updates `test/optim/test_lrscheduler.py` to test tensor LRs.
4. Changes type annotations for `_last_lr`, `get_last_lr()`, `base_lrs`, `get_lr()`, and `_get_closed_form_lr()` from `list[float]` to `list[float | Tensor]`; adds documentation.

Fixes #163103

LR schedulers can behave in unexpected ways when using a tensor LR due to patterns like this:
```python
self._last_lr: list[float] = [group["lr"] for group in self.optimizer.param_groups]
```

This PR adds a helper to address this:
```python
def _param_groups_val_list(optimizer: Optimizer, key: str) -> list[Any]:
    """Create a list containing group[key] for each optimizer param_group.
    Prevents aliasing when group[key] could be a Tensor.
    Raises a KeyError when group[key] does not exist.
    """
    return [
        group[key].clone() if isinstance(group[key], Tensor) else group[key]
        for group in optimizer.param_groups
    ]
```


filipviz:
Sounds interesting @vadimkantorov. Does this come up often? I'd be open to adding something like this soon, but I want to expand lr_scheduler test coverage first (once this and #163122 land).

janeyx99 (commented):
Thanks for clarifying what's going on! One highlevel thing I would ask for this PR to clarify (and document) is whether users expect get_lr or get_last_lr to return aliases vs copies of the Tensor lrs. We should at the very least document what is happening now (e.g., after this change, get_lr would return copies instead of aliases). And as that behavior is changing from previous releases, it is BC-breaking. Though, for this case, I think that it is fair to consider this a bug fix, as get_lr used to return immutable floats, and so we should follow that semantic and return un-aliased Tensors.

janeyx99:
What @vadimkantorov is talking about is something out of the scope for this PR, and what our get_closed_form functions attempt to help solve. (Vadim correct me if i'm misunderstanding!) At one time there was an effort to deprecate all the nonclosed forms, but it was unsuccessful because there were still use cases we didn't cover, and the person who maintained LRScheduler stopped doing so. So if you want a good can of worms, this situation may fall under that category 😛

filipviz:
Yeah, great points @janeyx99 - the PR was underbaked. Should be better now! Here's an overview:

### Current behavior

There are several rough edges caused by patterns like this in `lr_scheduler.py`:
```python
self._last_lr: list[float] = [group["lr"] for group in self.optimizer.param_groups]
```

1. When `group["lr"]` is a tensor, `_last_lr` and `get_last_lr()` are aliased tensors *despite their type annotation*, and are silently mutated in future calls to `.step()` (which updates `group["lr"]` in-place). If the user modifies the outputs of `get_last_lr()` in place, they silently corrupt their learning rates! For example:
```python
import torch
from torch import optim, nn

lr = torch.tensor(0.1)
opt = optim.SGD([nn.Parameter(torch.tensor(1.0))], lr=lr)
scheduler = optim.lr_scheduler.ConstantLR(opt, total_iters=1)

x = scheduler.get_last_lr()[0]
print(x)  # tensor(0.0333)

opt.step()
scheduler.step()  # mutates x
print(x)  # tensor(0.1000)

x /= 2  # corrupts param_groups[0]["lr"]
```

This causes unexpected behavior in common workflows:
```python
history = []
for step in total_iters:
    opt.step()
    scheduler.step()
    history.append(scheduler.get_last_lr()[0])
print(history)  # All values == history[-1]
```

2. `get_lr()` also had an inaccurate type annotation, and behaved inconsistently - `MultiplicativeLR`, `StepLR`, `MultiStepLR`, `ConstantLR`, `LinearLR`, `ExponentialLR`, `PolynomialLR`, and `CosineAnnealingLR` *sometimes* returned `group["lr"]` tensors directly, depending on the circumstances. For example, in `StepLR.get_lr`:
```python
@override
def get_lr(self) -> list[float | Tensor]:
    """Compute the learning rate of each parameter group."""
    _warn_get_lr_called_within_step(self)

    if (self.last_epoch == 0) or (self.last_epoch % self.step_size != 0):
        return [group["lr"] for group in self.optimizer.param_groups]  # Aliased
    return [group["lr"] * self.gamma for group in self.optimizer.param_groups]  # New tensors
```

This isn't a huge issue since `get_lr` isn't really user-facing, but I think it's worth fixing.

Aside: it's not clear to me why `get_lr` is a public method. It seems [to cause confusion](https://discuss.pytorch.org/t/whats-the-difference-between-get-lr-and-get-last-lr/121681). Should we consider moving it to `_get_lr` and adding a deprecation warning to `get_lr`?

3. `base_lrs` can be tensors which alias `group["initial_lr"]`, violating its type annotation. The aliasing is less of a concern here as neither is supposed to change. But when `base_lrs` are tensors, this propagates through the math in `_get_closed_form_lr()`, meaning its type annotation is inaccurate as well.

### New behavior

This PR makes the existing type behavior explicit, but without the alias footguns. `_last_lr`, `get_last_lr()`, `base_lrs`, `get_lr()`, and `_get_closed_form_lr()` are now all `list[float | Tensor]` with each entry matching the type of the corresponding `group["lr"]`.

Note that this PR doesn't add tests to enforce matching types, and that that `CyclicLR` and `OneCycleLR` still have slightly idiosyncratic behavior. I plan to address these in a follow-up PR to make reviewing easier, but can add them here if desired.

### Alternatives

1. If we instead wanted to satisfy the annotations by changing the behavior, we could do something like this:
```diff
- self._last_lr = [group["lr"] for group in self.optimizer.param_groups]
+ self._last_lr = [float(group["lr"]) for group in self.optimizer.param_groups]
```
Now that we use `_update_param_group_val` from #163098 this might not trigger recompilations each step, but I haven't tested this.

2. `get_last_lr()` returns `self._last_lr` directly. I think this is less of a concern since `_last_lr` is re-assigned each iteration, but we could consider defensively returning copies.

### Aside

What's the process for particularly marginal fixes? For example, I noticed `ConstantLR` currently checks:
```python
if factor > 1.0 or factor < 0:
    raise ValueError(
        "Constant multiplicative factor expected to be between 0 and 1."
    )
```
Which allows `factor=0`, which will cause a `ZeroDivisionError` error once `self.last_epoch == self.total_iters`. Should check for `factor <= 0`. Separate PR?

torch/optim/lr_scheduler.py
```diff
@@ -177,12 +188,26 @@ def load_state_dict(self, state_dict: dict[str, Any]):
...
+    def get_last_lr(self) -> list[float | Tensor]:
+        """Get the most recent learning rates computed by this scheduler.
+
+        Returns:
+            list[float | Tensor]: A :class:`list` of learning rates with entries for each of the optimizer's
+            :attr:`~torch.optim.Optimizer.param_groups`, with the same types as their ``group["lr"]``s.
+
+        .. note:: The returned :class:`~torch.Tensor`s are copies, and never alias the optimizer's ``group["lr"]``s.
+        """
         return self._last_lr
```
janeyx99, line 225: is the self._last_lr here actually a copy?

janeyx99 (approved):
Overall looks great! Thank you for looking into it. I agree let's land this now + build more on top later.

With get_lr...yea deprecating + privatizing it sounds good given the context you've mentioned. Can you check through GH/some issues to ensure that no one is relying on get_lr in a way we don't expect though? (This is for a future PR, the current one is good as is.)

torch/optim/lr_scheduler.py
filipviz: `_last_lr` is now always updated with:
```python
self._last_lr = _param_groups_val_list(self.optimizer, "lr")
```
So it's a `.clone()` of the `group["lr"]`s. But we *are* returning `_last_lr` directly - it's technically scheduler state, so if we want to be more defensive, we could clone again. I think this is less of a concern since `_last_lr` is re-assigned each step and is only used in `get_last_lr()`

filipviz:
> Also, is there a reason get_lr is not returning copies like get_last_lr? Does our internal usage depend on mutating the lr? (I would guess that we'd also want get_lr to return copies)

`get_lr()` now *does* return copies in cases where it previously would have returned the `group["lr"]` tensors directly. For example, in `StepLR.get_lr`:
```diff
@override
def get_lr(self) -> list[float | Tensor]:
    """Compute the learning rate of each parameter group."""
    _warn_get_lr_called_within_step(self)

    if (self.last_epoch == 0) or (self.last_epoch % self.step_size != 0):
-        return [group["lr"] for group in self.optimizer.param_groups]  # Returned group["lr"]s
+        return _param_groups_val_list(self.optimizer, "lr") # Uses .clone()
    return [group["lr"] * self.gamma for group in self.optimizer.param_groups]  # New tensors
```

torch/optim/lr_scheduler.py
```diff
@@ -177,12 +188,26 @@ def load_state_dict(self, state_dict: dict[str, Any]):
...
 
-    def get_lr(self) -> list[float]:
-        """Compute learning rate using chainable form of the scheduler."""
+    def get_lr(self) -> list[float | Tensor]:
+        """Compute the next learning rate for each of the optimizer's :attr:`~torch.optim.Optimizer.param_groups`.
+
+        Returns:
+            list[float | Tensor]: A :class:`list` of learning rates for each of the optimizer's
```
janeyx99: Ah I asked my question cuz I would expect get_lr here to also mention that it is a copy and not alias

torch/optim/lr_scheduler.py
```diff
@@ -177,12 +188,26 @@ def load_state_dict(self, state_dict: dict[str, Any]):
...
-    def get_last_lr(self) -> list[float]:
-        """Return last computed learning rate by current scheduler."""
+    def get_last_lr(self) -> list[float | Tensor]:
+        """Get the most recent learning rates computed by this scheduler.
+
+        Returns:
+            list[float | Tensor]: A :class:`list` of learning rates with entries for each of the optimizer's
+            :attr:`~torch.optim.Optimizer.param_groups`, with the same types as their ``group["lr"]``s.
+
+        .. note:: The returned :class:`~torch.Tensor`s are copies, and never alias the optimizer's ``group["lr"]``s.
```
janeyx99: Oh your docs are likely failing cuz you use quotation marks here! Maybe try backslash escaping them?

torch/optim/lr_scheduler.py
janeyx99: Ah, can you add a comment at that return saying 
// self._last_lr should be a clone of the group["lr"] already

I agree it's less of a concern, more wanted to make sure that if I read this later I understand what;s going on better!

torch/optim/lr_scheduler.py
filipviz: Hmm good point. I didn't note that because I felt the method semantics already suggested these would be new tensors. Want me to add a note?

torch/optim/lr_scheduler.py
filipviz: That should be fine, right? The [`load_state_dict`](https://docs.pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.LRScheduler.html#torch.optim.lr_scheduler.LRScheduler.load_state_dict) methods already have unescaped quotes and aren't using raw strings. Same for e.g. [EmbeddingBag](https://docs.pytorch.org/docs/stable/generated/torch.nn.EmbeddingBag.html).

I'm noticing the error comes up for `SWALR` even though I didn't modify `swa_utils.py`, meaning this must come from the inherited `get_last_lr()`. I'll keep looking into this.

filipviz: Ahhh those were just warnings. The actual issue was:
```log
ParseError: KaTeX parse error: Expected 'EOF', got '_' at position 12: \texttt{eta_̲min} + \frac{1}…
    at new ParseError (/usr/local/share/.config/yarn/global/node_modules/katex/dist/katex.js:108:18)
    at Parser.expect (/usr/local/share/.config/yarn/global/node_modules/katex/dist/katex.js:17762:13)
    at Parser.parseArgumentGroup (/usr/local/share/.config/yarn/global/node_modules/katex/dist/katex.js:18548:10)
    at Parser.parseGroupOfType (/usr/local/share/.config/yarn/global/node_modules/katex/dist/katex.js:18296:21)
    at Parser.parseArguments (/usr/local/share/.config/yarn/global/node_modules/katex/dist/katex.js:18261:24)
    at Parser.parseFunction (/usr/local/share/.config/yarn/global/node_modules/katex/dist/katex.js:18210:14)
    at Parser.parseGroup (/usr/local/share/.config/yarn/global/node_modules/katex/dist/katex.js:18603:21)
    at Parser.parseAtom (/usr/local/share/.config/yarn/global/node_modules/katex/dist/katex.js:18035:23)
    at Parser.parseExpression (/usr/local/share/.config/yarn/global/node_modules/katex/dist/katex.js:17889:25)
    at Parser.parse (/usr/local/share/.config/yarn/global/node_modules/katex/dist/katex.js:17821:26) {
  position: 11,
  length: 1,
  rawMessage: "Expected 'EOF', got '_'"
}
```
Looks like one has to backslash escape underscores in the KaTeX `\texttt` environment. Should work now!
image

torch/optim/lr_scheduler.py
filipviz: Added!

torch/optim/lr_scheduler.py
```diff
@@ -582,18 +647,40 @@ def __init__(
...
+        """
         _warn_get_lr_called_within_step(self)
 
         if self.last_epoch not in self.milestones:
-            return [group["lr"] for group in self.optimizer.param_groups]
+            return _param_groups_val_list(self.optimizer, "lr")
         return [
             group["lr"] * self.gamma ** self.milestones[self.last_epoch]
             for group in self.optimizer.param_groups
         ]
```
vadimkantorov, line 757: For closed-factor stuff, I propose to factor out these to e.g. function `torch.optim.scheduler._step_scheduler(...)` - so that they can in the future be used directly without having to create `LRScheduler` first. (and existing `_get_closed_form_lr` could also then use `torch.optim.scheduler._step_scheduler(...)` internally)

filipviz: I'm not sure I understand. Intuitively, wouldn't we want to keep most computation in some functional `def compute_lr(step, total_iters, base_lr, ...)` and turn the "stateful per-iteration stepping" into a thin wrapper around that? Meaning we wouldn't need these incremental forms anymore.

Maybe we can discuss further in a separate issue - #125531, #67760, or a new one.

torch/optim/lr_scheduler.py
janeyx99: Adding a note similar to get_last_lr would be great!

torch/optim/lr_scheduler.py
```diff
@@ -1992,8 +1994,8 @@ def get_lr(self) -> list[float | Tensor]:
         Computes learning rates for the optimizer's :attr:`~torch.optim.Optimizer.param_groups` following:
 
         .. math::
-            \texttt{eta_min} + \frac{1}{2}(\texttt{base_lr} - \texttt{eta_min})\left(1 +
-            \cos\left(\frac{\texttt{T_cur}}{\texttt{T_i}}\pi\right)\right)
+            \texttt{eta\_min} + \frac{1}{2}(\texttt{base\_lr} - \texttt{eta\_min})\left(1 +
+            \cos\left(\pi \cdot \frac{\texttt{T\_cur}}{\texttt{T\_i}}\right)\right)
```
janeyx99: It was the underscore that was the issue? Dang

janeyx99: This is surprising cuz the underscores were there ebfore!

filipviz: The underscores that were there in class-level `CosineAnnealingLR` docstring were fine - mine were problematic because they're in the `\texttt` environment which has some quirks in KaTeX (used by Sphinx).

torch/optim/lr_scheduler.py
filipviz: Done!

torch/optim/lr_scheduler.py
filipviz: Okay, now seeing that even warnings cause docs CI to fail. I should be able to figure these out - I'll get the docs working locally next time so we can avoid the back-and-forth. Sorry!

janeyx99: Ping me when it's ready or if you have any qs on how I can help! I do think our docs build is a bit finnicky to get right locally.

filipviz: @janeyx99 should be fixed now - sorry for the delay! I documented the issues I ran into in #163789

PR #163121, opened 2025-09-16
Turn on capture_scalar_outputs when fullgraph=True

bobrenjc93:
* #163246
* #163123
* __->__ #163121


test/dynamo/test_error_messages.py
```diff
@@ -91,29 +91,6 @@ def fn():
     return torch.linalg.lstsq(torch.rand(10, 10), torch.rand(10, 10))""",
             )
 
-    def test_data_dependent_operator(self):
```
bobrenjc93, line 94: this test is no longer relevant since we don't expect to ever throw an error anymore. in the case of fullgraph=False we graph break (but don't error, which is what this test suite is for)

laithsakka:
Can we add a comment closer to the when fullgraph=True documentation
this is opt into unbacked semantics implicitly?

laithsakka (approved):

PR #163123, opened 2025-09-16
Turn on capture_dynamic_output_shape_ops when fullgraph=True

bobrenjc93:
* #163246
* __->__ #163123
* #163121


test/dynamo/test_error_messages.py
```diff
@@ -48,27 +48,6 @@ def __exit__(self, exc_type, exc_value, traceback):
 
 
 class ErrorMessagesTest(LoggingTestCase):
-    def test_dynamic_shape_operator(self):
```
bobrenjc93, line 51: this test is no longer relevant since we don't expect to ever throw an error anymore. in the case of fullgraph=False we graph break (but don't error, which is what this test suite is for)

laithsakka (approved):
Can we add a comment closer to the when fullgraph=True documentation
this is opt into unbacked semantics implicitly?

laithsakka:
Accepted please address my comment before landing this

laithsakka:
Would be nice to print a log info that says " bla bla bla was turned on potentially opting in unbacked semantics because fullgraph=True"

PR #163125, opened 2025-09-16
Use computed buffer sizes of torch for cusparseLt metadata

aartbik:
Making sure buffer allocation matches what is computed by cusparseLt compression

### DIFF
--- a/aten/src/ATen/native/sparse/cuda/cuSPARSELtOps.cpp
+++ b/aten/src/ATen/native/sparse/cuda/cuSPARSELtOps.cpp
@@ -115,8 +115,13 @@ at::Tensor _cslt_compress(const Tensor& sparse_input) {
       &compressed_buffer_size));
 
   // create a new compressed tensor with the same dtype as the input,
-  // and with sufficient size for the packed data/metadata (given in bytes)
-  auto compressed_tensor = sparse_input.new_empty(compressed_size / sparse_input.itemsize());
+  // and with packed data/metadata stored in an array with original
+  // number of rows, and sufficient columns to provide compressed_size
+  // buffer (in bytes)
+  size_t orig_m = sparse_input.size(0);
+  size_t div = orig_m * sparse_input.itemsize();
+  size_t new_n = (compressed_size + div - 1) / div; // floor
+  auto compressed_tensor = sparse_input.new_empty({(int64_t)orig_m, (int64_t)new_n});
 
   auto& allocator = *::c10::cuda::CUDACachingAllocator::get();
   auto compressedBufferPtr = allocator.allocate(compressed_buffer_size);
@@ -162,7 +167,6 @@ std::tuple<at::Tensor, int64_t, int64_t, int64_t, int64_t> _cslt_sparse_mm_impl(
   cudaDataType output_type;
   cudaDataType C_type;
   cusparseComputeType compute_type;
-  auto compression_factor = 9;
 
   #ifdef USE_ROCM
   TORCH_CHECK(isHipSparseLtSupported());
@@ -174,7 +178,6 @@ std::tuple<at::Tensor, int64_t, int64_t, int64_t, int64_t> _cslt_sparse_mm_impl(
       output_type = CUDA_R_8I;
       C_type = CUDA_R_8I;
       compute_type = CUSPARSE_COMPUTE_32I;
-      compression_factor = 10;
       break;
 
 // cuSPARSELt v0.5.2 onwards changes CUSPARSE_COMPUTE_TF32, CUSPARSE_COMPUT_16F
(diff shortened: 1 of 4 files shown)


jcaip (approved):
lgtm, thanks for the PR!

aartbik:
I actually think we will need to add a compression_factor = 10 to the at::ScalarType::Float case (4 bytes) to make it work for float32. I am hoping the other branches work as well.

Are these paths tested anywhere that you know?

aartbik:
Assuming this is all good, can someone please issue the magical "@pytorchbot merge" for me?

jcaip:
> Yes, float is 1:2 not 2:4 like the others. But how about this change? Rather than computing the compressed size in terms of original length (which is complicated by having different sizeof(T) for the different T for the matrix, while metadata is always in bits), we simply use the computed size (in bytes) directly (in sizeof(T)). Since T is restricted to 1,2,4 bytes, and we have powers of two in rows/cols, this will always give a proper division. Looks a lot simpler and safer?

Sorry, I think this might cause some issues for us. This will work fine for `A_sparse = torch._cslt_compress`, which will return a compressed tensor of the right size. Usually this is just 10/16 or 9/16 based on the dtype, but for small tensors I think this this formula breaks down (there's a min size iirc). 

but if we pass this to `torch._cslt_sparse_mm(A_sparse, B)`, then we could run into an error because we rely on the inverse formula to calculate the original matmul shape (see [here](https://github.com/pytorch/pytorch/blob/7a0f93344e2c851b9bcf2b9c3225a323d48fde26/aten/src/ATen/native/sparse/cuda/cuSPARSELtOps.cpp#L305))
and there's no way we can invert the `compressedSizeCalculation`. 

So as a hack, I just used a fixed 10/16 compression ratio, since we can always do calculate the reverse. In the worst case, we over allocate some memory, but that's okay. Let me know if this makes sense to you. 

I think as an option we could also pass the original `m` value in  `torch.cslt_sparse_mm` so we don't have to recalculate as well, but that would require more design. Or if you have other ideas for how to better design this I'm happy to hear suggestions.

jcaip:
I think we should use compressed size and then throw an error or warning when the compressed size is not equal to 9/16 or 10/16. 

>However, this still hits cases where we have under allocated, which actually seems to indicate potential overrun bugs?

Yes this is currently an issue - I apologize I think I switched up the under / over allocation in my comment above. 

The problem is that the metadata for cuSPARSELt is calculated for blocks of (128, 64), so an input of say shape (512, 32) will cause `_cslt_compress` to buffer overflow, as the 9/16 calculation is not correct since the input shape will be padded. This is really dependent on the cuSPARSELt version I think, as I don't have insight into how the metadata size is calculated and I think this may have changed across versions (the shape at the least if not the actual metadata format itself).

These numbers were failing on an earlier version but seem to be working now with the latest version. But you can run the below script to reproduce this bug for the current version, just with different shapes

```python
import random
import torch
from torch.sparse import to_sparse_semi_structured, SparseSemiStructuredTensor

def rand_sparse_semi_structured_mask(
    r, c, dtype=torch.float16, device="cuda", choice=None
):
    """
    This function returns a 1:2 sparse matrix of size (r, c).
    Note that this means this matrix will also be 2:4 and 4:8 sparse as well.
    """

    choices = [[0, 1], [1, 0]]
    mask_entries = [choice or random.choice(choices) for i in range(r * c // 2)]

    return (
        torch.tensor(mask_entries, dtype=dtype, device=device)
        .reshape(r, c)
        .contiguous()
    )

m, k, n = 512, 16, 512

SparseSemiStructuredTensor._FORCE_CUTLASS = False
A = rand_sparse_semi_structured_mask(m, k, dtype=torch.float16)
A_sparse = to_sparse_semi_structured(A)

B = torch.rand((k, n), device=A_sparse.device).to(torch.float16)

dense_result = torch.mm(A, B)
print(dense_result)
print(A_sparse.packed.data_ptr())
cloned_A = A_sparse.packed.clone()
print(cloned_A.data_ptr())
sparse_result = torch._cslt_sparse_mm(cloned_A, B)
print(sparse_result)
print(sparse_result.shape)
assert torch.allclose(dense_result, sparse_result, rtol=1e-3, atol=1e-3)
```

aartbik:
Yes, any attempt to compute the amount of compressed storage for cuSPARSELt is dangerous, since it may change over the different releases and may indeed be irregular for smaller sizes (i.e. not a closed formula computation). And *under* allocating is of course super dangerous and may result buffer overrun and incorrect results.

I came up with the following
(1) use computed value for buffer allocation
(2) encode the original M in the metadata, which is now MxS for sufficient large S

This will
(1) always have correct size, i.e. never under allocate
(2) typically be exact, i.e. not too much over allocation
(3) completely remove the 9/16 10/16 compression logic

Let me know if you like this new approach! I also change the title of this PR.

jcaip:
Yup this all looks good - thanks for the fix!

PR #163130, opened 2025-09-17
[FSDP2] idempotent reset_sharded_param: no-op if _local_tensor is already padded

### LINKED ISSUE #1136: Inconsistent loss when resume training with vocab size that is not divisible by world size.
### Bug description

When I use an tokenizer that has vocabulary size that is not divisible by parallel (or world) size, the training loss will become inconsistent after resuming.

### Versions

can be reproduced using torch2.6

Reproduce:
Use any tokenizer that has a vocabulary size that is not divisible by parallel (or world) size.
train from step 0 to step 20:

Image
load step 10 checkpoint and resume training:

Image
As shown, step 11 and following steps have inconsistent loss.

weifengpy:
resolves https://github.com/pytorch/torchtitan/issues/1136

torchtitan use cached state dict for ft. reset_sharded_param should be idempotent if model.parameters() are padded already

```
# pad DTensor._local_tensor
fully_shard(model)
sd = fsdp_model.state_dict()
# reset_sharded_param should be a no-op in lazy_init
loss = fsdp_model(inp).sum()
```

this PR make `reset_sharded_param` idempotent by checking storage data ptr and return early

unit test
```
pytest -s test/distributed/_composable/fsdp/test_fully_shard_state_dict.py -k test_cached_state_dict
```
* __->__ #163130

### DIFF
--- a/torch/distributed/fsdp/_fully_shard/_fsdp_param.py
+++ b/torch/distributed/fsdp/_fully_shard/_fsdp_param.py
@@ -837,6 +837,7 @@ class FSDPParam:
         # 1st time in fully_shard(model)
         # 2nd time in model(input) lazy_init
         # 2nd time should be no-op if parameters remain unchanged
+        # 2nd time shouldn't be no-op if people call model.load_state_dict(...) before lazy_init
         # this makes it possible for trainer to call `sd = model.state_dict()` before the training loop
         # and use `sd` without calling .state_dict() per iteration
         same_local_tensor = (
--- a/torch/distributed/fsdp/_fully_shard/_fsdp_param.py
+++ b/torch/distributed/fsdp/_fully_shard/_fsdp_param.py
@@ -253,11 +253,6 @@ class FSDPParam:
                 lambda *args, **kwargs: self.reset_sharded_param()
             )
         )
-        self._state_dict_pre_hook_handle = (
-            module_info.module.register_state_dict_pre_hook(
-                lambda *args, **kwargs: self.reset_sharded_param()
-            )
-        )
 
     @torch.no_grad()
     def _init_sharded_param(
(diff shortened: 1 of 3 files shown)


weifengpy:
ci error is not relevant

fegin:
@awgu The current issue is that some training frameworks, including TorchTitan, calls `model.state_dict()` before the first `forward()` and uses that result throughout the entire training without calling `model.state_dict()` again. This will result in incorrect checkpoints being saved.

torch/distributed/fsdp/_fully_shard/_fsdp_param.py
```diff
@@ -833,10 +833,20 @@ def reset_sharded_param(self):
         if local_tensor.is_meta:
             return
         updated_local_tensor = False
+        # `reset_sharded_param` can be called twice
+        # 1st time in sd = model.state_dict()
```
tianyu-l: Please update comments based on offline discussions:
- first time should be during `fully_shard` call
- 2nd time could happen with / without state dict load. If with load, the 2nd time should not be a no-op.

tianyu-l (approved):
Thanks for the fix!

weifengpy:
> where did the padding get lost in the first place / why was there not padding re-added at that point

good question! no padding is getting losted. here is what I want to achieve

fully_shard(model) with padded local_tensor  -> model.state_dict() -> model(input) **reset_sharded_param should be no-op**

I just need reset_sharded_param to be idempotent. without the PR, we are always creating new padded tensors, because `local_tensor.size() != padded_sharded_size` is always true

No need to call reset_sharded_param in state dict hooks. I modified the PR

torch/distributed/fsdp/_fully_shard/_fsdp_param.py
weifengpy: updated. good catch!

fegin:
@weifengpy The test failure looks real.

wwwjn:
> model.state_dict() -> model(input) reset_sharded_param should be no-op

Thanks for the fix @weifengpy ! By reading the code and comments, the `reset_sharded_param()` should only pad again if the model local tensor changed, right?

PR #163134, opened 2025-09-17
Remove data_source argument from Sampler

cyyever:
`data_source` is declared being removed in PT 2.2 but not.


ezyang (approved):

ezyang:
fbcode looks safe-ish
https://www.internalfb.com/code/search?q=case%3Ayes%20regex%3Aon%20lang%3Apython%20repo%3Afbcode%20Sampler.*data_source

PR #163145, opened 2025-09-17
[Triton] [Inductor] Enable Epilogue Subtiling in the blackwell ws template

njriasan:
Summary: Enables support for epilogue subtiling in the blackwell ws template. This requires the ability to call `store_output` twice in the same kernel and reuse the same tensor descriptor across allocations.

Test Plan:
Tested with test_max_autotune.py on a Blackwell server.

Rollback Plan:

Differential Revision: D82610077


test/inductor/test_max_autotune.py
```diff
@@ -271,12 +271,14 @@ def next_multiple_16(a: int) -> int:
     @parametrize("b_transposed", (False, True))
     @parametrize("dynamic", (False, True))
     @parametrize("tma_store", (False, True))
+    @parametrize("epilogue_subtile", (False, True))
```
PaulZhang12, line 274: For these tests, do we know if the epilogue subtile wins? If so, can we assert that the correct codegen for epilogue subtiling exists

njriasan: At this time I was checking that the code doesn't crash, but I should just look for the store count (2 vs 1). I can update the test. Thanks for the feedback.

torch/_inductor/codegen/simd.py
```diff
@@ -416,6 +416,8 @@ def __init__(
         )
         self.no_x_dim = self.want_no_x_dim()
         self.code_hash: Optional[str] = None
+        # Info to enable multiple store_output calls for epilogue subtiling
+        self.store_output_ctr = itertools.count()
```
PaulZhang12, line 418: This is a bit confusing to follow, can we make this just an int?

torch/_inductor/codegen/triton.py
```diff
@@ -2485,42 +2487,49 @@ def codegen_block_ptr(
...
-            else:
-                self.body.writeline(line_body)
-
-            if isinstance(indexing, BlockPtrOptions):
-                # Store for later use. If the buffer is removed the below advancements
-                # are no longer necessary
-                self.block_ptr_to_buffer[block_descriptor] = name
+                block_descriptor_id = next(self.block_ptr_id)
+                if isinstance(indexing, BlockPtrOptions):
+                    block_descriptor = f"block_ptr{block_descriptor_id}"
```
PaulZhang12, line 2497: Did you test with block pointer? I thought we don't really use block_ptr anymore in triton

njriasan: Block pointers are used by other workloads, typically with out of tree third-parties. That is to avoid breaking existing tests.

torch/_inductor/codegen/simd.py
njriasan: I'm following the standard practice for other variable names. If that style is not preferable I'm happy to change it, but I want to be consistent as much as possible.

torch/_inductor/kernel/mm.py
```diff
@@ -645,14 +645,34 @@ def _compute_pid(tile_id, num_pid_in_group, grid_m, GROUP_M: tl.constexpr, NUM_S
         )
         offs_cm = pid_m * BLOCK_M
         offs_cn = pid_n * BLOCK_N
-        # TODO: Add EPILOGUE_SUBTILE
+        {%- if EPILOGUE_SUBTILE %}
+        acc = tl.reshape(accumulator, (BLOCK_M, 2, BLOCK_N // 2))
```
eellison, line 650: nit: assert BLOCK_N is divisible by 2

torch/_inductor/kernel/mm.py
```diff
@@ -645,14 +645,34 @@ def _compute_pid(tile_id, num_pid_in_group, grid_m, GROUP_M: tl.constexpr, NUM_S
...
+        acc = tl.permute(acc, (0, 2, 1))
+        acc0, acc1 = tl.split(acc)
```
eellison, line 652: just asking - would it ever make sense to configure # of subtiles ? do we even want 4 for instance ? E.g. with more expensive epilogue fusion. potentially this would avoid multiple store_outputs if done in a loop.

eellison (commented):
looks good ! would it be possible to test this with @drisspg's recent b200 enablement ?

torch/_inductor/config.py
```diff
@@ -1440,6 +1440,8 @@ class triton:
     # Should TMA store be enable from templates. TODO: Remove once we
     # can autotune over the result.
     enable_template_tma_store = os.environ.get("ENABLE_TEMPLATE_TMA_STORE", "0") == "1"
+    # Use epilogue subtiling. TODO: Remove once we determine the heuristic.
+    enable_epilogue_subtiling = os.environ.get("ENABLE_EPILOGUE_SUBTILING", "0") == "1"
```
eellison: When is this not profitable ?  I guess this is the TODO...

njriasan: Based on my data I think this should be the default. Given that this requires a separate integration component for fusions. I like having a fallback to disable it in case we hit correctness issues, but I will set the default to True if/when I test with the B200 enablement.

Initially I had it disabled by default because I was worried about sufficient testing.

torch/_inductor/kernel/mm.py
njriasan: It could, but I haven't explored it. The general logic would be that it reduces shared memory utilization so we can have a deeper pipeline so maybe. I will add this as a TODO to explore.

eellison (approved):

PR #163151, opened 2025-09-17
remove unnecessary sync point in AveragedModel update

gl3lan:
Summary:
The test `bool(self.n_averaged == 0)` is a CPU/GPU synchronization point that is called for each update.
This test is only meant to know whether the AveragedModel copy has been initialized or not.
This diff introduces a CPU-based variable for that purpose.
When loading from checkpoint we also make sure the parameter is refreshed.

After this fix, each `update_parameter` call is reduced to 6ms from 333ms (98% reduction).

Test Plan:
Test plan from GitHub:
contbuild & OSS CI
Test plan from GitHub:
CI

Rollback Plan:

Differential Revision: D82621539

### DIFF
--- a/test/optim/test_swa_utils.py
+++ b/test/optim/test_swa_utils.py
@@ -76,7 +76,6 @@ class TestSWAUtils(TestCase):
             # Check that AveragedModel is on the correct device
             self.assertTrue(p_swa.device == swa_device)
             self.assertTrue(p_avg.device == net_device)
-        self.assertTrue(averaged_dnn.n_averaged.device == swa_device)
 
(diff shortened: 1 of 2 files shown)


janeyx99 (commented):
Wait weren;t we going to land the other solution with the module hook?

janeyx99:
@gl3lan DCP adding an FC constraint is not okay here and the new DCP does not have this behavior! I much prefer the correct fix to this problem (vs adding yet newer state to work around unideal state), so I would propose the following:
- change the internal code where this breaks to ignore loading the entry and land the original change
- move the internal code to use the newer DCP when it's ready

PR #163152, opened 2025-09-17
[SymmMem] Fix NVSHMEM plugin + Triton 3.5

kwen2501:
* #163194
* __->__ #163152

1. The dispatch signatures defined in `core.extern_elementwise` call must match the C signature of the NVSHMEM functions, in particular the dtypes. Otherwise, there would be weird errors, such as IMA or hang. When matched, most of time the NVSHMEM device function will be inlined into the generated PTX. When not matched, it is represented as a function call in the PTX (not sure if it is the function call that goes wrong).

2. When calling the `core.extern` wrappers from the `triton.jit` kernels, the input must be cast to match the signatures defined in 1, e.g. via `nbytes.to(tl.int64)`. Otherwise, Triton will report a key error when searching for such kernel.

### DIFF
--- a/torch/distributed/_symmetric_memory/_nvshmem_triton.py
+++ b/torch/distributed/_symmetric_memory/_nvshmem_triton.py
@@ -1,3 +1,4 @@
+import logging
 import os
 import subprocess
 import sysconfig
@@ -5,6 +6,9 @@ from typing import Any, Optional
 
 import torch.distributed as dist
 from torch.utils._triton import has_triton
+
+
+logger = logging.getLogger(__name__)
 
 
 def _find_nvshmem_device_library() -> str:
@@ -98,7 +102,7 @@ def enable_triton(lib_dir: Optional[str] = None) -> dict[str, str]:
             key = kwargs["key"]
             device = kwargs["compile"]["device"]
             jit_function = kwargs["fn"].jit_function
-            kernel_cache, *rest = jit_function.device_caches[device]
+            kernel_cache = jit_function.device_caches[device][0]
             kernel = kernel_cache.get(key, None)
             if kernel is not None:
                 kernel.run
@@ -1006,9 +1010,11 @@ if has_triton():
             _semantic=_semantic,
         )
 
+    # Utility for inspecting Triton kernels
+
     triton_kernels: dict = {}
 
-    def log_triton_kernel(kernel) -> None:  # type: ignore[no-untyped-def]
+    def _log_triton_kernel(kernel) -> None:  # type: ignore[no-untyped-def]
         import atexit
         import tempfile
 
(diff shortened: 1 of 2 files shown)


torch/distributed/_symmetric_memory/_nvshmem_triton.py
```diff
@@ -97,7 +98,7 @@ def nvshmem_init_hook(*args, **kwargs) -> None:  # type: ignore[no-untyped-def]
             key = kwargs["key"]
             device = kwargs["compile"]["device"]
             jit_function = kwargs["fn"].jit_function
-            kernel_cache, _, _, _ = jit_function.device_caches[device]
+            kernel_cache, *rest = jit_function.device_caches[device]
```
ngimel:
```suggestion
            kernel_cache = jit_function.device_caches[device][0]
```

torch/distributed/_symmetric_memory/_nvshmem_triton.py
```diff
@@ -980,3 +1005,25 @@ def reduce_extern_wrapper(
...
+    def log_triton_kernel(kernel) -> None:  # type: ignore[no-untyped-def]
+        import atexit
+        import tempfile
+
+        if dist.is_initialized() and dist.get_rank() != 0:
+            return
+
+        def on_exit() -> None:
+            print("PTX files:")
+            for kernel in triton_kernels:
```
ngimel, line 1026: should it be triton_kernels.keys()?

torch/distributed/_symmetric_memory/_nvshmem_triton.py
```diff
@@ -980,3 +1005,25 @@ def reduce_extern_wrapper(
             is_pure=False,
             _semantic=_semantic,
         )
+
+    triton_kernels: dict = {}
+
+    def log_triton_kernel(kernel) -> None:  # type: ignore[no-untyped-def]
```
ngimel: _log_triton_kernel

ngimel (approved):
Thanks for the fix!

torch/distributed/_symmetric_memory/_nvshmem_triton.py
```diff
@@ -980,3 +1005,25 @@ def reduce_extern_wrapper(
...
+        import atexit
+        import tempfile
+
+        if dist.is_initialized() and dist.get_rank() != 0:
+            return
+
+        def on_exit() -> None:
+            print("PTX files:")
```
Skylion007: Probably shouldn't be print? Use logger.logs?

torch/distributed/_symmetric_memory/_nvshmem_triton.py
```diff
@@ -980,3 +1005,25 @@ def reduce_extern_wrapper(
...
+            for kernel in triton_kernels:
+                f = tempfile.NamedTemporaryFile(dir="/tmp", delete=False)
```
Skylion007: Use with context here for safety

kwen2501:
Bad PTX:
https://gist.github.com/kwen2501/0699ac17ceb08477fbe6c524ea7849e8

Good PTX:
https://gist.github.com/kwen2501/f564f409799f0865b05baf52937ea6de

PR #163154, opened 2025-09-17
Fixes floating point exception in torch.nn.PixelShuffle

arkadip-maitra:
Fixes #162251

**Previous Output:**
`Floating point exception (core dumped)`

**Now Output:**
`RuntimeError: upscale factor is too large, (upscale_factor}^2 overflowed: upscale_factor=545460846592`

### DIFF
--- a/aten/src/ATen/native/PixelShuffle.h
+++ b/aten/src/ATen/native/PixelShuffle.h
@@ -1,5 +1,6 @@
 #include <ATen/core/Tensor.h>
 #include <c10/util/Exception.h>
+#include <c10/util/safe_numerics.h>
 
 namespace at::native {
 
@@ -11,8 +12,10 @@ inline void check_pixel_shuffle_shapes(const Tensor& self, int64_t upscale_facto
               "pixel_shuffle expects a positive upscale_factor, but got ",
               upscale_factor);
   int64_t c = self.size(-3);
-  int64_t upscale_factor_squared = upscale_factor * upscale_factor;
-  TORCH_CHECK(c % upscale_factor_squared == 0,
+  uint64_t upscale_factor_squared;
+  TORCH_CHECK(!(c10::mul_overflows(upscale_factor, upscale_factor, &upscale_factor_squared)), 
+        "upscale factor is too large, (upscale_factor}^2 overflowed: upscale_factor=", upscale_factor);
+  TORCH_CHECK(c % static_cast<int64_t>(upscale_factor_squared) == 0,
               "pixel_shuffle expects its input's 'channel' dimension to be divisible by the square of "
               "upscale_factor, but input.size(-3)=", c, " is not divisible by ", upscale_factor_squared);
 }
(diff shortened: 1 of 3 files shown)


aten/src/ATen/native/PixelShuffle.h
```diff
@@ -11,8 +12,10 @@ inline void check_pixel_shuffle_shapes(const Tensor& self, int64_t upscale_facto
               "pixel_shuffle expects a positive upscale_factor, but got ",
               upscale_factor);
   int64_t c = self.size(-3);
-  int64_t upscale_factor_squared = upscale_factor * upscale_factor;
-  TORCH_CHECK(c % upscale_factor_squared == 0,
+  uint64_t upscale_factor_squared;
+  TORCH_CHECK(!(c10::mul_overflows(upscale_factor, upscale_factor, &upscale_factor_squared)),
```
Skylion007: TORCH_CHECK_VALUE to make this raise a ValueException

arkadip-maitra:
@Skylion007 thanks for the review. i added the suggested change.

arkadip-maitra:
@Skylion007 can you check and approve please. thanks

cyyever (approved):

cyyever:
But you need others' approval to merge. 
@Skylion007 take a look?

aten/src/ATen/native/PixelShuffle.h
```diff
@@ -11,8 +12,10 @@ inline void check_pixel_shuffle_shapes(const Tensor& self, int64_t upscale_facto
               "pixel_shuffle expects a positive upscale_factor, but got ",
               upscale_factor);
   int64_t c = self.size(-3);
-  int64_t upscale_factor_squared = upscale_factor * upscale_factor;
-  TORCH_CHECK(c % upscale_factor_squared == 0,
+  uint64_t upscale_factor_squared;
+  TORCH_CHECK_VALUE(!(c10::mul_overflows(upscale_factor, upscale_factor, &upscale_factor_squared)),
+        "upscale factor is too large, (upscale_factor)^2 overflowed: upscale_factor=", upscale_factor);
+  TORCH_CHECK(c % static_cast<int64_t>(upscale_factor_squared) == 0,
```
albanD: Wouldn't the uint64 -> int64 cast here be able to overflow as well?

arkadip-maitra: Yes, there are some particular values where this will fail. An option is to add a double check here if upscale_factor_squared is less than the int64 max value, since the mul_overflows returns the value in uint64.

Adding that check

arkadip-maitra: or better i am adding check to see if the upscale_factor given is < int64 max / 2. That's a lot cleaner

cyyever: You can check `c>=0` and cast `c` to uint64_t instead. Actually c must not be negative here or we would have a big trouble.

arkadip-maitra: c = self.size(-3) so its always positive.  the overflow in this implementation might happen for particular values of upsample_factor which when squared if it falls between int64 max - uint64 max. so its better to check this instead:

```
  TORCH_CHECK_VALUE(upscale_factor <= std::numeric_limits<decltype(upscale_factor)>::max() / upscale_factor,
                    "upscale factor is too large, (upscale_factor)^2 overflowed: upscale_factor=", upscale_factor);
```

albanD (approved):
Thanks for the update!

PR #163160, opened 2025-09-17
[ncclx][c10d] check if process group nccl is using nccl

riklas:
Test Plan:
OSS CI

Rollback Plan:

Differential Revision: D82639915


torch/csrc/distributed/c10d/init.cpp
```diff
@@ -3355,8 +3358,15 @@ options :class:`~torch.distributed.ProcessGroupNCCL.Options`).
...
                 return std::make_tuple(NCCL_MAJOR, NCCL_MINOR, NCCL_PATCH);
               })
-          .def_static("get_runtime_nccl_version", [] {
-            return ::c10d::getNcclVersionTuple();
+          .def_static(
+              "get_runtime_nccl_version",
+              [] { return ::c10d::getNcclVersionTuple(); })
+          .def_static("is_nccl_available", [] {
+#if defined(NCCL_H_) && !defined(IS_NCCLX)
+            return true;
```
Skylion007, line 3366: indentation nit

torch/csrc/distributed/c10d/init.cpp
```diff
@@ -2498,9 +2499,9 @@ communication mechanism.
             "barrier",
               [](const c10::intrusive_ptr<::c10d::ProcessGroup>& self,
                 std::optional<std::chrono::milliseconds> timeout) {
-                    ::c10d::BarrierOptions opts;
-                    opts.timeout = timeout.value_or(::c10d::kUnsetTimeout);
-                    return self->barrier(opts);
+                ::c10d::BarrierOptions opts;
```
Skylion007, line 2502: Why this indentation?

torch/csrc/distributed/c10d/init.cpp
riklas: this was the result of applying lintrunner. i previously had a lintrunner failure

torch/csrc/distributed/c10d/init.cpp
riklas: ditto lintrunner formatting

torch/csrc/distributed/c10d/init.cpp
Skylion007: Can we make sure the cpp files have been opted into clang-format then from the .lintrunner.toml file?

torch/csrc/distributed/c10d/init.cpp
```diff
@@ -3355,8 +3358,15 @@ options :class:`~torch.distributed.ProcessGroupNCCL.Options`).
               [] {
                 return std::make_tuple(NCCL_MAJOR, NCCL_MINOR, NCCL_PATCH);
               })
-          .def_static("get_runtime_nccl_version", [] {
-            return ::c10d::getNcclVersionTuple();
+          .def_static(
+              "get_runtime_nccl_version",
+              [] { return ::c10d::getNcclVersionTuple(); })
```
Skylion007, line 3363: Why not just a reference binding here instead of a temporary lambda? Just make the arg "&::c10d::getNcclVersionTuple" instead of the lambda

torch/csrc/distributed/c10d/init.cpp
riklas: yeah the cpp file's opted into clangformat https://github.com/riklas/pytorch/blob/export-D82639915/.lintrunner.toml#L87

i also tried to reformat with `lintrunner --take CLANGFORMAT` and it produces the same formatting

torch/csrc/distributed/c10d/init.cpp
riklas: makes sense. im going to put out another PR which improves these get_build/runtime version apis because i noticed they're declared as staticmethods but they take a self parameter. since this PR is about checking if nccl is available mind if i include these changes in the next one?

torch/csrc/distributed/c10d/init.cpp
```diff
@@ -2437,7 +2437,8 @@ communication mechanism.
                 std::optional<std::chrono::milliseconds> timeout) {
                 ::c10d::AllToAllOptions opts;
                 opts.timeout = timeout.value_or(::c10d::kUnsetTimeout);
-                return self->alltoall_base(output, input, outputSplitSizes, inputSplitSizes, opts);
+                return self->alltoall_base(
+                    output, input, outputSplitSizes, inputSplitSizes, opts);
```
ezyang, line 2441: Please don't randomly reformat lines when making changes, it makes review harder

torch/csrc/distributed/c10d/init.cpp
ezyang: What does `.lintbin/clang-format --version` say on your system

ezyang:
TBH, I'm not really sure this is necessary, as we already do a 'can we import nccl bindings' test to see if NCCL is available, as they are not bound whe it is not.

torch/csrc/distributed/c10d/init.cpp
riklas: 19.1.4. do i need to update it? 

```
.lintbin/clang-format --version

clang-format version 19.1.4 (https://github.com/llvm/llvm-project.git aadaa00de76ed0c4987b97450dd638f63a385bed)
```

PR #163171, opened 2025-09-17
[cuDNN][conv][64-bit] Disable cuDNN for 64-bit depthwise convs again

eqy:
test is breaking, will check if there's an older version that we can enable on to avoid completely dropping support

cc @csarofeen @ptrblck @xwang233 @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168

### DIFF
--- a/test/nn/test_convolution.py
+++ b/test/nn/test_convolution.py
@@ -4083,6 +4083,8 @@ class TestConvolutionNNDeviceType(NNTestCase):
         y = m.to(device=device)(x.to(device=device))
         self.assertEqual(yref, y)
 
+    # TODO(eqy): remove once cuDNN depthwise support is fixed
+    @unittest.expectedFailure
     @skipCUDAIfRocm
     @onlyCUDA
     @largeTensorTest("20GB")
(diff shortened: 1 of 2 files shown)


test/nn/test_convolution.py
```diff
@@ -4040,6 +4040,8 @@ def test_conv3d_64bit_indexing(self, device):
         y = m.to(device=device)(x.to(device=device))
         self.assertEqual(yref, y)
 
+    # TODO(eqy): remove once cuDNN depthwise support is fixed
+    @unittest.expectedFailure
```
Skylion007: @eqy Can we alternatively update CUDNN? Or is not fixed in the latest version?

eqy: It's broken in the latest version, seems to be ok in 9.10.2

test/nn/test_convolution.py
```diff
@@ -4045,22 +4045,33 @@ def test_conv3d_64bit_indexing(self, device):
...
+            x = x.reshape(100, 2, 3280, 3280)
+            x = x.contiguous(memory_format=torch.channels_last)
+            yref = c(x)
+            y = c.to(device=device)(x.to(device=device))
+            self.assertEqual(yref, y, atol=1e-3, rtol=1e-4)
+
+        cudnn_version = torch.backends.cudnn.version()
+
+        # TODO(eqy): Remove this once it is fixed in cuDNN and we can dispatch to it again
+        if cudnn_version > 91000:
```
Skylion007: Wouldn't this be cleaner as a conditional xfail?

ngimel (approved):

malfet (approved):

PR #163173, opened 2025-09-17
[pcache] Generalize testing + All Caches thread-safe

nmacchioni:
Summary:
1. Generalized testing by auto-detecting Cache types and splitting testing by abstract base class
- Now checks that all Cache types are thread-safe
- Will fail tests if any new Cache is added and is untested (for example, any cache with non-str key or non-bytes value)
2. All Caches are thread-safe
- InMemoryCache was the only one not thread-safe, so added a lock for access
- Realized that to implement MultiCache we should just have this requirement.
* Also, OnDiskCache is now a functioning AsyncCache with a default base_dir using Python's tempfile.gettempdir, i.e. OnDiskCache is no longer an abstract cache class

Test Plan:
```
[nmacchioni@*** / ()]$ buck test fbcode//mode/opt caffe2/test/inductor:pcache
Tests finished: Pass 28. Fail 0. Fatal 0. Skip 0. Build failure 0
[nmacchioni@*** / ()|remote/fbcode/warm_gpu_od_stable...)]$
```

Rollback Plan:

Differential Revision: D82660240

### DIFF
--- a/torch/_inductor/pcache.py
+++ b/torch/_inductor/pcache.py
@@ -3,6 +3,8 @@ from __future__ import annotations
 from functools import cached_property
 from os import getenv
 from pathlib import Path
+from tempfile import gettempdir
+from threading import Lock
 from typing import Generic, TYPE_CHECKING, TypeVar
 from typing_extensions import override, Self
 
@@ -79,6 +81,7 @@ class InMemoryCache(Cache[str, bytes]):
         Initialize the in-memory cache.
         """
         self._cache: dict[str, bytes] = {}
+        self._lock: Lock = Lock()
 
     @override
     def get(self: Self, key: str) -> bytes | None:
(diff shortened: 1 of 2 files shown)


torch/_inductor/pcache.py
```diff
@@ -239,11 +244,8 @@ def base_dir(self: Self) -> Path:
 
         Returns:
             Path: The base directory for the on-disk cache.
-
-        Raises:
-            NotImplementedError: If not implemented by a subclass.
         """
-        raise NotImplementedError
+        return Path(gettempdir())
```
masnesral, line 248: What's going on here? I'm expecting you to return an empty cache object, not a Path?

masnesral: Lol, sorry. Looking at wrong method

masnesral (approved):

PR #163182, opened 2025-09-17
[inductor] can_free check against graph outputs directly instead of outputNode as users

xuanzhang816:
In the example in unittest, `x**2` are offloaded to CPU and thus should be deleted right afterwards via `del buf0`. However, there is no `del buf0` in the generated code, as shown below:

```
    def call(self, args):
        arg0_1, arg1_1 = args
        args.clear()
        assert_size_stride(arg0_1, (32, 32), (32, 1))
        assert_size_stride(arg1_1, (32, 32), (32, 1))
        with torch.cuda._DeviceGuard(0):
            torch.cuda.set_device(0)
            buf0 = empty_strided_cuda((32, 32), (32, 1), torch.float32)
            # Topologically Sorted Source Nodes: [x], Original ATen: [aten.pow]
            stream0 = get_raw_stream(0)
            triton_poi_fused_pow_0.run(arg0_1, buf0, 1024, stream=stream0)
            del arg0_1
            buf1 = empty_strided_cuda((32, 32), (32, 1), torch.float32)
            # Topologically Sorted Source Nodes: [y], Original ATen: [aten.mm]
            extern_kernels.mm(buf0, arg1_1, out=buf1)
            del arg1_1
            # Topologically Sorted Source Nodes: [], Original ATen: []
            buf2 = torch.ops.aten.to.device(buf0, device=device(type='cpu'), dtype=torch.float32)
        buf3 = buf2
        assert_size_stride(buf3, (32, 32), (32, 1), 'torch.ops.aten.to.device')
        assert_alignment(buf3, 16, 'torch.ops.aten.to.device')
        with torch.cuda._DeviceGuard(0):
            torch.cuda.set_device(0)
            buf4 = empty_strided_cuda((32, 32), (32, 1), torch.float32)
            # Topologically Sorted Source Nodes: [z], Original ATen: [aten.mm]
            extern_kernels.mm(buf1, buf1, out=buf4)
            del buf1
        return (buf3, buf4, )
```

With the change, we now have the following generated code, where differences are highlighted.
```
    def call(self, args):
        arg0_1, arg1_1 = args
        args.clear()
        assert_size_stride(arg0_1, (32, 32), (32, 1))
        assert_size_stride(arg1_1, (32, 32), (32, 1))
        with torch.cuda._DeviceGuard(0):
            torch.cuda.set_device(0)
            buf0 = empty_strided_cuda((32, 32), (32, 1), torch.float32)
            # Topologically Sorted Source Nodes: [x], Original ATen: [aten.pow]
            stream0 = get_raw_stream(0)
            triton_poi_fused_pow_0.run(arg0_1, buf0, 1024, stream=stream0)
            del arg0_1
            buf1 = empty_strided_cuda((32, 32), (32, 1), torch.float32)
            # Topologically Sorted Source Nodes: [y], Original ATen: [aten.mm]
            extern_kernels.mm(buf0, arg1_1, out=buf1)
            del arg1_1
            # Topologically Sorted Source Nodes: [], Original ATen: []
            buf2 = torch.ops.aten.to.device(buf0, device=device(type='cpu'), dtype=torch.float32)
        buf3 = buf2
        assert_size_stride(buf3, (32, 32), (32, 1), 'torch.ops.aten.to.device')
        assert_alignment(buf3, 16, 'torch.ops.aten.to.device')
        **del buf0**
        **del buf2**
        with torch.cuda._DeviceGuard(0):
            torch.cuda.set_device(0)
            buf4 = empty_strided_cuda((32, 32), (32, 1), torch.float32)
            # Topologically Sorted Source Nodes: [z], Original ATen: [aten.mm]
            extern_kernels.mm(buf1, buf1, out=buf4)
            del buf1
        return (buf3, buf4, )
```

* __->__ #163182


torch/_inductor/scheduler.py
```diff
@@ -173,9 +173,8 @@ def can_free(self) -> bool:
             self.node
         ):
             return False
-        for use in self.users:
```
eellison, line 176: I want to be careful about this change since we have all these other reasons we add OutputNode: https://github.com/pytorch/pytorch/blob/ebfc87e303559d14f5a7ab536b26118f8a7ab5f9/torch/_inductor/scheduler.py#L2701-L2728.

The bug is actually that we consider `torch.ops.aten.to.device` to be aliasing. Couple things:

- we should fix this for torch.ops.aten.to.device within compile.

Secondly - for actual operators that are aliasing and return a new tensor, theres no particular reason why we need to forward along the output dep.

E.g.
```
def foo():
    a = op()
    b = aliasing_op(a)
    return b
```
It's okay to delete intermediary aliases
```
def foo():
    a = op()
    b = aliasing_op(a)
    del a
    return b
```

Although we would need to be careful to not then reuse the a buffer... This only matters to the extent that aliasing annotations are incorrect. Potentially we could trust the fake impl more than the annotation. 

Cc @zou3519 , @ezyang, @bdhirsh for thoughts.

Edit: jk, this is because this is a manual pass which inserted `to.device` instead of `prims.device_put`.

eellison (commented):
We should investigate incorrect aliasing assumptions. I think there is a general issue that the pytorch annotations are conservative - may-alias instead of must-alias. Potentially it would be more accurate to consider the actual fake impl.

torch/_inductor/scheduler.py
eellison: I have seen some user ops where the annotation was correct and the implementation was incorrect. where here, prim.device the annotation is incorrect and the meta is correct for this instance. in this case we could conservatively disable buffer reuse, but still delete the original alias.

test/inductor/test_memory.py
```diff
@@ -462,6 +462,53 @@ def replace_foreach(gm):
...
+                op="call_function", target=torch.ops.aten.pow.Tensor_Scalar
+            )[0]
+            mm_nodes = graph.find_nodes(
+                op="call_function", target=torch.ops.aten.mm.default
+            )
+            output_node = graph.find_nodes(op="output")[0]
+
+            with graph.inserting_after(mm_nodes[0]):
+                pow_1_on_cpu = graph.call_function(
+                    torch.ops.aten.to.device,
```
eellison, line 490: I guess the only reason this is incorrect here is because `to.device` is maybe-aliasing. if it had been traced through, this would have been correct, with a non-aliasing node `torch.ops.prims.device_put.default`

torch/_inductor/scheduler.py
eellison: Note that this was a similar reason why the custom op wasn't deleted: https://github.com/pytorch/pytorch/pull/163227. cc @BoyuanFeng 

I suppose we could assert on aliasing.

xuanzhang816: Confirming that using `prims.device_put` actually fixes the my issue that buffers are not deleted without this change.

PR #163183, opened 2025-09-17
[torchrec][LocalShardsWrapper] Implement tensor padding for local shards wrapper

jeffkbkim:
Differential Revision: D82663766

This patch implements the constant padding functionality (aten.constant_pad_nd.default) for LocalShardsWrapper. The method applies constant padding to the local shards based on the provided padding specification.

Padding may be used to resize tensors to the same size in order to perform an all gather. LocalShardsWrapper, a wrapper around a DTensor's local_tensor, can now support padding across its local shards.

Depending on the sharding type (RW, CW), the padding on [left, right, top, bottom] directions will be either applied to the first/last shard, or all local shards.

### DIFF
--- a/test/distributed/tensor/test_shards_wrapper.py
+++ b/test/distributed/tensor/test_shards_wrapper.py
@@ -1,0 +1,366 @@
+# Copyright (c) Meta Platforms, Inc. and affiliates
+# Owner(s): ["oncall: distributed"]
+import torch
+from torch.distributed.tensor._shards_wrapper import LocalShardsWrapper
+from torch.testing._internal.common_utils import run_tests, TestCase
+
+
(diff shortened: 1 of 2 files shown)


ezyang (changes requested):
If we're doing this here, the tests need to live in this repo. But I also need some more context on this change.

jeffkbkim:
> @ezyang Thanks for the response. I have updated the PR description.
> 
> I do have a test file that I can move - where would you recommend the test file to be located?
> 
> I am not too familiar of this codebase and was unable to find an appropriate directory.

@ezyang I have added a new test under test/distributed/tensor/test_shards_wrapper.py

ezyang:
I don't know if we actually have a reviewer for this code

PR #163185, opened 2025-09-17
[CP] Introduce flex_cp_forward custom op for FlexAttention CP

fegin:
* #165039
* __->__ #163185
* #164500

The custom op will fetch the required K and V. Currently, the forward pass is just an all-gather, and the backward pass is a reduce-scatter.  While the logic is the same as all_gather_tensor_autograd, the custom op avoids the Autograd warning that wait_tensor() is registered to autograd.

For the next step, we should explore how to interpolate the required communication based on the information from BlockMask.

cc @H-Huang @awgu @wanchaol @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci


XilunWu (approved):

torch/distributed/tensor/experimental/_cp_custom_ops.py
```diff
@@ -0,0 +1,88 @@
+from typing import Any
+
+import torch
+import torch.distributed._functional_collectives as funcol
+import torch.distributed.distributed_c10d as c10d
+
+
+@torch.library.custom_op("cplib::flex_cp_forward", mutates_args=())
```
tianyu-l: Looking at the name `flex_cp_forward`, I would expect it to perform both communication and computation.
I wonder if you have considered other names, e.g. `flex_cp_forward_comms`.

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1193,10 +1196,16 @@ class AttentionType(Enum):
...
+    def __init__(
+        self,
+        seq_dim: int,
+        attention_type: AttentionType,
+        dispatch_mode: _DispatchMode = _DispatchMode.MODULE_WRAPPER,
+    ) -> None:
         super().__init__()
         self.seq_dim = seq_dim
         self.attention_type = attention_type
+        self.dispatch_mode = dispatch_mode
```
tianyu-l: why do you need this?

fegin: Merge conflict, will remove it.

PR #163188, opened 2025-09-17
[aoti] AOTI mingw cross compilation 

yushangdi:
To run this, you need to install `mingw64-gcc-c++` and download windows cuda library toolkit. 

See design doc and demo instructions in https://docs.google.com/document/d/1iDaChqA5nNKkBFTzsdkmoomvQlXHbnlb1Z4yEp7xaJA/edit?tab=t.0

If cross_platform_target is windows, we do the following:

- do not link to `sleef`. This can be improved in the future if we need it. Currently I avoid it because that requires extra setup on the linux side
- Use `mingw64-gcc-c++` to compile
- Use `WINDOWS_CUDA_HOME` instead of `CUDA_HOME` when linking to cuda

```
 python test/inductor/test_aot_inductor_windows.py -k so
 ```
 
 
 Other changes:
 - de-couples compile_standalone config and dynamic link flag
 - create a new aot_inductor_mode config module, which is used to control configs in aot_inductor.

### CODE REVIEW

File: torch/_inductor/codegen/cpp_wrapper_gpu.py [hunk 1]
```diff
@@ -319,7 +319,7 @@ def codegen_inputs(self):
                 )
                 self.prefix.splice(
                     f"""
-                    if ((long({input_name}.data_ptr()) & ({GPU_ALIGN_BYTES} -1)) != 0) {{
+                    if ((reinterpret_cast<std::uintptr_t>({input_name}.data_ptr()) & ({GPU_ALIGN_BYTES} -1)) != 0) {{
```
yushangdi, line 322: has to change this for windows cross-compilation. This should also work for linux.

torch/_inductor/config.py
```diff
@@ -1609,6 +1607,32 @@ class aot_inductor:
     # TODO: should consolidate this flag with compile_standalone
     libtorch_free_headers: Optional[list[str]] = None
 
+    # compile to TorchStandalone
+    compile_with_torchstandalone: int = False
+
+    # If set to "windows", we will compile from WSL and generate a C++ project file for
+    # further Windows native compilation
```
yushangdi: wrong description. I'll change it

torch/csrc/inductor/aoti_runtime/model_base.h
```diff
@@ -1,6 +1,6 @@
 #pragma once
 #ifdef _WIN32
-#include <Windows.h>
+#include <windows.h>
```
yushangdi, line 3: has to use lower case for cross-compilation. windows is not case-sensitive, but linux is

ezyang:
Neat! Who is using AOTI on Windows? Can you show evidence that this is running on our Windows CI? Thanks!

(Not a full review, deferring to AOTI peeps)

yushangdi:
> Neat! Who is using AOTI on Windows? Can you show evidence that this is running on our Windows CI? Thanks!
> 
> (Not a full review, deferring to AOTI peeps)

@ezyang This is for Executorch (aka limited unifed runtime) to use AOTI as a backend for windows. I haven't added this to windows CI yet, but that's next step!

torch/_inductor/config.py
```diff
@@ -1593,21 +1597,35 @@ class aot_inductor:
...
+    # We'll use x86_64-w64-mingw32-gcc to cross-compile a .dll file
+    # If using cuda, you also need to set WINDOWS_CUDNN_HOME env var
+    # to point to windows CUDA toolkit.
+    # Example: WINDOWS_CUDNN_HOME=cuda-windows-base/cuda_cudart/cudart/
+    # The path should contain lib cuda and lib cudart
+    cross_target_platform: Optional[str] = None
+
+    # If link_libtorch is False and cross_target_platform is windows,
+    # a library needs to be provided to provide the shim implementations.
+    aoti_shim_library: Optional[str] = None
```
albanD, line 1616: I wouldn't expect you need this.
You don't need it as a runtime dependency because the runtime that loads the AOTI binary will always have a shim implementation for you.
And at build time, you always have libtorch that you can use as an example lib.

Am I missing something?

torch/_inductor/config.py
```diff
@@ -1593,21 +1597,35 @@ class aot_inductor:
...
 
-    # If None, the default torch headers such as torch/include
-    # will be used. Otherwise, the provided path will be used instead.
-    # This is needed for torchnative to load libtorch-free .so.
-    # Such as [f"{torchnative_dir}/standalone",f"{torchnative_dir}/",].
-    # TODO: should consolidate this flag with compile_standalone
-    libtorch_free_headers: Optional[list[str]] = None
+    # Currently the only valid option is "windows".
+    # We'll use x86_64-w64-mingw32-gcc to cross-compile a .dll file
+    # If using cuda, you also need to set WINDOWS_CUDNN_HOME env var
```
albanD: You meant WINDOWS_CUDA_HOME ?

torch/_inductor/config.py
```diff
@@ -1593,21 +1597,35 @@ class aot_inductor:
     # custom op libs that have implemented C shim wrappers
     custom_op_libs: Optional[list[str]] = None
 
-    compile_standalone: bool = False
```
albanD, line 1609: Not sure what's the BC policy on these configs but I assume changing them is ok?

torch/_inductor/cpp_builder.py
```diff
@@ -1097,6 +1108,11 @@ def _get_torch_related_args(
                 libraries.append("torch_python")
     else:
         libraries_dirs = []
+        if config.aot_inductor.cross_target_platform == "windows":
+            assert config.aot_inductor.aoti_shim_library
+            assert AOTI_SHIM_LIB
```
albanD: You can through nice errors with pointers at what's missing here

torch/_inductor/cpp_builder.py
```diff
@@ -2028,6 +2049,8 @@ def build(self) -> None:
         _create_if_dir_not_exist(_build_tmp_dir)
 
         build_cmd = self.get_command_line()
+        # print(build_cmd)
+        # breakpoint()
```
albanD: old comment

torch/utils/cpp_extension.py
```diff
@@ -1558,20 +1561,27 @@ def library_paths(device_type: str = "cpu", torch_include_dirs=True) -> list[str
         if HIP_HOME is not None:
             paths.append(os.path.join(HIP_HOME, 'lib'))
     elif device_type == "cuda":
-        if IS_WINDOWS:
+        if config.aot_inductor.cross_target_platform == "windows":
             lib_dir = os.path.join('lib', 'x64')
+            if WINDOWS_CUDA_HOME is None:
+                raise RuntimeError("Need to set WINDOWS_CUDA_HOME for windows cross-compilation")
+            assert WINDOWS_CUDA_HOME is not None
```
albanD: nit can be removed

torch/utils/cpp_extension.py
```diff
@@ -1558,20 +1561,27 @@ def library_paths(device_type: str = "cpu", torch_include_dirs=True) -> list[str
         if HIP_HOME is not None:
             paths.append(os.path.join(HIP_HOME, 'lib'))
     elif device_type == "cuda":
-        if IS_WINDOWS:
+        if config.aot_inductor.cross_target_platform == "windows":
```
albanD: It would be cleaner to add a concept of cross compilation to this and not leak the inductor info.
@msaroufim do you think it is worth it? Or should we just do this one for now?

test/inductor/test_aot_inductor_windows.py
```diff
@@ -0,0 +1,70 @@
...
+
+            with tempfile.TemporaryDirectory() as tmpdir:
+                with zipfile.ZipFile(package_path, "r") as zf:
+                    zf.extractall(tmpdir)
+
+
+if __name__ == "__main__":
+    from torch._inductor.test_case import run_tests
+
+    if HAS_GPU or sys.platform == "darwin":
```
albanD: Don't you never have a GPU on macos? :D

torch/_inductor/config.py
yushangdi: If I use libtorch as an example lib, then the .so will have a dependency on libtorch (it won't load unless it see a libtorch), and we'll need to use the `patchelf` hack to remove that dependency later. 

Moreover, we'll need to link against the windows version of torch.lib (not libtorch.so!), which requires a build of torch on windows. Seems easier to require an executorch.lib?

torch/_inductor/config.py
yushangdi: oh yeah, typo, good catch.

torch/_inductor/config.py
yushangdi: no one is using this config except me and Bin for testing now. Should be ok.

torch/utils/cpp_extension.py
yushangdi: I added a cross_target_platform arg now.

torch/_inductor/config.py
albanD: Ho yes, you don't have a torch.lib on hand indeed.
That sounds good to me then! We can iterate on thi later if needed.

test/inductor/test_aot_inductor_windows.py
```diff
@@ -0,0 +1,69 @@
...
+
+    def forward(self, x):
+        x = self.fc1(x)
+        x = self.relu(x)
+        x = self.fc2(x)
+        x = self.sigmoid(x)
+        return x
+
+
+class TestAOTInductorWindowsCrossCompilation(TestCase):
```
albanD, line 29: I am really not sure how to test this cross compilation workflow in CI.

@seemethere for context: we build a binary on linux with mingw and then run it on windows.
any recommendation on how to test that?

The only thing I can think of would be to have a two workflows one after the other, but that might be a lot of setup work?

albanD (commented):
Sounds good to me!
We can follow up with Eli on the testing.

I will let Bin approve this one though since I might have missed some things here.

test/inductor/test_aot_inductor_windows.py
yushangdi: If we have WSL on the windows CI, we can build this in WSL, and then run the rest on windows. This is how I'm testing it locally as well.

torch/_inductor/cpp_builder.py
```diff
@@ -66,6 +66,7 @@ def use_global_cache() -> bool:  # type: ignore[misc]
 _IS_LINUX = sys.platform.startswith("linux")
 _IS_MACOS = sys.platform.startswith("darwin")
 _IS_WINDOWS = sys.platform == "win32"
+AOTI_SHIM_LIB = os.environ.get("AOTI_SHIM_LIB")  # used for AOTI cross-compilation
```
desertfire: I feel this naming does not distinguish from `config.aot_inductor.aoti_shim_library`.

torch/_inductor/config.py
```diff
@@ -1593,21 +1597,35 @@ class aot_inductor:
...
+# a convenient class that automatically sets a group of the configs in aot_inductor
+# it should only control the flags in aot_inductor.
+# it should not do anything else.
+class aot_inductor_mode:
+    # dynamic_linkage=False
+    # link_libtorch=False
+    # package_cpp_only=True
+    # embed_kernel_binary=True
+    # emit_multi_arch_kernel=True
+    compile_standalone: bool = False
```
desertfire, line 1629: Just to confirm, this is for TorchNative, not ExecuTorch?

yushangdi: correct

torch/_inductor/cpp_builder.py
yushangdi: Maybe `AOTI_SHIM_LIBRARY_PATH`?

desertfire: Better. Also why don't we have a corresponding config for this one. Ad hoc environment variable makes it hard for users to discover.

yushangdi: make sense. I updated this to a config now.

desertfire (approved):

PR #163195, opened 2025-09-17
[testing] Add test owner labels for some tests (some core stuff?)

clee2000:
I am trying to give some test files better owner labels than `module: unknown`.  I am not sure them, but they seem pretty reasonable

### DIFF
--- a/test/test_pruning_op.py
+++ b/test/test_pruning_op.py
@@ -1,4 +1,4 @@
-# Owner(s): ["module: nn"]
+# Owner(s): ["oncall: quantization"]
 
 import hypothesis.strategies as st
 from hypothesis import given
--- a/test/test_pruning_op.py
+++ b/test/test_pruning_op.py
@@ -1,4 +1,4 @@
-# Owner(s): ["module: unknown"]
+# Owner(s): ["module: nn"]
 
 import hypothesis.strategies as st
 from hypothesis import given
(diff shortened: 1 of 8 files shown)


janeyx99 (approved):
I'd recommend adding the subscribers of each of these modules as the reviewers for this PR. I support changing unknown to known, but I wouldn't land this until the module owners agree.

test/test_pruning_op.py
```diff
@@ -1,4 +1,4 @@
-# Owner(s): ["module: unknown"]
+# Owner(s): ["module: nn"]
```
mikaylagawarecki:
```suggestion
# Owner(s): ["oncall: quantization"]
```

jgong5 (approved):

test/test_pruning_op.py
```diff
@@ -1,4 +1,4 @@
-# Owner(s): ["module: unknown"]
+# Owner(s): ["oncall: quantization"]
```
Xia-Weiwen, line 1: Not sure if this should be classified as quantization. CC @jerryzh168

test/test_static_runtime.py
```diff
@@ -1,4 +1,4 @@
-# Owner(s): ["module: unknown"]
+# Owner(s): ["module: cpp"]
```
albanD: This is actually `oncall: jit`

albanD (approved):
Thanks!

test/test_static_runtime.py
clee2000:
```suggestion
# Owner(s): ["oncall: jit"]
```

jgong5 (approved):

PR #163198, opened 2025-09-17
[inductor] rename kernel inputs scalars to kwargs

coconutruben:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #163201
* #163200
* #163199
* __->__ #163198

# why

- evolve the kernel inputs to be nodes + kwargs
- clearer that this is not limited to just scalars

- this is 1/n of removing the extra_kwargs and overriders from
  get_template_configs in favor of simplifying it to just being
  - template
  - kernel inputs
  - heuristic generated kwargs

# what

- rename scalars related stuff to kwargs
- replace invocations in addmm related configs

# testing

```
python3 -bb -m pytest test/inductor/test_max_autotune.py -v -k "_addmm"
```

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @jataylo @chenyang78

PR #163199, opened 2025-09-17
[inductor][heuristics] move out_dtype into kernel_inputs

coconutruben:
* #163201
* #163200
* __->__ #163199
* #163198

# why

- step closer towards simplifying the interface to not need kwargs
  overriders

# what

- avoid passing out_dtype through kwargs_overriders
- extract where needed for ATen templates through kernel_inputs

# testing

```
```

### DIFF
--- a/torch/_inductor/kernel/mm.py
+++ b/torch/_inductor/kernel/mm.py
@@ -1251,9 +1251,7 @@ def tuned_scaled_mm(
 
     if use_aten_gemm_kernels():
         templates_to_use.append(aten__fp8_mm)
-        kwarg_overrides[aten__fp8_mm.uid] = dict(
-            out_dtype=out_dtype, use_fast_accum=use_fast_accum
-        )
+        kwarg_overrides[aten__fp8_mm.uid] = dict(use_fast_accum=use_fast_accum)
 
     _, is_nonzero = _is_static_problem(layout)
 
--- a/torch/_inductor/kernel_inputs.py
+++ b/torch/_inductor/kernel_inputs.py
@@ -187,6 +187,15 @@ class KernelInputs(ABC):
             The output dtype
         """
 
+    def out_dtype_explicit(self) -> Optional[torch.dtype]:
+        """
+        Get the explicitly set output dtype without any inference
+
+        Returns:
+            The explicit output dtype or None if not set
+        """
+        return self._out_dtype
+
     def kwargs(self) -> dict[str, Union[float, int, bool]]:
         """
         Get the kwargs values for all input nodes.
@@ -286,12 +295,6 @@ class MMKernelInputs(KernelInputs):
         return (m, n, k)
 
     def out_dtype(self) -> torch.dtype:
-        """
-        Get the output dtype, whether passed in or inferred from the nodes
-
-        Returns:
-            The output dtype
-        """
         if self._out_dtype is not None:
             return self._out_dtype
         return self.mat1mat2()[0].get_dtype()
--- a/torch/_inductor/template_heuristics/aten.py
+++ b/torch/_inductor/template_heuristics/aten.py
@@ -21,12 +21,9 @@ if TYPE_CHECKING:
 # These are all labeled as device type None to indicate that they
 # are valid for all device types
 @register_template_heuristic(aten_mm.uid, None)
-@register_template_heuristic(aten__fp8_mm.uid, None)
 @register_template_heuristic(aten__int_mm.uid, None)
 @register_template_heuristic(aten_bmm.uid, None)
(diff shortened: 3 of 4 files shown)


torch/_inductor/kernel/mm.py
```diff
@@ -1251,9 +1251,7 @@ def tuned_scaled_mm(
 
     if use_aten_gemm_kernels():
         templates_to_use.append(aten__fp8_mm)
-        kwarg_overrides[aten__fp8_mm.uid] = dict(
-            out_dtype=out_dtype, use_fast_accum=use_fast_accum
-        )
+        kwarg_overrides[aten__fp8_mm.uid] = dict(use_fast_accum=use_fast_accum)
```
eellison, line 1254: Can you explain what this is doing?

torch/_inductor/kernel_inputs.py
```diff
@@ -187,6 +187,15 @@ def out_dtype(self) -> torch.dtype:
             The output dtype
         """
 
+    def out_dtype_explicit(self) -> Optional[torch.dtype]:
```
eellison, line 190: Do we need both `out_dtype_explicit` and `out_dtype` ?

torch/_inductor/template_heuristics/aten.py
```diff
@@ -44,8 +41,23 @@ def _get_template_configs_impl(
         yield dict()
 
 
-# None here indicates that this is valid for all device types on that op
-# Note (None, op) takes precedence over (device_type, None)
+@register_template_heuristic(aten_bmm_dtype.uid, "cuda")
+@register_template_heuristic(aten__fp8_mm.uid, None)
+class ATenOutDtypeConfigHeuristics(ATenConfigHeuristics):
+    def get_extra_kwargs(
```
eellison, line 47: This feels harder to for me to follow, that the extra kwargs/dtype is now not in line with the actual op lowering.

eellison (commented):
I am having a bit of a hard time following this

torch/_inductor/kernel/mm.py
PaulZhang12: ^ Yeah I am a bit confused, out_dtype is only used for bmm?

eellison (commented):
I'm not sure why we want to remove kwargs. Ideally we would not need to add an abstract base function for every particular kwarg.

torch/_inductor/template_heuristics/aten.py
coconutruben: I agree it's harder to follow, though it reduces the places of where a kwarg can now come from, which is they can only come from
- the config params 
- the extra kwargs (op wide, template specific kwarg)
we can/should probably move the templates closer to these heuristics so you can see it clearer which parts this maps to

torch/_inductor/kernel_inputs.py
coconutruben: we do not, though we do need a mechanism to say _do we have an explicit out_dtype or is it just inferred_. Would you suggest
- having a get/property to see whether it's set or inferred
- taking an extra bool whether to infer or just to return None

torch/_inductor/kernel/mm.py
coconutruben: out_dtype is also used for aten_fp8, and this is just moving it into the heuristic. So that kwarg is now coming from the call to `get_extra_kwargs` rather than coming from the overriders.

torch/_inductor/template_heuristics/aten.py
eellison: Is there a way to put this logic more inline with the lowering ? I think we should optimize for readability.

coconutruben: **option a:** remove the `get_extra_kwargs` logic form the heuristics entirely and only use the kwarg passer I'm trying to remove here. This might make it more error prone when editing/adding a template (every site that uses the template needs to be adjusted) but easier to read as you know inside the op lowering which which template is receiving which extra kwargs. this would affect
- addmm/scaledmm for epilogues
- tma for tma workspace arg
- and potentially ALLOW_TF32 for all the triton templates

**option b:** keep this as is (remove the extra kwargs) and move the templates next to their heuristics and params, asking readers to see which template in their op their care about and then go read that template/heuristic to see which kwargs are passed how

I favor option b, because I worry something like tma and allow_tf32 needing to be specified every time (mm, addmm) is going to invite people to write small helper functions that do it correctly once, and then we have some proliferation again of where which kwargs are coming from, and because removing kwargs right now isn't going to be painful, but if we build out more things/expand on this InductorChoices interface, changing the method signature will be more painful down the road.

but up to you, I'll implement either option that's fine with you, I just want to get rid of the possibility of there being two sources for extra_kwargs

PR #163200, opened 2025-09-17
[inductor][heuristics] move use_fast_accum into kernel_inputs

coconutruben:
* (to be filled)

# why

- remove last user of kwarg overriders

# what

- all scaled_mm kernels (triton, aten) that need this, are now getting
  it from the kwargs inside kernel_inputs

# testing

```
```


mlazos (approved):
looks good, but I'm not sure if we previously had an option where use_fast_accum could be None indicating you'd autotune across both true and false. (although to be fair I think it being true would just always be faster so not sure if this matters, just something to consider.)

PR #163204, opened 2025-09-17
[CI] Build slimmer torch for SymmMem CI

kwen2501:
* __->__ #163204


malfet (changes requested):
All CI build artifacts are globally cached, so disabling those dependencies should not make any significant change, but will be painful to debug later on

PR #163207, opened 2025-09-17
[DO NOT LAND] Make LSTM exportable

yiming0416:
Summary:
LSTM was not exportable as it failed at `_detect_attribute_assignment` check.

This is because the `_flat_weights` attribute in LSTM is a list of parameters that could be updated by the `_update_flat_weights` method, and this could cause silent fake tensor leakage.

We add `torch.compiler.is_exporting()` check in the `_update_flat_weights` to make sure that `_flat_weights` won't be updated during export so that no fake tensor leakage will happen.

Test Plan:
buck2 run mode/dev-nosan caffe2/test:test_export -- -r test_lstm_export

Rollback Plan:

Differential Revision: D82682791

### DIFF
--- a/test/export/test_export.py
+++ b/test/export/test_export.py
@@ -16026,6 +16026,17 @@ def forward(self, q, k, v):
         ):
             export(Foo(), (torch.randn(1, 33, 256, 128), k, v))
 
+    def test_lstm_export(self):
+        mod = torch.nn.LSTM(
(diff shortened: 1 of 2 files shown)


vadimkantorov:
One other thing:

Is it possible to directly allocate the weights as flattened in the init? And then just create individual parameters as views over that single flat param space? As long as the user doesn't manually tinker with the params, it probably should be no worse than doing single-time flattening in forward?

PR #161224, opened 2025-08-21
[device_mesh] Implement  `_unflatten` on top of CuTe layout bookkeeping

fduwjj:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #163358
* __->__ #161224
* #164510

cc @H-Huang @awgu @wanchaol @fegin @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci

PR #160429, opened 2025-08-12
Simplify internal bookkeeping of DeviceMesh

lw:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #160429

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci

PR #159482, opened 2025-07-30
[DeviceMesh] Add `_unflatten_` api for device mesh to support better UX for some use cases like EP and replicate

fduwjj:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #159482

After some initial feedback on the implementation of `_split`, we realize that we can first implement `_unflatten` for urgent use cases ask for now. And we will do more refactoring and iterations based on the discussions from this PR and this RFC: https://github.com/pytorch/pytorch/issues/159013. We will also ensure that all changes won't cause regression to DTensor's CPU overhead as well.

This PR is trying to:
1. For unflatten, we don't support flatten a unflattened mesh which might not be necessary because by the time when user decide to flatten a unflatten, users might essentially redo the unflatten operations which could make bookeeping complicated to handle and we don't see that use cases for now. (We throw a NotImplementError for now)
3. We need some extra book-keeping for unflatten api, what we want to keep track is which sub-mesh contains the unflattened `dim_name` so that users can slice these dim_names from root mesh as well. And we need to swap the mesh to slice when slicing from root mesh. Also to make sure we don't unflatten same `dim_name` into different sizes, we need to keep the total accumulated numel in the root for that dim_name as well.
4. We want to reuse PGs already created for the same `dim_name`. For the case when a different `dim_name` happens to have different shapes, we will create new PG because with a different name, users might want to use that dimension for different purposes, so we'd better not to reuse. (This assumption can be changed, so I am open to suggestions)
5. Added unit test to two situation: 1. we directly do unflatten on one 2D device mesh. 2. we first create a dummy 1D device mesh and then split into two 3D device mesh.

cc @H-Huang @awgu @wanchaol @fegin @wz337 @wconstab @d4l3k @pragupta

PR #163212, opened 2025-09-17
[DeviceMesh] Introduce CuTe layout into devicemesh code base for internal bookkeeping

fduwjj:
* #163358
* #161224
* #163213
* #163367
* #163288
* __->__ #163212

DeviceMesh essentially is a way to specify how devices interact with each other or device layout. They are all integers but because they can have various shapes and meshes, it make internal bookkeeping internally way more challenging. Currently our internal bookkeeing inside DeviceMesh is not scalable, so in order to support new functions like `_unflatten`, we need to introduce very complicated logics inside DeviceMesh as pointed out per comment (https://github.com/pytorch/pytorch/pull/159482/files#r2256025452). So thanks to @lw 's suggestion and PoC PR (https://github.com/pytorch/pytorch/pull/160429), we realize that by leveraging CuTe layout algebra([ref](https://docs.nvidia.com/cutlass/media/docs/cpp/cute/02_layout_algebra.html)) from Cutlass will greatly simply our internal mechanical bookkeeping for and make the abstraction ops way easier on top of it. So to make things go incrementally, we propose couple steps here https://github.com/pytorch/pytorch/issues/160337#issuecomment-3195106243.

On top of what we have been doing about PyCute we want to continue add methods into the wrapper class so that we can get rank indexes needed for ProcessGroup Creation with a layout object. We also added detailed explanations and comments (thanks to llm) and unit test to show case the code indeed is working as expected. 

More PRs are on the way.

This is a continue of https://github.com/pytorch/pytorch/pull/161016 (originally messed with EasyCLA)


ezyang (approved):
rip my comments lol

fegin (approved):

torch/distributed/_mesh_layout.py
```diff
@@ -59,13 +75,138 @@ def __getitem__(self, i: int) -> "_MeshLayout":
...
+
+        In distributed terms, complement() is often used to derive the "other"
+        rank grouping when splitting processes into 2D meshes.
+
+        For a visualized explanation, see https://x.com/ezyang/status/1962364978393981433/
+        """
         layout = complement(self, world_size)
         return _MeshLayout(layout.shape, layout.stride)
+
+    def member_ranks(self) -> list[int]:
```
lw: Nit: the name "member ranks" might imply that some ranks are members of the layout while others aren't. In reality a layout is just a way of "organizing" all the ranks into some structure. Perhaps a more fitting name for this function could be "global_ranks_of_first_group" or "global_ranks_of_group_with_rank_0". How does that sound?

lw (approved):
Nice! Thanks!

torch/distributed/_mesh_layout.py
fduwjj: global_ranks is one concept with a world_size. Otherwise, why do we call it "global". So I tend to just call it ranks. Also per comment from Will here: https://github.com/pytorch/pytorch/pull/161016#discussion_r2314420367. Although he also mentioned first group, he also pointed that we should not mention process group yet. maybe "ranks_from_zero"?

fduwjj: ok as discussed offline, let's call it `all_ranks_from_zero`

PR #159371, opened 2025-07-29
Allow controlling PG backend and options via init_device_mesh

lw:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #159371

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta

PR #163213, opened 2025-09-17
[DeviceMesh] Simplifying internal bookkeeping with CuTe layout

fduwjj:
* #164510
* __->__ #163213

We want to refactor the internal bookkeeping of DeviceMesh so that:
Simply the bookkeeping logics and make it generic enough so that it is easy to support new transformations like flatten noncontiguous dim, reshape and unflatten. (We leveraged the CuTe layout). This new layout also let us handle non-contiguous slicing, flatten, transpose possible.

Concretely, in this PR, we do the following:
1. Use the `_MeshLayout` to handle all index operations rather use a map to record mesh dims.
2. Removed `flatten_name_to_root_dims`, because now we can directly get layout from a flattened device mesh.
3. Replaced `_get_slice_mesh_dims` with `_get_slice_mesh_layout`.
4. Use the newly added function `check_overlap` to check layout overlap.
5. Use a new function `to_remapping_tensor` to use layout ranks as indices when the mesh tensor is not representable as CuTe. The reason is that layout acts as a backend of mesh tensor bookkeeping (indexing indices), it needs to be used as indices for remap back to the mesh tensor for new DeviceMesh generation and backend init. For example, in the case of 2K to 4K, the underlying layout is (2K, 1) but the actual value of the mesh tensor is [2K, 2K+1, ....,]. While flattening, slicing, we need to remap the layout back to the new mesh tensor so it maps the actual device allocation. For example, in the 2K to 4K case, if the shape is (1K, 1K) with dim_names ("dp", "tp"). Then when slicing "tp", the mesh tensor should be (2K, 2K+1, ..., 3K-1) or (3K, 3K+1, ... 4K-1). not the global ranks generated from the layout. (1K, 1).

Verified that loss curve is very close for DeepSeekV3 on torchtitan, note that exact same match is challenging because even if we run the baseline twice, the loss curve does not exactly match.

image

The PR looks big indeed but we don't change any existing behavior of DeviceMesh, so it is a pure refactor.

With this refactoring we also enabled the slicing and flatten of non-contiguous dims of a device mesh which is hard to implement without cute layout.

This is a continue of https://github.com/pytorch/pytorch/pull/161106 (original one got messed with EasyCLA)


torch/distributed/device_mesh.py
```diff
@@ -447,6 +417,10 @@ def __init__(
             self.mesh_dim_names = tuple(mesh_dim_names) if mesh_dim_names else None
             if backend_override is None:
                 backend_override = ((None, None),) * self.mesh.ndim
+            # Internal bookkeeping for the device mesh.
+            self._layout = (
+                layout if layout else _MeshLayout(self.mesh.size(), self.mesh.stride())
+            )
```
tianyu-l, line 465: It looks like if `layout` is given, it'll override the fallback `_MeshLayout(self.mesh.size(), self.mesh.stride())`. I wonder if it's good to do some basic check to make sure this is legal, e.g. shape is matched and pass `check_overlap`.

torch/distributed/device_mesh.py
```diff
@@ -308,29 +274,27 @@ def _get_slice_mesh_dims(
...
+            layout_sliced = _MeshLayout(sliced_sizes, sliced_strides)
+            if not layout_sliced.check_overlap():
+                raise RuntimeError(
+                    f"slicing overlapping dim_names {mesh_dim_names} is not allowed"
+                )
 
-            return slice_mesh_dims
+            return layout_sliced
 
         def _get_all_submeshes(
```
tianyu-l, line 336: Is this function ever used, other in test?

torch/distributed/device_mesh.py
```diff
@@ -339,10 +303,10 @@ def _get_all_submeshes(
...
-                -1, device_mesh.mesh.size(mesh_dim)
-            )
-
+            layout = device_mesh._layout[mesh_dim]
+            pg_ranks_by_dim = layout.to_remapping_tensor(
+                device_mesh.mesh, get_world_size()
+            )  # type: ignore[arg-type]
             cur_rank = device_mesh.get_rank()
             res_submeshes = []
             for mesh_1d in pg_ranks_by_dim:
```
tianyu-l, line 349: curious why the element in `pg_ranks_by_dim` is also 1d mesh?

torch/distributed/device_mesh.py
```diff
@@ -352,11 +316,14 @@ def _get_all_submeshes(
...
                 )
-                submesh._dim_group_names = (
+                submesh._dim_group_names = (  # type: ignore[has-type]
                     [device_mesh._dim_group_names[mesh_dim]]  # type: ignore[has-type]
                     if cur_rank in mesh_1d
                     else []
                 )
+                submesh._layout = (
+                    device_mesh._layout if cur_rank in mesh_1d else _MeshLayout(0, 0)
+                )
```
tianyu-l: could you elaborate what this code is doing?

torch/distributed/_mesh_layout.py
```diff
@@ -210,3 +221,141 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
...
+        previous_span = -1
+        previous_stride = -1
+        for size, stride in sorted(self.sizes_and_strides, key=lambda x: x[1]):
+            if previous_stride == stride or stride < previous_span:
+                return False
+            previous_stride = stride
+            previous_span = size * stride
+        return True
+
+    def to_remapping_tensor(
```
tianyu-l: This function seems very useful, although I'm puzzled by what it actually does.
1. the naming `to_remapping_tensor` is not very clear
2. the examples in the docstring didn't explain why they are the desired behaviors

In general as a reader I guess I'm missing contexts.

torch/distributed/device_mesh.py
fduwjj: it is used internally by shampoo. cc: @wz337

torch/distributed/device_mesh.py
```diff
@@ -171,7 +125,7 @@ def create_flatten_mesh(
                 return device_mesh
```
fduwjj, line 142: To answer @ezyang 's comment here in https://github.com/pytorch/pytorch/pull/161106/files#r2357456843.

> I know that it's valid to use the name from mesh_dim_names. But what if the user requests a different name, that is NOT the one in that list? Now that seems like trouble.

So if user pass in a different name in `_flatten`, we will not enter into this if and we will actually create a new device mesh. 

For example, mesh3d= init_device_mesh((2,2,2), ("dp", "cp", "tp"))
so you can do: 
mesh3d["dp", "cp"]._flatten("dp_cp")
mesh3d["dp", "cp"]._flatten("dp_cp_2")
but you cannot do
mesh3d["dp", "cp", "tp"]._flatten("dp_cp")
or
mesh3d["dp", "cp"]._flatten("dp")

torch/distributed/device_mesh.py
fduwjj: For this case, we are getting submesh per dim, so each dim is a 1d mesh? Even with nested one, we have to flatten it because per definition of mesh dim, that dimension only has one mesh.

torch/distributed/device_mesh.py
fduwjj: actually this is to match the logic of L322 because this one will generate all submesh for a given device mesh so even if this submesh does not exist in the current rank. So we set the pg to be None.. Let me add a comment here.

torch/distributed/_mesh_layout.py
fduwjj: `remapping_tensor` is a name we agreed on during the discussion last Monday. but I am open to other proposals. Maybe can you elaborate more on what exact contexts you are missing here?

Basically what we wrote in the comment:
> With this method, the cute layout serves as the backend of indices bookkeeping for the
        mesh tensor when it comes to flatten, unflatten and slicing operations. The actual mesh
        tensor still represents the actual device assignment and ranks.
        
which part sound not clear to you? Or are asking why we want to do this conversion? It is because we need an actual mesh tensor to specify device allocation and also for backend creation.

torch/distributed/device_mesh.py
ezyang: sgtm

torch/distributed/device_mesh.py
fduwjj: Actually after a second thought, we can remove this part, since a mesh maps to a layout, so this line of change is not needed.

torch/distributed/_mesh_layout.py
```diff
@@ -210,3 +221,144 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
             [group_offset + group_rank for group_rank in self.member_ranks()]
             for group_offset in self.complement(world_size).member_ranks()
         ]
+
+    def check_non_overlap(self) -> bool:
```
fegin, line 216: I feel this function alone and its test can be a PR since the meaning of this API itself is pretty straightforward, like coalesce. but it's up to you.

torch/distributed/_mesh_layout.py
```diff
@@ -210,3 +221,144 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
...
+                continue
+            if previous_stride == stride or stride < previous_span:
+                return False
+            previous_stride = stride
+            previous_span = size * stride
+        return True
+
+    def to_remapping_tensor(
+        self,
+        original_mesh_tensor: torch.Tensor,
```
fegin: The current `original_mesh_tensor` must have the same dimensionality as the layout, is this correct? In the future, if we would like to let users pass the device list/tensor into `DeviceMesh`, can it be a one-dimensional list/tensor? So the access is layout->coordinates->indices-> device_list[indices].

torch/distributed/_mesh_layout.py
```diff
@@ -210,3 +221,144 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
...
+                dtype=original_mesh_tensor.dtype,
+            ),
+        ):
+            return tensor
+
+        # This is important because the indices generated by the layout will be larger than the original mesh tensor
+        # when the original mesh tensor does not contain all ranks in the world. So we need to scale the layout's stride
+        # by world_size // mesh_tensor.numel() so that the indices generated by the layout will be within the range of
+        # the original mesh tensor.
+        if original_mesh_tensor.numel() != world_size:
```
fegin: Again, is this because the current DeviceMesh implementation so that layout covers more than the mesh tensor?

torch/distributed/_mesh_layout.py
```diff
@@ -210,3 +221,144 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
...
+        for size, stride in sorted(self.sizes_and_strides, key=lambda x: x[1]):
+            if size == 1:
+                continue
+            if previous_stride == stride or stride < previous_span:
+                return False
+            previous_stride = stride
+            previous_span = size * stride
+        return True
+
+    def to_remapping_tensor(
```
fegin: This API is less intuitive than `check_non_overlap` but I also think it can be a standalone PR as long as we can make the input/out clear enough. Then this PR can solely focus on the DeviceMesh refactor using `_MeshLayout`. But this is still up to your.

torch/distributed/device_mesh.py
fegin: Let's add TODO to these private APIs used by other components. This mostly represents our DeviceMesh should public these use cases in the future.

torch/distributed/device_mesh.py
```diff
@@ -444,6 +398,7 @@ def __init__(
             ] = None,
             _init_backend: bool = True,
             _rank: Optional[int] = None,
+            layout: Optional[_MeshLayout] = None,
```
fegin: Since this is a keyword argument, can we move it a bit earlier, at least earler than `_init_backend` and `_ranks` which look like private arguments or arguments that are going to be deprecated, lol.

torch/distributed/device_mesh.py
```diff
@@ -317,29 +271,27 @@ def _get_slice_mesh_dims(
...
+
+            sliced_sizes = tuple(l.sizes for l in layout_sliced)
+            sliced_strides = tuple(l.strides for l in layout_sliced)
+            # When users sliced dim_names outside from current mesh, we will check whether
+            # there is layout overlap. Eventually we will just directly throw error here because
+            # we will deprecate the slicing of flattened dim_name from root mesh.
+            layout_sliced = _MeshLayout(sliced_sizes, sliced_strides)
+            if not layout_sliced.check_non_overlap():
+                raise RuntimeError(
+                    f"slicing overlapping dim_names {mesh_dim_names} is not allowed"
```
fegin:
```suggestion
                    f"Slicing overlapping dim_names {mesh_dim_names} is not allowed."
```

torch/distributed/device_mesh.py
```diff
@@ -317,29 +271,27 @@ def _get_slice_mesh_dims(
...
+                        device_mesh._layout[device_mesh.mesh_dim_names.index(name)]
                     )
-                curr_idx = next_idx
+                elif name in flatten_name_to_root_layout:
+                    layout_sliced.append(flatten_name_to_root_layout[name])
+
+            sliced_sizes = tuple(l.sizes for l in layout_sliced)
+            sliced_strides = tuple(l.strides for l in layout_sliced)
+            # When users sliced dim_names outside from current mesh, we will check whether
+            # there is layout overlap. Eventually we will just directly throw error here because
```
fegin: Maybe add a TODO.

torch/distributed/_mesh_layout.py
fduwjj: sure, I can split them into a separate PR.

torch/distributed/_mesh_layout.py
fduwjj: > In the future, if we would like to let users pass the device list/tensor into DeviceMesh

We already allow that today. 

> The current original_mesh_tensor must have the same dimensionality as the layout

Same world size otherwise when we do original_mesh_tensor[idx] will collapse.

> can it be a one-dimensional list/tensor?

Yes, that's why I did a flatten in the end before indexing, it does not matter what shape original_mesh_tensor is, the final shape will be based on the given layout.

torch/distributed/_mesh_layout.py
fduwjj: yes, because we still needs to maintain the slicing of flattened tensor from root. That way we will need to keep the original layout not the layout derived from the submesh's mesh tensor.

torch/distributed/_mesh_layout.py
fduwjj: sure, I can split them.

torch/distributed/device_mesh.py
```diff
@@ -65,84 +67,41 @@ def get_current_mesh(self) -> "DeviceMesh":
...
             for mesh_nd in pg_ranks_by_dim:
                 submesh = DeviceMesh(
                     device_mesh.device_type,
                     mesh_nd,
                     mesh_dim_names=submesh_dim_names,
                     _init_backend=False,
+                    _layout=layout,
                 )
                 if cur_rank in mesh_nd:
                     res_submesh = submesh
```
lw: Can you explain this code? I know that it wasn't added by this PR, but I find it odd. Basically, we create a potentially very large number of DeviceMesh objects (e.g., on 1024 GPUs with TP=2, this could be up to 512 DeviceMeshes), only to throw them all away except for one. Why do we do that?

I suspect this means that constructing a DeviceMesh objects must have some global side-effects. Is that so? What kind of effects are we talking about?

lw: Note that in this snipped `init_backend` is False so we're not even creating sub-PGs...

torch/distributed/device_mesh.py
```diff
@@ -180,50 +134,48 @@ def create_flatten_mesh(
...
-            for flatten_dim_in_root in flatten_dims_in_root:
-                remained_dims_in_root.remove(flatten_dim_in_root)
-
-            pg_ranks_by_dim = root_mesh.mesh.permute(
-                *remained_dims_in_root, *flatten_dims_in_root
-            ).reshape(-1, flattened_mesh_dim_size)
-
             cur_rank = root_mesh.get_rank()
+            # Due to the limitation of ProcessGroup api, we need to start from root mesh so that all ranks call the
+            # new_group api to avoid potential hang.
```
lw, line 172: I feel bad that we need to make the DeviceMesh code more complicated in order to deal with these low-level concerns. Moreover, if the backend supports splitting (e.g., NCCL) then all of this isn't even actually needed!

May I suggest that we add a helper function that "splits" the root PG into an arbitrary "tiling" (i.e., N subgroups which cover everything) which abstracts away these complexities by internally deciding whether to call split or do this for loop?

torch/distributed/device_mesh.py
```diff
@@ -52,9 +54,9 @@ def __init__(self) -> None:
                 int, tuple[Optional[str], Optional[C10dBackend.Options]]
             ] = {}
             self.root_to_flatten_mapping: dict[DeviceMesh, dict[str, DeviceMesh]] = {}
-            # Record flatten mesh name to its mesh dim index in root mesh.
-            self.flatten_name_to_root_dims: dict[
-                DeviceMesh, dict[str, tuple[int, ...]]
+            # Record flatten mesh name to its flattened layout in root mesh.
+            self.flatten_name_to_root_layout: dict[
```
lw: Can I ask if you intend to keep these global (thread-local) mappings, or if the plan is to eventually drop them? I still don't understand exactly why we need them, but if they're slated for removal I won't worry about them too much.

torch/distributed/device_mesh.py
```diff
@@ -317,41 +271,41 @@ def _get_slice_mesh_dims(
...
-                        "Mesh dim indices should be in ascending order."
+            layout_sliced = []
+            for name in mesh_dim_names:
+                if name in device_mesh.mesh_dim_names:
+                    layout_sliced.append(
+                        device_mesh._layout[device_mesh.mesh_dim_names.index(name)]
                     )
-                curr_idx = next_idx
+                elif name in flatten_name_to_root_layout:
+                    layout_sliced.append(flatten_name_to_root_layout[name])
```
lw, line 298: Should there be an `else` here to catch remaining cases? (Maybe just throwing)

torch/distributed/device_mesh.py
lw: Could we get clarity on which context this is needed in and whether the same can be achieved in other ways?

torch/distributed/device_mesh.py
```diff
@@ -317,41 +271,41 @@ def _get_slice_mesh_dims(
...
+            sliced_strides = tuple(l.strides for l in layout_sliced)
+            # When users sliced dim_names outside from current mesh, we will check whether
+            # there is layout overlap.
+            # TODO: Eventually we will just directly throw error here because
+            # we will deprecate the slicing of flattened dim_name from root mesh.
+            layout_sliced = _MeshLayout(sliced_sizes, sliced_strides)
+            if not layout_sliced.check_non_overlap():
+                raise RuntimeError(
+                    f"Slicing overlapping dim_names {mesh_dim_names} is not allowed."
+                )
```
lw, line 331: Nice, this has become very clean and simple!

torch/distributed/device_mesh.py
```diff
@@ -317,41 +271,41 @@ def _get_slice_mesh_dims(
...
-                        f"Found mesh dim indices to slice: {slice_mesh_dims}. "
-                        "Mesh dim indices should be in ascending order."
+            layout_sliced = []
+            for name in mesh_dim_names:
+                if name in device_mesh.mesh_dim_names:
+                    layout_sliced.append(
+                        device_mesh._layout[device_mesh.mesh_dim_names.index(name)]
                     )
-                curr_idx = next_idx
+                elif name in flatten_name_to_root_layout:
```
lw, line 297: Maybe this will get addressed in later PRs, but what I see here is that we make a distinction between two "types" of dimensions: the ones that come from flattening and all the other ones. This distinction feels artificial to me. Once a dimension has been flattened, we compute its (nested?) sizes and strides, just like any other dimension, and then it should become indistinguishable and we should have a single code path. Why does it still require special handling?

lw (approved):
This is going in the right direction, LGTM! I left some comments on how I think we can push this further, but perhaps you're already taking care of them in the rest of the stack

torch/distributed/device_mesh.py
fduwjj: @lw my understanding is that inside Shampoo implementation it needs to know all other sub-meshes as well. cc: @wz337

torch/distributed/device_mesh.py
fduwjj: this is a very fair ask, so it is mostly due a constraint of c10d api, all ranks have to call the new_group api even that rank is not in that rank otherwise it will hang.

> constructing a DeviceMesh objects must have some global side-effects. Is that so?

Yes unfortunately we also want to fix this but definitely outside DeviceMesh

torch/distributed/device_mesh.py
fduwjj: > Moreover, if the backend supports splitting (e.g., NCCL) then all of this isn't even actually needed!

Well you don't get this for free because you still need to create the default PG first otherwise you cannot split. Also for BC reason not everyone is using split so we still need to support legacy path. But I agree the code here is nasty and ugly to some degree

torch/distributed/device_mesh.py
fduwjj: > Can I ask if you intend to keep these global (thread-local) mappings, or if the plan is to eventually drop them? I still don't understand exactly why we need them, but if they're slated for removal I won't worry about them too much.

1. These mappings are needed for BC reason because we still need to support slicing of flattened mesh which was a bad decision. But since we already support this feature, it will take at least two cycles to remove it unfortunately. Eventually they will go away.
2. We do have plans to move them out into DeviceMesh object itself, but that will be a separate PR. To do things incrementally is better than a super giant PR.

Does this sound make sense to you?

torch/distributed/device_mesh.py
fduwjj: Ahhh this is a legacy reason. For BC reason we need to keep this because we need to support slicing from root for the flattened dim. For example:
```
root_mesh = init_device_mesh((2,2,2), ("dp", "cp", "tp"))
root_mesh["dp", "cp"]._flatten()
```
we need to support the case:
```
root_mesh["dp_cp"]
```
This is a wrong decision we made in the past, however we do need to support it for BC reason. After we deprecate and remove it, we can clean this up. Maybe I can add a TODO for this one as well.

torch/distributed/device_mesh.py
fduwjj: No not needed because we checked the keys in L265. So else will never be hit.

torch/distributed/device_mesh.py
ezyang: You've mentioned legacy problems a bit, I would like it to be spelled out somewhere (e.g., PR desc) what exactly the legacy problem is, because I don't understand it right now.

test/distributed/test_device_mesh.py
```diff
@@ -875,10 +883,14 @@ def test_flatten_mesh_3d(self):
         self.assertEqual(flattened_dp_cp_mesh.mesh_dim_names[0], "dp_cp")
         root_mesh = _mesh_resources.get_root_mesh(dp_cp_mesh)
         self.assertEqual(root_mesh, mesh_3d)
-        flatten_mesh_root_dims = _mesh_resources.flatten_name_to_root_dims[root_mesh][
+        flatten_mesh_layout = _mesh_resources.flatten_name_to_root_layout[root_mesh][
             "dp_cp"
         ]
-        self.assertEqual(flatten_mesh_root_dims, (0, 1))
```
ezyang, line 881: so, uh, is this assert still True though

torch/distributed/device_mesh.py
```diff
@@ -65,17 +65,18 @@ def _init_device_mesh_stub():
...
+    torch.serialization.add_safe_globals([_MeshLayout])
+
     class _MeshEnv(threading.local):
         def __init__(self) -> None:
             self.mesh_stack: list[DeviceMesh] = []
             self.child_to_root_mapping: dict[DeviceMesh, DeviceMesh] = {}
-            self.mesh_dim_group_options: dict[
-                int, tuple[Optional[str], Optional[C10dBackend.Options]]
-            ] = {}
+            self.mesh_dim_group_options: dict[int, BackendConfig] = {}
```
ezyang, line 77: Just for reference, this is something you could put in an earlier stacked PR, easy BE and easy stamp

torch/distributed/device_mesh.py
ezyang: If this is legacy, help me understand why storing MeshLayout over mesh dim index is preferable? For legacy things I would have naively expected you to keep them exactly as is and route around the damage.

torch/distributed/device_mesh.py
```diff
@@ -477,6 +428,20 @@ def __init__(
             self.mesh_dim_names = tuple(mesh_dim_names) if mesh_dim_names else None
             if backend_override is None:
                 backend_override = ((None, None),) * self.mesh.ndim
+            # Internal bookkeeping for the device mesh.
+            self._layout = (
+                _layout
+                if _layout
+                else _MeshLayout(self.mesh.size(), self.mesh.stride())
```
ezyang, line 456: Does the layout always have to be consistent with self.mesh.size()/stride()?

torch/distributed/device_mesh.py
ezyang: Here's what the code usage in Shampoo looks like:

```
        # NOTE: We get all submeshes along the "replicate" dimension, then pick out
        # the sub-mesh that the optimizer state is assigned to.
        #
        # For the example above, this would give me submeshes [[3, 27], [11, 35], [19, 43]].
        # Note that the group source rank must belong to {0, 1, 2} in this case.
        # Suppose the group_source_rank = 1, then this would get the submesh [11, 35].
        replicate_submesh = _mesh_resources._get_all_submeshes(
            device_mesh_2d, "replicate"
        )[group_source_rank]
```

ezyang: This feels like a legitimate use case although I don't know exactly how to spell this. It feels like what is happening here is they want to do a submesh slice on replicate, but if they do it directly as `_mesh_resources["replicate"]` they will get the submesh for the CURRENT rank. But instead, they want to run it for `group_source_rank`. I feel this should be directly supportable, but it does interact with the current idea that a submesh is "only" a Layout, but it is also a particular coordinate picked out of the complement.

torch/distributed/device_mesh.py
fduwjj: no, by legacy I mean to have BC. And we eventually do plan to remove it. To answer your question, why we prefer layout over mesh dim index, because mesh dim index does not support flatten of un-contiguous dims. And if we want internal bookkeeping to use layout, we cannot just only replace part of it not all of them?

torch/distributed/device_mesh.py
fduwjj: It is the example I mentioned above:
```
root_mesh["dp_cp"]
```

This is something we already support and we plan to deprecate it. Since this is a refactor, we need to make sure we are not changing the current behavior.

test/distributed/test_device_mesh.py
fduwjj: we are not using root_dims for bookkeeping, so this check makes no sense now.

torch/distributed/device_mesh.py
fduwjj: sure, let me split

torch/distributed/device_mesh.py
fduwjj: Its numel has to be the same, but the value of size and stride no. That's exactly why we need to scale in the remapping.

torch/distributed/device_mesh.py
fduwjj: Split and merged in https://github.com/pytorch/pytorch/pull/163928

torch/distributed/device_mesh.py
fduwjj: no, I want it to be the last one and private for now.

torch/distributed/device_mesh.py
fduwjj: So in an ideal world, layout always has to be consistent with self.mesh.size()/stride(). But currently we have couple constraints:
1. All ranks need to call backend init, even if that ranks are not part it. So we will want to keep the layout to be same as before the submesh is sliced. This is also why we need to do scale.
2. Another thing we need to support slicing out flattened mesh from root mesh.

We believe both will be gone eventually and we will be able to eventually come back to this consistency and get rid of scaling as well.

torch/distributed/_mesh_layout.py
```diff
@@ -243,3 +244,90 @@ def check_non_overlap(self) -> bool:
         """
         ranks = self.all_ranks_from_zero()
         return len(ranks) == len(set(ranks))
+
+    def remap_to_tensor(
+        self,
+        mesh_tensor: torch.Tensor,
+        world_size: int,
```
fegin: If 1) we don't need all ranks to participate in sub pg creation, 2) all operation, including flatten and slicing are functional, can I claim that world_size should be needed for this function because both mesh_tensor and layout will be scaled appropriately. So world_size is always the same as mesh_tensor.numel() and we don't need to worry about sub PG creation.

fegin: If my new understanding is correct, then my original comment was incorrect that this `remap_to_tensor()` can be a standalone PR because its argument is tightly coupled with DeviceMesh and PG details. Then we may need to describe the restriction (sub PG creation and slicing is not functional)

torch/distributed/device_mesh.py
```diff
@@ -342,41 +288,41 @@ def _get_slice_mesh_dims(
...
+            layout_sliced = _MeshLayout(sliced_sizes, sliced_strides)
+            if not layout_sliced.check_non_overlap():
+                raise RuntimeError(
+                    f"Slicing overlapping dim_names {mesh_dim_names} is not allowed."
+                )
 
-            return slice_mesh_dims
+            return layout_sliced
 
+        # TODO: to make this use case by other components public API in the future.
```
fegin, line 335: If we try to make all operations functional and only keep the `root` mesh tensor, is this still doable?

torch/distributed/device_mesh.py
fegin: Probably not in this PR, and I'm not sure if you have done this in the unflatten PR, but we should add some TODOs and description about why these code exist and not the same as the unflatten design. BC is mostly the reason.

fegin (approved):

ezyang:
I still don't really understand the BC concerns but I am happy to defer to @lw and @fegin here!

torch/distributed/device_mesh.py
```diff
@@ -86,109 +87,49 @@ def get_current_mesh(self) -> "DeviceMesh":
...
                 cur_rank,
                 submesh_dim_names,
                 _init_backend=False,
+                _layout=layout,
             )
-
-            res_submesh._dim_group_names = slice_dim_group_name  # type: ignore[possibly-undefined, has-type]
-            self.child_to_root_mapping[res_submesh] = device_mesh
-
+            res_submesh._dim_group_names = slice_dim_group_name
```
lw, line 123: I'm quite confused by what this attribute is/does. `slice_dim_group_name` isn't annotated above, but it seems to be `List[int]` (we append to it the result of calling `.index()` on a list). However, in other circumstances, `_dim_group_names` seems to be a `List[str]`:
https://github.com/pytorch/pytorch/blob/4e8687921702ed0f5d36a785573d7ceb70d9ed24/torch/distributed/device_mesh.py#L965

Which of the two is it?

torch/distributed/device_mesh.py
```diff
@@ -208,47 +149,50 @@ def create_flatten_mesh(
...
-            for flatten_dim_in_root in flatten_dims_in_root:
-                remained_dims_in_root.remove(flatten_dim_in_root)
-
-            pg_ranks_by_dim = root_mesh.mesh.permute(
-                *remained_dims_in_root, *flatten_dims_in_root
-            ).reshape(-1, flattened_mesh_dim_size)
-
             cur_rank = root_mesh.get_rank()
+            # Due to the limitation of ProcessGroup api, we need to start from root mesh so that all ranks call the
+            # new_group api to avoid potential hang.
```
lw, line 172: Is this comment still relevant _here_, or should it be moved inside the helper function you introduced?

torch/distributed/device_mesh.py
```diff
@@ -208,47 +149,50 @@ def create_flatten_mesh(
...
+            pg_ranks_by_dim = flattened_mesh_layout.remap_to_tensor(
+                root_mesh.mesh,
+                get_world_size(),
+            )
+            # Flatten each mesh_nd tensor before passing to the helper method
+            # this is needed for flatten non-contiguous mesh dims.
+            flattened_pg_ranks_by_dim = torch.stack(
+                [mesh_nd.flatten() for mesh_nd in pg_ranks_by_dim]
+            )
```
lw: `pg_ranks_by_dim` is a Tensor, so isn't this expression equivalent to `pg_ranks_by_dim.flatten(start_dim=1)`?

torch/distributed/device_mesh.py
lw: I agree with @ezyang, it'd be very valuable to get a clear description of the legacy constraints (documented in the code, not as GitHub comments) so we can keep them in mind when proposing changes.

As I understand it, the legacy behavior you're describing occurs during lookup of a named mesh dim (i.e., inside the `__getitem__` method): on the _root_ mesh (and only on that one!) needs to be aware of dimensions that are not part of the original dims at creation, but which were obtained via flattening some of them. Is that so?

If that's the case then I'd expect to see some ad-hoc code inside the `__getitem__` method (plus some bookkeeping elsewhere), but the rest of the code to be relatively unimpacted.

torch/distributed/device_mesh.py
```diff
@@ -66,16 +66,17 @@ def _init_device_mesh_stub():
...
             self.child_to_root_mapping: dict[DeviceMesh, DeviceMesh] = {}
             self.mesh_dim_group_options: dict[int, BackendConfig] = {}
             self.root_to_flatten_mapping: dict[DeviceMesh, dict[str, DeviceMesh]] = {}
-            # Record flatten mesh name to its mesh dim index in root mesh.
-            self.flatten_name_to_root_dims: dict[
-                DeviceMesh, dict[str, tuple[int, ...]]
+            # Record flatten mesh name to its flattened layout in root mesh.
+            self.flatten_name_to_root_layout: dict[
+                DeviceMesh, dict[str, _MeshLayout]
             ] = {}
```
lw: [MAJOR] I really want to get rid of these attributes, as I dislike any form of global state. `mesh_dim_group_options` is already unofficially deprecated since https://github.com/pytorch/pytorch/pull/159371, so let's focus on the others. I believe that reworking this will help simplify the rest of the code too, so please hear me out.

The keys of both `root_to_flatten_mapping` and `flatten_name_to_root_layout` are all and only the _root_ meshes. So, instead of storing this in a single centralized location (as a dict), we could store this information directly as raw _attributes_ on the root mesh _objects_. Concretely:
```diff
- _mesh_resources.root_to_flatten_mapping[_mesh_resources.get_root_mesh(m)][mesh_dim]
+ _mesh_resources.get_root_mesh(m)._root_to_flatten_mapping[mesh_dim]
```

Now, we could add these attributes to the DeviceMesh class, which means that _all_ meshes get them, and we just leave these empty on non-root ones. However, I think there's a cleaner solution: we could define a new class, let's call it `SharedState` for now, and store it on each DeviceMesh under the `_shared_state` attribute. When we create a new _root_ mesh we initialize a new fresh ShareState for it, but when we create a _child_ mesh (via slicing, flattening, ...) we make it reuse the same SharedState as the parent. This way all meshes will have the same SharedState as their root. This allows to simplify the above expression even further:
```diff
- _mesh_resources.get_root_mesh(m)._root_to_flatten_mapping[mesh_dim]
+ m._shared_state._root_to_flatten_mapping[mesh_dim]
```
Note how we've managed to get rid of all global state! Yay!

With all this preliminary work, we can now leverage this SharedState class to simplify many more things:
- we could add a `root_mesh` attribute to SharedState, so that we replace the global state `_mesh_resources.child_to_root_mapping[m]` with `m._shared_state.root_mesh`.
- instead of storing `_dim_group_names` on _each_ mesh, and copying most of it over when we slice/flatten/..., we could have a common dictionary which maps the (sub-)MeshLayout to the name of the ProcessGroup that backs it;
- we could add a field that tracks which dimension names are the result of flattening, so that the `__getitem__` method on the root device mesh can access this and easily implement the legacy behavior;
- I think we could even centralize some of the name-to-(sub-)layout mapping, but let's maybe do the other changes first to see what they give.

Does this make sense? Do you see any issues with this proposal that I might have missed?

lw (commented):
This is a pretty big change so I want to make sure I get it. There are a few aspects I still don't understand, and I'd like to tackle them one by one to avoid getting confused. The first one is the infamous legacy behavior that we need to preserve, and how that is preventing us from simplifying the bookkeeping of flattened dims and remove global state. I left one major comment where I propose a step towards a simplification.

Once that's addressed, I'll also want to discuss the `remap_to_tensor` method, as I still don't get it fully.

ezyang (dismissed):
blocking merge pending BC considerations discussion

torch/distributed/device_mesh.py
fduwjj: Yes it is because we need `__getitem__` to work with flatten for BC reason. And to clarify, this requires the bookeeping to have hack both at flatten and `__getitem__`.

torch/distributed/device_mesh.py
fduwjj: it is always List[str] because it is pg_group_name which is a string, it can be a str of a global counter (again c10d related) or it can be a hash of all ranks within a pg. This is controlled by a kwarg in generating pg_name. Code pointer: https://github.com/pytorch/pytorch/blob/95be302889b8683b7ec7793a69ffa8891b6b5af8/torch/distributed/distributed_c10d.py#L5017

torch/distributed/device_mesh.py
fduwjj: yes it is always like that unless we fix it on the c10d level

torch/distributed/device_mesh.py
fduwjj: oh right, this is simplier.

torch/distributed/device_mesh.py
fduwjj: I think the change you proposed here itself should be a separate PR because we will need to add one more comparison attribute to DeviceMesh Hash which could make DTensor CPU overhead increase. I mean we need to do benchmarking for this as well. Or you really feel strong about we should do this in this PR?

Also to make things more complicated `_mesh_resources ` today is also directly being used so more test and code changes are needed. This PR is to first bring in layout into DeviceMesh so that we can have unflatten as well.

ezyang:
bouncing request changes per VC

torch/distributed/device_mesh.py
fduwjj: OK, for this part I will send out a separate PR to address it.

torch/distributed/_mesh_layout.py
```diff
@@ -243,3 +244,53 @@ def check_non_overlap(self) -> bool:
...
+
+        Returns:
+            torch.Tensor: A tensor representing the actual device allocation from mesh_tensor
+        """
+
+        # Create indexing tensor of the mesh mesh
+        ranks_indices_list = self.global_ranks(mesh_tensor.numel())
+        indices = torch.tensor(ranks_indices_list, device="cpu", dtype=torch.int).view(
+            -1,
+            *flatten(self.sizes),  # type: ignore[arg-type]
```
lw: I think it's incorrect to use `flatten(self.sizes)` here, because in case of a IntTuple with depth of 2+ this will break.

You can obtain such a DeviceMesh by starting with a (DP=2, CP=2, TP=2) mesh and flattening the DP+TP dimensions. The `mesh["dp_tp", "cp"]` mesh is 2-dimensional but it will have size ((2, 2), 2) and stride ((4, 1), 2), thus `flatten` will return a tuple with _3_ elements, and the result of this method will be a 3-d tensor, while it should be 2-d.

We need to add a test for this to the test suite.

torch/distributed/device_mesh.py
```diff
@@ -86,109 +88,49 @@ def get_current_mesh(self) -> "DeviceMesh":
...
-                        self.root_to_flatten_mapping[device_mesh][
-                            mesh_dim_name
-                        ]._dim_group_names[0]  # type: ignore[has-type]
+                        device_mesh._dim_group_names[  # type: ignore[has-type]
+                            not_none(device_mesh.mesh_dim_names).index(name)
+                        ]
                     )
                 else:
-                    slice_dim_idx.append(mesh_dim_indices[0] - num_dims_flatten)
+                    flatten_mesh = self.root_to_flatten_mapping[device_mesh][name]
```
lw, line 105: Should the key here be `device_mesh` or `root_mesh`? In all other usages of `root_to_flatten_mapping` in this file that dictionary is always indexed with `root_mesh`.

If this is a bug we need to add a regression test to the test suite.

torch/distributed/device_mesh.py
```diff
@@ -86,109 +88,49 @@ def get_current_mesh(self) -> "DeviceMesh":
...
-                    device_mesh.device_type,
-                    mesh_nd,
-                    mesh_dim_names=submesh_dim_names,
-                    _init_backend=False,
-                )
-                if cur_rank in mesh_nd:
-                    res_submesh = submesh
+            pg_ranks_by_dim = layout.remap_to_tensor(
+                root_mesh.mesh,
+            )
```
lw, line 114: We do _not_ need to invoke `remap_to_tensor` here. We're creating a view of the current mesh, which is thus able to share the same mesh tensor as the root mesh, without materializing a fresh one. For that we will probably have to introduce a new private constructor for DeviceMesh, and use that instead of `_create_mesh_from_ranks`, but that shouldn't be a big deal.

torch/distributed/device_mesh.py
```diff
@@ -86,109 +88,49 @@ def get_current_mesh(self) -> "DeviceMesh":
...
-                    mesh_dim,
-                    1,
-                )
-                for mesh_dim in submesh_dims
-            ]
-
-            mesh_tensor = device_mesh.mesh
-            # slice_dim_idx could be different from submesh_dims, as we may need to flatten out some dims.
-            slice_dim_idx = []
+            root_mesh = _mesh_resources.get_root_mesh(device_mesh)
```
lw: Nit: this is a method of `_MeshEnv`, hence instead of accessing the global variable `_mesh_resources` we can/should just use `self`.

torch/distributed/device_mesh.py
```diff
@@ -86,109 +88,49 @@ def get_current_mesh(self) -> "DeviceMesh":
...
                 submesh_dim_names,
                 _init_backend=False,
+                _layout=layout,
             )
-
-            res_submesh._dim_group_names = slice_dim_group_name  # type: ignore[possibly-undefined, has-type]
-            self.child_to_root_mapping[res_submesh] = device_mesh
-
+            res_submesh._dim_group_names = slice_dim_group_name
+            self.child_to_root_mapping[res_submesh] = self.get_root_mesh(device_mesh)
```
lw: Nit: We already have a variable `root_mesh`, no need to fetch it again.

torch/distributed/device_mesh.py
```diff
@@ -86,109 +88,49 @@ def get_current_mesh(self) -> "DeviceMesh":
...
             self,
             device_mesh: "DeviceMesh",
             mesh_dim_name: Optional[str] = None,
-            backend_override: BackendConfig = (
-                None,
-                None,
-            ),
+            backend_override: BackendConfig = (None, None),
         ) -> "DeviceMesh":
             root_mesh = _mesh_resources.get_root_mesh(device_mesh)
```
lw: Nit: same here

torch/distributed/device_mesh.py
```diff
@@ -208,47 +150,46 @@ def create_flatten_mesh(
                     f"Please specify another valid mesh_dim_name.",
                 )
 
+            flattened_mesh_layout = device_mesh._layout.coalesce()
```
lw, line 153: I think that coalescing isn't enough, as that preserves the order of dims/strides, whereas we want to normalize the layout into a canonical form that disregards that.

Concretely, we should make sure that `mesh["dp", "cp"].flatten()` and `mesh["cp", "dp"].flatten()` are able to reuse each other's mesh.

If this is indeed a bug, we should add a test for it.

torch/distributed/device_mesh.py
```diff
@@ -208,47 +150,46 @@ def create_flatten_mesh(
...
-            pg_ranks_by_dim = root_mesh.mesh.permute(
-                *remained_dims_in_root, *flatten_dims_in_root
-            ).reshape(-1, flattened_mesh_dim_size)
-
             cur_rank = root_mesh.get_rank()
+            # Due to the limitation of ProcessGroup api, we need to start from root mesh so that all ranks call the
+            # new_group api to avoid potential hang.
+            pg_ranks_by_dim = flattened_mesh_layout.remap_to_tensor(
+                root_mesh.mesh,
+            )
```
lw, line 175: Same as above: flattening is just a different way of viewing the same mesh, thus we do _not need_ to materialize the new mesh tensor thus we should _not_ do it, as it can be very expensive. Just like before, we just need to add a new constructor.

torch/distributed/device_mesh.py
```diff
@@ -208,47 +150,46 @@ def create_flatten_mesh(
...
             )
-            self.child_to_root_mapping[res_flattened_mesh] = root_mesh  # type: ignore[possibly-undefined]
+            self.child_to_root_mapping[res_flattened_mesh] = root_mesh
             self.root_to_flatten_mapping.setdefault(root_mesh, {})[mesh_dim_name] = (
-                res_flattened_mesh  # type: ignore[possibly-undefined]
+                res_flattened_mesh
+            )
+            self.flatten_name_to_root_layout[root_mesh][mesh_dim_name] = (
+                res_flattened_mesh._layout
             )
```
lw: It seems the `flatten_name_to_root_layout` is very redundant wrt `root_to_flatten_mapping`, why do we need to keep both? It seems to me that we can replace each access to `self.flatten_name_to_root_layout[foo][bar]` with `self.root_to_flatten_mapping[foo][bar]._layout`. Isn't that so?

test/distributed/test_device_mesh.py
lw, line 897: In reference to one of the issues I flagged below, we should add a check here like `self.assertIs(mesh_3d["cp", "dp"]._flatten(), flattened_dp_cp_mesh)`. I suspect this check currently fails.

torch/distributed/_mesh_layout.py
```diff
@@ -243,3 +244,53 @@ def check_non_overlap(self) -> bool:
...
+        )
+
+        return mesh_tensor.flatten()[indices]
```
lw: I think this method can be rewritten to make it more efficient by only using Tensor operations and avoid Python (nested) lists:
```py
outer_layout = self.complement(world_size)
return mesh_tensor.flatten().as_strided(
    flatten(outer_layout.sizes) + flatten(self.sizes),
    flatten(outer_layout.strides) + flatten(self.strides),
).reshape(-1, *(self[i].numel() for i in range(len(self))))
```

lw (commented):
Thanks for the changes! I believe this has helped a lot the readability of the code!

I took the time to read this carefully in full. In doing so I believe I found 3-4 bugs which, apparently, weren't caught by the current test suite. We should add tests for them immediately, and fix them.

I also think that we're constructing too many unnecessary mesh tensors. This is an expensive operation (proportional to the world size) and the point of switching to CuTe layouts was precisely to avoid doing so. It should be extremely easy to remove these intermediate mesh tensors, provided we introduce a new (private) constructor for DeviceMesh.

torch/distributed/device_mesh.py
fduwjj: I think previously we don't allow the case like `mesh["cp", "dp"].flatten()` because we have a check during slicing on the dim order and we will throw a KeyError("Mesh dim indices should be in ascending order"). With PR, I was hoping we enable this behavior.

We can continue forbidden it or we can just define the behavior like you said, they share the same mesh. Since this is a refactoring, I think we should still throw error. And I will address this behavior in future PR so that the discussion will be focused on this specific topc.

torch/distributed/device_mesh.py
fduwjj: We do need to materialize tensors for backend creation. I agree we can eventually only have one root tensors. But the API of c10d needs a materialized tensor for group creation. That unfortunately is unavoidable.

torch/distributed/_mesh_layout.py
fduwjj: With your suggestions here: https://github.com/pytorch/pytorch/pull/163213#discussion_r2394945514. I think this case is solved. And I added a test case for it.

torch/distributed/device_mesh.py
fduwjj: Ahh good question, this is indeed intentional, I can add a comment to it. We only (although we will deprecate it soon) support slicing from root of flattened mesh dim. So if device_mesh is not root, this is empty and we check this condition in `_get_slice_mesh_layout`. I agree the code might be a bit confusing if you read it without L271.. let me see if I make it cleaner.

Edit: I will add comments around this code.

torch/distributed/device_mesh.py
fduwjj: For flatten we do need it because it's a requirement of c10d API.

torch/distributed/device_mesh.py
fduwjj: that is true... I think previously with indexes, it is not possible to consolidate, but now with layout, we can do that. Another point to support having CuTe layout to simply internal bookkeeping.

test/distributed/test_device_mesh.py
fduwjj: as discussed above, original device_mesh does not support this way of slicing, I will address this behavior in follow-up PR because this is a refactoring PR. But IIUC, self.assertIs(mesh_3d["cp", "dp"]._flatten(), flattened_dp_cp_mesh) this should not share the same mesh tensor, why?

the former has mesh tensor: (for example on rank 0)
[[0, 4], [2, 6]]

while flattened_dp_cp_mesh has the mesh tensor:
[[0, 2], [4, 6]]

the point is the order of dim in the device mesh matters in the original implementation. OK here is what I am going to do, same as the feature of slicing transposed order which was not supported in previous code, I will move this to a different PR and error out the case when users try to flatten a non-contigous dims, this is the behavior in the current code and I will also add comments there as well.

torch/distributed/_mesh_layout.py
fduwjj: Thanks for your suggestions I think this also fix the bug you mentioned in https://github.com/pytorch/pytorch/pull/163213#discussion_r2394685106?

PR #163214, opened 2025-09-17
[CI] reuse old whl: fix metadata file not getting version replaced

clee2000:
In the .dist-info/METADATA file, the version was not being written with the new sha.

On python <3.11 (I think), the glob `**` will only match directories, so change this to `*`, which I checked that it will match both files and directories on py3.9 and py3.13

There's probably also a bunch of mismatches in RECORD but thats a problem for later


huydhn (approved):
Oh, is this the proper fix for https://github.com/pytorch/pytorch/pull/163232, feel free to revert mine change if that's the case

PR #163215, opened 2025-09-18
[inductor] do comm compute overlap at aten fx level

eellison:
* #163960
* #163959
* #163754
* __->__ #163215

This is first part of the stack that does comm/compute reordering, and then uses the exposure analysis to do bucketing.

Subsequent prs will handle:
- use of exposure analysis to do bucketing
- make sure inductor respects comm/compute overlapping done at fx level
- non-profiling mm estimation/rank broadcasting of profile results 

Other mis:
- Validate accuracy of nccl estimations  ( use ruisi's profiling instead ?)

For a llama 2d parallelism test, on forward, we overlap all but 2 of potentially hidden collectives. For backward, we overlap 217/269 of potentially hidden collectives. If you increase `compute_overlap_multipler` (for fudge factor of inaccurate comms estimation), that goes down to all but 16 of potentially hidden collectives.

fwd example: https://gist.github.com/eellison/76209c49d8829c5f1e323d34a3f040c3

bwd example: https://gist.github.com/eellison/6cfc2285df53a94cfa4012f5fdae5c51


ezyang:
Don't forget to post the scripts you were using to validate 2d llama!

torch/_inductor/fx_passes/post_grad.py
```diff
@@ -197,7 +197,15 @@ def post_grad_passes(gm: torch.fx.GraphModule, is_inference: bool):
                 pass_name = "custom_backend_passes_" + device
                 GraphTransformObserver(gm, pass_name).apply_gm_pass(custom_backend_pass)
 
+    if torch._inductor.config.test_configs.aten_fx_overlap_scheduling:
+        from torch._inductor.fx_passes.overlap_scheduling import (
+            schedule_overlap_bucketing,
```
ezyang: It would be nice to have a test that also exercises this from outside of Inductor, to show that we aren't overly reliant on Inductor configs

ezyang:
oh no merge conflicts

torch/_inductor/comm_analysis.py
```diff
@@ -37,11 +38,7 @@ def get_gpu_type() -> NVIDIA_GPU_TYPE:
         return NVIDIA_GPU_TYPE.AMPERE
 
 
-def get_collective_type(node: ir.IRNode) -> NCCL_COLL:
-    if not isinstance(node, ir._CollectiveKernel):
-        raise ValueError(f"node is not a collective kernel: {node}")
-
-    kernel_name = node.python_kernel_name
+def get_collective_type_from_kernel_name(kernel_name: str) -> NCCL_COLL:
```
ezyang, line 47: str matching seems pretty sus, any reason we're not matching on OpOverload?

ezyang:
@eellison when do you think you'll be able to rebase the PR and resolve the merge conflicts? (Not a request, just trying to plan around this.)

torch/_inductor/comm_analysis.py
eellison: this is a refactoring with existing logic, but i agree we should improve

torch/_inductor/comm_analysis.py
```diff
@@ -318,3 +323,70 @@ def estimate_nccl_collective_runtime(node: ir.IRNode) -> float:
 ################################################################################################################
 # The above code and constants are adapted from https://github.com/NVIDIA/nccl/blob/master/src/graph/tuning.cc #
 ################################################################################################################
+
+
+def estimate_nccl_collective_runtime(node: ir.IRNode) -> float:
+    """
+    Returns estimated NCCL collective runtime in nanoseconds (ns).
```
IvanKobzarev, line 330: `estimate_nccl_collective_runtime_impl` returns in millis , so we should either change here to millis or to change everywhere else to nanos.

torch/_inductor/fx_passes/overlap_scheduling.py
```diff
@@ -0,0 +1,655 @@
...
+        max_coll_distance: int = 1000,
+    ):
+        self.gm = gm
+        self.graph = gm.graph
+        self.compute_overlap_multipler = compute_overlap_multipler
+        self.max_node_distance = max_coll_distance
+        self.max_in_flight_bytes: int = int(max_in_flight_gb * 1024 * 1024 * 1024)
+
+        # Build structures
+        stable_topological_sort(self.graph)
```
IvanKobzarev, line 197: Do we have passes in inductor that does not guarantee correct ordering?

torch/_inductor/comm_analysis.py
```diff
@@ -318,3 +323,70 @@ def estimate_nccl_collective_runtime(node: ir.IRNode) -> float:
...
+    return estimate_nccl_collective_runtime_impl(
+        tensor_storage_size_bytes, group_size, coll
+    )
+
+
+def estimate_fx_collective_size(fx_node: torch.fx.Node) -> int:
+    size = 0
+    for node in fx_node.all_input_nodes:
+        if (t := node.meta.get("val")) is not None:
+            size += t.numel() * t.element_size()
```
IvanKobzarev, line 353: For `_out` variants of ops output will be in input arguments.

torch/_inductor/fx_passes/overlap_scheduling.py
```diff
@@ -0,0 +1,655 @@
...
+        """Handle scheduling compute and finding overlaps."""
+
+        compute_time = benchmark_node(node)
+        available_compute = compute_time * self.compute_overlap_multipler
+
+        # First reduce exposed time of in-flight collectives
+        for info in self.in_flight.values():
+            if info.exposed_time_ms == 0:
+                continue
+            overlap_amount = min(info.exposed_time_ms, available_compute)
```
IvanKobzarev, line 422: 'in-flight' fsdp collectives can be overlaped  by non-fsdp collectives that will be waited immediately. 
 (Without counting this, overlap can be undercalculated and we will be adding more compute nodes than needed)
Should we count those non-fsdp collectives as `is_compute_node()` ?

torch/_inductor/fx_passes/post_grad.py
eellison: tests are done on aten graphs

torch/_inductor/fx_passes/overlap_scheduling.py
eellison: good point!

torch/_inductor/fx_passes/overlap_scheduling.py
eellison: yea, the pattern matcher does not guarantee it.

IvanKobzarev (approved):

torch/_inductor/ir.py
```diff
@@ -9013,6 +9013,7 @@ def create_inplace(
         assert not unbacked_bindings, f"{kernel} {unbacked_bindings}"
         for tensor_arg in tensor_args:
             tensor_arg.realize()
+            V.graph.mark_buffer_mutated(tensor_arg.get_name())
```
ezyang, line 9016: Is this a bug fix?

eellison: yes. there is a test that exercises it.

torch/_inductor/fx_passes/overlap_scheduling.py
eellison: not yet handled, something we can do in the future

torch/_inductor/comm_analysis.py
eellison: at this point,  it should be functional without `out`, but follow up would be to assert this.

PR #163227, opened 2025-09-18
[Graph Partition] improve custom op output alias

BoyuanFeng:
For a custom op with multiple outputs, we will see the following generated code:
```
buf1 = op1(arg0)
buf3 = buf0[0]
buf4 = buf0[1]
del buf1 # <--- if buf1 is not accessed in the future
```

If `buf1` is not accessed in the future, it's good to deallocate early. So we don't delay `del` until both buf3 and buf4 are not used anymore. Note that buf3 and buf4 hold reference to the data such that `del buf1` does not prevent their usage.

However, when there are mutating args, we don't see `del buf1` immediately.

```python
@torch.library.custom_op(
    "mylib::op1",
    mutates_args=["x"],
    schema="(Tensor(a!)?  x) -> (Tensor, Tensor)",
    device_types="cuda",
)
def op1(x) -> tuple[torch.Tensor, torch.Tensor]:
    x = x + 1
    return (x + 1, x + 2)
```

image

Why? Because `buf3` is a MultiOutput with `buf1` as input and believes `buf1` (an output of FallbackKernel op1) has inputs that alias output.
https://github.com/pytorch/pytorch/blob/72fedf05752069c9e8b97c64397aedf6ee2bf5ec/torch/_inductor/ir.py#L7976-L7982

According to `[NOTE: FallbackKernel supported operators]`, as a mutating op that are auto-functionalizable, buf1's output should NOT alias any of the inputs. This PR improves get_inputs_that_alias_output of Fallback Kernel.

Use case: [moe custom op in vllm](https://github.com/vllm-project/vllm/blob/main/vllm/model_executor/layers/fused_moe/layer.py#L2057-L2064)

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @zou3519

### DIFF
--- a/test/inductor/test_perf.py
+++ b/test/inductor/test_perf.py
@@ -1157,7 +1157,7 @@ class InplacingTests(TestCase):
             )
 
             # Check that we are not allocate intermediate buffers
-            # which can be reinplaced.
+            # which can be reused.
             matches = re.findall(r"empty_strided_\w+\(", code)
             self.assertEqual(len(matches), 0)
             self.assertEqual("in_out" in code, True)
--- a/test/inductor/test_perf.py
+++ b/test/inductor/test_perf.py
@@ -1156,11 +1156,13 @@ class InplacingTests(TestCase):
                 torch.compile(f, fullgraph=True),
             )
 
-            # Check that we are allocating the minimum number of intermediate buffers
+            # Check that we are not allocate intermediate buffers
+            # which can be reinplaced.
             matches = re.findall(r"empty_strided_\w+\(", code)
-            self.assertEqual(len(matches), 1)
+            self.assertEqual(len(matches), 0)
+            self.assertEqual("in_out" in code, True)
 
-            self.assertExpectedInline(count_numel(f), """39""")
+            self.assertExpectedInline(count_numel(f), """45""")
 
     @requires_cuda_and_triton
     def test_inplace_triton_kernel_v1(self):
--- a/torch/_inductor/ir.py
+++ b/torch/_inductor/ir.py
@@ -7550,7 +7550,25 @@ class FallbackKernel(ExternKernelAlloc):
         return get_schema_info(self.op_overload).is_mutable()
 
     def get_inputs_that_alias_output(self) -> Sequence[str]:
-        return self.alias_names
+        assert isinstance(
+            self.op_overload, (torch._ops.OpOverload, torch._ops.HigherOrderOperator)
+        ), (
+            f"Fails to create FallbackKernel for {self.op_overload}: "
+            f"{type(self.op_overload)} not supported"
+        )
+
+        # See [Note: FallbackKernel supported operators]: for a mutating
+        # op that is auto-functionalizable, its outputs does NOT
+        # alias any of the inputs.
+        if (
+            not isinstance(self.op_overload, torch._ops.HigherOrderOperator)
+            and "_c10d_functional" not in self.op_overload.name()
+            and self.op_overload._schema.is_mutable
+            and can_auto_functionalize(self.op_overload)
+        ):
+            return []
+        else:
+            return self.alias_names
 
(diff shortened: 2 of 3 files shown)


test/inductor/test_perf.py
```diff
@@ -1156,11 +1156,13 @@ def f():
                 torch.compile(f, fullgraph=True),
             )
 
-            # Check that we are allocating the minimum number of intermediate buffers
+            # Check that we are not allocate intermediate buffers
```
BoyuanFeng, line 1159: the intermediate buffer `buf3` is reused for `buf8` so we don't allocate `buf8`. [code comparison (left: after this pr)](https://www.internalfb.com/intern/diffing/?paste_number=1952488322)

zou3519 (approved):

torch/_inductor/ir.py
```diff
@@ -7550,7 +7550,25 @@ def has_side_effects(self) -> bool:
...
+        )
+
+        # See [Note: FallbackKernel supported operators]: for a mutating
+        # op that is auto-functionalizable, its outputs does NOT
+        # alias any of the inputs.
+        if (
+            not isinstance(self.op_overload, torch._ops.HigherOrderOperator)
+            and "_c10d_functional" not in self.op_overload.name()
+            and self.op_overload._schema.is_mutable
+            and can_auto_functionalize(self.op_overload)
```
eellison, line 7567: Can you say more about this ? can you also add tests for collective operators ?

eellison (commented):
I understand that we want to del multi output as soon as possible, but it's not clear to me how that relates to the changes you've made here.

torch/_inductor/ir.py
```diff
@@ -7550,7 +7550,25 @@ def has_side_effects(self) -> bool:
...
     def get_inputs_that_alias_output(self) -> Sequence[str]:
-        return self.alias_names
+        assert isinstance(
+            self.op_overload, (torch._ops.OpOverload, torch._ops.HigherOrderOperator)
+        ), (
+            f"Fails to create FallbackKernel for {self.op_overload}: "
+            f"{type(self.op_overload)} not supported"
+        )
+
+        # See [Note: FallbackKernel supported operators]: for a mutating
```
eellison, line 7560: Shouldn't we just update this so as we do not create incorrect alias_names in the first place ?

torch/_inductor/ir.py
BoyuanFeng: For the 1st&2nd conditions:

```python
            not isinstance(self.op_overload, torch._ops.HigherOrderOperator)
            and "_c10d_functional" not in self.op_overload.name()
```

They are added since FallbackKernel just early returns when creating a `ir.FallbackKernel`:

https://github.com/pytorch/pytorch/blob/f329f14409de90ce50fb965b135fe9f65294ca11/torch/_inductor/ir.py#L7395-L7406

For the 3rd&4th conditions:

```python
            and self.op_overload._schema.is_mutable
            and can_auto_functionalize(self.op_overload)
```

If these conditions hold, the op is a "mutating op that is auto-functionalizable". `[Note: FallbackKernel supported operators]` says its outputs do not alias any of the inputs. So `get_inputs_that_alias_output` should return an empty list.

torch/_inductor/ir.py
BoyuanFeng: `self.alias_names` are arguments that are aliased. It seems to be right.

The issue is that `def get_inputs_that_alias_output(self) -> Sequence[str]` treats all `inputs arguments that are aliased` as `inputs arguments that alias output`. This is wrong for `mutating op that is auto-functionalizable`.

PR #163234, opened 2025-09-18
[Inductor-FX] Support torch.cond

blaine-rister:
# Feature

Support `torch.cond` in the FX converter. The generated FX IR is conceptually indentical to what would come from `torch.export`:
- Submodules as stored as attributes, and accessed via `getattr`.
- The conditional is represented as `torch.ops.higher_order.cond`, which takes in the subgraphs, a predicate and submodule inputs.

# Implementation overview

The FX backend generates code for subgraphs using the following steps:
1. When `codegen_conditional` is called in `WrapperFxCodegen`, we emit a `ConditionalLine`. 
   a. We also codegen the true/false subgraphs at this time, storing their subgms for later.
2. At the beginning of FX conversion, generate `get_attr` nodes accessing each subgraph. It's important to do this at the start, before registering the node metadata hook. This also matches the convention followed by torch.export.
3. When we see the `ConditionalLine` in the FX converter, we generate a corresponding `torch.ops.higher_order.cond`.

# Implementation details
This ended up being a substantial change, as wrapper codegen has some special logic for subgraphs. 

Certain methods of `PythonWrapperCodegen` are overridden by `SubgraphPythonWrapperCodegen`. To apply these overrides, we use multiple inheritance with the registered subclass of `WrapperFxCodegen`.

Unlike most other wrapper codegen methods, which map 1:1 to Wrapper IR lines, subgraph codegen generates a number of wrapper lines including `EnterSubgraphLine` and `ExitSubgraphLine`, along with Python or C++ code calling the subgraph as a function. These lines are used for some backends' memory planning.

In contrast, FX IR typically represents a subgraph call as a single HOP node, or a `call_module` op. To account for this difference, this PR introduces a new wrapper IR line called `ConditionalLine`, which is only used by the FX backend. We override the `codegen_conditional` method to emit this line. This sidesteps having to port the existing subgraph codegen and associated memory planning to Wrapper IR. (In principle, it seems possible to adapt the existing backends to `ConditionalLine`, but it could be a larger refactor, since we'd also have to update the memory planning.)

Some of the lower-level subgraph codegen methods are still shared between the FX and Python backends, such as `generate_subgraph_common`. Those were easier to port to Wrapper IR.

This also required generalizing the way the FX converter handles graph inputs and outputs. Previously, it assumed the IO signature was the same as `V.graph.module`, but this is only true for the parent graph, and not subgraphs. Instead, we need to call `get_graph_inputs` and `get_graph_outputs` to populate the inputs and outputs for subgraphs.

# Test plan
This PR adds a couple of tests using torch.cond. Here's an example graph generated by one of them:
```
graph():
    %arg0_1 : [num_users=1] = placeholder[target=arg0_1]
    %arg1_1 : [num_users=1] = placeholder[target=arg1_1]
    %true_graph_0 : [num_users=1] = get_attr[target=true_graph_0]
    %false_graph_0 : [num_users=1] = get_attr[target=false_graph_0]
    %cond : [num_users=1] = call_function[target=torch.ops.higher_order.cond](args = (%arg0_1, %true_graph_0, %false_graph_0, (%arg1_1,)), kwargs = {})
    %buf1 : [num_users=2] = call_function[target=operator.getitem](args = (%cond, 0), kwargs = {})
    %triton_kernel_wrapper_mutation : [num_users=0] = call_function[target=torch.ops.higher_order.triton_kernel_wrapper_mutation](args = (), kwargs = {kernel_idx: 6, constant_args_idx: 6, grid: [(1, 1, 1)], tma_descriptor_metadata: {}, kwargs: {in_out_ptr0: %buf1, xnumel: 6, XBLOCK: 8}})
    return buf1
```

It also removes an existing negative test which checked that a certain error was raised when subgraphs were encountered.


torch/_inductor/codegen/wrapper_fxir.py
```diff
@@ -184,11 +242,27 @@ def create(
...
         if is_subgraph:
-            raise NotImplementedError(
-                "Subgraphs are not yet supported by FX conversion"
+            assert subgraph_name is not None
+            assert parent_wrapper is not None
+
+            # Subgraphs override some methods of PythonWrapperCodegen.
+            # Apply these overrides to the user-provided class, with priority given to
+            # user-provided methods.
+            class SubgraphFxWrapperCodegen(cls, SubgraphPythonWrapperCodegen):
```
blaine-rister: mypy doesn't seem to support using a variable as a base class, but it's valid Python. Without this, derived classes would need their own `create` method, which would need to define a new class to handle subgraphs. This trick lets us handle everything in one class, without custom backends needing to know about subgraphs.

test/inductor/test_fxir_backend.py
```diff
@@ -116,8 +115,19 @@ def _compile_and_check(
     def setUpClass(cls):
         super().setUpClass()
 
-        # Register the FX backend.
-        register_backend_for_device(cls.device, TritonScheduling, WrapperFxCodegen)
+        # Register the FX backend, storing the default for later.
```
blaine-rister, line 120: This is a slight upgrade to fix local testing. AOTI tests could break if they were run after the torch.compile tests, since the latter modifies `device_codegens`. This restores the old backend when torch.compile tests conclude.

torch/_inductor/codegen/wrapper_fxir.py
```diff
@@ -147,12 +149,53 @@ class WrapperFxCodegen(PythonWrapperCodegen):
...
+    def is_subgraph(self) -> bool:
+        return isinstance(self, SubgraphPythonWrapperCodegen)
+
+    def get_fx_graph_inputs(
+        self,
+    ) -> dict[str, Union[ir.TensorBox, ir.TorchBindObject, sympy.Expr, None]]:
+        """
+        Get the input nodes corresponding to FX graph placeholders.
+        """
+        if V.aot_compilation and not self.is_subgraph:
```
blaine-rister, line 191: Moved this logic over from the FX converter, since we need to call `get_graph_inputs` for subgraphs.

test/inductor/test_fxir_backend.py
```diff
@@ -630,21 +640,47 @@ def foo(out, index, src):
...
 
-        cond = torch.tensor([True], device=self.device)
-        x = torch.ones([2, 3], device=self.device)
+        def foo(pred):
+            return torch.cond(pred, true_fn, false_fn, ())
 
-        with self.assertRaisesRegex(BackendCompilerFailed, "Subgraph"):
-            self._compile_and_check(foo, [cond, x])
+        pred_tensor = torch.tensor([pred], device=self.device)
+        self._compile_and_check(foo, [pred_tensor], expected_num_triton_kernels=2)
```
angelayi, line 685: maybe we can add an AOTI test which checks the graph contents using `assertExpectedInline` like [this](https://github.com/pytorch/pytorch/blob/a0d2d848465e125d0a3f4ece8a70443cb4d8f4b8/test/dynamo/test_higher_order_ops.py#L441-L458)

torch/_inductor/utils.py
```diff
@@ -666,6 +666,13 @@ def clear_cache(self: Any) -> None:
     return wrapper  # type: ignore[return-value]
 
 
+def cache_property_on_self(fn: Callable[P, RV]) -> CachedMethod[P, RV]:
```
blaine-rister, line 669: This fixes a longstanding issue where `cache_on_self`'s type signature was incompatible with `@property`. In a follow-up, we can remove a lot of `@no_type_check` decoratorations like [this one](https://github.com/pytorch/pytorch/blob/main/torch/_inductor/codegen/simd.py#L137).

torch/_inductor/codegen/wrapper_fxir.py
```diff
@@ -312,24 +391,22 @@ def _generate_graph_inputs(self) -> None:
...
-                    if isinstance(ir_node, (int, float, sympy.Integer, sympy.Float))
-                    else self._get_buffer(ir_node)
-                )
-                placeholder_node = self.gm.graph.placeholder(buffer.get_name())
-                placeholder_node.meta["val"] = buffer.get_example()
-                self._record_allocation(buffer, placeholder_node)
-
-            elif V.aot_compilation:
+        for name, ir_node in self.graph_inputs.items():
+            ir_node = self.graph_inputs[name]
```
angelayi: delete this line? you already have `ir_node` from the `.item()` right

angelayi (approved):
thanks so much for the detailed PR description! made reviewing the PR a lot easier! 

had some minor nits

torch/_inductor/codegen/wrapper_fxir.py
blaine-rister: Good catch. Removed in [3e07d5d](https://github.com/pytorch/pytorch/pull/163234/commits/3e07d5d7ca14ddf26b4aa545e5c47de618602061).

test/inductor/test_fxir_backend.py
blaine-rister: Sure. Added such a test in [cc09ac7](https://github.com/pytorch/pytorch/pull/163234/commits/cc09ac717cec326dc74b577b3cd9c426476dab41).

jansel (approved):

PR #163236, opened 2025-09-18
[export] Tag lifted constants with source object id.

zhxchen17:
Summary:
Adds a metadata to node which should achieve the same end goal as https://github.com/pytorch/pytorch/pull/162926

Basically we want to have a generic way to associate graph inputs with its original source. To achieve this, we can store an additional field "source_id" on the placeholder node so that export downstream can track the mapping between tensor objects and its placeholder node.

Test Plan:
buck2 run --flagfile fbcode//mode/opt fbcode//caffe2/test:test_export -- -r test_source_id

Rollback Plan:

Differential Revision: D82698063

### DIFF
--- a/torch/_export/passes/lift_constants_pass.py
+++ b/torch/_export/passes/lift_constants_pass.py
@@ -337,6 +337,11 @@ def lift_constants_pass(
                         f"tried to lift unsupported type {type(constant_val)} from node {node.format_node()}"
                     )
 
+                # source id is not meant to be serialized and expected to be
+                # used for tracking correspondence between placeholder nodes and
+                # the it's original constant objects. It's up to downstream to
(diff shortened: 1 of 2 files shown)


patrick-toulme:
Thanks for this! I will test it out on my workload

torch/_export/passes/lift_constants_pass.py
```diff
@@ -337,6 +337,7 @@ def lift_constants_pass(
                         f"tried to lift unsupported type {type(constant_val)} from node {node.format_node()}"
                     )
```
angelayi, line 339: maybe leave a comment saying what this is used for

angelayi (approved):

PR #163237, opened 2025-09-18
[pcache] MultiCache implementation

nmacchioni:
Summary:
Initial implementation of MultiCache + MemoizedOnDiskCache as an example

Still a WIP, but wanted to get the design out for rough feedback

Not sure if MultiCache can be thread-safe, given that insert_async is a Generator (so how would we prevent concurrent writes?)

Still need to add additional unit tests;
Should have the same functionality as AsyncCache instances, with different return types for insert and insert_async

Will also want to check:
Get with a single populated cache hits
Get with all populated caches hits
Insert with all empty caches updates all caches
Insert with some empty caches updates some caches

Should we allow different values in different caches to exist for the same key? That would be weird behavior

Rollback Plan:

Differential Revision: D82698606

### DIFF
--- a/test/inductor/test_pcache.py
+++ b/test/inductor/test_pcache.py
@@ -5,6 +5,7 @@ from concurrent.futures import ThreadPoolExecutor
 from inspect import isclass
 from os import environ
 from random import randint
+from typing import Generic
 from typing_extensions import Self
 
 from torch._inductor import pcache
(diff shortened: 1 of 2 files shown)


aorenste:
Should insert_async() return `Future[Generator[...]]` or `Generator[Future[...]]`?  If it's the first one then you'd need to wait for everything to finish before marking the future as complete - in which case why not just `Future[tuple[...]]`?  As far as the thread-safety it's not important for the returned value itself to be thread-safe - just the underlying cache object & its operations. (Saying that insert_async() is thread-safe doesn't make any ordering guarantees - it just means that you don't have to worry about corrupting the cache object)

PR #163238, opened 2025-09-18
Massive hack to make autograd shut up about threaded PG mutations

ezyang:
* __->__ #163238

See the Note for explanation.


ezyang:
@albanD if you're unhappy we can try to get multithreaded PG to route through the c10d ops but no one uses multithreaded PG for anything real and I hope to obsolete it with LocalTensor so I would prefer to spend less time on this. We also need to come up with a plan on how to drag the collectives kicking and screaming into correct mutation annotations, but I'm less certain on how to do this. In particular, one problem is that marking the entire thing as mutating is wrong for some ops; for example, if you broadcast, every tensor in the input is mutated EXCEPT for the original source tensor.

torch/testing/_internal/distributed/multi_threaded_pg.py
```diff
@@ -71,6 +71,24 @@ def bitwise_reduce(tensors, op):
...
+# but not very convincing reasons include that Threaded PG is implemented
+# in Python but you can't override Backend in Python, you can only override
+# ProcessGroup in Python), thereby bypassing the dispatch step.  Now we have
+# a problem: c10d's signatures are LIES, they mutate their (output) tensor
+# arguments but their annotations don't have mutations on them so we don't
+# actually update any view metadata if you do differentiation.  This
+# ordinarily "doesn't matter" because distributed collectives aren't
+# differentiable anyway, but it's possible to tickle this in testing if
+# someone tries to touch the grad_fn of a Tensor.  There a few ways to
+# fix this, but the easiest way was to use the .data trick to hide
```
albanD: nit: .detach()

torch/testing/_internal/distributed/multi_threaded_pg.py
```diff
@@ -224,14 +248,17 @@ def work(self, data):
...
                 else:
-                    dest_tensor_on_rank_i[0].add_(to_scatter[i].to(dst_tensor_device))
+                    # See Note [Hide collectives mutation from autograd]
+                    dest_tensor_on_rank_i[0].detach().add_(to_scatter[i].to(dst_tensor_device))
         if self.op == dist.ReduceOp.AVG:
             num_ranks = len(data)
             for each_rank_data in data:
-                each_rank_data[0][0] /= num_ranks
+                # See Note [Hide collectives mutation from autograd]
+                each_rank_data[0][0].data /= num_ranks
```
albanD:
```suggestion
                each_rank_data[0][0].detach() /= num_ranks
```

albanD (approved):
One slipped through, but SGTM once updated!

PR #150217, opened 2025-03-28
Fix DLPack stream logic.

ysiraichi:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #150691
* #150218
* __->__ #150217
* #150216

This PR fixes the logic for dealing with CUDA and ROCm streams whenever
we are trying to create a DLPack capsule from a tensor.

In summary, this PR:

- Uses the legacy default stream if `tensor.__dlpack__(stream=None)` is
  called for a CUDA tensor.
- Errors if `tensor.__dlpack__(stream=2)` is called for a CUDA tensor:
  PyTorch doesn't support the per-thread default stream.
- Errors if `tensor.__dlpack__(stream=stream)`, where `stream` is 1 or
  2, is called for a CUDA tensor using ROCm.

For more details, see [the documentation][1].

[1]: https://data-apis.org/array-api/latest/API_specification/generated/array_api.array.__dlpack__.html

PR #163242, opened 2025-09-18
Make `Tensor.__dlpack__(stream=None)` capture-safe during CUDA Graph capture

eee4017:
Many extensions (including pybind helpers) call `Tensor.__dlpack__()` without a stream argument. Before #150217, `stream=None` behaved like “no cross-stream sync” and was safe inside CUDA Graph capture. After #150217, `stream=None` maps to the legacy default stream, adding a cross-stream wait that invalidates capture when running on a non-default stream.

See this example

```
import torch
s = torch.cuda.Stream()
x = torch.randn(8, device="cuda")
g = torch.cuda.CUDAGraph()

with torch.cuda.stream(s):
    with torch.cuda.graph(g):
        _ = x + 1
        cap = x.__dlpack__()
        _ = torch.utils.dlpack.from_dlpack(cap)
```

This PR partially reverts #150217 that stream=None defaults to no sync.

cc @mcarilli @ezyang @eellison @penguinwu @BoyuanFeng

### DIFF
--- a/torch/_tensor.py
+++ b/torch/_tensor.py
@@ -1750,12 +1750,7 @@ class Tensor(torch._C.TensorBase):
             is_cuda = not is_rocm
 
             if stream is None:
-                if torch.cuda.is_current_stream_capturing():
-                    # During capture, default-stream ops would invalidate capture.
-                    # Route to the current (capturing) stream instead.
-                    stream = torch.cuda.current_stream()
-                else:
-                    stream = torch.cuda.default_stream()
+                stream = torch.cuda.current_stream()
             elif (is_rocm and stream == 0) or (is_cuda and stream == 1):
                 stream = torch.cuda.default_stream()
             else:
--- a/torch/_tensor.py
+++ b/torch/_tensor.py
@@ -1772,7 +1772,7 @@ class Tensor(torch._C.TensorBase):
                 event.record(current_stream)
                 stream.wait_event(event)
         elif self.device.type == "cpu":
-            assert stream is None or stream == -1, "stream should be None or -1 on cpu."
+            assert stream is None or stream == -1, "stream should be None on cpu."
 
         if self.device.type == "xla":
             import torch_xla
--- a/torch/_tensor.py
+++ b/torch/_tensor.py
@@ -1749,7 +1749,14 @@ class Tensor(torch._C.TensorBase):
             is_rocm = torch.version.hip is not None
             is_cuda = not is_rocm
 
-            if stream is None or (is_rocm and stream == 0) or (is_cuda and stream == 1):
+            if stream is None:
+                if torch.cuda.is_current_stream_capturing():
+                    # During capture, default-stream ops would invalidate capture.
+                    # Route to the current (capturing) stream instead.
+                    stream = torch.cuda.current_stream()
+                else:
+                    stream = torch.cuda.default_stream()
+            elif (is_rocm and stream == 0) or (is_cuda and stream == 1):
                 stream = torch.cuda.default_stream()
             else:
                 if is_cuda and stream == 2:
--- a/torch/_tensor.py
+++ b/torch/_tensor.py
@@ -1671,7 +1671,7 @@ class Tensor(torch._C.TensorBase):
     def __dlpack__(
         self,
         *,
-        stream: Optional[Any] = None,
+        stream: Optional[Any] = -1,
         max_version: Optional[tuple[int, int]] = None,
         dl_device: Optional[tuple[enum.IntEnum, int]] = None,
         copy: Optional[bool] = None,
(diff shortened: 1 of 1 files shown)


ngimel:
Why don't we want to keep no sync behavior if stream is None? Having different behavior during capture and outside of it is nightmare

eqy:
CC @tqchen as we are considering changing the behavior to use `current_stream()` when the passed `stream` is `None`, which effectively partially reverts https://github.com/pytorch/pytorch/pull/150217

Note that the with the PR which makes `stream == None` assume `default_stream()` it breaks graph capture due to the sync

ngimel:
>The problem is that we are not allowed to sync the default stream during capture (cudaErrorStreamCaptureInvalidated). Before the https://github.com/pytorch/pytorch/pull/150217 we don't have this behavior.

Exactly, and I'm questioning the changes introduced by #150217, as they effectively make capture impossible, and I'm strongly against introducing different behavior under capture and without capture.

tqchen:
Note that there is stream = -1 which indicate no sync, however, seems most impl still defaults to None. 

On a related note, the new speed convert RFC https://github.com/pytorch/pytorch/issues/162845 adds a stream query, which will make the behavior more streamlined (and compatible with cudagraph)

Of course we are talking about default behavior of dlpack, the original reasoning is that None is a safe choice (without considering cudagraph), and the dlpack interface would require explicit sync to the default stream. Considering cudagrpah, i think indeed in that mode default to current seems more sensible. 

cc @leofang

ngimel:
Please fix PR description to match what PR does

tqchen:
Just to elaborate a bit further on the choices here for clarity

Canonical way of using dlpack is to call `mylib.from_dlpack` where the mylib decides on the stream parameter to pass in. In early version of the dlpack we don't have stream

```python
s = torch.cuda.Stream()
x = torch.randn(8, device="cuda")
g = torch.cuda.CUDAGraph()

with torch.cuda.stream(s):
    with torch.cuda.graph(g):
        _ = x + 1
        mylib_tensor = mylib.from_dlpack(x)
        mylib_kernel(mylib_tensor)
```

We have two choices for mylib to implement cudagraph support

## C0: simply call `torch.__dlpack__()` without passing in stream = None, and stream=None defaults to no sync in torch 

This would requires an update in torch(this PR). 

- Pros: many of intended behavior of mylib.from_dlpack are indeed stream preserving, so `mylib` do not need update to support cuda graph.
- Cons: divergence to the data-api spec, which some frameworks `myframework` may rely on, specifically those who may not expect stream preserving behavior (the usecase where another framework do not expect stream context and use their own internal stream) and would like to "play it safe" (although not as efficient).

Of course in C0(Cons) case `mylib` can also update their call to `__dlpack__()` to pass in stream, indicating legacy default. But the default behavior(no stream passing) will results in potential issue when calling from_dlpack under a torch stream context.

## C1: simply call `torch.__dlpack__()` with stream = -1

This would requires updates to mylib in their implementation of `from_dlpack`. By explicitly passing in `stream=-1`. 
- Cons: `mylib` will not be stream preserving without an update, although the original intention of many `mylib` are stream preserving
- Pros: for other `myframework` who have a different stream handling mechanism, their behavior under torch stream context won't be broken, and for now it aligns with the data-api spec, although one could argue that the importance of stream passing would need an update of the spec.


## Discussions and Possible Action Items

Note that regardless of C0 or C1, assuming `mylib` or `myframework` is aware of their intention in `from_dlpack` and pass in `stream` parameter accordingly, we will get cudagraph compact for cases of `mylib`and correctness for the case of `myframework`.  

The only difference is course of action and whether the legacy default option should be preserved as conservative. The same thing applies to `__dlpack_versioned__` dunder in the latest post 1.0 version, which will be used for exchange onwards.

One action item that i think falls from the discussion is that data-api spec should be updated to discuss the stream preserving behavior and provide guidelines for `mylib` usecases to pass in stream correctly when stream preserving is the intention

eee4017:
How about we change `Tensor.__dlpack__`’s default stream to -1 (stream-preserving, no sync) so implicit calls are CUDA Graph–friendly and avoid default-stream traffic; this matches what most extensions intend, requires no library changes unless they explicitly want default-stream sync (which they can pass), and keeps behavior unchanged for the large majority of users—while it does slightly diverge from the Array API/DLPack guidance, we can document the deviation and note that anyone who needs default-stream synchronization should pass it explicitly.

ysiraichi:
The motivation behind PR #150217 was to make PyTorch DLPack compliant to DLPack 1.0 spec, specifically the behavior on different values for `stream` parameter. It currently does support `stream=-1`, so that could be used as an work around (as mentioned by @tqchen).

ngimel (approved):

PR #163246, opened 2025-09-18
Add support for name kwarg in mark_dynamic

bobrenjc93:
* #163421
* __->__ #163246

Ergonomic improvement to allow sharing symbols without having to do the complex torch._check paradigm as described by @anijain2305 in his recent UED:

```
Different symbols for KV cached tensors - One property in my case was that the KV cache for different attention blocks had the same seq length, but there is no API to enforce that. The only way is to add torch._check but torch.compile must trace those functions to instruct the dynamic shape infra. This required me to change the model code.
Changing model code is not the best experience. Lets see how transformers maintainers react to my PR. Maybe an API with fullgraph=True is a better bet here.
```

### DIFF
--- a/test/dynamo/test_misc.py
+++ b/test/dynamo/test_misc.py
@@ -7955,6 +7955,26 @@ utils_device.CURRENT_DEVICE == None""".split("\n"):
         with self.assertRaises(ConstraintViolationError):
             torch.compile(my_dyn_fn, backend="eager")(y)
 
+    def test_mark_dynamic_with_name(self):
+        def fn(x, y):
+            return torch.randn(1) * x, torch.randn(1) * y
+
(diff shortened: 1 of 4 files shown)


ezyang:
I think doing it with strings is unwise, as in a large enough codebase it can be difficult to avoid collisions, which will cause extremely strange errors. The only use case for strings is if there is a single global configuration spot for dynamic that applies everywhere, but mark_dynamic works both for fullgraph and graph break cases, and also it can propagate unpredictably as its data flow. Easy fix: explicitly allocate Dim symbols (similar to how export does it) and then use those to dedupe by object identity.

Speaking of which, why don't we use export's Dim directly? cc @avikchaudhuri

PR #163251, opened 2025-09-18
[xpu][feature] Enable triton online softmax kernels on XPU.

weishi-deng:
This pr is to enable triton online softmax kernels for xpu devices, so we add a device check in prepare_softmax_extra_check.


torch/_inductor/fx_passes/post_grad.py
```diff
@@ -322,8 +322,10 @@ def prepare_softmax_extra_check(match):
     """
     return (
         config.online_softmax
-        and match.kwargs["x"].meta["val"].device.type == "cuda"
-        and config.cuda_backend == "triton"
+        and (match.kwargs["x"].meta["val"].device.type == "cuda"
```
etaf: Suggest to make the logic more clear:
```        
and (
   (match.kwargs["x"].meta["val"].device.type == "cuda" and config.cuda_backend == "triton")         
or (match.kwargs["x"].meta["val"].device.type == "xpu" and config.xpu_backend == "triton")
)
```

etaf:
Is there any cuda test case we can reuse for the online-softmax?  Please enable the corresponding UT.

torch/_inductor/fx_passes/post_grad.py
weishi-deng: Fixed

etaf:
> > Is there any cuda test case we can reuse for the online-softmax? Please enable the corresponding UT.
> 
> The ut for online softmax on xpu is enabled in test_online_softmax.py. Note that the split reduction heuristic is relative to the device hardware info, so the output code with a shape of 2048 on xpu devices is different from that of CUDA. BMG has num_sm of 20, so the check in https://github.com/pytorch/pytorch/blob/main/torch/_inductor/ir.py#L1362 (as 2048 > 20 * 2 *32) will make num_split 1.

Great! could you add these explanations as comments in the test case so that others are aware of the differences from CUDA?

weishi-deng:
> > > Is there any cuda test case we can reuse for the online-softmax? Please enable the corresponding UT.
> > 
> > 
> > The ut for online softmax on xpu is enabled in test_online_softmax.py. Note that the split reduction heuristic is relative to the device hardware info, so the output code with a shape of 2048 on xpu devices is different from that of CUDA. BMG has num_sm of 20, so the check in https://github.com/pytorch/pytorch/blob/main/torch/_inductor/ir.py#L1362 (as 2048 > 20 * 2 *32) will make num_split 1.
> 
> Great! could you add these explanations as comments in the test case so that others are aware of the differences from CUDA?

Done

etaf (approved):

torch/_inductor/fx_passes/post_grad.py
```diff
@@ -322,8 +322,10 @@ def prepare_softmax_extra_check(match):
     """
     return (
         config.online_softmax
-        and match.kwargs["x"].meta["val"].device.type == "cuda"
-        and config.cuda_backend == "triton"
+        and (
+            (match.kwargs["x"].meta["val"].device.type == "cuda" and config.cuda_backend == "triton")
+            or (match.kwargs["x"].meta["val"].device.type == "xpu" and config.xpu_backend == "triton")
```
EikanWang: Could it be something like this?
```python
match.kwargs["x"].meta["val"].device.type in ["cuda", "xpu"] and "triton" in [config.cuda_backend, config.xpu_backend]
```

test/inductor/test_online_softmax.py
```diff
@@ -310,5 +312,5 @@ def f(x, y):
 instantiate_parametrized_tests(TestOnlineSoftmax)
 
 if __name__ == "__main__":
-    if IS_LINUX and HAS_CUDA_AND_TRITON:
+    if IS_LINUX and (HAS_CUDA_AND_TRITON or HAS_XPU_AND_TRITON):
```
EikanWang: Could it be something like `HAS_TRITON` and `HAS_GPU`?

EikanWang (approved):
Pls. rebase this PR.

mlazos (approved):

weishi-deng:
@EikanWang @etaf I updated the ut to add the check with device info as the split reduction is enabled on XPU PVC but disabled on XPU BMG devices. Therefore, we should add a condition check to determine when to use split reduction. Pls help review again.

torch/_inductor/fx_passes/post_grad.py
```diff
@@ -351,10 +351,9 @@ def prepare_softmax_extra_check(match):
     """
     We only have triton online softmax kernels currently.
     """
-    return (
-        config.online_softmax
-        and match.kwargs["x"].meta["val"].device.type == "cuda"
-        and config.cuda_backend == "triton"
+    return config.online_softmax and (
```
etaf: from torch._inductor.utils import get_current_backend
device_type = match.kwargs["x"].meta["val"].device.type
return config.online_softmax and device_type in ("cuda", "xpu") and "triton" == get_current_backend(device_type)

etaf (changes requested):

torch/_inductor/fx_passes/post_grad.py
weishi-deng: Updated.

torch/_inductor/fx_passes/post_grad.py
```diff
@@ -351,11 +351,9 @@ def prepare_softmax_extra_check(match):
     """
     We only have triton online softmax kernels currently.
     """
-    return (
-        config.online_softmax
-        and match.kwargs["x"].meta["val"].device.type == "cuda"
-        and config.cuda_backend == "triton"
-    )
+    from torch._inductor.utils import get_current_backend
```
etaf: Oh, I'm not mean to put the import line here, I just try to tell where you can get the `get_current_backend`. it'll be better to move the import line to the file header.

etaf (changes requested):

torch/_inductor/fx_passes/post_grad.py
weishi-deng: updated.

etaf (approved):

EikanWang (approved):

EikanWang (approved):

EikanWang (approved):

weishi-deng:
> @weishi-deng would you mind taking a look at #172048 I would appreciate your input on the changes I made. Should I have kept the fallback to num_kernels = 1 for XPUs? Thanks

I made a test for #172048, and it also works for XPU devices with/without num_sm > 20. I think this looks good to me.

PR #163252, opened 2025-09-18
Add SDPA patterns for T5 variants when batch size is 1

CaoE:
As mentioned in
https://github.com/pytorch/pytorch/blob/main/torch/_inductor/fx_passes/fuse_attention.py#L838, this PR generates patterns  for the cases batch size == 1.


torch/_inductor/fx_passes/fuse_attention.py
```diff
@@ -723,6 +687,42 @@ def _sfdp_replacement_23(query, key, value):
     )
 
 
+def _sfdp_pattern_24(query, key, value, attention_mask):
```
Valentine233, line 690: Is `pattern_24` just moved here for ordering, without any code change?

torch/_inductor/fx_passes/fuse_attention.py
```diff
@@ -643,7 +607,7 @@ def _sfdp_replacement_21(query, key, value, attn_mask):
         query,
         key,
         value,
-        attn_mask=attn_mask,
+        attn_mask=attn_mask.to(dtype=query.dtype),
```
Valentine233, line 610: As reduced q/k/v and fp32 mask are supported in SDPA, why need type conversion here?

torch/_inductor/fx_passes/fuse_attention.py
CaoE: Thanks for your comments. Yes. Just moved `pattern_24 `.

torch/_inductor/fx_passes/fuse_attention.py
CaoE: Followed other patterns defined in this file. Add type conversion here to make it more robust, because cuda `EFFICIENT_ATTENTION` does not support mixed types of mask and q/k/v.

Valentine233 (approved):

jansel (approved):

PR #163263, opened 2025-09-18
[Autograd] The behavior of Tensor.requires_grad=True is different from that of Tensor.requires_grad_(True).

fffrog:
* #163264
* __->__ #163263

As the title stated.

```Python
>>> a = torch.randn(3, 3, device="cpu", requires_grad=True)
>>> b = torch.pow(a, 2)
>>>
>>> b.requires_grad=False
RuntimeError: you can only change requires_grad flags of leaf variables. If you want to use a computed variable in a subgraph that doesn't require differentiation use var_no_grad = var.detach().
>>> b.requires_grad=True
RuntimeError: you can only change requires_grad flags of leaf variables.

>>> b.requires_grad_(False)
RuntimeError: you can only change requires_grad flags of leaf variables. If you want to use a computed variable in a subgraph that doesn't require differentiation use var_no_grad = var.detach().
>>> b.requires_grad_(True)
tensor([[4.1244, 3.3084, 0.1038],
        [0.3303, 0.0662, 0.1792],
        [0.0080, 0.0127, 0.4904]], grad_fn=<PowBackward0>)
```


soulitzer:
Isn't it clearer to have an explicit on t.requires_grad=True rather than silently doing nothing. I agree that consistency is good, but the ideal state is to error on both, rather than not error, so I'm not sure this change in this PR is necessarily better...

ezyang:
@soulitzer I am going to defer to you

FFFrog:
> Yeah I think we should keep current behavior. As for deprecation, I don't think its worth the pain. They are different code paths one is a native_function and one is not, etc. This inconsistent behavior is not that horrible imo considering that the user is already doing something weird by setting it to True.

@soulitzer Thank you for your suggestion, but we think it's better to keep things as they are and I will close this PR right now.

PR #163269, opened 2025-09-18
Update pytorch_sphinx_theme2 to latest hash

svekars:
The updated theme:
- Fixes articleBody in the json+ld that caused previous Google Search issues
- Other minor fixes
- 404.html fixes


albanD (approved):
Sounds good. Please mention in this PR what was added since the last pin so we can track things more easily.

PR #163270, opened 2025-09-18
[export] add node.meta["val"] when HOO's output is empty

yushangdi:
Summary: as title. Otherwise the verifier will complain that node.meta has no val

Test Plan: 

Differential Revision: D82691627

### DIFF
--- a/torch/_export/serde/serialize.py
+++ b/torch/_export/serde/serialize.py
@@ -2717,6 +2717,10 @@ class GraphModuleDeserializer(metaclass=Final):
     def deserialize_outputs(self, serialized_node: Node, fx_node: torch.fx.Node):
         # Check single value return
         if len(serialized_node.outputs) == 0:
+            if isinstance(
+                fx_node.target, torch._higher_order_ops.wrap.WrapWithSetGradEnabled
+            ):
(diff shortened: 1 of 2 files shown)


torch/_export/serde/serialize.py
```diff
@@ -2717,6 +2717,10 @@ def deserialize_sym_op_outputs(self, serialized_node: Node, fx_node: torch.fx.No
     def deserialize_outputs(self, serialized_node: Node, fx_node: torch.fx.Node):
         # Check single value return
         if len(serialized_node.outputs) == 0:
+            if isinstance(
+                fx_node.target, torch._higher_order_ops.wrap.WrapWithSetGradEnabled
+            ):
+                fx_node.meta["val"] = tuple()
```
angelayi, line 2723: I feel like maybe we should be modifying [this part](https://fburl.com/code/mer867kl) to add a dummy output?

yushangdi: @angelayi yeah. actually with your fix in https://github.com/pytorch/pytorch/pull/163295, I think we don't have to land this. Do you want to just add a dummy output there in your PR?

PR #163272, opened 2025-09-18
support a few features for vllm inductor precompile

dolpm:
add support to directly serialize to byes and deserialize from bytes so we can store in vllm's serializable callable.

### DIFF
--- a/torch/_inductor/async_compile.py
+++ b/torch/_inductor/async_compile.py
@@ -240,6 +240,12 @@ def async_compile_pool_manager():
         AsyncCompile.quiesce()
 
 
+def reload_kernel_in_parent(load_fn):
+    # Benchmark how often this happens
+    with dynamo_timed("reload_kernel_in_parent"):
+        return load_fn()
+
+
 class AsyncCompile:
     """
     Utilities to compile in thread pools or subprocess pools (in the case of Triton).
@@ -383,11 +389,6 @@ class AsyncCompile:
             _load_triton_kernel_from_source, kernel_name, source_code
         )
 
-        def reload_kernel_in_parent():
-            # Benchmark how often this happens
-            with dynamo_timed("reload_kernel_in_parent"):
-                return load_kernel()
-
         counters["inductor"]["async_compile_cache_miss"] += 1
 
         kernel_code_log.info("Triton Kernel:\n%s", source_code)
@@ -410,7 +411,9 @@ class AsyncCompile:
             if isinstance(future, StaticAutotunerFuture):
                 # Remove the future now that we've cache hit
                 CompiledTritonKernels.remove_future(source_code)
-                future.reload_kernel_from_src = reload_kernel_in_parent
+                future.reload_kernel_from_src = functools.partial(
+                    reload_kernel_in_parent, load_kernel
+                )
             if is_parallel:
                 return future
             else:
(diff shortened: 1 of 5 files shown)


torch/_inductor/standalone_compile.py
```diff
@@ -62,6 +62,75 @@ def __init__(
     def __call__(self, *args: Any) -> Any:
         return self._compiled_fn(*args)
 
+    def to_bytes(self) -> bytes:
```
zhxchen17: I think in the long term we don't want to use standalone_compile in vllm, but fine to do for now I guess. 

Could we have some basic unittest to verify to_bytes() and from_bytes() and finish a save/load roundtrip?

torch/_inductor/async_compile.py
```diff
@@ -240,6 +240,12 @@ def async_compile_pool_manager():
         AsyncCompile.quiesce()
 
 
+def reload_kernel_in_parent(load_fn):
```
dolpm, line 243: so that this can be pickled

torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -668,6 +668,7 @@ def prepare_for_caching(self) -> None:
                 result.kernel.cubin_raw = None
 
     def __getstate__(self) -> dict[str, Any]:
+        self.prepare_for_pickle()
```
dolpm: otherwise this assert will be thrown when pickling.

jamesjwu: Hmm I'm not 100% sure the behavior of putting this function here; does `copy.deepcopy` trigger `__getstate__`? 

I would double check `triton_bundler.py` and see how it calls `prepare_for_caching`/`prepare_for_pickle`; you don't want to change the existing caching behavior. 

(On a separate note, can you not just call this function before pickling instead of putting it in the pickle setup? I suppose that is a little more gross overall?)

torch/utils/_sympy/functions.py
```diff
@@ -1411,7 +1411,10 @@ def eval(cls, a, b):
                 return sympy.Integer(getattr(operator, real_op_name)(int(a), int(b)))
             return None
 
-    BitwiseFn.__name__ = "BitwiseFn_" + name
+    nm = "BitwiseFn_" + name
```
jamesjwu, line 1455: nit: I think this needs to be rebased past your next diff

torch/_inductor/standalone_compile.py
```diff
@@ -256,6 +259,9 @@ def handle_node(node: torch.fx.Node) -> None:
         )
         assert callable(compiled_fn)
 
+        if not save_artifacts:
+            return compiled_fn
```
jamesjwu, line 264: I think this changes existing behavior to return a different type for existing vllm callers, which will be an annoying BC issue between versions. Instead, let's introduce a new option in `options` for returning a serializable callable directly instead of artifacts? 

(I think there are other configs you will need to make this fully work anyway — things like bundling AOTAUtogradCache

dolpm: it's defaulted behavior will be the same, and this logic is gonna change the return type either way (unless we had a separate function all together, which seems excessive). i can update it, but not sure it will get rid of the BC concern.

torch/_inductor/runtime/triton_heuristics.py
dolpm: yeah, let's just add a custom reducer in vllm.

zhxchen17 (approved):
linters

PR #163276, opened 2025-09-18
[opaque_obj] Add set_payload + docs

angelayi:
* #163278
* #163277
* __->__ #163276
* #162660


torch/_library/opaque_object.py
```diff
@@ -3,14 +3,97 @@
...
+        ...     make_opaque,
+        ...     get_payload,
+        ...     set_payload,
+        ... )
+        >>>
+        >>> class OpaqueQueue:
+        >>>     def __init__(self, queue: list[torch.Tensor], init_tensor_: torch.Tensor) -> None:
+        >>>         super().__init__()
+        >>>         self.queue = queue
+        >>>         self.init_tensor_ = init_tensor_
```
zou3519: I would not recommend OpaqueQueue as the example for users. Might be better to include something that isn't a container of tensors. The pytree thing is usually better for container of tensors

For example, maybe Point with x, y being floats.

torch/csrc/jit/python/init.cpp
```diff
@@ -1883,6 +1883,16 @@ void initJITBindings(PyObject* module) {
         return customObj->getPayload();
       },
       R"doc(Returns the Python object stored on the given opaque object.)doc");
+  m.def(
+      "_set_opaque_object_payload",
+      [](py::object obj, py::object payload) {
+        auto typePtr =
+            torch::getCustomClass("__torch__.torch.classes.aten.OpaqueObject");
+        auto ivalue = torch::jit::toIValue(std::move(obj), typePtr);
```
zou3519, line 1893: what does the error message look like when you try to _set_opaque_object_payload on a scriptobject that is not a OpaqueObject? We want to check that that error message is reasonable

zou3519 (approved):
lgtm but would recommend a non-OpaqueQueue example. We should try to do something that is not a container of Tensors.

torch/csrc/jit/python/init.cpp
angelayi: It errors with
> RuntimeError: Tried to convert an IValue of type __torch__.torch.classes._TorchScriptTesting._TensorQueue (of Python compilation unit at: 0) to custom class type __torch__.torch.classes.aten.OpaqueObject (of Python compilation unit at: 0)

angelayi: I updated the error message on the python side to look a little nicer!

torch/_library/opaque_object.py
angelayi: Updated the example with an "rng" example! I couldn't think of a good example with "point" 😅

torch/csrc/jit/python/init.cpp
zou3519: nice

PR #163277, opened 2025-09-18
[opaque obj] Error for torch.library.custom_op infer_schema

angelayi:
Unsure how we can get infer_schema to infer the scriptObject type from just the type annotation, so for now will just error clearly and ask users to specify a schema.

* #163936
* #163714
* #163284
* #163278
* __->__ #163277
* #163279


torch/_library/infer_schema.py
```diff
@@ -124,7 +124,13 @@ def unstringify_type(ty: Union[type[object], str]) -> tuple[typing.Any, bool]:
         annotation_type, _ = unstringify_type(param.annotation)
 
         if annotation_type not in SUPPORTED_PARAM_TYPES:
-            if annotation_type.__origin__ is tuple:
+            if annotation_type == torch._C.ScriptObject:
```
zou3519, line 129: I think we should be able to get this to work.
The idea is to provide some new NewType annotation that behaves like "ScriptObject" (or Any) during typing, but can be uniquely identified by this function. We can call that torch.library.OpaqueObject or torch.library.OpaqueType. Bonus points if this is able to become a constructor that returns a ScriptObject but I'm not sure if that will work

https://mypy.readthedocs.io/en/stable/more_types.html#newtypes

torch/library.py
```diff
@@ -41,6 +42,7 @@
     "triton_op",
     "wrap_triton",
     "infer_schema",
+    "OpaqueType",
```
zou3519: Can you keep this private for now? (don't import into torch.library yet). We should have one future PR where all of these APIs (get/set payload too) become public (assuming they're not public yet, it's been a while since I looked at the code)

zou3519 (approved):

torch/library.py
angelayi: yes theyre not private yet!

PR #163278, opened 2025-09-18
[opaque_obj] Add make_fx tracing support

angelayi:
* #163936
* #163714
* #163284
* __->__ #163278
* #163277
* #163279


torch/_library/fake_class_registry.py
```diff
@@ -134,55 +135,67 @@ def maybe_to_fake_obj(
...
-                x,
+    if str(x._type()) == "__torch__.torch.classes.aten.OpaqueObject":
+        from torch._library.opaque_object import FakeOpaqueObject, get_payload
+
+        # We can't implement an __obj_flatten__ function because
+        # CustomClassHolder only allow us to register function signatures for
+        # objs that are valid IValues, and arbitrary py::objects/handles are not
+        # valid IValues (they need to be a CustomClassHolder).
+        payload = get_payload(x)
+        fake_x = FakeOpaqueObject(payload)
```
angelayi: technically I don't even need to fakify it.. but I'm unsure if I should just to match the programming model of the ScriptObjects

torch/_library/opaque_object.py
```diff
@@ -1,6 +1,20 @@
...
 
 import torch
+from torch._library.fake_class_registry import FakeScriptObject
+
+
+@torch._library.register_fake_class("aten::OpaqueObject")
+class FakeOpaqueObject:
+    def __init__(self, payload: Any) -> None:
+        self.payload: Any = payload
+
```
zou3519: FakeOpaqueObject should not actually have the payload on it. When someone tries to access the payload during fake tensor prop, we should error out.

The idea is that the Fake kernel should not depend on the payload at all. That makes the object truly opaque to compile and means that compile doesn't need to guard on it

test/test_opaque_obj.py
```diff
@@ -62,6 +79,14 @@ def pop_impl(q: torch._C.ScriptObject) -> torch.Tensor:
 
         self.lib.impl("queue_pop", pop_impl, "CompositeExplicitAutograd")
 
+        def pop_impl_fake(q: torch._C.ScriptObject) -> torch.Tensor:
+            # This is not accurate
+            ctx = torch._custom_op.impl.get_ctx()
+            u0 = ctx.create_unbacked_symint()
+            return torch.empty(u0)
```
zou3519: what's not accurate? that we don't know anything about the tensors in the queue? (put small explanation into the code)

torch/_library/opaque_object.py
```diff
@@ -81,6 +96,12 @@ def get_payload(opaque_object: torch._C.ScriptObject) -> Any:
         payload (Any): The Python object stored in the opaque object. This can
         be set with `set_payload()`.
     """
+    if isinstance(opaque_object, FakeScriptObject):
+        raise ValueError(
+            "OpaqueObjects are opaque, so therefore the contents should not be "
+            "visible to torch.compile, and the fake kernel should not depend "
+            "on the contents of the OpaqueObject at all."
+        )
```
zou3519, line 106: nit:
- can deduplicate this with the below
- ValueError("get_payload: this function was called with a FakeScriptObject. The implication is that you are calling get_payload inside of a fake kernel. The fake kernel should not depend on the contents of the OpaqueObject at all, so we're erroring out. if you need this functionality, consider creating a custom TorchBind Object instead (but note that this is more difficult, <link to tutorial>)") (something like that that has more context)

torch/_library/opaque_object.py
```diff
@@ -104,6 +125,13 @@ def set_payload(opaque_object: torch._C.ScriptObject, payload: Any) -> None:
         torch._C.ScriptObject: The opaque object that stores the given Python object.
         payload (Any): The Python object to store in the opaque object.
     """
+    if isinstance(opaque_object, FakeScriptObject):
```
zou3519, line 130: Can we add a test for this? a fake kernel that calls get/set payload and do an assertraisesregex for it

torch/_library/fake_class_registry.py
```diff
@@ -134,55 +135,64 @@ def maybe_to_fake_obj(
...
-                "Detected script object %s has aliasing relationship among its tensors. "
-                "Flattened obj: %s. Aliasing tensor indices: %s. "
-                "This is not supported and may cause unexpected behavior.",
-                x,
+    if str(x._type()) == "__torch__.torch.classes.aten.OpaqueObject":
+        from torch._library.opaque_object import FakeOpaqueObject
+
+        # In order to make OpaqueObjects truly opaque, the fake kernel should
+        # not depend on the contents of the OpaqueObject at all.
+        fake_x = FakeOpaqueObject()
```
zou3519, line 143: Add test that fakifying an OpaqueObject leads to something that doesn't have a real_obj?

zou3519 (approved):
lgtm, have minor comments, please read

test/test_opaque_obj.py
angelayi: I was thinking this isn't accurate because I'm assuming the tensors in the queue have rank 1?

zou3519: yup

torch/_library/fake_class_registry.py
angelayi: Hmm I don't think that path exists.. theres always a real_obj

PR #163279, opened 2025-09-18
[opaque_obj] Add __eq__ and __deepcopy__

angelayi:
* #163936
* #163714
* #163284
* #163278
* #163277
* __->__ #163279


test/test_opaque_obj.py
```diff
@@ -35,6 +36,14 @@ def size(self) -> int:
         self._size_counter += 1
         return len(self.queue)
 
+    def __eq__(self, other):
```
zou3519, line 26: Why do we need `__eq__` and `__deepcopy__` again? if it is for Dynamo guarding, I don't think Dynamo needs to guard against OpaqueObject much. Dynamo should just emit a guard that the object is an OpaqueObject

angelayi: Originally the way I implemented it, I ran into some uses of these functions. But now I don't think I need it anymore. But it seems like it would still be helpful to have these functions?

angelayi: 1. when adding support for infer_schema, [this test](https://github.com/pytorch/pytorch/blob/4535254c28242848a99d12d4248d90cedf164f1f/test/test_custom_ops.py#L986) compares the ScriptObject inputs
2. when we create FakeScriptObjects we deepcopy the ScriptObject ([code ptr](https://github.com/pytorch/pytorch/blob/4535254c28242848a99d12d4248d90cedf164f1f/torch/_library/fake_class_registry.py#L22-L29)). This won't error but it'll just warn.

zou3519: why does the fake script object carry around a copy of the real one? @ydwu4

zou3519: The way I expected Fakification of an OpaqueObject is that we return some sort of FakeScriptObject that disallows access of the payload (and any other fields) of the OpaqueObject.

zou3519 (commented):
code looks reasonable, i'm gonna want tests for the fakeification behavior later

zou3519 (approved):

test/test_opaque_obj.py
angelayi: After chatting with Yidi and looking around the codebase, the real obj is mainly used for getting the schema of call_torchbind, and used by inductor to get the number of bytes in the buffer. It doesn't seem to be used by any meta kernels

PR #163281, opened 2025-09-18
[MPS] Compute `offset2bag/bag_size/max_indices` in `_embedding_bag`

kurtamohler:
* __->__ #163281

Part of #162270


aten/src/ATen/native/mps/kernels/EmbeddingBag.metal
```diff
@@ -22,55 +22,65 @@ struct ReductionOpInit<EmbeddingBagMode::MAX, T> {
...
 struct ReductionOp {
+  inline opmath_t<T> operator()(opmath_t<T> weight_val, opmath_t<T> out_val) {
+    return weight_val + out_val;
+  }
+};
+
+template <typename T>
+struct ReductionOp<EmbeddingBagMode::MAX, T> {
+  inline opmath_t<T> operator()(opmath_t<T> weight_val, opmath_t<T> out_val) {
+    return max(weight_val, out_val);
```
malfet: Q: Which max are you using there? one from `metal::` or one from `c10::metal::`? 
I don't know if embedding bug supposed to carry about NaN, but if it is, make sure to use `c10::metal::` wrapper, as regular one will not be able to handle NaN correctly

aten/src/ATen/native/mps/kernels/EmbeddingBag.metal
```diff
@@ -22,55 +22,65 @@ struct ReductionOpInit<EmbeddingBagMode::MAX, T> {
...
+      thread I& max_idx,
+      I weight_idx,
+      bool pad) {}
 };
 
-template <typename T>
-struct ReductionOp<EmbeddingBagMode::MAX, T> {
+template <typename T, typename I>
+struct MaybeCalcMaxIndex<EmbeddingBagMode::MAX, T, I> {
   inline opmath_t<T> operator()(
```
malfet: Should it return void?

aten/src/ATen/native/mps/kernels/EmbeddingBag.metal
```diff
@@ -22,55 +22,65 @@ struct ReductionOpInit<EmbeddingBagMode::MAX, T> {
...
       opmath_t<T> out_val,
-      uint32_t,
-      constant T*,
-      uint32_t) {
-    return max(static_cast<opmath_t<T>>(weight_val), out_val);
+      opmath_t<T> new_out_val,
+      thread I& max_idx,
+      I weight_idx,
+      bool pad) {
+    max_idx = (pad || new_out_val == out_val) ? max_idx : weight_idx;
```
malfet: I'm not sure if compiler is smart enough, but wouldn't it be better to do something like
```suggestion
   if (!pad && new_out_val != out_val) {
         max_idx = weight_idx;
   }
```

test/test_mps.py
```diff
@@ -6940,6 +6940,61 @@ def helper(value, dim, index, idx_dtype=torch.int32):
         with self.assertRaisesRegex(RuntimeError, "Index to scalar can have only 1 value"):
             helper(22, 0, [])
 
+    # TODO: This test can be removed once the backward pass of embedding_bag is
```
malfet, line 6943: Hmm, what stops us from running existing forward tests from op_info and just add emebdding_bag to GRAD_FAILURES?

kurtamohler: We're already running the `torch.nn.functional.embedding_bag` opinfo tests, but that function does not return offset2bag, bag_size, and max_indices. There currently is no forward mode opinfo test that checks those

aten/src/ATen/native/mps/kernels/EmbeddingBag.metal
kurtamohler: I'm not sure if the compiler is smart enough to avoid the thread divergence either, but the Metal documentation does recommend avoiding if statements that could potentially cause divergence: [link](https://developer.apple.com/documentation/metal/creating-threads-and-threadgroups?utm_source=chatgpt.com#:~:text=Avoid%20writing%20code%20that%20could%20cause%20your%20kernel%20function%20to%20diverge%3B%20that%20is%2C%20to%20follow%20different%20code%20paths.%20A%20typical%20example%20of%20divergence%20is%20caused%20by%20using%20an%20if%20statement.)

Apparently, XCode has a thread divergence counter in the profiler, so it would be possible to check this. But I don't have access to a graphical interface on the machine I'm using

aten/src/ATen/native/mps/kernels/EmbeddingBag.metal
kurtamohler: Good catch. Looks like the only way to make it match the CPU impl is to use `metal::max` and also initialize `out_val` with nan. I'll make those changes and add some nans to the test

kurtamohler: Ok I've updated it to have the same nan behavior as the CPU impl. It doesn't use `max` at all, and instead uses comparison. Let me know what you think

aten/src/ATen/native/mps/kernels/EmbeddingBag.metal
malfet: Let me experiment locally and I'll let you know

aten/src/ATen/native/mps/kernels/EmbeddingBag.metal
malfet: By the way, I don't know which one is faster: ternary or max

malfet (approved):

aten/src/ATen/native/mps/kernels/EmbeddingBag.metal
```diff
@@ -23,54 +23,72 @@ struct ReductionOpInit<EmbeddingBagMode::MAX, T> {
...
+  }
+};
+
+template <typename T>
+struct ReductionOp<EmbeddingBagMode::MAX, T> {
+  inline opmath_t<T> operator()(
+      opmath_t<T> weight_val,
+      opmath_t<T> out_val,
+      bool is_first) {
+    return (is_first || weight_val > out_val) ? weight_val : out_val;
```
malfet, line 39: This would also be a non-nan preserving,  but I guess it's the same behavior as CPU

PR #163282, opened 2025-09-18
[Fix] Restrict stride normalization to 1D tensors on export

### LINKED ISSUE #163274: [DLPack] Further restrict the normalization condition for DLPack Export
Previously in https://github.com/pytorch/pytorch/issues/83069, we decided to normalize the DLPack tensor if the tensor's shape[i] equals 1, however. Such normalization can cause issue for down stream consumers. Specifically, say a column major tensor

```python
shape=(1024, 1), stride=(1, 1024)
```

Exporting this tensor will trigger normalization, and as a result stride becomes `stride=(1, 1)`, because it is non-contiguous, and there is opportunity to do so. While it is not wrong(as a matter of fact, any `stride=(1, x)` is valid),  `stride=(1, 1)` is usually harder to handle for many cases, on one thing, it is harder to detect for example the tensor is in column major. Stride settings in cases like GPU tma also do not like such duplicated strides.

Reading the original post, seems the original intention is to enable the use-case of single dimension array`shape=(1,)`, `stride=(x,)` and normalize that to  `stride=(1,)`. Which could make sense. While the multi dimensional normalization can be counter productive.

As a result, we think ideally to restrict normalization to one dimension array, it will also further speedup the conversion

Kathryn-cat:
This change restricts the DLPack stride normalization to apply only to 1D tensors of shape (1,).

### Rationale
The previous implementation normalized the strides for any multi-dimensional tensor containing a dimension of size 1. While well-intentioned, this "over-normalization" discards critical memory layout information, causing issues for downstream consumers who rely on strides to infer alignment and contiguity.

For example:

* A row-major tensor with `shape=(1, 128)` and `stride=(128, 1)` would be incorrectly normalized to `stride=(1, 1)`.

* A column-major tensor with `shape=(1024, 1)` and `stride=(1, 1024)` would also be normalized to `stride=(1, 1)`.

This loss of stride information makes it impossible for consumers to detect the original memory layout (e.g., row-major vs. column-major) and breaks assumptions about memory alignment needed for optimized indexing or specialized hardware APIs like GPU TMA.

The original intent of the normalization was to handle the simple case of a 1D tensor with shape=(1,) and a non-standard stride. This fix reverts to that specific, non-problematic behavior, ensuring that multi-dimensional tensors retain their precise stride information during DLPack export.

### Related Issues
#163274

### DIFF
--- a/aten/src/ATen/DLConvertor.cpp
+++ b/aten/src/ATen/DLConvertor.cpp
@@ -401,7 +401,7 @@ T* toDLPackImpl(const Tensor& src) {
   // The following code detects whether the src follows
   // a continuous pattern. If the src follows such pattern (common-case)
   // then we do not need to normalize the strides.
-  bool need_normalize_strides = src.ndim() == 1 && src.size(0) == 1 && src.stride(0) != 1;
+  bool need_normalize_strides = src.dim() == 1 && src.size(0) == 1 && src.stride(0) != 1;
   // less common case, try normalizing the strides
   if (need_normalize_strides) {
     // create a new tensor with possibly normalized strides
--- a/aten/src/ATen/DLConvertor.cpp
+++ b/aten/src/ATen/DLConvertor.cpp
@@ -401,30 +401,13 @@ T* toDLPackImpl(const Tensor& src) {
   // The following code detects whether the src follows
   // a continuous pattern. If the src follows such pattern (common-case)
   // then we do not need to normalize the strides.
-  bool need_normalize_strides = false;
-  int64_t expected_stride = 1;
-  for (int i = src.dim() - 1; i >= 0; i--) {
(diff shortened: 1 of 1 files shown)


eqy (approved):

Kathryn-cat:
hi @ngimel , we have restricted normalizing strides only on 1-D tensor with shape (1,). In your example, tensor of size[1,1,1] is not normalized mainly due to loss of information on the original memory layout information, which might break some assumptions on downstream usage. For example, GPU TMA instructions do not allow tensor with strides [1,1,1].

ngimel:
Then please let's not normalize!

tqchen:
that sounds good to me, maybe @Kathryn-cat can send a followup pr?

PR #163284, opened 2025-09-18
[effects] Add register_effectful_op

angelayi:
* #167364
* #167363
* #167245
* #167231
* #167230
* #163936
* #163714
* __->__ #163284

Refactored register_effectful_op to return a handler to match how fake kernels are registered. This makes it easier to deregister effects


torch/_library/effects.py
```diff
@@ -0,0 +1,33 @@
...
+]
+
+
+def register_effectful_op(
+    op: _op_identifier,
+    effect: Optional[_EffectType] = _EffectType.ORDERED,
+):
+    r"""
+    Registers an effect for this operator.
+
```
zou3519, line 16: should expose in torch.library docs somewhere, https://docs.pytorch.org/docs/stable/library.html

torch/_library/effects.py
```diff
@@ -0,0 +1,33 @@
...
+_op_identifier = Union[
+    str,
+    "torch._ops.OpOverload",
+    "torch._library.custom_ops.CustomOpDef",
+    "torch._ops.HigherOrderOperator",
+]
+
+
+def register_effectful_op(
```
zou3519: As a torch.library API, this should probably accept an optional torch.library object to tie the lifetime of the registration to. See torch.library.register_fake for inspiration.

I would recommend doing something like the following https://github.com/pytorch/pytorch/blob/main/torch/_library/simple_registry.py#L16.

Basically, when an operator gets deleted, it also needs to delete the side effect annotation. SimpleRegistry manages that a bit but it is kind of complicated

test/higher_order_ops/test_with_effects.py
```diff
@@ -300,49 +301,30 @@ def f(x):
...
 
-            # Pytorch custorm op implementation
-            @torch.library.impl(
-                "mylib::record_scalar_tensor",
-                "CompositeExplicitAutograd",
-                lib=lib,
-            )
-            def record_scalar_tensor(x, prefix):
+            # Pytorch custom op implementation
+            @torch.library.custom_op("mylib::record_scalar_tensor", mutates_args=())
```
angelayi, line 315: I turned this custom op into a @torch.library.custom_op since I didnt see this anywhere else in the file

torch/library.py
```diff
@@ -1065,6 +1072,48 @@ def register(func):
         return register(func)
 
 
+def _register_effectful_op(
+    op: _op_identifier,
+    effect: Optional[EffectType],
+    *,
+    lib: Optional[Library] = None,
+) -> None:
```
zou3519, line 1090: do you have a plan to make this public?

torch/_library/custom_ops.py
```diff
@@ -471,6 +472,11 @@ def register_fake(self, fn: Callable, /) -> Callable:
         self._abstract_fn = fn
         return fn
 
+    def register_effect(self, effect: Optional[EffectType]) -> None:
+        from torch._higher_order_ops.effects import _register_effectful_op
+
+        _ = _register_effectful_op(self, effect)
```
zou3519: why `_ =` ?

torch/_higher_order_ops/effects.py
```diff
@@ -17,38 +18,90 @@
...
+    "torch._library.custom_ops.CustomOpDef",
+    "torch._ops.HigherOrderOperator",
+]
+OpType = Union["torch._ops.HigherOrderOperator", "torch._ops.OpOverload"]
 
+_EffectType = EffectType
 
-OpType = Union[torch._ops.HigherOrderOperator, torch._ops.OpOverload]
 
+class SideEffectRegistry:
```
zou3519: I'm confused, can't we just add a `effect` field here? https://github.com/pytorch/pytorch/blob/main/torch/_library/simple_registry.py#L53 
 
That's what the register_fake stuff is doing, it sets a field there (and the simple_registry is the source of truth)

torch/library.py
```diff
@@ -398,6 +399,12 @@ def fallback(self, fn, dispatch_key="", *, with_keyset=False):
 
         self.m.fallback(dispatch_key, fn, with_keyset)
 
+    def _register_effectful_op(self, op_name: str, effect: Optional[EffectType]):
```
zou3519, line 402: Add some docs, the other functions here have a doc string

zou3519 (commented):
I'd prefer the source of truth to live in simple_registry.py (that's where our other sources of truth live) instead of maintaining a separate store for this one.

That way there's a central way to see the registered metadata for an operator

torch/library.py
angelayi: yes! although I wasn't sure if that should be done now?

torch/_higher_order_ops/effects.py
angelayi: sorry that makes sense! updated!

torch/library.py
zou3519: well you have until 2.10 to decide

torch/_library/simple_registry.py
```diff
@@ -51,6 +54,23 @@ def __init__(self, qualname: str) -> None:
...
+            overload = ""
+        try:
+            schema = torch._C._get_schema(qualname, overload)
+        except RuntimeError:
+            # In the HOO case
+            return
+        for arg in schema.arguments:
+            if isinstance(arg.type, torch.ClassType):
+                self.effect = EffectType.ORDERED
+                break
```
zou3519: this might be the first time i've seen a break statement in python lol

torch/_library/simple_registry.py
```diff
@@ -51,6 +54,23 @@ def __init__(self, qualname: str) -> None:
...
+        if len(qualname.split(".")) > 1:
+            overload = qualname.split(".")[1]
+            qualname = qualname.split(".")[0]
+        else:
+            overload = ""
+        try:
+            schema = torch._C._get_schema(qualname, overload)
+        except RuntimeError:
+            # In the HOO case
+            return
```
zou3519: er I think we can assume that all higher order ops have the namespace "higher_order". Might be nicer than a try-catch esp if the get_schema fails for another reason

torch/_library/custom_ops.py
```diff
@@ -471,6 +472,10 @@ def register_fake(self, fn: Callable, /) -> Callable:
         self._abstract_fn = fn
         return fn
 
+    def register_effect(self, effect: Optional[EffectType]) -> None:
+        entry = torch._library.simple_registry.singleton.find(self._qualname)
+        entry.effect = effect
```
zou3519: why isn't this registering something to a library?

zou3519: the main thing is: if someone deletes this torch.libary.custom_op (or if it goes out of scope), I want to make sure the entry goes away (or the effect is reset)

zou3519 (approved):

PR #163285, opened 2025-09-18
Raise TypeError in torch.export.load when input isn't a file or buffer

### LINKED ISSUE #163040: [torch.export.load] Passing an `ExportedProgram` raises `AttributeError: 'ExportedProgram' object has no attribute 'tell'` instead of a clear `TypeError`
### Summary

Calling `torch.export.load(ep)` where `ep` is an `ExportedProgram` produces an `AttributeError` originating from file-like handling (`.tell`) inside the PT2 archive loader. The API does print a warning saying the argument must be a buffer or a `.pt2` file, but it still proceeds and then crashes with a non-actionable error.

This should fail early with a clear `TypeError` (or `ValueError`) stating the accepted input types, rather than attempting to treat an `ExportedProgram` as a file object.

### Minimal Reproducer
```python
# minimal_repro_export_load_misuse.py
import io
import torch
import torch.nn as nn

class Tiny(nn.Module):
    def forward(self, x):
        return F.relu(x)

def main():
    print("torch.__version__ =", torch.__version__)
    m = Tiny().eval()
    x = torch.randn(2, 3)

    # 1) Normal export: returns ExportedProgram
    ep = torch.export.export(m, (x,))
    print("type(ep) =", type(ep).__name__)

    # 2) Misuse: pass ExportedProgram directly to load()
    print("\n[Misuse] torch.export.load(ep) -> expecting a clean TypeError...")
    try:
        torch.export.load(ep)
    except Exception as e:
        print("Caught:", type(e).__name__, e)

if __name__ == "__main__":
    main()
```

### Observed Output (excerpt)
```
torch.__version__ = 2.9.0a0+git31d5c67
type(ep) = ExportedProgram

[Misuse] torch.export.load(ep) -> expecting a clean TypeError...
W**** torch/export/pt2_archive/_package.py:943] Unable to load package. f must be a buffer or a file ending in .pt2. Instead got {ExportedProgram: ...}
Caught: AttributeError 'ExportedProgram' object has no attribute 'tell'
```

### Versions

```
PyTorch version:  2.8.0a0+gitba56102 & 2.9.0a0+git31d5c67
Is debug build: True

CMake version: version 3.22.1
Libc version: glibc-2.35

Is CUDA available: True
```

dsashidh:
Fixes #163040 

Replaces a vague AttributeError with a clearer TypeError when torhc.export.load( ) is misused. For example, calling it with an ExportedProgram now raises immediately insteald of hitting .tell() downstream. The check I added matches the existed warning and turns it into a proper error as suggested in the TODO comment that existed before my change.

### DIFF
--- a/torch/export/pt2_archive/_package.py
+++ b/torch/export/pt2_archive/_package.py
@@ -989,20 +989,25 @@ def load_pt2(
 
     from torch._inductor.cpp_builder import normalize_path_separator
 
-    is_buffer = isinstance(f, (io.IOBase, IO))
-    is_path = isinstance(f, (str, os.PathLike))
-
-    if not ((is_buffer and f.readable() and f.seekable()) or is_path):
+    if isinstance(f, (io.IOBase, IO)):
+        if not (f.readable() and f.seekable()):
+            raise TypeError(
+                "Unable to load package. `f` must be a buffer (i.e. a readable and "
+                "seekable file-like object) or a string path. "
+                f"Instead got {type(f).__name__}."
+            )
+    elif isinstance(f, (str, os.PathLike)):
+        if not os.fspath(f).endswith(".pt2"):
+            logger.warning("Path does not end with .pt2; proceeding anyway: {%s}", f)
+        f = os.fspath(f)
+    else:
         raise TypeError(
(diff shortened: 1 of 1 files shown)


angelayi (approved):
Thanks so much!

PR #163288, opened 2025-09-18
[DeviceMesh] Add extra check in flatten result cache lookup

fduwjj:
* #163358
* #161224
* #163213
* #163367
* __->__ #163288
* #163212

while refactoring DeviceMesh bookkeeping, we found that there is one corner case which we just don't check whether the dims to be flattened into is same as the dims which an existing flattened name maps to. So we need to add extra cases in the unit test and extra check logic in the code.


wz337:
LGTM! Thanks for adding extra check

wz337 (approved):

ezyang (approved):

test/distributed/test_device_mesh.py
```diff
@@ -903,6 +903,18 @@ def test_flatten_mesh_3d(self):
         cp_tp_mesh._flatten("dummy")
         self.assertEqual(mesh_3d["dummy"].mesh_dim_names[0], "dummy")
 
+        # Test flatten into an existing mesh_dim_name inside the mesh
+        with self.assertRaisesRegex(
+            RuntimeError,
+            "dp already exists for submesh of the DeviceMes",
```
fegin:
```suggestion
            "dp already exists for submesh of the DeviceMesh",
```

fegin (approved):
Minor typo, which you can fix in another PR.

torch/distributed/device_mesh.py
```diff
@@ -185,7 +185,16 @@ def create_flatten_mesh(
...
                 and mesh_dim_name in self.root_to_flatten_mapping[root_mesh]
             ):
-                return self.root_to_flatten_mapping[root_mesh][mesh_dim_name]
+                if (
+                    tuple(flatten_dims_in_root)
+                    == self.flatten_name_to_root_dims[root_mesh][mesh_dim_name]
+                ):
+                    return self.root_to_flatten_mapping[root_mesh][mesh_dim_name]
+                else:
+                    raise RuntimeError(
```
Skylion007: Should this raise a RuntimeError, a ValueError, or AssertionError?

Skylion007: I think the latter two are more appropriate than RuntimeError, right?

test/distributed/test_device_mesh.py
fduwjj: opps, I can fix it in this PR.

torch/distributed/device_mesh.py
fduwjj: sure, I can make it `ValueError`. We don't really pay attention to Error type inside DeviceMesh. maybe I will need to have a round of audit of its error type down the road.

PR #163291, opened 2025-09-18
[DTensor] Fix adding Partial(sum) with Scalar

### LINKED ISSUE #163193: Incorrect results when adding partial DTensor with a scalar
### 🐛 Describe the bug

```python
from torch.distributed._tensor import Shard, distribute_tensor, init_device_mesh, DTensor
import torch
from torch.nn import functional as F
from torch import distributed as dist


if __name__ == '__main__':
    dist.init_process_group(backend="gloo")
    rank = dist.get_rank()
    mesh = init_device_mesh('cpu', (2,))
    tensor = torch.arange(10).reshape(10, 1).float()
    dt = distribute_tensor(tensor, device_mesh=mesh, placements=[Shard(0)])

    tensor10 = torch.tensor(10)
    value10 = 10

    sss = dt.sum()
    res = [sss + tensor10, sss + value10]
    s = '\n'.join(str((r, r.full_tensor() if isinstance(r, DTensor) else r)) for r in res)
    if rank == 1:
        import time; time.sleep(1)
    print(s)
```
prints:
```
(DTensor(local_tensor=15.0, device_mesh=DeviceMesh((2,), device: 'cpu', stride: (1,)), placements=(Partial(sum),)), tensor(55.))
(DTensor(local_tensor=20.0, device_mesh=DeviceMesh((2,), device: 'cpu', stride: (1,)), placements=(Partial(sum),)), tensor(65.))
(DTensor(local_tensor=40.0, device_mesh=DeviceMesh((2,), device: 'cpu', stride: (1,)), placements=(Partial(sum),)), tensor(55.))
(DTensor(local_tensor=45.0, device_mesh=DeviceMesh((2,), device: 'cpu', stride: (1,)), placements=(Partial(sum),)), tensor(65.))
```

### Versions

with latest torch nightly

SherlockNoMad:
Fixes #163193

When adding Partial(Sum) with a ScalarTensor, a few thing are happening
1. ScalarTensor is auto converted as Replicated DTensor
2. Propagator decide to redistribute R -> Partial(sum)
3. This result as the div by num_shards
```
    aten::add.Tensor(dt: f32[][P], t: i64[])
      redistribute_input(1, [R] -> [P])
        aten::div.Tensor(t: i64[], 4)
      aten::add.Tensor(t: f32[], t: f32[])
```

When adding Partial(Sum) with a Scalar
1. Scalar remains as scalar 
2. Propagator only sees a single Partial DTensor, and sets output placement also as Partial
3. As a result, scalar is added locally in all ranks, resulting in a wrong global result.
```
    aten::add.Tensor(dt: f32[][P], 2)
      aten::add.Tensor(t: f32[], 2)
```

In contrast, aten.sub(Partial, ScalarTensor) would redistribute Partial tensor as Replicate first. Although, it could have follow the same strategy used in aten.add. 
```
    aten::sub.Tensor(dt: f32[][P], t: i64[])
      redistribute_input(0, [P] -> [R])
        _c10d_functional::all_reduce(t: f32[], sum, 0)
        _c10d_functional::wait_tensor(t: f32[])
      aten::sub.Tensor(t: f32[], t: i64[])
```

### DIFF
--- a/torch/distributed/tensor/_dispatch.py
+++ b/torch/distributed/tensor/_dispatch.py
@@ -49,6 +49,22 @@ def is_same_size_handler(
     return lhs.shape == rhs.shape
 
 
+def add_handler_predicate(
+    op_call: torch._ops.OpOverload, args: tuple[object, ...], kwargs: dict[str, object]
+) -> bool:
+    return not isinstance(args[1], torch.Tensor)
+
+
+def add_scalar_handler(
+    op_call: torch._ops.OpOverload,
+    args: tuple[object, ...],
+    kwargs: dict[str, object],
+) -> object:
+    assert add_handler_predicate(op_call, args, kwargs)
+    new_args = (args[0], torch.tensor(args[1]))
+    return op_call(*new_args, **kwargs)
+
+
 def found_inf_reduce_handler(
     op_call: torch._ops.OpOverload,
     args: tuple[object, ...],
@@ -125,6 +141,10 @@ class OpDispatcher:
             aten._amp_foreach_non_finite_check_and_unscale_.default: found_inf_reduce_handler,
         }
 
+        self._custom_op_handlers_with_condition = {
+            aten.add.Tensor: (add_handler_predicate, add_scalar_handler),
+        }
+
     # This flag is used internally to control whether we treat the torch.Tensor(non-DTensor)
     # as implicitly replicated or we throw error to user.
     # NOTE: It is EXTREMELY UNSAFE to turn this flag on by default so we intentionally leave
@@ -151,6 +171,11 @@ class OpDispatcher:
         """
         if op_call in self._custom_op_handlers:
             return self._custom_op_handlers[op_call](op_call, args, kwargs)  # type: ignore[operator]
+
+        if op_call in self._custom_op_handlers_with_condition:
+            condition, handler = self._custom_op_handlers_with_condition[op_call]
+            if condition(op_call, args, kwargs):
+                return handler(op_call, args, kwargs)
 
         # extract local tensor and sharding infos to a OpInfo
         op_info = self.unwrap_to_op_info(op_call, args, kwargs)
--- a/torch/distributed/tensor/_dispatch.py
+++ b/torch/distributed/tensor/_dispatch.py
@@ -61,7 +61,7 @@ def add_scalar_handler(
     kwargs: dict[str, object],
 ) -> object:
     assert add_handler_predicate(op_call, args, kwargs)
-    new_args = (args[0], torch.tensor(args[1]))
+    new_args = (args[0], torch.tensor(args[1], device=args[0].device))  # type: ignore[attr-defined]
     return op_call(*new_args, **kwargs)
 
 
--- a/test/distributed/tensor/test_dtensor_ops.py
+++ b/test/distributed/tensor/test_dtensor_ops.py
@@ -649,16 +649,34 @@ class TestDTensorOps(DTensorOpTestBase):
                 else:
                     print(f"xfail('{opinfo.name}'),")
 
-    def test_partial_sum_add_scalar(self):
+    def test_pointwise_partial_scalar(self):
         self.mesh = init_device_mesh(DEVICE_TYPE, (self.world_size,))
 
-        tensor = torch.arange(10, dtype=torch.float, device=DEVICE_TYPE)
+        # Avoid having 0, so that tensor.prod() is not zero.
+        tensor = torch.arange(1, 12, dtype=torch.float, device=DEVICE_TYPE)
         dt = distribute_tensor(tensor, self.mesh, [Shard(0)])
 
-        def func(t):
-            return t.sum() + 1
+        funcs = [
+            lambda t: t.sum() + 2,
+            lambda t: t.sum() - 2,
+            lambda t: t.sum() * 2,
+            lambda t: t.sum() / 2,
+            lambda t: t.prod() + 2,
+            lambda t: t.prod() - 2,
+            lambda t: t.prod() * 2,
+            lambda t: t.prod() / 2,
+            # TODO: fix the following cases
+            # lambda t: t.mean() + 2,  # AssertionError: only support replicate to PartialSUM for now!
+            # lambda t: t.mean() - 2,  # AssertionError: Scalars are not close!
+            # lambda t: t.mean() * 2,  # AssertionError: Scalars are not close!
+            # lambda t: t.mean() / 2,  # AssertionError: Scalars are not close!
+            # lambda t: t.sum().copy_(2),  # AssertionError: Scalars are not close!
+            # lambda t: t.sum().clamp_(max=2),  # AssertionError: Scalars are not close!
+            # lambda t: t.sum().clamp_(max=torch.tensor(2)),  # AssertionError: Scalars are not close!
+        ]
 
-        self.assertEqual(func(dt).full_tensor(), func(tensor))
+        for func in funcs:
+            self.assertEqual(func(dt).full_tensor(), func(tensor))
 
     def test_one_hot(self):
         ops = [op for op in op_db if op.name == "nn.functional.one_hot"]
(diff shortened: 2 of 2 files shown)


dcci (approved):

torch/distributed/tensor/_dispatch.py
```diff
@@ -125,6 +141,10 @@ def __init__(self) -> None:
             aten._amp_foreach_non_finite_check_and_unscale_.default: found_inf_reduce_handler,
         }
 
+        self._custom_op_handlers_with_condition = {
+            aten.add.Tensor: (add_handler_predicate, add_scalar_handler),
```
ppwwyyxx, line 145: will the same issue happen for other ops? I wonder whether this is sufficient, or do we have a more comprehensive way to test more ops?

SherlockNoMad: I tried aten.sub(Paritla(sum), scalar) and aten.mul(Parital(prod), scalar), neither of them have the problem.

aten.add is special, coz it has a "linearity" of 1. See https://github.com/pytorch/pytorch/blob/main/torch/distributed/tensor/_ops/_pointwise_ops.py#L414-L425

For aten.add, it's currently redistributing ScalarTensor from Replicated() to Partial(), resulting as a "divide by number shard" op. 

Both aten.sub and atens.mul are redistributing from Partial() to Replicate(), which is not ideal. 

I will add some test case coverage.

dcci: I would suggest we fix the case at hand and also think about more extensive coverage. I have several ideas I can share.

ppwwyyxx: Thanks for the explanation! It would be helpful to briefly comment here why `add` is special.

SherlockNoMad: > I would suggest we fix the case at hand and also think about more extensive coverage. I have several ideas I can share.

yea, let me land this first, and let's talk about coverage. 

Here's the failure cases I have collected so far. 
```
        funcs = [
            lambda t: t.sum() + 2,
            lambda t: t.sum() - 2,
            lambda t: t.sum() * 2,
            lambda t: t.sum() / 2,
            lambda t: t.prod() + 2,
            lambda t: t.prod() - 2,
            lambda t: t.prod() * 2,
            lambda t: t.prod() / 2,
            # TODO: fix the following cases
            # lambda t: t.mean() + 2,  # AssertionError: only support replicate to PartialSUM for now!
            # lambda t: t.mean() - 2,  # AssertionError: Scalars are not close!
            # lambda t: t.mean() * 2,  # AssertionError: Scalars are not close!
            # lambda t: t.mean() / 2,  # AssertionError: Scalars are not close!
            # lambda t: t.sum().copy_(2),  # AssertionError: Scalars are not close!
            # lambda t: t.sum().clamp_(max=2),  # AssertionError: Scalars are not close!
            # lambda t: t.sum().clamp_(max=torch.tensor(2)),  # AssertionError: Scalars are not close!
        ]
```

torch/distributed/tensor/_dispatch.py
```diff
@@ -49,6 +49,22 @@ def is_same_size_handler(
...
+    return not isinstance(args[1], torch.Tensor)
+
+
+def add_scalar_handler(
+    op_call: torch._ops.OpOverload,
+    args: tuple[object, ...],
+    kwargs: dict[str, object],
+) -> object:
+    assert add_handler_predicate(op_call, args, kwargs)
+    new_args = (args[0], torch.tensor(args[1]))
```
ppwwyyxx: should it be created on the same device as args[0]?

SherlockNoMad: good catch.

tianyu-l:
@ezyang for review

torch/distributed/tensor/_dispatch.py
```diff
@@ -152,6 +172,11 @@ def dispatch(
         if op_call in self._custom_op_handlers:
             return self._custom_op_handlers[op_call](op_call, args, kwargs)  # type: ignore[operator]
 
+        if op_call in self._custom_op_handlers_with_condition:
+            condition, handler = self._custom_op_handlers_with_condition[op_call]
+            if condition(op_call, args, kwargs):
+                return handler(op_call, args, kwargs)
```
ezyang, line 178: Hmm, I'm not a huge fan of adding another custom op handler here. Is this just the easiest way you thought of how to do it, or is there a reason we cannot fix the preexisting logic?

ezyang: Two reasons I'm not a fan:
1. This is an extra dict lookup in the hotpath for DTensor where we have been explicitly working on reducing overhead here (cc @swolchok). It's not a big cost, but it is a cost.
2. Falling through here there is code that "handles" the scalar case, but it is buggy and we are just bandaiding over it, so I'd like to be convinced that the inner code is unfixable.

torch/distributed/tensor/_dispatch.py
ezyang: I don't think we should rush this PR, especially since we're not fixing all of the broken cases. Let's talk it through.

torch/distributed/tensor/_dispatch.py
ezyang: fwiw, it doesn't matter too much, pointwise ops also support 0d cpu scalar. We do have to be careful that we didn't change the promotion logic though

torch/distributed/tensor/_dispatch.py
ezyang: > For aten.add, it's currently redistributing ScalarTensor from Replicated() to Partial(), resulting as a "divide by number shard" op.

I didn't understand the explanation here. What exactly is the series of events that leads to the incorrect result?

SherlockNoMad: See updated summary for what's going now.

torch/distributed/tensor/_dispatch.py
swolchok: > extra dict lookup in the hotpath for DTensor where we have been explicitly working on reducing overhead here

supposing your other concerns about this PR don't end up holding, it seems like at minimum we could widen `_custom_op_handlers` to always include a condition instead of introducing a new dict, make the condition `None` if it is meant to always apply, and then check `if condition is None or condition(op_call, args, kwargs):`.

(by the way, I was going to complain about the double dictionary lookup in the success case, but it seems that it's still faster than using dict.get:

```
> python3.13 -m timeit --setup 'k = "key"; d = {"key": 7}
                         def noop(x): pass' 'if k in d: noop(d[k])'
10000000 loops, best of 5: 24.4 nsec per loop
> python3.13 -m timeit --setup 'k = "key"; d = {"key": 7}
                         def noop(x): pass' 'v = d.get(k)
                         if v is not None: noop(v)'
10000000 loops, best of 5: 28.1 nsec per loop
```

SherlockNoMad: @ezyang 

I took a closer look, there are 2 ways to support Partial(sum) + Scalar
1. Convert 1st input, from partial to replicate. 
This would incur a communication, but I think I can adjust shard propagator to change the the input placement.
2. Convert 2nd input from "replicated" into "Partial(sum)". 
As 2nd input is a Scalar, we need to convert it into a Tensor somehow. It gets pretty cumbersome if I were to do the scalar-to-tensor conversion int the middle of ShardingPropagation. Sharding Propagation also assume that scalar does not affect shard propagation, i.e. Scalar doesn't even have a DTensorSpce in OpSchema. See
```
class OpSchema:
    def args_spec(self) -> tuple[DTensorSpec, ...]:
        """
        args_spec: Tuple[DTensorSpec, ...]: contains a clean list of args spec list
            with NO non-DTensor positional arguments (i.e. int/float/tuple, etc)
            mainly used by sharding propagation to propagate the output spec
        """
```
This implies that DTensor compute is likely wrong for many ops with Scalar inputs.

ezyang: So if I understand your description correctly, it seems like shard prop says Partial to Partial for add scalar is OK. Which clearly it is not? So we should fix shard prop to not allow this thereby forcing a redistribute. I agree that in principle you can cleverly divide the scalar so that it does work, but (1) I don't think I care that much and (2) this should probably be expressed by just having some special shortcuts in the DTensor dispatch where when you have a specific placement you might override the implementation to do something special (similarly, you might consider redirecting to ring attention when it is possible.)

PR #163293, opened 2025-09-18
Change DebugMode record_torchfunction default to False

SherlockNoMad:
Result is too noisy with `record_torchfunction = True`. Change it to False, to make it clean. 


zpcore (approved):
LGTM!

Issue #163202, opened 2025-09-17
[CUDA][Memory Snapshot] `test_memory_compile_regions` appears to interfere with subsequent `record_memory_history` tests


eqy:
### 🐛 Describe the bug

Running `test_memory_compile_regions` before other memory tests in `test_cuda.py` e.g., with `python test/test_cuda.py -k memory` seems to cause them to fail, apparently due to corrupting/preventing their snapshots from reflecting later activity.

Consider `test_memory_plots_free_segment_stack`, which checks that `empty_cache` is captured in the memory history in the `all` context. This function as intended when the test is run standalone, or if we exit `test_memory_compile_regions` before `ss = torch.cuda.memory._snapshot()["device_traces"]`. However, runnign the full test causes the `test_memory_plots_free_segment_stack` snapshot to elide the `empty_cache` call.

### Versions

Current upstream build.

PR #163299, opened 2025-09-18
[CUDA] Cleanup persistent cuBLASLt workspaces before compile-regions test

eqy:
Fixes some tests that seemed to start flaking out as reported in #163202, due to cuBLASLt workspaces becoming persistent following that change.

It's relatively obvious why the workspaces/allocations corresponding to them should be cleaned up for `test_memory_snapshot_script` but less obvious for `test_memory_plots_free_segment_stack`?  Why does not cleaning up workspace prevent `empty_cache` from showing up?

cc @ptrblck @msaroufim @jerryzh168 @mruberry @csarofeen @xwang233

### DIFF
--- a/test/test_cuda.py
+++ b/test/test_cuda.py
@@ -4305,7 +4305,6 @@ class TestCudaMallocAsync(TestCase):
             pass
         finally:
             torch.cuda.memory._record_memory_history(None)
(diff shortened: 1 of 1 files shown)


albanD (commented):
You actually clear before the test here?

albanD (approved):
Ok ok

PR #163301, opened 2025-09-18
[rfc] Supporting exporting a model with DTensor params/inputs

suo:
As title, one issue was that our fake mode detection didn't understand dtensor.

RFC because:
- I'm a dtensor noob so I don't know if this is the right way to use dtensor
- I don't like making torch/_guards.py aware of DTensor, looking for suggestions on alternative ways to structure the code.


ezyang (approved):
I'll approve, but you can generalize by testing for and using `__tensor_flatten__` when it exists to grab inner tensors from subclasses

ezyang:
Yes, we should also do that :)

PR #163209, opened 2025-09-17
[inductor][choices] move extra kwargs out of get_template_configs

coconutruben:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #163206
* __->__ #163209
* #163305

# why

- extra kwargs are input/op dependent and not config dependent. We don't
  plan to serialize/deserialize them, and so they need to be fed in
  later beore making the KTC, rather than when getting the config values
  directly

# what

- move extra_kwargs into the KTC and get_ktc interface directly

# testing

```
python3 -bb -m pytest test/inductor/test_max_autotune.py -v -k "_addmm"
```

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov

Differential Revision: [D82871310](https://our.internmc.facebook.com/intern/diff/D82871310)

PR #163206, opened 2025-09-17
[inductor][template heuristics] decompose k as first KernelTemplateParams

coconutruben:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #163206
* #163209
* #163305

# why

- start moving schema-less dicts to structured KernelTemplateParams so
  that its clear which params each template has and which are
  required/defaulted

# what

- modify the get_template_configs to expect a dict or a
  KernelTemplateParams object, and only wrap the dict in that case but
  forward the KernelTemplateParams directly

- implement DataclassTemplateParams
- add DecomposeKTemplateParams as a version of that with k_split params

# testing

```
python3 -bb -m pytest test/inductor/test_max_autotune.py -v -k "decompose_k"
```

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @jataylo @chenyang78

Differential Revision: [D82871309](https://our.internmc.facebook.com/intern/diff/D82871309)

PR #163305, opened 2025-09-19
[inductor][triton heuristics] move allow tf32 out of config params

coconutruben:
* #163206
* #163209
* __->__ #163305

# why

- this is not directly controlled by the config arg but rather by the
  input and by the inductor wide setting
- it's always the same for every choice
- we want the config kwargs to be *programable* and this is not
  programable in that sense but rather needs to use inductor config

# what

- move generating the ALLOW_TF32 kwarg in Triton templates into
  get_extra_kwargs

# testing

with some annotations, this is now the kwargs and extra_kwargs on addmm

```
{'EVEN_K': True, 'USE_FAST_ACCUM': False, 'ACC_TYPE': 'tl.float32', 'num_stages': 1, 'num_warps': 2, 'BLOCK_M': 32, 'BLOCK_N': 32, 'BLOCK_K': 16, 'hint_override': None, 'GROUP_M': 8} # choice/config kwargs
{'ALLOW_TF32': True, 'epilogue_fn': <function addmm_epilogue.<locals>.epilogue at 0x7f64d54ff600>, 'epilogue_fn_hash': "['addmm_epilogue', torch.float32, 1, 1]", 'prefix_args': 1} # extra kwargs
```

they're both passed onto the template

Differential Revision: [D82871312](https://our.internmc.facebook.com/intern/diff/D82871312)


torch/_inductor/template_heuristics/triton.py
```diff
@@ -1446,6 +1446,23 @@ class MMTemplateConfigMixin(GemmMaxAutotuneTemplateConfigHeuristics):
     ]
     _filter_configs: Callable[[list[BaseConfig]], list[BaseConfig]]
 
+    def get_extra_kwargs(
+        self,
+        kernel_inputs: KernelInputs,
+        op_name: str,
+    ) -> dict[str, Any]:
+        assert isinstance(kernel_inputs, MMKernelInputs)
```
nmacchioni, line 1540: Should it be `kernel_inputs: MMKernelInputs` if we then assert the same?

nmacchioni (approved):
LGTM pending the one comment

torch/_inductor/template_heuristics/triton.py
coconutruben: for overriding I'm leaving it as KernelInputs right now, we have the asserts throughout the code, but I'll try to look for a cleaner way to do this, you're right that littering them around is unfortunate

coconutruben:
@coconutruben has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D82871312).

Issue #148114, opened 2025-02-27
We should never throw vanilla C++ exceptions


albanD:
We have a lot of them at the moment in a few places in the codebase: https://github.com/search?q=repo%3Apytorch%2Fpytorch%20std%3A%3Aruntime_error(&type=code

We should migrate ALL of them to TORCH_CHECK-like macros and add a lint rule to enforce it going forward.

cc @jbschlosser @Skylion007 in case you would be interested in this cleanup!

Nithurshen:
I have worked on the following files:
```
./c10/util/TypeCast.cpp
./c10/util/Registry.h
./c10/util/sparse_bitset.h
./c10/util/Exception.h
```

And have opened a [PR](https://github.com/pytorch/pytorch/pull/164910)

PR #163308, opened 2025-09-19
Better error handling in torch/nativert/*

licy666:
Replace the **runtime_error** of the vallina C++ exceptions with **TORCH_CEHCK** in **torch/nativert/***

The vallina C++ exception should not exist in the core part of pytorch for its corss-languanges trait. Comparing with the vallina C++ exceptions, TORCH_CHECK have the richer error context and It has the unified error handling mechanism. This commit replace the runtime_error with TORCH_CHECK of the files in
torch/nativert/*   .

Fixes part of #148114

### DIFF
--- a/torch/nativert/common/FileUtil.cpp
+++ b/torch/nativert/common/FileUtil.cpp
@@ -12,6 +12,7 @@
 #endif
 #include <cerrno>
 
+#include <c10/util/Exception.h>
 #include <fmt/core.h>
 
 namespace torch::nativert {
@@ -130,14 +131,15 @@ File::File(int fd, bool ownsFd) noexcept : fd_(fd), ownsFd_(ownsFd) {
 
 File::File(std::string_view name, int flags, mode_t mode)
     : fd_(::open(std::string(name).c_str(), flags, mode)), ownsFd_(false) {
-  if (fd_ == -1) {
-    throw std::runtime_error(fmt::format(
-        "open(\"{}\", {}, 0{}) failed with errno {}.",
-        name,
-        flags,
-        mode,
-        errno));
-  }
+  TORCH_CHECK(
+      fd_ != 1,
+      "open(\"",
+      name,
+      "\", ",
+      flags,
+      ", 0",
+      mode,
+      ") returned stdout.")
   ownsFd_ = true;
 }
 
@@ -166,15 +168,11 @@ File::~File() {
 /* static */ File File::temporary() {
   // make a temp file with tmpfile(), dup the fd, then return it in a File.
   FILE* tmpFile = tmpfile();
-  if (!tmpFile) {
-    throw std::runtime_error("tmpfile() failed");
-  }
+  TORCH_CHECK(tmpFile != nullptr, "tmpfile() failed");
   auto guard = c10::make_scope_exit([&]() { fclose(tmpFile); });
 
   int fd = ::dup(fileno(tmpFile));
-  if (fd == -1) {
-    throw std::runtime_error("dup() failed");
-  }
+  TORCH_CHECK(fd != -1, "dup() failed");
 
   return File(fd, true);
 }
@@ -193,9 +191,7 @@ void File::swap(File& other) noexcept {
 }
 
 void File::close() {
-  if (!closeNoThrow()) {
-    throw std::runtime_error("close() failed");
-  }
+  TORCH_CHECK(closeNoThrow(), "close() failed");
 }
 
 [[nodiscard]] bool File::closeNoThrow() {
--- a/torch/nativert/executor/triton/CpuTritonKernelManager.cpp
+++ b/torch/nativert/executor/triton/CpuTritonKernelManager.cpp
@@ -1,5 +1,6 @@
 #include <torch/nativert/executor/triton/TritonKernelManager.h>
 
+#include <c10/util/Exception.h>
 #include <c10/util/FbcodeMaps.h>
 #include <c10/util/Logging.h>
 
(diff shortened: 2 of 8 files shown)


licy666:
@FFFrog PTAL

torch/nativert/graph/passes/SubgraphRewriter.cpp
```diff
@@ -66,7 +67,12 @@ bool compareConstants(const Constant& a, const Constant& b) {
         // Unsupported types (Graph)
         LOG(ERROR) << "Unsupported Constant types for pattern matching: "
                    << typeid(lhs).name() << " vs " << typeid(rhs).name();
-        throw std::runtime_error("Unsupported Constant types.");
+        TORCH_CHECK(
+            false,
+            fmt::format(
+                "Unsupported Constant types for pattern matching: {} vs {}",
+                typeid(lhs).name(),
+                typeid(rhs).name()))
```
can-gaa-hou: Is it duplicated with the above error log?

licy666: Yes, because the behavior between LOG() and the TORCH_CHECK() is different. The Log() can provide additional error context.

torch/nativert/graph/passes/SubgraphRewriter.cpp
```diff
@@ -66,7 +67,12 @@ bool compareConstants(const Constant& a, const Constant& b) {
         // Unsupported types (Graph)
         LOG(ERROR) << "Unsupported Constant types for pattern matching: "
                    << typeid(lhs).name() << " vs " << typeid(rhs).name();
-        throw std::runtime_error("Unsupported Constant types.");
+        TORCH_CHECK(
```
Skylion007, line 69: This can be lazily concatenated with positional arguments in TORCH_CHECK

torch/nativert/graph/passes/pass_manager/GraphPassRegistry.h
```diff
@@ -63,9 +64,9 @@ class GraphPassRegistry {
 
   const GraphPass& get_pass(const GraphPassIdentifier& name) {
     auto it = registry_.find(name);
-    if (it == registry_.end()) {
-      throw std::runtime_error("Pass " + name + " not registered to get");
-    }
+    TORCH_CHECK(
+        it != registry_.end(),
+        fmt::format("Pass {} not registered to get", name));
```
Skylion007: Same here, no need for eager concatenation with fmt::format

janeyx99:
Hi! I suspect that the usage of raw exception is intended, as nativert may not want to have a dependency on libtorch. ccin'g @desertfire to confirm.

torch/nativert/common/FileUtil.cpp
```diff
@@ -130,14 +131,14 @@ File::File(int fd, bool ownsFd) noexcept : fd_(fd), ownsFd_(ownsFd) {
...
-    throw std::runtime_error(fmt::format(
-        "open(\"{}\", {}, 0{}) failed with errno {}.",
-        name,
-        flags,
-        mode,
-        errno));
-  }
+  TORCH_CHECK(
+      fd_ != -1,
+      fmt::format(
```
dolpm: no need for fmt here. just use TORCH_CHECK(cond, str_1, str_2, str_n)

torch/nativert/executor/triton/CpuTritonKernelManager.cpp
```diff
@@ -28,7 +29,8 @@ void* _dlsym(void* handle, const char* name) {
 
 char* _dlerror() {
 #if defined(_WIN32)
-  throw std::runtime_error("dlerror not supported on Windows");
+  TORCH_CHECK(false, "dlerror not supported on Windows");
+  return nullptr; // This line will never be reached, but needed for compilation
```
dolpm: __builtin_unreachable() ?

torch/nativert/kernels/AutoFunctionalizeKernel.cpp
```diff
@@ -37,10 +38,12 @@ void UnsafeAutoFunctionalizeKernel::computeInternal(
...
     // to us, we'll have to change this up a little.
     auto stackTrace = node_->getMetadata("stack_trace");
-    throw std::runtime_error(fmt::format(
-        "Original Python stacktrace:\n{}\n{}",
-        stackTrace ? *stackTrace : "<no stack trace>",
-        ex.what()));
+    TORCH_CHECK(
+        false,
+        fmt::format(
+            "Original Python stacktrace:\n{}\n{}",
```
dolpm: same as above - just use built in variadic strcat from TORCH_CHECK

torch/nativert/kernels/C10Kernel.cpp
```diff
@@ -31,15 +32,17 @@ void C10Kernel::computeInternal(ExecutionFrame& executionFrame) const {
...
-    throw std::runtime_error(fmt::format(
-        "Exception while executing node: {}\n"
-        "with args:\n{}\n"
-        "{}\n"
-        "Original Python stacktrace:\n{}",
-        fmt::streamed(*node_),
-        readableArgs(op_.schema(), stack),
-        ex.what(),
-        stackTrace ? *stackTrace : "<no stack trace>"));
+    TORCH_CHECK(
```
dolpm, line 35: same.

dolpm (changes requested):
thanks for taking the time to do this. overall looks good, but let's try to avoid using fmt and just use the builtin strcat like @Skylion007 said.

will stamp after that

dolpm (approved):

licy666:
Thanks for your kind advice!

I have replace fmt::format by positional argument.

Please have a check.

licy666:
@desertfire @SherlockNoMad fmt::format has been taken place by positional arguments in TORCH_CHECK. Please take a look. 😊

PR #162651, opened 2025-09-10
[DTensor] Add guide for what to do about mixed torch.Tensor and DTensor operations

zou3519:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #162651
* #162307
* #162117

Also updates the error message to point to the guide.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci

PR #163310, opened 2025-09-19
[graph partition] Add way to register custom rule

zou3519:
* __->__ #163310
* #162651
* #162307
* #162117

This PR adds an experimental way to register a custom rule for if
inductor should partition the graph around an operator.

Test Plan:
- new test

### DIFF
--- a/torch/_inductor/scheduler.py
+++ b/torch/_inductor/scheduler.py
@@ -22,6 +22,8 @@ if TYPE_CHECKING:
     from collections.abc import Sequence
     from types import ModuleType
 
+import weakref
+
 import sympy
 
 import torch
@@ -84,6 +86,28 @@ compute_dependencies_log = torch._logging.getArtifactLogger(
 PartitionType: TypeAlias = list["BaseSchedulerNode"]
 _T = TypeVar("_T")
 _P = ParamSpec("_P")
+
+
+_custom_should_partition_fns: weakref.WeakKeyDictionary[
+    Union[torch._ops.OpOverload, torch._ops.HigherOrderOperator], Callable[..., bool]
+] = weakref.WeakKeyDictionary()
+
+
+def register_should_partition_rule(
+    op: Union[torch._ops.OpOverload, torch._ops.HigherOrderOperator],
+    func: Callable[..., bool],
+) -> None:
+    """Register a function that says if Inductor should partition the graph on this op.
+
+    The function should be have the same signature as the operator.
+    Inductor will invoke the function with FakeTensors when it needs to decide
+    if the graph should be partitioned.
+
+    `register_should_partition_rule` is currently private and experimental.
+    Use at your own risk.
+    """
+    assert isinstance(op, (torch._ops.OpOverload, torch._ops.HigherOrderOperator))
+    _custom_should_partition_fns[op] = func
 
 
 @dataclasses.dataclass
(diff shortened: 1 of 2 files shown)


ProExpertProg (approved):
This looks good for dynamically deciding whether to partition nodes for custom ops! The only other thing here would be data-dependent dynamic sizes

BoyuanFeng (approved):

torch/_inductor/scheduler.py
```diff
@@ -86,6 +88,28 @@
...
+    """Register a function that says if Inductor should partition the graph on this op.
+
+    The function should be have the same signature as the operator.
+    Inductor will invoke the function with FakeTensors when it needs to decide
+    if the graph should be partitioned.
+
+    `register_should_partition_rule` is currently private and experimental.
+    Use at your own risk.
+    """
+    assert isinstance(op, (torch._ops.OpOverload, torch._ops.HigherOrderOperator))
```
eellison: When would this be torch._ops.HigherOrderOperator ?

torch/_inductor/scheduler.py
```diff
@@ -4347,6 +4371,21 @@ def should_partition(
...
+        # Allow users to manually specify if a node should be partitioned
+        # Can only do this for FallbackKernels
+        ir_node = node.node
+        if isinstance(ir_node, torch._inductor.ir.FallbackKernel):
+            operator = ir_node.op_overload
+            if operator is not None and operator in _custom_should_partition_fns:
+                should_partition_fn = _custom_should_partition_fns[operator]
+                fx_node = ir_node.get_origin_node()
+                assert fx_node is not None
+                _, fake_args, fake_kwargs = (
```
eellison: nit, this wont always succeed.. assert on first returned arg ? tbh not sure why i didnt just make this optional return..

eellison: Nit: if the operator doesnt have fixed require strides, it's possible the strides of the inputs that will be passed in at runtime are different than the strides in `fake_args` `fake_kwargs`. Not sure it really matters though, non-blocking for land.

torch/_inductor/scheduler.py
```diff
@@ -86,6 +88,28 @@
 _P = ParamSpec("_P")
 
 
+_custom_should_partition_fns: weakref.WeakKeyDictionary[
```
eellison, line 91: should there be anyway of resetting this ? it's fine if not since this is very particular use case and vllm can do it manually. also note this wont be part of fx graph cache but i guess that's fine.

eellison (approved):

torch/_inductor/scheduler.py
zou3519: no clue lol. It is very likely that the HOP part of this doesn't work

torch/_inductor/scheduler.py
zou3519: Oh I didn't realize the first returned arg was a bool for if this succeeded or not

torch/_inductor/scheduler.py
zou3519: I am going to assert against HOP to start. There's not many HOPs that would be here anyways (they're not FallbackKernel)

torch/_inductor/scheduler.py
zou3519: easiest way is torch._inductor.scheduler._custom_should_partition_fns.reset() lol

I'm not sure

torch/_inductor/scheduler.py
zou3519: yike

eellison: @IvanKobzarev has an example showing how to get the args with correct strides here: https://github.com/pytorch/pytorch/pull/162377/files#diff-8dfc175f817a04eb10c183509564129a91f32db31190e7b4ad3fd9653337c6f6R3672. but again not landing blocking.

PR #163315, opened 2025-09-19
[Inductor][Intel GPU] Save `threads_per_warp` from tirton compiled kernel for launching kernel correctly in cpp wrapper.

etaf:
* __->__ #163315

On the Inductor XPU backend, `threads_per_warp` is not always 32. For Intel GEMM Triton kernels, it can be 16. This information must be preserved for XPU so that the Cpp wrapper can launch the kernel with the correct configuration.
 

### DIFF
--- a/torch/_inductor/runtime/triton_heuristics.py
+++ b/torch/_inductor/runtime/triton_heuristics.py
@@ -1114,10 +1114,9 @@ class CachingAutotuner(KernelInterface):
             # For Intel GEMM Triton kernels, it can be 16.
             # This information must be preserved so that the Cpp wrapper
             # can launch the kernel with the correct configuration.
-            threads_per_warp = 32
-            if hasattr(launcher.bin.metadata, "threads_per_warp"):
-                threads_per_warp = launcher.bin.metadata.threads_per_warp
-            params["threads_per_warp"] = threads_per_warp
+            params["threads_per_warp"] = getattr(
+                launcher.bin.metadata, "threads_per_warp", 32
+            )
 
         from torch._inductor.codecache import CudaKernelParamCache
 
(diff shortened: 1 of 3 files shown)


EikanWang (approved):

torch/csrc/inductor/aoti_runtime/sycl_runtime_wrappers.h
```diff
@@ -128,12 +128,10 @@ static std::unique_ptr<sycl::kernel> _createKernel(
     uint32_t numWarps,
     uint32_t sharedMemory,
     void** params,
-    sycl::queue* queuePtr) {
+    sycl::queue* queuePtr,
+    uint32_t threadsPerWarp) {
```
chengjunlu, line 132: This order looks better to me.
```
uint32_t gridX,
uint32_t gridY,
uint32_t gridZ,
uint32_t numWarps,
uint32_t threadsPerWarp,
uint32_t sharedMemory,
void** params,
```

etaf: Yes, the order is better, but the code generation of these parameters are shared with cuda, so I would prefer keep the `theadPerWarp` as an extra paramter here.

torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -1109,6 +1109,16 @@ def save_gpu_kernel(self, stream, launcher):
...
             "profile_scratch": launcher.profile_scratch,
         }
+        if self.device_props.type == "xpu":
+            # On the XPU backend, threads_per_warp is not always 32.
+            # For Intel GEMM Triton kernels, it can be 16.
+            # This information must be preserved so that the Cpp wrapper
+            # can launch the kernel with the correct configuration.
+            threads_per_warp = 32
+            if hasattr(launcher.bin.metadata, "threads_per_warp"):
+                threads_per_warp = launcher.bin.metadata.threads_per_warp
```
desertfire: Nit: You can drop if-checking and use `getattr` with `32` as the default value.

desertfire (approved):

torch/_inductor/runtime/triton_heuristics.py
etaf: @desertfire : Thank you very much for your suggestion. I’ve simplified this piece of code into a single line.

PR #163320, opened 2025-09-19
[DTensor] Handle Broadcast properly for aten.expand

SherlockNoMad:
Fixes #160968

Think about following case
```
        mesh = init_device_mesh(DEVICE_TYPE, (4,))
        tensor = torch.randn(1, 4)
        dt = distribute_tensor(tensor, mesh, [Shard(0)])
        out = dt.expand(4, 4)
```
Tensor of shape[1, 4] is unevenly sharded on 0-th dim into 4 shard. Only the first rank holds the local_tensor [1, 4], other ranks holds local_tensor of shape[0, 4]. 

When expanding to shape of [4, 4], (or more generally [N, 4]), there could be 2 potential resulting placements
1. [S(0)].  Each rank would result in local_tensor of [1, 4]. This would require one-to-all broadcast.
2. [Replicated]. Each rank would result in local_tensor of [4, 4]. But in practice, it would first replicate the input of shape [1, 4], and then expand locally. 

In both cases, this would require a broadcasting collective, and the redistribution cost is the same. 
However, we are using "AllGather followed by Slice" to simulate "one-to-all broadcasting" collective. 

So I am choosing Option 2, to replicate the input tensor first, and then expand locally. 

Generally, for view operation on DTensor, we would need some kind of "lazy communication": record the op, and only materialize communication when need. 

### DIFF
--- a/test/distributed/tensor/test_dtensor_ops.py
+++ b/test/distributed/tensor/test_dtensor_ops.py
@@ -691,6 +691,24 @@ class TestDTensorOps(DTensorOpTestBase):
             else:
                 self.assertTrue("redistribute_input" in debug_mode.debug_string())
 
+    def test_expand(self):
+        self.mesh = init_device_mesh(DEVICE_TYPE, (self.world_size,))
+
+        tensor = torch.arange(4, dtype=torch.float, device=DEVICE_TYPE).reshape(1, 4)
+        dt0 = distribute_tensor(tensor, self.mesh, [Shard(0)])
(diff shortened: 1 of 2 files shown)


ezyang:
While the PR looks plausible, I'm not terribly familiar with this part of the view algorithm, so I have a basic sanity check question: how can you argue that no case where we previously *soundly* avoided a redistribute with broadcast, wasn't impeded from happening now?

Issue #162246, opened 2025-09-05
SAC's _CachingTorchDispatchMode does not support HOPs


xmfan:
### 🐛 Describe the bug

Running into this error https://github.com/pytorch/pytorch/blob/5da573c42c332bc68d4b7946c69f690a876d951a/torch/_ops.py#L423-L431

When attempting to use a HOP (local_map) inside of a SAC'd function. My current repro requires setting up autoparallel, but this error looks like it would affect all HOPs that try to control dispatch order.

```python
(Pdb++) bt
  [ 0] /home/xmfan/core/a/autoparallel/examples/example_local_map.py(167)<module>()
       with AutoParallel(model, input_fn, mesh, mp_policy, compile=True) as autop:
  [ 1] /home/xmfan/core/a/autoparallel/autoparallel/api.py(223)__enter__()
       self.build_model_graph()
  [ 2] /home/xmfan/core/a/autoparallel/autoparallel/api.py(287)build_model_graph()
       ep = torch.export.export(self.model, inputs, strict=True)
  [ 3] /home/xmfan/core/a/pytorch/torch/export/__init__.py(277)export()
       return _export(
  [ 4] /home/xmfan/core/a/pytorch/torch/export/_trace.py(1126)wrapper()
       ep = fn(*args, **kwargs)
  [ 5] /home/xmfan/core/a/pytorch/torch/export/exported_program.py(124)wrapper()
       return fn(*args, **kwargs)
  [ 6] /home/xmfan/core/a/pytorch/torch/export/_trace.py(2195)_export()
       ep = _export_for_training(
  [ 7] /home/xmfan/core/a/pytorch/torch/export/_trace.py(1126)wrapper()
       ep = fn(*args, **kwargs)
  [ 8] /home/xmfan/core/a/pytorch/torch/export/exported_program.py(124)wrapper()
       return fn(*args, **kwargs)
  [ 9] /home/xmfan/core/a/pytorch/torch/export/_trace.py(2058)_export_for_training()
       export_artifact = export_func(
  [10] /home/xmfan/core/a/pytorch/torch/export/_trace.py(1412)_strict_export()
       gm_torch_level = _export_to_torch_ir(
  [11] /home/xmfan/core/a/pytorch/torch/export/_trace.py(806)_export_to_torch_ir()
       gm_torch_level, _ = torch._dynamo.export(
  [12] /home/xmfan/core/a/pytorch/torch/_dynamo/eval_frame.py(1979)inner()
[...]

PR #163322, opened 2025-09-19
[hop] support local_map + SAC

xmfan:
* #163602
* __->__ #163322

Some ops like local_map hop's deferred mode are not desugared by make_fx, this means that when we apply SAC tags, we will need to define dispatch rules for the SAC torch dispatch modes as pointed out here: https://github.com/pytorch/pytorch/issues/162246#issuecomment-3259176721. This PR adds those rules.

Additionally it fixes a pre-existing issue where we weren't coercing tangent layout (that AOTAutograd typically does) when partitioning the HOP joint.

### DIFF
--- a/test/higher_order_ops/test_local_map.py
+++ b/test/higher_order_ops/test_local_map.py
@@ -23,7 +23,6 @@ if torch.distributed.is_available():
     from torch.distributed.tensor.placement_types import Replicate, Shard
 
 from torch.testing._internal.common_utils import run_tests, TEST_WITH_CROSSREF, TestCase
-from torch.testing._internal.triton_utils import requires_cuda_and_triton
 
 
 nested_compile_region = torch.compiler.nested_compile_region
@@ -149,7 +148,6 @@ class TestLocalMap(TestCase):
     def tearDown(self):
         self.exit_stack.close()
 
-    @requires_cuda_and_triton
     @unittest.skipIf(
         not torch.distributed.is_available(), "Torch distributed not available."
     )
@@ -165,14 +163,14 @@ class TestLocalMap(TestCase):
 
         backend = EagerAndRecordGraphs()
 
-        model = create_model(cp_decorated, nheads, dim1, dim2).cuda()
-        inputs = (torch.randn(bs, seq_len, dim1, requires_grad=True).cuda(),)
+        model = create_model(cp_decorated, nheads, dim1, dim2)
+        inputs = (torch.randn(bs, seq_len, dim1, requires_grad=True),)
         with LocalMapWrappedHigherOrderVariable.enable():
             out = torch.compile(model, backend=backend)(*inputs)
         out.sum().backward()
 
-        model = create_model(cp_function, nheads, dim1, dim2).cuda()
-        inputs = (torch.randn(bs, seq_len, dim1, requires_grad=True).cuda(),)
+        model = create_model(cp_function, nheads, dim1, dim2)
+        inputs = (torch.randn(bs, seq_len, dim1, requires_grad=True),)
         with LocalMapWrappedHigherOrderVariable.enable():
             out = torch.compile(model, backend=backend)(*inputs)
         out.sum().backward()
@@ -238,7 +236,6 @@ class GraphModule(torch.nn.Module):
 """,
             )
 
-    @requires_cuda_and_triton
     @unittest.skipIf(
         not torch.distributed.is_available(), "Torch distributed not available."
     )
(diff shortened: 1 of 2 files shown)


xmfan:
hold on, updating test to use aot eager backend

test/higher_order_ops/test_local_map.py
```diff
@@ -42,7 +44,17 @@ def context_parallel_attention(query, key, value):
...
 
 
-def create_model(attention_fn, nheads, dim1, dim2):
+def save_attention(ctx, op, *args, **kwargs):
+    if (
+        op == torch.ops.aten._scaled_dot_product_flash_attention.default
+        or op == torch.ops.aten._scaled_dot_product_efficient_attention.default
+    ):
+        # NOTE: we can't save nondeterministic_seeded ops, the run with rng wrapper is not traceable yet
+        return torch.utils.checkpoint.CheckpointPolicy.PREFER_SAVE
```
soulitzer: Prefer save is completely ignored by the compiler btw, so generally its recommended to use `MUST_SAVE`, but probably fine I if we're testing eager only

test/higher_order_ops/test_local_map.py
```diff
@@ -42,7 +44,17 @@ def context_parallel_attention(query, key, value):
     return out
 
 
-def create_model(attention_fn, nheads, dim1, dim2):
+def save_attention(ctx, op, *args, **kwargs):
+    if (
+        op == torch.ops.aten._scaled_dot_product_flash_attention.default
+        or op == torch.ops.aten._scaled_dot_product_efficient_attention.default
+    ):
+        # NOTE: we can't save nondeterministic_seeded ops, the run with rng wrapper is not traceable yet
```
soulitzer: Wait do you mean "cannot recompute RNG ops"

xmfan: oh ignore this, this is autoparallel frontend specific

test/higher_order_ops/test_local_map.py
```diff
@@ -193,10 +238,183 @@ def forward(self, q_1: "f32[8, 16, 16, 6]", k_1: "f32[8, 16, 16, 6]", v_1: "f32[
 """,
             )
 
+    @requires_cuda_and_triton
```
ezyang: Why does this test require CUDA?

test/higher_order_ops/test_local_map.py
```diff
@@ -193,10 +238,183 @@ def forward(self, q_1: "f32[8, 16, 16, 6]", k_1: "f32[8, 16, 16, 6]", v_1: "f32[
...
+    @unittest.skipIf(
+        not torch.distributed.is_available(), "Torch distributed not available."
+    )
```
ezyang: Ditto this, doesn't seem like you need distributed

test/higher_order_ops/test_local_map.py
```diff
@@ -193,10 +238,183 @@ def forward(self, q_1: "f32[8, 16, 16, 6]", k_1: "f32[8, 16, 16, 6]", v_1: "f32[
...
+        ).cuda()
+        inputs = (torch.randn(bs, seq_len, dim1, requires_grad=True).cuda(),)
+        try:
+            with (
+                LocalMapWrappedHigherOrderVariable.enable(),
+                torch._higher_order_ops.local_map.defer_inlining(),
+            ):
+                out = torch.compile(model, backend=backend)(*inputs)
+            out.sum().backward()
+        except AttributeError as e:
```
ezyang, line 372: What's going on here?

test/higher_order_ops/test_local_map.py
```diff
@@ -193,10 +238,183 @@ def forward(self, q_1: "f32[8, 16, 16, 6]", k_1: "f32[8, 16, 16, 6]", v_1: "f32[
...
+                    actual = node.meta["recompute"]
+                    self.assertEqual(expected, actual)
+                    if actual == torch.utils.checkpoint.CheckpointPolicy.MUST_SAVE:
+                        self.assertTrue(node.name in fw_outs)
+                    #     self.assertTrue(node.name in fw_outs and node.name in bw_ins)
+                    # elif (
+                    #     actual == torch.utils.checkpoint.CheckpointPolicy.MUST_RECOMPUTE
+                    # ):
+                    #     # can still be in fw_outs for post-graph bytecode
+                    #     self.assertFalse(node.name in bw_ins)
```
ezyang, line 423: Is there a reason we can't use an expect test here?

ezyang (approved):
The actual impl changes are plausible

test/higher_order_ops/test_local_map.py
xmfan: nice it works on cpu now

test/higher_order_ops/test_local_map.py
xmfan: i need it to import local_map/device mesh

test/higher_order_ops/test_local_map.py
xmfan: local_map HOP currently only works for AP style compile which interprets the nodes and directly accessses their target. But the graph's codegen is currently wrong, it should be `torch._higher_order_ops.local_map.<locals>.call_local_map`. I'll fix this later as it's not needed for AP.

test/higher_order_ops/test_local_map.py
xmfan: we could, but we'd need to manually check for these properties whenever we update the expecttest anyway

test/higher_order_ops/test_local_map.py
ezyang: oh ok, please look forward to https://www.internalfb.com/diff/D82283623

test/higher_order_ops/test_local_map.py
ezyang: I guess I'd prefer to make it obvious which code is live and which code is dead, with a comment with what you said here

PR #163335, opened 2025-09-19
[dynamo, nested graph breaks] enable basic nested graph break wrapped tests

williamwen42:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* #171826
* #163503
* #179844
* #179843
* #165597
* __->__ #163335

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @kadeng @chauhang @amjames @jataylo @azahed98 @Lucaskabela @chenyang78


torch/_dynamo/test_case.py
```diff
@@ -102,16 +102,25 @@ def tearDown(self) -> None:
...
+        if config.debug_disable_compile_counter:
+            if isinstance(x, utils.CompileCounterInt) or isinstance(
+                y, utils.CompileCounterInt
+            ):
+                return
+            # skip checks like self.assertEqual(len(counters["graph_break"]), 1)
+            if (
+                (cur_frame := inspect.currentframe())
+                and (upper_frame := cur_frame.f_back)
+                and (upper_code := inspect.getframeinfo(upper_frame).code_context)
```
fxdawnn, line 124: Quick question: Does this include the case when there are 3 layer of nested function?
``f(g(h(x)))``

torch/_dynamo/config.py
```diff
@@ -485,7 +485,8 @@
 # traced FX graph is empty when RETURN_* is traced.
 allow_empty_graphs = False
 
-# Used for testing - forces all top-level functions to be nested when traced with Dynamo
+# Used for testing - forces all top-level functions to be nested when traced with Dynamo.
+# There are slight differences between this config and wrap_top_frame.
```
Lucaskabela, line 624: Can we elaborate what those differences would be here?

Lucaskabela (approved):
Generally looks gtm; seeking a bit of elaboration on a config comment though

torch/_dynamo/test_case.py
williamwen42: The current frame would be `self.assertEqual` and the second frame would be the test file - what other case are you thinking about?

This is also admittedly hacky - I'll probably remove this once `nested_graph_breaks` is enabled on dynamo unittests.

mlazos (approved):

PR #163344, opened 2025-09-19
Simplify _compute_local_shape_and_global_offset and make it SPMD.

ezyang:
* __->__ #163344

There is only one substantive change: the branch on
`global_offset[shard_dim] <= local_offset[shard_dim]`
is removed because it is unnecessary: you can always treat the
first shard uniformly with the rest of the shards, because your
global offset is guaranteed to be zero in this case anyway.

I also switch the shard_size case to sym_ite, to make it possible
for LocalTensor to deal with the MPMD-ness here, but it's equivalent
to the old if-then-else.

I tried to rewrite the comments to be more clear what is going on
algorithmically here.

### DIFF
--- a/torch/distributed/tensor/_utils.py
+++ b/torch/distributed/tensor/_utils.py
@@ -169,7 +169,6 @@ def _compute_local_shape_and_global_offset(
 
             local_shape[shard_dim] = shard_size
 
-
             global_offset[shard_dim] = torch.sym_ite(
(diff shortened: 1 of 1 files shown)


albanD (approved):
SGTM

zpcore (approved):
LGTM!

tianyu-l (approved):

PR #163358, opened 2025-09-19
[DeviceMesh] Implement a device mesh concatenate api for submesh and SPMD use case

fduwjj:
* #165492
* __->__ #163358
* #166264

Today FSDP needs to slicing out spmd mesh from root mesh here: https://github.com/pytorch/pytorch/blob/main/torch/distributed/fsdp/_fully_shard/_fsdp_param.py#L301. But essentially, users want is a concatenate of some submesh into a big mesh and used as a spmd mesh. This PR is tentatively trying to implement this API for users.

One thing to note is that, all sub-mesh needs to slicing/flatten or unflatten from same root mesh otherwise the indices make no sense when it comes to mesh indexing and device allocation.

Differential Revision: [D85409698](https://our.internmc.facebook.com/intern/diff/D85409698)


torch/distributed/device_mesh.py
```diff
@@ -1159,6 +1159,52 @@ def _unflatten(
...
+            concatenate_dim_group_name: list[str] = []
+            root_mesh = _mesh_resources.get_root_mesh(device_mesh_list[0])
+            for mesh in device_mesh_list:
+                concatenate_dim_names.extend(not_none(mesh.mesh_dim_names))
+                concatenate_sizes.extend(list(flatten(mesh._layout.sizes)))
+                concatenate_strides.extend(list(flatten(mesh._layout.strides)))
+                concatenate_dim_group_name.extend(not_none(mesh._dim_group_names))
+                # Concatenate device mesh having different root mesh tensors are meaningless
+                # because the concatenated indices should be indexed by the same root mesh tensor.
+                if not torch.equal(
```
ezyang: You don't want to do object identity for the remapping? This is O(N) right here!

ezyang: Also, you don't need to test the mesh itself, you just need to know the remapping is consistent and you can concatenate without reference to the root mash.

torch/distributed/device_mesh.py
```diff
@@ -1159,6 +1159,52 @@ def _unflatten(
...
+                ):
+                    raise RuntimeError(
+                        "Cannot concatenate DeviceMeshes with different root mesh tensors"
+                    )
+            concatenate_mesh_layout = _MeshLayout(
+                tuple(concatenate_sizes), tuple(concatenate_strides)
+            )
+            if not concatenate_mesh_layout.check_non_overlap():
+                raise RuntimeError(
+                    f"Cannot concatenate overlapping meshes: {device_mesh_list}"
```
ezyang, line 1206: I do wonder if someone is going to ask for a concatenate that implicitly drops duplicated dims. This would be coherent I think, though I don't know if you need it.

torch/distributed/device_mesh.py
```diff
@@ -1159,6 +1159,52 @@ def _unflatten(
...
+                )
+            cur_rank = root_mesh.get_rank()
+            pg_ranks_by_dim = concatenate_mesh_layout.to_remapping_tensor(
+                root_mesh.mesh,
+                get_world_size(),
+            )
+
+            for mesh_nd in pg_ranks_by_dim:
```
ezyang: ?!?! Why do you need to do it for every mesh_nd? Is this because you're triggering comms to initialize PGs?

torch/distributed/device_mesh.py
fduwjj: We do want the root mesh tensor is the same to ensure that remapping makes sense for each one in the mesh list.

torch/distributed/device_mesh.py
fduwjj: I honestly think that is not a valid use case for concatenate..

torch/distributed/device_mesh.py
fduwjj: so long story short, we need all ranks to call `new_group` which is hidden very deep in the stack to initialize PGs. Otherwise the code will hang.

torch/distributed/device_mesh.py
ezyang: Yes, but I argue we should do object identity for that

torch/distributed/device_mesh.py
```diff
@@ -1176,6 +1176,41 @@ def _unflatten(
...
+                    concat_sizes.append(dm._layout[i].sizes)
+                    concat_strides.append(dm._layout[i].strides)
+                concat_dim_names.extend(not_none(dm.mesh_dim_names))
+                concat_dim_group_name.extend(not_none(dm._dim_group_names))
+                # Concatenate device mesh having different root mesh tensors are meaningless
+                # because the concatenated indices should be indexed by the same root mesh tensor.
+                if dm._flatten_rank_map != flatten_rank_map:
+                    raise RuntimeError(
+                        "Cannot concatenate DeviceMeshes with different root mesh tensors"
+                    )
```
fegin, line 1202: We don't define root mesh tensors to users, nor do we have a way to get the root mesh tensor (we want to kill it). So should we have a more meaningful description here, like the inputs DeviceMesh must be derived from the same DeviceMesh?

fegin (approved):

torch/distributed/device_mesh.py
fduwjj: sure, let me update it.

PR #163367, opened 2025-09-19
[CuTe] Add layout overlap checking util function in _MeshLayout

fduwjj:
* #163213
* __->__ #163367
* #163930
* #163928
* #163288
* #163212

While refactoring the bookkeeping for DeviceMesh while leveraging CuTe layout, we found that we need to have two more util functions. One is to check whether one layout has overlap inside it or not. For example, (2,2):(2:1) has no overlap while (2,2):(2:2) has overlap.


torch/distributed/_mesh_layout.py
```diff
@@ -210,3 +211,144 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
...
+        - Dim 1: stride=1, span=3*1=3, covers addresses [0,1,2]
+        - Dim 0: stride=2, span=2*2=4, covers addresses [0,2]
+        → Overlap! stride=2 < span=3, so addresses [0,2] are duplicated
+
+        Returns:
+            bool: True if no overlap exists (valid layout), False if overlap detected
+        """
+        previous_span = -1
+        previous_stride = -1
+        for size, stride in sorted(self.sizes_and_strides, key=lambda x: x[1]):
```
ezyang: It's kind of goofy and never shows up for device meshes, but technically you need to handle zero here too.

torch/distributed/_mesh_layout.py
```diff
@@ -210,3 +211,144 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
...
+            if size == 1:
+                continue
+            if previous_stride == stride or stride < previous_span:
+                return False
```
ezyang: You don't need to test previous_stride == stride. Suppose previous_stride == stride. Then previous_span = size * previous_stride, where size > 1. This then guarantees that previous_span is greater than stride.

ezyang: (this is why I commented about zeros above, technically size can be zero in which case this reasoning doesn't hold, but with zero numel there is never any overlap anyway)

torch/distributed/_mesh_layout.py
```diff
@@ -210,3 +211,144 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
...
+        3. If consecutive: return the logical groups directly
+        4. If non-consecutive or partial world: map logical indices to actual ranks
+
+        Examples:
+
+        Case 1 - Consecutive ranks, full world:
+        original_mesh_tensor = [[0,1],[2,3]]  # 2x2 mesh, ranks 0-3
+        world_size = 4
+        layout = Layout(2:2)
+        → Returns logical groups directly: [[0,2],[1,3]]
```
ezyang: Yeah, the implementation here seems overly complicated. My intuition is that you always generate assuming arange(world_size) device mapping first. Then, to apply a remapping, simply take every device and replace it with mapping[device]. Now you are done. Because the mappings are injective, you can always invert it into a partial function as well.

ezyang: Is this function claude code generated? If it is I feel less bad telling you to rewrite it all  :P

torch/distributed/_mesh_layout.py
Skylion007: The lambda for the key here should just be generated with 'key=operator.itemgetter(1)' as it's a cfunction in Python and will let the interpreter optimize this call a bit more.

torch/distributed/_mesh_layout.py
```diff
@@ -210,3 +211,144 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
             [group_offset + group_rank for group_rank in self.member_ranks()]
             for group_offset in self.complement(world_size).member_ranks()
         ]
+
+    def check_non_overlap(self) -> bool:
+        """
+        Check if the layout has any overlap between the ranks it generates. If there is overlap,
+        we return False, otherwise True.
+
+        Aside from indice 0, indices from each dim of the layout must be non-overlapping.
```
lw: For the math-curious readers, could we mention that this is equivalent to whether the layout is injective?

torch/distributed/_mesh_layout.py
```diff
@@ -210,3 +211,144 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
...
+    ) -> torch.Tensor:
+        """
+        Convert this layout into a tensor representation that maps the logical mesh
+        structure to actual device ranks, handling cases where the mesh doesn't use
+        consecutive ranks or doesn't span the full world size (Neither is CuTe representible).
+
+        With this method, the cute layout serves as the backend of indices bookkeeping for the
+        mesh tensor when it comes to flatten, unflatten and slicing operations. The actual mesh
+        tensor still represents the actual device assignment and ranks. We need this function
+        to specify device allocation and create backend for a mesh.
```
lw: I'm somewhat confused by the purpose of this function and its exact "specification". Notably, I didn't expect the documented behavior for case #3 when reading this docstring.

When I chatted with @ezyang I had understood that the issue we're trying to solve here is when the "codomain" of the Layouts isn't the range `[0, world_size)`, but is either shifted or shuffled. Concretely, it could mean that the coordinate `(0, ..., 0)` does _not_ map to 0 but to some higher value (hence the Layout is "affine" rather than linear) or that the coordinate `n` might map to a _higher_ value than the coordinate `n-1` (i.e., the mapping isn't linear/monotonic at all).

The mental model I had constructed was that we represent everything as a composition of a regular linear CuTe Layout _and_ a single global arbitrary non-linear layout. (I mean this as _functional_ composition). When we manipulate Layouts we only manipulate the linear part, and leave the rest unchanged, which means that all Layout will share the same common non-linear component.

Another way of seeing this is that the linear CuTe layouts are the "tensors" (with sizes and strides), which are different ways of indexing into the elements of a shared "storage" (which is a 1-d sequences of values) which would be our arbitrary non-linear remapping of ranks.

Is this the right mental model to have? If so, could you explain what this function is supposed to perform?

lw (commented):
I'm a bit perplexed by the second part of this PR

torch/distributed/_mesh_layout.py
fegin: > Another way of seeing this is that the linear CuTe layouts are the "tensors" (with sizes and strides), which are different ways of indexing into the elements of a shared "storage" (which is a 1-d sequences of values) which would be our arbitrary non-linear remapping of ranks.

This what I understand and `original_mesh_tensor` is the storage and I don't think it is a 1-d sequences for now but it should be in the future after refactoring?

@fduwjj  please correct me if I'm wrong.

lw: Alright, in that case I have some follow-up questions:
- Why do we allow `original_mesh_tensor` to have more than 1 dimensions? Storages are 1-d, and it seems that any "structure" of `original_mesh_tensor` (ndim, shape, strides, ...) could be folded into the Layout, thus allowing `original_mesh_tensor` to be 1-d as well.
- If we're doing functional composition, then in example no. 3 why are we clamping the strides? The rules of composition (IMO) should dictate that if a large stride gets us out-of-bounds for the underlying storage we either error out or we ignore that stride (i.e., we clamp the _size_, not the stride).

torch/distributed/_mesh_layout.py
fduwjj: sure.

torch/distributed/_mesh_layout.py
fduwjj: right, let me make remove `previous_stride == stride`

torch/distributed/_mesh_layout.py
fduwjj: > When we manipulate Layouts we only manipulate the linear part, and leave the rest unchanged, which means that all Layout will share the same common non-linear component.

So it is not the case when it comes to slicing, we need to have a new "storage". So it is not a view rather than a split or reshape.

fduwjj: @lw 

> Why do we allow original_mesh_tensor to have more than 1 dimensions? Storages are 1-d, and it seems that any "structure" of original_mesh_tensor (ndim, shape, strides, ...) could be folded into the Layout, thus allowing original_mesh_tensor to be 1-d as well.

A mesh tensor can be any shape, so it can be N-D tensor. And indeed today after init of a device mesh, we store a flattened mesh tensor for device mesh comparison. But in order to create backend (PG) correctly we cannot directly return a 1D tensor here. Also this is because today all ranks have to call `new_group` even if one rank is not in that group. c10d legacy reason.

> If we're doing functional composition, then in example no. 3 why are we clamping the strides? The rules of composition (IMO) should dictate that if a large stride gets us out-of-bounds for the underlying storage we either error out or we ignore that stride (i.e., we clamp the size, not the stride).

oh the reason for clamping is that when you slicing from a submesh. You might think that we should just infer layout from passed in tensor and we will never run into out of bounds issue. Let me give you one example here, when slicing from a mesh (2,2,2) ("dp", "cp", "tp") ["dp", "cp"]. The layout is (2,2):(4,2) not (2,2):(2,1). then when you further slicing ["dp"]. If we use the inferred layout (2,2):(2,1), the index storage tensor generated from will be wrong. 

To make things more complicated, if users directly create the dp_cp mesh from PGs and then slicing ["dp"]. We also need to use (2,2):(4,2) this layout.

torch/distributed/_mesh_layout.py
fduwjj: Sure, let me rewrite, I did use claude to generate the comments for me. It failed the Turing test, lol. The code logic is not claude generated.

fduwjj: > Yeah, the implementation here seems overly complicated. My intuition is that you always generate assuming arange(world_size) device mapping first.

For legacy reason, we will need to return a N-D mesh tensor first and then get the tensor which contains current rank for backend creation.. If we fix the legacy part later, this function will look simpler.

fduwjj:
@ezyang 

> While flattening, slicing, we need to remap the layout back to the new mesh tensor so it maps the actual device allocation.

today when flattening or slicing, the mesh tensor actually is changed. I mean you can always say they are views or subset of original tensors (yes that is correct) but we do need to materialize that view or subset for backend creation, especially for flatten.

torch/distributed/_mesh_layout.py
fduwjj: And you are right the `torch.equal(...)` that part can be removed. so the code looks simpler now.

torch/distributed/_mesh_layout.py
```diff
@@ -211,3 +213,125 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
...
+        The layout is supposed to be injective i.e, aside from indice 0, indices from each
+        dim of the layout must be non-overlapping.
+
+        Here is how it works:
+        1. Sort dimensions by stride (smallest stride first)
+        2. For each dimension, check if:
+           - It has the same stride as previous dimension (duplicate mapping)
+           - Its stride overlaps with the previous dimension's span
+
+        A dimension's "span" is size * stride, representing the address space it covers.
```
fegin: `Address` seems to be confusing, should it be index?

torch/distributed/_mesh_layout.py
```diff
@@ -211,3 +213,125 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
...
+
+        Example 1 - Valid (no overlap):
+        Layout: sizes=(2,3), strides=(6,1)
+        - Dim 1: stride=1, span=3*1=3, covers addresses [0,1,2]
+        - Dim 0: stride=6, span=2*6=12, covers addresses [0,6]
+        → No overlap since 6 > 3
```
fegin, line 227: Do you want to have an example that the span equals to the previous so that we can understand the boundary case? Or simply change this to be the equal case, whichever you prefer.

torch/distributed/_mesh_layout.py
```diff
@@ -211,3 +213,125 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
...
+            if size == 0 or size == 1:
+                continue
+            if stride < previous_span:
+                return False
+            previous_span = size * stride
+        return True
+
+    def remap_to_tensor(
+        self,
+        original_mesh_tensor: torch.Tensor,
```
fegin: Can you add a clarification why this `original_mesh_tensor` is not 1D? Like it is the mesh tensor that is provided by users when users call `init_device_mesh(mesh_tensor)`?

fegin: And do we want to keep `original_` prefix? It sounds like we have a new mesh tensor.

torch/distributed/_mesh_layout.py
```diff
@@ -211,3 +213,125 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
...
+            layout = Layout(2:2)
+            Return: [[0,2],[1,3]]
+
+        Case 2 - Non-consecutive ranks:
+            original_mesh_tensor = [[10,20],[30,40]]  # custom rank assignment
+            world_size = 4
+            layout = Layout(2:2)
+            Return: [[[10,30],[20,40]]]
+
+        Case 3 - Partial world (stride scaling needed):
```
fegin: I'm confused by the usage of `world_size`. Why do we need world_size? I thought the layout will be "scaled" after any operations, e.g., flatten, unflatten, same as mesh tensor?

torch/distributed/_mesh_layout.py
```diff
@@ -211,3 +213,125 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
             [offset + rank for rank in self.all_ranks_from_zero()]
             for offset in self.complement(world_size).all_ranks_from_zero()
         ]
+
+    def check_non_overlap(self) -> bool:
+        """
+        Check if the layout has any overlap between the ranks it generates. If there is overlap,
+        we return False, otherwise True.
```
ezyang, line 218: This is not quite accurate. The problem is that size stride pairs can encode arbitrary Diophantine equations, which are generally difficult to solve. For example, consider (3, 2):(2, 3). This looks like it overlaps. But it actually does not: 

```
	•	(0,0) → 0
	•	(0,1) → 3
	•	(1,0) → 2
	•	(1,1) → 5
	•	(2,0) → 4
	•	(2,1) → 7
```

ezyang: I don't know of a technical term for this particular condition, but ChatGPT suggested using the adjective "monotonic" somewhere, which seems good here.

torch/distributed/_mesh_layout.py
```diff
@@ -211,3 +213,125 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
...
+
+        Example 2 - Invalid (overlap):
+        Layout: sizes=(2,3), strides=(2,1)
+        - Dim 1: stride=1, span=3*1=3, covers addresses [0,1,2]
+        - Dim 0: stride=2, span=2*2=4, covers addresses [0,2]
+        → Overlap! stride=2 < span=3, so addresses [0,2] are duplicated
+
+        Returns:
+            bool: True if no overlap exists (valid layout), False if overlap detected
+        """
```
ezyang, line 243: The LLM generated comment is too long. Make it shorter.

torch/distributed/_mesh_layout.py
```diff
@@ -211,3 +213,125 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
...
+        Leverage layout as an index for mesh tensor that re-maps the indexes after layout
+        transformation to actual device ranks.
+
+        With this method, the cute layout serves as the backend of indices bookkeeping for the
+        mesh tensor when it comes to flatten, unflatten and slicing operations. The actual mesh
+        tensor still represents the actual device assignment and ranks. We need this function
+        to specify device allocation and create backend for a mesh.
+
+        When the layout span is larger than the size of actual tensor, we need to clamp the stride
+        to the size of tensor to avoid out of boundary error. Although any transform of mesh tensors
```
ezyang: ?????? sounds like LLM slop

ezyang: There is something going on with stride scaling but you're definitely not clamping.

torch/distributed/_mesh_layout.py
```diff
@@ -211,3 +213,125 @@ def global_ranks(self, world_size: int) -> list[list[int]]:
...
+        """
+
+        def scale_stride(scale: int, strides: IntTuple) -> IntTuple:
+            """
+            Recursively scale down strides by a factor to fit within smaller mesh.
+
+            When layout expects world_size=8 but mesh only has 4 elements,
+            we need to scale strides down by factor of 2 to generate valid indices.
+
+            Example: stride=4 with scale=2 → stride=2 (or keep as-is if stride < scale)
```
ezyang: This is a horrible example because you'd have the same result if you did `strides % scale`. Which, actually, I would really like to know, why is this not modulus?

fduwjj: ok maybe I can use a different example here.

fduwjj:
hmm still more comments to address.. Will do it tomorrow.

torch/distributed/_mesh_layout.py
fduwjj: index, yes. Let me fix it.

torch/distributed/_mesh_layout.py
fduwjj: Sure added one more example.

torch/distributed/_mesh_layout.py
fduwjj: sure I will add it in the refactor PR where this code is moved into

fduwjj: no this is the mesh_tensor from previous device mesh..

torch/distributed/_mesh_layout.py
fduwjj: no we need to keep the original stride, because of two reasons. One we support from_group which the group are from first two dims for a 3D device mesh. Also for flatten a sliced submesh, you also need to keep the original stride, otherwise you get wrong ranks for PG creation.

The test I am talking about is `test_from_group_with_mesh_shape_3d` and also `mesh_3d["dp","cp"]._flatten()`

torch/distributed/_mesh_layout.py
fduwjj: this is actually an interesting case... and the current logic fails on this case.

torch/distributed/_mesh_layout.py
fduwjj: ok I misunderstood the meaning of `clamping ` when luca mentioned it. We are doing scaling indeed.

torch/distributed/_mesh_layout.py
fduwjj: I think we can just leverage `all_ranks_from_zero` to generate all indices? let me think about it.

fduwjj: ok I added this case into Unit test as well.

fduwjj:
Ready for review again.

fegin (approved):

PR #163368, opened 2025-09-19
refactor: pull _replace_node common functionality out of Scheduler.finalize_multi_template_buffers

aorenste:
Pull replace_node function out of Scheduler.finalize_multi_template_buffers(). This is needed by the next PR (#163369). As part of this also pull the _replace_operation_buffer() up to top-level since it needed no self references.

* #163369
* __->__ #163368


PaulZhang12 (approved):

PaulZhang12 (changes requested):
Actually I think this needs a bit more context.. we are moving fields from object level to class level?

PaulZhang12 (approved):
Right, I understand, but I guess my initial assumption was that we usually just annotate class level variables (static) and assign instance level at init time. This refactor seems fundamentally better though, stamped.

PR #163377, opened 2025-09-19
Fix prims.broadcast_in_dim functionalization

jansel:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.12.0) (oldest at bottom):
* __->__ #163377

Fixes #163037

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @chenyang78

PR #163386, opened 2025-09-20
Better decomp for torch.eye

jansel:
* #163377
* #163482
* #163520
* #163481
* #163422
* #163412
* #163393
* #163434
* #163419
* #163415
* #163414
* #163387
* #163398
* __->__ #163386


torch/_refs/__init__.py
```diff
@@ -5599,12 +5599,16 @@ def eye(
     torch._check(n >= 0, lambda: f"n must be greater or equal to 0, got {n}")
     torch._check(m >= 0, lambda: f"m must be greater or equal to 0, got {m}")
 
-    range_n = torch.arange(n, dtype=torch.int64, device=device, requires_grad=False)
-    range_m = torch.arange(m, dtype=torch.int64, device=device, requires_grad=False)
+    if max(n, m) <= torch.iinfo(torch.int32).max:
+        range_dtype = torch.int32  # optimization for small tensors
```
vadimkantorov, line 5603: 2 billon items is quite a lot :) so for small tensors, maybe also supporting uint16 (65536) would be better?

jansel: True, though I'm mainly trying to avoid int64 math on CUDA (which is much slower).  While the difference between int16 and int32 is smaller.

vadimkantorov:
btw does uint8/uint16 also work here? or no impls for unsigned?

jansel:
Not sure if all backends implement those types.

torch/_refs/__init__.py
```diff
@@ -5599,12 +5599,19 @@ def eye(
     torch._check(n >= 0, lambda: f"n must be greater or equal to 0, got {n}")
     torch._check(m >= 0, lambda: f"m must be greater or equal to 0, got {m}")
 
-    range_n = torch.arange(n, dtype=torch.int64, device=device, requires_grad=False)
-    range_m = torch.arange(m, dtype=torch.int64, device=device, requires_grad=False)
+    range_dtype = torch.int64
```
Skylion007, line 5616: Slight nit: feels like this strength reduction helper optimization would make a good reusable utility method if you make the 'max(m, n)' the input and have it return the range dtype, might be useful to future decomps.

Skylion007:
> Not sure if all backends implement those types.

More reason to factor this out in their own utility. It would be good to have a robust check here we could grab the min dtype that is supported on target device.

Skylion007 (commented):
This is subtly complicated to do, and we should factor it out into its own function.

torch/_refs/__init__.py
```diff
@@ -5585,6 +5585,13 @@ def empty_strided(
     )
 
 
+def _strength_reduce_integer(val: int) -> torch.dtype:
+    for possible_dtype in (torch.uint8, torch.uint16, torch.int32):
+        if val <= torch.iinfo(possible_dtype).max:
```
eellison, line 5590: nit: fix for unbacked. If adding guards here is intentional: `guard_or_False`. Otherwise: `statically_known_true(val <= torch.iinfo(possible_dtype).max)`

eellison (approved):

torch/_refs/__init__.py
jansel: This is only called for `int`, for SymInt we use int64.

PR #163390, opened 2025-09-20
[dynamo] Fix handling of kwargs in exception constructor

rtimpe:
* __->__ #163390


rtimpe:
> Are there any test cases that are failing because of this? If I recall correctly, there are some exception subclasses (OSError, SyntaxError) that accept kwargs.
>

No current tests are failing, but the one I added in this PR would fail.  Currently the kwargs are passed to the base VariableTracker constructor.
 
> We could maybe parse the init ctor of the exception to get the named arguments and bind them correctly. Not sure how viable this is.

Yeah, I was thinking something like this, but not sure how to do it.  There's `exc_type.__dict__` but I'm not sure how to identify which of those are kwargs to `__init__`.  Otherwise we could have explicit support for each exception type, but that's a bit more work.

guilhermeleobas (approved):

PR #163394, opened 2025-09-20
Combine strong and weak refcounts in intrusive_ptr in a single refcount

mcfi:
Summary:
Currently, we assume that refcount_ and weakcount_ are always stored in an 8-byte aligned address right next to each other. Based on this assumption, we load 8 bytes in intrusive_ptr::reset_ to check the values of both counts. However, that assumption is not part of C++ language standard so it's essentially undefined behavior.

This change eliminates that assumption by combining refcount_ and weakcount_ in a single 64-bit count and we use the lower 32 bits for refcount_ and upper 32 bits for the weakcount_.

In addition to eliminating the undefined behavior, the change also eliminates the read of weakcount_ after decrementing refcount_ in intrusive_ptr::reset_. This claws back lost performance introduced in https://github.com/pytorch/pytorch/pull/162784 for non-final refcount_ decrementing.

Reviewed By: yfeldblum

Differential Revision: D82869192

### DIFF
--- a/c10/util/intrusive_ptr.h
+++ b/c10/util/intrusive_ptr.h
@@ -27,7 +27,78 @@ struct DontIncreaseRefcount {};
 } // namespace raw
 
 namespace detail {
-constexpr uint32_t kImpracticallyHugeReferenceCount = 0x0FFFFFFF;
+constexpr uint64_t kImpracticallyHugeReferenceCount = 0x0FFFFFFF;
+constexpr uint64_t kImpracticallyHugeWeakReferenceCount =
+    (kImpracticallyHugeReferenceCount << 32);
+constexpr uint64_t kReferenceCountOne = 1;
+constexpr uint64_t kWeakReferenceCountOne = (kReferenceCountOne << 32);
+constexpr uint64_t kUniqueRef = (kReferenceCountOne | kWeakReferenceCountOne);
+
+template <class TTarget>
+struct intrusive_target_default_null_type final {
+  static constexpr TTarget* singleton() noexcept {
+    return nullptr;
+  }
+};
+
+template <class TTarget, class ToNullType, class FromNullType>
+TTarget* assign_ptr_(TTarget* rhs) {
+  if (FromNullType::singleton() == rhs) {
+    return ToNullType::singleton();
+  } else {
+    return rhs;
+  }
+}
+
+inline uint32_t refcount(uint64_t combined_refcount) {
+  return static_cast<uint32_t>(combined_refcount);
+}
+
+inline uint32_t weakcount(uint64_t combined_refcount) {
+  return static_cast<uint32_t>(combined_refcount >> 32);
+}
+
+// The only requirement for refcount increment is that it happens-before
+// decrement, so no additional memory ordering is needed.
+inline uint64_t atomic_combined_refcount_increment(
+    std::atomic<uint64_t>& combined_refcount,
+    uint64_t inc) {
+  return combined_refcount.fetch_add(inc, std::memory_order_relaxed) + inc;
+}
+
+inline uint32_t atomic_refcount_increment(
+    std::atomic<uint64_t>& combined_refcount) {
+  return detail::refcount(atomic_combined_refcount_increment(
+      combined_refcount, kReferenceCountOne));
+}
+
+inline uint32_t atomic_weakcount_increment(
+    std::atomic<uint64_t>& combined_refcount) {
+  return detail::weakcount(atomic_combined_refcount_increment(
+      combined_refcount, kWeakReferenceCountOne));
+}
+
+// The requirement is that all modifications to the managed object happen-before
+// invocation of the managed object destructor, and that allocation of the
+// managed object storage happens-before deallocation of the storage.
+//
+// To get this ordering, all non-final decrements must synchronize-with the
+// final decrement. So all non-final decrements have to store-release while the
+// final decrement has to load-acquire, either directly or with the help of
+// fences. But it's easiest just to have all decrements be acq-rel. And it turns
+// out, on modern architectures and chips, it's also fastest.
+inline uint64_t atomic_combined_refcount_decrement(
+    std::atomic<uint64_t>& combined_refcount,
+    uint64_t dec) {
+  return combined_refcount.fetch_sub(dec, std::memory_order_acq_rel) - dec;
+}
+
+inline uint32_t atomic_weakcount_decrement(
+    std::atomic<uint64_t>& combined_refcount) {
+  return detail::weakcount(atomic_combined_refcount_decrement(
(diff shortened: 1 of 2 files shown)


c10/util/intrusive_ptr.h
```diff
@@ -80,8 +85,15 @@ class C10_API intrusive_ptr_target {
...
   //    If it is not, you must report that the storage is dead.
   //
-  mutable std::atomic<uint32_t> refcount_;
-  mutable std::atomic<uint32_t> weakcount_;
+  //.We use a single combined count for refcount and weakcount so that
+  // we can atomically operate on both at the same time for performance
+  // and defined behaviors.
+  //
+  mutable std::atomic<uint64_t> combined_refcount_;
+  static_assert(
```
Skylion007: Why not split into multiple static asserts so it's obvious which invariant is violated.

c10/util/intrusive_ptr.h
```diff
@@ -177,6 +189,15 @@ class C10_API intrusive_ptr_target {
    * destructed), this function WILL NOT be called.
    */
   virtual void release_resources() {}
+
+  uint32_t refcount(std::memory_order order = std::memory_order_relaxed) const {
+    return (uint32_t)combined_refcount_.load(order);
```
Skylion007: Please use C++ style casts here, is this a static_cast / reinterpret cast etc.

c10/util/intrusive_ptr.h
```diff
@@ -198,12 +219,23 @@ TTarget* assign_ptr_(TTarget* rhs) {
...
+    std::atomic<uint64_t>& combined_refcount) {
+  return (uint32_t)(atomic_combined_refcount_increment(
+      combined_refcount, kReferenceCountOne));
+}
+
+inline uint32_t atomic_weakcount_increment(
+    std::atomic<uint64_t>& combined_refcount) {
+  return (uint32_t)(atomic_combined_refcount_increment(
+                        combined_refcount, kWeakReferenceCountOne) >>
+                    32);
```
Skylion007: The 32 offset is used a lot as a magic number. Better to make it a constexpr variable later in case we want to change the offset at a later time, improve code as documentation about why?

c10/util/intrusive_ptr.h
```diff
@@ -215,12 +247,17 @@ inline uint32_t atomic_weakcount_increment(std::atomic<uint32_t>& weakcount) {
...
+    std::atomic<uint64_t>& combined_refcount,
+    uint64_t dec) {
+  return combined_refcount.fetch_sub(dec, std::memory_order_acq_rel) - dec;
 }
 
-inline uint32_t atomic_weakcount_decrement(std::atomic<uint32_t>& weakcount) {
-  return weakcount.fetch_sub(1, std::memory_order_acq_rel) - 1;
+inline uint32_t atomic_weakcount_decrement(
+    std::atomic<uint64_t>& combined_refcount) {
+  return (uint32_t)(atomic_combined_refcount_decrement(
```
Skylion007: Should this be static cast or reinterpret cast?

c10/util/intrusive_ptr.h
```diff
@@ -590,11 +611,11 @@ class intrusive_ptr final {
 #ifdef NDEBUG
     expected_decrefs = 0;
 #endif
-    result.target_->refcount_.store(
-        detail::kImpracticallyHugeReferenceCount + expected_decrefs,
+    result.target_->combined_refcount_.store(
+        (uint32_t)(detail::kImpracticallyHugeReferenceCount +
```
Skylion007: Same here

Skylion007 (commented):
Please replace the c style casts with more explicit C++ style casts here especially since we are doing very niche things with bit level representation optimization.

c10/util/intrusive_ptr.h
swolchok: > The 32 offset is used a lot as a magic number.

personally I think it's fine in this case because we are using uint32_t and uint64_t,so the sources for the numbers are right there

Skylion007: Yeah, wouldn't normally flag it here, just thinking ahead in case we need to support exotic architectures at some point because it's such an important part of the code

mcfi: Introduced utility functions refcount/weakcount in namespace::detail to abstract away how refcount and weakcount are extracted from the combined_refcount

c10/util/intrusive_ptr.h
mcfi: Converting uint64_t to uint32_t should always use static_cast.

mcfi:
> Please replace the c style casts with more explicit C++ style casts here especially since we are doing very niche things with bit level representation optimization.

Sure, I've replaced all C-style casts to C++, thanks.

c10/util/intrusive_ptr.h
```diff
@@ -27,7 +27,78 @@ struct DontIncreaseRefcount {};
 } // namespace raw
 
 namespace detail {
```
Skylion007, line 29: Really minor readability suggestion, creating a struct which only stores the uint64_t and methods would be a bit cleaner, you could add some static asserts to make it sure it's POD, the right size, properly aligned etc, and get some type safety. Might be overkill for this though, your call.
```cpp
struct RefCounts {
    std::atomic<uint64_t> v{0};

    static constexpr uint64_t RefOne  = 0x00000001ULL;
    static constexpr uint64_t WkOne   = 0x00000001ULL << 32;
    static constexpr uint64_t RefMask = 0x00000000FFFFFFFFULL;
    static constexpr uint64_t WkMask  = 0xFFFFFFFF00000000ULL;
    static constexpr uint64_t Unique  = RefOne | WkOne;

    uint32_t ref(std::memory_order mo = std::memory_order_relaxed) const {
        return static_cast<uint32_t>(v.load(mo));
    }
    uint32_t weak(std::memory_order mo = std::memory_order_relaxed) const {
        return static_cast<uint32_t>(v.load(mo) >> 32);
    }

    uint32_t inc_ref()  { return static_cast<uint32_t>(v.fetch_add(RefOne,  std::memory_order_relaxed) + RefOne); }
    uint32_t inc_weak() { return static_cast<uint32_t>((v.fetch_add(WkOne, std::memory_order_relaxed) + WkOne) >> 32); }

    uint64_t dec_ref_acqrel() { return v.fetch_sub(RefOne, std::memory_order_acq_rel) - RefOne; }
    uint32_t dec_weak_acqrel(){ return static_cast<uint32_t>((v.fetch_sub(WkOne, std::memory_order_acq_rel) - WkOne) >> 32); }
};
```
Would make it easier to type check and organize the methods

Skylion007 (approved):
Looks like a strict improvement to me whether or not you go with the struct. It is great that does not exploit UB and is usable on more platforms and improves the performance over the last prev shared_pointer optimization.

PR #163407, opened 2025-09-20
Slightly improve speed of AdamW by rearranging order of bias_correction2_sqrt

ad8e:
Source of idea: https://x.com/Tim_Dettmers/status/1969131103798567295

> Looking closer, PyTorch also uses FP32, but here's the real reason why bnb Adam is better: we optimized for float numerics, order does matter! Computing sqrt(v) + eps*c2 then dividing avoids amplifying errors vs PyTorch's sqrt(v)/c2 + eps. Same math, better stability!

I don't believe this is true. The stability benefit should be meaningless; there is some discussion at https://discord.com/channels/729741769192767510/1079865324087803985/1418777002592174190 (accessing this requires joining the [EleutherAI discord](https://discord.gg/zBGx3azzUn))

However, Dettmer's version is faster than PyTorch's, because it does 2 tensor-wise multiplications instead of 1 elementwise division on sqrt(v). So I copied it to the torch implementation. My own optimizer does not have an eps/beta2 interaction, so I have no prior testing of this change. If momentum buffers are changed to fp16/bf16, instead of float32 as they are now, this new formulation will have higher precision than the old formulation.

Changed:
_single_tensor_adam, non-capturable
_multi_tensor_adam, non-capturable
_multi_tensor_adam, capturable

Not changed:
_single_tensor_adam, capturable. The existing code is extremely inefficient for reasons I don't understand. Thus, I left it alone.
fused adam

Not tested:
Whether "capturable multi tensor adam" still satisfies being "capturable". I don't know the contract it must satisfy, so I can't test it.

Tested:
It's faster for normal usecases, but testing is not comprehensive. If there are only 2 tensors, the change to capturable `_multi_tensor_adam` is slower.
Loss curves match exactly with the changed versions. In this testcase, `eps` is set to a deliberately high number to make differences obvious:
```
import torch
import torch.nn as nn
from torch.optim.adamw import AdamW

torch.manual_seed(42)
torch.cuda.manual_seed_all(42)

class SimpleModel(nn.Module):
    def __init__(self):
        super(SimpleModel, self).__init__()
        self.linear = nn.Linear(10, 10)
        self.linear2 = nn.Linear(10, 1)

    def forward(self, x):
        return self.linear2(self.linear(x))

model = SimpleModel()

learning_rate = 0.1
weight_decay = 0.01

# to change: capturable, foreach
optimizer = AdamW(model.parameters(), lr=learning_rate, weight_decay=weight_decay, eps=1.0, capturable=True)

criterion = nn.MSELoss()

input_data = torch.randn(32, 10)
target_data = torch.randn(32, 1)

# optional: cuda
model = model.cuda()
input_data = input_data.cuda()
target_data = target_data.cuda()

num_epochs = 10
for epoch in range(num_epochs):
    outputs = model(input_data)
    loss = criterion(outputs, target_data)

    optimizer.zero_grad()
    loss.backward()
    optimizer.step()

    print(f'Epoch [{epoch+1}/{num_epochs}], Loss: {loss.item():.4f}')
```

Unknown:
Why the old code used `(lr / bc) * -1`. I changed it instead to `-(lr / bc)` (and then `-(lr * bc2 / bc1)`), but maybe there is some hidden behavior?

cc @jerryzh168

### DIFF
--- a/torch/optim/adam.py
+++ b/torch/optim/adam.py
@@ -524,17 +524,20 @@ def _single_tensor_adam(
             step_size = lr / bias_correction1
 
             bias_correction2_sqrt = bias_correction2**0.5
+            # We use this identity:
+            # 1 / (sqrt(v) / beta2_correction_sqrt + eps) =
+            # beta2_correction_sqrt / (sqrt(v) + eps * beta2_correction_sqrt)
 
             if amsgrad:
                 # Maintains the maximum of all 2nd moment running avg. till now
                 torch.maximum(max_exp_avg_sqs[i], exp_avg_sq, out=max_exp_avg_sqs[i])
 
                 # Use the max. for normalizing running avg. of gradient
-                denom = (max_exp_avg_sqs[i].sqrt() / bias_correction2_sqrt).add_(eps)
+                denom = max_exp_avg_sqs[i].sqrt().add_(eps * bias_correction2_sqrt)
             else:
-                denom = (exp_avg_sq.sqrt() / bias_correction2_sqrt).add_(eps)
+                denom = exp_avg_sq.sqrt().add_(eps * bias_correction2_sqrt)
 
-            param.addcdiv_(exp_avg, denom, value=-step_size)  # type: ignore[arg-type]
+            param.addcdiv_(exp_avg, denom, value=-step_size * bias_correction2_sqrt)  # type: ignore[arg-type]
 
         # Lastly, switch back to complex view
         if amsgrad and torch.is_complex(params[i]):
@@ -738,6 +741,13 @@ def _multi_tensor_adam(
             step_size = bias_correction1
             bias_correction2_sqrt = bias_correction2
 
+            # now step_size = - lr * sqrt(1 - beta2 ^ t) / (1 - beta1 ^ t)
+            torch._foreach_mul_(step_size, bias_correction2_sqrt)
+
+            # bias_correction2_sqrt unused after this; reuse its buffer for eps
+            torch._foreach_mul_(bias_correction2_sqrt, eps)
+            eps_bias_corrected = bias_correction2_sqrt
+
             if amsgrad:
                 device_max_exp_avg_sqs = cast(list[Tensor], device_max_exp_avg_sqs_)
                 # Maintains the maximum of all 2nd moment running avg. till now
@@ -748,8 +758,7 @@ def _multi_tensor_adam(
             else:
                 exp_avg_sq_sqrt = torch._foreach_sqrt(device_exp_avg_sqs)
 
-            torch._foreach_div_(exp_avg_sq_sqrt, bias_correction2_sqrt)
-            torch._foreach_add_(exp_avg_sq_sqrt, eps)
+            torch._foreach_add_(exp_avg_sq_sqrt, eps_bias_corrected)
             torch._foreach_div_(exp_avg_sq_sqrt, step_size)
 
             # at this point, exp_avg_sq_sqrt = - (1 - beta^t) * [sqrt(exp_avg_sq / (1 - beta2^t)) + eps] / lr
@@ -762,9 +771,9 @@ def _multi_tensor_adam(
                 1 - beta2 ** _get_value(step) for step in device_state_steps
             ]
 
-            step_size = _stack_if_compiling([(lr / bc) * -1 for bc in bias_correction1])
-
             bias_correction2_sqrt = [bc**0.5 for bc in bias_correction2]  # type: ignore[arg-type]
+
+            step_size = _stack_if_compiling([-(lr * bc2 / bc1) for bc1, bc2 in zip(bias_correction1, bias_correction2_sqrt)])
 
             if amsgrad:
                 device_max_exp_avg_sqs = cast(list[Tensor], device_max_exp_avg_sqs_)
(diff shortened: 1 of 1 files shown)


torch/optim/adam.py
```diff
@@ -524,17 +524,20 @@ def _single_tensor_adam(
...
+            # 1 / (sqrt(v) / beta2_correction_sqrt + eps) =
+            # beta2_correction_sqrt / (sqrt(v) + eps * beta2_correction_sqrt)
 
             if amsgrad:
                 # Maintains the maximum of all 2nd moment running avg. till now
                 torch.maximum(max_exp_avg_sqs[i], exp_avg_sq, out=max_exp_avg_sqs[i])
 
                 # Use the max. for normalizing running avg. of gradient
-                denom = (max_exp_avg_sqs[i].sqrt() / bias_correction2_sqrt).add_(eps)
+                denom = max_exp_avg_sqs[i].sqrt().add_(eps * bias_correction2_sqrt)
```
Skylion007, line 536: Would the Fused Adam CPP files need to be updated? What about other highly related optimizers? See: https://github.com/search?q=repo%3Apytorch%2Fpytorch%20bias_correction2_sqrt&type=code

ad8e: Yes, the same change applies to the three files found by this search (FusedAdamKernel.cpp, fused_adam_utils.cuh, FusedOptimizerOps.metal). I didn't change them because my setup makes it difficult to test these files and because I don't know CUDA, but the same transformation should be applied there.

Skylion007: My understanding is that this fix should be universal across all these backends (CPU/GPU/Metal). Regardless, we should align the implementations if you observe a speedup

ad8e: _multi_tensor_adam
capturable old: 93.7 ms
capturable new: 109 ms (slower!)
non-capturable old: 88 ms
non-capturable new: 77 ms

_single_tensor_adam
non-capturable old: 90 ms
non-capturable new: 78 ms

So it's a speedup except for capturable `_multi_tensor_adam`. Unclear why; need advice. I'm suspecting `_foreach` overhead. Benchmark:

```
import torch
import torch.nn as nn
from torch.optim.adamw import AdamW

torch.manual_seed(42)
torch.cuda.manual_seed_all(42)

dim = 4096

class SimpleModel(nn.Module):
    def __init__(self):
        super(SimpleModel, self).__init__()
        self.linear = nn.Linear(dim, dim)
        self.linear2 = nn.Linear(dim, 1)

    def forward(self, x):
        return self.linear2(self.linear(x))


model = SimpleModel()

learning_rate = 0.1
weight_decay = 0.01

# to change: capturable, foreach
optimizer = AdamW(model.parameters(), lr=learning_rate, weight_decay=weight_decay, eps=1.0, foreach=True, capturable=True)

criterion = nn.MSELoss()

input_data = torch.randn(1, dim)
target_data = torch.randn(1, 1)

# optional: cuda
model = model.cuda()
input_data = input_data.cuda()
target_data = target_data.cuda()

num_epochs = 100
outputs = model(input_data)
loss = criterion(outputs, target_data)
loss.backward()

start = torch.cuda.Event(enable_timing=True)
end = torch.cuda.Event(enable_timing=True)
start.record()

for epoch in range(num_epochs):
    optimizer.step()

end.record()
torch.cuda.synchronize()
print(start.elapsed_time(end))
```

Skylion007: @janeyx99 Will have more insight into this

janeyx99 (commented):
Thanks for kicking this off. @albanD, @ngimel and I have been discussing replacing all eps logic with taking the `max(val, eps)` instead of `val + eps` entirely.

ad8e:
`max(val, eps)` will have marginally tighter tolerances on eps, maybe 10%? Loss curves with semi-tuned `eps` should be mostly equivalent.

Point of caution though: there are many possible variants on the beta2 buffer and eps, which bypass these interactions with bias_correction2. So if the purpose of `max` is just for a small speedup, it's very likely that it'll be a maintenance burden for a non-standard variant.

For example, here's one of the many choices: the second moment buffer can be initialized with `eps`, in which case the eps addition is removed entirely. I have not tested this for models with poor gradient flow or with fp16, for which this might be disastrous.

ngimel:
Can you provide a tl;dr here on why stability benefit is meaningless?

torch/optim/adam.py
janeyx99: It looks like this PR introduces one extra foreach op call in the capturable case, so Im guessing it's due to that.

ad8e: If I bump up the tensor count in the benchmark from 2 to 4 (and make them all 4096x4096), then old is 630 ms and new is 595 ms.

```
import torch
import torch.nn as nn
from torch.optim.adamw import AdamW
import sys

torch.manual_seed(42)
torch.cuda.manual_seed_all(42)

dim = 4096

layer_count = 10


class SimpleModel(nn.Module):
    def __init__(self, num_layers):
        super(SimpleModel, self).__init__()
        layers = []
        for _ in range(num_layers):
            layers.append(nn.Linear(dim, dim))
        self.layers = nn.ModuleList(layers)

    def forward(self, x):
        for layer in self.layers:
            x = layer(x)
        return x


model = SimpleModel(layer_count)

learning_rate = 0.1
weight_decay = 0.01

# to change: capturable, foreach
optimizer = AdamW(model.parameters(), lr=learning_rate, weight_decay=weight_decay, eps=1.0, foreach=True, capturable=True)

criterion = nn.MSELoss()

input_data = torch.randn(1, dim)
target_data = torch.randn(1, dim)

# optional: cuda
model = model.cuda()
input_data = input_data.cuda()
target_data = target_data.cuda()

num_epochs = 100
outputs = model(input_data)
loss = criterion(outputs, target_data)
loss.backward()

start = torch.cuda.Event(enable_timing=True)
end = torch.cuda.Event(enable_timing=True)
start.record()

for epoch in range(num_epochs):
    optimizer.step()

end.record()
torch.cuda.synchronize()
print(f"Time with {layer_count} layers: {start.elapsed_time(end)} ms")
```

So the constant-time `_foreach` overhead is causing it.

I'm blindly assuming the _foreach usage is necessary in the `capturable` case. It's being used to multiply 1-element tensors together.

Skylion007: Yeah makes sense, the foreach kernel runtime gets amortized as there are more tensors in play and as they get larger.

PR #163418, opened 2025-09-20
[BE]: Add typing utils to copy signatures from methods or signatures

Skylion007:
Testing out this little shim to improve typing through out the codebase.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci @lolpack @maggiemoss @ndmitchell @kinto0 @samwgoldman

### DIFF
--- a/torch/utils/_typing_utils.py
+++ b/torch/utils/_typing_utils.py
@@ -51,23 +51,26 @@ def copy_method_params(
     return return_func
 
 
-# stricter variants to preserve the origin callers Return Type too
-_RSrc = TypeVar("_RSrc")  # source callable's return type for stricter variants
-
-
+# stricter variants to preserve the origin callers Return Type too.
+# TODO: consider folding both these into the above variants with an optional
+# parameter to control whether to copy the return type or not.
 def copy_func_sig(
-    source_func: Callable[_P, _RSrc],
-) -> Callable[[Callable[..., _RSrc]], Callable[_P, _RSrc]]:
-    def _return(func: Callable[..., _RSrc]) -> Callable[_P, _RSrc]:
-        return cast(Callable[_P, _RSrc], func)
+    source_func: Callable[_P, _R],
+) -> Callable[[Callable[..., _R]], Callable[_P, _R]]:
+    """Cast the decorated function's call signature and return type to the source_func's."""
+
+    def _return(func: Callable[..., _R]) -> Callable[_P, _R]:
+        return cast(Callable[_P, _R], func)
 
     return _return
 
 
 def copy_method_sig(
-    source_method: Callable[Concatenate[_A1, _P], _RSrc],
-) -> Callable[[Callable[..., _RSrc]], Callable[Concatenate[_A1, _P], _RSrc]]:
-    def _return(func: Callable[..., _RSrc]) -> Callable[Concatenate[_A1, _P], _RSrc]:
-        return cast(Callable[Concatenate[_A1, _P], _RSrc], func)
+    source_method: Callable[Concatenate[_A1, _P], _R],
+) -> Callable[[Callable[..., _R]], Callable[Concatenate[_A1, _P], _R]]:
+    """Cast the decorated *method*'s call signature to the source_method and return type."""
+
+    def _return(func: Callable[..., _R]) -> Callable[Concatenate[_A1, _P], _R]:
+        return cast(Callable[Concatenate[_A1, _P], _R], func)
 
     return _return
--- a/torch/utils/_typing_utils.py
+++ b/torch/utils/_typing_utils.py
@@ -1,6 +1,8 @@
 """Miscellaneous utilities to aid with typing."""
 
-from typing import TypeVar
+from collections.abc import Callable
+from typing import Any, cast, Concatenate, TypeVar
+from typing_extensions import ParamSpec
 
 
 # Helper to turn Optional[T] into T when we know None either isn't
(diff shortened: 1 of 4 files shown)


ezyang (approved):

rec (commented):
So this is very cool!

Do typecheckers actually understand this and get the right type for the decorated functions?

Quibble: calling a type variable `_A1` is unintuitive, unless you're talking about steak. 😉 Why not `_T` like we do everywhere else?

torch/utils/_typing_utils.py
```diff
@@ -12,3 +13,63 @@ def not_none(obj: Optional[T]) -> T:
...
         raise TypeError("Invariant encountered: value was None when it should not be")
     return obj
+
+
+_P = ParamSpec("_P")
+_R = TypeVar("_R")
+_A1 = TypeVar("_A1")
+
+
+def copy_func_params(
```
benjaminglass1, line 24: My only feedback is that I agree with your comment below; please merge the `params`/`sig` methods together. Maybe a bool called `validate_return` would be sufficiently clear in intent?

benjaminglass1 (approved):

torch/utils/_typing_utils.py
Skylion007: It's not validate return though, it's copy/override return though? It will not validate return it will copy the return type from the signature

torch/utils/_typing_utils.py
```diff
@@ -12,3 +13,63 @@ def not_none(obj: Optional[T]) -> T:
     if obj is None:
         raise TypeError("Invariant encountered: value was None when it should not be")
     return obj
+
```
rec, line 17: This is kind of a holy grail for Python type checking, the ability to just copy signatures from place to place.

Well done!

Maybe you should start off with a paragraph of documentation explaining all of them?

Reading the names elsewhere out of context, they might not be entirely clear. What about:

* `copy_function_sig`
* `copy_function_sig_params_only`
* `copy_method_sig`
* `copy_method_sig_params_only`

?

Those are pretty unambiguous.

torch/utils/_typing_utils.py
```diff
@@ -12,3 +13,63 @@ def not_none(obj: Optional[T]) -> T:
...
+_A1 = TypeVar("_A1")
+
+
+def copy_func_params(
+    source_func: Callable[_P, Any],
+) -> Callable[[Callable[..., _R]], Callable[_P, _R]]:
+    """Cast the decorated function's call signature to the source_func's.
+
+    Usage:
+        def upstream_func(a: int, b: float, *, double: bool = False) -> float: ...
```
rec, line 30: Perhaps call this `source` or `source_func`, no need to introduce this new "upstream" idea.

Also consider throwing a blank line after this definition....

torch/utils/_typing_utils.py
```diff
@@ -12,3 +13,63 @@ def not_none(obj: Optional[T]) -> T:
...
+        @copy_func_params(upstream_func)
+        def enhanced(a: int, b: float, *args: Any, double: bool = False, **kwargs: Any) -> str: ...
+    """
```
rec, line 33: So wait, what is the effect of this? 

It seems as if "enhanced"'s signature would be restricted by this, and if you tried to pass any `kwargs` to it, the type checker would complain.

torch/utils/_typing_utils.py
```diff
@@ -12,3 +13,63 @@ def not_none(obj: Optional[T]) -> T:
...
+
+    def return_func(func: Callable[..., _R]) -> Callable[_P, _R]:
+        return cast(Callable[_P, _R], func)
+
+    return return_func
+
+
+def copy_method_params(
```
rec, line 41: It's not clear to me however that it is possible to merge the "params" and "sig" versions, I'd be excited to see it.

The ultimate ultimate would be if you had a single decorator:

`def copy_signature(params_only: bool = False, is_method: bool = False)`

rec (commented):
Great stuff!

torch/utils/_typing_utils.py
Skylion007: Oh thi swould basically check if double was passed as a kwarg only and complain if a kwarg was passed that was not in upstream function. So if you wanted to restrict the kwargs it's useful. Honestly this doc string was generated by codex so probably not the best example here.

torch/utils/_typing_utils.py
Skylion007: Awesome, yeah I'll spend some time merging this in now that we care more about Python type checking

torch/utils/_typing_utils.py
Skylion007: You can do it, but it'd be truly unreadable;
```suggestion
from collections.abc import Callable
from typing import Any, Concatenate, Literal, TypeVar, overload, cast
from typing_extensions import ParamSpec

_P = ParamSpec("_P")
_R = TypeVar("_R")
_A1 = TypeVar("_A1")

# ----------------------------
# copy_signature decorator
# ----------------------------

# ---- functions: params only ----
@overload
def copy_signature(
    source: Callable[_P, Any],
    *,
    params_only: Literal[True] = True,
    is_method: Literal[False] = False,
) -> Callable[[Callable[..., _R]], Callable[_P, _R]]: ...

# ---- functions: full signature (return tied to source) ----
@overload
def copy_signature(
    source: Callable[_P, _R],
    *,
    params_only: Literal[False],
    is_method: Literal[False] = False,
) -> Callable[[Callable[..., _R]], Callable[_P, _R]]: ...

# ---- methods: params only (keep first arg type of decorated method) ----
@overload
def copy_signature(
    source: Callable[Concatenate[Any, _P], Any],
    *,
    params_only: Literal[True] = True,
    is_method: Literal[True],
) -> Callable[[Callable[..., _R]], Callable[Concatenate[_A1, _P], _R]]: ...

# ---- methods: full signature (return tied to source) ----
@overload
def copy_signature(
    source: Callable[Concatenate[Any, _P], _R],
    *,
    params_only: Literal[False],
    is_method: Literal[True],
) -> Callable[[Callable[..., _R]], Callable[Concatenate[_A1, _P], _R]]: ...

def copy_signature(
    source: Callable[..., Any],
    *,
    params_only: bool = True,
    is_method: bool = False,
):
    """
    Static typing trick: casts the decorated callable to appear to have the same
    parameter list as `source`. If params_only=False, also ties the return type
    to `source`'s return type.

    - is_method=False: function-like, copies full params (_P).
    - is_method=True:  method-like, copies params after the first argument, while
                       preserving the decorated method's first arg type (_A1).
    """

    # runtime implementation is just a cast; overloads do all the real work
    def _decorator(func: Callable[..., Any]):
        return cast(Any, func)

    return _decorator
```

Skylion007:
I'll merge this so we have something play around with:
Would validate_return or ParamOnly place more nicely together?

Skylion007 (commented):
FYI regarding the other suggestions

PR #163423, opened 2025-09-20
[SymmMem] Promote `@requires_nvshmem` instead of `enable_triton`

kwen2501:
* __->__ #163423

### Issue
The previous `enable_triton` UI requires the user-defined Triton kernel have a "nvshmem" in its name.
If users did not do so, the kernel would miss the NVSHMEM init, and silently hit CUDA IMA. 

The `@require_nvshmem` decorator eliminates the above name requirement (and the `enable_triton` call).

### Usage:
```
@requires_nvshmem
@triton.jit
def foo(...):
    ...

foo[(1, 1)](...)
```
It also remove the need of passing `extern_lib` to `foo` (handled by the decorator now).

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @gujinghui @PenghuiCheng @jianyuh @min-jean-cho @yanbing-j @Guobing-Chen @Xia-Weiwen @snadampal @mcarilli @ptrblck @leslie-fang-intel @voznesenskym @penguinwu @EikanWang @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela

### DIFF
--- a/torch/distributed/_symmetric_memory/_nvshmem_triton.py
+++ b/torch/distributed/_symmetric_memory/_nvshmem_triton.py
@@ -5,7 +5,6 @@ import sysconfig
 from typing import Any, Optional
 
 import torch.distributed as dist
-from torch.distributed._distributed_c10d import _nvshmemx_cumodule_init
 from torch.utils._triton import has_triton
 
 
@@ -155,6 +154,7 @@ def _nvshmem_init_hook(*args, **kwargs) -> None:  # type: ignore[no-untyped-def]
     A hook function to initialize the CUModule created by `triton.jit` with
     NVSHMEM device context
     """
+    from torch._C._distributed_c10d import _nvshmemx_cumodule_init
     jit_function = kwargs["fn"].jit_function
     fn_name = jit_function.fn.__name__
 
--- a/torch/distributed/_symmetric_memory/_nvshmem_triton.py
+++ b/torch/distributed/_symmetric_memory/_nvshmem_triton.py
@@ -176,6 +176,27 @@ def _nvshmem_init_hook(*args, **kwargs) -> None:  # type: ignore[no-untyped-def]
             )
 
 
+if has_triton():
+    from triton.runtime.jit import JITFunction, KernelInterface
+
+    # Create a new Callable class that follows the KernelInterface protocol so
+    # that the Callable works with the subscript operator, e.g. `foo[(1, 1)]`
+    class GridCallableWithExtern(KernelInterface):
+        """
+        `KernelInterface` invokes `self.run` in `__getitem__`, i.e. [].  We
+        implement a `run` method by directing the call to `JITFunction.run`,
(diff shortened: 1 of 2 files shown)


torch/distributed/_symmetric_memory/_nvshmem_triton.py
```diff
@@ -5,114 +5,235 @@
...
-                            if p and p not in paths:
-                                paths.append(p)
-            except subprocess.CalledProcessError:
-                pass
 
-    except ImportError:
-        pass
+def enable_triton(lib_dir: Optional[str] = None) -> dict[str, str]:
+    raise NotImplementedError(
+        "`enable_triton` is deprecated. "
```
ezyang, line 104: Err, do you have to go through a real deprecation cycle? Otherwise it's not really deprecated now lol, you removed it!

torch/distributed/_symmetric_memory/_nvshmem_triton.py
```diff
@@ -5,114 +5,235 @@
...
+        with added extern_libs kwarg, so that users don't have to pass it
+        """
+
+        def run(self, *args, grid, warmup, **kwargs):  # type: ignore[no-untyped-def]
+            # Call the JITFunction.run with added extern_libs kwarg
+            return jit_func.run(
+                *args, **kwargs, grid=grid, warmup=warmup, extern_libs=extern_libs
+            )
+
+    return MyGridCallable()
```
ezyang: The changes here and above are pretty involved, so I guess I'll let @ngimel  look.

torch/distributed/_symmetric_memory/_nvshmem_triton.py
kwen2501: The feature is not released yet (in whatever alpha/beta sense). This is so that we improve it for the first release.

torch/distributed/_symmetric_memory/_nvshmem_triton.py
ngimel: This looks good, but I'd structure it in the same way triton native decorators are structured https://github.com/triton-lang/triton/blob/1c03c462795670f6c85faff6d798d91a7eb7168e/python/triton/runtime/autotuner.py#L462 (it's very close)

ngimel (approved):

torch/distributed/_symmetric_memory/_nvshmem_triton.py
kwen2501: Thanks! Modified. Wish I have seen it earlier!

kwen2501 (commented):
the rebase looks good to me

Issue #163314, opened 2025-09-19
Flex Attention Decoding Silently Incorrect with H=None Broadcasting and Sliced mask_mod


zaptrem:
## 🐛 Describe the bug

Two bugs in `torch.nn.attention.flex_attention` cause silent incorrect results when used for autoregressive decoding with BlockMask slicing.

### Bug 1: H=None Broadcasting Produces Incorrect Results
When creating a BlockMask with `H=None` (which should broadcast to any head count), the attention computation produces silently incorrect results.

```python
# This should work but produces wrong results
block_mask = create_block_mask(causal_mask, B=4, H=None, Q_LEN=512, KV_LEN=512, ...)
```

**Expected:** H=None should broadcast correctly to match the actual number of attention heads
**Actual:** Results differ significantly from ground truth (max error ~2.77)

### Bug 2: Sliced BlockMask mask_mod Becomes Noop
After slicing a BlockMask, the `mask_mod` attribute is replaced with a noop function. If users attempt to modify the sliced mask's `mask_mod` directly (instead of using the base mask's `mask_mod`), results are silently incorrect.

```python
base_mask = create_block_mask(...)
sliced_mask = base_mask[:, :, block_offset]
# sliced_mask.mask_mod is now noop_mask, not the original mask function!
sliced_mask.mask_mod = get_mask_mod_with_offset(sliced_mask.mask_mod, offset)  # WRONG - uses noop
sliced_mask.mask_mod = get_mask_mod_with_offset(base_mask.mask_mod, offset)    # CORRECT
```

**Expected:** Clear documentation or error when mask_mod is incorrectly modified (or just maintain the correct mask mod function)
**Actual:** Silent incorrect results (max error ~0.18)

## Impact
These bugs affect autoregressive generation use cases where BlockMask slicing is used for efficient decoding, as described in the [PyTorch blog post on flex decoding](hhttps://pytorch.org/blog/flexattention-for-inference/).

## Reproduction

https://gist.github.com/zaptrem/bd47b71fa2908ebbc5b2a26c6fcf5070

Test script attached demonstrates all three scenarios:
1. ✅ Correct implementation (H=8, base mask_mod)
2. ❌ H=None broadcast bug (max error: 2.77)
3. ❌ Wrong mask_mod application bug (max error: 0.18)

All scenarios are validated against `F.scaled_dot_product_attention` as ground truth.

## Versions

```
PyTorch version: 2.8.0+cu126
Is debug build: False
[...]

PR #163426, opened 2025-09-20
[Flex attention] Fix flex attention head broadcast

Isalia20:
Fixes part of #163314

In particular bug: **Bug 1: H=None Broadcasting Produces Incorrect Results**

This fixes a shape bug when slicing BlockMask on the Q-tile axis with an int (**mask[:, :, i]**). That form of indexing collapses the Q dimension, so kv_num_blocks/kv_indices lose their expected [B, H, Q_tiles, …] shape. Due to them losing shape, even though the mask_mod remains "interpretable", the kernel’s stride math then reads wrong offsets. Due to this we get silent numerical mismatches compared to regular SDPA, especially when single position decoding/H broadcasting.

The B=None, H=None works case is accidental: with singleton batch/head the kernel maps to index 0 via `sparse_idx_z = off_zq % 1` and `sparse_idx_hq = off_hq % 1` and with a single Q tile `q_start // SPARSE_Q_MULTIPLE = 0`. The missing Q-tiles stride is multiplied by 0, so the bad offset from the collapsed Q axis doesn’t move the pointer and it happens to read the first tile correctly. Once H > 1 or there are multiple Q tiles, those terms become nonzero and the kernel indexes with wrong strides which causes silent error

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Chillee @drisspg @yanboliang @BoyuanFeng

### DIFF
--- a/torch/nn/attention/flex_attention.py
+++ b/torch/nn/attention/flex_attention.py
@@ -650,7 +650,13 @@ class BlockMask:
         """
         index = (index,) if not isinstance(index, tuple) else index
         padded = (*index, slice(None), slice(None), slice(None))[:3]
-        index = tuple(slice(i, i + 1) if isinstance(i, int) else i for i in padded)
+        sizes = self.kv_num_blocks.shape[:3]
+        index = tuple(
+            (slice(i + n, i + n + 1) if -n <= i < 0 else slice(i, i + 1))
+            if isinstance(i, int)
+            else i
+            for i, n in zip(padded, sizes)
+        )
         new_kv_num_blocks = self.kv_num_blocks[index]
         new_kv_indices = self.kv_indices[index]
         if self.full_kv_num_blocks is not None:
--- a/torch/nn/attention/flex_attention.py
+++ b/torch/nn/attention/flex_attention.py
@@ -648,6 +648,9 @@ class BlockMask:
                 assert new_block_mask.kv_num_blocks.shape == (2, 1, 1)
                 assert new_block_mask.kv_indices.shape == (2, 1, 1, 4)
         """
+        index = (index,) if not isinstance(index, tuple) else index
+        padded = (*index, slice(None), slice(None), slice(None))[:3]
+        index = tuple(slice(i, i + 1) if isinstance(i, int) else i for i in padded)
         new_kv_num_blocks = self.kv_num_blocks[index]
         new_kv_indices = self.kv_indices[index]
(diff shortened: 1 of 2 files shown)


Skylion007:
Good backport candidate into our latest RC!

torch/nn/attention/flex_attention.py
```diff
@@ -648,6 +648,9 @@ def causal_mask(b, h, q_idx, kv_idx):
                 assert new_block_mask.kv_num_blocks.shape == (2, 1, 1)
                 assert new_block_mask.kv_indices.shape == (2, 1, 1, 4)
         """
+        index = (index,) if not isinstance(index, tuple) else index
+        padded = (*index, slice(None), slice(None), slice(None))[:3]
+        index = tuple(slice(i, i + 1) if isinstance(i, int) else i for i in padded)
```
drisspg: Does this handle negative ints correctly?

drisspg (commented):
Also I think this might change the shape of the return tensors right? that actually might be a good thing but would double check this doesn't break other blockmask tests

torch/nn/attention/flex_attention.py
Isalia20: It handles all of them except for -1 unfortunately, updated it

Isalia20:
Yes it changes shape of the getitem on blockmask, updated other tests

drisspg (approved):
Thank you!

drisspg (commented):
FWIW the main usage for this AFAIK is in gpt-fast: https://github.com/meta-pytorch/gpt-fast/blob/6ecad9b5b6b987d17ac4303965545873d0192086/generate.py#L74

and this is using tensors as an index and so we keep the dim. The slicing operation feels a little weird tbh I would prefer if users manually edited the bits and then created a new BM from_kv_blocks and we wouldn't run into the `not setting a score mod` problem 

This seems to be a pretty  big foot gun for what is essentially syntactic sugar.  So, IMO I think its okay to land and call it a bug fix.

Can you also update the PR w/ a description as to why this fixes the issue. AFAIK this looks like essentially a bad interaction where the sliced mask_mod w/ 1 less shape is interpretable even though it shouldn't be and so the kernel is reading bogus values

PR #163430, opened 2025-09-21
[CI][CUDA][B200] Skip tests to unblock #159494

nWEIdia:
Help unblock #159494  which fixes the issue that there are no per-commit B200 tests even though runners are available in https://hud.pytorch.org/runners/pytorch 

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @ptrblck @eqy @tinglvv @atalman @malfet @drisspg @ngimel

### DIFF
--- a/test/inductor/test_max_autotune.py
+++ b/test/inductor/test_max_autotune.py
@@ -103,6 +103,7 @@ class FailChoiceCaller(ChoiceCaller):
         raise RuntimeError("This choice caller will always throw")
 
 
+# xfail on SM100 due to https://github.com/pytorch/pytorch/issues/163429
 @xfailIfSM100OrLater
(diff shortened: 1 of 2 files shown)


malfet (changes requested):
Hmm, is it possible to create a decorator specifically for sm_120 rather than skip test unconditionally? And also prefer xfail to skip

nWEIdia:
Note:  to test this change, we need this to be merged to #159494 to test on B200 as the trunk is missing B200.

PR #163431, opened 2025-09-21
[BE] Remove multinterpreter logic from pythonFallbackKernel and python_dispatch

PaliC:
Removes some logic that assumes we have multiple interpreters running around and instead uses the global interpreter.

Also remove a giant comment about pyinterpreter tags I forgot to remove earlier.


torch/csrc/utils/python_dispatch.cpp
```diff
@@ -155,37 +156,35 @@ class PythonKernelHolder : public c10::OperatorKernel {
     const auto& schema = op.schema();
     const auto num_arguments = schema.arguments().size();
 
-    // Otherwise, find a PyInterpreter on a Tensor IF if has Python key (which
-    // means it's a nontrivial tensor subclass)
+    // Check if any tensors have Python key and use global interpreter
```
albanD, line 155: You can remove a lot more here. Even the mode logic above is just to try and fin an interpreter !

torch/csrc/utils/python_dispatch.cpp
```diff
@@ -155,37 +156,35 @@ class PythonKernelHolder : public c10::OperatorKernel {
...
+            has_python_tensor = true;
+            break;
           }
         }
+        if (has_python_tensor)
+          break;
+      }
+    }
+
+    if (has_python_tensor) {
```
albanD, line 177: Is this really expected to be a noop if there are not subclass?

aten/src/ATen/core/PythonFallbackKernel.cpp
```diff
@@ -64,47 +65,24 @@ void pythonFallback(const c10::OperatorHandle& op, c10::DispatchKeySet dispatch_
...
-  // It is safe to dispatch on the very first Tensor with a pyobj_interpreter
-  // without checking the interpreters of any of the arguments, because when
-  // we actually run dispatch(), we will take out PyObjects in the context
-  // of that interpreter, and this will ensure that everyone is on the same
-  // interpreter.
   bool tensors_with_python_key_present = false;
-  c10::impl::PyInterpreter* interpreter = nullptr;
+  c10::impl::PyInterpreter* interpreter = c10::impl::getGlobalPyInterpreter();
+
+  // Still need to check if any tensors have the Python key
```
albanD, line 74: Please update all the comments to not reflect what you are doing in the PR but the new state of the world.
"still" or "instead" shouldn't really be there.

torch/csrc/utils/python_dispatch.cpp
```diff
@@ -146,46 +146,38 @@ class PythonKernelHolder : public c10::OperatorKernel {
     if (mode_stack_len > 0) {
       const auto& cur_torch_dispatch_mode_state =
           c10::impl::TorchDispatchModeTLS::get_stack_at(mode_stack_len - 1);
-      cur_torch_dispatch_mode_state->pyinterpreter()
```
albanD, line 149: That doesn't look right

PR #163432, opened 2025-09-21
[precompile] Add option to disable guard check on aot-compiled function.

zhxchen17:
Summary:
Under circumstances it seems reasonable to return a callable directly without guard check when user use aot_compile on a function with single compilation result.

When having multiple entries (aot_compile_module), we should start enabling guard check to differetiate different compiled functions apart.

Test Plan: CI


dolpm (approved):
i feel this is error-prone, but ig it is better than just disabling by default.

mlazos (approved):

PR #163422, opened 2025-09-20
[inductor] Fix error from custom CUDA allocators

jansel:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.12.0) (oldest at bottom):
* #163584
* #163377
* #163482
* #163520
* #163481
* __->__ #163422
* #163412
* #163393
* #163434
* #163419

Fixes #163257

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben

PR #163393, opened 2025-09-20
[inductor] Support out_dtype arg to matmul

jansel:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.12.0) (oldest at bottom):
* #163584
* #163377
* #163482
* #163520
* #163481
* #163422
* #163412
* __->__ #163393
* #163434
* #163419

Fixes #163275

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @mlazos

PR #163434, opened 2025-09-21
[inductor] Freeze layouts in FlexAttention

jansel:
* #163584
* #163377
* #163482
* #163520
* #163481
* #163422
* #163412
* #163393
* __->__ #163434
* #163419

Fixes #163300


torch/_inductor/kernel/flex/flex_attention.py
```diff
@@ -212,6 +215,9 @@ def flex_attention(
     score_mod_other_buffers = maybe_realize(score_mod_other_buffers)
```
drisspg, line 221: so we have something similar ish w/ realizing input nodes, for my own understanding even though we realize these buffers it is possible that the layouts for these nodes are changed but subsequent nodes?

jansel: In inductor you have 3 levels:
1) Unrealized: no buffer is allocated usages are inlined into users
2) Realized (FlexibleLayout): a buffer is allocated, but the strides can be changed by optimizations
3) Realized (FrozenLayout): a buffer is allocated, with fixed strides

`.realize()` is about going from 1 to 2.   `.freeze_layout()` goes from 2 to 3.

drisspg (approved):

PR #163437, opened 2025-09-21
[Code Clean] Replace std::runtime_error with TORCH_CHECK

### LINKED ISSUE #148114: We should never throw vanilla C++ exceptions

zhudada0120:
Replace the runtime_error of the vallina C++ exceptions with TORCH_CEHCK
Including:
- torch/csrc/export
- torch/csrc/cuda

Fixes #148114

### DIFF
--- a/torch/csrc/cuda/utils.cpp
+++ b/torch/csrc/cuda/utils.cpp
@@ -1,3 +1,4 @@
+#include <c10/util/Exception.h>
 #include <torch/csrc/Stream.h>
 #include <torch/csrc/cuda/THCP.h>
 #include <torch/csrc/python_headers.h>
@@ -8,18 +9,17 @@
 // whatever the current stream of the device the input is associated with was.
 std::vector<std::optional<at::cuda::CUDAStream>>
 THPUtils_PySequence_to_CUDAStreamList(PyObject* obj) {
-  if (!PySequence_Check(obj)) {
-    throw std::runtime_error(
-        "Expected a sequence in THPUtils_PySequence_to_CUDAStreamList");
-  }
+  TORCH_CHECK(
+      PySequence_Check(obj),
+      "Expected a sequence in THPUtils_PySequence_to_CUDAStreamList");
   THPObjectPtr seq = THPObjectPtr(PySequence_Fast(obj, nullptr));
-  if (seq.get() == nullptr) {
-    throw std::runtime_error(
-        "expected PySequence, but got " + std::string(THPUtils_typename(obj)));
-  }
+  TORCH_CHECK(
+      seq.get() != nullptr,
+      "expected PySequence, but got " + std::string(THPUtils_typename(obj)));
 
   std::vector<std::optional<at::cuda::CUDAStream>> streams;
   Py_ssize_t length = PySequence_Fast_GET_SIZE(seq.get());
+  streams.reserve(length);
   for (Py_ssize_t i = 0; i < length; i++) {
     PyObject* stream = PySequence_Fast_GET_ITEM(seq.get(), i);
 
@@ -34,7 +34,8 @@ THPUtils_PySequence_to_CUDAStreamList(PyObject* obj) {
     } else if (stream == Py_None) {
       streams.emplace_back();
     } else {
-      throw std::runtime_error(
+      TORCH_CHECK(
+          false,
           "Unknown data type found in stream list. Need torch.cuda.Stream or None");
     }
   }
--- a/torch/csrc/export/upgrader.cpp
+++ b/torch/csrc/export/upgrader.cpp
@@ -1,3 +1,4 @@
+#include <c10/util/Exception.h>
 #include <torch/csrc/export/upgrader.h>
 #include <limits>
 #include <map>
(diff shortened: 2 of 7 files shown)


zhudada0120:
PTAL @eqy @syed-ahmed @FFFrog 
Thanks.

cyyever:
@zhudada0120 Please fix the linter.

torch/csrc/cuda/utils.cpp
```diff
@@ -3,20 +3,15 @@
...
-    throw std::runtime_error(
-        "Expected a sequence in THPUtils_PySequence_to_CUDAStreamList");
-  }
+  TORCH_CHECK(PySequence_Check(obj), "Expected a sequence in THPUtils_PySequence_to_CUDAStreamList");
   THPObjectPtr seq = THPObjectPtr(PySequence_Fast(obj, nullptr));
-  if (seq.get() == nullptr) {
-    throw std::runtime_error(
-        "expected PySequence, but got " + std::string(THPUtils_typename(obj)));
-  }
+  TORCH_CHECK(seq.get(), "expected PySequence, but got " + std::string(THPUtils_typename(obj)));
```
Skylion007: Don't rely on implicit Boolean conversion here: explicitly check against nullptr please

torch/csrc/cuda/utils.cpp
```diff
@@ -3,20 +3,15 @@
...
 
   std::vector<std::optional<at::cuda::CUDAStream>> streams;
```
Skylion007, line 20: While we are here, call streams.reserve(length);

torch/csrc/export/upgrader.cpp
```diff
@@ -27,9 +28,7 @@ static nlohmann::json getFieldByKeypath(
     const std::vector<std::string>& keypath) {
   nlohmann::json current = obj;
```
Skylion007: Get field by keypath should be pass by value and obj should be std::move in

torch/csrc/export/upgrader.cpp
```diff
@@ -42,14 +41,10 @@ static void setFieldByKeypath(
   nlohmann::json* current = &obj;
```
Skylion007, line 40: I don't think current can be nullptr, why isn't it just a reference?

torch/csrc/export/upgrader.cpp
```diff
@@ -176,16 +171,14 @@ void throwUpgraderError(
     error_stream << "\nProblematic object: " << problematic_object.dump(2);
   }
 
-  throw std::runtime_error(error_stream.str());
+  TORCH_CHECK(false, error_stream.str());
 }
 
 nlohmann::json upgrade(const nlohmann::json& artifact, int target_version) {
   auto current_artifact = artifact;
```
Skylion007: A lot of these should be move by value and copy into their local temporary

Skylion007 (approved):
Minus the nullptr check looks good. These files need a better engineering pass though as there are a lot unnecessary copies particularly on potentially really large JSONs

torch/csrc/export/upgrader.cpp
```diff
@@ -23,13 +24,11 @@ static const std::multiset<Upgrader>& getUpgrader(int current_version) {
 }
 
 static nlohmann::json getFieldByKeypath(
-    const nlohmann::json& obj,
+    nlohmann::json obj,
     const std::vector<std::string>& keypath) {
-  nlohmann::json current = obj;
```
cyyever, line 28: Why not just rename inferences of `current` to `obj` and remove the `current` variable?

torch/csrc/export/upgrader.cpp
```diff
@@ -38,19 +37,15 @@ static nlohmann::json getFieldByKeypath(
 static void setFieldByKeypath(
     nlohmann::json& obj,
     const std::vector<std::string>& keypath,
-    const nlohmann::json& value) {
-  nlohmann::json* current = &obj;
+    nlohmann::json value) {
+  nlohmann::json& current = obj;
```
cyyever: Same as above

torch/csrc/export/upgrader.cpp
```diff
@@ -176,16 +171,14 @@ void throwUpgraderError(
     error_stream << "\nProblematic object: " << problematic_object.dump(2);
   }
 
-  throw std::runtime_error(error_stream.str());
+  TORCH_CHECK(false, error_stream.str());
 }
 
-nlohmann::json upgrade(const nlohmann::json& artifact, int target_version) {
-  auto current_artifact = artifact;
+nlohmann::json upgrade(nlohmann::json artifact, int target_version) {
```
cyyever, line 178: Same as above

cyyever (changes requested):

torch/csrc/cuda/utils.cpp
zhudada0120: Thanks for the suggestion! I've updated the code to explicitly check against nullptr (instead of relying on implicit Boolean conversion). PTAL. Thanks!

torch/csrc/cuda/utils.cpp
zhudada0120: Added streams.reserve(length) to pre-allocate memory for the vector, which should improve performance by avoiding multiple reallocations. PTAL, thanks!

torch/csrc/export/upgrader.cpp
zhudada0120: Thanks for the suggestion! I’ve fixed multiple related places: changed the parameters here to pass by value, and used std::move to move them into local temporary variables. PTAL, thanks!

torch/csrc/export/upgrader.cpp
zhudada0120: Thanks for the suggestion! Updated to pass `obj` by value and use `std::move` to leverage move semantics, which should reduce copying overhead for the json object. PTAL, thanks!

torch/csrc/export/upgrader.cpp
zhudada0120: You’re right, current is never actually null. But later, we do need to rebind current as we go through the keypath, and pointers work better here because they let us reassign them to point to different nested parts. So I think using a pointer here makes sense.

torch/csrc/export/upgrader.cpp
zhudada0120: Thanks for your suggestion! I have removed current_artifact here and replaced it with artifact. PTAL.

torch/csrc/export/upgrader.cpp
zhudada0120: Thanks for your suggestion! I have removed current here and replaced it with obj. PTAL.

torch/csrc/export/upgrader.cpp
```diff
@@ -23,34 +24,29 @@ static const std::multiset<Upgrader>& getUpgrader(int current_version) {
...
-  nlohmann::json current = obj;
   for (const auto& key : keypath) {
-    if (!current.contains(key)) {
-      throw std::runtime_error("Keypath not found: " + key);
-    }
-    current = current[key];
+    TORCH_CHECK(obj.contains(key), "Keypath not found: " + key);
+    obj = obj[key];
   }
   return current;
```
cyyever: return obj?

cyyever (changes requested):

torch/csrc/export/upgrader.cpp
zhudada0120: Sorry, I’ve updated it as suggested.

cyyever (approved):

PR #163438, opened 2025-09-21
[BC breaking] Remove deprecated imports for torch.utils.data.datapipes.iter.grouping

cyyever:
This PR removes import tricks of `SHARDING_PRIORITIES` and  `ShardingFilterIterDataPipe` from `torch.utils.data.datapipes.iter.grouping`. They are declared to be removed in PyTorch 2.1 but not.
Before change:
```
import torch.utils.data.datapipes.iter.grouping.SHARDING_PRIORITIES
import torch.utils.data.datapipes.iter.grouping.ShardingFilterIterDataPipe
```
works
After change:
there is an import error exception.

### DIFF
--- a/test/test_datapipe.py
+++ b/test/test_datapipe.py
@@ -3356,63 +3356,6 @@ class TestSharding(TestCase):
         with self.assertRaises(Exception):
             dp.apply_sharding(2, 1, sharding_group=SHARDING_PRIORITIES.DEFAULT)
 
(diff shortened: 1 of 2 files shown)


janeyx99 (approved):
Nice, but please write up a BC breaking note in the PR Description similar to in https://docs.google.com/document/d/14OmgGBr1w6gl1VO47GGGdwrIaUNr92DFhQbY_NEk8mQ/edit?tab=t.0#heading=h.xhn730wluw68

PR #163446, opened 2025-09-21
[AMP][Refactor] Simplify dtype support logic in autocast context manager

KarhouTam:
## Description:

This PR refactors the autocast context manager in `autocast_mode.py` to simplify and centralize the logic for checking supported dtypes for each device. The previous implementation repeated similar checks for multiple device types. Now, a single mapping `device_supported_dtypes` is used to associate device types with their supported dtypes, and the validation logic is unified. 

In my view, this makes the code easier to maintain and extend for new devices.

Please share any suggestions and comments with me.

BTW, in the original `xla` branch, the `supported_dtype` are `[torch.float16, torch.bfloat16]`, https://github.com/pytorch/pytorch/blob/5d8a226e23339e7243a2a84afd174f685f145b68/torch/amp/autocast_mode.py#L358-L363 but the warning message has only `torch.bfloat16`.

cc @mcarilli @ptrblck @leslie-fang-intel @jgong5

### DIFF
--- a/torch/amp/autocast_mode.py
+++ b/torch/amp/autocast_mode.py
@@ -278,6 +278,20 @@ class autocast:
             else self.device.upper()
         )
         if enabled:
+            # Special case for CUDA AMP and bfloat16 support
+            if self.device == "cuda":
+                if torch.cuda.amp.common.amp_definitely_not_available():
+                    warnings.warn(
+                        "CUDA is not available or torch_xla is imported. AMP disabled."
+                    )
+                    enabled = False
+                elif (
+                    self.fast_dtype == torch.bfloat16
+                    and not torch.cuda.is_bf16_supported()
+                ):
+                    raise RuntimeError(
+                        "Current CUDA Device does not support bfloat16. Please switch dtype to float16."
+                    )
             if self.fast_dtype not in supported_dtypes:
                 error_message = (
                     f"In {device_name} autocast, but the target dtype is not supported. Disabling autocast.\n"
@@ -299,20 +313,6 @@ class autocast:
                 )
                 warnings.warn(error_message)
                 enabled = False
-            # Special case for CUDA AMP and bfloat16 support
-            elif self.device == "cuda":
-                if torch.cuda.amp.common.amp_definitely_not_available():
-                    warnings.warn(
-                        "CUDA is not available or torch_xla is imported. AMP disabled."
-                    )
-                    enabled = False
-                elif (
-                    self.fast_dtype == torch.bfloat16
-                    and not torch.cuda.is_bf16_supported()
-                ):
-                    raise RuntimeError(
-                        "Current CUDA Device does not support bfloat16. Please switch dtype to float16."
-                    )
         self._enabled = enabled
 
     def __enter__(self):
--- a/torch/amp/autocast_mode.py
+++ b/torch/amp/autocast_mode.py
@@ -284,18 +284,18 @@ class autocast:
         }
 
         if self.device in device_supported_dtypes:
-            supported_dtype = device_supported_dtypes[self.device]
+            supported_dtypes = device_supported_dtypes[self.device]
             device_name = (
                 self.device
                 if self.device == self.custom_backend_name
                 else self.device.upper()
             )
 
-            if self.fast_dtype not in supported_dtype and enabled:
+            if self.fast_dtype not in supported_dtypes and enabled:
                 error_message = (
                     f"In {device_name} autocast, but the target dtype is not supported. Disabling autocast.\n"
                     f"{device_name} Autocast only supports dtypes of "
-                    + ", ".join(str(dtype) for dtype in supported_dtype)
+                    + ", ".join(map(str, supported_dtypes))
                     + " currently."
                 )
                 warnings.warn(error_message)
--- a/torch/amp/autocast_mode.py
+++ b/torch/amp/autocast_mode.py
@@ -230,9 +230,7 @@ class autocast:
             raise ValueError(
                 f"Expected `device_type` of type `str`, got: `{type(device_type)}`"
             )
-        if dtype is None:
-            dtype = torch.get_autocast_dtype(device_type)
-        self.fast_dtype = dtype
+        self.fast_dtype = torch.get_autocast_dtype(device_type) if not dtype else dtype
         if torch._jit_internal.is_scripting():
             self._enabled = enabled
             self.device = device_type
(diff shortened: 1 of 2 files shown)


torch/amp/autocast_mode.py
```diff
@@ -273,96 +271,53 @@ def __init__(
...
-                )
-        elif self.device == "mps":
-            supported_dtype = [torch.bfloat16, torch.float16]
-            if self.fast_dtype not in supported_dtype:
                 error_message = (
-                    "In MPS autocast, but the target dtype is not supported. Disabling autocast.\n"
-                    "MPS Autocast only supports dtype of torch.bfloat16 and torch.float16 currently."
+                    f"In {device_name} autocast, but the target dtype is not supported. Disabling autocast.\n"
+                    f"{device_name} Autocast only supports dtypes of "
+                    + ", ".join(str(dtype) for dtype in supported_dtype)
```
Skylion007: Call '.join(map(str, supported_dtype))' marginally faster. Also shouldn't supported dtype be supported dtypes now?

KarhouTam: Thnaks. Applied.

torch/amp/autocast_mode.py
```diff
@@ -273,96 +271,53 @@ def __init__(
...
+
+        if self.device in device_supported_dtypes:
+            supported_dtypes = device_supported_dtypes[self.device]
+            device_name = (
+                self.device
+                if self.device == self.custom_backend_name
+                else self.device.upper()
+            )
+
+            if self.fast_dtype not in supported_dtypes and enabled:
```
janeyx99:
```suggestion
            if self.fast_dtype not in supported_dtypes and enabled:
```
This is the same if as in line 278 but every other instance does not have the `and enabled`. That appears to be different from the original behavior, yes?

KarhouTam: Yeah, I noticed that and have some questions: Is it intentional that only the `cpu` branch has the `and enabled` condition? I see that most other branches have the same code logic (concatenating the `error_message` and throwing a `warnings.warn(error_message`). So I don't know if there are some cases that I missed. 

In my view, if `enabled=False`, then these if branches can be skipped (like the `cpu` branch, line 278 previously). I just added `and enabled` within the if branches for most devices. Please correct me if I missed something or got something wrong. Thanks.

torch/amp/autocast_mode.py
```diff
@@ -273,96 +271,53 @@ def __init__(
         if cache_enabled is not None:
             self._cache_enabled = cache_enabled
```
fffrog:
```suggestion
        if (
            enabled
            and self.device == "cuda"
            and torch.cuda.amp.common.amp_definitely_not_available()
        ):
            warnings.warn(
                "User provided device_type of 'cuda', but CUDA is not available. Disabling"
            )
            enabled = False
        self._cache_enabled = cache_enabled if cache_enabled is not None else torch.is_autocast_cache_enabled()
```

torch/amp/autocast_mode.py
```diff
@@ -273,96 +271,53 @@ def __init__(
...
+            "cpu": [torch.bfloat16, torch.float16],
+            "mtia": [torch.bfloat16, torch.float16],
+            "maia": [torch.bfloat16, torch.float16],
+            "xpu": [torch.bfloat16, torch.float16],
+            "ipu": [torch.bfloat16, torch.float16],
+            "hpu": [torch.bfloat16, torch.float16],
+            "mps": [torch.bfloat16, torch.float16],
+            "xla": [torch.bfloat16, torch.float16],
+            self.custom_backend_name: self.custom_device_mod.get_amp_supported_dtype(),
+        }
```
fffrog:
```suggestion
default_dtypes = [torch.bfloat16, torch.float16]

device_supported_dtypes = defaultdict(
    lambda: default_dtypes,
    {
        self.custom_backend_name: self.custom_device_mod.get_amp_supported_dtype(),
    }
)
```

torch/amp/autocast_mode.py
```diff
@@ -273,96 +271,53 @@ def __init__(
...
+
+        if self.device in device_supported_dtypes:
```
fffrog: I don`t need this one, because the we have check the validation for device in this [line](https://github.com/pytorch/pytorch/blob/27164b6788cab6e6d8095012839e51c958a819d6/torch/amp/autocast_mode.py#L242)

torch/amp/autocast_mode.py
fffrog: Yes, not all branches will check the "enable" field, but I think adding this check is reasonable and won't cause any compatibility issues. If the user sets "enable" to False, then AMP shouldn't throw a data type error, right?

However, for clarity, I suggest placing "enable" at the beginning of that line.

```suggestion
if enabled:
    if self.fast_dtype not in supported_dtypes:
```

torch/amp/autocast_mode.py
```diff
@@ -273,96 +271,53 @@ def __init__(
...
+                    f"In {device_name} autocast, but the target dtype is not supported. Disabling autocast.\n"
+                    f"{device_name} Autocast only supports dtypes of "
+                    + ", ".join(map(str, supported_dtypes))
+                    + " currently."
                 )
                 warnings.warn(error_message)
                 enabled = False
-            elif self.fast_dtype == torch.bfloat16:
+            # Special case for MPS bfloat16 support on macOS < 14
+            elif self.device == "mps" and self.fast_dtype == torch.bfloat16 and enabled:
```
fffrog:
```suggestion
            if self.device == "mps" and self.fast_dtype == torch.bfloat16:
```

torch/amp/autocast_mode.py
```diff
@@ -273,96 +271,53 @@ def __init__(
...
-                enabled = False
+        elif (
+            self.device == "cuda"
+            and self.fast_dtype == torch.bfloat16
+            and not torch.cuda.is_bf16_supported()
+            and enabled
+        ):
+            raise RuntimeError(
+                "Current CUDA Device does not support bfloat16. Please switch dtype to float16."
+            )
```
fffrog:
```suggestion
            if (
            self.device == "cuda"
            and self.fast_dtype == torch.bfloat16
        ):
            if not torch.cuda.is_bf16_supported():
                raise RuntimeError(
                "Current CUDA Device does not support bfloat16. Please switch dtype to float16."
            )
```

torch/amp/autocast_mode.py
```diff
@@ -260,109 +258,61 @@ def __init__(
...
-                )
-                warnings.warn(error_message)
-                enabled = False
+                elif (
+                    self.fast_dtype == torch.bfloat16
+                    and not torch.cuda.is_bf16_supported()
+                ):
+                    raise RuntimeError(
+                        "Current CUDA Device does not support bfloat16. Please switch dtype to float16."
+                    )
```
KarhouTam, line 292: Hi, @FFFrog, followed your suggestion. About the `cuda` related if branches, I choose to merge them.

Also, I changed the error message from `User provided device_type of 'cuda', but CUDA is not available. Disabling` to `XLA Autocast only supports dtype of torch.bfloat16 currently`, according to function `amp_definitely_not_available()`:

```python
# torch/cuda/amp/common.py

def amp_definitely_not_available():
    return not (torch.cuda.is_available() or find_spec("torch_xla"))
```

Hope these changes are fine to you. And if any cases I've missed or got wrong, please tell me and I'll fix them asap.

torch/amp/autocast_mode.py
```diff
@@ -260,109 +258,61 @@ def __init__(
...
+        )
+
+        supported_dtypes = device_supported_dtypes[self.device]
+        device_name = (
+            self.device
+            if self.device == self.custom_backend_name
+            else self.device.upper()
+        )
+        if enabled:
             if self.fast_dtype not in supported_dtypes:
```
KarhouTam: Hi, @FFFrog , followed your suggestion, I extracted the `if enabled` and remove other `and enabled` codes below.

torch/amp/autocast_mode.py
```diff
@@ -260,109 +258,61 @@ def __init__(
...
+            if cache_enabled is None
+            else cache_enabled
+        )
+        device_supported_dtypes = collections.defaultdict(
+            lambda: [torch.bfloat16, torch.float16],  # default supported dtypes
+            {
+                # Add special cases here (device that not supports both torch.bfloat16 and torch.float16)
+                self.custom_backend_name: self.custom_device_mod.get_amp_supported_dtype()
+            },
+        )
```
KarhouTam: Hi, @FFFrog , followed your suggestion, the dict is changed to a `defaultdict` and I make some comments for better readability.

KarhouTam (commented):
Hi, @FFFrog , thanks for your valuable CRs. I've followed your suggestions and refactored the code. PTAL.

cc: @janeyx99 @Skylion007

fffrog (approved):
SGTM, thank you.

Hi @albanD @ezyang @Skylion007, please help to look at this one.

albanD (approved):
Nice cleanup, thanks!

ezyang:
test fails look real

KarhouTam:
Thanks for the reminder! Sorry, I can't handle this in time because I'm on vacation. I'll deal with the CI-failed tests as soon as I return. (BTW, I watched your YT video about PyTorch Dynamo; it was very inspiring. Thank you for that!)

ezyang:
mps failure probably real

PR #163459, opened 2025-09-21
switch from stack based to graph based aproach

laithsakka:
* __->__ #163459


bobrenjc93 (approved):
Not reviewing very closely since experimental and we'll be doing some heart surgery as discussed offline.

PR #163460, opened 2025-09-22
Triton template IMA reads on B200

drisspg:
* #159494
* #163537
* __->__ #163460


eqy (approved):
I guess this is OK, I was thinking maybe some kind of `xFail` when simulating `--repeat` but an IMA would corrupt the CUDA context anyway so it would need to be run in a subprocess

alexsamardzic (approved):

PR #163464, opened 2025-09-22
[BE] Remove HermeticPyObjectTLS and Simplify PythonOpRegistrationTrampoline

PaliC:
Removes HermeticPyObjectTLS as we no longer need since torch deploy is no longer supported. PythonOpRegistrationTrampoline is also drastically simplified as and being prepped for removal in a future PR.


PaliC:
* (to be filled)

c10/core/impl/HermeticPyObjectTLS.h
```diff
@@ -1,62 +0,0 @@
-#pragma once
-
-#include <c10/macros/Export.h>
-#include <atomic>
-
-namespace c10::impl {
```
albanD, line 6: Please do a quick github search to make sure we don't have big users of this

albanD (approved):
SGTM

Skylion007 (approved):

PR #163473, opened 2025-09-22
remove allow-untyped-defs from ./torch/utils/benchmark/op_fuzzers/unary.py

bobrenjc93:
* #163476
* __->__ #163473
* #163475
* #163474
* #163471
* #163478
* #163477


torch/utils/benchmark/op_fuzzers/unary.py
```diff
@@ -14,7 +25,9 @@
 
 
 class UnaryOpFuzzer(Fuzzer):
-    def __init__(self, seed, dtype=torch.float32, cuda=False):
+    def __init__(self, seed: Optional[int], dtype: _dtype | None = None, cuda: bool = False) -> None:
+        if dtype is None:
+            dtype = _DEFAULT_DTYPE
```
Skylion007, line 30:
```suggestion
            dtype = _DEFAULT_DTYPE
            assert dtype is not None
```
Instead of the weird TYPE_CHECKING stuff you are doing above.

PR #163474, opened 2025-09-22
remove allow-untyped-defs from ./torch/distributions/half_normal.py

bobrenjc93:
* #163476
* #163473
* #163475
* __->__ #163474
* #163471
* #163478
* #163477


ezyang (approved):

torch/distributions/half_normal.py
```diff
@@ -58,26 +59,28 @@ def mean(self) -> Tensor:
 
     @property
     def mode(self) -> Tensor:
-        return torch.zeros_like(self.scale)
+        return self.scale * 0.0
```
ezyang, line 62: nuh uh, don't do that

ezyang (changes requested):
.

PR #163476, opened 2025-09-22
remove allow-untyped-defs from ./torch/utils/benchmark/op_fuzzers/sparse_unary.py

bobrenjc93:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #163476
* #163471
* #163475
* #163478

PR #163478, opened 2025-09-22
remove allow-untyped-defs from ./torch/fx/experimental/unification/multipledispatch/core.py

bobrenjc93:
* #163476
* #163471
* #163475
* __->__ #163478

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci @EikanWang @jgong5 @wenzhe-nrv


ezyang (approved):

torch/fx/experimental/unification/core.py
```diff
@@ -77,7 +77,7 @@ def reify(e, s):
 seq = tuple, list, Iterator
 
 
-@dispatch(seq, seq, dict)
+@dispatch((tuple, list, Iterator), (tuple, list, Iterator), dict)  # type: ignore[arg-type]
```
ezyang: this is pretty pointless change if you're just going to ignore this line

torch/fx/experimental/unification/multipledispatch/core.py
```diff
@@ -1,15 +1,17 @@
...
 
-global_namespace = {}  # type: ignore[var-annotated]
+global_namespace: dict[str, Dispatcher] = {}
 
 __all__ = ["dispatch", "ismethod"]
 
 
-def dispatch(*types, **kwargs):
+def dispatch(
+    *types: type[Any], **kwargs: Any
```
Skylion007: Hold on, this any should be a TypeVar that is the return type of Callable I'm guessing, right? Annotate it as such. Actually, this should probably be a TypeVarTuple, that can be expanded. That way we can check if the Callable supports the common ancestor type.

PR #163480, opened 2025-09-22
fix: handle zero local shards correctly in distributed checkpoint

rphmeier:
This fixes a subtle issue that can occur when there are zero local shards for a sharded tensor in a saved distributed checkpoint.

The following sequence of events could occur:
  1. `_flatten_sharded_tensors` would filter out the item from the state dict
  2. Then `DefaultLoadPlanner.create_local_plan` would wrongly treat the key as missing
  3. Which would then lead it to fall back to the old key format and fail

I fixed this by having `_flatten_sharded_tensors` return a set of paths that were being omitted so they wouldn't count as missing keys in the checkpoint.

(note: i observed this issue when saving/loading parameters of shape `1, N, D` and `1, D`, but it seems like a general issue)

### DIFF
--- a/torch/distributed/checkpoint/_sharded_tensor_utils.py
+++ b/torch/distributed/checkpoint/_sharded_tensor_utils.py
@@ -11,15 +11,19 @@ from torch.distributed.remote_device import _remote_device
 from ._traverse import OBJ_PATH, set_element, STATE_DICT_ITEM, traverse_state_dict
 from .utils import _element_wise_add, _normalize_device_info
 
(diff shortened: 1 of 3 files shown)


ezyang:
This seems like it could be a legitimate fix but it will be difficult to evaluate without a test

rphmeier:
I don't know this codebase at all so I suspect the maintainers would be able to write a test much more quickly than I could. If you leave some pointers about where to add the test I will get around to it eventually

PR #163484, opened 2025-09-22
Add weights_only=True to torch.load

cyyever:
This PR adds `weights_only=True` to remaining `torch.load` calls and enables `TOR102` in pyflake8.

### DIFF
--- a/torch/distributed/checkpoint/_experimental/checkpoint_reader.py
+++ b/torch/distributed/checkpoint/_experimental/checkpoint_reader.py
@@ -78,7 +78,7 @@ class CheckpointReader:
 
         if state_dict is None:
             result: tuple[STATE_DICT, list[str]] = (
-                torch.load(file_path, map_location=map_location),
+                torch.load(file_path, map_location=map_location, weights_only=True),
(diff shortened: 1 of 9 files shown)


torch/distributed/checkpoint/_experimental/checkpoint_reader.py
```diff
@@ -78,7 +78,7 @@ def read(
 
         if state_dict is None:
             result: tuple[STATE_DICT, list[str]] = (
-                torch.load(file_path, map_location=map_location),
+                torch.load(file_path, map_location=map_location, weights_only=True),
```
albanD, line 81: I'm not sure if this one is ok to change.
We need to either:
- put the two in this file to False and we can merge this PR
- wait for someone from the dcp team to tell us if this is ok.

cyyever: `weights_only` defaults to `True` in `torch.load`, so the semantics doesn't change after the fix. Or it should be `False` and the current behaviour is wrong?

PR #163485, opened 2025-09-22
symintify fill_diagonol_

bobrenjc93:
* #163607
* __->__ #163485

Fixes #162271


aten/src/ATen/native/Fill.cpp
```diff
@@ -121,19 +120,19 @@ Tensor& fill_diagonal_(Tensor& self, const Scalar& fill_value, bool wrap) {
...
 
   if (wrap && nDims == 2 && height > width + 1) {
-    std::vector<int64_t> wrap_sizes;
+    std::vector<SymInt> wrap_sizes;
 
-    int64_t step = width + 1;
-    int64_t wrap_size = ((self.numel() + step - 1) / step) - size;
+    auto step = width + 1;
+    auto wrap_size = ((self.numel() + step - 1) / step) - size;
     wrap_sizes.push_back(wrap_size);
```
Skylion007, line 131: Nit should be emplace_back or better yet just construct using a list initialize since it's a single element vector that's not referenced earlier, just construct the vector here with brace syntax.

aten/src/ATen/native/Fill.cpp
```diff
@@ -121,19 +120,19 @@ Tensor& fill_diagonal_(Tensor& self, const Scalar& fill_value, bool wrap) {
   strides.push_back(stride);
```
Skylion007, line 120: Same here, might as will construct as initializer list since its length 1

Skylion007 (approved):

aten/src/ATen/native/Fill.cpp
bobrenjc93: addressed both here https://github.com/pytorch/pytorch/pull/163607

PR #163486, opened 2025-09-22
Add type annotations to MPS profiler utilities

bobrenjc93:
## Summary
- drop the local mypy allow-untyped-defs escape hatch in the MPS profiler helpers
- annotate the context managers and bool helpers so they type-check cleanly

## Testing
- python -m mypy torch/mps/profiler.py --config-file mypy-strict.ini

------
https://chatgpt.com/codex/tasks/task_e_68d0ce4df2e483268d06673b65ef7745

### DIFF
--- a/torch/mps/profiler.py
+++ b/torch/mps/profiler.py
@@ -1,5 +1,5 @@
 import contextlib
-from typing import Iterator, cast
+from typing import Iterator, Literal
 
 import torch
 
@@ -14,7 +14,10 @@ __all__ = [
 ]
 
 
-def start(mode: str = "interval", wait_until_completed: bool = False) -> None:
+ProfilerMode = Literal["interval", "event", "interval,event"]
+
+
+def start(mode: ProfilerMode = "interval", wait_until_completed: bool = False) -> None:
     r"""Start OS Signpost tracing from MPS backend.
 
     The generated OS Signposts could be recorded and viewed in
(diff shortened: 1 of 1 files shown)


torch/mps/profiler.py
```diff
@@ -35,16 +35,18 @@ def start(mode: str = "interval", wait_until_completed: bool = False) -> None:
...
-def stop():
+def stop() -> None:
     r"""Stops generating OS Signpost tracing from MPS backend."""
-    torch._C._mps_profilerStopTrace()
+    torch._C._mps_profilerStopTrace()  # type: ignore[attr-defined]
 
 
 @contextlib.contextmanager
-def profile(mode: str = "interval", wait_until_completed: bool = False):
+def profile(mode: str = "interval", wait_until_completed: bool = False) -> Iterator[None]:
```
Skylion007: Perfect: mode should be replaced with a literal of valid mode down the line though

torch/mps/profiler.py
```diff
@@ -72,16 +74,16 @@ def is_metal_capture_enabled() -> bool:
     """Checks if `metal_capture` context manager is usable
     To enable metal capture, set MTL_CAPTURE_ENABLED envvar
     """
-    return torch._C._mps_isCaptureEnabled()  # type: ignore[attr-defined]
+    return cast(bool, torch._C._mps_isCaptureEnabled())  # type: ignore[attr-defined]
```
Skylion007: I'd probably just add it to the type ignore. We can remove it when we add proper type stubs here.

Skylion007: Actually hold on, it doesn't give an error on CI and actually complains about redundant cast...

Skylion007 (approved):

PR #163488, opened 2025-09-22
[pt2][cache] rework cache for true generic usage + better tests

nmacchioni:
Differential Revision: D82933509

over the weekend I realized that some of the cache implementation was a bit silly, and too constrained to be actually generic. for example, InMemoryCache[str, bytes] was odd since we'd probably want to be able to store more than just str keys with bytes values. so tldr; everything is now generic, with the one constraint being that Key and Value must both be pickle-able types. this makes things a lot simpler for us, since all caches can now be str -> bytes caches under the hood if we'd like, and Key/Value just get pickled on the way in and out.

with this change, there were also some improvements made to the testing; mainly better coverage, but now we also test each cache across every combination of Key/Value types to ensure that they will work with the types we might specify later

I also hardened some things here and there, for example we now use literal_eval (forgot who mentioned this on the first PR, but thank you for the suggestion!), and all errors coming from the caching will be wrapped in CacheError from now on (although we still raise from the original error context where possible)

putting this PR up now for feedback, in the process of generalizing the code I did remove the documentation since it was becoming outdated but I will add that back in after the PR is green

I have the next PR ready as well (implements a fresh cache context manager), will export once this lands


test/inductor/test_cache.py
```diff
@@ -0,0 +1,737 @@
...
+)
+
+
+class TestMixin:
+    @staticmethod
+    def abstract_cache_types() -> set[type[icache.Cache]]:
+        return {icache.Cache, icache.AsyncCache}
+
+    @staticmethod
+    def cache_types() -> list[type[icache.Cache]]:
```
aorenste: Sequence? or set? (here and the methods below)

test/inductor/test_cache.py
```diff
@@ -0,0 +1,737 @@
...
+    ) -> list[icache.Key]:
+        keys = []
+        while len(keys) < num:
+            if (key := self.key_not_in(cache, key_fn)) not in keys:
+                keys.append(key)
+        return keys
+
+    def key(self: Self, key_type: type[icache.Key]) -> icache.Key:
+        if key_type == str:
+            return f"{randint(0, 2**32)}"
```
aorenste: nit: consider prefixing with something like "k" so reading it as debugging output makes it a bit easier to see it's a string?

test/inductor/test_cache.py
```diff
@@ -0,0 +1,737 @@
...
+    ) -> list[icache.Value]:
+        values = []
+        while len(values) < num:
+            if (value := value_fn()) not in values:
+                values.append(value)
+        return values
+
+    def value(self: Self, value_type: type[icache.Value]) -> icache.Value:
+        if value_type == str:
+            return f"{randint(0, 2**32)}"
```
aorenste: nit: like above, consider prefixing with 'v' to make debugging easier...

torch/_inductor/cache.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+class CacheError(ValueError):
+    pass
+
+
+# TypeVars can't be recursive, so I'm not sure how to type
+# that Key tuples should be tuple[Key, ...] and Value
+# tuples/dicts/lists should follow the same pattern
+Key = TypeVar("Key", str, int, tuple[Any, ...])
```
aorenste: Yeah - looks like this needs the 3.12 syntax to do properly...

torch/_inductor/cache.py
```diff
@@ -0,0 +1,229 @@
...
+            kv_pair = kv_pair.strip()
+
+            if not kv_pair:
+                # kv_pair could be '' if env_val is '' or has ; suffix
+                continue
+
+            try:
+                key_bytes_repr, value_bytes_repr = kv_pair.split(",", 1)
+                # ignore whitespace prefix/suffix, again
+                key_bytes_repr, value_bytes_repr = (
```
aorenste: nit: maybe move the strip out of the try/except?

torch/_inductor/cache.py
```diff
@@ -0,0 +1,229 @@
...
+    def from_file_path(cls, fpath: Path) -> Self:
+        cache = cls()
+
+        if not fpath.is_file():
+            # fpath doesn't exit = empty cache
+            return cache
+
+        try:
+            with open(fpath, "rb") as fp:
+                cache._cache = pickle.load(fp)
```
aorenste, line 202: Should we add a simple header version byte? Just so we can update without worrying about breaking BC?

torch/_inductor/cache.py
```diff
@@ -0,0 +1,229 @@
...
+                if not isinstance(cache._cache, dict):
```
aorenste: nit: move out of the 'with open...' block...

torch/_inductor/cache.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+class OnDiskCache(AsyncCache[Key, Value]):
+    @cached_property
+    def base_dir(self: Self) -> Path:
+        return Path(gettempdir()) / "cache"
+
+    def _fpath_from_key(self: Self, key: Key) -> Path:
+        try:
+            return self.base_dir / sha256(pickle.dumps(key)).hexdigest()[:32]
```
aorenste, line 290: Maybe this should have an additional cache_name path? So instead of dumping the files in "/tmp/cache/hexdigest" they're put into "/tmp/cache/mybeautifulcache/hexdigest"?

torch/_inductor/cache.py
```diff
@@ -0,0 +1,229 @@
...
+        return FileLock(str(fpath.parent / "locks" / fpath.name[:4]) + ".lock")
+
+    @override
+    def get(self: Self, key: Key) -> Value | None:
+        fpath = self._fpath_from_key(key)
+        flock = self._flock_from_fpath(fpath)
+
+        with flock:
+            if not fpath.is_file():
+                return None
```
aorenste, line 338: nit: could just blindly do the open() and catch the FileNotFoundError...

aorenste (approved):
approving with some minor comments

torch/_inductor/cache.py
masnesral, line 290: FYI in codecache.py I think we do `base64.b32encode(hashlib.sha256(data).digest())[:51].decode("utf-8").lower()` for this operation in case that's somehow better. (I know because I refactored this code once, but I didn't dig into why we do it this way)

masnesral (approved):

torch/_inductor/cache.py
nmacchioni: Thanks for pointing that out; I think either method should be fine. codecache takes the raw binary hash, converts it to B32, trims to 51 and decodes (still outputs base32). This method here just skips some steps, going straight to hex. I think it is just a matter of choice here

torch/_inductor/cache.py
nmacchioni: Can you point me to that syntax? I'm interested to see how it is solved

test/inductor/test_cache.py
nmacchioni: Sorry, not sure the question here

torch/_inductor/cache.py
nmacchioni: Great idea, adding a check now

test/inductor/test_cache.py
aorenste: I should have marked as 'nit': Just suggesting that instead of returning a list you return a Sequence or set. In general it's better to return an immutable Sequence (unless you think the caller needs to be able to mutate it) - for example that way you could cache the result without changing the API

torch/_inductor/cache.py
nmacchioni: So what happens if we update the version of the cache, but the files still exist? get should obviously return None, since the values stored by the previous cache version can't be safely parsed, but should we also remove that file so that the key can be re-cached?

torch/_inductor/cache.py
aorenste: Something like this:
https://www.reddit.com/r/Python/comments/1jvaaw4/recursive_generic_type_hints_python_312/
I didn't look too deeply since we still have to support 3.10 for a while.

torch/_inductor/cache.py
aorenste: Hm. Yeah - good question. I guess the issue is that if you're running two different versions then you'll ping-pong between them.  
I guess another reasonable option is to fold the version into the hash of the filename instead...

torch/_inductor/cache.py
nmacchioni: Ah, I still think that only works for type aliases and not for type vars which we need since generic expects actual types and not aliases

torch/_inductor/cache.py
nmacchioni: I think in this case if there is overlap I'll just prefer to delete the old version cache, if the user wants to fiddle with the version then they can be ping ponged all they want haha

PR #163503, opened 2025-09-22
[dynamo] prevent recompilations on external_utils.wrap_inline

williamwen42:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.15.0) (oldest at bottom):
* #171826
* __->__ #163503
* #185524
* #186100
* #191389
* #195958
* #195494
* #192863


torch/_dynamo/external_utils.py
```diff
@@ -58,15 +60,32 @@ def is_compiling() -> bool:
...
     """
 
     @functools.wraps(fn)
     def inner(*args: _P.args, **kwargs: _P.kwargs) -> _R:
         return fn(*args, **kwargs)
 
+    if fn not in _wrap_inline_cache:
+        _wrap_inline_cache[fn] = deepcopy_code(inner.__code__)  # type: ignore[attr-defined]
+
+    inner.__code__ = _wrap_inline_cache[fn]  # type: ignore[attr-defined]
```
StrongerXi, line 89: Checking my understanding:

Given different functions `f1` and `f2`, your PR makes `wrap_inline(f1)` and `wrap_inline(f2)` have different code objects. Wouldn't that _trigger_ recompilation (as opposed to "don't recompile" as you documented)?

On the other hand, given the same `f`, existing code (without your PR) already have `inner.__code__` being the same for multiple `wrap_inline(f)` calls, so there'll be no recompilation anyways, e.g.,
```python
import torch

def f(x):
    return x + 1

f1 = torch._dynamo.external_utils.wrap_inline(f)
f2 = torch._dynamo.external_utils.wrap_inline(f)
f1 = torch.compile(f1, backend="eager")
f2 = torch.compile(f2, backend="eager")

f1(torch.rand(10))
f2(torch.rand(10))
```

Could you create a minimal example to show what exactly is causing unwanted recompilation over `wrap_inline(...)`?

StrongerXi: > Given different functions f1 and f2, your PR makes wrap_inline(f1) and wrap_inline(f2) have different code objects. Wouldn't that trigger recompilation (as opposed to "don't recompile" as you documented)?

I think I got it, by returning a new code object, you are changing recompilations into compilation over _different_ frames (i.e., code objects).

So this is mainly for compiling a `wrap_inline`-ed function at top level (or after graph breaks, however that might get triggered). But you still want to treat different instances of `wrap_inline(f)`s (with same input function object) as the same frame for compilation (basically my example), thus the cache.

Curious, why are we hitting this case now?

StrongerXi (approved):
I'd like to understand why we are hitting this problem for `wrap_inline`, etc. now. Because in theory this is a dynamo problem for all non-inplace function wrappers.

torch/_dynamo/external_utils.py
williamwen42: Minimal example is in the added unit test.

I'm hitting this case now because my wrapped nested graph break tests add an additional frame to each top-level compile call through wrap_inline - but then this causes recompilation limits to be exceeded. I then figured that this problem could happen more generally with wrap_inline - if we compile enough top level functions that are wrap_inline'd, then we would suddenly start hitting the recompile limit.

The current problem with this PR is that we create a new wrapper with a new code object for each _instance_ of the function, which is problematic for two instances of the function with the same code object. This PR would then compile the inner function twice.

StrongerXi: > I'm hitting this case now because my wrapped nested graph break tests add an additional frame to each top-level compile call through wrap_inline

Makes sense, I assume this is only for the tests, i.e., not fundamental to nested graph break use cases in the wild? If so then I think it's totally okay to lazy evaluate the "different func objects but same code objects" problem.

williamwen42: Yeah, this is not fundamental to nested graph breaks.

mlazos (approved):

Issue #163346, opened 2025-09-19
Stable headers could provide helpers for automagic boxing of functions


lw:
### 🚀 The feature, motivation and pitch

Registering an operator in the stable API requires passing a boxed function, whereas in the regular API we must pass an unboxed one.

Writing a boxing wrapper is easy but tedious, and the result is verbose and prone to going stale when the underlying function changes.

However, with some C++ template metaprogramming it should be possible to automatically generate such boxing wrappers! It would be amazing if the stable API provided such a functionality out-of-the-box.

I made an attempt at writing such a wrapper, leveraging many of the components that are already included in PyTorch:
```cpp
#include <c10/util/Metaprogramming.h>

namespace {

template<class... T, std::size_t... I>
std::tuple<T...> unbox_to_tuple_impl(StableIValue* stack, std::index_sequence<I...>)
{
    return std::make_tuple(to<T>(stack[I])...);
}

template<class... T>
std::tuple<T...> unbox_to_tuple(StableIValue* stack)
{
    return unbox_to_tuple_impl<T...>(stack, std::make_index_sequence<sizeof...(T)>());
}

template<class... T, std::size_t... I>
void box_from_tuple_impl(StableIValue* stack, std::tuple<T...> vals, std::index_sequence<I...>)
{
    (stack[I] = from<T>(std::get<I>(vals)), ...);
}

template<class... T>
void box_from_tuple(StableIValue* stack, std::tuple<T...> vals)
{
    box_from_tuple_impl<T...>(stack, vals, std::make_index_sequence<sizeof...(T)>());
}

template <typename ReturnType, typename ParameterTypeList, typename FuncT, FuncT* func>
struct boxer_impl {};

template <typename... ReturnTypes, typename... ParameterTypes, typename FuncT, FuncT* func>
struct boxer_impl<std::tuple<ReturnTypes...>, c10::guts::typelist::typelist<ParameterTypes...>, FuncT, func> {
  void operator()(
    StableIValue* stack,
    uint64_t num_args,
    uint64_t num_outputs
  ) {
    assert(num_args == sizeof...(ParameterTypes));
    assert(num_outputs == sizeof...(ReturnTypes));
    std::tuple<ParameterTypes...> args = unbox_to_tuple<ParameterTypes...>(stack);
    auto res = std::apply(func, args);
    box_from_tuple<ReturnTypes...>(stack, res);
  }
};

[...]

PR #163505, opened 2025-09-22
Provide boxing helper to stable API

lw:
* __->__ #163505
* #163635
* #163634

Fixes https://github.com/pytorch/pytorch/issues/163346

cc @malfet @zou3519 @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @albanD


test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
```diff
@@ -75,44 +75,28 @@ Tensor sgd_out_of_place(
   return out;
 }
 
-void boxed_sgd_out_of_place(StableIValue* stack, uint64_t num_args, uint64_t num_outputs) {
-  Tensor res = sgd_out_of_place(
-    to<Tensor>(stack[0]),
-    to<Tensor>(stack[1]),
-    float(to<double>(stack[2])),
```
lw, line 82: Any reason this was first converted to double? Does `to<float>()` not work?

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
```diff
@@ -528,26 +381,10 @@ int64_t test_stream(int32_t device_index) {
   return torch::stable::accelerator::getCurrentStream(device_index).id();
 }
 
-void boxed_test_stream(
-    StableIValue* stack,
-    uint64_t num_args,
-    uint64_t num_outputs) {
-  int64_t res = test_stream(static_cast<int64_t>(to<int64_t>(stack[0])));
```
lw, line 547: Why the extra static_cast after the `to<int64_t>(...)`?

torch/csrc/stable/library.h
```diff
@@ -168,3 +170,77 @@ class StableTorchLibraryInit final {
...
+    assert(num_args == sizeof...(ParameterTypes));
+    assert(num_outputs == sizeof...(ReturnTypes));
+    std::tuple<ParameterTypes...> args = unbox_to_tuple<ParameterTypes...>(stack);
+    auto res = std::apply(func, args);
+    box_from_tuple<ReturnTypes...>(stack, res);
+  }
+};
+
+template <typename ReturnType, typename... ParameterTypes, typename FuncT, FuncT* func>
+struct boxer_impl<ReturnType, c10::guts::typelist::typelist<ParameterTypes...>, FuncT, func> {
```
lw, line 221: Does any C++ expert know whether this needs an `enable_if` (or whatever) to exclude it from the candidates when ReturnType is a tuple?

torch/csrc/stable/library.h
```diff
@@ -168,3 +170,110 @@ class StableTorchLibraryInit final {
...
+struct boxer_impl<
+    std::tuple<ReturnTypes...>,
+    c10::guts::typelist::typelist<ParameterTypes...>,
+    FuncT,
+    func> {
+  void operator()(
+      StableIValue* stack,
+      uint64_t num_args,
+      uint64_t num_outputs) {
+    assert(num_args == sizeof...(ParameterTypes));
```
lw: TODO: this should become TORCH_INTERNAL_ASSERT (or whatever its header-only equivalent is)

torch/csrc/stable/library.h
```diff
@@ -168,3 +170,110 @@ class StableTorchLibraryInit final {
...
+    assert(num_outputs == 1);
+    std::tuple<ParameterTypes...> args =
+        unbox_to_tuple<ParameterTypes...>(stack);
+    auto res = std::apply(func, args);
+    box_from_tuple<std::tuple<ReturnType>>(stack, std::make_tuple(res));
```
lw: TODO: simplify it to `stack[0] = from<ReturnType>(res);`

torch/csrc/stable/library.h
```diff
@@ -168,3 +170,110 @@ class StableTorchLibraryInit final {
...
+    auto res = std::apply(func, args);
+    box_from_tuple<ReturnTypes...>(stack, res);
+  }
+};
+
+template <
+    typename ReturnType,
+    typename... ParameterTypes,
+    typename FuncT,
+    FuncT* func>
```
lw, line 249: Does any C++ expert know whether this needs an enable_if (or whatever) to exclude it from the candidates when ReturnType is a tuple?

janeyx99 (commented):
To make this easy for reviewing + properly chunk up changes, can you stack the following PRs:
1. a PR moving Metaprogramming.h + TypeList.h to torch/headeronly.
2. a codemod PR changing <c10/util/TypeList.h> -> <torch/headeronly/util/TypeList.h> (and similar one for Metaprogramming)
3. the final PR with the boxed impl + tests

I'm thinking 1+2 should be easily landable (not break tests). Let's also add tests for the APIs we rely on being headeronly into `test/cpp/aoti_abi_check`

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
lw: The test failed without this, but I believe it's more foolproof if the special-handling of floats is done by `to<...>` directly rather than by each user separately, hence I added a new specialization for `to<...>`.

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
lw: The tests seem to pass without it...

torch/csrc/stable/library.h
```diff
@@ -168,3 +170,126 @@ class StableTorchLibraryInit final {
...
+    std::tuple<ParameterTypes...> args =
+        unbox_to_tuple<ParameterTypes...>(stack);
+    auto res = std::apply(func, args);
+    stack[0] = from<ReturnType>(res);
+  }
+};
+
+template <typename FuncT, FuncT* func>
+struct boxer {
+  using FunctionTraits = c10::guts::infer_function_traits_t<FuncT>;
```
janeyx99, line 276:
```suggestion
  using FunctionTraits = torch::headeronly::guts::infer_function_traits_t<FuncT>;
```

torch/csrc/stable/library.h
```diff
@@ -168,3 +170,126 @@ class StableTorchLibraryInit final {
...
+};
+
+template <
+    typename ReturnType,
+    typename... ParameterTypes,
+    typename FuncT,
+    FuncT* func>
+struct boxer_impl<
+    ReturnType,
+    c10::guts::typelist::typelist<ParameterTypes...>,
```
janeyx99, line 252: and here

torch/csrc/stable/library.h
```diff
@@ -168,3 +170,126 @@ class StableTorchLibraryInit final {
...
+};
+
+template <
+    typename... ReturnTypes,
+    typename... ParameterTypes,
+    typename FuncT,
+    FuncT* func>
+struct boxer_impl<
+    std::tuple<ReturnTypes...>,
+    c10::guts::typelist::typelist<ParameterTypes...>,
```
janeyx99, line 219: and here

torch/csrc/stable/library.h
```diff
@@ -168,3 +170,126 @@ class StableTorchLibraryInit final {
...
+      stack, vals, std::make_index_sequence<sizeof...(T)>());
+}
+
+template <
+    typename ReturnType,
+    typename ParameterTypeList,
+    typename FuncT,
+    FuncT* func>
+struct boxer_impl {
+  static_assert(c10::guts::false_t<ReturnType>::value, "Unsupported");
```
janeyx99, line 209: and here

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
janeyx99: Likely unnecessary indeed

torch/csrc/stable/library.h
lw: Are you sure about this? In https://github.com/pytorch/pytorch/pull/163634, when I moved the header files, I didn't change the namespace of the symbols, hence these still live in `c10::`. Do you want me to change that?

torch/csrc/stable/stableivalue_conversions.h
```diff
@@ -314,6 +314,14 @@ struct ToImpl<torch::stable::Tensor> {
...
 };
 
+// Specialization for StableIValue => float
+template <>
+struct ToImpl<float> {
+  static float call(StableIValue val) {
+    return float(to<double>(val));
+  }
+};
+
```
janeyx99, line 324: Ah yea,  it's cuz the `float` type in the schema actually corresponds to `double` in libtorch. (So an IValue that was a float will be converted to a double).

I'm trying to work out all the implications of this specialization, and I'm figuring that it assumes that:
1. the code never ever calls a float -> StableIValue conversion (i.e., that there are only double -> StableIValue conversions) OR
2. that calling `float(to<double>(a_raw_memory_copied_float_into_uint64)` is equivalent behavior to `float(a_raw_memory_copied_float_into_uint64)`. 

I'm not convinced either of these assumptions are always true (I think 1 may be true in practice for now)...so I don't think we should add this specialization.

What _is_ true is that there should _never_ be any floats in the user code (we should just change the signature of the sgd kernel to take in a double + downcast in the kernel itself--my bad from earlier). And the boxing kernel should just call to<double> when it sees "float" in the schema.

torch/csrc/stable/library.h
janeyx99: Oop I missed this 2 days ago sorry! But yes, we want to maintain the invariant that everything in torch/csrc/stable uses either other stable things or header only APIs. And then we are maintaining that only APIs in `torch::headeronly` are actualy header-only and tested to be header only.

So my request with this change is to:
1. expose the APIs in torch::headeronly namespace
2. add tests for them (or migrate existing)
3. only use torch::headeronly or torch::stable or a C shim API in these files

janeyx99:
Closing this in favor of https://github.com/pytorch/pytorch/pull/167582. thanks @lw for the foundational work!

PR #163507, opened 2025-09-22
[BUG] MaxUnpool2d/3d should check output dim before accessing its elements

can-gaa-hou:
Fixes #163409

### DIFF
--- a/aten/src/ATen/native/cuda/MaxUnpooling.cu
+++ b/aten/src/ATen/native/cuda/MaxUnpooling.cu
@@ -125,8 +125,6 @@ Tensor& max_unpooling2d_forward_out_cuda(const Tensor& self_,
   TORCH_CHECK(
       indices_.scalar_type() == at::ScalarType::Long,
       "elements in indices should be type int64 but got: ", indices_.scalar_type());
-  auto oheight = output_size[0];
-  auto owidth = output_size[1];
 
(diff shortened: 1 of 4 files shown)


aten/src/ATen/native/cuda/MaxUnpooling.cu
```diff
@@ -505,13 +508,14 @@ at::Tensor& max_unpooling3d_backward_out_cuda(const Tensor& grad_output_,
...
-  int64_t oT = output_size[0];
-  int64_t oH = output_size[1];
-  int64_t oW = output_size[2];
 
   max_unpooling3d_shape_check(
     self_, grad_output_, indices_, output_size, stride, padding, "max_unpooling3d_backward_out_cuda()");
 
+  int64_t oT = output_size[0];
+  int64_t oH = output_size[1];
+  int64_t oW = output_size[2];
```
eqy, line 517: What is the purpose of moving the declarations around?

malfet (approved):

aten/src/ATen/native/cuda/MaxUnpooling.cu
malfet: Don't dereference 2nd index of output_size if output_size.length() is 0 or something (which I guess is validated by next TORCH_CHECK)

Skylion007 (approved):

PR #163525, opened 2025-09-22
ci: Add a way to lint all files in a PR from label

seemethere:
* #163526
* __->__ #163525

Works by either applying `lint-all-files` and should automatically apply when `Reverted` appears on PRs as well

### DIFF
--- a/.github/workflows/_get-changed-files.yml
+++ b/.github/workflows/_get-changed-files.yml
@@ -26,11 +26,12 @@ jobs:
             # Get the PR number from the github context
             PR_NUMBER="${{ github.event.number }}"
 
-            # Check if the PR has the lint-all-files label
-            HAS_LINT_ALL_LABEL=$(gh api repos/${{ github.repository }}/pulls/$PR_NUMBER --jq '.labels[] | select(.name == "lint-all-files") | .name')
+            # Check if the PR has the lint-all-files or Reverted label
+            # Reverted PRs probably have a high likelihood of needing to have all files linted, so we should lint all files
+            HAS_LINT_ALL_LABEL=$(gh api repos/${{ github.repository }}/pulls/$PR_NUMBER --jq '.labels[] | select(.name == "lint-all-files" or .name == "Reverted") | .name')
 
             if [ -n "$HAS_LINT_ALL_LABEL" ]; then
-              echo "Found 'lint-all-files' label, linting all files"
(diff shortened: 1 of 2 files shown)


.github/workflows/_get-changed-files.yml
```diff
@@ -26,17 +26,26 @@ jobs:
...
 
-            # Use gh CLI to get changed files in the PR with explicit repo
-            CHANGED_FILES=$(gh api repos/${{ github.repository }}/pulls/$PR_NUMBER/files --paginate --jq '.[] | select(.status != "removed") | .filename' | tr '\n' ' ' | sed 's/ $//')
-
-            if [ -z "$CHANGED_FILES" ]; then
-              echo "No changed files found, setting to '*'"
-              CHANGED_FILES="*"
+            # Check if the PR has the lint-all-files or Reverted label
+            # Reverted PRs probably have a high likelihood of needing to have all files linted, so we should lint all files
+            HAS_LINT_ALL_LABEL=$(gh api repos/${{ github.repository }}/pulls/$PR_NUMBER --jq '.labels[] | select(.name == "lint-all-files" or .name == "Reverted") | .name')
```
ZainRizvi: With these changes we're coupling this workflow very tightly to the linter jobs.  That's fine on it's own, but consider renaming this workflow to reflect that it shouldn't be used for anything other than lint

Or a better fix could be to pass in a list of "bypass_labels", and if any of those are on the PR then you return all files. The lint workflow can then pass in "revert" and "lint-all-files" as the bypass labels

ZainRizvi (approved):

.github/workflows/_get-changed-files.yml
seemethere: This is a good suggestion, I refactored to do this

.github/workflows/lint.yml
```diff
@@ -31,6 +31,8 @@ jobs:
     if: github.repository_owner == 'pytorch'
     name: Get changed files
     uses: ./.github/workflows/_get-changed-files.yml
+    with:
+      all_files: ${{ contains(github.event.pull_request.labels.*.name, 'lint-all-files') || contains(github.event.pull_request.labels.*.name, 'Reverted') }}
```
malfet, line 35: Would be nice to also respect `ci-no-td` label

PR #163527, opened 2025-09-22
Add magic TORCH_MAKE_PYBIND_ENUM_FASTER macro

swolchok:
* __->__ #163527
* #164974

See comment on the macro definition. In short, pybind11 3.x
added `py::native_enum`, and also had to add overhead for that new way
to bind enums on the critical path for calling functions that take
regular old `py::enum_`s as arguments (for example, `__eq__`).

Differential Revision: [D82873169](https://our.internmc.facebook.com/intern/diff/D82873169/)

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @EikanWang @jgong5 @wenzhe-nrv @sanchitintel @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @kadeng @chauhang @amjames @Lucaskabela @jataylo @ezyang @chenyang78


swolchok:
> Add a link to the pybind11 issue so we can remove this when it's resolved over there

I haven't filed an issue and I don't see a particular reason to think it's structurally fixable. type_casters are keyed on the type they are casting, not on whether the pybind11 user registered the type with `py::enum_` vs `py::native_enum`. Their clear recommendation is for new code to use `py::native_enum`.

Skylion007:
I maintain pybind11 so happy to discuss if there is a better way to optimize the casters or refactor it better in the long term

ezyang (approved):
Approving, but this will be annoyingly difficult to get people to remember to do lol

PR #163534, opened 2025-09-22
Rename to _debug_mode.py to make it private

SherlockNoMad:
rename debug_mode.py to _debug_mode.py to make it private, per @alban's request. 


albanD (approved):
Thanks

PR #163559, opened 2025-09-22
[MTIA] Enable deserialization for FP8 checkpoint loading

PatriceVignola:
Summary: It looks like loading FP8 checkpoints goes through that path which wasn't enabled for MTIA beforehand, whereas loading BF16 checkpoints didn't.

Differential Revision: D82997140

### DIFF
--- a/torch/testing/_internal/common_utils.py
+++ b/torch/testing/_internal/common_utils.py
@@ -1463,6 +1463,10 @@ TEST_WITH_ROCM: bool = TestEnvironment.def_flag(
     "TEST_WITH_ROCM",
     env_var="PYTORCH_TEST_WITH_ROCM",
 )
+TEST_WITH_MTIA: bool = TestEnvironment.def_flag(
+    "TEST_WITH_MTIA",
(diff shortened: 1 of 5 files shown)


mikaylagawarecki (commented):
Are there tests that exercise this change?

torch/testing/_internal/common_utils.py
```diff
@@ -1463,6 +1463,10 @@ def split_if_not_empty(x: str):
     "TEST_WITH_ROCM",
     env_var="PYTORCH_TEST_WITH_ROCM",
 )
+TEST_WITH_MTIA: bool = TestEnvironment.def_flag(
```
mikaylagawarecki, line 1466: n00b qn on my end: is this PR adding the only / first MTIA test within pytorch?

mikaylagawarecki (approved):

torch/testing/_internal/common_utils.py
PatriceVignola: We have some in pytorch already I believe, but not a lot (and they may not need to use this variable). Most of our pytorch tests are not in OSS.

PR #163547, opened 2025-09-22
[torchfuzz] Encapsulate fuzzing and codegen logic into ops

bobrenjc93:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #163560
* #163558
* #163557
* #163556
* #163555
* #163554
* #163553
* __->__ #163547

PR #163560, opened 2025-09-22
[torchfuzz] introduce multi process fuzzer

bobrenjc93:
* __->__ #163560
* #163558
* #163557
* #163556
* #163555
* #163554
* #163553
* #163547

### DIFF
--- a/tools/experimental/dynamic_shapes/torchfuzz/fuzzer.py
+++ b/tools/experimental/dynamic_shapes/torchfuzz/fuzzer.py
@@ -1,5 +1,6 @@
 # mypy: ignore-errors
 import logging
+import multiprocessing as mp
 import os
 import random
 import sys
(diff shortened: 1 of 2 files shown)


tools/experimental/dynamic_shapes/torchfuzz/fuzzer.py
```diff
@@ -177,63 +178,56 @@ def log(success: bool) -> None:
...
-                continue
-
-            print(f"Test failed with error: {error_message}")
-            return
-
-
 if __name__ == "__main__":
     import argparse
 
+    try:
```
laithsakka, line 184: why in main not globally.

tools/experimental/dynamic_shapes/torchfuzz/fuzzer.py
```diff
@@ -177,63 +178,56 @@ def log(success: bool) -> None:
...
+    parser.add_argument(
+        "--max-depth", type=int, help="Maximum depth for operation stack (1-20)"
+    )
+
+    # Multi-process fuzzing arguments
     parser.add_argument(
-        "--seed", type=int, help="Random seed for reproducible generation"
+        "--start", type=int, help="Starting seed value for multi-process fuzzing"
     )
     parser.add_argument(
```
laithsakka, line 210: maybe call it steps or itterations?

tools/experimental/dynamic_shapes/torchfuzz/fuzzer.py
```diff
@@ -250,15 +244,59 @@ def known_issue(error_message: str) -> bool:
...
+        # Single seed execution mode
         print("Running single fuzz_and_execute...")
         seed, success, error_message = fuzz_and_execute(
             seed=args.seed, max_depth=args.max_depth
         )
         print(f"Result: seed={seed}, success={success}")
         if not success:
             print(f"Error: {error_message}")
+            sys.exit(1)
+    elif args.start is not None or args.count is not None:
```
laithsakka, line 258: what does start have to be not None? why not just always use seed instead ? or remove seed and just keep start either, also when its none shall we not just assign it randomly and do .

tools/experimental/dynamic_shapes/torchfuzz/multi_process_fuzzer.py
```diff
@@ -0,0 +1,379 @@
...
+    except BrokenPipeError:
+        import os
+
+        os.makedirs("/tmp/torchfuzz", exist_ok=True)
+        with open("/tmp/torchfuzz/crash.log", "a") as f:
+            f.write(f"BrokenPipeError: {msg}\n")
+
+
+# List of regex patterns for ignore bucket
+IGNORE_PATTERNS: list[re.Pattern] = [
```
laithsakka, line 38: why are we ignoring
is not define
was not declared in this scope
are those bugs in the fuzzer or bugs found in the fuzzer

also its nice to keep those ad dict so we track know there is a gh issue associated with each

tools/experimental/dynamic_shapes/torchfuzz/multi_process_fuzzer.py
```diff
@@ -0,0 +1,379 @@
...
+    re.compile(
+        r"Dynamo failed to run FX node with fake tensors: call_method fill_diagonal_"
+    ),
+    re.compile(r"zuf\d+ is not defined"),
+    re.compile(r"'zuf\d+' was not declared in this scope"),
+    re.compile(r"out_eager_sum: 0\.0"),
+    re.compile(r"error: operand #\d+ does not dominate this use"),
```
laithsakka: this is compile error ? seems like some of those are eager errors?
no?

tools/experimental/dynamic_shapes/torchfuzz/multi_process_fuzzer.py
```diff
@@ -0,0 +1,379 @@
...
+
+    Returns:
+        Tuple of (seed, success, output, duration, ignored_pattern_idx)
+        ignored_pattern_idx: -1 if not ignored, otherwise index of IGNORE_PATTERNS
+    """
+    start_time = time.time()
+
+    try:
+        # Run fuzzer.py with the specified seed
+        cmd = [sys.executable, "fuzzer.py", "--seed", str(seed)]
```
laithsakka: ahh thats we pass start ok ok i see, but when you pass seed here we already deleted fuzz_and_test path what would be called?

laithsakka: ok i stop the review here seems not complete?

laithsakka: ok ok we chatted pass --single here?

tools/experimental/dynamic_shapes/torchfuzz/multi_process_fuzzer.py
```diff
@@ -0,0 +1,387 @@
...
+        # Use multiprocessing Pool to distribute work
+        with mp.Pool(processes=num_processes) as pool:
+            # Submit all seeds to the process pool
+            future_results = []
+            for seed in seeds:
+                future = pool.apply_async(run_fuzzer_with_seed, (seed,))
+                future_results.append(future)
+
+            # Set up progress bar
+            if HAS_TQDM:
```
laithsakka, line 168: accepting to unblock but this code is complicated probably could use simplification in the future

laithsakka (approved):

tools/experimental/dynamic_shapes/torchfuzz/multi_process_fuzzer.py
```diff
@@ -0,0 +1,387 @@
...
+                                f"❌ FAILURE - Seed {seed} (duration: {duration:.2f}s):"
+                            )
+                            for line in output.split("\n"):
+                                if line.strip():
+                                    persist_print(f"   {line}")
+                            persist_print("")
+                    elif not success and ignored_pattern_idx != -1:
+                        # Optionally, print ignored failures if desired
+                        if verbose:
+                            if HAS_TQDM and pbar:
```
laithsakka, line 240: for example you can create a function for the 


```
 if HAS_TQDM and pbar:
else
```

and pass to it lambdas/strings? 

also there is so much repetition 
```
 for line in output.split("\n"):
   if line.strip():
   pbar.write(f"   {line}")
                                    
```
put it in function?

tools/experimental/dynamic_shapes/torchfuzz/multi_process_fuzzer.py
```diff
@@ -0,0 +1,387 @@
...
+        sys.exit(130)
+
+    total_time = time.time() - start_time
+
+    # Print summary
+    persist_print("=" * 60)
+    persist_print("📈 SUMMARY")
+    persist_print("=" * 60)
+
+    successful = [r for r in results if r[1]]
```
laithsakka, line 351: what is r?
shall we use better name for it?
fields of r ca be named for clarity hard to understand what are we looking at here

PR #138395, opened 2024-10-19
Make OpaqueUnaryFn pickleable

ezyang:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #138395

Fixes https://github.com/pytorch/pytorch/issues/138070

Signed-off-by: Edward Z. Yang <ezyang@meta.com>

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10

PR #163571, opened 2025-09-22
fix pickling for BitwiseFn

dolpm:
Summary:
ran into AttributeError: Can't get local object 'make_opaque_bitwise_fn.<locals>.BitwiseFn'

looks like it was fixed for UnaryFn but not BitwiseFn in https://github.com/pytorch/pytorch/pull/138395

Fixes #147841

### DIFF
--- a/test/inductor/test_compile_subprocess.py
+++ b/test/inductor/test_compile_subprocess.py
@@ -206,7 +206,8 @@ class TestSubprocess(TestCase):
 
             start = time.time()
             last_report = start
-            while _AsyncFxCompile._stat_compiled_runs < 4:
+            while True:
+                start_stat_compiled_runs = _AsyncFxCompile._stat_compiled_runs
                 # Sleep a bit so we don't drive the CPU unnecessarily.
                 time.sleep(0.25)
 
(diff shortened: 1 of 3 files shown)


jamesjwu (approved):
Good find, thanks!

dolpm:
these tests seem to be broken on trunk (e.g., PYTORCH_TEST_WITH_ASAN=1 PYTORCH_TEST_WITH_UBSAN=1 python test/inductor/test_compile_subprocess.py CpuTests.test_async)

will wait to land this.

test/inductor/test_compile_subprocess.py
```diff
@@ -219,6 +220,9 @@ def model_add(x, y):
                 # Backward pass
                 output.sum().backward()
 
+                if _AsyncFxCompile._stat_compiled_runs - start_stat_compiled_runs == 2:
```
malfet, line 223: Why not break if it's greater than 2?

test/inductor/test_compile_subprocess.py
```diff
@@ -206,7 +206,8 @@ def model_add(x, y):
 
             start = time.time()
             last_report = start
-            while _AsyncFxCompile._stat_compiled_runs < 4:
+            while True:
```
malfet, line 209: This looks very unsafe

test/inductor/test_compile_subprocess.py
dolpm: because it's possible the second compile run hasn't finished (and the stride will be 1). once both finish, each run will increment this counter by 2, so we want to wait for that.

it's possible you have something like...

_stat_compiled_runs:
0
0
...
0
0
0
0
...
1
2
3
4
5
...
7
9
11

there is no deterministic way to figure out when the second async compile will complete so we can't just choose a value.

Issue #163539, opened 2025-09-22
conv3d bfloat16 wrong result


ZejiaZheng:
### 🐛 Describe the bug

conv3d is producing wrong results in large input shapes after the cudnn upgrade (~https://github.com/pytorch/pytorch/pull/155122/commits~) // Update: issue is from cudnn 9.10.2
To repro
```
import torch
import torch.nn as nn
import torch.nn.functional as F


def test_conv2d_bf16_vs_fp32():
    """Test case for large spatial dimension conv2d with bf16 vs fp32"""

    # Check for CUDA availability
    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")
    print(f"Using device: {device}")

    # Set up the problematic configuration
    batch_size = 1
    in_channels = 16
    out_channels = 16
    in_height, in_width = 1282, 722
    out_height, out_width = 1280, 720
    kernel_size = (1, 3, 3)
    stride = (1, 1, 1)
    dilation = (1, 1, 1)

    # Note: This is a 3D convolution based on kernel/stride/dilation having 3 values
    # Adjusting dimensions accordingly
    in_depth = 124
    out_depth = 124

    # Create random input on CUDA
    torch.manual_seed(42)
    x_fp32 = torch.randn(batch_size, in_channels, in_depth, in_height, in_width,
                        dtype=torch.float32, device=device)
    x_bf16 = x_fp32.to(torch.bfloat16)

    # Create conv layer with random weights
    conv = nn.Conv3d(
        in_channels=in_channels,
        out_channels=out_channels,
        kernel_size=kernel_size,
        stride=stride,
        padding=0,  # No padding to match input/output sizes
        dilation=dilation,
        bias=False
    ).to(device)

    # Initialize weights
    nn.init.kaiming_normal_(conv.weight)

    # Forward pass in fp32
    conv_fp32 = conv.to(torch.float32)
    with torch.no_grad():
        y_fp32 = conv_fp32(x_fp32)

    # Forward pass in bf16
    conv_bf16 = conv.to(torch.bfloat16)
    with torch.no_grad():
        y_bf16 = conv_bf16(x_bf16)

    # Convert bf16 output to fp32 for comparison
    y_bf16_as_fp32 = y_bf16.to(torch.float32)

    # Compute differences
    diff = torch.abs(y_fp32 - y_bf16_as_fp32)
    mean_abs_diff = diff.mean().item()
    l2_diff = torch.sqrt((diff ** 2).sum()).item()
[...]

PR #163581, opened 2025-09-22
[cuDNN][Convolution] Disable cuDNN for 3D convolutions with kernel size != 1 for cuDNN 9.8+

eqy:
To workaround #163539

Still confirming whether 9.10 is affected. The original test states that the convolution is "large," but note that the input size does not apepar to require 64-bit indexing.

cc @csarofeen @ptrblck @xwang233 @msaroufim @jerryzh168 @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @voznesenskym @penguinwu @EikanWang @Guobing-Chen @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben

### DIFF
--- a/aten/src/ATen/native/Convolution.cpp
+++ b/aten/src/ATen/native/Convolution.cpp
@@ -421,6 +421,16 @@ struct ConvParams {
                         " Consider upgrading cuDNN and/or enabling the V8 API for better efficiency.");
         return false;
       }
+      // broken on cuDNN 9.8 - 9.10
+      if (cudnn_version >= 90800 && cudnn_version < 91300) {
+        if (input.scalar_type() == at::kBFloat16 || input.scalar_type() == at::kHalf) {
+          for (auto it = weight.sizes().begin(); it != weight.sizes().end(); it++) {
+            if (*it != 1) {
+              return false;
+            }
+          }
+        }
+      }
     }
     if (!input.is_cuda() || !cudnn_enabled) {
       return false;
--- a/aten/src/ATen/native/Convolution.cpp
+++ b/aten/src/ATen/native/Convolution.cpp
@@ -424,8 +424,8 @@ struct ConvParams {
       // broken on cuDNN 9.8
       if (cudnn_version >= 90800) {
         if (input.scalar_type() == at::kBFloat16 || input.scalar_type() == at::kHalf) {
-          for (auto it = weight.sizes().begin(); it != weight.sizes().end(); it++) {
-            if (*it != 1) {
+          for (auto val : weight.sizes()) {
+            if (val != 1) {
               return false;
             }
           }
--- a/aten/src/ATen/native/Convolution.cpp
+++ b/aten/src/ATen/native/Convolution.cpp
@@ -413,23 +413,23 @@ struct ConvParams {
     if (!detail::getCUDAHooks().compiledWithCuDNN()) {
       return false;
     }
+    static long cudnn_version = detail::getCUDAHooks().versionCuDNN();
+    // broken on cuDNN 9.8
+    if (cudnn_version >= 90800) {
+      if (input.scalar_type() == at::kBFloat16 || input.scalar_type() == at::kHalf) {
+        for (auto val : weight.sizes()) {
+          if (val != 1) {
+            return false;
+          }
+        }
+      }
+    }
     if (needs_64bit_indexing_no_split(input, weight)) {
-      static long cudnn_version = detail::getCUDAHooks().versionCuDNN();
       if (!(cudnn_version >= 90300 && at::native::cudnnv8_enabled_check_debug())) {
         TORCH_WARN_ONCE("cuDNN cannot be used for large non-batch-splittable convolutions"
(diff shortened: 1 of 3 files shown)


aten/src/ATen/native/Convolution.cpp
```diff
@@ -421,6 +421,16 @@ struct ConvParams {
                         " Consider upgrading cuDNN and/or enabling the V8 API for better efficiency.");
         return false;
       }
+      // broken on cuDNN 9.8
+      if (cudnn_version >= 90800) {
+        if (input.scalar_type() == at::kBFloat16 || input.scalar_type() == at::kHalf) {
+          for (auto it = weight.sizes().begin(); it != weight.sizes().end(); it++) {
+            if (*it != 1) {
```
malfet: We are using c++17
```suggestion
          for (auto val : weight.sizes()) {
            if (val != 1) {
```

aten/src/ATen/native/Convolution.cpp
```diff
@@ -421,6 +421,16 @@ struct ConvParams {
                         " Consider upgrading cuDNN and/or enabling the V8 API for better efficiency.");
         return false;
       }
+      // broken on cuDNN 9.8
+      if (cudnn_version >= 90800) {
+        if (input.scalar_type() == at::kBFloat16 || input.scalar_type() == at::kHalf) {
```
malfet: Is it broken just for inputs or for weights as well? Also, don't we have a `reduced_precision_float` bredicate?

aten/src/ATen/native/Convolution.cpp
```diff
@@ -421,6 +421,16 @@ struct ConvParams {
...
+          for (auto it = weight.sizes().begin(); it != weight.sizes().end(); it++) {
+            if (*it != 1) {
+              return false;
+            }
+          }
+        }
+      }
     }
     if (!input.is_cuda() || !cudnn_enabled) {
```
malfet, line 437: Q: Shouldn't this check be at the very beginning of the function

eqy: Yes, trying not to create a merge conflict for myself: https://github.com/pytorch/pytorch/pull/163171

aten/src/ATen/native/Convolution.cpp
eqy: I do not think we have convolutions that are mixing input and weight types as IIRC this fails checks

Are you referring to reduced precision reductions? Those are for cuBLAS matmuls only and control the use of split-k.

Or if you're referring to automatic mixed-precision/AMP that's doing casts of weight and input before the computation.

ngimel (approved):
Approving to unblock, but if the sizes in the issue don't require 64-bit indexing, how will that help?

aten/src/ATen/native/Convolution.cpp
```diff
@@ -413,8 +413,18 @@ struct ConvParams {
     if (!detail::getCUDAHooks().compiledWithCuDNN()) {
       return false;
     }
+    static long cudnn_version = detail::getCUDAHooks().versionCuDNN();
+    // broken on cuDNN 9.8
+    if (cudnn_version >= 90800) {
+      if (input.scalar_type() == at::kBFloat16 || input.scalar_type() == at::kHalf) {
+        for (auto val : weight.sizes()) {
+          if (val != 1) {
```
ngimel: this is also checking channels of weight, not just filter? Also, this disables all cudnn convolutions, not just 3d?

ngimel (commented):
Ouch, so this should disable cudnn for all of 3d cases?

aten/src/ATen/native/Convolution.cpp
eqy: That's a good point, in theory it would only need to be disabled for channels-first cases...

aten/src/ATen/native/Convolution.cpp
```diff
@@ -413,8 +413,19 @@ struct ConvParams {
     if (!detail::getCUDAHooks().compiledWithCuDNN()) {
       return false;
     }
+    static long cudnn_version = detail::getCUDAHooks().versionCuDNN();
+    // broken on cuDNN 9.8
+    if (cudnn_version >= 90800) {
+      if (cudnn_conv_suggest_memory_format(input, weight) == at::MemoryFormat::Contiguous &&
+          input.scalar_type() == at::kBFloat16 || input.scalar_type() == at::kHalf) {
+        for (int i = 2; i < weight.dim(); i++) {
```
ngimel, line 422: does it affect both 2d and 3d, or 3d only?

eqy: I'll check with the team to confirm but my understand is it's 3d only as the user who reported said they could work around it with a 2D conv

ngimel: so you need to also check `weight.dim() == 5`?

malfet (approved):

PR #163586, opened 2025-09-22
[pt2][cache] local and global fresh cache helpers

nmacchioni:
Summary:

tldr;
```
from torch._inductor.cache import InMemoryCache, with_fresh_cache

cache = InMemoryCache()
cache.insert(key, value)

with cache.with_fresh_cache():
    assert cache.get(key) is None

with with_fresh_cache():
    assert cache.get(key) is None

with cache.with_fresh_cache():
    cache.insert(key_2, value_2)
    with with_fresh_cache():
        assert cache.get(key_2) is None
    assert cache.get(key_2) == value_2

assert cache.get(key) == value
```

Now that I type out an example, I realize that calling it "with_fresh_cache" is repetitive, so taking ideas on more suitable names

+ more preference towards Sequence over list
+ helper functions in unit tests to improve readability
+ some shuffling on names from foo to _foo for class methods which should not be called directly by the user

Test Plan:
```
buck test fbcode//mode/opt caffe2/test/inductor:cache
```

Differential Revision: D83013658


torch/_inductor/cache.py
```diff
@@ -0,0 +1,637 @@
...
+
+if TYPE_CHECKING:
+    from concurrent.futures import Future, ThreadPoolExecutor
+
+
+# TypeVars can't be recursive, so generic types that fall within
+# Key or Value can't be bound properly; for example, Key should
+# only take tuples of other Key types: tuple[Key, ...]. this is
+# a known shortcoming of torch's typing
+Key = TypeVar("Key", str, int, tuple[Any, ...])
```
aorenste, line 28: nit: Should probably be _Key...
Same for _Value, _GLOBAL*

Not a big deal since _inductor is private so these aren't part of the public API but still indicates private to the file...

torch/_inductor/cache.py
```diff
@@ -0,0 +1,637 @@
...
+@contextmanager
+def with_fresh_cache() -> Iterator[None]:
+    """
+    Context manager for a global fresh cache context.
+    Increments the global fresh cache index upon entry and decrements it upon exit.
+    Ensures that all cache instances are updated and unused fresh cache instances
+    are removed from the global registry when the context is exited.
+    """
+    global GLOBAL_FC_IDX
+    with GLOBAL_FC_IDX_LOCK:
```
aorenste, line 52: nit: I would just use the GLOBAL_CACHE_REGISTRY_LOCK for both of these. The way you have this works but if someone else accidentally adds a "with REGISTRY_LOCK, IDX_LOCK" then you risk deadlock.  Only one lock = no possibility for deadlock.

torch/_inductor/cache.py
```diff
@@ -0,0 +1,637 @@
...
+        GLOBAL_FC_IDX += 1
+    try:
+        yield
+    finally:
+        with GLOBAL_FC_IDX_LOCK, GLOBAL_CACHE_REGISTRY_LOCK:
+            # we need to 1) remove unreferenced fresh cache
+            # instances from all cache instances and 2) update
+            # the global cache registry to remove the now deleted
+            # fresh cache instances
+            _fc_to_del: list["Cache"[Any, Any]] = []
```
aorenste, line 62: Since you import annotations I'm surprised you need to quote "Cache".

torch/_inductor/cache.py
```diff
@@ -0,0 +1,637 @@
...
+                    continue
+                _fc = cache._fc_idx_to_cache[cache._fc_idx]
+                del cache._fc_idx_to_cache[cache._fc_idx]
+                _fc_to_del.append(_fc)
+            for _fc in _fc_to_del:
+                try:
+                    GLOBAL_CACHE_REGISTRY.remove(_fc)
+                except ValueError:
+                    pass
+                del _fc
```
aorenste, line 74: This is no different from (and more confusing than)
```
_fc = None
```
what's the intention?

torch/_inductor/cache.py
```diff
@@ -0,0 +1,637 @@
...
+
+    def __new__(cls) -> Self:
+        """
+        Create a new cache instance and register it in the global cache registry.
+        Returns:
+            Self: The newly created cache instance.
+        """
+        global GLOBAL_CACHE_REGISTRY
+        self = super().__new__(cls)
+        with GLOBAL_CACHE_REGISTRY_LOCK:
```
aorenste, line 99: nit: technically doesn't need the lock - python dict insertions are defined as atomic
(https://docs.python.org/3.10/faq/library.html#what-kinds-of-global-value-mutation-are-thread-safe)
(But obv. having the lock makes it easier to reason about the interaction between this and the context finally above)

torch/_inductor/cache.py
```diff
@@ -0,0 +1,637 @@
...
+    
+    def _construct_fc(self: Self) -> Self:
+        """
+        Construct a new fresh cache instance.
+        Subclasses may override this method to provide custom behavior
+        for fresh cache instantiation.
+        Returns:
+            Self: A new fresh cache instance.
+        """
+        return type(self)()
```
aorenste, line 142: This seems kind of odd to me - I feel like you've got two separate concepts which are mushed together in a single class. Wouldn't it be better to separate them and have a generic wrapper ("Cache") which does all the FC stuff and then have sub-types ("CacheImpl") which implement the implementation specific logic?

For example for an InMemoryCache the thing the user would get would effectively be a `Cache(InMemoryCacheImpl)`.  Cache.get() would do the FC stuff to ensure it's a fresh cache and then forward to the appropriate InMemoryCacheImpl.get()...

eellison (commented):
deferring to @aorenste and @masnesral on this one

PR #163589, opened 2025-09-23
[Inductor] deterministic mode

shunting314:
* #164533
* #164532
* #164525
* __->__ #163589

Add a deterministic mode to skip the on device benchmarking that we know should affect numeric. This include
- pad-mm
- dynamic rblock scaling
- template autotuning
- coordinate descent tuning for reduction
- reduction config autotuning in CachingAutotuner.  For reduction both RBLOCK, num_warps should affect numeric. XBLOCK does not. We can still autotune XBLOCK for reductions.
- benchmarking for computation communication reordering pass

The mode definitely has perf hit.


torch/_inductor/config.py
```diff
@@ -700,6 +700,10 @@ def use_autoheuristic(name: str) -> bool:
 # enabling both of these will implicitly disable split_reductions
 split_reductions = True
 
+# A deterministic mode that skips any on device benchmarking in Inductor
+# if we know they affect numerics.  WARNING: Expect perf hit in this mode.
+deterministic = os.getenv("TORCHINDUCTOR_DETERMINISTIC") == "1"
```
eellison, line 711: bike-shedding - is it worth making this a bit more structured so that we can extend it for future cases ?
```
class determinism:
      
	  ...
 	 static_shape_numerics = ...
      ...
     # nyi 
      batch_invariant_deterministic = ...
      ...
```
      
Or maybe
```
class determinism:

    static_shape_numerics = False

    
    joint_graph_passes = False

    post_grad_passes = False

    inductor_codegen = False

    inductor_config_selection = False
```

With the latter, we could potentially set as context manager throughout stack.

cc @jansel for thoughts

torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -1142,13 +1145,26 @@ def coordinate_descent_tuning(self, launcher, *args, **kwargs):
...
         ):
             # skip triton template
             return launcher
 
+        if self.deterministic_mode and self.heuristic_type in (
+            HeuristicType.REDUCTION,
+            HeuristicType.PERSISTENT_REDUCTION,
+            HeuristicType.SPLIT_SCAN,
+        ):
+            # Not only RBLOCK size matters for numericals of reduction.
```
eellison, line 1214: Is it worth tuning the xblock ?

torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -2807,6 +2824,15 @@ def adapt_config_for_tiling(
     )
 
 
+def filter_reduction_configs_for_determinism(configs: list[Config]) -> list[Config]:
```
eellison: Could be interesting to have some benchmarks for this, to see how well we do.

torch/_inductor/fx_passes/pad_mm.py
```diff
@@ -411,6 +411,10 @@ def _should_pad_bench(
     op: torch._ops.OpOverloadPacket,
     input: Optional[Tensor] = None,
 ) -> bool:
+    if torch._inductor.config.deterministic:
```
eellison, line 414: This doesnt handle interop with existing cache, cc @nmacchioni .

torch/_inductor/codegen/triton.py
```diff
@@ -4159,6 +4159,7 @@ def inductor_meta_common():
             "min_split_scan_rblock": config.triton.min_split_scan_rblock,
             "spill_threshold": config.triton.spill_threshold,
             "store_cubin": config.triton.store_cubin,
+            "deterministic": config.deterministic,
```
eellison, line 4258: This will also mess up caching. In end state we would need to check if a cache exists for this with `deterministic=False`.  but not needed initially.

torch/_inductor/config.py
eellison: asking because for auto parallel, we want joint_graph, and post_grad_passes to be deterministic, but not necesarily inductor_codegen and inductor_config_selection.

eellison (commented):
Looks good! lets get takes from other folks too before landing. this doesnt handle interaction with caching, but that can be done subsequently, I know @nmacchioni has some early starts there.

torch/_inductor/codegen/triton.py
shunting314: yea, I think these deterministic flags can be add as part of the cache-key.

torch/_inductor/config.py
eellison: It could also potentially be `True / False / ranks_synchronize`  or something along those lines which indicates to sync ranks 🤷  The last option can also be used for auto parallel.

torch/_inductor/runtime/triton_heuristics.py
shunting314: I create a script to verify the triton reduction determinism regarding xblock/rblock/num_warps: https://gist.github.com/shunting314/8364e7857425b4b8812549dd449a0523 .

It shows that rblock/num_warps indeed affect numerics . And in this test, xblock does not affect numeric. but I'm not sure if this is in general true. Not sure if triton would pick different layout when xblock changes. cc @htyu any thoughts? Basically the question is, can the numeric of a reduction change in Triton if the only change is the XBLOCK.

torch/_inductor/runtime/triton_heuristics.py
shunting314: Tried on a HF model BertForMaskedLM:

- torch.compile out of the box perf: 1.544x (against eager)
- deterministic mode: 1.334x
- deterministic mode (but with padmm): 1.456x
- deterministic mode (but with padmm, and not filter reduction configs): 1.557x

padmm and the reduction config are roughly equally important for this model (since vocabulary size is an odd number)

torch/_inductor/config.py
jansel: A single flag is easier to enable.  If we add more flags, then enabling them 1-by-1 would be a pain.

Supporting a string version seem fine though.

torch/_inductor/runtime/triton_heuristics.py
jansel: Hrm, might be worth spending a few hours trying to come up with a heuristic that is better than picking configs[0].

torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -1182,6 +1198,7 @@ def benchmark_one_config(config):
             config2launcher[config] = launcher
 
             out = self.bench(launcher, *args, **kwargs)
+            counters["inductor"]["coordest_tuning_bench"] += 1
```
PaulZhang12: Nit: rename coord_desc_tuning_bench

torch/_inductor/runtime/triton_heuristics.py
PaulZhang12: configs[0] case here should already pick quite good configs given the updated heuristics, can you maybe retry also on top of https://github.com/pytorch/pytorch/pull/162447 ? I'm curious where the gap is

torch/_inductor/select_algorithm.py
```diff
@@ -2600,6 +2600,16 @@ def cache_clear(self) -> None:
...
 
+    def pick_deterministic_choice(self, choices: list[ChoiceCaller]) -> ChoiceCaller:
+        assert len(choices) >= 2
+        externs = [
+            choice for choice in choices if isinstance(choice, ExternKernelChoice)
+        ]
+        if len(externs) > 0:
+            return externs[0]
+        else:
+            return choices[0]
```
eellison, line 2621: Maybe the "cache interop" is a better solve but similarly, i imagine this will be pretty slow for flex attention - cc @drisspg .

torch/_inductor/runtime/triton_heuristics.py
htyu: Rblock/num_warps affects reduction order. Triton by default chooses to divide a row among multiple warps horizontally, as described in https://docs.google.com/document/d/1-D4cWt3QVtPXcOQxDVSuOBCUkNKrbJ9Vl8s-QIA0fes/edit?tab=t.0

XBLOCK shouldn't, as thread distribution along the columns takes priority.

torch/_inductor/runtime/triton_heuristics.py
shunting314: Changed this part a bit by:
1. do autotuning for xblock
2. pick a group with the smallest rblock

Now, the perf in deterministic mode is 1.416x .  This may be further improved if we consider features like xnumel/rnumel/num-load/num-store/num-contiguous-acess/num-non-contiguous-access etc. 

I think picking configs[0] mainly not cover the case that a reduction have both contiguous and non-contiguous access

torch/_inductor/config.py
v0i0: for my understanding (and maybe to elaborate the comment): this flag is intended as a kind of override, right? i.e. it takes precedence over all other config options that might trigger benchmark-guided optimization?

torch/_inductor/runtime/triton_heuristics.py
v0i0: i am not sure this is sufficient for non-determinism, specifically when rblock is small, have you checked? i'd think that tuning xblock can change the # of elements per thread, which might impact how the reduction is performed?

v0i0 (approved):
grepping through the code suggests some other configs as well (e.g. runtime_estimations_mms_benchmark) that should be overridden?

torch/_inductor/runtime/triton_heuristics.py
shunting314: > i am not sure this is sufficient for non-determinism, specifically when rblock is small, have you checked?

Why it's insufficient? We do step 2 with deterministic heuristics. In this version, I just return the one with smallest RBLOCK.

torch/_inductor/config.py
shunting314: yes

eellison:
> grepping through the code suggests some other configs as well (e.g. runtime_estimations_mms_benchmark) that should be overridden?

Would it be safer to make do_bench error when determinism is set ? And we can explicitly allow it in portions of code we have vetted as equivalent.

shunting314:
> Would it be safer to make do_bench error when determinism is set ? And we can explicitly allow it in portions of code we have vetted as equivalent.

make sense. This would also guard against benchmarking code added in future as long as people do benchmarking with the small set of benchmarking functions in Inductor. I'll add this in a stacked PR

shunting314:
> Would it be safer to make do_bench error when determinism is set ? And we can explicitly allow it in portions of code we have vetted as equivalent.

hmm, people call triton.testing.do_bench directly sometimes. So even if we add this, it would still be best effect but can not always guaranteed no benchmarking happening.

torch/_inductor/runtime/triton_heuristics.py
v0i0: ah got it

torch/_inductor/select_algorithm.py
drisspg: So by default Flex only picks 1 choice and then we dump them all for max-autotune. That choice is forsure not always the best but I dont think this will be a delta unless max-autotune is enabled

torch/_inductor/runtime/triton_heuristics.py
shunting314: I collected the autotuning results for huggingface benchmark suite ( https://gist.github.com/shunting314/a52aa2b436229d1b92d6d116163d1a4a ) and analyze manually 5 of the models (https://gist.github.com/shunting314/c9322d90ce405ec0e24aac3f97eb1963 ) to come up with some heuristics.  These heuristics make BertForMaskedLM having on par perf with or without filtering.

shunting314: The heuristic works even better than I think. I ran a perf test forcing TORCHINDUCTOR_FORCE_FILTER_REDUCTION_CONFIGS to be true (i.e. do reduction config filtering. But mm-padding etc are still active). [result](https://hud.pytorch.org/benchmark/compilers?dashboard=torchinductor&startTime=Thu%2C%2025%20Sep%202025%2000%3A24%3A47%20GMT&stopTime=Thu%2C%2002%20Oct%202025%2000%3A24%3A47%20GMT&granularity=hour&mode=training&dtype=amp&deviceName=cuda%20(h100)&lBranch=gh/shunting314/228/head&lCommit=582267e91d65739a43cd77434746bbdf31e059c6&rBranch=main&rCommit=77b9aac6c2ab41e07ca2643260ca3e6f059abc4a).

TB: 3% perf hit, 3s compilation time speedup
HF: 6% perf hit, 1s compilation time speedup
TIMM: 1% perf hit, 6s compilation time speedup.

Looks like this is very beneficial for TIMM.

Screenshot 2025-10-01 at 5 26 17 PM

nmacchioni:
> @nmacchioni
> 
> > the cache will spit out an override file that users can copy/paste to replay the same results (and I guess this is where folks can choose to meddle with the results)
> 
> Is there a way to force determinism and error if the override file doesn't have enough information?

Yeah, in my mind we can either throw an exception if the override file doesn't have the required information to be deterministic or we can fallback to synchronizing off of the remote db. OSS will have to throw an exception by default, as we're not providing the remote db there. Internally we'll just have a toggle for whatever behavior is wanted

PR #160380, opened 2025-08-11
[AOTInductor] Add input information for Triton Kernels in AOTI

muchulee8:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #160380
* #163593

Summary:
We use record_function to pass in input information to let Kineto show
input information.

Test Plan:
Before:
Screenshot 2025-09-19 at 10 45 10 AM

After:
Screenshot 2025-09-19 at 10 44 53 AM

Reviewers:

Subscribers:

Tasks:

Tags:

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @amjames @chauhang @aakhundov @coconutruben

PR #163593, opened 2025-09-23
[Kineto] Add list of string parsing for profiler

muchulee8:
* #160380
* __->__ #163593

Summary:
We add the parsing for list of string. This is needed for AOTInductor
profiling for input information of Triton kernels.

Test Plan:
Included in commit.
test_profiler_op_event_kwargs_list_of_strings

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/torch/csrc/autograd/profiler_kineto.cpp
+++ b/torch/csrc/autograd/profiler_kineto.cpp
@@ -264,16 +264,38 @@ struct AddGenericMetadata : public MetadataBase {
         continue;
       }
 
-      // Until needed, lets limit the kwargs to only ints, doubles, strings and
-      // bools
-      if (!val.isInt() && !val.isDouble() && !val.isString() && !val.isBool()) {
-        LOG(WARNING) << "Inputted kwarg: " << key
-                     << " is not an int, double, string, or bool for op: "
-                     << op_event.name_ << " skipping";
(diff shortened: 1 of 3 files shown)


torch/csrc/autograd/profiler_kineto.cpp
```diff
@@ -264,16 +264,38 @@ struct AddGenericMetadata : public MetadataBase {
...
+        LOG(WARNING)
+            << "Inputted kwarg: " << key
+            << " is not an int, double, string, bool, or list of strings for op: "
+            << op_event.name_ << " skipping";
         continue;
       }
-      bool isString = val.isString();
-      addMetadata(key, ivalueToStr(val, isString));
+
+      if (isStringList) {
```
sraikund16, line 290: why only string list instead of allowing for int list and double list?

sraikund16 (approved):
Overall, LGTM. Just one question

torch/csrc/autograd/profiler_kineto.cpp
muchulee8: I'm just trying to play safe, since I saw the change of limiting kwargs type was triggered by a SEV a while ago, so just to limit it to list of string, and also only testing on this.

PR #163596, opened 2025-09-23
[Inductor-FX] Support symbol and dynamic scalar graph inputs and outputs

blaine-rister:
# Problems
This PR fixes a few edge cases that the FX converter missed related to dynamic shapes. 

1. Inductor graphs can sometimes take `sympy.Symbol` inputs. We have logic to convert these to FX placeholder nodes. However, this logic did not update the `self.expr_to_proxy` table mapping symbols to proxy nodes. (There was existing logic to do this for `ir.TensorBox` inputs, but not `sympy.Symbol`.) This caused sympy tracing to fail when these symbol inputs were used in other expressions.

2. We lacked codegen for `ShapeAsConstantBuffer`. This IR node is seen when the graph input or output is a scalar computed from dynamic shapes.

# Fixes
a. Update `self.expr_to_proxy` when generating placeholders for `sympy.Symbol` inputs. Change `SymbolBuffer.get_example` to convert the symbol to a `torch.SymInt`, so we can populate `meta["val"]` correctly and use the value in other computations.
b. Support `ShapeAsConstantBuffer` by tracing the sympy expression.
c. Move output generation inside the metadata hook, allowing us to populate `meta["val"]` for the nodes computing `ShapeAsConstantBuffer`.

# Test plan
Added several new CI tests:
 1. `torch.cond` with dynamic shapes. This exposes both issues, as the predicate is a `ShapeAsConstantBuffer` and one of the subgraphs uses a symbol input, due to the closure. Also tests when the parent and subgraphs have different input shapes.
 2. Output dynamic shape scalar. This tests `ShapeAsConstantBuffer` as an output.


torch/_inductor/codegen/wrapper_fxir.py
```diff
@@ -475,9 +494,7 @@ def codegen_proxy() -> torch.fx.Proxy:
                     undefined_symbol_expr
                 ]
 
-        for node in V.graph.module.graph.find_nodes(op="placeholder"):  # type: ignore[operator, union-attr]
```
blaine-rister, line 478: This could cause an issue with subgraphs, as `V.graph.module` comes from the parent graph. Subgraph inputs may not have the same name.

blaine-rister: TODO test case for this? Pass a dynamic shaped tensor computed in the parent to the subgraph.

blaine-rister: Never mind. `V.graph.module` adapts to the subgraph. Enhanced the test case anyways.

angelayi (approved):

jansel (approved):

PR #163599, opened 2025-09-23
[BE] Using std::move to reduce copy constructor calls by one.

thenumberouscode:
inspired by https://github.com/pytorch/pytorch/pull/163416
cc @jerryzh168 @Skylion007


aten/src/ATen/native/Linear.cpp
```diff
@@ -632,15 +632,15 @@ Tensor einsum(std::string_view equation, TensorList operands, at::OptionalIntArr
       for (auto dim = perm_index - 1; dim >= out_num_dim; --dim) {
         sizes.erase(sizes.begin() + dim);
       }
-      return ops[0].view_symint(sizes);
+      return std::move(ops[0].view_symint(sizes));
     } else {
       std::vector<int64_t> sum_dims(perm_index - out_num_dim);
       std::iota(sum_dims.begin(), sum_dims.end(), out_num_dim);
-      return ops[0].sum(sum_dims);
+      return std::move(ops[0].sum(sum_dims));
```
Skylion007: Doesn't sum return an rvalue already? Isn't this redundant?

aten/src/ATen/native/TensorShape.cpp
```diff
@@ -2132,10 +2132,11 @@ static Tensor select_sparse(const Tensor& self, int64_t dim, int64_t index) {
     if (sparse_dim == 1) {
       // return dense part:
       if (new_values.size(0) == 1) {
-        return new_values[0];
+        return std::move(new_values[0]);
       } else {
         // sum promotes integral type to int64 when dtype is not specified.
-        return at::sum(new_values, 0, false, new_values.scalar_type());
+        return std::move(
```
Skylion007: Same here, should be rvalue, right?

aten/src/ATen/native/TensorShape.cpp
```diff
@@ -2149,23 +2150,23 @@ static Tensor select_sparse(const Tensor& self, int64_t dim, int64_t index) {
                             .view(-1);
       auto new_indices =
           indices.index_select(1, nzIndices).index_select(0, dimIndices);
-      return _sparse_coo_tensor_with_dims_and_tensors(
+      return std::move(_sparse_coo_tensor_with_dims_and_tensors(
```
Skylion007: Same.

aten/src/ATen/native/TensorShape.cpp
```diff
@@ -2149,23 +2150,23 @@ static Tensor select_sparse(const Tensor& self, int64_t dim, int64_t index) {
...
           new_sizes,
           new_indices,
           new_values,
-          self.options());
+          self.options()));
     }
   } else {
     auto new_values = values.select(dim - sparse_dim + 1, index);
-    return _sparse_coo_tensor_with_dims_and_tensors(
+    return std::move(_sparse_coo_tensor_with_dims_and_tensors(
```
Skylion007: Same, these should be return rvalues already, right??? Or it it just return an lvalue which we are casting to an xvalue?

aten/src/ATen/native/Linear.cpp
```diff
@@ -632,15 +632,15 @@ Tensor einsum(std::string_view equation, TensorList operands, at::OptionalIntArr
       for (auto dim = perm_index - 1; dim >= out_num_dim; --dim) {
         sizes.erase(sizes.begin() + dim);
       }
-      return ops[0].view_symint(sizes);
+      return std::move(ops[0].view_symint(sizes));
```
Skylion007: Same here.

Skylion007 (changes requested):
Most of these moves don't make sense, a few do though.

aten/src/ATen/native/Linear.cpp
```diff
@@ -632,15 +632,15 @@ Tensor einsum(std::string_view equation, TensorList operands, at::OptionalIntArr
...
     } else {
       std::vector<int64_t> sum_dims(perm_index - out_num_dim);
       std::iota(sum_dims.begin(), sum_dims.end(), out_num_dim);
-      return ops[0].sum(sum_dims);
+      return std::move(ops[0].sum(sum_dims));
     }
   }
 
-  return ops[0];
+  return std::move(ops[0]);
```
Skylion007, line 643: This the only move that looks valid in this PR...

Skylion007 (changes requested):

aten/src/ATen/native/TensorShape.cpp
thenumberouscode: Sorry, I will change it immediately.

aten/src/ATen/native/TensorShape.cpp
```diff
@@ -2148,7 +2148,7 @@ static Tensor select_sparse(const Tensor& self, int64_t dim, int64_t index) {
     if (sparse_dim == 1) {
       // return dense part:
       if (new_values.size(0) == 1) {
-        return new_values[0];
+        return std::move(new_values[0]);
```
thenumberouscode: @Skylion007 It looks like this line only works in gcc, failed in clang in workflow:https://github.com/pytorch/pytorch/actions/runs/18460272315/job/52763454958?pr=163599

thenumberouscode: oh, I have found the real reason:
The above changes indeed prevent copy elision working. for new_value is a tensor not vector. so the clang is clever enough to check it out by options -Wmove open., while gcc is not.

thenumberouscode:
> The error message indicates that Clang could do it
> 
> ```
> error: moving a temporary object prevents copy elision
> ```
> 
> You can wrap your changes in conditional macros to check non-Clang scenarios.

@cyyever 
After careful analysis, I need to correct my previous explanation:

When returning an element from a vector that is a local variable within a function, RVO (Return Value Optimization) does apply to the element itself. However, there's an additional copy constructor call that occurs when extracting the element from the vector. The element gets copied out of the vector as an rvalue first, and then RVO can optimize the return. By using std::move, we invoke the move constructor instead of the copy constructor, thereby reducing the overhead.

 The error message is raised for I wrongly thought new_values is a vector which is a tensor,  and using std::move on it caused the copy elision error. cc @Skylion007

Skylion007 (approved):

PR #163602, opened 2025-09-23
[dynamo] trace local_map with local shapes for AP

xmfan:
* #164437
* #164433
* #164431
* __->__ #163602
* #164340
* #164420
* #164419
* #164321
* #164296

Context is in https://www.internalfb.com/excalidraw/EX519691 and https://docs.google.com/document/d/1qnuXLZk_GYt_PksHTwkn7L2ELRDnYlIRPkHAlXTyuhw/edit?tab=t.0. And the description of the previous PR: https://github.com/pytorch/pytorch/pull/164340.

The previous PR adds the support on the HOP side for eager execution and AOTAutograd. Dynamo is still passing the HOP a subgraph with wrong shapes. This PR fixes that. This is similar to the HOP implementation, however we additionally need to manually keep the TensorVariable metadata in sync.


ezyang:
still draft?

test/higher_order_ops/test_local_map.py
```diff
@@ -460,12 +517,12 @@ def test_local_map_dynamo_mismatch_placements(self):
...
             redistribute_inputs=True,
             in_grad_placements=None,
-            device_mesh=None,
+            device_mesh=self.mesh,
         )
         def mismatch_input(x, scalar):
             return x + scalar, scalar
 
-        x = torch.randn(36, 36, requires_grad=True)
+        x = torch.randn(64, 64, 64, requires_grad=True)
```
xmfan, line 520: changed these so that it's evenly shardable

torch/_dynamo/variables/higher_order_ops.py
```diff
@@ -3576,6 +3576,63 @@ def _call_function(
...
+            # This is okay since accessing grad_fn is a graph break.
+            local_tensor = redistribute_fw_inputs(
+                (global_tensor,),
+                (placements,),
+                mesh,
+            )
+            local_tensor = local_tensor[0]
+
+            priors[vt] = global_tensor
+            vt.as_proxy().node.meta["example_value"] = local_tensor
```
ydwu4, line 3630: Mutating an existing variable tracker looks suspicious as some other subsystems might assume the mutation always come from interpreting user code that actually does the mutation. There could be some future risk if we don't differentiate these two cases e.g. these two syncs might get recorded as two attribute mutations.

The alternative is to create new tensor variables with desirable shapes e.g.like in [flex_attention](https://github.com/pytorch/pytorch/blob/main/torch/_dynamo/variables/higher_order_ops.py#L2844-L2850), we create new_empty tensors and enclosing them with discard_graph_changes().

ydwu4 (approved):

PR #163605, opened 2025-09-23
Move non-tensor nodes on boundary after split

openzig:
Summary: See details in the attached doc
[AOTI Non-Tensor on Boundary Issue.pdf](https://github.com/user-attachments/files/22608882/AOTI.Non-Tensor.on.Boundary.Issue.pdf)

Meta internal post: https://fb.workplace.com/groups/aoti.productionization/permalink/743060011864813/

Test Plan: unit test

Bifferential Revision: D80903201

Steps to show how this feature works:
- checkout commit: 01eb5dd5c3dc3e5d62fc975da9bcfee4befbd421
- run this local lowering command: P2072906967
You will see error like below, this means there is a node on the boundary whose dtype is `torch.dtype`, the corresponding model code is [this line](https://www.internalfb.com/code/fbsource/[e9ae35402388]/fbcode/minimal_viable_ai/umia_v1/ig/omni/root_model.py?lines=2409)
```
Found <class 'torch.dtype'> in output, which is not a known type.
```
- Then search `move_non_tensor_nodes_on_boundary` in file `fbcode/caffe2/torch/fb/model_transform/lower_presets/aoti/ig_ss_t2i_retrieval_aot_inductor.py`, you will see it is commented, now set `move_non_tensor_nodes_on_boundary` to true, and rerun the command in last step, you will see the tracing is done and `_exported_run_on_acc_1_readable.txt` is generated. (although there is IMA error which is not related to this diff)

### DIFF
--- a/torch/fx/passes/split_utils.py
+++ b/torch/fx/passes/split_utils.py
@@ -8,10 +8,16 @@ from torch.fx._compatibility import compatibility
 from torch.fx.graph import map_arg
 from torch.fx.passes.utils import HolderModule, lift_subgraph_as_module
 
-from .tools_common import NodeList
+from .tools_common import CALLABLE_NODE_OPS, is_node_output_tensor, NodeList
 
 
-__all__ = ["getattr_recursive", "setattr_recursive", "Component", "split_by_tags"]
+__all__ = [
+    "getattr_recursive",
+    "setattr_recursive",
+    "Component",
+    "split_by_tags",
+    "move_non_tensor_nodes_on_boundary",
+]
 
 
 @compatibility(is_backward_compatible=False)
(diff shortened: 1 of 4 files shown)


torch/fx/passes/splitter_base.py
```diff
@@ -119,6 +120,16 @@ def __init__(
             "we might not care about non-tensor data flow and we can set this option "
             "to true to disable the functionality that prevent non-tensor data flow.",
         )
+        parser.add_argument(
+            "--move-non-tensor-nodes-on-boundary",
+            "--move_non_tensor_nodes_on_boundary",
+            required=False,
+            type=bool,
+            help="AOTI does not support non-tensor nodes on acc->acc, acc->gpu and gpu->acc boundary. "
+            "For non-tensor nodes on acc->acc boundary and acc->gpu, we move the nodes from upstream to downstream. "
```
sevenEng: i'm thinking will acc->gpu also already handled by `reduce_acc_nodes_non_tensor_output`

inside `reduce_acc_nodes_non_tensor_output`, we'll find these nodes who are in acc but produce non-tensor output and have users in non-acc, and then move them out of acc

torch/fx/passes/splitter_base.py
```diff
@@ -1092,6 +1104,8 @@ def split(self, remove_tag: bool = False) -> torch.fx.GraphModule:
     def __call__(self) -> torch.fx.GraphModule:
         subgraphs = self.put_nodes_into_subgraphs()
         subgraphs = self.remove_small_acc_subgraphs(subgraphs)
+        if self.settings.move_non_tensor_nodes_on_boundary:
+            move_non_tensor_nodes_on_boundary(subgraphs)
```
sevenEng: i'm wondering if it's possible to integrate this with the node tracker in some way,

as here we're changing the split result, and the node movements we tracked in it will actually be wrong

```
        self.tracker = NodeEventTracker(int(TRACKER_MODE), DUMP_PREFIX)
```

torch/fx/passes/split_utils.py
```diff
@@ -310,3 +316,192 @@ def remap_func(x):
         return result_gm, orig_to_split_fqn_mapping
 
     return result_gm
+
+
+@compatibility(is_backward_compatible=False)
+def move_non_tensor_nodes_on_boundary(subgraphs) -> None:
```
ColinPeppler, line 322: there should be a type for this: `list[Subgraph]`

torch/fx/passes/split_utils.py
```diff
@@ -310,3 +316,192 @@ def remap_func(x):
...
+    """
+    Move non-tensor nodes on the boundary between subgraphs.
+
+    For each subgraph:
+
+    1. Find nodes whose type is not tensor and any of its children is in another
+       subgraph, put them in a queue for next step
+
+    2. Do a BFS on those nodes, for each node, let's say node X and it is in subgraph A:
```
ColinPeppler: I'm thinking you're referring to `can_move_node_and_dependencies` here.

```suggestion
    2. Do a DFS on those nodes, for each node, let's say node X and it is in subgraph A:
```

torch/fx/passes/split_utils.py
```diff
@@ -310,3 +316,192 @@ def remap_func(x):
...
+            # Skip if node is no longer in this subgraph (may have been moved)
+            if (
+                current_node not in node_to_subgraph
+                or node_to_subgraph[current_node] != subgraph_idx
+            ):
+                continue
+
+            children = get_children_in_graph(current_node)
+            if not children:
+                continue
```
ColinPeppler: nit: this can be an assertion given we only add nodes into the queue if `has_children_in_other_subgraph` is True (and we know it to be true).

torch/fx/passes/split_utils.py
```diff
@@ -310,3 +316,192 @@ def remap_func(x):
...
+
+            # Find target subgraph. The children should all be in the same subgraph except current subgraph
+            target_subgraph_candidates = set()
+            for child in children:
+                child_subgraph = node_to_subgraph[child]
+                if child_subgraph != subgraph_idx:
+                    target_subgraph_candidates.add(child_subgraph)
+            if len(target_subgraph_candidates) != 1:
+                continue
```
ColinPeppler, line 477: Worth a comment to call out any cases not covered i.e. multiple children living in different dependent subgraphs.

torch/fx/passes/split_utils.py
```diff
@@ -310,3 +316,192 @@ def remap_func(x):
...
+            can_move, nodes_to_move = can_move_node_and_dependencies(
+                current_node, subgraph_idx, target_subgraph
+            )
+
+            if can_move:
+                # Move all nodes in nodes_to_move to target subgraph
+                for node_to_move in nodes_to_move:
+                    if (
+                        node_to_move in node_to_subgraph
+                        and node_to_subgraph[node_to_move] == subgraph_idx
```
ColinPeppler: nit: better to assert this given `nodes_to_move` should only contain nodes that need to be moved.

torch/fx/passes/split_utils.py
```diff
@@ -310,3 +316,192 @@ def remap_func(x):
...
+                parents = get_parents_in_graph(current_node)
+                for parent in parents:
+                    if (
+                        not is_node_output_tensor(parent)
+                        and parent not in processed
+                        and parent in node_to_subgraph
+                        and node_to_subgraph[parent] == subgraph_idx
+                    ):
+                        # Check if parent meets step 1 requirement: any children in another subgraph
+                        if has_children_in_other_subgraph(parent, subgraph_idx):
```
ColinPeppler: nit: this could be assertion given we just moved `current_node` and the selected dependencies.

torch/fx/passes/splitter_base.py
```diff
@@ -131,6 +142,7 @@ def __init__(
             args.allow_non_tensor if args.allow_non_tensor else allow_non_tensor
         )
         self.max_acc_splits: int = max_acc_splits
+        self.move_non_tensor_nodes_on_boundary: bool = move_non_tensor_nodes_on_boundary
```
ColinPeppler: following other args

```suggestion
  self.move_non_tensor_nodes_on_boundary: bool = (
      args.move_non_tensor_nodes_on_boundary
      if args.move_non_tensor_nodes_on_boundary
      else move_non_tensor_nodes_on_boundary
  )
```

torch/fx/passes/splitter_base.py
ColinPeppler: Is there a good reason to call `move_non_tensor_nodes_on_boundary` after `remove_small_acc_subgraphs`? I think it would make more sense to call it before because it can modify the acc subgraphs (e.g. non-tensor node move causes acc subgraph to shrink).

torch/fx/passes/split_utils.py
openzig: > I'm thinking you're referring to `can_move_node_and_dependencies` here.

you are right, this is what can_move_node_and_dependencies does, I put it here just want to give the user a brief overview of the whole process

torch/fx/passes/split_utils.py
openzig: done

torch/fx/passes/split_utils.py
openzig: seems that the check is redundant, removed

torch/fx/passes/split_utils.py
openzig: done

torch/fx/passes/splitter_base.py
openzig: you are right, run it before removing small acc graph is better as it will change the graph size

torch/fx/passes/split_utils.py
openzig: I cannot use explicit type list[Subgraph] here as bc linter check does not allow deleting class ([error](https://github.com/pytorch/pytorch/actions/runs/20197443551/job/57983619414)), and adding suppress seems not working. And I cannot import it from the splitter_base as it will cause circular dependency

torch/fx/passes/split_utils.py
```diff
@@ -310,3 +316,200 @@ def remap_func(x):
...
+            if (
+                current_node not in node_to_subgraph
+                or node_to_subgraph[current_node] != subgraph_idx
+            ):
+                continue
+
+            children = get_children_in_graph(current_node)
+            assert children is not None, (
+                "Only node that has children in other subgraph can be moved"
+            )
```
ColinPeppler, line 464: If get_children_in_graph returns empty list it will pass the `is not None` check..


```suggestion
            children = get_children_in_graph(current_node)
            assert len(children) > 0, (
                "Only node that has children in other subgraph can be moved"
            )
```

torch/fx/passes/splitter_base.py
```diff
@@ -120,6 +119,16 @@ def __init__(
             "we might not care about non-tensor data flow and we can set this option "
             "to true to disable the functionality that prevent non-tensor data flow.",
         )
+        parser.add_argument(
+            "--move-non-tensor-nodes-on-boundary",
+            "--move_non_tensor_nodes_on_boundary",
+            required=False,
+            type=bool,
```
ColinPeppler: From [argparse docs](https://docs.python.org/3/library/argparse.html#:~:text=The%20bool()%20function%20is%20not%20recommended%20as%20a%20type%20converter.%20All%20it%20does%20is%20convert%20empty%20strings%20to%20False%20and%20non%2Dempty%20strings%20to%20True.%20This%20is%20usually%20not%20what%20is%20desired.)

> The [bool()](https://docs.python.org/3/library/functions.html#bool) function is not recommended as a type converter. All it does is convert empty strings to False and non-empty strings to True. This is usually not what is desired.

```suggestion
            action='store_true',
```

ColinPeppler (approved):
lgtm with 2 follow ups

PR #163607, opened 2025-09-23
[ez] use list initializer syntax in fill_diagonal_

bobrenjc93:
* __->__ #163607
* #163485

### DIFF
--- a/aten/src/ATen/native/Fill.cpp
+++ b/aten/src/ATen/native/Fill.cpp
@@ -109,26 +109,22 @@ Tensor& fill_diagonal_(Tensor& self, const Scalar& fill_value, bool wrap) {
   }
 
   auto storage_offset = self.sym_storage_offset();
-  std::vector<SymInt> sizes;
-  std::vector<SymInt> strides;
   auto size = std::min(height, width);
 
   int64_t stride = 0;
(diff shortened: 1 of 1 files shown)


aten/src/ATen/native/Fill.cpp
```diff
@@ -109,26 +109,22 @@ Tensor& fill_diagonal_(Tensor& self, const Scalar& fill_value, bool wrap) {
...
   auto size = std::min(height, width);
 
   int64_t stride = 0;
   for (const auto i : c10::irange(nDims)) {
     stride += self.stride(i);
   }
-  strides.push_back(stride);
-  sizes.push_back(size);
+  std::vector<SymInt> strides{stride};
+  std::vector<SymInt> sizes{size};
```
Skylion007: Size is a SymInt now right? Should call std::move here.

Skylion007 (approved):

aten/src/ATen/native/Fill.cpp
bobrenjc93: how come? std::move(size) will cast it to an rvalue reference, but constructing a std::vector<SymInt> from a single element already copies (or moves) it in.

also size is used again after this line and after we std::move it we can't safely use it anymore?

PR #163317, opened 2025-09-19
[Code Clean] Replace `std::runtime_error` with `TORCH_CHECK`

KarhouTam:
Including:
- `torch/csrc/instruction_counter`
- `torch/csrc/lazy`
- `torch/csrc/monitor`
- `torch/csrc/profiler`
- `torch/csrc/dynamo`

Fixes part of [#148114](https://github.com/pytorch/pytorch/issues/148114)

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela

PR #163610, opened 2025-09-23
[Code Clean] Replace `std::runtime_error` with `TORCH_CHECK`

KarhouTam:
Including:
- `torch/csrc/instruction_counter`
- `torch/csrc/lazy`
- `torch/csrc/monitor`
- `torch/csrc/profiler`
- `torch/csrc/dynamo`

Fixes part of #148114 

Personal mistake about (PR #163317), this PR does the same thing **and PR #163317 has already been approved by @albanD.**

This is a personal mistake on my part, and I'm so sorry about that. Hope you won't mind @albanD. 🥹

### DIFF
--- a/torch/csrc/dynamo/guards.cpp
+++ b/torch/csrc/dynamo/guards.cpp
@@ -2,6 +2,7 @@
 #include <ATen/autocast_mode.h>
 #include <c10/core/SafePyObject.h>
 #include <c10/core/impl/PyInterpreter.h>
+#include <c10/util/Exception.h>
 #define PY_SSIZE_T_CLEAN
 #include <ATen/EmptyTensor.h>
 #include <ATen/SparseCsrTensorUtils.h>
@@ -64,8 +65,7 @@ int open_counter() {
 
 uint64_t count_instructions(const std::function<void()>& fn) {
   int fd = open_counter();
-  if (fd == -1)
-    throw std::runtime_error("perf_event_open failed");
+  TORCH_CHECK(fd != -1, "perf_event_open failed");
 
   ioctl(fd, PERF_EVENT_IOC_RESET, 0);
   ioctl(fd, PERF_EVENT_IOC_ENABLE, 0);
@@ -568,7 +568,7 @@ static PyMethodDef TensorGuards_methods[] = {
     {nullptr} /* Sentinel */
 };
 
-static PyTypeObject TensorGuardsType = { PyVarObject_HEAD_INIT(nullptr, 0)
+static PyTypeObject TensorGuardsType = {PyVarObject_HEAD_INIT(nullptr, 0)
 };
 
 struct AutocastState {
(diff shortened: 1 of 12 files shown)


albanD (approved):
No worries at all!

torch/csrc/dynamo/guards.cpp
```diff
@@ -2882,9 +2883,8 @@ class GuardManager {
       return py::type::of(py::none());
     }
 
-    if (!PyCallable_Check(_weak_type.ptr())) {
-      throw std::runtime_error("_weak_type is not callable");
-    }
+    TORCH_CHECK(
```
Skylion007: Should be TORCH_CHECK_TYPE

torch/csrc/dynamo/guards.cpp
```diff
@@ -2896,10 +2896,8 @@ class GuardManager {
   }
 
   void mark_tag_safe_root() {
-    if (!_is_tag_safe) {
-      throw std::runtime_error(
-          "Marking a node tag_safe_root when its not tag safe");
-    }
+    TORCH_CHECK(
```
Skylion007, line 2899: Feels like this should be an TORCH_CHECK_ASSERT

torch/csrc/dynamo/guards.cpp
```diff
@@ -1783,8 +1782,10 @@ class LAMBDA_GUARD_NO_FRAMELOCALS : public LAMBDA_GUARD {
   }
```
Skylion007, line 1782: The pyobject here should be moved into their parent class in the ctor btw

torch/csrc/monitor/counters.cpp
```diff
@@ -21,7 +22,8 @@ const char* aggregationName(Aggregation agg) {
     case Aggregation::MIN:
       return "min";
     default:
-      throw std::runtime_error(
+      TORCH_CHECK(
+          false,
           "unknown aggregation: " + std::to_string(static_cast<int>(agg)));
```
Skylion007: Make it two args and use torch check's lazy concatenation

Skylion007 (approved):
Some last minute nits, but looks good once those are fixed.

torch/csrc/dynamo/guards.cpp
KarhouTam: Got it. Thanks.

torch/csrc/monitor/counters.cpp
KarhouTam: Got it. Thanks

torch/csrc/dynamo/guards.cpp
KarhouTam: TORCH_CHECK_ASSERT is not an available macro. Or you mean `TORCH_INTERNAL_ASSERT` ?

torch/csrc/dynamo/guards.cpp
KarhouTam: I think we can open another PR for this refactor. Wdyt? Because this PR's goal is to replace `std::runtime_error` with `TORCH_CHECK`.

PR #160938, opened 2025-08-19
[WOQ] Add XPU kernel for _weight_int8pack_mm

xiaowangintel:
Summary:
This issue proposes implementing a XPU kernel for aten._weight_int8pack_mm, a weight-only quantized (WOQ) linear operation that is currently only supported on CPU and CUDA.  

Motivation:
Same as https://github.com/pytorch/pytorch/pull/159325.

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168

PR #163615, opened 2025-09-23
[xpu][feature] Supports woq_int8 inductor pattern on Intel GPU

xiaowangintel:
Summary:

Supports woq_int8 inductor pattern on Intel GPU. When using torch.compile, woq_int8 will be lowering to _weight_int8pack_mm instead of being falled back mul().sum(). The Intel GPU backend of _weight_int8pack_mm was supported in https://github.com/pytorch/pytorch/pull/160938.

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @voznesenskym @penguinwu @EikanWang @Guobing-Chen @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @chenyang78

### DIFF
--- a/test/inductor/test_mkldnn_pattern_matcher.py
+++ b/test/inductor/test_mkldnn_pattern_matcher.py
@@ -4280,14 +4280,19 @@ class TestPatternMatcher(TestPatternMatcherBase):
             def matcher_check_fn():
                 self.assertEqual(counters["inductor"]["woq_matcher_count"], 1)
 
-            self._test_common(
-                mod,
-                (x, w, s),
-                matcher_check_fn,
-                check_quantization=False,
-                atol=0.001,
-                rtol=0.07,
-            )
+                def matcher_check_fn():
+                    self.assertEqual(
+                        counters["inductor"]["woq_matcher_count"], 0 if TEST_ACL else 1
+                    )
+
+                self._test_common(
+                    mod,
+                    (x.to(device), w.to(device), s.to(device)),
+                    matcher_check_fn,
+                    check_quantization=False,
+                    atol=0.001,
+                    rtol=0.07,
+                )
 
     @skipIfNoDynamoSupport
     def test_woq_int4_cpu(self):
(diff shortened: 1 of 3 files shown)


test/inductor/test_mkldnn_pattern_matcher.py
```diff
@@ -4133,26 +4133,30 @@ def forward(self, x, weight, scales):
...
-            mod = M(is_permute=is_permute).eval()
-            x = torch.randn(x_shape, dtype=torch.bfloat16).as_strided(x_shape, x_stride)
-            w_shape = (12, 256)
-            w = torch.randint(-128, 127, w_shape, dtype=torch.int8)
-            s = torch.randn(s_shape, dtype=torch.bfloat16)
-
-            def matcher_check_fn():
-                self.assertEqual(
-                    counters["inductor"]["woq_matcher_count"], 0 if TEST_ACL else 1
+        devices = ["cpu", "xpu"]
```
etaf: Hi, this will fail on device without xpu.
I suggest you move this case to `TestPatternMatcherGeneric`, which is already designed to be reused by both cpu and xpu.

etaf (changes requested):

test/inductor/test_mkldnn_pattern_matcher.py
xiaowangintel: Gets it, I have modified.

etaf (approved):

test/inductor/test_mkldnn_pattern_matcher.py
EikanWang, line 1: Does this PR need to change this test file?

xiaowangintel: Yes， https://github.com/pytorch/pytorch/blob/70ce9561a0e1f5a8fd2f9950e870bcb82e9509e2/test/inductor/test_mkldnn_pattern_matcher.py#L4781 the original test is only for cpu, without xpu.

EikanWang (approved):

desertfire (approved):

jansel (approved):

PR #163616, opened 2025-09-23
Update doc for torch.index_select, which can accept IntTensor as a parameter

catswe:
Description said "entries in index which is a LongTensor" but index_select can accept an IntTensor as the parameter

### DIFF
--- a/torch/_torch_docs.py
+++ b/torch/_torch_docs.py
@@ -5308,7 +5308,7 @@ add_docstr(
 index_select(input, dim, index, *, out=None) -> Tensor
 
 Returns a new tensor which indexes the :attr:`input` tensor along dimension
-:attr:`dim` using the entries in :attr:`index` which is either an `IntTensor` or a `LongTensor`.
+:attr:`dim` using the entries in :attr:`index`.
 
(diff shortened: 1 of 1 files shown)


torch/_torch_docs.py
```diff
@@ -5308,7 +5308,7 @@ def merge_dicts(*dicts):
 index_select(input, dim, index, *, out=None) -> Tensor
 
 Returns a new tensor which indexes the :attr:`input` tensor along dimension
-:attr:`dim` using the entries in :attr:`index` which is a `LongTensor`.
+:attr:`dim` using the entries in :attr:`index` which is either an `IntTensor` or a `LongTensor`.
```
jbschlosser:
```suggestion
:attr:`dim` using the entries in :attr:`index`.
```

jbschlosser (commented):
why not just remove the reference here and let the arg docs do the talking :)

torch/_torch_docs.py
catswe: Makes sense. Thank you!

jbschlosser (approved):

PR #161062, opened 2025-08-20
[ContextParallel] add `_LoadBalancer` classes, and load-balance interface to Context Parallel APIs

XilunWu:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #163806
* #163053
* #163617
* __->__ #161062

**Summary**
This PR provides an interface for users to specify how to load-balance the attention
input. The load-balance is essentially a rearrangement of the input tensor(s) over the
seq_dim before sharding and can be specified via an index tensor `rearrange` such
that Q[rearrange] is the balanced Q users want (i.e. `rearrange[i] == j` where `i` is the new
index of `Q[j]` in the balanced Q). An example is the `_generate_round_robin_indices()` added
in https://github.com/pytorch/pytorch/pull/155442. 

**New `_LoadBalancer` classes**
New `_LoadBalancer` class (defined in `torch/distributed/tensor/experimental/_load_balancer.py`)
provides one interface for defining load-balance behavior: `_generate_indices(self, restore: bool = False)`.

When `restore == False`, this method should output an index Tensor (namely `rearrange_idx`) such
that QKV will be transformed into Q' K' V' in a way that `Q'[i] == Q[rearrange_idx[i]]` (same applies
to K and V).

When `restore == True`, this method outputs an index Tensor (namely `restore_idx` such that
`Q'[restore_idx] == Q` (same applies to K and V).

**Impact**
2 public CP APIs and 1 private CP API is modified. This PR should be backward-compatible by:
- For uses w/ SDPA, existing users must be using the `context_parallel()` API which does not
take in the extra `load_balancer` argument and solely determines from the global var
`_cp_options.enable_load_balance`.
- For new users including who want to try `flex_attention()`, we require to use the new API
`_context_parallel_buffers` to explicitly shard the QKV input instead of using `context_parallel()`
because we no longer rely on TorchDispatchMode nor TorchFunctionMode for op replacement. And
we also require users to explicitly pass in a `load_balancer` argument if load-balancing is demanded.

**Load-Balance Behavior**
`context_parallel_unshard()`, and `create_cp_block_mask()` APIs now take an extra optional argument
`load_balancer`. This argument is optional because of backward compatibility but we require new users
to explicitly pass in a `load_balancer` if load-balancing is demanded:
- if `load_balancer == None` and `_cp_options.enable_load_balance == False`, CP performs
no load-balancing on input Tensors.
- if `load_balancer == None` and `_cp_options.enable_load_balance ==True`, CP performs
[...]

PR #163617, opened 2025-09-23
[ContextParallel] add process-time based Round-Robin load-balance to CP

XilunWu:
* #165222
* #164919
* #163806
* #163053
* __->__ #163617

**Summary**
The load-balancing problem can be modeled as [identical-machines scheduling](https://en.wikipedia.org/wiki/Identical-machines_scheduling) problem. We already provided an easy-to-extend interface in #161062 for
implementing load-balancing and in this PR we start with adding a Round-Robin solution as an example
and also a verification. This can be easily adapted to other solutions like Shortest-processing-time-first/
Longest-processing-time-first with extra padding added for collectives.

- Added a new type of `_LoadBalancer` implementation `_PTRRLoadBalancer` which is designed for
`flex_attention()`. This load-balance strategy analyzes the `BlockMask` sparsity info and perform
Round-Robin (unlike traditional Round-Robin doing it in circular order, we do in zig-zag order). 
- Make `_context_parallel_buffers` and `context_parallel_unshard` handle batched load-balance
index (previously it can only handle non-batched load-balance index), like in `create_cp_block_mask`.

**Test**
`pytest test/distributed/tensor/test_attention.py`


torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1177,12 +1177,29 @@ def _context_parallel_buffers(
...
+                buffer_batch_size = buffer.size(dim=0)
+
+                if lb_indices_batch_size == buffer_batch_size:
+                    for i in range(buffer_batch_size):
+                        # NOTE: assuming batch dim is 0
+                        buffer_batch_i = torch.index_select(
+                            buffer[i], dim=seq_dim - 1, index=load_balance_indices[i]
+                        )
+                        buffer[i] = buffer_batch_i
+                elif lb_indices_batch_size == 1:
```
drisspg: can we just put the 0 index if condition into the main loop and not have two?

torch/distributed/tensor/experimental/_load_balancer.py
```diff
@@ -288,6 +291,169 @@ def _generate_indices_for_batch(self, seq_length_per_doc, restore) -> torch.Tens
...
+        block_mask = self.block_mask
+        kv_num_blocks = block_mask.kv_num_blocks
+        full_kv_num_blocks = block_mask.full_kv_num_blocks
+        non_sparse_kv_num_blocks = (
+            kv_num_blocks + full_kv_num_blocks
+            if full_kv_num_blocks is not None
+            else kv_num_blocks
+        )
+        B, H, Q = non_sparse_kv_num_blocks.shape
+        # assumption: the masking is identical across heads
```
drisspg: we shouldnt assert but we should make this very visible in the docs

torch/distributed/tensor/experimental/_load_balancer.py
```diff
@@ -288,6 +291,169 @@ def _generate_indices_for_batch(self, seq_length_per_doc, restore) -> torch.Tens
...
+        tasks_in_group = tasks_in_group.view(-1, group_size).transpose(
+            0, 1
+        )  # (group_size, n // group_size)
+
+        # sort each group. This step should not have impact on correctness
+        # nor execution run time, but it helps users visualize the mask
+        tasks_in_group, _ = torch.sort(tasks_in_group, dim=1)
+        return tasks_in_group
+
+    def _generate_indices(self, restore: bool = False) -> Tensor:
```
drisspg, line 378: Very pretty :)

we dont have a DtoH sync here right?

XilunWu: no, all operations are tensor ops and happen on `block_mask`'s device.

test/distributed/tensor/test_attention.py
fegin: We don't need this, we can just extend load_balance_type to `["_PerDocumentHeadTailLoadBalancer", "_PTRRLoadBalancer", None]`

test/distributed/tensor/test_attention.py
```diff
@@ -660,13 +666,26 @@ def test_cp_flex_attention_document_mask(self) -> None:
...
             # generate load balancer
-            load_balancer = (
-                _PerDocumentHeadTailLoadBalancer(
-                    lengths, self.world_size, torch.device(self.device_type)
-                )
-                if enable_load_balance
-                else None
-            )
+            load_balancer = None  # no load-balance
+            if enable_load_balance:
```
fegin: We can remove this as mentioned above.

test/distributed/tensor/test_attention.py
```diff
@@ -660,13 +666,26 @@ def test_cp_flex_attention_document_mask(self) -> None:
...
+                        create_block_mask(
+                            document_causal_mask,
+                            B=batch_size,
+                            H=1,
+                            Q_LEN=max_seq_len,
+                            KV_LEN=max_seq_len,
+                        ),
+                        self.world_size,
+                        self.device_type,
+                    )
```
fegin: We should move `load_balancer` creation to `test_cp_flex_attention`. Then we don't need to create two global mask and it also gives us a better view of how the UX is.

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1074,21 +1074,28 @@ def _context_parallel_buffers(
...
-                    )
-                else:
-                    # load_balance_indices has shape (batch_size, seq_length)
-                    # TODO: this for-looop can be done in a smarter way
-                    for i in range(load_balance_indices.size(dim=0)):
-                        # NOTE: assuming batch dim is 0
+                # NOTE: assuming batch dim is 0
+                idx_batch_size = load_balance_indices.size(0)
+                data_batch_size = buffer.size(0)
+                assert idx_batch_size == 1 or idx_batch_size == data_batch_size, (
```
fegin: ValueError may be a better excpetion.

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1074,21 +1074,28 @@ def _context_parallel_buffers(
...
+                    "Cannot rearrange buffer: "
+                    f"load_balance_indices has shape {load_balance_indices.shape}, "
+                    f"but buffer has shape {buffer.shape}."
+                )
+
+                for i in range(data_batch_size):
```
fegin, line 1082: We can do this to simplify the code:

```
index = load_balance_indices[0] if idx_batch_size == 1 else load_balance_indices[i]
```

You don't need a branch after this branch.

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1585,19 +1592,26 @@ def context_parallel_unshard(
...
+            # NOTE: assuming batch dim is 0
+            idx_batch_size = restore_indices.size(0)
+            data_batch_size = unsharded_b.size(0)
+            assert idx_batch_size == 1 or idx_batch_size == data_batch_size, (
+                "Cannot restore buffer: "
+                f"restore_indices has shape {restore_indices.shape}, "
+                f"but unsharded_b has shape {unsharded_b.shape}."
+            )
+
+            for i in range(data_batch_size):
```
fegin, line 1600: ditto. We can simplify the code by getting the index first.

torch/distributed/tensor/experimental/_load_balancer.py
```diff
@@ -301,6 +304,178 @@ def _generate_indices_for_batch(self, seq_length_per_doc, restore) -> torch.Tens
...
+class _PTRRLoadBalancer(_LoadBalancer):
+    """
+    Processing-Time based Round-Robin (PTRR) load balancer.
+    """
+
+    def __init__(
+        self,
+        block_mask: BlockMask,
+        world_size: int,
+        device: Union[str, torch.device],
```
fegin: We are on Python 3.10 and above, | is the recommended way instead of Union.

torch/distributed/tensor/experimental/_load_balancer.py
```diff
@@ -301,6 +304,178 @@ def _generate_indices_for_batch(self, seq_length_per_doc, restore) -> torch.Tens
...
+                load balance indices that rearrange original positions to PTRR pattern.
+
+            Returns:
+                The generated indices of shape `(1, seq_len)` if the load-balancing is
+                identical within the batch (i.e. `BlockMask.shape[0] == 1`), or
+                `(batch_size, seq_len)` if the load-balancing should vary within the batch.
+
+        Warning:
+            For Multi-Head Attention, we require the masks over the head dimension are identical
+            (i.e. `self.block_mask` must have shape (B, 1, seq_len, seq_len) or (1, 1, seq_len, seq_len)).
```
fegin, line 395: We should add the check in `__init__()`.

torch/distributed/tensor/experimental/_load_balancer.py
fegin: Should we remove the `device` requirement from `_PTRRLoadBalancer`? The device must be the same as `block_mask`, which we can infer automatically. Also, this is the only load balancer that cannot be used with SDPA, we should mention this.

torch/distributed/tensor/experimental/_load_balancer.py
```diff
@@ -301,6 +304,178 @@ def _generate_indices_for_batch(self, seq_length_per_doc, restore) -> torch.Tens
...
+            "for now only support q_blk_size == kv_blk_size"
+        )
+
+        indices = torch.arange(
+            q_blk_size * ptrr_indices.size(1), device=ptrr_indices.device
+        ).view(-1, q_blk_size)  # (NUM_BLOCKS, BLOCK_SIZE)
+        indices = indices[ptrr_indices].view(B, -1)  # (B, qkv_size)
+
+        if restore:
+            indices = torch.vmap(torch.argsort)(indices)
```
fegin, line 474: I'm thinking that should we put the logic to a separate function? The main reason is that I am worried about the performance indication and am thinking if we should compile the code.

fegin (approved):
LGTM, please address the comments before merging the PR.

torch/distributed/tensor/experimental/_load_balancer.py
XilunWu: Yes, simply add `@torch.compile(fullgraph=True)` to the function definition. But we'll need to remove "raise" and "assert" since they would break graph when hit.

PR #163619, opened 2025-09-23
Add `inference_mode` hint message to use `eval` with inference.

### LINKED ISSUE #162923: Method `.inference_mode()` can be misleading
### 📚 The doc issue

Hi

The name of [inference_mode](https://docs.pytorch.org/docs/stable/generated/torch.autograd.grad_mode.inference_mode.html) can be confusing.
At first glance, the name suggests that it might also handle tasks commonly associated with inference, such as disabling dropout or switching batch normalization layers to evaluation mode. While the documentation does not mention such behavior, this could easily lead to confusion for users (as it did for me).

### Suggest a potential alternative/fix

The easiest solution would be to add a clear note in the documentation stating that, for actual inference, the model should also be set to evaluation mode using `.eval()`. This would help prevent users from assuming that `inference_mode` handles those responsibilities automatically.

cc @svekars @sekyondaMeta @AlannaBurke @albanD

zeshengzong:
Fixes #162923

### Before

image

### After

image

### DIFF
--- a/torch/autograd/grad_mode.py
+++ b/torch/autograd/grad_mode.py
@@ -237,6 +237,12 @@ class inference_mode(_DecoratorContextManager):
        Unlike some other mechanisms that locally enable or disable grad,
        entering inference_mode also disables :ref:`forward-mode AD <forward-mode-ad>`.
 
+    .. warning::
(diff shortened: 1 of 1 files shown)


jbschlosser (approved):
Looks great, thanks for the update!

PR #163621, opened 2025-09-23
[Quant] extend the op list for quant lift up

Valentine233:
Add `aten.reshape.default` into the op list of quant lift up, in order to fuse more potential quantized kernels.

### DIFF
--- a/test/inductor/test_mkldnn_pattern_matcher.py
+++ b/test/inductor/test_mkldnn_pattern_matcher.py
@@ -4703,7 +4703,6 @@ class TestDynamicPatternMatcher(TestPatternMatcherBase):
                 self.k_proj = torch.nn.Linear(input_dim, input_dim, bias=False)
                 self.v_proj = torch.nn.Linear(input_dim, input_dim, bias=False)
                 self.softmax = torch.nn.Softmax(dim=-1)
-                self.transpose_for_score = transpose_for_score
                 self.num_attention_heads = num_attention_heads
                 self.attention_head_size = attention_head_size
                 self.all_head_size = self.num_attention_heads * self.attention_head_size
--- a/test/inductor/test_mkldnn_pattern_matcher.py
+++ b/test/inductor/test_mkldnn_pattern_matcher.py
@@ -4750,7 +4750,7 @@ class TestDynamicPatternMatcher(TestPatternMatcherBase):
                 )
                 self.assertEqual(
                     counters["inductor"]["quant_lift_up_count"],
-                    4 if annotate_matmul and not TEST_ACL else 1,
+                    4 if (annotate_matmul and not TEST_ACL and IS_X86) else 1,
                 )
 
             quantizer = X86InductorQuantizer()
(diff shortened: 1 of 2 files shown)


test/inductor/test_mkldnn_pattern_matcher.py
```diff
@@ -4723,32 +4720,38 @@ def forward(self, x):
...
-                    counters["inductor"]["qlinear_weight_prepack_matcher_count"], 3
+                    counters["inductor"]["qlinear_weight_prepack_matcher_count"], 4
                 )
                 self.assertEqual(
                     counters["inductor"]["qlinear_unary_matcher_count"],
                     3 if annotate_matmul and not TEST_ACL else 0,
                 )
+                self.assertEqual(
+                    counters["inductor"]["quant_lift_up_count"],
+                    4 if annotate_matmul and not TEST_ACL else 1,
```
Valentine233: The original count is `3` instead of `4`.

test/inductor/test_mkldnn_pattern_matcher.py
```diff
@@ -4694,7 +4694,6 @@ class SelfAttnLikeModule(torch.nn.Module):
             def __init__(
                 self,
                 input_dim,
-                transpose_for_score=False,
```
Xia-Weiwen, line 4697: Why is this always True now?

mingfeima (approved):
LGTM. Just address weiwen's comment will be fine.

test/inductor/test_mkldnn_pattern_matcher.py
```diff
@@ -4723,32 +4720,38 @@ def forward(self, x):
...
-                transpose_for_score=True,
                 num_attention_heads=16,
                 attention_head_size=64,
             ).eval()
             v = torch.randn(2, 384, 1024)
 
             def matcher_check_fn():
                 self.assertEqual(
-                    counters["inductor"]["qlinear_weight_prepack_matcher_count"], 3
+                    counters["inductor"]["qlinear_weight_prepack_matcher_count"], 4
```
Xia-Weiwen, line 4745: Where is the additional qlinear_weight_prepack from?

test/inductor/test_mkldnn_pattern_matcher.py
Valentine233: In the original UT, `transpose_for_score` is always True: https://github.com/pytorch/pytorch/blob/main/test/inductor/test_mkldnn_pattern_matcher.py#L4738. So I remove the flag for simplicity. In addition, there is no need to test `transpose_for_score=False` because we always have this in the model.

test/inductor/test_mkldnn_pattern_matcher.py
Valentine233: This is from the new added `torch.nn.Linear`: `self.dense`, which would become a `qlinear`.

Xia-Weiwen (approved):

Valentine233:
@jansel @ezyang Could you please help review the PR?

jansel (approved):

PR #163623, opened 2025-09-23
Remove dataclass_slots

cyyever:
`dataclass` now has `slots` kwarg.


torch/_dynamo/bytecode_transformation.py
```diff
@@ -39,8 +38,7 @@
     from .output_graph import DynamoTracerOutput
 
 
-@dataclass_slots
-@dataclasses.dataclass
+@dataclasses.dataclass(slots=True)
```
Skylion007, line 41: Mypy's min Python version needs to be updated first

cyyever: I just noticed that @malfet had updated it.

Skylion007 (approved):

PR #163632, opened 2025-09-23
[3/N] Use std::filesystem in inductor

cyyever:
Continued work to use std::fs in inductor.

### DIFF
--- a/torch/csrc/inductor/aoti_torch/shim_common.cpp
+++ b/torch/csrc/inductor/aoti_torch/shim_common.cpp
@@ -28,6 +28,7 @@
 #include <c10/core/Device.h>
 #include <c10/core/DeviceGuard.h>
 #include <c10/core/Stream.h>
+#include <c10/util/FileSystem.h>
 
 #ifndef AT_PER_OPERATOR_HEADERS
 #include <ATen/Functions.h>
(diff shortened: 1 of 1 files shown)


torch/csrc/inductor/aoti_torch/shim_common.cpp
```diff
@@ -1248,21 +1199,22 @@ void aoti_torch_save_tensor_handle(
...
+    if (!c10::filesystem::create_directories(tmp_folder, ec)) {
       std::cout << "aoti_torch_save_tensor_handle: Error creating directory: "
-                << tmp_folder << '\n';
+                << tmp_folder << " error:" << ec.message() << '\n';
       return;
     }
   }
-  std::string tensor_filepath_to_save = tmp_folder + launch_prefix + "_" +
-      kernel_name + "_" + tensor_name + "_" + t->device().str() + ".pt";
+  std::string tensor_filepath_to_save = tmp_folder.string() + launch_prefix +
```
Skylion007, line 1216: Would be way cleaner with fmtlib

Skylion007 (approved):

torch/csrc/inductor/aoti_torch/shim_common.cpp
cyyever: It should be, but let's format it later to avoid introducing additional bugs.

PR #163634, opened 2025-09-23
Move TypeTraits/TypeList/Metaprogramming.h to header-only

lw:
* #163635
* __->__ #163634


torch/headeronly/util/Metaprogramming.h
```diff
@@ -0,0 +1,224 @@
...
+}
+} // namespace detail
+
+template <class Mapper, class... Args>
+auto tuple_map(std::tuple<Args...>&& tuple, const Mapper& mapper) {
+  return detail::tuple_map(
+      std::move(tuple), mapper, std::index_sequence_for<Args...>());
+}
+
+} // namespace c10::guts
```
janeyx99, line 224: Can you add a torch::headeronly namespace exposing the APIs you will be using? (These should match the ones tested)

This way we can better guarantee that everything in torch::headeronly is truly headeronly.

janeyx99 (commented):
Also I believe we can delete the corresponding .cpp files for these

janeyx99:
This is now done in #167386, closing!

PR #163636, opened 2025-09-23
[AOTI] [XPU] add null-pointer-deference check for `malloc` in `sycl_runtime_wrappers.h`

shaoyuyoung:
Fixes #163624
I use `TORCH_CHECK` because I think this `npd` issue can be caused by users when they try to allocate too large memory

### DIFF
--- a/torch/csrc/inductor/aoti_runtime/sycl_runtime_wrappers.h
+++ b/torch/csrc/inductor/aoti_runtime/sycl_runtime_wrappers.h
@@ -1,6 +1,7 @@
 // NOLINT
 #pragma once
 #ifdef USE_XPU
+#include <c10/util/Exception.h>
 #include <c10/xpu/XPUFunctions.h>
 #include <level_zero/ze_api.h>
 #include <sycl/sycl.hpp>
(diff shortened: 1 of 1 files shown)


torch/csrc/inductor/aoti_runtime/sycl_runtime_wrappers.h
```diff
@@ -46,6 +47,7 @@ static ze_module_handle_t _createModule(
     size_t szLog = 0;
     ZE_CHECK(zeModuleBuildLogGetString(buildLog, &szLog, nullptr));
     char* strLog = (char*)malloc(szLog);
+    TORCH_CHECK(strLog != nullptr, "Failed to allocate log buffer, size=", szLog);
```
jbschlosser: `TORCH_INTERNAL_ASSERT` seems more appropriate here than `TORCH_CHECK` as the latter is more for validation of user inputs, etc.

jbschlosser (commented):
thanks for the update :) one minor note

torch/csrc/inductor/aoti_runtime/sycl_runtime_wrappers.h
shaoyuyoung: thanks! have updated

shaoyuyoung:
hi, any update?

PR #163639, opened 2025-09-23
Revert to old behaviour of not padding strides if shape or stride is dynamic

nandesuka:
Differential Revision: D83053287

### DIFF
--- a/torch/_inductor/ir.py
+++ b/torch/_inductor/ir.py
@@ -3741,6 +3741,15 @@ class Layout(OutputSpec):
         ):
             return in_strides
 
+        # Skip padding the strides for dynamic shapes based on config.pad_dynamic_shape
+        # Checking both shape and strides, as there are cases where only one is dynamic
+        is_dynamic = not all(
+            isinstance(s, (int, sympy.Integer))
+            for s in itertools.chain(in_strides, size)
+        )
+        if not config.pad_dynamic_shapes and is_dynamic:
+            return in_strides
+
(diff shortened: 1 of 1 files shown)


blaine-rister (approved):
LGTM. Left a nit about cleaning up some related code.

torch/_inductor/ir.py
```diff
@@ -3740,6 +3740,15 @@ def _pad_strides(
...
         ):
             return in_strides
+        
+        # Skip padding the strides for dynamic shapes based on config.pad_dynamic_shape
+        # Checking both shape and strides, as there are cases where only one is dynamic
+        is_dynamic = not all(
+            isinstance(s, (int, sympy.Integer))
+            for s in itertools.chain(in_strides, size)
+        )
+        if not config.pad_dynamic_shapes and is_dynamic:
```
blaine-rister, line 3750: nit: is it safe to remove the `or config.pad_dynamic_shapes` on line 3784? Now that we're returning early, that might be redundant.

blaine-rister (approved):
LGTM. Left a nit about possibly removing a redundant check further down in the function.

torch/_inductor/ir.py
nandesuka: let's just leave it for now, can clean up later once we have a root cause

laithsakka: are we padding with out guards if we remove this?
shall we instead make sure we guard properly .

nandesuka: Can you clarify, guard what values with what constraints?

PR #163642, opened 2025-09-23
Use cuda nvrtc so file based on cuda version used by torch

### LINKED ISSUE #162367: libnvrtc.so failure in TestCompileKernel.test_compile_kernel

atalman:
Fixes https://github.com/pytorch/pytorch/issues/162367

cc @msaroufim @tinglvv @nWEIdia

### DIFF
--- a/torch/cuda/_utils.py
+++ b/torch/cuda/_utils.py
@@ -78,7 +78,7 @@ def _get_nvrtc_library() -> ctypes.CDLL:
         ]
     else:
         nvrtc_libs = [
-            "libnvrtc.so.{major_version}",
+            f"libnvrtc.so.{major_version}",
             "libnvrtc.so",  # Fallback to unversioned
         ]
     for lib_name in nvrtc_libs:
--- a/torch/cuda/_utils.py
+++ b/torch/cuda/_utils.py
@@ -71,10 +71,22 @@ def _get_hiprtc_library() -> ctypes.CDLL:
 
 
 def _get_nvrtc_library() -> ctypes.CDLL:
+    major_version = int(torch.version.cuda.split(".")[0])
     if sys.platform == "win32":
-        return ctypes.CDLL("nvrtc64_120_0.dll")
+        nvrtc_libs = [
+            f"nvrtc64_{major_version}0_0.dll",
+        ]
     else:
-        return ctypes.CDLL("libnvrtc.so")
+        nvrtc_libs = [
+            "libnvrtc.so.{major_version}",
+            "libnvrtc.so",  # Fallback to unversioned
+        ]
+    for lib_name in nvrtc_libs:
+        try:
+            return ctypes.CDLL(lib_name)
+        except OSError:
+            continue
+    raise OSError("Could not find any NVRTC library")
 
 
 def _get_gpu_rtc_library() -> ctypes.CDLL:
--- a/torch/cuda/_utils.py
+++ b/torch/cuda/_utils.py
@@ -71,7 +71,7 @@ def _get_hiprtc_library() -> ctypes.CDLL:
 
 
 def _get_nvrtc_library() -> ctypes.CDLL:
-    major_version = int(torch.version.cuda.split(".")[0])
+    major_version = int(torch.version.cuda.split(".")[0])  # type: ignore[union-attr]
     if sys.platform == "win32":
         nvrtc_libs = [
             f"nvrtc64_{major_version}0_0.dll",


msaroufim (approved):

torch/cuda/_utils.py
```diff
@@ -71,10 +71,22 @@ def _get_hiprtc_library() -> ctypes.CDLL:
 
 
 def _get_nvrtc_library() -> ctypes.CDLL:
+    major_version = int(torch.version.cuda.split(".")[0])  # type: ignore[union-attr]
```
Skylion007, line 74: Just use a torch.utils._typing_utils not_none here instead of a type ignore

torch/cuda/_utils.py
```diff
@@ -71,10 +71,22 @@ def _get_hiprtc_library() -> ctypes.CDLL:
...
+        nvrtc_libs = [
+            f"libnvrtc.so.{major_version}",
+            "libnvrtc.so",  # Fallback to unversioned
+        ]
+    for lib_name in nvrtc_libs:
+        try:
+            return ctypes.CDLL(lib_name)
+        except OSError:
+            continue
+    raise OSError("Could not find any NVRTC library")
```
Skylion007, line 89: Why not store the last OSError in an optional and chain from there to preserve the trace?

torch/cuda/_utils.py
```diff
@@ -71,10 +71,22 @@ def _get_hiprtc_library() -> ctypes.CDLL:
...
     else:
-        return ctypes.CDLL("libnvrtc.so")
+        nvrtc_libs = [
+            f"libnvrtc.so.{major_version}",
+            "libnvrtc.so",  # Fallback to unversioned
+        ]
+    for lib_name in nvrtc_libs:
+        try:
+            return ctypes.CDLL(lib_name)
+        except OSError:
```
Skylion007, line 87: Probably should at least be a logger.debug with exc_info=True

torch/cuda/_utils.py
atalman: Tried this, Looks like torch.utils._typing_utils not_none creates circular dependency issue.

PR #163647, opened 2025-09-23
Fix IValue from SymBool on big-endian system

AlekseiNikiforovIBM:
Skip test_compiled_autograd_attribution on s390x

It fails both on s390x and x86_64 at least under some circumstances. Disable it for now until on s390x until it works reliably.

### DIFF
--- a/aten/src/ATen/core/ivalue.h
+++ b/aten/src/ATen/core/ivalue.h
@@ -624,7 +624,14 @@ struct TORCH_API IValue final {
   IValue(const c10::SymBool& i) {
     if (auto mi = i.maybe_as_bool()) {
       tag = Tag::Bool;
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
       payload.u.as_int = *mi;
+#elif __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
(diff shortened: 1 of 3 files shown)


aten/src/ATen/core/ivalue.h
```diff
@@ -624,7 +624,14 @@ struct TORCH_API IValue final {
   IValue(const c10::SymBool& i) {
     if (auto mi = i.maybe_as_bool()) {
       tag = Tag::Bool;
+#if __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__
       payload.u.as_int = *mi;
+#elif __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__
+      /* due to byteorder if value assigned as_int, as_bool actually is not set correctly */
+      payload.u.as_bool = *mi;
```
AlekseiNikiforovIBM, line 631: I didn't just set `as_bool` for every case due to

https://github.com/pytorch/pytorch/blob/39c340ec9e2ee3011f1d260f581b5a95f3c99039/aten/src/ATen/core/ivalue.h#L683-L690

I can remove these checks and just set `as_bool` value for every case if that would be preferred way to fix this code.

malfet (approved):
LGTM

PR #163652, opened 2025-09-23
Allow unbacked to unbacked replacements if rhs unbacked symbols are all inputs

laithsakka:
* __->__ #163652

This partially solve the issue https://github.com/pytorch/pytorch/issues/163641. We do not need to ban unbacked to unbacked replacement if all rhs symbols are inputs since we know those symbols are seen by the whole program.

This issue was found as i was tracing some vllm models with unbacked, namely  Qwen/Qwen2-1.5B-Instruct it makes reasoning logic easier to do those replacements. 

as for data dependent similar pattern, I am thinking to create a set of replacements that we apply only during static eval
instead of none. to make reasoning better. 

### DIFF
--- a/torch/fx/experimental/symbolic_shapes.py
+++ b/torch/fx/experimental/symbolic_shapes.py
@@ -5490,6 +5490,10 @@ class ShapeEnv:
         def track_symint(
             source: Source, val: IntLikeType, constraint: DimConstraint = None
         ) -> None:
+            # We should not generate guards for unbacked inputs.
+            if isinstance(val, SymInt) and self.is_unbacked_symint(val.node.expr):
+                return
+
             log.debug("track_symint %s %s %s", LazyString(source.name), val, constraint)
             assert not isinstance(val, SymInt) or is_symbolic(val)
 
--- a/torch/fx/experimental/symbolic_shapes.py
+++ b/torch/fx/experimental/symbolic_shapes.py
@@ -973,7 +973,7 @@ def free_unbacked_symbols(x: IterateExprs) -> OrderedSet[sympy.Symbol]:
     )
 
 
-def free_non_source_unbacked_symbols(
+def _free_non_source_unbacked_symbols(
     x: IterateExprs, unbacked_inputs: OrderedSet[sympy.Symbol]
 ) -> OrderedSet[sympy.Symbol]:
     """Unbacked symbols that are not inputs to the graph. These are symbols that originated from
@@ -6950,7 +6950,7 @@ class ShapeEnv:
                     if isinstance(lhs, sympy.Symbol):
                         if free_unbacked_symbols(
                             lhs
-                        ) and not free_non_source_unbacked_symbols(
+                        ) and not _free_non_source_unbacked_symbols(
                             rhs, self.unbacked_inputs
                         ):
                             return True
--- a/torch/fx/experimental/symbolic_shapes.py
+++ b/torch/fx/experimental/symbolic_shapes.py
@@ -5492,11 +5492,7 @@ class ShapeEnv:
             source: Source, val: IntLikeType, constraint: DimConstraint = None
         ) -> None:
             # We should not generate guards for unbacked inputs.
-            if (
-                isinstance(val, SymInt)
-                and isinstance(val.node.expr, sympy.Symbol)
-                and self.is_unbacked_symint(val.node.expr)
-            ):
+            if isinstance(val, SymInt) and has_free_unbacked_symbols(val.node.expr):
                 return
 
             log.debug("track_symint %s %s %s", LazyString(source.name), val, constraint)
--- a/torch/fx/experimental/symbolic_shapes.py
+++ b/torch/fx/experimental/symbolic_shapes.py
@@ -4939,8 +4939,6 @@ class ShapeEnv:
         symbolic_context: Optional[StatelessSymbolicContext] = None,
     ) -> sympy.Expr:
         """Create a new symbol which is tracked by this ShapeEnv"""
-        # import fbvscode
-        # fbvscode.set_trace()
         # check if constraint_dim is actually static integer
         if (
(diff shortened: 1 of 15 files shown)


torch/fx/experimental/symbolic_shapes.py
```diff
@@ -3717,6 +3727,8 @@ def _init(
         self.var_to_range_sloc: dict[sympy.Symbol, ValueRangesSLoc] = {}
         self.source_name_to_debug_name: dict[str, str] = {}
         self.var_to_sources: dict[sympy.Symbol, list[Source]] = {}
+        # A set of unabcked symbols that are inputs (i.e: not data dependent).
```
aorenste: nit: spelling

torch/fx/experimental/symbolic_shapes.py
```diff
@@ -4851,6 +4863,8 @@ def create_unbacked_symint(self, source: Optional[Source] = None) -> SymInt:
         self._log_create_unbacked_symbol(
             "create_unbacked_symint", symbol, vr, source, sym_node=sym_node
         )
+        if "u96" in str(symbol):
```
aorenste: debugging cruft?

torch/fx/experimental/symbolic_shapes.py
```diff
@@ -6900,7 +6920,10 @@ def _smart_symbol_sort(x: sympy.Symbol) -> tuple[int, int, str]:
                 ):
                     raise NotImplementedError
 
-                # Never replace unbacked symbols with other unbacked symbols.
+                # Never replace unbacked symbols that with other unbacked symbols that are
```
aorenste: nit: extra word 'that'

torch/fx/experimental/symbolic_shapes.py
```diff
@@ -973,6 +973,16 @@ def free_unbacked_symbols(x: IterateExprs) -> OrderedSet[sympy.Symbol]:
     )
 
 
+def free_non_source_unbacked_symbols(
```
aorenste:
```suggestion
def _free_non_source_unbacked_symbols(
```

torch/fx/experimental/symbolic_shapes.py
```diff
@@ -4969,6 +4983,11 @@ def create_symbol(
         if dynamic_dim in (DimDynamic.SIZE_LIKE_UNBACKED, DimDynamic.OBLIVIOUS_SIZE):
             out = self.create_unbacked_symint(source).node.expr
             self._constrain_range_for_size(out)
+            from torch._dynamo.source import LocalSource
+
+            if isinstance(source.base, LocalSource) and source.base.is_input:
```
laithsakka: ok this is NOT the right way to tell that this is an unbacked input.

laithsakka: or maybe it is let me double check

ezyang:
I haven't reviewed the code but the high level reasoning LGTM

torch/fx/experimental/symbolic_shapes.py
```diff
@@ -4913,8 +4924,18 @@ def create_unspecified_symbol(
...
+        category=FutureWarning,
+    )
+    @record_shapeenv_event()
+    def create_symbol(self, *args: Any, **kwargs: Any) -> sympy.Expr:
+        """Create a new symbol which is tracked by this ShapeEnv"""
+        return self.create_non_data_dependent_symbol(*args, **kwargs)
+
     @record_shapeenv_event()
-    def create_symbol(
+    def create_non_data_dependent_symbol(
```
aorenste: What's the long term naming scheme? We have backed, unbacked, now non_data_dependent...

torch/fx/experimental/symbolic_shapes.py
```diff
@@ -973,6 +973,16 @@ def free_unbacked_symbols(x: IterateExprs) -> OrderedSet[sympy.Symbol]:
     )
 
 
+def _free_non_source_unbacked_symbols(
```
aorenste, line 976: nit (maybe future PR): I know this is named to match "free_unbacked_symbols" but maybe we should call it something like "_get_free..." or "_compute_free..." - it's not obvious from the name whether it's a getter or an action (free the non-source unbacked symbols) (and then maybe rename "free_unbacked_symbols" too)

aorenste (commented):
Code looks reasonable but some tests are failing

torch/fx/experimental/symbolic_shapes.py
laithsakka: mm i mean there is no much places were we distinguish those

so we have
backed : non_data_dependent
unbacked
  - inputs: non_data_dependent
  - non inputs: data dependent. (item, tolist ..etc)
  
for the  purpose of creating symbols this function was named create_symbol, which is very general. This is really meant for non_data_dependent symbols.

bobrenjc93: This seems so overly complicated. Can you explain why we can't just have create_backed_symbol and create_unbacked_symbol where unbacked are implicitly data dependent?

If the only concern is around the input striding, I'd much prefer banning on non-contiguous striding on unbacked vs all of this additional complexity around backed symbols, unbacked symbols, data dependent symbols, non data depdenent symbols...

laithsakka: we already have this API called
create_symbol now, i do not want people to use it to create non input symbols it will fail
does that make sense?
just renamed it to create_non_data_dependent_symbol , is create_graph_input_symbol better? i think that would be confusing.

I could have kept it create_symbol but its not safer to rename.

bobrenjc93: Discussed offline - laith will split out naming changes in a separate PR since we want to take the time to get that right.

bobrenjc93 (approved):

PR #163653, opened 2025-09-23
[BE] Delete all pre py-3.10 checks

malfet:
* __->__ #163653

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci @EikanWang @jgong5 @wenzhe-nrv @sanchitintel @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela


jansel (approved):

torch/_dynamo/bytecode_transformation.py
```diff
@@ -577,7 +537,7 @@ def linetable_310_writer(
     See https://github.com/python/cpython/blob/main/Objects/lnotab_notes.txt
     This is the internal format of the line number table for Python 3.10
     """
-    assert sys.version_info >= (3, 10) and sys.version_info < (3, 11)
+    assert sys.version_info < (3, 11)
```
Skylion007: Sigh, I'd keep the lower bound here for readability but make it one compound comparison

torch/_inductor/tiling_utils.py
```diff
@@ -373,19 +371,8 @@ def try_split(self, pw: Split, red: Split) -> Optional[tuple[Split, Split]]:
...
-    def zip_equal(it1: Iterable[T], it2: Iterable[U]) -> Iterator[tuple[T, U]]:
-        """
-        Zip two iterables, raising ValueError if their lengths differ.
-        """
-        if len(it1) != len(it2):
-            raise ValueError(f"Lengths differ: {len(it1)} != {len(it2)}")
-        return zip(it1, it2)
-
+# On Python 3.10+ we can use zip(strict=True)
+zip_equal = functools.partial(zip, strict=True)
```
Skylion007: We should enable the zip strict ruff rule codebase wide gradually with a todo suppression now. It's the best code quality feature of 3.10

torch/_inductor/utils.py
```diff
@@ -3335,12 +3335,7 @@ def __delitem__(self, key: KeyType) -> None:
 @dataclass_transform(frozen_default=True)
```
Skylion007, line 3335: Actually this whole function should be replaced with a functools.partial on dataclass, would be easier to type through and less error prone. It's just dataclass with no default kwargs...

torch/utils/_config_module.py
```diff
@@ -178,10 +177,7 @@ def visit(
         prefix: str,
     ) -> None:
         """Walk the module structure and move everything to module._config"""
-        if sys.version_info[:2] < (3, 10):
-            type_hints = getattr(source, "__annotations__", {})
-        else:
-            type_hints = inspect.get_annotations(source)
+        type_hints = inspect.get_annotations(source)
```
Skylion007, line 180: We should just backport the 3.14 features too so it can read the newer type_extension stuff. Replace with typing_extension.get_annotations . This should have been the polyfill here btw

torch/_inductor/utils.py
Skylion007: Also now that we support 3.10, slots should also equal True, no reason not to use it really if we already require frozen

torch/_jit_internal.py
```diff
@@ -1257,12 +1250,9 @@ def _get_named_tuple_properties(
         defaults = []
     # In 3.10 recommended way to get annotations is to call `inspect.get_annotations` function
     # Also, annotations from base class are not inherited so they need to be queried explicitly
-    if sys.version_info[:2] < (3, 10):
-        obj_annotations = getattr(obj, "__annotations__", {})
-    else:
-        obj_annotations = inspect.get_annotations(obj)
-        if len(obj_annotations) == 0 and hasattr(obj, "__base__"):
-            obj_annotations = inspect.get_annotations(obj.__base__)
+    obj_annotations = inspect.get_annotations(obj)
```
Skylion007, line 1253: Using typing_extensions backport here too

torch/__init__.py
```diff
@@ -2819,10 +2819,7 @@ def _import_device_backends():
     from importlib.metadata import entry_points
 
     group_name = "torch.backends"
-    if sys.version_info < (3, 10):
```
Skylion007, line 2822: Mypy's config will need its min Python version updated too. Another reason to have it read from pyproject.toml

torch/_inductor/tiling_utils.py
malfet: Yes, this will be done as followup PR (B905 suppression)

torch/utils/_config_module.py
malfet: Again, this change purely addresses ruff violations and tries not to intro any code

torch/_dynamo/bytecode_transformation.py
malfet: Hmm, let me change it to `sys.version_info[:2] == (3, 10)` which would be a correct one

PR #163654, opened 2025-09-23
[AMP] Add deprecated decorator for torch.xxx.amp.autocast class

fffrog:
* __->__ #163654

As the title stated.

**Changes:**
- torch.cuda.amp.autocast
- torch.cpu.amp.autocast
- add explicit `__new__` and `__init_subclass__` for those class above for inspect.signature to retrieve correct signature


Skylion007 (approved):

Flamefire:
This leads to numerous errors in at least the tests on Python 3.13

The `__new__` method wrapper added by `@deprecated` causes it to be picked up by the JIT passes which then fails in 2 places:
1. https://github.com/pytorch/pytorch/blob/24cdfbdd4ce3d3b080e64e6a8b03a8ecda547bd0/torch/_jit_internal.py#L593 doesn't detect it as builtin anymore which makes `inspect.getsource` in `inspect.getsource` fail
2. It now appears in `cls.__dict__` so it is not filtered anymore at https://github.com/pytorch/pytorch/blob/f605bd7dcc16ff97328db88b1005a9aecff2fed4/torch/jit/frontend.py#L263 which makes it fail in `get_source_lines_and_files` on calling `inspect.getsourcefile(obj)`

Reproducible by running `python test/test_jit_autocast.py TestAutocast.test_callees` on a CUDA build


Is there no CI with Python 3.13?

PR #163656, opened 2025-09-23
Speedup __instancecheck__ by moving VariableTrackerMeta to C++

guilhermeleobas:
* __->__ #163656

Also, caches the result of `__instancecheck__`

Reduces tracing time of test_ziplongest by 0.7 seconds.

```
                              Comparison Results
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Benchmark                    ┃ Baseline Time ┃ New Time ┃ Speedup/Slowdown ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ TestBasicOps.test_ziplongest │ 45.5060s      │ 44.8077s │            0.98x │
└──────────────────────────────┴───────────────┴──────────┴──────────────────┘
```

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @kadeng @chauhang @amjames @Lucaskabela @jataylo


jansel (changes requested):
Failing tests?

guilhermeleobas:
@jansel Should be good now. It was a memory leak and a missing decref.

torch/csrc/dynamo/init.cpp
```diff
@@ -36,6 +36,11 @@ using torch::dynamo::autograd::torch_c_dynamo_compiled_autograd_init;
 
 namespace {
 
+std::map<std::pair<PyTypeObject*, PyTypeObject*>, bool> _isinstance_cache;
+std::vector<PyObject*> _all_subclasses;
+PyTypeObject* VT_tp = nullptr;
+PyTypeObject* LazyVT_tp = nullptr;
```
jansel, line 33: Are these owned pointers?  My main worry is we get the recounting right.  Could we use something like `THPObjectPtr` to ensure we don't have pointers to dead objects?

torch/csrc/dynamo/init.cpp
```diff
@@ -152,6 +157,135 @@ struct IsValidVarName {
   }
 };
 
+PyObject* _fast_isinstance_check(PyObject* cls, PyObject* obj) {
+  if (LazyVT_tp == nullptr || VT_tp == nullptr) {
+    PyObject* base_mod = PyImport_ImportModule("torch._dynamo.variables.base");
```
jansel: Throughout the PR lets use THPObjectPtr for RAII ref-counting correctness.

jansel (changes requested):

torch/csrc/dynamo/init.cpp
guilhermeleobas: Only the last two are owned. The pointers inside `_isinstance_cache` and `_all_subclasses` are weak references. I'll update the code to use THPObjectPtr.

PR #163660, opened 2025-09-23
[inductor] fix issue for example value with unbacked strides

sevenEng:
## Issue

During autotune, we're not applying size hints atomically for the example inputs used for benchmarking.

If there is unbacked symint showing up in inputs' strides, this might lead to CUDA IMA,

and this could be reproduced by the added unittest, with stride being `[128 * u0, 128, 1]` and unbacked fallback being 8192, after calling `benchmark_example_value`, we get back a tensor with stride as `[8192, 128, 1]` as opposed to `[128 * 8192, 128, 1]`

## Fix

Using the atomic API when trying to apply size hints to input tensor' strides.

### DIFF
--- a/test/inductor/test_unbacked_symints.py
+++ b/test/inductor/test_unbacked_symints.py
@@ -653,6 +653,28 @@ class TestUnbackedSymints(InductorTestCase):
         expected = fn(*example_inputs)
         torch.testing.assert_close(actual, expected)
 
+    @skipGPUIf(not HAS_GPU, "requires gpu and triton")
+    @inductor_config.patch({"max_autotune": True})
+    @dynamo_config.patch({"capture_scalar_outputs": True})
+    def test_autotune_with_unbacked_stride(self, device):
+        def fn(x, y, a):
+            u0 = a.item()
+            torch._check(u0 != 1)
+            unbacked = x.expand(8, u0, *x.shape).clone()
+            unbacked = torch.permute(unbacked, [0, 2, 1])
+            y = y.expand(8, *y.shape)
+            bmm = torch.ops.aten.bmm(unbacked, y)
+            return bmm
+
+        example_inputs = (
+            torch.randn((32,), dtype=torch.bfloat16, device=device),
+            torch.randn((128, 64), dtype=torch.bfloat16, device=device),
+            torch.tensor(128, device=device),
+        )
+        actual = torch.compile(fn, fullgraph=True)(*example_inputs)
+        expected = fn(*example_inputs)
+        torch.testing.assert_close(actual, expected)
+
 
 instantiate_device_type_tests(TestUnbackedSymints, globals(), allow_xpu=True)
 
--- a/torch/_inductor/select_algorithm.py
+++ b/torch/_inductor/select_algorithm.py
@@ -3622,10 +3622,12 @@ class AlgorithmSelectorCache(PersistentCache):
                 fallback=config.unbacked_symint_fallback,
                 hint_override=hint_override,
             ),
-            V.graph.sizevars.size_hints(
-                node.get_stride(),
-                fallback=config.unbacked_symint_fallback,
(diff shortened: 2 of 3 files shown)


test/inductor/test_unbacked_symints.py
```diff
@@ -607,6 +607,30 @@ def fn(x):
         expected = fn(*example_inputs)
         torch.testing.assert_close(actual, expected)
 
+    @skipGPUIf(not HAS_GPU, "requires gpu and triton")
+    @inductor_config.patch({"max_autotune": True})
+    @dynamo_config.patch({"capture_scalar_outputs": True})
+    def test_autotune_with_unbacked_stride(self, device):
+        if device == "cpu":
+            raise unittest.SkipTest("Only testing on GPU")
```
ColinPeppler: nit: this unittest should work for CPU too

torch/_inductor/select_algorithm.py
```diff
@@ -3620,9 +3622,12 @@ def key_of(node):
...
-            *sizevars.size_hints(
-                node.get_stride(),
-                fallback=config.unbacked_symint_fallback,
+            tuple(
+                V.graph.sizevars.atomically_apply_size_hint(
+                    stride,
+                    fallback=config.unbacked_symint_fallback,
+                )
+                for stride in node.get_stride()
             ),
```
ColinPeppler: Should keep the tuple unpacking

```suggestion
            *tuple(
                V.graph.sizevars.atomically_apply_size_hint(
                    stride,
                    fallback=config.unbacked_symint_fallback,
                )
                for stride in node.get_stride()
            ),
```

ColinPeppler:
Dumping some thoughts here, since this would be used for torch.compile GEMM autotuning too:
1. If we graph break, do the sub-graphs share a single SizeVarAllocator?
2. If we recompile, do both graphs share a single SizeVarAllocator?

Why do I care about `SizeVarAllocator` so much? We use it to cache unbacked substitutions when calling `atomically_apply_size_hint`.

https://github.com/pytorch/pytorch/blob/1495b35d29512f303ab37780760c5e692158514b/torch/_inductor/sizevars.py#L750-L769

I believe the answer to both (1) and (2) is Yes.

AFAIK everytime we run Inductor, we need to create a brand new `GraphLowering`.
https://github.com/pytorch/pytorch/blob/1495b35d29512f303ab37780760c5e692158514b/torch/_inductor/compile_fx.py#L1391

When we create `GraphLowering` we also create `SizeVarAllocator`.
https://github.com/pytorch/pytorch/blob/1495b35d29512f303ab37780760c5e692158514b/torch/_inductor/graph.py#L375

test/inductor/test_unbacked_symints.py
sevenEng: let me check what happens for the cpu test

sevenEng: test for cpu also passed :D

torch/_inductor/select_algorithm.py
sevenEng: thanks for catching this!

torch/_inductor/select_algorithm.py
```diff
@@ -3582,10 +3582,12 @@ def benchmark_example_value(node, hint_override: Optional[int] = None):
                 fallback=config.unbacked_symint_fallback,
                 hint_override=hint_override,
             ),
-            V.graph.sizevars.size_hints(
-                node.get_stride(),
-                fallback=config.unbacked_symint_fallback,
-                hint_override=hint_override,
```
ColinPeppler, line 3628: Yeah, we should definitely keep `hint_override` here.

We should either make atomically_apply_size_hint available/an option in size_hints or the other way around.

cc: @bobrenjc93 @pianpwk @ezyang  for any thoughts

ColinPeppler (approved):

PR #163665, opened 2025-09-23
Fake process group Direct construction error

### LINKED ISSUE #162129: Direct construction of FakeProcessGroup doesn't raise errors but has silent incorrectness
### 🐛 Describe the bug

Repro:

```
import torch
from torch.utils._python_dispatch import TorchDispatchMode
from torch.distributed._distributed_c10d import FakeProcessGroup


class SimpleTensorMode(TorchDispatchMode):
    def __torch_dispatch__(self, func, types, args=(), kwargs=None):
        print(func)
        if kwargs is None:
            kwargs = {}
        return func(*args, **kwargs)


fake_pg = FakeProcessGroup(rank=0, world_size=3)
with SimpleTensorMode():
    tensor = torch.tensor([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]])
    dist.all_reduce(tensor, group=fake_pg)
```

This prints 

```
aten.lift_fresh.default
```

But it's expected to also print c10d.allreduce_.default

If you replace fake_pg with a proper initialized PG ala

```
        fake_store = FakeStore()
        torch.distributed.init_process_group(
            "fake", store=fake_store, rank=0, world_size=3
        )
```

it works.

We should error more loudly when you use FakeProcessGroup directly in this way.

### Versions

main

ahkush:
Fixes #162129. Added validation in _rank_not_in_group() to check if ```FakeProcessGroup``` is properly initialized before use, raising a clear error message if ```torch.distributed.init_process_group(backend='fake')``` hasn't been called first. 
This prevents silent failures and ensures proper dispatch system integration for all distributed operations. 

Added test case test_fake_process_group_direct_usage_error() that validates the error is raised for ```all_reduce``` and ```all_to_all_single``` operations. 

Please let me know if additional distributed operators should be tested or if any other updates are needed.

### DIFF
--- a/torch/csrc/distributed/c10d/FakeProcessGroup.hpp
+++ b/torch/csrc/distributed/c10d/FakeProcessGroup.hpp
@@ -28,11 +28,15 @@ class FakeProcessGroup : public Backend {
     bool error_on_collective = false;
   };
 
-  FakeProcessGroup(
+  // Static factory method for official APIs
+  static c10::intrusive_ptr<FakeProcessGroup> _create_internal(
       int rank,
       int size,
-      c10::intrusive_ptr<Options> options = c10::make_intrusive<Options>())
-      : Backend(rank, size), options_(std::move(options)) {}
+      c10::intrusive_ptr<Options> options = c10::make_intrusive<Options>()) {
+    return c10::intrusive_ptr<FakeProcessGroup>(
+        new FakeProcessGroup(rank, size, std::move(options)),
+        c10::raw::DontIncreaseRefcount{});
+  }
 
   const std::string getBackendName() const override {
     return "fake";
@@ -235,6 +239,9 @@ class FakeProcessGroup : public Backend {
   }
 
  private:
+  // Private constructor used by official APIs
+  FakeProcessGroup(int rank, int size, c10::intrusive_ptr<Options> options)
+      : Backend(rank, size), options_(std::move(options)) {}
   c10::intrusive_ptr<Options> options_;
 
   void checkCollectiveError() {
--- a/torch/distributed/fsdp/_flat_param.py
+++ b/torch/distributed/fsdp/_flat_param.py
@@ -543,7 +543,7 @@ class FlatParamHandle:
         self._device_handle = _FSDPDeviceHandle.from_device(self.device)
         self.process_group = process_group
         if self._use_fake_all_gather or self._use_fake_reduce:
-            self._fake_process_group = FakeProcessGroup(
+            self._fake_process_group = FakeProcessGroup._create_internal(
                 rank=process_group.rank(), world_size=process_group.size()
             )
         self.rank = process_group.rank()
--- a/torch/testing/_internal/distributed/fake_pg.py
+++ b/torch/testing/_internal/distributed/fake_pg.py
@@ -22,7 +22,7 @@ def _create_fake_pg(common_opts, backend_opts):
     for every collective. It should be used as a convenient tool when playing
     with distributed but don't care about the actual data.
     """
-    return FakeProcessGroup(
+    return FakeProcessGroup._create_internal(
         common_opts.group_rank, common_opts.group_size, backend_opts
     )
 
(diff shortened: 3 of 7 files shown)


ezyang:
The test feels too late to me. Why can't you discover something bad happened earlier?

ezyang:
Yeah, this is cure is worse than the disease, I think.

What if we blocked direct construction of FakeProcessGroup entirely? Instead, the "official" APIs would have to do some private API that gets around this blockage.

torch/testing/_internal/distributed/fake_pg.py
```diff
@@ -22,7 +22,7 @@ def _create_fake_pg(common_opts, backend_opts):
     for every collective. It should be used as a convenient tool when playing
     with distributed but don't care about the actual data.
     """
-    return FakeProcessGroup(
+    return FakeProcessGroup._create_internal(
         common_opts.group_rank, common_opts.group_size, backend_opts
     )
```
kwen2501, line 27: Looks like a bc break?

kwen2501 (commented):
I think a solution to the issue filed by @ezyang  is for `dist.all_reduce` to directly call into `torch.ops.c10d.all_reduce_`, instead of `pg.all_reduce`.

ahkush:
@kwen2501 
While that would ensure dispatch integration, it wouldn't solve the core issue. The dispatch system requires process groups to be registered in the GroupRegistry (via ```resolve_process_group()```), but directly constructed ```FakeProcessGroup``` instances are never registered. So ```torch.ops.c10d.all_reduce_``` would fail with "Could not resolve the process group" for direct constructions. 

Also, should users be able to construct ```FakeProcessGroup``` directly at all, or is it better to guide them toward the official ```init_process_group(backend='fake')``` API for proper integration?

Do you have any suggestions for a better approach that would address the dispatch integration issue while handling the registration requirement? I'd appreciate your thoughts on this.

torch/testing/_internal/distributed/fake_pg.py
ahkush: The ```_create_fake_pg()``` function itself has no BC break - same signature, same behavior, still returns a ```FakeProcessGroup```. Only direct ```FakeProcessGroup()``` construction breaks (intentionally), which gets a clear error message directing users to the proper API. Internal utilities and type checking continue working unchanged.

torch/distributed/fsdp/_flat_param.py
```diff
@@ -543,7 +543,7 @@ def __init__(
         self._device_handle = _FSDPDeviceHandle.from_device(self.device)
         self.process_group = process_group
         if self._use_fake_all_gather or self._use_fake_reduce:
-            self._fake_process_group = FakeProcessGroup(
+            self._fake_process_group = FakeProcessGroup._create_internal(
```
ezyang, line 546: I'm actually kind of skeptical about this use site, it sort of feels like potentially this is buggy LOL

torch/csrc/distributed/c10d/FakeProcessGroup.hpp
```diff
@@ -26,11 +26,15 @@ class FakeProcessGroup : public Backend {
...
 
-  FakeProcessGroup(
+  // Static factory method for official APIs
+  static c10::intrusive_ptr<FakeProcessGroup> _create_internal(
       int rank,
       int size,
-      c10::intrusive_ptr<Options> options = c10::make_intrusive<Options>())
-      : Backend(rank, size), options_(std::move(options)) {}
+      c10::intrusive_ptr<Options> options = c10::make_intrusive<Options>()) {
+    return c10::intrusive_ptr<FakeProcessGroup>(
```
ezyang, line 36: nit: make_intrusive_ptr

torch/csrc/distributed/c10d/FakeProcessGroup.hpp
```diff
@@ -214,6 +218,9 @@ class FakeProcessGroup : public Backend {
   }
 
  private:
+  // Private constructor used by official APIs
+  FakeProcessGroup(int rank, int size, c10::intrusive_ptr<Options> options)
+      : Backend(rank, size), options_(std::move(options)) {}
```
ezyang, line 244: I don't think it is as important to hide the ctor on the C++ side

ezyang (approved):

PR #163671, opened 2025-09-23
Fix double dispatch to Python for detach

### LINKED ISSUE #71725: When someone calls detach() on a __torch_dispatch__ object, detach() gets called twice
### 🐛 Describe the bug

See https://gist.github.com/zou3519/6a673f7b624ca2365361e45ce7843439 .
```
with capture_logs() as logs:
    x = LoggingTensor(torch.tensor([3.0], requires_grad=True))
    log_input("x", x)
    x.detach()
print(logs)
```
prints:
```
["$0 = input('x')", '$1 = torch._ops.aten.detach($0)', '$2 = torch._ops.aten.detach($1)']
```
on PyTorch 1.10 and PyTorch master

### Versions

1.10 and master

cc @albanD @zou3519

swolchok:
* __->__ #163671

This fixes #71725.

Differential Revision: [D83857880](https://our.internmc.facebook.com/intern/diff/D83857880)


torch/csrc/autograd/VariableTypeManual.cpp
```diff
@@ -450,6 +450,17 @@ static const Tensor& resize_as_(
 static Tensor detach(c10::DispatchKeySet ks, const Tensor& self) {
   auto out = ([&]() {
     at::AutoDispatchBelowADInplaceOrView guard;
+    // HACK: avoid doubling cost if we're going to dispatch to Python,
```
ezyang: yes, dog science. let me think if I have a better idea for you

ezyang: So, one problem with doing this is detach() will never get called at all if base.is_inference(), and this is undesirable:

```
  // Note [View of inference tensor]
  // For inference tensor this code can only be hit outside InferenceMode
  // since ADInplaceOrView is in the default_included_set.
  // If Inplace and View were separate dispatch keys we can just put Inplace
  // in the default_included_set, so that view ops on inference tensor doesn't
  // have to go through as_view even outside InferenceMode.
  if (base.is_inference())
    return tensor;
```

ezyang: The key operation we do in as_view is this:

```
// See NOTE [ Autograd View Variables ] for details.
// Non-differentiable view. Just share version counter.
inline Variable make_variable_non_differentiable_view(
    const Variable& base,
    const at::Tensor& data,
    bool allow_tensor_metadata_change = true) {
  if (data.defined()) {
    // Currently all of non-differentiable view ops(detach/_indices/_values)
    // share the same TensorImpl as their base Tensor. Thus a new TensorImpl
    // allocation here is required.
    auto data_impl_copy = data.getIntrusivePtr()->shallow_copy_and_detach(
        /*version_counter=*/impl::version_counter(base),
        /*allow_tensor_metadata_change=*/allow_tensor_metadata_change);
    data_impl_copy->set_autograd_meta(nullptr);
    return Variable(data_impl_copy);
  }
  return Variable();
}
```

The payload for detach() in normal code is:

```
Tensor detach(const Tensor& self) {
  // NB: detach() is not the same thing as alias()! The main difference is that
  // detach does not allow metadata change while alias does.
  return Tensor(self.getIntrusivePtr()->shallow_copy_and_detach(
      // NB: The ADInplaceOrView logic will overwrite these with the
      // appropriate values if it runs; otherwise these are the values.
      /*version_counter=*/0,
      /*allow_tensor_metadata_change=*/false));
}
```

But shallow_copy_and_detach on subclass does this:

```
template <typename VariableVersion>
c10::intrusive_ptr<TensorImpl> TensorImpl::shallow_copy_and_detach_core(
    VariableVersion&& version_counter,
    bool allow_tensor_metadata_change) const {
  c10::intrusive_ptr<TensorImpl> r;
  const auto mode_stack_len = c10::impl::TorchDispatchModeTLS::stack_len();
  // TODO: do we have to exclude after Python dispatch key set?
  if (mode_stack_len > 0 &&
      !c10::impl::tls_is_dispatch_key_excluded(DispatchKey::Python)) {
    const auto& cur_torch_dispatch_mode_state =
        c10::impl::TorchDispatchModeTLS::get_stack_at(mode_stack_len - 1);
    r = cur_torch_dispatch_mode_state->pyinterpreter()->detach(this);
  } else if (
      key_set_.has(DispatchKey::Python) &&
      !c10::impl::tls_is_dispatch_key_excluded(DispatchKey::Python)) {
    r = (pyobj_slot_.load_pyobj_interpreter())->detach(this);
  }
  if (r) {
    if (!r->is_inference()) {
      r->set_version_counter(std::forward<VariableVersion>(version_counter));
    }
    r->set_allow_tensor_metadata_change(allow_tensor_metadata_change);
    return r;
  }
  // otherwise just copy the TensorImpl and not the PyObject.  Since
  // the interpreter is dead no one can call us out on it
  auto impl = c10::make_intrusive<TensorImpl>(
      // No need to populate Storage; copy_tensor_metadata will do it for us.
      key_set_,
      data_type_,
      device_opt_);
  copy_tensor_metadata(
      /*src_impl=*/this,
      /*dest_impl=*/impl.get(),
      /*version_counter=*/std::forward<VariableVersion>(version_counter),
      /*allow_tensor_metadata_change=*/allow_tensor_metadata_change);
  return impl;
}
```

I think my spicy proposal is that we should just NEVER redispatch to the underlying detach. It seems pointless.

@albanD wdyt?

ezyang (approved):
OK with the dog science version too

torch/csrc/autograd/VariableTypeManual.cpp
albanD: Ok from a bit of investigation, my current state of mind is:
- shallow_copy_and_detach() triggering subclass dispatch() is GOOD. We actually want the user to see whenever a new Tensor object is created (the id of what is returned from their torch_dispatch handler should be the same as the id they get back at the caller).
- we do call make_variable() (and thus shallow_copy_and_detach) on all factory function, why no detach there? Because we check refcount lol. If you have a logging mode that puts the return in a global, keeping it alive, then you will see both the factory function AND a detach, which is WRONG! (issue on this will come when I have a minute).
- Not re dispatching is generally BAD, if you look at the history of this file, this is what it used to do. And we changed it. The most obvious problem red flag from checking the dispatch entries for detach() is the Functionalization impl that will now be skipped. And that might be (very) bad for FakeTensor.
- Other views do NOT have this issue because when creating these non-differentiable views, we just smash the existing Tensor object: https://github.com/pytorch/pytorch/blob/4c2c401ccf18da57f7fb68778687322ad20df34f/torch/csrc/autograd/variable.h#L801-L830

My take for detach is that we should rely on the contract that the underlying redispatch did actually return a fresh Tensor and make the as_view not create a new Tensor (and thus no detach).
Given that the ONLY non-differentiable view in our world is detach(), this is as simple as updating https://github.com/pytorch/pytorch/blob/4c2c401ccf18da57f7fb68778687322ad20df34f/torch/csrc/autograd/variable.h#L835-L850 to remove the shallow_copy_and_detach :)

swolchok: > this is as simple as updating ... to remove the shallow_copy_and_detach()

so the only thing it's doing is set the autograd meta to nullptr, right? If we go with this approach, I'm inclined to just directly implement detach rather than going through as_view (and then clean up the special non-differentiable case in as_view) because as_view is special-cased for non-differentiable views anyway. would that be a desirable cleanup from your perspective?

albanD: > so the only thing it's doing is set the autograd meta to nullptr, right?

Yes.

> would that be a desirable cleanup from your perspective?

It is all quite convoluted one way or the other....
But keeping this one next to make_variable_differentiable_view (that has the same assumptions) feels ok to me (and less refactor work for you :D )

c10/core/TensorImpl.cpp
```diff
@@ -500,6 +500,11 @@ c10::intrusive_ptr<TensorImpl> TensorImpl::shallow_copy_and_detach_core(
   c10::intrusive_ptr<TensorImpl> r;
   const auto mode_stack_len = c10::impl::TorchDispatchModeTLS::stack_len();
   // TODO: do we have to exclude after Python dispatch key set?
+
+  // These first two if/else if conditions are duplicated in detach()
```
albanD: This is an old comment that can be removed

torch/csrc/autograd/variable.h
```diff
@@ -837,14 +837,11 @@ inline Variable make_variable_non_differentiable_view(
     const at::Tensor& data,
     bool allow_tensor_metadata_change = true) {
   if (data.defined()) {
-    // Currently all of non-differentiable view ops(detach/_indices/_values)
-    // share the same TensorImpl as their base Tensor. Thus a new TensorImpl
-    // allocation here is required.
-    auto data_impl_copy = data.getIntrusivePtr()->shallow_copy_and_detach(
-        /*version_counter=*/impl::version_counter(base),
-        /*allow_tensor_metadata_change=*/allow_tensor_metadata_change);
```
albanD, line 859: Just like the one above, you need     data_impl->set_allow_tensor_metadata_change(allow_tensor_metadata_change);

This is for the metadata change errors in CI

torch/csrc/autograd/variable.h
```diff
@@ -837,14 +837,11 @@ inline Variable make_variable_non_differentiable_view(
     const at::Tensor& data,
     bool allow_tensor_metadata_change = true) {
   if (data.defined()) {
-    // Currently all of non-differentiable view ops(detach/_indices/_values)
-    // share the same TensorImpl as their base Tensor. Thus a new TensorImpl
-    // allocation here is required.
-    auto data_impl_copy = data.getIntrusivePtr()->shallow_copy_and_detach(
-        /*version_counter=*/impl::version_counter(base),
```
albanD, line 858: You need to set the version counter of the base as the version counter of data as well (this is all the inplace didn't raise errors in CI).

albanD (commented):
.

swolchok:
@albanD looks like your proposal didn't work? https://github.com/pytorch/pytorch/actions/runs/18144312914/job/51645283222?pr=163671#step:25:1353

torch/csrc/autograd/variable.h
```diff
@@ -837,14 +837,15 @@ inline Variable make_variable_non_differentiable_view(
...
-        /*allow_tensor_metadata_change=*/allow_tensor_metadata_change);
-    data_impl_copy->set_autograd_meta(nullptr);
-    return Variable(data_impl_copy);
+    // Currently all of non-differentiable view ops (just detach)
+    // redispatch first, so they should have already allocated a new
+    // TensorImpl. No need to shallow_copy_and_detach here.
+    auto* data_impl = data.unsafeGetTensorImpl();
+    data_impl->set_version_counter(
+        base.unsafeGetTensorImpl()->version_counter());
```
swolchok: changing this to impl::version_counter; sorry about the reading comprehension failure

swolchok:
@swolchok has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D83857880).

swolchok:
new sparse test failure (python test/test_torch.py TestTorchDeviceTypeCPU.test_grad_scaling_unscale_sparse_cpu_float32 assertion failure on new_nnz <= nnz()) is definitely not associated with an ASAN problem, double-checked

torch/csrc/autograd/variable.h
```diff
@@ -837,14 +837,14 @@ inline Variable make_variable_non_differentiable_view(
...
-    // share the same TensorImpl as their base Tensor. Thus a new TensorImpl
-    // allocation here is required.
-    auto data_impl_copy = data.getIntrusivePtr()->shallow_copy_and_detach(
-        /*version_counter=*/impl::version_counter(base),
-        /*allow_tensor_metadata_change=*/allow_tensor_metadata_change);
-    data_impl_copy->set_autograd_meta(nullptr);
-    return Variable(data_impl_copy);
+    // Currently all of non-differentiable view ops (detach/_indices/_values)
+    // redispatch first, so they should have already allocated a new
+    // TensorImpl. No need to shallow_copy_and_detach here.
```
albanD: Bail out option: while these 3 functions do that, others might not.
I'm ok with adding a new flag here: "is_fresh_tensor = false".
And set this to True in VariableTypeManual for detach() to fix only this one.

We can then follow up in other PRs to gradually change this flag for different functions in the codegen until they're all fixed and we can remove it.

torch/csrc/autograd/VariableTypeManual.cpp
```diff
@@ -453,20 +453,18 @@ static Tensor detach(c10::DispatchKeySet ks, const Tensor& self) {
...
-      /* output */ out,
-      /* is_bw_differentiable */ false,
-      /* is_fw_differentiable */ false,
-      /* view_func */ nullptr,
-      /* rev_view_func */ nullptr,
-      /* creation_meta */ CreationMeta::DEFAULT,
-      /*allow_tensor_metadata_change=*/false);
-
-  return result;
+  if (self.is_inference()) {
```
albanD: nit: move this under the comment as this logic is usually part of as_view()

albanD (approved):
SGTM

izaitsevfb:
@pytorchbot revert -m "breaks export tests" -c nosignal

export/test_export_training_ir_to_run_decomp.py::TrainingIRToRunDecompExportNonStrictTestExport::test_tag_ac_export_training_ir_to_decomp_nonstrict [GH job link](https://github.com/pytorch/pytorch/actions/runs/18329076162/job/52202433523) [HUD commit link](https://hud.pytorch.org/pytorch/pytorch/commit/c32118dc3e50505fd285e6e448a90883fce11535)

export/test_export.py::TestExport::test_tag_ac_export [GH job link](https://github.com/pytorch/pytorch/actions/runs/18329076162/job/52202692797) [HUD commit link](https://hud.pytorch.org/pytorch/pytorch/commit/c32118dc3e50505fd285e6e448a90883fce11535)


Looks like the test expectations need to be updated.

swolchok:
looking at HUD, https://github.com/pytorch/pytorch/actions/runs/18484981746/job/52668577498?pr=163671 is a KP from yesterday that was fixed later. Merging with -i.

PR #83558, opened 2022-08-16
Preserve stack trace for backward nodes over AOTAutograd

SherlockNoMad:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #83558

For the following program. 
```
def my_relu(a):
    return a.relu()

def func(a, b):
    a = torch.nn.Linear(10, 10)(a)
    d = torch.square(b)
    d = my_relu(d)
    loss = d.sum()

    return loss

with torchdynamo.optimize("aot_nop"):
    x = torch.rand(10, 10, requires_grad=True)
    y = torch.rand(10, 10, requires_grad=True)
    out = func(x, y)
``` 

It would generate the following fx graph with stack_trace populated in both forward and backward nodes. 
```
def forward(self, primals, tangents):
    primals_1, primals_2, primals_3, primals_4, tangents_1, = fx_pytree.tree_flatten_spec([primals, tangents], self._in_spec)
    t_default = torch.ops.aten.t.default(primals_3);  primals_3 = None
    addmm_default = torch.ops.aten.addmm.default(primals_4, primals_1, t_default);  primals_4 = primals_1 = t_default = None
    pow_tensor_scalar = torch.ops.aten.pow.Tensor_Scalar(primals_2, 2)
    relu_default = torch.ops.aten.relu.default(pow_tensor_scalar);  pow_tensor_scalar = None
    detach_default = torch.ops.aten.detach.default(relu_default)
    sum_default = torch.ops.aten.sum.default(relu_default);  relu_default = None
    is_same_size_default = torch.ops.aten.is_same_size.default(sum_default, tangents_1)
    expand_default = torch.ops.aten.expand.default(tangents_1, [10, 10]);  tangents_1 = None
    detach_default_1 = torch.ops.aten.detach.default(detach_default);  detach_default = None
    threshold_backward_default = torch.ops.aten.threshold_backward.default(expand_default, detach_default_1, 0);  expand_default = detach_default_1 = None
    pow_tensor_scalar_1 = torch.ops.aten.pow.Tensor_Scalar(primals_2, 1.0);  primals_2 = None
    mul_scalar = torch.ops.aten.mul.Scalar(pow_tensor_scalar_1, 2.0);  pow_tensor_scalar_1 = None
    mul_tensor = torch.ops.aten.mul.Tensor(threshold_backward_default, mul_scalar);  threshold_backward_default = mul_scalar = None
[...]

PR #163673, opened 2025-09-23
Preserve user annotation in graph

SherlockNoMad:
```py
import torch
import torch.fx.traceback as fx_traceback
import torch.export


class M(torch.nn.Module):
    def forward(self, x):
            with fx_traceback.annotate({"fdsp_bucket": 0}):
                x = x + 1
            x = x - 2
            with fx_traceback.annotate({"cuda_stream": 2, "fsdp_bucket": 1}):
                x = x * 2
        x = x / 3
        return x

m = M()

with fx_traceback.preserve_node_meta():
    ep = torch.export.export(m, (torch.randn(10),))

for node in ep.graph.nodes:
    if node.op == "call_function":
        print(f"{node.target}, {node.meta.get("custom", {})}")


```

prints 

```Shell
aten.add.Tensor, {'pp_stage': 0, 'fdsp_bucket': 0}
aten.sub.Tensor, {'pp_stage': 0}
aten.mul.Tensor, {'pp_stage': 0, 'cuda_stream': 2, 'fsdp_bucket': 1}
aten.div.Tensor, {}
```

TODOs:
- run_decomposition is failing
- Need to test with the new full graph capture + aot_export_joint apis
- Need to make the annotation propagate through autograd engine to reach the bw nodes. Sample impl here: https://github.com/pytorch/pytorch/pull/83558
- Edward want to restrict the key in custom field to be top-level singleton objects only
- also need to take care of metadata merging when passes are fusing nodes

Thanks @angelayi  for contributing the dynamo fixes.

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1118,7 +1118,6 @@ coverage_ignore_functions = [
     "is_consistent",
     "is_more_precise",
     # torch.fx.traceback
-    "annotate",
     "format_stack",
     "get_current_meta",
     "has_preserved_node_meta",
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -1118,6 +1118,7 @@ coverage_ignore_functions = [
     "is_consistent",
     "is_more_precise",
     # torch.fx.traceback
+    "annotate",
     "format_stack",
     "get_current_meta",
     "has_preserved_node_meta",
--- a/test/export/test_export.py
+++ b/test/export/test_export.py
@@ -15138,6 +15138,9 @@ def forward(self, x):
             test_serdes=True,
         )
 
+    # TODO: following tests should be fixed
+    @testing.expectedFailureTrainingIRToRunDecomp
+    @testing.expectedFailureTrainingIRToRunDecompNonStrict
     def test_preserve_annotation(self):
         class M(torch.nn.Module):
             def forward(self, x):
@@ -15154,8 +15157,6 @@ def forward(self, x):
 
         with fx_traceback.preserve_node_meta():
             ep = export(m, (torch.randn(10),))
-            # TODO: following line is failing
-            # ep = ep.run_decompositions({})
 
         for node in ep.graph.nodes:
             if node.target == torch.ops.aten.add.default:
--- a/test/export/test_export.py
+++ b/test/export/test_export.py
@@ -21,6 +21,7 @@ from unittest.mock import MagicMock, patch
 
 import torch
 import torch._dynamo as torchdynamo
+import torch.fx.traceback as fx_traceback
 import torch.nn.functional as F
 import torch.utils._pytree as pytree
 from functorch.experimental.control_flow import cond, map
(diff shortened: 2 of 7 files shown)


torch/fx/traceback.py
```diff
@@ -240,6 +240,22 @@ def set_stack_trace(stack: list[str]):
...
 
+@contextmanager
+def set_annotation(key, value):
+    global current_meta
+
+    has_old_value = key in current_meta
+    old_value = current_meta.get(key, None)
+
+    try:
+        current_meta[key] = value
```
angelayi: why not throw everything under current_meta["custom"]? then you don't have to deal with marking every key as being preserved

SherlockNoMad: oh, I didn't notice this key. 
sure

test/export/test_export.py
```diff
@@ -16104,6 +16105,41 @@ def forward(self, q, k, v):
...
+                        fx_traceback.set_custom_annotation("fsdp_bucket", 1),
+                    ):
+                        x = x * 2
+                x = x / 3
+                return x
+
+        m = M()
+
+        with fx_traceback.preserve_node_meta():
+            ep = torch.export.export(m, (torch.randn(10),))
```
angelayi: This will also test strict=True/False and retraceability
```suggestion
            ep = export(m, (torch.randn(10),))
```

torch/fx/traceback.py
```diff
@@ -241,6 +241,35 @@ def set_stack_trace(stack: list[str]):
         current_meta["stack_trace"] = "".join(stack)
 
 
+@contextmanager
+def set_custom_annotation(key, value):
```
angelayi: would it be useful for this to just take in a dict?

test/export/test_export.py
```diff
@@ -15137,6 +15138,38 @@ def forward(self, x, y):
...
+                        x = x + 1
+                    x = x - 2
+                    with fx_traceback.annotate({"cuda_stream": 2, "fsdp_bucket": 1}):
+                        x = x * 2
+                x = x / 3
+                return x
+
+        m = M()
+
+        with fx_traceback.preserve_node_meta():
```
angelayi, line 15158: Should we just always turn this on in export? right now I think we only turn it on when exporting a GraphModule (through the fx.Interpreter).

angelayi: also ep.run_decompositions passes if you remove this context

docs/source/conf.py
```diff
@@ -1118,6 +1118,7 @@
     "is_consistent",
     "is_more_precise",
     # torch.fx.traceback
+    "annotate",
```
albanD: It is not allowed to add anything to this file. Either make the function private or document it properly.

albanD (changes requested):

docs/source/conf.py
SherlockNoMad: alright, added doc.

angelayi (approved):

torch/fx/traceback.py
```diff
@@ -241,6 +242,62 @@ def set_stack_trace(stack: list[str]):
         current_meta["stack_trace"] = "".join(stack)
 
 
+@compatibility(is_backward_compatible=False)
+@contextmanager
+def annotate(annotation_dict: dict):
```
fmassa, line 247: Quick question: can we nest `annotate` regions inside each other?

What happens if we do
```python
with annotate({"fsdp_bucket": 0}):
    with annotate({"ac_region": 1}):
        # do something
```

SherlockNoMad: yes, see PR description. 

Node will have both tags in such case.

docs/source/conf.py
SherlockNoMad: https://docs-preview.pytorch.org/pytorch/pytorch/163673/fx.html#module-torch.fx.annotate

doc looks good tome

albanD (approved):
Thanks for updating the website!

PR #163683, opened 2025-09-23
Add scaffolding for aoti_torch_call_dispatcher BC with native ops

mikaylagawarecki:
Part 1 of plan in https://docs.google.com/document/d/1MaX51H5aEQE5XnOlnZIpf9oCYwzGrTWkgBACxNzsmWE/edit?usp=sharing

- Upgrade `aoti_torch_call_dispatcher` to v2 with an `extension_build_version`
- Allow registration of StableIValue stack  --> IValue stack adapters for schema changes

#### Note: This PR does not include a linter that tells the user to add the upgrader if the schema changes, which is an important piece that will be added in a separate PR

* #165285
* #165284
* #165283
* #164332
* __->__ #163683
* #166373
* #164356


torch/csrc/inductor/aoti_torch/c/shim.h
```diff
@@ -537,6 +537,17 @@ AOTI_TORCH_EXPORT AOTITorchError aoti_torch_call_dispatcher(
...
+// Has the same semantic as aoti_torch_call_dispatcher, but takes an
+// additional argument for the extension build version. This is
+// needed for backward compatibility when calling native functions via
+// the dispatcher. The caller should pass in its build version (not target
+// version).
+AOTI_TORCH_EXPORT AOTITorchError aoti_torch_call_dispatcher_v2(
+    const char* opName,
+    const char* overloadName,
+    StableIValue* stack,
+    uint64_t extension_build_version);
```
mikaylagawarecki: @janeyx99 After thinking more about "we need to propagate both build and target versions", I landed on the header should always pass in the TORCH_ABI_VERSION macro 

#### (1) If a new kwarg is added to a native_function
This is not a "BC-breaking" change, but if the version that the extension builds with is old, the ops.h header will not have the kwarg, so we need an upgrader. 

We could have an alternative design (e.g. pass the num args and the schema corresponding to the op in the build version) and read new kwargs off the libtorch schema but 
(-) This would require exposing FunctionSchema to the frontend 
(-) This makes it more tricky to handle case (2) below

I think this design more future proof (e.g. in the unlikely case BC-break changes the return signature, the current design gives us the flexibility to handle that)

#### (2) If the default on a kwarg is flipped
This is a "BC-breaking" change, BUT, we will check TORCH_FEATURE_VERSION to pass in the the correct default for the argument that corresponds to the version that the user targets in the ops.h header


The remaining case for which we might need to know the version that the extension targets is for StableIValue conversions (`from_ivalue`/`to_ivalue` calls). However, I think StableIValues should fall under one of two cases

1. header-only implementation: BC/FC of conversion depends on `extension_build_version` not target_abi
2. shimmed shared/unique_ptr to OpaqueHandle --> target abi version does not matter 

The only case where we might need the ABI version that the extension targets for StableIValue conversions is when we have something like the following. I find it hard to fathom that we would end up in this scenario, so I chose not to add `extension_target_abi_version` as an argument

```cpp
class Foo {
private:
  std::shared_ptr<FooOpaque, DeleterFnPtr> a;
  #if TORCH_FEATURE_VERSION > xyz
  std::shared_ptr<BarOpauqe, DeleterFnPtr> b;
  #endif
}
```

EDIT: Adding a 3 here

#### (3) A new op is added to native_functions.yaml, we have a wrapper in ops.h that calls it via `aoti_torch_call_dispatcher`

The op definition will be gated by the following in ops.h (it's not available to users unless they target a version that is guaranteed to have it).

```
#if TORCH_FEATURE_VERSION > ...
Tensor op(...) {
aoti_torch_call_dispatcher_v2("op", ...)
}
#endif
```

torch/csrc/inductor/aoti_torch/c/shim.h
```diff
@@ -537,6 +537,17 @@ AOTI_TORCH_EXPORT AOTITorchError aoti_torch_call_dispatcher(
     const char* overloadName,
     StableIValue* stack);
 
+// Has the same semantic as aoti_torch_call_dispatcher, but takes an
+// additional argument for the extension build version. This is
+// needed for backward compatibility when calling native functions via
+// the dispatcher. The caller should pass in its build version (not target
+// version).
```
janeyx99:
```suggestion
// the dispatcher. The caller should pass in the libtorch version the extension is
// building with (NOT target version).
```
making sure I understand

torch/csrc/inductor/aoti_torch/c/shim.h
janeyx99: My conclusion from reading this comment is:
- TORCH_FEATURE_VERSION is used in our headers to bar code that calls into shim APIs that do not exist for a certain version.
- TORCH_ABI_VERSION is the "real" status of APIs, so we need to adapt based on reality (the struct or API at the time of build) vs an older version of reality.

In this new framework, we are relying on TORCH_FEATURE_VERSION to be the gatekeeper. Once we're "through the gate", we assume that the APIs we are using are what we have, so we use TORCH_ABI_VERSION to communicate that information.

torch/csrc/inductor/aoti_torch/shim_common.cpp
```diff
@@ -1675,6 +1678,234 @@ AOTITorchError aoti_torch_call_dispatcher(
   });
 }
 
+// Schema Adapter Infrastructure
+// SchemaAdapterRegistry contains the adapters registered via
+// register_schema_adapter that define how to convert the StableIValue argument
+// stack to an IValue stack when changes are made to the schema of a function.
```
janeyx99:
```suggestion
// stack to an IValue stack when changes are made to the schema of an ATen
// function. This should only be relevant in the context of calling
// aoti_torch_call_dispatcher.
```

torch/csrc/inductor/aoti_torch/c/shim.h
mikaylagawarecki: Yes, I think that's right!

Then there's the odd case I mention at the end which I just feel like is "too unrealistic to happen"

test/cpp_extensions/schema_adapter_test_extension/schema_adapter_test/csrc/test_extension.cpp
```diff
@@ -0,0 +1,151 @@
+#include <pybind11/pybind11.h>
+#include <torch/csrc/inductor/aoti_torch/c/shim.h>
+#include <torch/csrc/jit/frontend/tracer.h>
```
janeyx99: Is this extension supposed to be stable? If so, I wouldn't expect to have to depend on pybind11, torch/csrc/jit/frontend/tracer.h and <torch/extension.h>

test/cpp_extensions/schema_adapter_test_extension/schema_adapter_test/csrc/test_extension.cpp
```diff
@@ -0,0 +1,151 @@
...
+STABLE_TORCH_LIBRARY_IMPL(schema_adapter_test, CPU, m) {
+  m.impl("test_schema_upgrader_v1", &boxed_test_schema_upgrader_v1);
+  m.impl("test_schema_upgrader_v2", &boxed_test_schema_upgrader_v2);
+  m.impl("test_schema_upgrader_v3", &boxed_test_schema_upgrader_v3);
+  m.impl("test_schema_upgrader_v4", &boxed_test_schema_upgrader_v4);
+}
+
+// Pybind11 module - test adapters are automatically registered via static
+// initialization
+PYBIND11_MODULE(TORCH_EXTENSION_NAME, m) {
```
janeyx99: do we need this?

torch/csrc/inductor/aoti_torch/shim_common.cpp
```diff
@@ -1672,6 +1675,234 @@ AOTITorchError aoti_torch_call_dispatcher(
...
+
+// Currently this only adapts the argument stack.
+// C++ default argument resolution will happen at compile time in the
+// torch/csrc/stable/ops.h header, so extensions always pass complete argument
+// lists for the version they build against's schema. As such, this is only
+// needed if a new keyword argument is added to the schema
```
janeyx99:
```suggestion
// needed if a new argument is added to the schema
```
This would work for any new argument, yes?

torch/csrc/inductor/aoti_torch/shim_common.cpp
```diff
@@ -1672,6 +1675,234 @@ AOTITorchError aoti_torch_call_dispatcher(
...
+//
+// This is not declared in the stable shim.h,
+// so we **do not make any guarantees that the signature of this will not
+// change**. If there is a need to define similar infrastructure for the returns
+// we can update this.
+
+namespace {
```
janeyx99: Is the anonymous namespace here our way of making the code within it "static"?

torch/csrc/inductor/aoti_torch/shim_common.cpp
```diff
@@ -1672,6 +1675,234 @@ AOTITorchError aoti_torch_call_dispatcher(
...
+        std::string(op_name), applies_to_versions_below, std::move(adapter_fn));
+  });
+}
+
+// Test adapters for _test_schema_upgrader - demonstrating schema adapter
+// functionality
+
+// Version constants for _test_schema_upgrader using PyTorch version encoding
+// Similar to TORCH_ABI_VERSION but for specific schema versions
+#define MAKE_VERSION(major, minor, patch) \
```
janeyx99: should we make this a more public util

torch/csrc/inductor/aoti_torch/shim_common.cpp
```diff
@@ -1672,6 +1675,234 @@ AOTITorchError aoti_torch_call_dispatcher(
...
+        auto arg_type = schema.arguments()[idx].type();
+        torch::jit::push(ivalue_stack, to_ivalue(arg_type, stable_ivalue));
+      }
+    }
+
+    op.callBoxed(ivalue_stack);
+
+    // there should then be num_returns IValues on the stack, which
+    // we will convert to StableIValue and repopulate user input stack
+    for (const auto idx : c10::irange(num_returns)) {
```
janeyx99: I'd anticipate there is a future possibility we may need adapters for returns too. From what I understand, the approach should be the same as with the inputs and thus can be added later without changing any APIs?

janeyx99 (approved):
Thanks! LGTM. According to our in person discussion I also agree we should move out the testing code.

torch/csrc/inductor/aoti_torch/shim_common.cpp
mikaylagawarecki: To capture what we discussed offline, yes, there's a comment above that mentions this

torch/csrc/inductor/aoti_torch/shim_common.cpp
mikaylagawarecki: Will do in a followup since I'm deleting all the testing code from this PR

torch/csrc/inductor/aoti_torch/shim_common.cpp
mikaylagawarecki: Yep!

PR #163694, opened 2025-09-23
[MPS] [Sparse] unique_dim and sparse broadcast

Isalia20:
Implements unique_dim, sparse broadcast ops and adds dtypes for mps for tests where we expect to fail, otherwise they would always fail due to being run in double precision


test/test_sparse.py
```diff
@@ -456,6 +456,7 @@ def test_ctor_size_checks(self, device, dtype):
     @expectedFailureMPS
     @coalescedonoff
     @dtypes(torch.double)
```
Skylion007, line 458: Normal version doesn't support torch float32? Or is it just that's tested by optest somewhere else?

test/test_sparse.py
```diff
@@ -1626,6 +1635,7 @@ def test_shape(di, dj, dk, nnz):
     @expectedFailureMPS
```
Skylion007, line 1627: Remove expectedFailureMPS wrappers from these tests

test/test_sparse.py
```diff
@@ -3950,6 +3973,7 @@ def check_empty(sparse_shape, nnz, dense_shape, coalesce):
     @coalescedonoff
     @expectedFailureMPS
```
Skylion007, line 3952: A bunch of these need to be removed

test/test_sparse.py
Isalia20: most of the sparse functions on other devices(CPU/CUDA) are all tested in double, not sure why. Maybe it's due to gradcheck being imprecise in float32

test/test_sparse.py
Isalia20: why? We expect that test to fail

test/test_sparse.py
Isalia20: I think no? why should we remove

Skylion007: Because it's no longer an expectedFailure on MPS?

Isalia20: We expect this to fail. I added dtypesIfMPS to the test so expectFailureMPS triggers only when there’s an unexpected success. Without this, it would always fail because all tests using this decorator run in torch.float64, which always errors on MPS regardless of whether the op is implemented.

malfet: This is a good change, but do you mind submitting this one as part of separate PR?

aten/src/ATen/native/mps/operations/Unique.mm
```diff
@@ -305,4 +316,88 @@ static void runUniqueGraph(UniqueCachedGraph* uniqueGraph,
...
 }
 
+static Tensor lexsort_rows_perm_mps(const Tensor& mat_2d) {
+  const auto rows = mat_2d.size(0), cols = mat_2d.size(1);
+  if (rows <= 1 || cols == 0) {
+    return arange(rows, mat_2d.options().dtype(kLong));
+  }
+
+  Tensor perm;
+  for (int64_t c = cols - 1; c >= 0; --c) {
```
malfet: Nit
```suggestion
  for (auto c = cols - 1; c >= 0; --c) {
```

aten/src/ATen/native/mps/operations/Unique.mm
```diff
@@ -305,4 +316,88 @@ static void runUniqueGraph(UniqueCachedGraph* uniqueGraph,
...
+    auto keys = mat_2d.select(1, c);
+    if (perm.defined()) {
+      keys = keys.index_select(0, perm);
+    }
+    Tensor idx = argsort(keys, /*dim=*/0, /*descending=*/false);
```
malfet:
```suggestion
    const auto idx = argsort(keys, /*dim=*/0, /*descending=*/false);
```

aten/src/ATen/native/mps/operations/Unique.mm
```diff
@@ -305,4 +316,88 @@ static void runUniqueGraph(UniqueCachedGraph* uniqueGraph,
   return _unique_impl_mps(self, return_inverse, return_counts, false, std::nullopt);
 }
 
+static Tensor lexsort_rows_perm_mps(const Tensor& mat_2d) {
+  const auto rows = mat_2d.size(0), cols = mat_2d.size(1);
```
malfet, line 320: Aren't c++17 allows something like
```suggestion
   const auto [rows, cols] = mat_2d.sizes();
```

aten/src/ATen/native/mps/operations/Unique.mm
```diff
@@ -305,4 +316,88 @@ static void runUniqueGraph(UniqueCachedGraph* uniqueGraph,
...
+  }
+
+  Tensor perm;
+  for (int64_t c = cols - 1; c >= 0; --c) {
+    auto keys = mat_2d.select(1, c);
+    if (perm.defined()) {
+      keys = keys.index_select(0, perm);
+    }
+    Tensor idx = argsort(keys, /*dim=*/0, /*descending=*/false);
+    perm = perm.defined() ? perm.index_select(0, idx) : std::move(idx);
```
malfet:
```suggestion
    if (!perm.defined()) {
      perm = std::move(keys);
      continue;
    }

    keys = keys.index_select(0, perm);
    const auto idx = argsort(keys, /*dim=*/0, /*descending=*/false);
    perm =  perm.index_select(0, idx);
```

aten/src/ATen/native/mps/operations/Unique.mm
```diff
@@ -305,4 +316,88 @@ static void runUniqueGraph(UniqueCachedGraph* uniqueGraph,
...
+                                                                     bool return_counts) {
+  dim = maybe_wrap_dim(dim, self.dim());
+
+  auto sizes = self.sizes().vec();
+  auto num_zero_dims = std::count(sizes.begin(), sizes.end(), (int64_t)0);
+  if (self.size(dim) == 0) {
+    auto output = at::empty(sizes, self.options());
+    auto inverse_indices = at::empty({0}, self.options().dtype(kLong));
+    auto counts = at::empty({0}, self.options().dtype(kLong));
+    return std::make_tuple(output, inverse_indices, counts);
```
malfet: Nit
```suggestion
    return {output, inverse_indices, counts};
```

aten/src/ATen/native/mps/operations/Unique.mm
Isalia20: updated

aten/src/ATen/native/mps/operations/Unique.mm
Isalia20: updated

aten/src/ATen/native/mps/operations/Unique.mm
Isalia20: not supported from .sizes() I think, getting error when trying to compile:
```
/pytorch/aten/src/ATen/native/mps/operations/Unique.mm:320:15: error: cannot decompose private member 'Data' of 'c10::ArrayRef<long long>'
    320 |   const auto [rows, cols] = mat_2d.sizes();
```

aten/src/ATen/native/mps/operations/Unique.mm
Isalia20: rewrote in a simpler way

aten/src/ATen/native/mps/operations/Unique.mm
Isalia20: updated

test/test_sparse.py
Isalia20: removed it for this PR, will submit in a separate one

malfet (approved):

PR #163697, opened 2025-09-23
[NFC] fixed mistake in comment

aartbik:
I used "floor" instead of "ceil", so fix it. Also fixed other typo.


jcaip (approved):
lgtm, thanks for the fix!

PR #163705, opened 2025-09-23
Do not generate guards on unbacked SymInt inputs

laithsakka:
* __->__ #163705

We should not generate guards on unbacked symbols, i found that we do some recompilations 
with my changes in [](https://github.com/pytorch/pytorch/pull/163652) and also found the following error messages
on current state of pytorch when running the program with the graph below.

TODO:
Still figuring out what is the right approach here. 
We need to know what guards are side effects of torch checks and which ones are things that did not go through a torch 
check yet. (like striding relations...)

### DIFF
--- a/torch/fx/experimental/symbolic_shapes.py
+++ b/torch/fx/experimental/symbolic_shapes.py
@@ -5525,7 +5525,7 @@ class ShapeEnv:
 
                 # We do not want to guard on unbacked symbols. All replacements and
                 # bounds should have originated from torch._checks, so runtime
-                # assertions should be there to covor those guards.
+                # assertions should be there to cover those guards.
                 if not has_free_unbacked_symbols(s):
                     input_guards.append((source, s))
             else:
@@ -5852,7 +5852,7 @@ class ShapeEnv:
         for symbol, sources in symbol_to_source.items():
             # We do not want to guard on unbacked symbols. All replacements and
             # bounds should have originated from torch._checks, so runtime
-            # assertions should be there to covor those guards.
+            # assertions should be there to cover those guards.
             if self.is_unbacked_symint(symbol):
                 continue
 
--- a/torch/fx/experimental/symbolic_shapes.py
+++ b/torch/fx/experimental/symbolic_shapes.py
@@ -5246,7 +5246,8 @@ class ShapeEnv:
         default.
         """
         self.log.info("produce_guards")
-
+        # import fbvscode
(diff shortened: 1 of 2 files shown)


laithsakka:
add unit test

ezyang:
Don't expect test on the C++ guard printout, I expect it to be unstable.

I'm not convinced. I think you still need to generate runtime asserts for these "guards", no?

torch/fx/experimental/symbolic_shapes.py
```diff
@@ -5467,6 +5467,10 @@ def get_expression(tensor_dim_src: Source) -> sympy.Expr:
         def track_symint(
             source: Source, val: IntLikeType, constraint: DimConstraint = None
         ) -> None:
+            # We should not generate guards for unbacked inputs.
+            if isinstance(val, SymInt) and has_free_unbacked_symbols(val.node.expr):
```
aorenste: Is this right? So we just silently ignore unbacked in this case? I'm surprised this isn't either an assertion (you shouldn't have unbacked here - and then figure out why this is being called at all) or a DDE (which could be caught by a higher-level guard_or_false if needed).

laithsakka: This is a function that we call on each symint, the purpose is to add a guard between the original sym associated with the property and the new one observed post replacement.

For example.
if i have s0  as t.size()[0], then have a torch._check(s0, u0) the symnode will show up as u0 due to replacements.  then I would add a guard t.size()[0] == source_of(u0). here. (before this PR). 

 I could add this check at each call site but there is many of them.
 
 it also prepare  symbol_to_source[s] which is used later to add guards on symbols
 (like var range) min/max, but was trying to avoid all at once place. 
 but with my change i avoid prepping it for unbacked.

laithsakka: I could potentially move the check to somewhere else maybe after thus line let me try that and see if that more clear for readers. but i will have to be more other places.

laithsakka: ok code updated to be more explicit about what guards we are dropping.

laithsakka:
taking this off review for a bit i would like to play with it more and do more changes.

torch/fx/experimental/symbolic_shapes.py
laithsakka: I have to rethink this anyway current fix is not wrong,

PR #163714, opened 2025-09-24
[opaque obj] Allow non-effectful scriptobjs

angelayi:
Fixes functionalization so that we can run ops using ScriptObjects w/o needing effects. Previously we would run into an error when running functionalization on the TorchBindOpOverloads.

* #167364
* #167363
* #167245
* #167231
* #167230
* #163936
* __->__ #163714
* #163284


torch/_subclasses/functional_tensor.py
```diff
@@ -508,11 +508,25 @@ def unwrap(x):
...
                     # from the TLS in order to avoid infinite looping, but this would prevent us from coming
                     # back to PreDispatch later
-                    outs_unwrapped = func._op_dk(
-                        torch._C.DispatchKey.Functionalize,
-                        *args_unwrapped,
-                        **kwargs_unwrapped,
-                    )
+                    if isinstance(func, TorchBindOpOverload):
+                        with (
+                            torch.utils._python_dispatch._pop_mode_temporarily() as mode
```
ydwu4: what will happen if we just do 
```python
func(*fully_unwrapped_args, **fully_unwrapped_kwargs,)
```
My mental model is that this will go through the must_dispatch_in_python logic and automatically figure out the next dispatch key (e.g. Python, which then pop the mode out there) so we don't have to manually pop the mode and handle_dispatch_mode. If there is a key after functionalization but it's not a dispatch mode, the current logic will skip that key.

angelayi: yes! this works too!

test/test_opaque_obj.py
```diff
@@ -90,7 +90,7 @@ def pop_impl_fake(q: torch._C.ScriptObject) -> torch.Tensor:
             # This is not accurate since the queue could have tensors that are
             # not rank 1
             ctx = torch._custom_op.impl.get_ctx()
-            u0 = ctx.create_unbacked_symint()
+            u0 = ctx.new_dynamic_size()
```
zou3519, line 93: Add a description to the PR? I'm not sure what a "non-effectful scriptobjs" is right now

test/test_opaque_obj_v2.py
```diff
@@ -233,6 +235,68 @@ def bad_fake2(x, rng_state) -> torch.Tensor:
...
+        )
+
+        # By default, ops with ScriptObjects as inputs are registered as being
+        # effectful
+        _deregister_effectful_op("_TestOpaqueObject::noisy_inject.default")
+
+        # If we register with None, this means the ops do not have effect
+        torch.library._register_effectful_op(
+            "_TestOpaqueObject::noisy_inject.default", None
+        )
```
zou3519: What happens when you call register_effectful_op twice? Does it override or does it error?

test/test_opaque_obj_v2.py
```diff
@@ -233,6 +235,68 @@ def bad_fake2(x, rng_state) -> torch.Tensor:
...
+    mul = torch.ops.aten.mul.Tensor(getitem_1, getitem_1);  getitem_1 = None
+    with_effects_1 = torch.ops.higher_order.with_effects(getitem, torch.ops._TestOpaqueObject.noisy_inject.default, mul, arg1_1);  getitem = mul = arg1_1 = None
+    getitem_2 = with_effects_1[0]
+    getitem_3 = with_effects_1[1];  with_effects_1 = None
+    add = torch.ops.aten.add.Tensor(getitem_3, getitem_3);  getitem_3 = None
+    return (getitem_2, add)""",  # noqa: B950
+        )
+
+        # By default, ops with ScriptObjects as inputs are registered as being
+        # effectful
```
zou3519: @angelayi for opaque_object_v2, aren't we directly using PyObject* (instead of ScriptObject)? If so, we have an option for the PyObject* to be non-effectful by default.

test/test_opaque_obj_v2.py
angelayi: It should override

test/test_opaque_obj_v2.py
angelayi: yes! fixed this so the default is noneffectful

test/test_opaque_obj_v2.py
```diff
@@ -233,6 +236,62 @@ def bad_fake2(x, rng_state) -> torch.Tensor:
...
+    getitem_1 = with_effects[1];  with_effects = None
+    mul = torch.ops.aten.mul.Tensor(getitem_1, getitem_1);  getitem_1 = None
+    with_effects_1 = torch.ops.higher_order.with_effects(getitem, torch.ops._TestOpaqueObject.noisy_inject.default, mul, arg1_1);  getitem = mul = arg1_1 = None
+    getitem_2 = with_effects_1[0]
+    getitem_3 = with_effects_1[1];  with_effects_1 = None
+    add = torch.ops.aten.add.Tensor(getitem_3, getitem_3);  getitem_3 = None
+    return (getitem_2, add)""",  # noqa: B950
+        )
+
+        _deregister_effectful_op("_TestOpaqueObject::noisy_inject")
```
zou3519: Why is the deregister here? if it's to reset the state, you want a try-finally and this in the finally block. otherwise if there's an early return due to exception you might see flaky tests

zou3519: Also can I just  torch.library._register_effectful_op(
            "_TestOpaqueObject::noisy_inject", None
        ) to undo this?

torch/_subclasses/functional_tensor.py
```diff
@@ -504,65 +504,79 @@ def unwrap(x):
...
-                    outs_wrapped = pytree.tree_map_only(
-                        torch.Tensor, wrap, outs_unwrapped
-                    )
-            finally:
-                torch._disable_functionalization()
-                torch._functionalize_enable_reapply_views(old_apply_views)  # type: ignore[attr-defined]
+        if isinstance(func, TorchBindOpOverload):
+            # When the function is a TorchBindOpOverload, meaning some of the
+            # inputs are FakeScriptObjects, we need to skip c++ dispatcher and
+            # dispatch in python because C++ dispatcher will check the schema
```
zou3519, line 510: What happens if you have a torchbindopoverload that mutates an input tensor? Did this case work before? Does it work now?

zou3519 (commented):
seems reasonable, i just want to understand what happens in the case where we're mutating a tensor

torch/_subclasses/functional_tensor.py
angelayi: I tried with a torchbindopoverload that mutates an input tensor and it gets correctly wrapped with auto_functionalize. This works before though! I added an expecttest in the following PR

test/test_opaque_obj_v2.py
angelayi: yes!

torch/_subclasses/functional_tensor.py
zou3519: nice

zou3519 (approved):

PR #163729, opened 2025-09-24
[Inductor-FX] Support unbacked symbol definitions

blaine-rister:
# Problem
Inductor sometimes generates unbacked symints to handle things like mismatched branches of `torch.cond`. This code is represented by `pytree.KeyPath`, with special codegen logic to convert it to Python and C++. This was not previously supported by the FX backend.

# Feature
This PR adds support for unbacked symbol declarations to the FX backend. The implementation is fairly straightforward.
1. Instead of raw Python/C++, update the wrapper codegen method to emit a new Wrapper IR line called `UnbackedSymbolDefsLine`. This contains all the information needed to  generate the Python and C++ code.
2. Move the existing Python/C++ codegen to a private method, which is invoked by `UnbackedSymbolDefsLine.codegen()`.
3. Implement a method to generate FX IR from unbacked symbol definitions. The implementation is based on recursive descent, consuming some keypath entries, emitting an FX IR node, and recursing to the rest of the keypath. It is conceptually identical to the existing algorithm for Python and C++, except it generates FX nodes.
4. The FX backend currently relies on size hints to generate autotuning arguments, and consequently autotuning does not support unbacked SymInts. At some point, we would like to generalize the autotuning logic to support these. But for now, simply emit a warning and skip autotuning when we see them.
5. The new test case exposed some tricky issues reconciling Triton call args with constants stored in `triton_meta`. This PR rewrites the relevant helper function to do this in a more principled way.

# Test plan
This PR imports an existing control flow test to the FX backend's test suite. The test uses unbacked symbol definitions to handle mismatched dynamic shapes coming from `torch.cond` branches.


torch/_inductor/codegen/wrapper_fxir.py
```diff
@@ -981,30 +1003,36 @@ def add_constants_to_call_args(
...
+                    call_args[len(call_args) - len(extra_launcher_args) :]
+                )
 
             return tuple(new_call_args)
 
         kernel_config = tuner.compile_results[0].config
         call_args = add_constants_to_call_args(call_args, kernel_config)
         call_args, grid = tuner._interpret_args_grid(call_args, kernel_config)
         call_kwargs = dict(zip(signature, call_args))
+        assert not any(kwarg in kernel_config.kwargs for kwarg in call_kwargs), (
```
blaine-rister, line 1033: The test case in this PR exposed a very tricky bug related to arg/kwarg handling. Although this should no longer happen with the updated method, this assert guards against that bug if by some off chance we missed something.

jansel (approved):

PR #163746, opened 2025-09-24
fixes import error 'functionalize' from functorch

RiyaP2508:
Fixes #163637 


malfet (approved):

malfet:
> I hit the same issue. 
Do you have a reproducer?
But change is indeed ok, does it fix the problem for you?

PR #163749, opened 2025-09-24
[dynamo] [fix] fix weakref proxy in dynamo symbolic covert when triggering a graph break 

shaoyuyoung:
Similar to #161508
The repro can be found in my test (so I didn't draft an issue).
```python
import torch
import weakref
import torch._dynamo as dynamo
from torch._dynamo.symbolic_convert import code_context


class Dummy:
    pass


def target_fn(x: torch.Tensor) -> torch.Tensor:
    dynamo.graph_break()
    return x + 1



dummy = Dummy()
proxy = weakref.proxy(dummy)
code_context.get_context(target_fn.__code__)["orig_graphmodule"] = (lambda: proxy)
f_compiled = torch.compile(target_fn)
y = f_compiled(torch.tensor(1))
torch.testing.assert_close(y, torch.tensor(2))
```
err log
```
File "/root/ysy/pytorch/torch/_dynamo/symbolic_convert.py", line 2708, in create_call_resume_at
  code_context.get_context(new_code)["orig_graphmodule"] = weakref.ref(
                                                           ^^^^^^^^^^^^
torch._dynamo.exc.InternalTorchDynamoError: TypeError: cannot create weak reference to 'weakref.ProxyType' object
```

I think we should add a check for `context["orig_graphmodule"]` to avoid repeatedly creating weak references. Because this will throw the above `TypeError:`

### DIFF
--- a/test/inductor/test_cpu_repro.py
+++ b/test/inductor/test_cpu_repro.py
@@ -485,17 +485,6 @@ class CPUReproTests(TestCase):
             example_inputs = (torch.rand(1, 10),)
             self.common(Model(), example_inputs)
 
-    @torch._dynamo.config.patch(capture_scalar_outputs=True)
-    def test_fill_diagonal_item_scalar_cpu(self):
-        def fn():
-            x = torch.ones(3, 3)
-            x.fill_diagonal_(0)
-            return x.sum().item()
-
-        compiled = torch.compile(fn, backend="inductor", fullgraph=True)
-        eager = fn()
-        self.assertEqual(compiled(), eager)
-
     @unittest.skipIf(not torch.backends.mkldnn.is_available(), "MKLDNN is not enabled")
     @patch("torch.cuda.is_available", lambda: False)
     def test_linear_packed(self):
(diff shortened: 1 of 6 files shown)


torch/_dynamo/symbolic_convert.py
```diff
@@ -2701,7 +2701,7 @@ def create_call_resume_at(
             orig_graphmodule_maybe = code_context.get_context(cur_tx.f_code).get(
                 "orig_graphmodule", lambda: None
             )()
-            if orig_graphmodule_maybe is not None:
+            if orig_graphmodule_maybe and not isinstance(orig_graphmodule_maybe, weakref.ProxyType):
```
can-gaa-hou: `weakref.ProxyTypes` contains `weakref.ProxyType` and `weakref.CallableProxyType`, which would be better in this case I guess.

shaoyuyoung: thanks! resolved!

mlazos (approved):

shaoyuyoung:
Resolve some conflicts on the previous new commits

PR #124444, opened 2024-04-19
Add support for capturing tensors with score_mod

Chillee:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #124799
* __->__ #124444

```
import torch
from torch import nn
import torch.nn.functional as F
import torch._inductor.config as config
# torch.set_default_device('cuda')

import torch
from torch.nn.attention._templated_attention import _templated_attention as templated_attention
from triton.testing import do_bench
from torch.nn.attention import SDPBackend, sdpa_kernel


index = torch.ops.aten
torch.manual_seed(0)

B = 16
H = 16
S = 2048
D = 64

head_scale = torch.randn(H, device='cuda')
def alibi(score, batch, head, token_q, token_kv):
    return score + torch.ops.aten.index(head_scale, [head]) * (token_q - token_kv)
bias = torch.randn(H, S, S, dtype=torch.float16, device='cuda')

query = torch.randn(B, H, S, D, device="cuda", dtype=torch.float16)
key = torch.randn(B, H, S, D, device="cuda", dtype=torch.float16)
value = torch.randn(B, H, S, D, device="cuda", dtype=torch.float16)

compiled = torch.compile(templated_attention)
out = compiled(query, key, value, score_mod=alibi)
out2 = templated_attention(query, key, value,score_mod=alibi)
print((out - out2).abs().mean())
assert (out - out2).abs().mean() < 1e-3
print("Flash (no mask): ", do_bench(lambda: F.scaled_dot_product_attention(query, key, value)))
print("Flash (mask): ", do_bench(lambda: F.scaled_dot_product_attention(query, key, value, attn_mask=bias)))
print("flexattention: ", do_bench(lambda: compiled(query, key, value, score_mod=alibi)))
```
image

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @peterbell10 @ipiszy @yf225 @chenyang78 @kadeng @muchulee8 @aakhundov @ColinPeppler @amjames @desertfire @chauhang

Differential Revision: [D56583900](https://our.internmc.facebook.com/intern/diff/D56583900)

PR #163752, opened 2025-09-24
[inductor][templates] Distinguish between kernel input nodes and codegen input nodes

kundaMwiza:
If there is a single autotuner choice, the wrong type of input node is used to instantiate `TritonTemplateBuffer` through `TritonTemplateCaller.output_node`. This PR distinguishes the input nodes used in `AlgorithmSelectorCache.__call__` between the actual inputs passed to the kernel at runtime, vs the possibly viewed inputs that influence scheduling behaviour (e.g. `MemoryDeps`) and codegen. See the added unit test for more detail.

### DIFF
--- a/torch/_inductor/select_algorithm.py
+++ b/torch/_inductor/select_algorithm.py
@@ -1985,7 +1985,9 @@ class TritonTemplate(KernelTemplate):
 
         # `kernel_input_nodes` are the actual inputs that will be passed to the kernel,
         # so e.g. views of the same input are not included. `codegen_input_nodes`
-        # includes views of inputs to preserve the kernel semantics
+        # includes views of inputs to preserve the kernel semantics. The shape and
+        # strides of `codegen_input_nodes` will be used to infer read/writes in
+        # TemplateBuffer.extract_read_writes
         kernel_input_nodes = tuple(
             [V.graph.get_buffer(k) for k in result.input_call_args]
         )
--- a/torch/_inductor/select_algorithm.py
+++ b/torch/_inductor/select_algorithm.py
@@ -1992,7 +1992,9 @@ class TritonTemplate(KernelTemplate):
             [V.graph.get_buffer(k) for k in result.input_call_args]
         )
         # Here we have (*input_nodes, *captured_buffers)
-        codegen_input_nodes = tuple(input_nodes) + kernel_input_nodes[len(expected_input_args):]
+        codegen_input_nodes = (
+            tuple(input_nodes) + kernel_input_nodes[len(expected_input_args) :]
+        )
         extra_args = V.graph.sizevars.size_hints(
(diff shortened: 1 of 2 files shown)


kundaMwiza:
CC @Chillee original change in https://github.com/pytorch/pytorch/pull/124444 @eellison  for prologue fusion

torch/_inductor/select_algorithm.py
```diff
@@ -1977,9 +1977,13 @@ def generate(  # type: ignore[override]
             expected_input_args,
         )
 
-        full_input_nodes = tuple(
+        # `kernel_input_nodes` are the actual inputs that will be passed to the kernel,
+        # so e.g. views of the same input are not included. `codegen_input_nodes`
+        # includes views of inputs to preserve the kernel semantics
+        kernel_input_nodes = tuple(
             [V.graph.get_buffer(k) for k in result.input_call_args]
```
kundaMwiza, line 1992: Views of computed buffers are not stored as buffers on the graph

torch/_inductor/select_algorithm.py
```diff
@@ -1977,9 +1977,13 @@ def generate(  # type: ignore[override]
             expected_input_args,
         )
 
-        full_input_nodes = tuple(
+        # `kernel_input_nodes` are the actual inputs that will be passed to the kernel,
+        # so e.g. views of the same input are not included. `codegen_input_nodes`
+        # includes views of inputs to preserve the kernel semantics
```
eellison: Maybe add a comment that these will be the strides used to infer read/writes in `TemplateBuffer.extract_read_writes` ?

eellison (approved):
Looks good, thank you for pr!! cc @coconutruben as well

torch/_inductor/select_algorithm.py
```diff
@@ -1977,7 +1977,18 @@ def generate(  # type: ignore[override]
...
+        # so e.g. views of the same input are not included. `codegen_input_nodes`
+        # includes views of inputs to preserve the kernel semantics. The shape and
+        # strides of `codegen_input_nodes` will be used to infer read/writes in
+        # TemplateBuffer.extract_read_writes
+        kernel_input_nodes = tuple(
+            [V.graph.get_buffer(k) for k in result.input_call_args]
+        )
+        # Here we have (*input_nodes, *captured_buffers)
+        codegen_input_nodes = (
+            tuple(input_nodes) + kernel_input_nodes[len(expected_input_args) :]
```
kundaMwiza, line 1996: @eellison Merge failed since I did not include inputs for subgraphs and this change restores those inputs, so flex attention tests should now pass. From what I can see inputs for subgraphs don't take part in fusion so it should be fine to just use the graph inputs here? If they do, the template would also need to pass those inputs if there is a chance that views of them can be taken

PR #163755, opened 2025-09-24
[inductor] Improve bound on the number of dims to match for the block

kundaMwiza:
- Removes redundant broadcast code when `len(kernel.range_tree_nodes)` is much larger than `len(range_tree.nodes)`. For example:
```python
# before, the broadcast is to [1, 1, XBLOCK, R0_BLOCK]
tmp0 = tl.reshape(tl.broadcast_to(tl.load(block_ptr0, boundary_check=[2], padding_option='zero', eviction_policy='evict_last')[:, None, :, :], [(511 + XBLOCK) // 512, ((1) * ((1) <= ((511 + XBLOCK) // 512)) + ((511 + XBLOCK) // 512) * (((511 + XBLOCK) // 512) < (1))), ((512) * ((512) <= (XBLOCK)) + (XBLOCK) * ((XBLOCK) < (512))), R0_BLOCK]), [XBLOCK, R0_BLOCK])
# after
tmp0 = tl.reshape(tl.load(block_ptr0, boundary_check=[2], padding_option='zero', eviction_policy='evict_last'), [XBLOCK, R0_BLOCK])
```
- Fix: also save range_tree_nodes per subgraph 

### DIFF
--- a/torch/_inductor/codegen/triton.py
+++ b/torch/_inductor/codegen/triton.py
@@ -2234,7 +2234,9 @@ class TritonKernel(SIMDKernel[TritonCSEVariable]):
                 )
                 num_dims = max(
                     2,
-                    len(self.range_tree_nodes),
+                    # range_tree.nodes only includes the entries for the range tree
+                    # len(range_tree.nodes) <= self.range_tree_nodes
+                    len(range_tree.nodes),
                     (
                         index.count(FloorDiv(index_var, denom))
                         + index.count(ModularIndexing(index_var, denom, modulo))
--- a/torch/_inductor/select_algorithm.py
+++ b/torch/_inductor/select_algorithm.py
@@ -105,7 +105,7 @@ DEBUG = False
 if TYPE_CHECKING:
     import concurrent
 
-    from torch._inductor.codegen.simd import IterationRangesRoot
+    from torch._inductor.codegen.simd import IterationRangesEntry, IterationRangesRoot
 
     from .codegen.common import CSE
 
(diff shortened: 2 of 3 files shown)


torch/_inductor/codegen/triton.py
```diff
@@ -2234,7 +2234,9 @@ def match_mod_div_block(
                 )
                 num_dims = max(
                     2,
-                    len(self.range_tree_nodes),
+                    # range_tree.nodes only includes the entries for the range tree
+                    # len(range_tree.nodes) <= self.range_tree_nodes
+                    len(range_tree.nodes),
```
eellison, line 2239: I know this code less well but - would it make sense to try to match for multiple values of `num_dims` @blaine-rister ?

eellison (approved):
Nice! lets let @blaine-rister review as well

torch/_inductor/codegen/triton.py
kundaMwiza: @blaine-rister Can I please get a review when you're free? Thanks

torch/_inductor/select_algorithm.py
```diff
@@ -267,10 +267,16 @@ class SubgraphInfo:
 
     # only copied over if not None
     range_trees: Optional[list["IterationRangesRoot"]] = None
+    range_tree_nodes: Optional[dict[sympy.Symbol, "IterationRangesEntry"]] = None
```
blaine-rister, line 270: Just for my own understanding, how is this code related to the rest of the PR? It seems like you're only using 'self.range_trees`, but I could be missing something.

blaine-rister (approved):
LGTM. Thanks for the improvement.

torch/_inductor/select_algorithm.py
kundaMwiza: This is just book keeping to isolate the `range_tree_nodes` to a subgraph, similar to the `range_trees`. Without this a subgraph would have range entries for unrelated `range_trees` from another subgraph

blaine-rister: Makes sense! Thanks.

PR #163762, opened 2025-09-24
Pass through CMake compiler/linker args to setup.py

directhex:
Presumably if the arguments were worth giving to CMake, they were worth giving to setup.py's compile calls too. For example, if your specific environment requires use of a `-fuse-ld`  flag or `--target=` flag.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1435,9 +1435,13 @@ def configure_extension_build() -> tuple[
         ]
 
     main_compile_args: list[str] = []
+    if cmake_cache_vars["CMAKE_C_FLAGS"]:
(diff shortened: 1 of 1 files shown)


cyyever:
You can use `LDFLAGS`

directhex:
> You can use `LDFLAGS`,`CXXFLAGS` and `CFLAGS`. They should work on all UNIX-like systems.

Sorry, you're right. I didn't notice that another part of a wrapper script on my side was defining CMAKE_C_FLAGS, and cmake will ignore CFLAGS if CMAKE_C_FLAGS is non-null

PR #163763, opened 2025-09-24
Fix THP_PyObject_VirtualFree return type

guangyey:
* __->__ #163763

# Motivation
`void THP_PyObject_VirtualFree` should have no return value; otherwise, it would raise a build warning 
```bash
C:\Users\guangyey\pytorch\torch\csrc\dynamo\cpython_defs.c(264): warning C4098: 'THP_PyObject_VirtualFree': 'void' function returning a value
```
# Additional Context
Refer to
https://github.com/python/cpython/blob/c4f21d7c7c415a85a975fb878a1e578c12969d82/Include/cpython/objimpl.h#L59-L68
PyObjectArenaAllocator::free is defined with `void` return type.


albanD (changes requested):
This is taken from CPython codebase. Unless it is updated there, I prefer not to change it here to ensure perfect 1-1 matching between the two.

Skylion007 (dismissed):

williamwen42 (approved):

guangyey:
@albanD The original code is
https://github.com/python/cpython/blob/fad48ea1816be3125ea51edcdfe2f999d6ade796/Objects/obmalloc.c#L640-L644
I think here should be a typo.

albanD (approved):
Ok then

PR #163768, opened 2025-09-24
[CD] Add statically linked windows libraries to exclude list

### LINKED ISSUE #159514: [BE] Only package top level `.lib` files into whl on Windows
### 🐛 Describe the bug

PyTorch Windows wheel are packaged with `.lib` files for all libraries that are bundled into the wheel, but only ones pointing to a public facing APIs(c10.dll, torch_*.dll ) are needed to develop extensions against PyTorch, so other are just dead weight and should not be shipped

### Versions

CI

atalman:
Fixes: https://github.com/pytorch/pytorch/issues/159514

Seeing following in the Wheel build logs:
```
Linking CXX static library lib\kineto.lib
Linking CXX static library lib\dnnl.lib
....
```

These files are around 800MB uncompressed and 109MB compressed, hence provide ~50% size reduction for Windows CPU builds.

Test Plan: Build Pytorch Windows binary. Build vision, audio and torchcodec with this binary. Smoke test.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1706,13 +1706,16 @@ def main() -> None:
     }
     # windows libraries *.lib are excluded
     # these are statically linked
-    exclude_package_data = {
+    exclude_windows_libs = [
         "lib/dnnl.lib",
         "lib/kineto.lib",
         "lib/libprotobuf-lite.lib",
         "lib/libprotobuf.lib",
         "lib/libprotoc.lib",
-    }
+    ]
+    exclude_package_data = [
+        "torch": exclude_windows_libs,
+    ]
 
     if not BUILD_LIBTORCH_WHL:
         package_data["torchgen"] = torchgen_package_data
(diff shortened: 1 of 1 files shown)


albanD (approved):
Sounds good on principle, how do you test this?

setup.py
```diff
@@ -1704,7 +1704,18 @@ def main() -> None:
     package_data = {
         "torch": torch_package_data,
     }
-    exclude_package_data = {}
+    # windows libraries *.lib are excluded
```
malfet: This is a misleading statement, as `.libs` are not always static libraries, there are something called import libraries: https://learn.microsoft.com/en-us/cpp/build/reference/working-with-import-libraries-and-export-files

malfet (approved):

atalman:
Successful test of torchvision: https://github.com/pytorch/vision/actions/runs/18006184532/job/51234220133

PR #163773, opened 2025-09-24
Avoid fast path mask left-align check in compiled TransformerEncoder

### LINKED ISSUE #163640: [torch.compile] TransformerEncoder with boolean src_key_padding_mask fails under Dynamo fullgraph mode
### Describe the bug

Compiling a simple `nn.TransformerEncoder` model with `torch.compile(fullgraph=True)` fails when a boolean `src_key_padding_mask` is passed.

Eager mode works correctly, but `torch.compile` raises `torch._dynamo.exc.Unsupported` because of an internal call to `torch._nested_tensor_from_mask_left_aligned`, which returns a `bool` instead of a `Tensor`.

### Minimal Reproducible Example
```python
import torch
import torch.nn as nn

class TinyEnc(nn.Module):
    def __init__(self, d_model=512, nhead=8, num_layers=1):
        super().__init__()
        layer = nn.TransformerEncoderLayer(
            d_model=d_model, nhead=nhead, batch_first=True, dropout=0.1
        )
        self.enc = nn.TransformerEncoder(layer, num_layers=num_layers)
        self.proj = nn.Linear(d_model, 10)

    def forward(self, x, pad_mask):
        # Passing src_key_padding_mask triggers torch._nested_tensor_from_mask_left_aligned
        y = self.enc(x, mask=None, src_key_padding_mask=pad_mask)
        return self.proj(y)

def main():
    torch.manual_seed(0)
    m = TinyEnc().eval()

    B, T, C = 1, 41, 512
    x = torch.randn(B, T, C, dtype=torch.float32)
    pad_mask = (torch.rand(B, T) > 0.5)
    pad_mask[..., 0] = True

    # Eager is fine
    with torch.inference_mode():
        y = m(x, pad_mask)
    print("eager ok:", tuple(y.shape))

    # Compile (fullgraph=True required to reproduce)
    cm = torch.compile(m, backend="inductor", fullgraph=True)
    _ = cm(x, pad_mask)

if __name__ == "__main__":
    main()
```

### Observed error
```
eager ok: (1, 41, 10)

torch._dynamo.exc.Unsupported: torch.* op returned non-Tensor
  Explanation: torch.* ops that return a non-Tensor cannot be traced into the Dynamo FX graph output

Developer debug context: example_value type: bool; op: call_function; 
target: <built-in method _nested_tensor_from_mask_left_aligned of type object at 0x...>

from user code:
   File "test.py", line 16, in forward
    y = self.enc(x, mask=None, src_key_padding_mask=pad_mask)
   File "torch/nn/modules/transformer.py", line 467, in forward
    ) and not torch._nested_tensor_from_mask_left_aligned(
```

### Versions

```
PyTorch version:  2.8.0a0+gitba56102 & 2.9.0a0+git31d5c67
Is debug build: True

CMake version: version 3.22.1
Libc version: glibc-2.35

Is CUDA available: True
```

cc @albanD @mruberry @jbschlosser @walterddr @mikaylagawarecki @malfet @chauhang @penguinwu

jbschlosser:
* __->__ #163773

Fixes #163640

This PR avoids a mask left align check in the case that we're operating under torch.compile / torch.export. Originally, I planned to make a more invasive change to auto-disable the fast path entirely underneath torch.compile / torch.export, but I realized during testing that the fast path wasn't actually causing compile issues outside of the narrow issue identified here.

### DIFF
--- a/torch/nn/modules/transformer.py
+++ b/torch/nn/modules/transformer.py
@@ -443,6 +443,7 @@ class TransformerEncoder(Module):
         str_first_layer = "self.layers[0]"
         batch_first = first_layer.self_attn.batch_first
         is_fastpath_enabled = torch.backends.mha.get_fastpath_enabled()
+        do_mask_check = getattr(self, "mask_check", True)
 
         if not is_fastpath_enabled:
             why_not_sparsity_fast_path = (
@@ -464,15 +465,11 @@ class TransformerEncoder(Module):
             why_not_sparsity_fast_path = "src_key_padding_mask was None"
         # This check avoids a call to torch._nested_tensor_from_mask_left_aligned() that
         # breaks in torch.compile.
-        elif (
-            (not hasattr(self, "mask_check")) or self.mask_check
-        ) and torch.compiler.is_compiling():
+        elif do_mask_check and torch.compiler.is_compiling():
             why_not_sparsity_fast_path = (
                 "mask_check enabled with torch.compile or torch.export"
             )
-        elif (
-            (not hasattr(self, "mask_check")) or self.mask_check
-        ) and not torch._nested_tensor_from_mask_left_aligned(
+        elif do_mask_check and not torch._nested_tensor_from_mask_left_aligned(
             src, src_key_padding_mask.logical_not()
         ):
             why_not_sparsity_fast_path = "mask_check enabled, and src and src_key_padding_mask was not left aligned"
--- a/torch/nn/modules/transformer.py
+++ b/torch/nn/modules/transformer.py
@@ -462,6 +462,14 @@ class TransformerEncoder(Module):
             )
         elif src_key_padding_mask is None:
             why_not_sparsity_fast_path = "src_key_padding_mask was None"
+        # This check avoids a call to torch._nested_tensor_from_mask_left_aligned() that
+        # breaks in torch.compile.
+        elif (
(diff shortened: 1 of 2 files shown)


mikaylagawarecki (approved):
Thank you, just one question from me

torch/nn/modules/transformer.py
```diff
@@ -462,6 +462,14 @@ def forward(
             )
         elif src_key_padding_mask is None:
             why_not_sparsity_fast_path = "src_key_padding_mask was None"
+        # This check avoids a call to torch._nested_tensor_from_mask_left_aligned() that
+        # breaks in torch.compile.
+        elif (
+            (not hasattr(self, "mask_check")) or self.mask_check
```
mikaylagawarecki: I know this is used in the branch just below, but to confirm, if not hasattr(self, "mask_check"), this means the mask is always checked?

jbschlosser: yes that's correct.. the default is to do a mask check if for some reason the attribute isn't present. Doing a mask check is also the default if the attribute was unspecified upon construction.

jbschlosser: I made this clearer through the use of `getattr(self, "mask_check", True)` instead of `hasattr()`. Leaving it in for now to absolutely minimize risk

Issue #124480, opened 2024-04-19
Fused Linear and Cross-Entropy Loss `torch.nn.functional.linear_cross_entropy`


imoneoi:
### 🚀 The feature, motivation and pitch

It'd be great to have a fused linear and cross-entropy function in PyTorch, for example, `torch.nn.functional.linear_cross_entropy`. This function acts as a fused linear projection followed by a cross-entropy loss, e.g.

```python
def linear_cross_entropy(linear_weights, input, labels):
    logits = F.linear(input, linear_weights)
    return F.cross_entropy(logits, labels)
```

Compared to naive implementation, this fused function does not materialize the intermediate `logits`, saving a lot of VRAM.

This function is quite common in models, such as LLMs, classification, and so on. When the batch size or number of classes is large, the intermediate `logits` would take up the majority of VRAM.

For example, the latest LLMs, such as Llama 3 and Gemma, adopt extremely large vocabularies (128-256K). Thus, the size of logits can become very large, consuming a significant proportion of VRAM. The calculation below shows the VRAM size for Llama 3 8B with a batch size of 81920 tokens:

````
Logits size: 128256 (vocabulary size) * 81920 (batch size) * 2 bytes (bf16) = 19.57GiB
Hidden state size (checkpointed): 4096 (hidden size) * 81920 (batch size) * 32 (layers) * 2 bytes (bf16) = 20GiB
````

Therefore, a fused linear and cross-entropy loss function that does not require materializing full `logits` may reduce VRAM consumption by half.

Related: the same feature request for FlashAttention: https://github.com/Dao-AILab/flash-attention/issues/922

awgu:
cc: @Chillee would be curious to hear your thoughts

balancap:
Funny thing, I have been working on exactly that very recently (for Graphcore IPU, but nothing specific about the hardware in the technique).
It's definitely possible to serialize & fuse the linear + softmax + cross entropy combination, the only trade-off being the recomputation necessary for the backward pass. Happy to help!

YouJiacheng:
FWIW, I derived some formulas & wrote a (very rough) WIP prototype.
```python
import torch

import triton
import triton.language as tl

@triton.jit
def fwd_kernel(
    x_ND_ptr,
    w_DV_ptr,
    c_N_ptr,
    output_N_ptr,
    l_N_ptr,
    N, D, V,
    stride_xn, stride_xd,
    stride_wd, stride_wv,
    # Meta-parameters
[...]

Issue #163408, opened 2025-09-20
RFC: Inherit some modules for torchtune.


krammnic:
### 🚀 The feature, motivation and pitch

Hello from [pytorch/torchtune](https://github.com/pytorch/torchtune).

Support of the torchtune was stopped in order to propose something new! We know that our module was extremely valuable in the community, and it contains dozens of high-quality pure torch code. Therefore, we propose to inherit some specific modules like RoPE, LCE, etc., in the core.

Motivation: Each new framework/library (honestly every new abstraction) reimplements these modules and sometimes does it in not the best manner. Hence, we might face some inconsistencies between implementations; the amount of duplicating code is growing, and in the worst scenario, we can get critical bugs. Furthermore, any new library will and should not depend on any current complex library (even on torchtune) to use them. Standardization of such modules in the core will reduce the amount of new code, reduce the amount of fragmentation between different abstractions, and help to mitigate potential problems that might be caused by new implementation. 

Proposal: Inherit important submodules of torchtune under a new module in core like `torch.llm._functional`.

Credits to @vadimkantorov

mikaylagawarecki:
> LinearCrossEntropy

we have plan to add a kernel for this, @krocki is working on it

> Grouped expert Layer
> MoE and Top K routing

cc @slayton58 who I believe might be familiar with plans for the `_{*}_gemm` APIs and @ngimel on whether there would value of adding an nn layer for these

> Optimizer in Backward

cc @janeyx99 for thoughts on whether we should upstream this to optim

> Rotary Positional Embeddings

See the comment Vadim linked https://github.com/pytorch/pytorch/issues/149534#issuecomment-2748614292

PR #163774, opened 2025-09-24
Add fused linear cross entropy CPU implementation

krocki:
Summary:
  - add a fused CPU implementation of linear cross entropy with vocabulary/batch chunking so large logits never fully materialize
  - expose torch.nn.functional.linear_cross_entropy; the CPU path uses the fused kernel while other devices fall back to linear + cross_entropy (keeping behavior identical)
  - add unit tests covering forward/backward parity (auto/vocab/batch), reductions, ignore_index, label smoothing, and gradcheck

  Addresses: #124480
  GPU/Triton support will follow in a separate PR; please note this is the first installment toward #124480.

  Release Notes:
  - [New Feature] torch.nn.functional.linear_cross_entropy now provides a fused CPU implementation with chunking heuristics to avoid materializing large logit tensors.

  Test Plan:
  - python test/nn/test_linear_cross_entropy.py -v
  
  Expected output:

```
  test_batch_chunking (main.TestLinearCrossEntropyCPU.test_batch_chunking) ... ok
  test_forward_backward_matches_reference_auto (main.TestLinearCrossEntropyCPU.test_forward_backward_matches_reference_auto) ... ok
  test_gradcheck (main.TestLinearCrossEntropyCPU.test_gradcheck) ... ok
  test_parameter_validation (main.TestLinearCrossEntropyCPU.test_parameter_validation) ... ok
  test_reduction_and_options (main.TestLinearCrossEntropyCPU.test_reduction_and_options) ... ok
  test_vocab_chunking (main.TestLinearCrossEntropyCPU.test_vocab_chunking) ... ok
```

### DIFF
--- a/aten/src/ATen/native/LinearCrossEntropy.cpp
+++ b/aten/src/ATen/native/LinearCrossEntropy.cpp
@@ -1,0 +1,1245 @@
+#define TORCH_ASSERT_ONLY_METHOD_OPERATORS
+#include <ATen/core/Tensor.h>
+#include <ATen/Dispatch.h>
+#include <ATen/native/Resize.h>
+#include <ATen/TensorIterator.h>
+#include <ATen/TensorOperators.h>
+#include <c10/util/irange.h>
+#include <limits>
+#include <tuple>
+#include <vector>
+#include <optional>
+#include <algorithm>
+#include <string>
+
+#ifndef AT_PER_OPERATOR_HEADERS
+#include <ATen/Functions.h>
+#include <ATen/NativeFunctions.h>
+#else
+#include <ATen/ops/empty.h>
+#include <ATen/ops/linear.h>
+#include <ATen/ops/cross_entropy_loss.h>
+#include <ATen/ops/zeros.h>
+#include <ATen/ops/zeros_like.h>
+#include <ATen/ops/full.h>
+#include <ATen/ops/ones.h>
+#include <ATen/ops/max.h>
+#include <ATen/ops/exp.h>
+#include <ATen/ops/log.h>
+#include <ATen/ops/logsumexp.h>
+#include <ATen/ops/where.h>
+#include <ATen/ops/ge.h>
+#include <ATen/ops/lt.h>
+#include <ATen/ops/logical_and.h>
+#include <ATen/ops/logical_or.h>
+#include <ATen/ops/logical_not.h>
+#include <ATen/ops/masked_fill.h>
+#include <ATen/ops/sub.h>
+#include <ATen/ops/add.h>
+#include <ATen/ops/mul.h>
+#include <ATen/ops/index_select.h>
+#include <ATen/ops/gather.h>
+#include <ATen/ops/nonzero.h>
+#include <ATen/ops/maximum.h>
+#include <ATen/ops/gt.h>
+#include <ATen/ops/div.h>
+#include <ATen/ops/ne.h>
+#include <ATen/ops/sum.h>
+#include <ATen/ops/linear_cross_entropy_backward_native.h>
+#include <ATen/ops/linear_cross_entropy_native.h>
+#endif
+
+namespace at::native {
+
+// Strategy selection for optimal chunking approach
+enum class ChunkingStrategy {
+    NAIVE,           // No chunking - standard approach
+    VOCAB_CHUNKING,  // Chunk vocabulary dimension (existing)
+    BATCH_CHUNKING   // Chunk batch dimension (new)
+};
+
+constexpr int64_t kDefaultVocabChunkSize = 4096;
+constexpr int64_t kDefaultBatchChunkSize = 1024;
+
+struct LinearCrossEntropySavedForBackward {
+  Tensor logsumexp;        // [N]
+  Tensor grad_input;
+  Tensor grad_weight;
+  Tensor grad_bias;
+  ChunkingStrategy strategy{ChunkingStrategy::NAIVE};
+  int64_t chunk_size{0};
+};
+
+struct LinearCrossEntropyForwardResult {
+  Tensor loss;
+  std::optional<LinearCrossEntropySavedForBackward> saved;
+};
+
+inline ChunkingStrategy select_chunking_strategy(c10::string_view strategy) {
+  if (strategy == "none") {
+    return ChunkingStrategy::NAIVE;
+  } else if (strategy == "vocab") {
+    return ChunkingStrategy::VOCAB_CHUNKING;
+  } else if (strategy == "batch") {
+    return ChunkingStrategy::BATCH_CHUNKING;
+  }
+  TORCH_CHECK(false,
+              "Unknown chunking strategy: ",
+              strategy,
+              ". Valid options: 'vocab', 'batch', 'none'");
+}
+
+LinearCrossEntropyForwardResult linear_cross_entropy_forward_naive(
+    const Tensor& input,
+    const Tensor& linear_weight,
+    const Tensor& target,
+    const std::optional<Tensor>& linear_bias_opt,
+    int64_t reduction,
+    int64_t ignore_index,
+    double label_smoothing,
+    bool save_for_backward);
+
+LinearCrossEntropyForwardResult linear_cross_entropy_forward_vocab_chunking(
+    const Tensor& input,
+    const Tensor& linear_weight,
+    const Tensor& target,
+    const std::optional<Tensor>& linear_bias_opt,
+    int64_t reduction,
+    int64_t ignore_index,
+    double label_smoothing,
+    int64_t chunk_size,
+    bool save_for_backward);
+
+LinearCrossEntropyForwardResult linear_cross_entropy_forward_batch_chunking(
+    const Tensor& input,
+    const Tensor& linear_weight,
+    const Tensor& target,
+    const std::optional<Tensor>& linear_bias_opt,
+    int64_t reduction,
+    int64_t ignore_index,
+    double label_smoothing,
+    int64_t chunk_size,
+    bool save_for_backward);
+
+std::tuple<Tensor, Tensor> _linear_cross_entropy_vocab_chunking_cpu(
+    const Tensor& input,
+    const Tensor& linear_weight,
+    const Tensor& target,
+    const std::optional<Tensor>& linear_bias_opt,
+    int64_t reduction,
+    int64_t ignore_index,
+    double label_smoothing,
+    int64_t chunk_size);
+
+std::tuple<Tensor, Tensor, std::optional<Tensor>> _linear_cross_entropy_vocab_chunking_backward_cpu(
+    const Tensor& grad_output,
+    const Tensor& saved_logsumexp,
+    const Tensor& input,
+    const Tensor& linear_weight,
+    const Tensor& target,
+    const std::optional<Tensor>& linear_bias_opt,
+    int64_t reduction,
+    int64_t ignore_index,
+    double label_smoothing,
+    int64_t chunk_size);
+
+std::tuple<Tensor, Tensor, Tensor, Tensor, Tensor, Tensor> _linear_cross_entropy_batch_chunking_cpu(
+    const Tensor& input,
+    const Tensor& linear_weight,
+    const Tensor& target,
+    const std::optional<Tensor>& linear_bias_opt,
+    int64_t reduction,
+    int64_t ignore_index,
+    double label_smoothing,
+    int64_t chunk_size);
+
+std::tuple<Tensor, Tensor, std::optional<Tensor>> _linear_cross_entropy_batch_chunking_backward_cpu(
+    const Tensor& grad_output,
+    const Tensor& saved_grad_input,
+    const Tensor& saved_grad_weight,
+    const Tensor& saved_grad_bias,
+    const Tensor& grad_weight_valid,
+    const Tensor& grad_bias_valid,
+    const Tensor& input,
+    const Tensor& linear_weight,
+    const Tensor& target,
+    const std::optional<Tensor>& linear_bias_opt,
+    int64_t reduction,
+    int64_t ignore_index,
+    double label_smoothing,
+    int64_t chunk_size);
+
+inline const Tensor& contiguous_if_needed(const Tensor& tensor, Tensor& buffer) {
+  if (tensor.is_contiguous()) {
+    return tensor;
+  }
+  buffer = tensor.contiguous();
+  return buffer;
+}
+
+static LinearCrossEntropyForwardResult linear_cross_entropy_forward_impl(
+    const Tensor& input,
+    const Tensor& linear_weight,
+    const Tensor& target,
+    const std::optional<Tensor>& linear_bias_opt,
+    int64_t reduction,
+    int64_t ignore_index,
+    double label_smoothing,
+    ChunkingStrategy strategy,
+    int64_t vocab_chunk_size,
+    int64_t batch_chunk_size,
+    bool save_for_backward) {
+
+  TORCH_CHECK(input.dim() >= 2, "Expected input to have at least 2 dimensions, got ", input.dim());
+  TORCH_CHECK(linear_weight.dim() == 2, "Expected linear_weight to be 2-dimensional, got ", linear_weight.dim());
+  TORCH_CHECK(input.size(-1) == linear_weight.size(1),
+              "Expected input.size(-1) to match linear_weight.size(1), got ",
+              input.size(-1), " and ", linear_weight.size(1));
+
+  switch (strategy) {
+    case ChunkingStrategy::VOCAB_CHUNKING:
+      TORCH_CHECK(
+          vocab_chunk_size > 0,
+          "linear_cross_entropy: vocab_chunk_size must be positive, got ",
+          vocab_chunk_size);
+      return linear_cross_entropy_forward_vocab_chunking(
+          input,
+          linear_weight,
+          target,
+          linear_bias_opt,
+          reduction,
+          ignore_index,
+          label_smoothing,
+          vocab_chunk_size,
+          save_for_backward);
+    case ChunkingStrategy::BATCH_CHUNKING:
+      TORCH_CHECK(
+          batch_chunk_size > 0,
+          "linear_cross_entropy: batch_chunk_size must be positive, got ",
+          batch_chunk_size);
+      return linear_cross_entropy_forward_batch_chunking(
+          input,
+          linear_weight,
+          target,
+          linear_bias_opt,
+          reduction,
+          ignore_index,
+          label_smoothing,
+          batch_chunk_size,
+          save_for_backward);
+    default:
+      return linear_cross_entropy_forward_naive(
+          input,
+          linear_weight,
+          target,
+          linear_bias_opt,
+          reduction,
+          ignore_index,
+          label_smoothing,
+          save_for_backward);
+  }
+}
+
+Tensor linear_cross_entropy_cpu(
+    const Tensor& input,
+    const Tensor& linear_weight,
+    const Tensor& target,
+    const std::optional<Tensor>& linear_bias_opt,
+    int64_t reduction,
+    int64_t ignore_index,
+    double label_smoothing,
+    c10::string_view chunking_strategy,
+    std::optional<int64_t> vocab_chunk_size_opt,
+    std::optional<int64_t> batch_chunk_size_opt) {
+
+  const int64_t vocab_chunk_size = vocab_chunk_size_opt.value_or(kDefaultVocabChunkSize);
+  const int64_t batch_chunk_size = batch_chunk_size_opt.value_or(kDefaultBatchChunkSize);
+  const ChunkingStrategy strategy = select_chunking_strategy(chunking_strategy);
+
+  auto result = linear_cross_entropy_forward_impl(
+      input,
+      linear_weight,
+      target,
+      linear_bias_opt,
+      reduction,
+      ignore_index,
+      label_smoothing,
+      strategy,
+      vocab_chunk_size,
+      batch_chunk_size,
+      /*save_for_backward=*/false);
+
+  return std::move(result.loss);
+}
+
+LinearCrossEntropyForwardResult linear_cross_entropy_forward_naive(
+    const Tensor& input,
+    const Tensor& linear_weight,
+    const Tensor& target,
+    const std::optional<Tensor>& linear_bias_opt,
+    int64_t reduction,
+    int64_t ignore_index,
+    double label_smoothing,
+    bool save_for_backward) {
+
+  Tensor input_buffer;
+  const Tensor& input_ref = contiguous_if_needed(input, input_buffer);
+  Tensor linear_weight_buffer;
+  const Tensor& linear_weight_ref = contiguous_if_needed(linear_weight, linear_weight_buffer);
+  Tensor target_buffer;
+  const Tensor& target_ref = contiguous_if_needed(target, target_buffer);
+  Tensor linear_bias_tensor;
+  if (linear_bias_opt.has_value()) {
+    linear_bias_tensor = linear_bias_opt.value();
+  }
+  Tensor linear_bias_buffer;
+  const Tensor& linear_bias_ref = linear_bias_tensor.defined()
+      ? contiguous_if_needed(linear_bias_tensor, linear_bias_buffer)
+      : linear_bias_tensor;
+  std::optional<Tensor> linear_bias_use;
+  if (linear_bias_ref.defined()) {
+    linear_bias_use = linear_bias_ref;
+  }
+
+  auto logits = at::linear(input_ref, linear_weight_ref, linear_bias_use);
+  auto logits_flat = logits.reshape({-1, logits.size(-1)});
+  auto target_flat = target_ref.reshape({-1});
+  auto options = logits_flat.options();
+  const int64_t vocab_size = linear_weight_ref.size(0);
+
+  Tensor valid_mask = at::ne(target_flat, ignore_index);
+  Tensor logsumexp = at::logsumexp(logits_flat, {1}, false);
+  Tensor target_logits = at::zeros(logsumexp.sizes(), options);
+
+  auto valid_indices = valid_mask.nonzero().reshape({-1});
+  if (valid_indices.numel() > 0) {
+    auto gathered_targets = at::index_select(target_flat, 0, valid_indices);
+    auto gathered_logits = at::index_select(logits_flat, 0, valid_indices);
+    auto gathered = at::gather(gathered_logits, 1, gathered_targets.unsqueeze(1)).squeeze(1);
+    target_logits.scatter_(0, valid_indices, gathered);
+  }
+
+  Tensor losses = logsumexp;
+  if (label_smoothing > 0.0) {
+    const double smoothing = label_smoothing;
+    const double uniform = smoothing / static_cast<double>(vocab_size);
+    auto sum_logits = logits_flat.sum(-1);
+    losses = losses - target_logits.mul(1.0 - smoothing) - sum_logits.mul(uniform);
+  } else {
+    losses = losses - target_logits;
+  }
+
+  losses.masked_fill_(at::logical_not(valid_mask), 0);
+
+  Tensor loss_result;
+  if (reduction == Reduction::None) {
+    loss_result = losses.reshape(target.sizes());
+  } else {
+    auto total_loss = losses.sum();
+    if (reduction == Reduction::Sum) {
+      loss_result = total_loss;
+    } else {
+      auto denom_long = valid_mask.sum();
+      auto denom = denom_long.to(total_loss.scalar_type());
+      if (denom_long.item<int64_t>() == 0) {
+        loss_result = denom.div(denom);
+      } else {
+        loss_result = total_loss.div(denom);
+      }
+    }
+  }
+
+  LinearCrossEntropyForwardResult result;
+  result.loss = std::move(loss_result);
+
+  if (save_for_backward) {
+    LinearCrossEntropySavedForBackward saved;
+    saved.logsumexp = std::move(logsumexp);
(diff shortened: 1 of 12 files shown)


vadimkantorov:
@albanD also some existing implementations might have been developed in torchtune:
- https://github.com/pytorch/pytorch/issues/163408

and maybe torchtitan

If a C++ version gets added in Core, might also be good to also preserve some Python/torch.compilable or Triton example in the docs page (as "reference" implementation - could be flyout which is not shown by default)? This is useful as probably this trick will be tried for other kinds of losses - or for computing multiple losses/logits stats at the same time - so having a hackable "reference" version would be useful.

Skylion007:
Will either of these support the dtype aware cut cross entropy versions of this kernel? Or just the non-dtype aware original liger implementation. https://github.com/apple/ml-cross-entropy.

TLDR: you can exploit the low precision dtypes to do even less work in this kernel in the common cases: https://github.com/apple/ml-cross-entropy

aten/src/ATen/native/LinearCrossEntropy.cpp
```diff
@@ -0,0 +1,664 @@
...
+      grad_bias.add_(grad_chunk.sum(0));
+    }
+  }
+
+  grad_input = grad_input.view_as(input);
+  std::optional<Tensor> grad_bias_opt;
+  if (grad_bias.defined()) {
+    grad_bias_opt = std::move(grad_bias);
+  }
+  return std::make_tuple(grad_input, grad_weight, std::move(grad_bias_opt));
```
Skylion007: All these should have std::move called on it

test/nn/test_linear_cross_entropy.py
```diff
@@ -0,0 +1,195 @@
...
+    target_flat = target.view(-1)
+    loss = F.cross_entropy(
+        logits_flat,
+        target_flat,
+        reduction=reduction,
+        ignore_index=ignore_index,
+        label_smoothing=label_smoothing,
+    )
+    if reduction == "none":
+        loss = loss.view(target.shape)
```
vadimkantorov: could be `loss = loss.view_as(target)`?

test/nn/test_linear_cross_entropy.py
```diff
@@ -29,7 +31,7 @@ def _reference_linear_cross_entropy(
         label_smoothing=label_smoothing,
     )
     if reduction == "none":
-        loss = loss.view(target.shape)
+        loss = loss.reshape(target.shape)
```
vadimkantorov: or maybe `loss = reshape_as(target)`, if reshape must be used...

torch/testing/_internal/common_methods_invocations.py
```diff
@@ -14753,8 +14826,8 @@ def sample_inputs_alias_copy(op_info, device, dtype, requires_grad, **kwargs):
         dtypes=floating_types_and(torch.float16, torch.bfloat16),
         sample_inputs_func=sample_inputs_cross_entropy,
         supports_out=False,
-        supports_forward_ad=True,
-        supports_fwgrad_bwgrad=True,
+        supports_forward_ad=False,
+        supports_fwgrad_bwgrad=False,
```
benjaminglass1: @krocki Looks like an accidental change here.

test/nn/test_linear_cross_entropy.py
```diff
@@ -0,0 +1,256 @@
...
+        )
+
+    def test_vocab_chunking(self) -> None:
+        torch.manual_seed(0)
+        input = torch.randn(4, 16, requires_grad=True)
+        weight = torch.randn(5000, 16, requires_grad=True)
+        target = torch.randint(0, 5000, (4,))
+        self._compare_with_reference(
+            input, weight, target, None, chunking_strategy="vocab"
+        )
```
benjaminglass1: @krocki With the new OpInfo implementation, you should be able to remove all the tests in this file that do `_compare_with_reference`.

aten/src/ATen/native/LinearCrossEntropy.cpp
```diff
@@ -0,0 +1,652 @@
...
+    c10::string_view strategy) {
+
+    if (strategy == "none") {
+        return ChunkingStrategy::NAIVE;
+    } else if (strategy == "vocab") {
+        return ChunkingStrategy::VOCAB_CHUNKING;
+    } else if (strategy == "batch") {
+        return ChunkingStrategy::BATCH_CHUNKING;
+    } else if (strategy == "auto") {
+        // Empirically validated chunk sizes for optimal memory/compute balance
```
mikaylagawarecki: Where are these numbers from?

I'm not sure there is precedence for providing a heuristic as the default, would it make sense to make the user explicitly choose the strategy they want to use instead?

aten/src/ATen/native/native_functions.yaml
```diff
@@ -9488,6 +9488,14 @@
   dispatch:
     CompositeImplicitAutograd: cross_entropy_loss_symint
 
+- func: linear_cross_entropy(Tensor input, Tensor weight, Tensor target, Tensor? bias=None, int reduction=Mean, SymInt ignore_index=-100, float label_smoothing=0.0, str chunking_strategy="auto") -> Tensor
```
mikaylagawarecki: nit: 

```suggestion
- func: linear_cross_entropy(Tensor self, Tensor weight, Tensor target, Tensor? bias=None, int reduction=Mean, SymInt ignore_index=-100, float label_smoothing=0.0, str chunking_strategy="auto") -> Tensor
```

mikaylagawarecki: Since `cross_entropy_loss` also has a parameter called `weight`, to be clearer herer, wdyt of `weight` --> `linear_weight` and `bias` --> `linear_bias`

mikaylagawarecki: To make it easier to add more kwargs in the future, if necessary

```suggestion
- func: linear_cross_entropy(Tensor input, Tensor weight, Tensor target, *, Tensor? bias=None, int reduction=Mean, SymInt ignore_index=-100, float label_smoothing=0.0, str chunking_strategy="auto") -> Tensor
```

Also I think this should return the tensors we save for backward in addition to the result

aten/src/ATen/native/native_functions.yaml
```diff
@@ -9488,6 +9488,14 @@
...
+  dispatch:
+    CPU: linear_cross_entropy_cpu
+
+- func: linear_cross_entropy_backward(Tensor grad_output, Tensor input, Tensor weight, Tensor target, Tensor? bias=None, int reduction=Mean, SymInt ignore_index=-100, float label_smoothing=0.0, str chunking_strategy="auto") -> (Tensor, Tensor, Tensor?)
```
mikaylagawarecki:
```suggestion
- func: linear_cross_entropy_backward(Tensor grad_output, Tensor input, Tensor weight, Tensor target, *, Tensor? bias=None, int reduction=Mean, SymInt ignore_index=-100, float label_smoothing=0.0, str chunking_strategy="auto") -> (Tensor, Tensor, Tensor?)
```

torch/nn/functional.py
```diff
@@ -3503,6 +3503,156 @@ def cross_entropy(
...
+            reduction,
+            ignore_index,
+            label_smoothing,
+        )
+    else:
+        op = torch.ops.aten.linear_cross_entropy.default
+        # Only exercise the fused path when the operator is actually built for
+        # this runtime; otherwise fall back to the explicit composition so the
+        # behaviour matches older binaries.
+        if not op.has_kernel_for_dispatch_key("CPU"):
```
mikaylagawarecki: Why is this needed / what is the precedence for this?

torch/nn/functional.py
```diff
@@ -3503,6 +3503,156 @@ def cross_entropy(
...
+            reduction_enum = _Reduction.get_enum(reduction)
+            needs_grad = torch.is_grad_enabled() and (
+                input.requires_grad
+                or weight.requires_grad
+                or (bias is not None and bias.requires_grad)
+            )
+            # Some downstream builds may omit the generated autograd kernel; if
+            # that happens we still provide gradients by delegating to the
+            # unfused implementation.
+            if needs_grad and not op.has_kernel_for_dispatch_key("AutogradCPU"):
```
mikaylagawarecki: Similarly I'm not quite sure why we need this

torch/nn/functional.py
```diff
@@ -3503,6 +3503,156 @@ def cross_entropy(
...
+    )
+    if reduction == "none":
+        loss = loss.reshape(target.shape)
+    return loss
+
+
+def linear_cross_entropy(
+    input: Tensor,
+    weight: Tensor,
+    target: Tensor,
```
mikaylagawarecki, line 3546:
```suggestion
    target: Tensor,
    *,
```

torch/_decomp/decompositions.py
```diff
@@ -39,6 +39,11 @@
 
 DispatchKey = torch._C.DispatchKey  # type: ignore[attr-defined]
 
+
+_LINEAR_CROSS_ENTROPY_NAIVE: Optional[Callable[..., Tensor]] = getattr(
```
mikaylagawarecki: similarly I'm not sure why we need this, when could the version of decompositions.py and torch/nn/functional.py be out of sync?

aten/src/ATen/native/LinearCrossEntropy.cpp
```diff
@@ -0,0 +1,652 @@
...
+    int64_t ignore_index,
+    double label_smoothing) {
+
+  // Flatten multi-dimensional inputs for processing (standard PyTorch pattern)
+  // This allows handling both 2D [batch, hidden] and 3D [batch, seq, hidden] inputs
+  auto input_flat = input.reshape({-1, input.size(-1)});  // [N, H] where N = batch * seq_len
+  auto target_flat = target.reshape({-1});                // [N] flattened targets
+
+    const int64_t batch_size = input_flat.size(0);
+    const int64_t chunk_size = 1024;  // Empirically optimized for batch dimension chunking
```
mikaylagawarecki: Does the ideal number here also depend on the value of `H`? 

I think we should expose this as a config to the user

aten/src/ATen/native/LinearCrossEntropy.cpp
```diff
@@ -0,0 +1,652 @@
...
+  // Select optimal chunking strategy based on input characteristics and user preference
+  ChunkingStrategy selected_strategy = select_chunking_strategy(vocab_size, flattened_batch, chunking_strategy);
+
+  auto input_flat = input.reshape({-1, input.size(-1)});  // [N, H]
+  auto target_flat = target.reshape({-1});                // [N]
+  auto valid_mask = at::ne(target_flat, ignore_index);
+
+  // Execute selected chunking strategy
+  if (selected_strategy == ChunkingStrategy::VOCAB_CHUNKING) {
+    const int64_t chunk_size = 4096;  // Empirically validated chunk size for optimal memory/compute balance
```
mikaylagawarecki: This should be configurable by the end user

aten/src/ATen/native/LinearCrossEntropy.cpp
```diff
@@ -0,0 +1,652 @@
...
+    auto chunk_max = std::get<0>(logits_chunk.max(-1));
+    auto new_max = at::maximum(running_max, chunk_max);
+    auto exp_scale_old = at::exp(running_max.sub(new_max));
+    auto shifted_logits = logits_chunk.sub(new_max.unsqueeze(-1));
+    auto exp_chunk = at::sum(at::exp(shifted_logits), {-1});
+    exp_sums = at::add(at::mul(exp_sums, exp_scale_old), exp_chunk);
+    running_max = new_max;
+  }
+
+  Tensor logsumexp = running_max.add(exp_sums.log());
```
mikaylagawarecki: Why do we choose to recompute `logsumexp` rather than saving it from forward (by returning it from the fwd function)? 

This seems different from the reference and would increase the amount of recompute in the backward (which has been carefully tuned in the reference impl I shared)

aten/src/ATen/native/LinearCrossEntropy.cpp
```diff
@@ -0,0 +1,652 @@
...
+  const double uniform_component = label_smoothing > 0.0 ? label_smoothing / static_cast<double>(weight.size(0)) : 0.0;
+  const int64_t total = input_flat.size(0);
+
+  for (int64_t start_idx = 0; start_idx < total; start_idx += chunk_size) {
+    const int64_t slice = std::min<int64_t>(chunk_size, total - start_idx);
+    auto input_chunk = input_flat.narrow(0, start_idx, slice);
+    auto target_chunk = target_flat.narrow(0, start_idx, slice);
+    auto valid_mask_chunk = valid_mask.narrow(0, start_idx, slice);
+    auto logits_chunk = at::linear(input_chunk, weight, bias_opt);
+    auto logsumexp_chunk = at::_ops::logsumexp::call(logits_chunk, std::vector<int64_t>{1}, false);
```
mikaylagawarecki: Why `at::_ops::...::call` here?

aten/src/ATen/native/LinearCrossEntropy.cpp
```diff
@@ -0,0 +1,652 @@
...
+  if (grad_bias.defined()) {
+    grad_bias_opt = std::move(grad_bias);
+  }
+  return std::make_tuple(std::move(grad_input), std::move(grad_weight), std::move(grad_bias_opt));
+}
+
+// Computes gradients when we chunked the batch dimension (or not at all).  The
+// loop keeps the working set bounded by `chunk_size` rows so that we never
+// allocate a full [N, vocab] buffer even when the batch is very large.
+inline std::tuple<Tensor, Tensor, std::optional<Tensor>> backward_batch_chunking(
```
mikaylagawarecki, line 902: My understanding of the Liger-style batch chunking is that the gradients (up to a scaling factor) are [computed in the forward and saved for backward ](https://github.com/linkedin/Liger-Kernel/blob/2942b58f1cb6723fc89e0e339b1e175ebfb6b5ef/src/liger_kernel/ops/fused_linear_cross_entropy.py#L318-L322).

This looks different from what is done here though, could you please help me understand why

aten/src/ATen/native/LinearCrossEntropy.cpp
```diff
@@ -0,0 +1,652 @@
...
+      }
+
+      auto chunk_max = std::get<0>(logits_chunk.max(-1));
+      auto new_max = at::maximum(running_max, chunk_max);
+
+      auto exp_scale_old = at::exp(at::sub(running_max, new_max));
+      auto shifted_logits = at::sub(logits_chunk, new_max.unsqueeze(-1));
+      auto exp_chunk = at::sum(at::exp(shifted_logits), {-1});
+      exp_sums = at::add(at::mul(exp_sums, exp_scale_old), exp_chunk);
+      running_max = new_max;
```
mikaylagawarecki: To confirm my understanding -- the ref impl took log every iteration, but it's ok not to because we already have decent guarantees of numerical stability here (since the max is being subtracted)

Is that the right understanding?

aten/src/ATen/native/LinearCrossEntropy.cpp
```diff
@@ -0,0 +1,652 @@
...
+    if (strategy == "none") {
+        return ChunkingStrategy::NAIVE;
+    } else if (strategy == "vocab") {
+        return ChunkingStrategy::VOCAB_CHUNKING;
+    } else if (strategy == "batch") {
+        return ChunkingStrategy::BATCH_CHUNKING;
+    } else if (strategy == "auto") {
+        // Empirically validated chunk sizes for optimal memory/compute balance
+        const int64_t vocab_chunk_size = 4096;   // Same as existing implementation
+        const int64_t batch_chunk_size = 1024;   // Optimized for batch processing
```
mikaylagawarecki: Where is 1024 from?

I imagine the hardcoded numbers here would vary based on hardware etc. so I hesitant to have an "auto" option that uses a heuristic to choose the default. 

I would prefer to remove "auto" from the initial PR and land it in a follow up PR if we decide to go with it

mikaylagawarecki (commented):
Thanks @krocki for your hard work on this PR. I left a first pass of comments -- a couple overall comments

1. I am hesitant about `chunking_strategy="auto"` as an option and also as the default, let's remove it from this PR
2. Let's make `chunk_size` configurable from the frontend
3. If I understand correctly, the backward implementations in this PR are not "saving for backward" the same tensors that the reference / liger implementations do. I think we want to match the recompute behavior of the references and it's not super clear to me why we don't here. Do feel free to clarify if there are things I overlooked here (broader api surface forcing us to need to recompute some things or inconsistency with other conventions within core etc.).
4. I'm reviewing the vocab chunking impl by naively comparing against the reference so it would be helpful to me if you could add PR comments on the parts that differ and why  😅  (e.g. computing sum_exps and taking log at the end vs log_sum_exp of every chunk and re-exponentiating every time) Thank you!!

I'm not sure if there is a good way to validate the memory properties of the CPU kernel, but when you get to adding the CUDA impl using a [memory snapshot](https://docs.pytorch.org/docs/stable/torch_cuda_memory.html) to validate it would be great

aten/src/ATen/native/native_functions.yaml
krocki: this should be fixed now

krocki: this should be fixed now

krocki: committed a revised version

aten/src/ATen/native/native_functions.yaml
krocki: this should be fixed now

torch/nn/functional.py
krocki: ok

aten/src/ATen/native/native_functions.yaml
```diff
@@ -9568,6 +9568,14 @@
   dispatch:
     CompositeImplicitAutograd: cross_entropy_loss_symint
 
+- func: linear_cross_entropy(Tensor self, Tensor linear_weight, Tensor target, *, Tensor? linear_bias=None, int reduction=Mean, SymInt ignore_index=-100, float label_smoothing=0.0, str chunking_strategy="none", int? vocab_chunk_size=None, int? batch_chunk_size=None) -> Tensor
```
mikaylagawarecki, line 9613: hey, I didn't go over your changes in detail but noticed that you were using a custom autograd function in ATen

The proper way to save a tensor for backward is to return a tuple of Tensors from the function here

We can use a private _linear_cross_entropy and call it from the python `linear_cross_entropy` in order to only return the output we want in forward

Also please please feel free to ping me on slack or respond to my comments if they are unclear, I am sorry that you had to spend time on the custom autograd function

aten/src/ATen/native/LinearCrossEntropy.cpp
krocki: Chunking is now driven by explicit vocab_chunk_size / batch_chunk_size kwargs that flow from Python

aten/src/ATen/native/LinearCrossEntropy.cpp
krocki: Backward now saves the same tensors as the reference implementation. Forward caches the per-row logsumexp (and strategy metadata) whenever Autograd is active

aten/src/ATen/native/LinearCrossEntropy.cpp
krocki: Removed at::_ops::logsumexp::call and other at::_ops uses (the code now calls the public at::logsumexp, etc.)

aten/src/ATen/native/LinearCrossEntropy.cpp
krocki: Batch/vocab chunk sizes now come from the schema kwargs, so there’s no hidden 1024 default.

aten/src/ATen/native/LinearCrossEntropy.cpp
krocki: Thanks for double-checking! Yes, that’s the intent—the loop maintains a running max so every chunk is shifted by the largest logit seen so far, we accumulate exp_sums in that normalized scale, and only take the log once at the end. Algebraically it matches the reference’s per-chunk logsumexp, but we avoid extra log/exp work while keeping the same stability guarantees.

aten/src/ATen/native/native_functions.yaml
krocki: I looked at that option, but keeping a single public schema that just returns the loss felt cleaner than introducing a second _linear_cross_entropy variant. The bespoke Autograd node now saves only the per-row logsumexp (plus the strategy metadata) and reuses those in backward, so we match the reference recompute budget.

torch/_decomp/decompositions.py
benjaminglass1: @krocki Let's go ahead and remove this entirely, you can assume the torch repo files update in sync with each other.

torch/nn/functional.py
```diff
@@ -3510,6 +3510,170 @@ def cross_entropy(
...
+            reduction=reduction,
+            ignore_index=ignore_index,
+            label_smoothing=label_smoothing,
+            chunking_strategy=chunking_strategy,
+            vocab_chunk_size=vocab_chunk_size,
+            batch_chunk_size=batch_chunk_size,
+        )
+
+    if not isinstance(reduction, str):
+        if hasattr(reduction, "node"):
```
mikaylagawarecki, line 3582: why is this needed whereas other functions in this file that take in string arguments (e.g. `embedding_bag`) don't have this sort of logic

torch/nn/functional.py
```diff
@@ -3510,6 +3510,170 @@ def cross_entropy(
...
+        result = _linear_cross_entropy_naive(
+            input,
+            linear_weight,
+            target,
+            linear_bias,
+            reduction,
+            ignore_index,
+            label_smoothing,
+        )
+    else:
```
mikaylagawarecki, line 3633: please remove this, if the impl is only registered to the CPU key (as it currently is) the dispatcher will throw an error if the inputs are on another device

torch/nn/functional.py
```diff
@@ -3510,6 +3510,170 @@ def cross_entropy(
...
+        if not supports_chunk_sizes:
+            result = _linear_cross_entropy_naive(
+                input,
+                linear_weight,
+                target,
+                linear_bias,
+                reduction,
+                ignore_index,
+                label_smoothing,
+            )
```
mikaylagawarecki: similarly here, this is not needed, schema always has `vocab_chunk_size` ?

torch/nn/functional.py
```diff
@@ -3510,6 +3510,170 @@ def cross_entropy(
...
+        else:
+            reduction_enum = _Reduction.get_enum(reduction)
+            try:
+                result = torch.ops.aten.linear_cross_entropy(
```
mikaylagawarecki: the way to call this following the convention in this file would be `torch._C._nn.linear_cross_entropy`

torch/nn/functional.py
```diff
@@ -3510,6 +3510,170 @@ def cross_entropy(
...
+                linear_weight,
+                target,
+                linear_bias,
+                reduction,
+                ignore_index,
+                label_smoothing,
+            )
+        else:
+            reduction_enum = _Reduction.get_enum(reduction)
+            try:
```
mikaylagawarecki: please remove the try except, as above it's ok for the dispatcher to throw an unhandled error if the implementation has not been registered to the corresponding backend key yet.

torch/nn/functional.py
```diff
@@ -3510,6 +3510,170 @@ def cross_entropy(
...
+    batch_chunk_size: Optional[int] = None,
+) -> Tensor:
+    r"""Compute fused linear transformation and cross entropy loss on CPU.
+
+    This is a convenience wrapper around :func:`linear` followed by
+    :func:`cross_entropy`.  When the inputs live on CPU it uses a fused ATen
+    kernel that chunks the vocabulary or batch dimension to avoid materialising
+    large logit tensors; the chunk sizes can be overridden with
+    ``vocab_chunk_size`` and ``batch_chunk_size``.  For other devices it falls
+    back to the unfused composition.
```
mikaylagawarecki, line 3563: Per below comments, this does not need to fall back to unfused for other devices, also please document the arguments like how other docstrings in this file are written

aten/src/ATen/native/LinearCrossEntropy.cpp
```diff
@@ -0,0 +1,1103 @@
...
+
+    logsumexp = running_max.add(exp_sums.log());
+  }
+
+  Tensor grad_input = zeros_like_tensor(input_flat);
+  Tensor grad_weight = zeros_like_tensor(linear_weight);
+  Tensor grad_bias = zeros_like_or_undef(linear_bias_opt);
+
+  const double uniform_component = label_smoothing > 0.0 ? label_smoothing / static_cast<double>(vocab_size) : 0.0;
+  Tensor grad_output_tensor = cast_grad_output(grad_output, input);
```
mikaylagawarecki: why this cast, when do we expect the type of grad_output to differ from input?

aten/src/ATen/native/LinearCrossEntropy.cpp
```diff
@@ -0,0 +1,1103 @@
...
+
+  const auto options = input.options();
+  const Tensor& linear_bias = linear_bias_opt.value_or(Tensor());
+
+  TORCH_CHECK(chunk_size > 0, "linear_cross_entropy: batch_chunk_size must be positive, got ", chunk_size);
+  const int64_t total = input_flat.size(0);
+  const int64_t num_chunks = (total + chunk_size - 1) / chunk_size;
+
+  Tensor logsumexp;
+  if (has_saved) {
```
mikaylagawarecki: similarly here, when would this ever be False

aten/src/ATen/native/LinearCrossEntropy.cpp
```diff
@@ -0,0 +1,1103 @@
...
+        linear_weight,
+        target,
+        linear_bias_opt,
+        grad_output,
+        reduction,
+        ignore_index,
+        label_smoothing,
+        vocab_chunk_size_opt.value_or(kDefaultVocabChunkSize),
+        Tensor(),
+        /*has_saved=*/false);
```
mikaylagawarecki, line 1076: Why does the derivatives.yaml registration always goes through the `/*has_saved=*/false` path?

aten/src/ATen/native/LinearCrossEntropy.cpp
mikaylagawarecki: I think we want to follow the Liger approach here and save `grad_input`, `grad_weight` (if needed), and `grad_bias` (if needed) for backward so that we do not need to recompute the logits in backward. Further, we want to [write `grad_logits` in-place to the logits](https://github.com/linkedin/Liger-Kernel/blob/2942b58f1cb6723fc89e0e339b1e175ebfb6b5ef/src/liger_kernel/ops/fused_linear_cross_entropy.py#L134-L136) during the forward computation to save memory.

aten/src/ATen/native/native_functions.yaml
mikaylagawarecki: To clarify, we want to use a different recompute budget for batch chunking vs vocab chunking. This is an intentional decision to tradeoff memory and recompute. We can save more for batch chunking to avoid the logit recompute (see the other comment).

mikaylagawarecki (commented):
Thanks for the progress!

As a best practice, let us not use a custom autograd function within ATen, even if this requires multiple schemas to accommodate (these can be private by prefixing them with an underscore).

If the schemas are unclear, please ping me on Slack, and we can discuss before you make any implementation changes.

Also, sorry for any confusion around what we want for batch chunking vs. vocab chunking. We want batch chunking to save more and recompute less (following the Liger approach), and the vocab chunking can save less and recompute more. This is to give users optionality on their memory tradeoff.

test/nn/test_linear_cross_entropy.py
krocki: fixed, moved all _compare_with_reference tests to opinfo

torch/nn/functional.py
krocki: removed this

torch/nn/functional.py
krocki: removed

mikaylagawarecki:
@krocki 2 things from our meeting earlier

Firstly, if you find a different save vs recompute decision to be more optimal than what the liger implementation does based on the review of other similar kernels that you've done do feel free to raise that! :)


Secondly, just to give some guidance in case it's unclear how the ops should be registered

```
# In native_functions.yaml
- func: _linear_cross_entropy_vocab_chunking(...) -> (resultTensor, n items to save for backward)
    - CPU: ... # CompositeExplicit

- func: _linear_cross_entropy_batch_chunking(...) -> (resultTensor, m items to save for backward)
    - CPU: ... # CompositeExplicit
    
 # in derivatives.yaml
 register derivative formulas for both
```

From python

```
 def linear_cross_entropy(...): 
      // calls appropriate torch._C.linear_cross_entropy_foo based on barious conditions
```


You can take a look at the [`scaled_dot_product_attention` variants](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/native_functions.yaml#L15112-L15194) in native_functions.yaml for an example of this pattern (in the case of scaled_dot_product_attention, the [top level function that dispatches into the other ops is written in C and registered as a CompositeImplicit function in native_functions.yaml](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/transformers/attention.cpp?fbclid=IwY2xjawOE4lFleHRuA2FlbQIxMQBzcnRjBmFwcF9pZAEwCGNhbGxzaXRlATcAAR7iQUpHji35q4uR1jRNXV3fu0Ae6WPmQCctrhf2tel2QcptArMQp0oGFWBsVQ_aem_6dkK6tF4YoytNyJgiVI83A#L718), but that's not actually necessary and we can just do that part in python for this case is my understanding.

PR #163779, opened 2025-09-24
[aoti] Load metadata w/o loading package

angelayi:
Add a function to load the metadata stored in aoti without needing to load the .so. This can be used to store what platform we are compiling the .so on which we can check before loading the .so.

* #163792
* __->__ #163779


yushangdi (approved):

desertfire (approved):

test/inductor/test_aot_inductor_package.py
```diff
@@ -579,14 +579,39 @@ def forward(self, x, y):
...
+            with WritableTempFile(suffix=".pt2") as f:
+                ep = torch.export.export(model, example_inputs, strict=False)
+                package_path = torch._inductor.aoti_compile_and_package(
+                    ep, package_path=f.name, inductor_configs=inductor_configs
+                )  # type: ignore[arg-type]
+
+                # We can load the metadata w/o loading the actual package
+                loaded_metadata = (
+                    torch._C._aoti.AOTIModelPackageLoader.load_metadata_from_package(
+                        package_path, "model"
```
sayakpaul, line 604: How is the "model" string determined here?

angelayi: When you call [package_aoti](https://github.com/pytorch/pytorch/blob/c8e75c48b958a1a7b9f4a7ba168455b3cffc93be/torch/_inductor/package/package.py#L82) or [package_pt2](https://github.com/pytorch/pytorch/blob/main/torch/export/pt2_archive/_package.py#L543) directly to package your AOTI files, you can specify a mapping of model name to AOTI files, allowing you to package multiple AOTI files in one .pt2 archive. If you just have one model then by default the name is `"model"`.

PR #163781, opened 2025-09-24
[Triton] [Inductor] Prune template selection based on decompose_k

njriasan:
Summary:

Triton templates tend to perform very poorly on large K, hence the introduction of decompose_k. As a result, when decompose_k is selected will disable exploring the Triton templates. We may want to consider an override in the future. 

Note: Based on the timing results it may be desirable to better refine/prune the decompose k decisions.

Testing:

Tested by looking at the autotune/compilation time using a single shape in TritonBench.
`TORCHINDUCTOR_FORCE_DISABLE_CACHES=1 ENABLE_PERSISTENT_TMA_MATMUL=1 python run --op gemm --rep 1000 --sleep 1.0 --m 512 --n 512 --k 300000 --only pt2_matmul_maxautotune`
Before this change:
`SingleProcess AUTOTUNE benchmarking takes 13.5368 seconds and 0.1595 seconds precompiling for 38 choices`
With this change:
`SingleProcess AUTOTUNE benchmarking takes 9.9626 seconds and 0.0020 seconds precompiling for 11 choices`

### DIFF
--- a/test/inductor/test_max_autotune.py
+++ b/test/inductor/test_max_autotune.py
@@ -1562,7 +1562,9 @@ class TestMaxAutotune(TestCase):
         with mock.patch(
             "torch._inductor.kernel.mm.use_decompose_k_choice"
         ) as decomp_mock:
-            decomp_mock.return_value = True
+            decomp_mock.side_effect = (
+                lambda *args, **kwargs: kwargs.get("threshold_multiple", 1) == 1
+            )
 
             out, code = run_and_get_code(compiled_func, a, b)
             FileCheck().check("extern_kernels.bmm_dtype").check_regex(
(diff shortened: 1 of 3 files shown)


PaulZhang12:
Can we make it a bit more strict here? Like use_decompose_k has some threshold for M, N. In the case where we only want to autotune decomposeK, can we make it so that the threshold is even greater? Like instead of just 32 * M, N for using decomposeK, only autotuning decomposeK should be like 64 * M, N for example

PaulZhang12:
I have no data here, but just as a first iteration, we should be a little safer imo, but also ok with this change cc @eellison

eellison (approved):
Looks good to me, although maybe we should keep them if exhaustive is set.

PaulZhang12 (approved):
Looks great!

njriasan:
@pytorch merge

PR #163784, opened 2025-09-24
[DO NOT LAND] Make _flat_weights in RNN ParameterList

yiming0416:
Summary: Removed `_update_flat_weights()` in the `forward`. Let's see what CI says.

Test Plan: buck2 run mode/dev-nosan caffe2/test:test_export -- -r test_lstm_export

Differential Revision: D83170973


albanD (changes requested):
These can't be Parameters, they would end up on the state dict, etc

PR #163782, opened 2025-09-24
Add smoke tests to verify that stable ABI FA3 wheel runs w/ newer torch

janeyx99:
Passing CI: https://github.com/pytorch/pytorch/actions/runs/18141589975/job/51635340255?pr=163782

Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #163782

PR #163795, opened 2025-09-24
Add H100 runner to be recognized in actionlint

janeyx99:
* #163782
* __->__ #163795

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -23,6 +23,8 @@ self-hosted-runner:
     - linux.arm64.m7g.4xlarge.ephemeral
     - linux.arm64.r7g.12xlarge.memory
     - linux.aws.h100
+    - linux.aws.h100.4
+    - linux.aws.h100.8
     - linux.4xlarge.nvidia.gpu
     - linux.8xlarge.nvidia.gpu
     - linux.16xlarge.nvidia.gpu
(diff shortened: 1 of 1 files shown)


.github/actionlint.yaml
```diff
@@ -22,6 +22,7 @@ self-hosted-runner:
     - linux.arm64.m7g.4xlarge
     - linux.arm64.m7g.4xlarge.ephemeral
     - linux.arm64.r7g.12xlarge.memory
+    - linux.aws.h100
```
huydhn, line 25: There are also `linux.aws.h100.4` and `linux.aws.h100.8`

huydhn (approved):

.github/actionlint.yaml
seemethere: weird that they weren't already in there.

I feel like this may be a hole since we probably didn't use these in a `runs-on` context before

seemethere, line 25:
```suggestion
    - linux.aws.h100
    - linux.aws.h100.4
    - linux.aws.h100.8
```

seemethere (approved):
LGTM, would be nice to have all of the h100 skus

.github/actionlint.yaml
huydhn: After chatting with @janeyx99 earlier, I suspect that the loophole is that those runners are set in the test matrix's `runner`, i.e. https://github.com/pytorch/pytorch/blob/1495b35d29512f303ab37780760c5e692158514b/.github/workflows/h100-symm-mem.yml#L45 which the lint job doesn't check.  Only when you put them directly into `runs-on` that the actionlint recognizes them, i.e. `runs-on: linux.aws.h100`

PR #163803, opened 2025-09-24
[Inductor] add a new config fallback_embedding_bag_byte_unpack

hl475:
Differential Revision: D82988783

introduce an inductor config fallback_embedding_bag_byte_unpack so we can have options to not let inductor decompose the op

### DIFF
--- a/torch/_inductor/decomposition.py
+++ b/torch/_inductor/decomposition.py
@@ -885,6 +885,13 @@ def select_decomp_table() -> dict[Any, Callable[..., Any]]:
     """decomps can change based on config"""
     if config.fallback_random:
         return decompositions
+    if config.fallback_embedding_bag_byte_unpack is True:
+        # remove q_embedding_bag_byte_unpack_decomp from decompositions
+        for k in decompositions.keys():
+            if k == torch.ops.quantized.embedding_bag_byte_unpack.default:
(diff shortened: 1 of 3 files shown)


henryoier (approved):

torch/_inductor/decomposition.py
```diff
@@ -885,6 +885,13 @@ def select_decomp_table() -> dict[Any, Callable[..., Any]]:
     """decomps can change based on config"""
     if config.fallback_random:
         return decompositions
+    if config.fallback_embedding_bag_byte_unpack is True:
```
desertfire, line 888: Nit: `if config.fallback_embedding_bag_byte_unpack:`

torch/_inductor/decomposition.py
```diff
@@ -885,6 +885,13 @@ def select_decomp_table() -> dict[Any, Callable[..., Any]]:
...
+        # remove q_embedding_bag_byte_unpack_decomp from decompositions
+        for k in decompositions.keys():
+            if k == torch.ops.quantized.embedding_bag_byte_unpack.default:
+                del decompositions[k]
+                break
```
desertfire, line 893: `decompositions.pop(torch.ops.quantized.embedding_bag_byte_unpack.default, None)`

Issue #163798, opened 2025-09-24
tolist() does not graph when capture_scalar_outputs = False


laithsakka:
repo:
```
@torch.compile(fullgraph=False, backend="eager")
def func(a):
    u0, u1 = a.tolist()
    return a*u0*u1
func(torch.tensor([1,2]))
```

output graph:
```
class GraphModule(torch.nn.Module):
    def forward(self, L_a_: "i64[2][1]cpu"):
        l_a_ = L_a_
        
         # File: /home/lsakka/pytorch/pytorch/example.py:7 in func, code: u0, u1 = a.tolist()
        getitem = l_a_[0]
        item: "Sym(u0)" = getitem.item();  getitem = None
        getitem_1 = l_a_[1]
        item_1: "Sym(u1)" = getitem_1.item();  getitem_1 = None
        
         # File: /home/lsakka/pytorch/pytorch/example.py:8 in func, code: return a*u0*u1
        mul: "i64[2][1]cpu" = l_a_ * item;  l_a_ = item = None
        mul_1: "i64[2][1]cpu" = mul * item_1;  mul = item_1 = None
        return (mul_1,)
        
```
in contrast, the following program does graph break.
```
@torch.compile(fullgraph=False, backend="eager")
def func2(a):
    u0 = a.item()
    return a*u0
```
func2(torch.tensor([1]))

This is problematic because now we have item() in the graphs then we might re-trace 
in autograd and fail because capture_scalar_outputs=False and its too late to know where to graph break!
seen this recently with Ads cc @chauhang @penguinwu @ezyang @bobrenjc93 @pianpwk 

Note that with inductor backend fails to compile !
```
@torch.compile(fullgraph=False, backend="inductor")
def func(a):
    u0, u1 = a.tolist()
    return a*u0*u1
func(torch.tensor([1,2]))
```
a proposed [fix](https://github.com/pytorch/pytorch/pull/163807) but need to wait for CI and add unit test. 
cc @chauhang @penguinwu @ezyang @bobrenjc93

PR #163807, opened 2025-09-24
graph break on tolist if capture_scalar_outputs is false

laithsakka:
* __->__ #163807

address https://github.com/pytorch/pytorch/issues/163798

its problematic to not graph break because:
1. break current contract.  
2. well dynamo trace then we have .item call then if we ever re-trace later in autograd for example we hit a
 failure (We do not know where to graph break at that point)! see the added unit test. 
 

### DIFF
--- a/benchmarks/dynamo/timm_models.py
+++ b/benchmarks/dynamo/timm_models.py
@@ -451,9 +451,12 @@ class TimmRunner(BenchmarkRunner):
 
 
 def timm_main():
-    logging.basicConfig(level=logging.WARNING)
-    warnings.filterwarnings("ignore")
-    main(TimmRunner())
+    # Turn on capture_scalar_outputs for timm models by default.
+    # see https://github.com/pytorch/pytorch/pull/163807
+    with torch._dynamo.config.patch("capture_scalar_outputs", True):
+        logging.basicConfig(level=logging.WARNING)
+        warnings.filterwarnings("ignore")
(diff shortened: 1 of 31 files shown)


test/dynamo/test_misc.py
```diff
@@ -8146,6 +8146,18 @@ def my_dyn_fn(x, y):
         torch._dynamo.reset()
         torch.compile(my_dyn_fn, backend="eager")(y, y)
 
+    @torch._dynamo.config.patch(capture_scalar_outputs=False)
+    def test_tolist(self):
+        # This should compile with no faluire.
```
laithsakka, line 8152: used to fail to compile before this PR

bobrenjc93 (approved):

benchmarks/dynamo/timm_models.py
```diff
@@ -451,9 +451,12 @@ def forward_and_backward_pass(self, mod, inputs, collect_outputs=True):
 
 
 def timm_main():
-    logging.basicConfig(level=logging.WARNING)
-    warnings.filterwarnings("ignore")
-    main(TimmRunner())
+    # Turn on capture_scalar_outputs for timm models by default.
```
laithsakka: @BoyuanFeng @eellison  is any of you the right person to review this?

laithsakka:
Screenshot 2025-09-27 at 8 54 43 PM
error is on main alos

PR #163820, opened 2025-09-25
[dtensor] avoid shape recompilations on DTensorSpec

pianpwk:
skips DTensorSpec.sizes/strides in metadata guard checks


azahed98 (commented):
Changes lgtm. Let me know when final review is needed.

azahed98 (approved):
Lgtm! Thanks for the change

torch/distributed/tensor/_api.py
```diff
@@ -650,6 +650,15 @@ def __get_tensor_shard__(self, index):
         else:
             raise RuntimeError("Unsupported tensor type!")
 
+    @classmethod
+    def __metadata_guard__(cls, orig, other):
```
azahed98: Nit: return type annotation

azahed98:
> When I trace a DTensor with dynamic shapes, does the DTensorSpec size/stride become symbolic? 
I just double checked, and the DTensorSpec does contain symints during tracing.

> And if they do become symbolic, then why would we end up guarding on them?
I'm admittedly not that familiar with how the guards are constructed, but I was going off #152963, under "additional work" point 1, allegedly the symints are undesirable for constructing the guards and cause recompiles. @bdhirsh Could you provide more context?

ezyang:
It is a smell to me that `__tensor_flatten__` on DTensor returns stuff that shouldn't actually be compared against. Maybe it all works out but I would worry there are other places where we assume we can test against the metadata directly that cause problems.

PR #163822, opened 2025-09-25
Handle DDE in infer_size_impl

laithsakka:
* #163868
* __->__ #163822

hit this while running VLLM with unbacked for model Qwen/Qwen2-1.5B-Instruct

### DIFF
--- a/aten/src/ATen/InferSize.h
+++ b/aten/src/ATen/InferSize.h
@@ -45,7 +45,39 @@ inline void infer_size_impl(
     }
   }
 
-  auto set_infer_dim = [&]() {
+  if (infer_dim) {
+    // numel is the product of known sizes, it has to be divisible by newsize.
+    // newsize can't be zero.
+    if constexpr (std::is_same_v<NumelType, c10::SymInt>) {
+      if (auto v = newsize.maybe_as_int()) {
+        if (*v == 0) {
+          // Avoid div by 0 when sym_eq(numel % newsize, 0) is constructed!
+          // which may happen when newsize is not a symbol! if its a symbol
+          // division wont happen anyway during compile.
+          TORCH_MAYBE_SYM_CHECK(
+              numel == newsize,
+              "shape '",
+              shape,
+              "' is invalid for input of size ",
+              numel);
+        }
+      } else {
+        auto cond = sym_gt(newsize, 0)
+                        .sym_and(sym_eq(numel % newsize, 0))
+                        .sym_or(sym_eq(numel, newsize));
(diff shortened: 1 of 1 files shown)


aten/src/ATen/InferSize.h
```diff
@@ -64,7 +64,23 @@ inline void infer_size_impl(
     return;
   };
 
-  if (infer_dim && newsize > 0 && numel % newsize == 0) {
+  if (infer_dim) {
+    // numel is the product of known sizes it has to be divisible by newsize.
+    if constexpr (std::is_same_v<NumelType, c10::SymInt>) {
+      TORCH_MAYBE_SYM_CHECK(
+          sym_gt(numel, newsize).sym_and(sym_eq(numel % newsize, 0)),
```
laithsakka: oh no this shall be sym_gt(numel, 0).

aten/src/ATen/InferSize.h
```diff
@@ -64,7 +64,23 @@ inline void infer_size_impl(
...
+          shape,
+          "' is invalid for input");
+    } else {
+      TORCH_CHECK(
+          numel > 0 && numel % newsize == 0,
+          "shape '",
+          shape,
+          "' is invalid for input of size ",
+          numel);
+    }
```
ColinPeppler: noob question, can we just use TORCH_MAYBE_SYM_CHECK for both cases? or there's a performance hit so we need `if constexpr`?

laithsakka: does not compile. 
sym_gt(numel, 0) return type would be bool for non symint path so we cant call sym_and on it.

laithsakka: sym_gt(numel, 0) return type would be bool for non symint path so we cant call sym_and on it.

laithsakka:
ah need to handle the case for ==0 will take a look tomorrow

laithsakka:
ok need to fix issues before re-asking for reviews.

laithsakka:
ok all tests pass, will fix lint later.

aten/src/ATen/InferSize.h
laithsakka: mm let me try to intorduce && in Symbool and see if use only operators might work that way

laithsakka: did not work operator> specialize when second arg is int for SymInt,

laithsakka:
> test?

done

bobrenjc93 (approved):

Skylion007 (approved):

PR #163827, opened 2025-09-25
[dde] is_channels_last_strides_2d_s4 -> is_channels_last_strides_2d_s4_false to support unbacked shapes

ColinPeppler:
Adds is_channels_last_strides_2d_s4_or_false which will return False if it can't statically determine is_channels_last_strides_2d_s4.

## DDE traceback
```
  File "/caffe2/test/inductor/__unbacked_symints__/unbacked_symints#link-tree/torch/_dynamo/utils.py", line 3547, in run_node
    return node.target(*args, **kwargs)  # type: ignore[operator]
...
  File "<invalid>", line 0, in at::_ops::pixel_shuffle::call(at::Tensor const&, long)
...
  File "<invalid>", line 0, in at::TensorBase::suggest_memory_format(bool) const
  File "<invalid>", line 0, in c10::TensorImpl::is_strides_like_default(c10::MemoryFormat) const
  File "<invalid>", line 0, in c10::SymbolicShapeMeta::init_is_channels_last() const
  File "<invalid>", line 0, in bool c10::is_channels_last_strides_2d_s4<c10::SymInt>(c10::ArrayRef<c10::SymInt>, c10::ArrayRef<c10::SymInt>)
...
 File "torch/fx/experimental/symbolic_shapes.py", line 7593, in _evaluate_expr
    raise self._make_data_dependent_error(

torch._dynamo.exc.UserError: Could not guard on data-dependent expression u0 + 1 > 1

from user code:
   File "test/inductor/test_unbacked_symints.py", line 618, in fn
    return torch.nn.functional.pixel_shuffle(repeated, upscale_factor=2)
```

Differential Revision: D83094252

### DIFF
--- a/c10/core/MemoryFormat.h
+++ b/c10/core/MemoryFormat.h
@@ -1,5 +1,6 @@
 #pragma once
-
+#include <c10/core/SymBool.h>
+#include <c10/core/SymInt.h>
 #include <c10/util/ArrayRef.h>
 #include <c10/util/Exception.h>
 
@@ -126,20 +127,20 @@ inline std::vector<int64_t> get_channels_last_strides_3d(IntArrayRef sizes) {
 // 3. All helper functions have similar comments, only 1st helper function is
 // commented here.
 template <typename T>
-inline bool is_channels_last_strides_2d_s4(
+inline bool is_channels_last_strides_2d_s4_or_false(
     const ArrayRef<T> sizes,
     const ArrayRef<T> strides) {
   T min = 0;
   // special case for trivial C dimension. default to NCHW
-  if (strides[1] == 0) {
+  if (TORCH_GUARD_OR_TRUE(sym_eq(strides[1], 0))) {
     return false;
   }
   // loop strides indices
   for (auto& d : {1, 3, 2, 0}) {
-    if (sizes[d] == 0) {
+    if (TORCH_GUARD_OR_TRUE(sym_eq(sizes[d], 0))) {
       return false;
     }
-    if (strides[d] < min) {
+    if (TORCH_GUARD_OR_TRUE(sym_lt(strides[d], min))) {
       return false;
     }
     // Fallback to NCHW as default layout for ambiguous cases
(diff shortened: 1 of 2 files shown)


c10/core/MemoryFormat.h
```diff
@@ -249,7 +250,7 @@ inline bool is_channels_last_strides_2d(
     const ArrayRef<T> strides) {
   switch (sizes.size()) {
     case 4:
-      return is_channels_last_strides_2d_s4(sizes, strides);
+      return is_channels_last_strides_2d_s4_or_false(sizes, strides);
```
laithsakka, line 253: is_channels_last_strides_2d also should be renamed to is_channels_last_strides_2d_or_false.

c10/core/MemoryFormat.h
```diff
@@ -126,20 +127,20 @@ inline std::vector<int64_t> get_channels_last_strides_3d(IntArrayRef sizes) {
 // 3. All helper functions have similar comments, only 1st helper function is
 // commented here.
 template <typename T>
-inline bool is_channels_last_strides_2d_s4(
+inline bool is_channels_last_strides_2d_s4_or_false(
```
laithsakka, line 130: how hard would it be to do a 
sym_is_channels_last_strides_2d_s4
sym_is_channels_last_strides_2d

mm actually I am I am ok with this for now because
```
SymBool SymbolicShapeMeta::compute_channels_last_2d_dim5() const {
  init_is_channels_last_3d_contiguous();
  if (guard_or_false(is_channels_last_3d_contiguous(), __FILE__, __LINE__)) {
    return false;
  }
  return ~is_channels_last_3d_contiguous() &
      compute_strides_like_channels_last_2d();
}
```

already have guard_or_false here so it wont make an effect ,

but ideally this should be cleaned up to follow contiguous model, were only users can guard_or_false we do not implicitly  do it in APIs that do not have that naming.

can you add a comment and create gh issue to track this.

c10/core/MemoryFormat.h
```diff
@@ -160,7 +161,7 @@ inline bool is_channels_last_strides_2d_s4(
     //     [1, C, 1, H]@[HC, H, H, 1] transpose(1, 3)
     //     [1, H, 1, C]@[HC, 1, H, H] shouldn't be identified as channels_last
     min = strides[d];
-    if (sizes[d] > 1) {
+    if (TORCH_GUARD_OR_TRUE(sym_gt(sizes[d], 1))) {
```
laithsakka, line 164: can you add the comment that explains why this is ok
I think the reasoning if we do this when we dont the only effect is that min will be bigger than needed and this would break out the loop with false at line 153

PR #163835, opened 2025-09-25
Expose torch.nn.utils.parametrize

hanchchch:
`torch.nn.utils.parametrize` is not imported from `torch/nn/utils/__init__.py`, thus is not exposed and make it hard for code editors to statically analyze the code and provide auto-completion based on the function signature.

Screenshot 2025-09-25 at 12 01 52 PM

after the fix:

Screenshot 2025-09-25 at 12 02 16 PM


albanD (approved):
Good catch, thanks!

PR #163839, opened 2025-09-25
[fr] Skip the dtype check for some one to all or all to one collective

fduwjj:
* __->__ #163839

As title, in practice we found that sometimes, the dtype of gather does not match when it comes to output among all ranks, which is a undefined behavior. Same with broadcast and scatter. And they are all completed, so we should not think they are errors, we can skip it.


VieEeEw (approved):
#thanks for dtype mismatch fix!

PR #163842, opened 2025-09-25
Improve repeat op to a single copy

haifeng-jin:
In #163455 , the `reshape` was not a pure view op.

The `permute` before it created an non-contiguous tensor, which would trigger a data copy during the reshape.

This PR improved the implementation by remove the `urtensor` intermediate tensor completely.
By simply expanding the `xtensor` would achieve the `repeat` effect.

Before this PR, there were two data copies (in `urtensor.copy_` and `urtensor.reshape`).
Now, there is only one data copy in the `.copy_()`.
Reshape would not copy data because it is on a contiguous tensor.

One more note is that we do want at one copy because we want to duplicate the elements for the repeats.
User can inplace modify single elements without afffecting others.

### DIFF
--- a/aten/src/ATen/native/TensorShape.cpp
+++ b/aten/src/ATen/native/TensorShape.cpp
@@ -1880,43 +1880,30 @@ Tensor repeat(const Tensor& self, IntArrayRef repeats) {
 
   Tensor xtensor = self.expand(padded_size);
 
-  Tensor urtensor;
-  if (self.is_quantized()) {
-    urtensor = at::empty_quantized(target_size, self);
-  } else {
-    urtensor = at::empty(target_size, self.options());
-  }
-
-  // return an empty tensor if one of the repeat dimensions is zero
   if (zero_tensor) {
-    return urtensor;
+    return self.is_quantized() ? at::empty_quantized(target_size, self)
+                               : at::empty(target_size, self.options());
   }
 
+  Tensor view = xtensor;
+  // Create view of shape [1, s0, 1, s1, ...]
+  // where si is self.size(i).
   for (const auto i : c10::irange(xtensor.dim())) {
-    // can't unfold with step 0, so make sure step is at least 1
-    // (it doesn't matter what it is in that case, because the size is 0).
-    auto size_i = xtensor.sizes()[i];
-    urtensor = urtensor.unfold(i, size_i, std::max<int64_t>(size_i, 1));
+    view = view.unsqueeze(2 * i);
   }
 
-  urtensor.copy_(xtensor.expand_as(urtensor));
+  // Create view of shape [r0, s0, r1, s1, ...]
+  // where ri is repeat[i].
+  auto expand_shape = std::vector<int64_t>();
+  expand_shape.reserve(xtensor.dim() * 2);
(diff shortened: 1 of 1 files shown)


aten/src/ATen/native/TensorShape.cpp
```diff
@@ -1880,43 +1880,29 @@ Tensor repeat(const Tensor& self, IntArrayRef repeats) {
...
-  return urtensor;
+    view = view.unsqueeze(2 * i);
+    expand_shape.push_back(repeats[i]);
+    expand_shape.push_back(xtensor.size(i));
+  }
+  // expanded_view is non-contiguous because .expand set stride to 0.
+  auto expanded_view = view.expand(expand_shape);
+
+  // copy to contiguous tensor and reshape to [s0 * r0, s1 * r1, ...].
+  auto contiguous_copy = at::empty_like(expanded_view);
```
Skylion007: We don't need an extra arg here to override the memory_format?

haifeng-jin: As I tested, we do not need the `memory_format` arg though we should according to the doc.

For better readability, I changed it to explicitly use `at::MemoryFormat::Contiguous`.

Skylion007 (approved):

aten/src/ATen/native/TensorShape.cpp
```diff
@@ -1880,43 +1880,34 @@ Tensor repeat(const Tensor& self, IntArrayRef repeats) {
...
+  // copy to contiguous tensor.
+  auto contiguous_copy = at::empty(
+      expanded_view.sizes(),
+      expanded_view.options(),
+      at::MemoryFormat::Contiguous);
+  contiguous_copy.copy_(expanded_view);
+
+  // Reshape to [s0 * r0, s1 * r1, ...].
+  // No extra copy of data during reshape for a contiguous tensor.
+  return contiguous_copy.reshape(target_size);
```
Skylion007: Why don't we explicitly assert this with a view then? View will fail if the underlying tensor is not contiguous
```suggestion
  return contiguous_copy.view(target_size);

File: aten/src/ATen/native/TensorShape.cpp [hunk 3]
```diff
@@ -1880,43 +1880,34 @@ Tensor repeat(const Tensor& self, IntArrayRef repeats) {
...
+    return self.is_quantized() ? at::empty_quantized(target_size, self)
+                               : at::empty(target_size, self.options());
   }
 
+  // Create view of shape [r0, s0, r1, s1, ...]
+  // where ri is repeat[i], si is self.size(i).
+  Tensor view = xtensor;
+  auto expand_shape = std::vector<int64_t>();
+  expand_shape.reserve(xtensor.dim() * 2);
   for (const auto i : c10::irange(xtensor.dim())) {
```
Skylion007, line 1893:
```suggestion

  auto xdim = xtensor.dim();
  auto expand_shape = std::vector<int64_t>();
  expand_shape.reserve(xdim* 2);
  Tensor view = std::move(xtensor);
  for (const auto i : c10::irange(xdim)) {

haifeng-jin:
Also, cc @Skylion007 to see if there any insights.

PR #160266, opened 2025-08-10
Support of dtensor redistribute with device order

zpcore:
**RFC** [ordered device sharding RFC](https://docs.google.com/document/d/16ZSW8BHh2XgtryJEoSkptb-58R3PPrGGSB6lMxm1lVc/edit?usp=sharing)

### Summary
Now redistribute Dtensor honors the device ordering. We can specify `shard_order` as follow:
```python
sharded_dt = distribute_tensor(
                    tensor,
                    mesh,
                    # placements=[R, Shard(0), Shard(1), Shard(0)],   <<< specify `placements` 
                    #  or `shard_order` or both, but make sure they are not conflicted. If we 
                    #  specify `placements` only, it will automatically follow the default left-to-right
                    #  device order to place shards
                    shard_order=[0: ["tp","dp"], 1: ["sp"]]
                )
```
and
```python
out_dt = sharded_dt.redistribute(mesh, placement, shard_order)
```
If no order information specified, it will use the default left to right device order [0,1,2,...]. 
The detailed algorithm to redistribute dtensor from one spec to another with ordering information is documented in [Redistribute DTensor](https://docs.google.com/document/d/16ZSW8BHh2XgtryJEoSkptb-58R3PPrGGSB6lMxm1lVc/edit?tab=t.0#bookmark=id.6sq4dzpuf176). 

### Notes

- The original dtensor redistribute algorithm w/o device order is under `DTensorRedistributePlanner. generate_greedy_transform_infos`. The new algorithm is `DTensorRedistributePlanner.generate_optimal_transform_infos`, which is only enabled when user specify the device order that is not canonical.

- Note that device order information is added into the DTensorSpec. So `redistribute_local_tensor` doesn't need the `src_shard_order` and `dst_shard_order`. I leave them here as a reference for AutoParallel (cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @fmassa ). I will remove those order related args from redistributed related API in the next PR of the ghstack.

Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #163772
* #162294
* #160903
* __->__ #160266

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta

Differential Revision: [D83704483](https://our.internmc.facebook.com/intern/diff/D83704483)

PR #150490, opened 2025-04-01
[DTensor] StridedShard support uneven sharding

wconstab:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #150887
* #150862
* #150650
* __->__ #150490

This enables using FSDP+TP on parameters with dimensions that aren't
evenly divisible by the DP/TP mesh sizes.

- this may not support all possible combinations of strided shardings
  and shardings, but the support before this PR is not complete anyway

This contains several fixes for different aspects of DTensor behavior
relating to uneven strided sharding:
- original creation of the strided tensor requires fixes in
  StridedShard._split_tensor
- full_tensor() reconstruction requries fixes in
  StridedShard._to_replicate_tensor to correctly reshuffle the data into
  the original pre-sharded order
- Distributed Checkpointing support requires correct computation of the
  compute_local_shape_and_global_offset util so it knows how a local
  shard maps to the global tensor, for reconstruction during
  load/reshard.

This PR also adds a util `_explicit_order_placements` which converts a list of
placements with StridedSharding into a list of placements with only
regular sharding, with the order shuffled such that it is equivalent.

Builds on and completes the work started in https://github.com/pytorch/pytorch/pull/148894

Uneven Sharding Example
-------
(copied from _StridedShard._to_replicate_tensor docstring)

mesh = (DP=2, TP=2)
original = torch.arange(5)

**Applying Sharding**

Step 1 - Apply TP sharding
`tp = distribute_tensor(x, world_mesh['tp'], [Shard(0)])`

local_tensors:
rank0: [0,1,2]    rank1: [3,4]
rank1: [0,1,2]    rank3: [3,4]

Step 2 - Apply FSDP sharding
`dp_tp = ...` (the process of creating a strided-shard tensor is skipped over as it is hacky and complicated)
dp_tp has placement (_StridedShard(0, split_factor=2), Shard(0))
local_tensors:
rank0: [0,1]  rank1: [3]
rank1: [2]    rank3: [4]

**Reconstructing the Full Tensor**
Now, say someone wants to reconstruct dp_tp's full tensor. This will invoke 'redistribute' to replicate.
redistribute will first replicate the "Shard(0)" placement on the rightmost mesh dim, then replicate the
StridedShard placement second, which is implemented by this function.
So our starting point (`local_tensor` arg) is the result of replicating the Shard(0) placement across the
TP dim, which looks like this.

[...]

PR #163843, opened 2025-09-25
[DTensor] fix uneven _StridedShard

tianyu-l:
* __->__ #163843

Previous uneven `_StridedShard` in https://github.com/pytorch/pytorch/pull/150490 seems failing cases like sharding `tensor = torch.arange(6)` with FSDP 2, TP 2.

This PR attempts to reinvent `_StridedShard`.

I didn't test nested `_StridedShard`, because there shouldn't be any use cases. I think it will become quite messy when it comes to **nested uneven** `_StridedShard`. We are probably going to deprecate it anyway after @zpcore 's work https://github.com/pytorch/pytorch/pull/160266 on ordered sharding, so IMO not worth it to make it too general.

### DIFF
--- a/torch/distributed/tensor/placement_types.py
+++ b/torch/distributed/tensor/placement_types.py
@@ -111,7 +111,7 @@ class Shard(Placement):
         return shard_list, pad_sizes
 
     @staticmethod
-    def _local_shard_size_and_offset(
+    def local_shard_size_and_offset(
         curr_local_size: int,
         num_chunks: int,
         rank: int,
@@ -144,6 +144,14 @@ class Shard(Placement):
                 - shard_starting_idx
             )
             return local_shard_size, shard_starting_idx
+
+    def _local_shard_size_and_offset(
+        self,
+        curr_local_size: int,
+        num_chunks: int,
+        rank: int,
+    ) -> tuple[int, Optional[int]]:
+        return Shard.local_shard_size_and_offset(curr_local_size, num_chunks, rank)
 
     def _shard_tensor(
         self,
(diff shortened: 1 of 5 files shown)


torch/distributed/tensor/placement_types.py
```diff
@@ -460,33 +468,36 @@ def _split_tensor(
...
+        # and then second_split as the virtual "left" sharding, and finally assemble
+        # results in the transposed left-first order.
+        first_split, _ = super()._split_tensor(
+            tensor, self.split_factor, with_padding=False, contiguous=False
         )
-
-        # compute the chunk size inline with ``torch.chunk`` to calculate padding
-        full_chunk_size = (tensor.size(self.dim) + total_split - 1) // total_split
+        second_split = [
+            super(_StridedShard, self)._split_tensor(
```
ezyang, line 479: Nit: super() not super(_StridedShard, self)

torch/distributed/tensor/placement_types.py
```diff
@@ -460,33 +468,36 @@ def _split_tensor(
...
+                s, num_chunks=num_chunks, with_padding=False, contiguous=False
+            )[0]
```
ezyang, line 481: Note to self: This is list[torch.Tensor], it's not extracting the first tensor

ezyang (approved):

torch/distributed/tensor/placement_types.py
tianyu-l: IIRC I tried but didn't work.

I tried again with some agents
me:
> why the first call to super()._split_tensor is ok, but the second call to super()._split_tensor fails with "TypeError: super(type, obj): obj must be an instance or subtype of type"?

agent:
> The first call to super()._split_tensor is valid because it is called directly in the method of the _StridedShard class, so super() correctly refers to the parent class (Shard) and uses the current instance (self). However, in the second call, super()._split_tensor is used inside a list comprehension, which creates a new scope. In this new scope, super() without arguments does not have access to the correct class and instance context, leading to the error "TypeError: super(type, obj): obj must be an instance or subtype of type". To fix this, you should use the two-argument form: super(_StridedShard, self)._split_tensor(...), which explicitly specifies the class and instance.

zpcore: Weirdly, this works fine when I change second `super(_StridedShard, self)` into `super()._split_tensor`. Maybe some recent updates fixed it.

PR #163850, opened 2025-09-25
[OpenReg][BE] Replacing explicit prefix/suffix with CMake variables

can-gaa-hou:
As the title states, suffixes like`.dylib` and `lib` can be replaced by `CMAKE_SHARED_LIBRARY_SUFFIX`, and prefixes like `lib` can be replaced by `CMAKE_SHARED_LIBRARY_PREFIX` on Unix or `CMAKE_IMPORT_LIBRARY_PREFIX` on Windows.

cc @albanD


albanD (approved):
Ho very cool, I didn't knew these existed!
For a follow up, any way the prefix could be shared as well so we can remove the if statement all together?

can-gaa-hou:
> Ho very cool, I didn't knew these existed! For a follow up, any way the prefix could be shared as well so we can remove the if statement all together?

As far as I know, Windows does have SHARED LIBRARY, but it refers to ".dll", which is only used at runtime. For compiling, we use ".lib" files, which is named STATIC LIBRARY in Cmake.

PR #163865, opened 2025-09-25
AOTI MPS Shim Implementation

manuelcandales:
## MPS Shim API

*   Updated MPS shimification API with handles and function declarations:
    *   `AOTIMetalShaderLibraryHandle` and `AOTIMetalKernelFunctionHandle` types
    *   Library management: `aoti_torch_mps_create_shader_library`, `aoti_torch_mps_delete_shader_library`, `aoti_torch_mps_get_kernel_function`
    *   Kernel execution: `aoti_torch_mps_run_command_block`, `aoti_torch_mps_start_encoding`, `aoti_torch_mps_dispatch` variants, etc

## MPS Shader Codegen

*   Modified to generate source constants instead of direct `DynamicMetalShaderLibrary` instantiation:
    *   **Before**: `at::native::mps::DynamicMetalShaderLibrary mps_lib_0(R"MTL(...)MTL");`
    *   **After**: `const char* mps_lib_0_source = R"MTL(...)MTL";`
*   Updated kernel call generation  to use shimified functions:
    *   Generates calls to shimified API instead of direct libtorch calls

### Section 1: Shader Library
**Before (Direct Library Object)**
```cpp
at::native::mps::DynamicMetalShaderLibrary mps_lib_0(R"MTL(
    ...
)MTL");
```
**After (Source String)**
```cpp
const char* mps_lib_0_source = (R"MTL(
    ...
)MTL");
```

### Section 2: Getter Functions & RAII Management

**Before (Direct Library Access)**
```cpp
const std::shared_ptr<at::native::mps::MetalKernelFunction> get_mps_lib_0() {
    static const auto func = mps_lib_0.getKernelFunction("generated_kernel");
    return func;
}

AOTIMetalKernelFunctionHandle get_mps_lib_0_handle() {
    static const auto handle = AOTIMetalKernelFunctionHandle(get_mps_lib_0().get());
    return handle;
}
```

**After (Shim API + RAII Wrapper)**
```cpp
AOTIMetalKernelFunctionHandle get_mps_lib_0_handle() {
    static auto kernel_handle = []() {
        AOTIMetalShaderLibraryHandle lib_handle = nullptr;
        AOTIMetalKernelFunctionHandle kern_handle = nullptr;

        aoti_torch_mps_create_shader_library(mps_lib_0_source, &lib_handle);
        aoti_torch_mps_get_kernel_function(lib_handle, "generated_kernel", &kern_handle);

        // RAII wrapper with custom deleter
        auto lib_deleter = [](AOTIMetalShaderLibraryHandle h) {{
            if (h) aoti_torch_mps_delete_shader_library(h);
        }};

        using LibDeleter = decltype(lib_deleter);
        using LibPtr = std::unique_ptr<AOTIMetalShaderLibraryOpaque, LibDeleter>;

        // Return pair of kernel handle and library smart pointer for cleanup
        return std::make_pair(kern_handle, LibPtr(lib_handle, lib_deleter));
    }();
    return kernel_handle.first;
}
```

### Section 3: Runtime Execution

**Before (Direct Library Methods)**
```cpp
void AOTInductorModel::run_impl(...) {

    ...

    get_mps_lib_0()->runCommandBlock([&] {
        get_mps_lib_0()->startEncoding();
        aoti_torch_mps_set_arg_tensor(get_mps_lib_0_handle(), 0, buf0);
        aoti_torch_mps_set_arg_tensor(get_mps_lib_0_handle(), 1, arg0_1);
        aoti_torch_mps_set_arg_tensor(get_mps_lib_0_handle(), 2, arg1_1);
        get_mps_lib_0()->dispatch({static_cast<uint64_t>(10LL)});

    });

    ...

} // AOTInductorModel::run_impl
```

**After (Shim API with Lambda Pattern)**
```cpp
void AOTInductorModel::run_impl(...) {

    ...

    auto mps_lib_0_lambda_0 = [&](AOTIMetalKernelFunctionHandle handle) {
        aoti_torch_mps_start_encoding(handle);
        aoti_torch_mps_set_arg_tensor(handle, 0, buf0);
        aoti_torch_mps_set_arg_tensor(handle, 1, arg0_1);
        aoti_torch_mps_set_arg_tensor(handle, 2, arg1_1);
        aoti_torch_mps_dispatch_single(handle, static_cast<uint64_t>(10LL));
    };

    std::function<void(AOTIMetalKernelFunctionHandle)> mps_lib_0_func_wrapper_0 = mps_lib_0_lambda_0;
    aoti_torch_mps_run_command_block(get_mps_lib_0_handle(), aoti_torch_mps_shared_callback, &mps_lib_0_func_wrapper_0);

    ...

} // AOTInductorModel::run_impl
```


torch/_inductor/codegen/cpp_wrapper_mps.py
```diff
@@ -113,14 +186,34 @@ def _generate_kernel_call_helper(
             self.write_mps_kernel_call(kernel_name, new_args)
 
     def write_mps_kernel_call(self, name: str, call_args: list[str]) -> None:
-        # Initialization of the kernel function and kernel function handle
-        # variables have already been done at the beginning, which was
-        # codegen-ed in `codegen_mps_func_init`
-        self.writeline(f"get_{name}()->runCommandBlock([&] {{")
-        self.writeline(f"    get_{name}()->startEncoding();")
+        # Generate unique variable names to avoid duplicate declarations
+        # when the same MPS lib is used multiple times
```
JacobSzwejbka, line 190: Why cant you just call the same one?

torch/csrc/inductor/aoti_torch/shim_mps.cpp
```diff
@@ -27,3 +32,121 @@ AOTITorchError aoti_torch_mps_set_arg_int(
...
+AOTITorchError aoti_torch_mps_create_shader_library(
+    const char* metal_shader_source,
+    AOTIMetalShaderLibraryHandle* library_handle) {
+  AOTI_TORCH_CONVERT_EXCEPTION_TO_ERROR_CODE({
+    auto* library = new at::native::mps::DynamicMetalShaderLibrary(std::string(metal_shader_source));
+    *library_handle = reinterpret_cast<AOTIMetalShaderLibraryHandle>(library);
+  });
+}
+
+AOTITorchError aoti_torch_mps_delete_shader_library(
```
angelayi, line 41: when does this get called in the generated code?

torch/csrc/inductor/aoti_torch/shim_mps.cpp
```diff
@@ -27,3 +32,121 @@ AOTITorchError aoti_torch_mps_set_arg_int(
...
+    size_t group_size_size) {
+  AOTI_TORCH_CONVERT_EXCEPTION_TO_ERROR_CODE({
+    auto* function_ptr = reinterpret_cast<at::native::mps::MetalKernelFunction*>(func);
+    c10::ArrayRef<uint64_t> length_ref(length, length_size);
+    c10::ArrayRef<uint64_t> group_size_ref(group_size, group_size_size);
+    function_ptr->dispatch(length_ref, group_size_ref);
+  });
+}
+
+AOTITorchError aoti_torch_mps_synchronize_stream() {
```
angelayi: when does this get called in the generated code?

torch/_inductor/codegen/cpp_wrapper_mps.py
angelayi: I think it's bc the arguments are different, so you need a new lambda each time?

manuelcandales: Yeah, same MPS lib with different arguments.

torch/csrc/inductor/aoti_torch/shim_mps.cpp
manuelcandales: your are right, is not needed - deleted!

torch/csrc/inductor/aoti_torch/shim_mps.cpp
manuelcandales: updated codegen to generate a function to clean up all the shader libraries, and this function gets called in the AOTInductorModelContainer destructor

angelayi (approved):
thanks so much!

torch/_inductor/codegen/cpp_wrapper_mps.py
```diff
@@ -47,13 +48,16 @@ def _generate_kernel_call_helper(
...
-            aoti_torch_mps_set_arg(get_mps_lib_0_handle(), 1, arg0_1);
-            ...
-            get_mps_lib_0()->dispatch(9);
-        });
+        auto mps_lib_0_lambda = [&](AOTIMetalKernelFunctionHandle handle) {
+            aoti_torch_mps_start_encoding(handle);
+            aoti_torch_mps_set_arg_tensor(handle, 0, buf0);
+            aoti_torch_mps_set_arg_tensor(handle, 1, arg0_1);
+            aoti_torch_mps_set_arg_tensor(handle, 2, arg1_1);
+            aoti_torch_mps_dispatch_single(handle, static_cast<uint64_t>(10LL));
```
desertfire, line 56: Maybe this is a MPS specific question, but where does this constant 10 come from?

desertfire (commented):
It will be helpful if you can paste a before/after generated code example from a simple model.

torch/_inductor/codegen/cpp_wrapper_mps.py
```diff
@@ -133,48 +226,72 @@ def get_device_include_path(device: str) -> str:
...
-                )
-                self.prefix.writeline(
-                    f"    static const auto handle = AOTIMetalKernelFunctionHandle(get_{line.kernel_name}().get());"
-                )
-                self.prefix.writeline("    return handle;")
-                self.prefix.writeline("}")
+        # Generate shimified declarations and functions for each library
+        for lib_name in shader_libraries:
+            # Add handle declarations
+            self.prefix.writeline(
```
desertfire: You can use `self.prefix.splice` to write multiple lines. The current way is hard to read.

torch/csrc/inductor/aoti_torch/shim_mps.cpp
```diff
@@ -1,9 +1,14 @@
 #include <ATen/native/mps/MetalShaderLibrary.h>
 #include <torch/csrc/inductor/aoti_torch/c/shim_mps.h>
 #include <torch/csrc/inductor/aoti_torch/utils.h>
+#include <ATen/mps/MPSStream.h>
+#include <unordered_map>
 
 using namespace torch::aot_inductor;
 
+// Global storage to keep shared_ptr alive while raw pointers are used
+static std::unordered_map<at::native::mps::MetalKernelFunction*, std::shared_ptr<at::native::mps::MetalKernelFunction>> function_storage;
```
desertfire: Can you explain more on why this is needed?

desertfire (changes requested):

torch/_inductor/codegen/cpp_wrapper_mps.py
angelayi: The 10 comes from [this code](https://github.com/pytorch/pytorch/blob/1ab124e188b6968ba41a2b4f3595c14e8cd3d96a/torch/_inductor/codegen/mps.py#L976-L1006) which sets the thread size and group size

torch/_inductor/codegen/cpp_wrapper_mps.py
manuelcandales: done!

torch/_inductor/codegen/cpp_wrapper_mps.py
```diff
@@ -132,49 +225,77 @@ def get_device_include_path(device: str) -> str:
...
+                aoti_torch_mps_create_shader_library(mps_lib_0_source, &lib_handle);
+                aoti_torch_mps_get_kernel_function(lib_handle, "generated_kernel", &kern_handle);
+
+                // RAII wrapper with custom deleter
+                auto lib_deleter = [](AOTIMetalShaderLibraryHandle h) {
+                    if (h) aoti_torch_mps_delete_shader_library(h);
+                };
+
+                using LibDeleter = decltype(lib_deleter);
+                using LibPtr = std::unique_ptr<AOTIMetalShaderLibraryOpaque, LibDeleter>;
```
desertfire, line 246: This part is the same for all the functions and should be extracted as a common code.

torch/_inductor/codegen/cpp_wrapper_mps.py
```diff
@@ -132,49 +225,77 @@ def get_device_include_path(device: str) -> str:
...
+
+                // Return pair of kernel handle and library smart pointer for cleanup
+                return std::make_pair(kern_handle, LibPtr(lib_handle, lib_deleter));
+            }();
+            return kernel_handle.first;
```
desertfire, line 251: Since you only return `kernel_handle.first`, `kernel_handle.second` dies after `kernel_handle`'s lifetime ends here, which means `lib_deleter` is triggered to delete `lib_handle`. Is this expected?

torch/csrc/inductor/aoti_torch/shim_mps.cpp
manuelcandales: I added more comments to the code explaining why this is needed.
MetalShaderLibrary::getKernelFunction returns a shared_ptr to a MetalKernelFunction. But aoti_torch_mps_get_kernel_function needs to return a raw pointer for C compatibility. I do this to keep the shared_ptr alive.
I implemented RAII for the library in the codegen
aoti_torch_mps_delete_shader_library does all the clean up (library and functions belonging to the library)

torch/_inductor/codegen/cpp_wrapper_mps.py
manuelcandales: yes, we currently codegen one library per generated kernel

torch/_inductor/codegen/cpp_wrapper_mps.py
manuelcandales: done!

torch/_inductor/codegen/cpp_wrapper_mps.py
desertfire: What I meant to ask is, is kern_handle depend on lib_handle to be live? I guess the answer is no, but wanted to confirm.

desertfire: Related to my later question about function_storage, can you return std::shared_ptr or std::unique_ptr here? That you won't need to maintain a global table.

manuelcandales: Those are C++ features, right? The shim signature needs to be pure C. That has been the main limitation. At least, when I implemented aoti_torch_mps_run_command_block using std::function, sure it worked with libtorch, but didn't work with libtorch-free AOTI. That's why I have to do the trampoline with the callback (see aoti_torch_mps_run_command_block in shim_mps.h)

torch/_inductor/codegen/cpp_wrapper_mps.py
manuelcandales: I tried to extract that common code, but I am running into redefinition errors. It seems that for certain cases, the cpp_wrapper_mps.py codegen is being called twice. I tried a few things, but I still run into the error.
I am considering to revert that to how I had it (repeated in each getter) and we can figure out how to make it work in a future PR.

manuelcandales:
> It will be helpful if you can paste a before/after generated code example from a simple model.

done! (added to PR description)

torch/csrc/inductor/aoti_torch/shim_mps.cpp
manuelcandales: Alright, the last commit implements a new getCachedKernelFunctionPtr method in MetalShaderLibrary, avoiding the global storage.

aten/src/ATen/native/mps/OperationUtils.mm
```diff
@@ -917,6 +917,22 @@ void executeMPSAllocatorCallback(void* ptr, EventType event) override {}
   return std::make_shared<MetalKernelFunction>(cpl, func);
 }
 
+MetalKernelFunction* MetalShaderLibrary::getCachedKernelFunctionPtr(const std::string& name) {
+  // Check if kernel is already cached
+  auto it = kernelCache.find(name);
```
desertfire, line 922: Do we have to use another cache here?
1. getLibraryPipelineState already uses a cache;
2. The way I suggested was to return `new MetalKernelFunction(cpl, func)` directly, and then manage that MetalKernelFunction object lifetime in the generated code in a RAII way. You can look at how AtenRecordFunctionHandle and RAIIAtenRecordFunctionHandle work as a reference. As a matter of fact, I suspect you don't need to create this function, you only need to create a new object in `aoti_torch_mps_get_kernel_function`.

manuelcandales: Yes, I tried that, but I ran into a segmentation fault.
It seems that MTLFunction objects are references into the MTLLibrary, they cannot live independently. So, the library needs to be the one managing their lifetime

desertfire: I am not sure about the implementation details about MTLFunction and MTLLibrary, but I am ok if you want to use another cache here. One thing I will recommend is to rename `getCachedKernelFunctionPtr` to `borrowCachedKernelFunctionPtr`, so that the caller knows they don't own the returned pointer and thus not manage its lifetime.

desertfire (approved):

Issue #163462, opened 2025-09-22
GroupedMM triton template IMAs on B200


drisspg:
# Summary

Repro step
1. Change this line back: https://github.com/pytorch/pytorch/pull/163460/files#diff-417ae05acc8966294ad5ac5a23bd2406b1252166e99bad30fd735f49e5582c47L494
to <= sm90
2.  Use zsh func:
```
shell

function sanitize() {
    export CUDA_LAUNCH_BLOCKING=1
    export PYTORCH_NO_CUDA_MEMORY_CACHING=1
    compute-sanitizer --tool memcheck "$@"
}
```

3. run `sanitize python test/test_matmul_cuda.py -k test_grouped_gemm_compiled_op --repeat 2  `
```Shell
========= 
========= Invalid __global__ read of size 2 bytes
=========     at triton_tem_fused__grouped_mm_0+0x870 in cdqdlvbzvwav4iitwbiivm2tl4tekwwd55fxpogamtmpniuqpoxg.py:90
=========     by thread (172,0,0) in block (2,0,0)
=========     Access at 0x7fbfb52012c0 is out of bounds
=========     and is 189 bytes after the nearest allocation at 0x7fbfb5201200 of size 4 bytes
=========     Saved host backtrace up to driver entry point at kernel launch time
=========         Host Frame: (anonymous namespace)::launch_kernel(_object*, _object*) [0x979680] in libtorch_python.so
=========         Host Frame: cfunction_call in methodobject.c:548 [0x21bc28] in python
=========         Host Frame: _PyObject_MakeTpCall in call.c:240 [0x1ff17b] in python
=========         Host Frame: _PyEval_EvalFrameDefault.cold in bytecodes.c:2715 [0x112b22] in python
=========         Host Frame: _PyObject_FastCallDictTstate in call.c:133 [0x201b6d] in python
=========         Host Frame: _PyObject_Call_Prepend in call.c:508 [0x22a638] in python
=========         Host Frame: slot_tp_call in typeobject.c:8791 [0x2f70ca] in python
=========         Host Frame: _PyObject_MakeTpCall in call.c:240 [0x1ff17b] in python
=========         Host Frame: _PyEval_EvalFrameDefault.cold in bytecodes.c:2715 [0x112b22] in python
=========         Host Frame: _PyObject_FastCallDictTstate in call.c:133 [0x201b6d] in python
=========         Host Frame: _PyObject_Call_Prepend in call.c:508 [0x22a638] in python
[...]

PR #163895, opened 2025-09-25
Fix non-TMA loads in grouped MM Triton kernel

alexsamardzic:
* #164256
* __->__ #163895


alexsamardzic:
Should fix IMA reported in #163462.  Will look further into the remaining numerical failures on B200.

alexsamardzic:
The IMAs were coming from `tl.load`-s not properly masked along the inner dimension.  The `compute-sanitizer` would report IMAs even when the test run only once.

The non-deterministic results were coming from masked input values not all properly set to 0.  On the first run, typically the block where values would be `tl.load`-ed would be set to 0s, to the result would come up correct, but on the second run it seems that oftentimes it happens oftentimes the block would not be initially all 0s, so these bogus values would affect the results.

torch/_inductor/kernel/mm_grouped.py
```diff
@@ -335,11 +335,12 @@ def early_config_prune(g, m, configs, named_args):
...
                         + offs_k[None, :] * B_STRIDE_K
                     )
-                    a = tl.load(a_ptrs, mask=offs_am[:, None] < m_size)
-                    b = tl.load(b_ptrs, mask=offs_bn[:, None] < n_size)
-                    if k_offset + BLOCK_K > k_size:
-                        a = tl.where(group_offs_k < k_size, a, 0)
-                        b = tl.where(group_offs_k < k_size, b, 0)
+                    a_mask = (offs_am[:, None] < m_size) & (group_offs_k[None, :] < k_size)
+                    b_mask = (offs_bn[:, None] < n_size) & (group_offs_k[None, :] < k_size)
+                    a = tl.load(a_ptrs, mask=a_mask)
```
ngimel: can you do `other=0` here instead of tl.where?

ngimel:
btw why doesn't B200 use TMA loads?

alexsamardzic:
> btw why doesn't B200 use TMA loads?

It uses them in the same case as when H100 is using it: basically when both inputs are row-major.  This is because of TMA descriptor limitation, enforced [here](https://github.com/triton-lang/triton/blob/22b1a44d659b099313c7920fe70da9045a70f7d8/python/triton/tools/tensor_descriptor.py#L27) (note that familiar 16-byte alignment related limitations are also enforced in the same method).

So at the moment, either both inputs are TMA-loaded, or are loaded through `tl.load()`.  I should probably consider doing a TMA load also when only one input is row-major (so TMA load for that input, and `tl.load()` for the other one), but that will somewhat complicate the kernel code.

torch/_inductor/kernel/mm_grouped.py
alexsamardzic: I think `tl.where` is Triton idiom for doing this kind of stuff, it's not possible to write something like `a[!a_mask] = 0`.

alexsamardzic:
As far as numerical differences concerned:  It seems the problem is actually with CUTLASS-based kernel, not the Triton-based one.

Namely, I've simply added ([patch.txt](https://github.com/user-attachments/files/22564232/patch.txt)) saving inputs and both `_grouped_mm` (CUTLASS-based kernel) reference output and Triton-based kernel (well, when `max_autotune` is on for the test) output, for both runs of the test.  It seems that the wrong values are actually produced by the CUTLASS-based kernel, it's same problem with zero output not being properly generated when group size is 0, that we once faced on H100, and believed it is fixed by CUTLASS.  Here is an example of inputs and outputs for the two test runs:

[2d-2d-True-True-True.txt](https://github.com/user-attachments/files/22564296/2d-2d-True-True-True.txt)
[2d-2d-True-True-True-2.txt](https://github.com/user-attachments/files/22564301/2d-2d-True-True-True-2.txt)

So the test was for 2D/2D case (the differences now occur only in this case), both inputs were row-major, and `max_autotune` was on.  The first file lists inputs/outputs for the first test run, and the second file for the second run.  The `m` and `n` are 3 and 7, respectively, and the k-offsets are `[0,  8, 16, 16, 27]`.  So the output will be 5x3x7 and the first and fourth 3x7 sub-tensors should be all 0s, as corresponding groups are of size 0 along `k`.  The Triton-based kernel produces such groups in both runs, however for some reason the CUTLASS-based kernel do not on the second run.  Tried with PyTorch built against the latest CUTLASS, unfortunately still the same error.

alexsamardzic:
> Ok, can you switch tests to run with `fill_uninitialized_memory=True` then, that should uncover bugs like this without waiting for hardware update
> 

I don't think that would help here, namely the issue here was because of how GPU shared memory is filled.  On the other hand, the input tensors in the test are actually already filled with random values, including parts of these tensors that are for there for alignment only, exactly for the purpose of detecting related issues.


> Also, B200 (and H100 for bf16) should in principle support tma for all transpose combinations, tma has this limitation but then tl.dot should accept transposed tiles, no? E.g. for cutlass kernel perf of all bf16 variants is pretty close, and they all use the same kernel just with the different layout tags. But that's beyond the scope of this fix, for this fix let's just enable tests with uninitialized memory.

Makes sense indeed - do you want me to do this next?

torch/_inductor/kernel/mm_grouped.py
ngimel: but tl.load itself supports setting uninitialized values via `other` kwarg

alexsamardzic:
> Ugh, does this happen on H100 and B200, or only B200? Can you isolate and report to cutlass?

I believe I found the cause of the problem: it seems not to be about the shared memory, but about the accumulator initialization instead.  Was not able to simplify the reproducer that we have here in the meantime, but reported the issue (with a dummy fix) to the CUTLASS team: NVIDIA/cutlass#2674; hope it gets properly fixed quickly.

torch/_inductor/kernel/mm_grouped.py
lezcano: even more, `tl.where` would block pipelining. `tl.load` lowers to `cp.async` when the `other` value is zero. `tl.where` would make loading `a`, and `b` into registers.

lezcano (approved):

PR #163896, opened 2025-09-25
Fix preserve annotation with decomp

yushangdi:
If we use `fx_traceback.preserve_node_meta()`, we will have a few extra node.meta fields on nodes, such as "seq_nr", added from `fx/proxy.py`. As a result, there might be non-empty node.meta on graph nodes.


SherlockNoMad (approved):
lgtm, plz double check with @ydwu4

ydwu4 (approved):

PR #163903, opened 2025-09-25
[cuDNN][SDPA] Disable dropout for cuDNN SDPA on 9.11 - 9.13

eqy:
cuDNN introduced some broken heuristics for these cases so we need to disable dropout to avoid unexpected crashes due to heuristics refusing to proceed

### DIFF
--- a/aten/src/ATen/native/transformers/cuda/sdp_utils.cpp
+++ b/aten/src/ATen/native/transformers/cuda/sdp_utils.cpp
@@ -689,6 +689,14 @@ bool can_use_cudnn_attention(const sdp_params& params, bool debug) {
   }
   return false;
 #endif
+#if defined(CUDNN_VERSION) && CUDNN_VERSION > 91000 && CUDNN_VERSION < 91400
+  if (params.dropout > 0.0) {
+    if (debug) {
+      TORCH_WARN(CUDNN_VERSION, " cuDNN version does not support droppout in SDPA (9.11 - 9.13).");
+    }
+    return false;
+  }
+#endif
   // Define gate functions that determine if a flash kernel can be ran
   // Replace with std::to_array when we migrate to c++20
   constexpr auto general_constraints =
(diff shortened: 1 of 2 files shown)


aten/src/ATen/native/transformers/cuda/sdp_utils.cpp
```diff
@@ -688,6 +688,14 @@ bool can_use_cudnn_attention(const sdp_params& params, bool debug) {
     TORCH_WARN(CUDNN_VERSION, " cuDNN version too old to use cuDNN Attention (< v9.0.0)");
   }
   return false;
+#endif
+#if defined(CUDNN_VERSION) && CUDNN_VERSION > 91000 && CUDNN_VERSION < 91400
+  if (params.dropout > 0.0) {
+    if (debug) {
+      TORCH_WARN(CUDNN_VERSION, " cuDNN version does not support droppout in SDPA (9.11 - 9.13).");
```
ngimel, line 696: make it WARN_ONCE?

ngimel (approved):

aten/src/ATen/native/transformers/cuda/sdp_utils.cpp
ngimel: Also should it be runtime version check? If someone compiles against 9.10 and then updates just cudnn to 9.11 they might run into this?

eqy: We already don't really allow that unless `PYTORCH_SKIP_CUDNN_COMPATIBILITY_CHECK` is set, so we assume the user knows what they are doing

eqy: Will add this anyway as it looks simple enough

malfet (approved):

atalman (approved):

aten/src/ATen/native/transformers/cuda/sdp_utils.cpp
eqy: It will be `WARN_ONCE` in the sense that it is only printed for debugging (when no backends are available), this just follow the convention for the rest of the "why wasn't this backend chosen" messages here which Flash and Mem-Eff attention also use.

PR #162056, opened 2025-09-03
Expand on autotune support for persistent reduction kernels

jataylo:
After the removal of want_no_x_dim for persistent reduction kernels, we can improve the autotuning setup for persistent reduction kernels.

Currently even with tuning enable, filtering will only try a single config in many cases. Avoid filtering with autotune mode, and override MAX_BLOCK limit. Also we always include tiny_config when autotuning is enabled.

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @mlazos

PR #163908, opened 2025-09-25
[ROCm][inductor] autotune support for persistent reduction kernels

naromero77amd:
After the removal of want_no_x_dim for persistent reduction kernels, we can improve the autotuning setup for persistent reduction kernels.

Currently even with tuning enable, filtering will only try a single config in many cases. Avoid filtering with autotune mode, and override MAX_BLOCK limit. Also we always include tiny_config when autotuning is enabled.

Contributions from several members of the AMD Inductor and Triton teams: @jataylo @iupaikov-amd @AmdSampsa @xiaohuguo2023 

cc @mlazos @jeffdaily @sunway513 @jithunnair-amd @pruthvistony @ROCmSupport @dllehr-amd @jataylo @hongxiayang @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben


naromero77amd:
This is a redo of this PR: https://github.com/pytorch/pytorch/pull/162056

But branch was created in ROCm repo.

naromero77amd:
We demonstrate that the updated heuristics are generally applicable to other models such as the HuggingFace (HF) Inductor Dashboard benchmark suite.

Here is a high-level outline of the steps that were taken. We ran the HF benchmark suite from Inductor Dashboard using the command below:
```
python benchmarks/dynamo/runner.py --dtypes amp --suites huggingface --training --compilers inductor_no_cudagraphs --no-gh-comment --output-dir $outdir
```
Then we extract the relevant kernels and benchmark with and without the PR (baseline also used TORCHINDCTOR_MAX_AUTOTUNE_POINTWISE=1). There is a geomean speed-up of 1.08 X on over 400 kernels.

[kernel_comp_per_pr.csv](https://github.com/user-attachments/files/22938447/kernel_comp_per_pr.csv)

jansel (approved):

torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -2933,40 +2942,49 @@ def _persistent_reduction_configs(
...
-                    register_intensive=True,
-                )
-            ]
+        elif reduction_hint == ReductionHint.OUTER:
+            configs = configs[-1:]
+        elif reduction_hint == ReductionHint.OUTER_TINY:
+            configs = tiny_configs
+    else:
+        # If autotune is enabled append tiny configs
+        for conf in tiny_configs:
```
PaulZhang12: Why are we doing this here? We are adding extra autotuning to nvidia side, maybe do this just for amd?

naromero77amd: We can make that change. Will be pushing that out soon.

PaulZhang12 (approved):

PR #163910, opened 2025-09-26
[dtensor] set dropout probability to zero for testing purposes

dolpm:
Summary:
i noticed that nn.functional.multi_head_attention_forward was getting some intermittent failures. looks like it's related to dropout, which is inherently probabilistic. we can remove randomness by setting the probability to zero.


dolpm:
> Should set seed instead.
> 
> torch.manual_seed

tried that first, wasn't working. can look more into it tomorrow.

PR #163911, opened 2025-09-26
Speed up return_and_correct_aliasing by caching write_aliases for outputs

swolchok:
* __->__ #163911

We did some validation-type checks every time that only need to be
done once because they depend only on the schema.

Benchmark: https://gist.github.com/swolchok/92ef7dd1017b4fb4f17bbbf5ab59a510

Ran on M4 Mac. Output cleaned up to reduce visual noise and only show the return_and_correct_aliasing case.

```
Before:
> python torchdispatchbench.py
__torch_dispatch__ with return_and_correct_aliasing:
_ = t_aliasing.detach()
  Median: 3.69 us
  IQR:    0.03 us (3.68 to 3.71)
  4 measurements, 10000 runs per measurement, 1 thread

After:
> python torchdispatchbench.py
__torch_dispatch__ with return_and_correct_aliasing:
_ = t_aliasing.detach()
  Median: 3.33 us
  IQR:    0.02 us (3.33 to 3.34)
  4 measurements, 10000 runs per measurement, 1 thread
```

Observed similar results on Linux server, with before median of 19.92 usec and after of 18.78 usec.

### DIFF
--- a/torch/utils/_python_dispatch.py
+++ b/torch/utils/_python_dispatch.py
@@ -780,11 +780,11 @@ def return_and_correct_aliasing(func, args, kwargs, out):
 
     # Next: we need to make sure to return inputs directly, if the output is a mutable alias (e.g. add_()).
 
-    # simple case: none of our outputs have mutable aliases, so we can return the output as-is
     schema_info_outs_write_aliases = schema_info.outs_write_aliases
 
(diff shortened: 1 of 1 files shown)


swolchok:
hey @bdhirsh can you please take a look at this one?

PR #163922, opened 2025-09-26
[inductor] add subsystem to pattern matcher

avikchaudhuri:
Summary:
Running a toy example through `torch.compile(fullgraph=True, backend="inductor")` with default inductor config, I tried to see what passes are run in each of pre-grad, joint-graph, and post-grad phases by printing out the subsystem in `GraphTransformObserver`. However the subsystem showed up as None in a bunch of transforms that were run in each of those phases, so this PR adds some additional annotations.

Note that these annotations are probably not a complete set, since other transforms may run based on changes to the config that are not covered here.

Hopefully this doesn't change behavior. However, I did notice that bisecting relies on disabling various phases, which means that while before some passes would *not* be disabled (because their subsystem was `None`), now they would.

Test Plan: existing tests + manual test described in summary

Differential Revision: D83306676

### DIFF
--- a/torch/_inductor/fx_passes/pre_grad.py
+++ b/torch/_inductor/fx_passes/pre_grad.py
@@ -1,5 +1,6 @@
 # mypy: allow-untyped-defs
 import copy
+import functools
 import itertools
 import logging
 import types
@@ -14,7 +15,9 @@ from torch.fx.experimental.optimization import (
     matches_module_pattern,
     replace_node_module,
 )
-from torch.fx.passes.graph_transform_observer import GraphTransformObserver
+from torch.fx.passes.graph_transform_observer import (
+    GraphTransformObserver as GraphTransformObserverBase,
+)
 from torch.fx.passes.shape_prop import ShapeProp
 from torch.nn import functional as F
 from torch.nn.utils.fusion import fuse_conv_bn_eval, fuse_conv_bn_weights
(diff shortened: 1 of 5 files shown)


jansel (changes requested):
Test failures?

torch/_inductor/fx_passes/pre_grad.py
```diff
@@ -1,5 +1,6 @@
 # mypy: allow-untyped-defs
 import copy
+import functools
```
eellison, line 3: pre_grad_passes doesnt exist yet in bisector - mind adding it ?

eellison (commented):
nice ! but lets wait for test failures to be fixed..

torch/_inductor/fx_passes/pre_grad.py
avikchaudhuri: So basically add a line here?
https://github.com/pytorch/pytorch/blob/d7491fb1c1262597a77cbc4aa21d0e2f9fc7620d/torch/_inductor/compiler_bisector.py#L60

avikchaudhuri: Hmm, I'll do this in a separate PR.

qq: 
Looks like the bisector doesn't pick up something like:
```
with GraphTransformObserver(gm, "pre_grad_custom_pass"):
            config.pre_grad_custom_pass(gm.graph)
```

so instead I have to do:
```
GraphTransformObserver(gm, "pre_grad_custom_pass").apply_graph_pass(
            config.pre_grad_custom_pass
)
```

KP? @eellison

jansel (approved):

PR #163924, opened 2025-09-26
add support for ir scalar literal parsing for inf/-inf/True/False 

weinanliu:
Currently the ir parser doesn't support parse ir like 
```
graph():
  %12 : float = prim::Constant[value=-inf]()
  %13 : float = prim::Constant[value=inf]()
  %14 : bool = prim::Constant[value=True]()
  %15 : bool = prim::Constant[value=False]()
  return (%12)
```

So the python script below will throw error.

```
#!/bin/env python
import torch

def test():
    return [True, False]
f = torch.jit.script(test)
torch._C._jit_pass_constant_propagation(f.graph)
ts_str = f.graph.__repr__()
print(ts_str)
ts = torch.parse_ir(ts_str)
func = torch._C._create_function_from_graph("forward", ts)
ret = func()
assert ret == [True, False]


def test():
    return [float("inf"), float("-inf")]
f = torch.jit.script(test)
torch._C._jit_pass_constant_propagation(f.graph)
ts_str = f.graph.__repr__()
print(ts_str)
ts = torch.parse_ir(ts_str)
func = torch._C._create_function_from_graph("forward", ts)
ret = func()
assert ret == [float("inf"), float("-inf")]
```

I add "inf" and bool cases for IRParser::parseScalarLiteral in irparser.cpp.

cc @EikanWang @jgong5 @wenzhe-nrv @sanchitintel @ezyang

### DIFF
--- a/test/jit/test_misc.py
+++ b/test/jit/test_misc.py
@@ -433,6 +433,17 @@ class TestMisc(JitTestCase):
         self.assertTrue(ret.numel() == 1)
         self.assertTrue(len(ret.size()) == 1)
 
+    def test_parse_ir_single_inf(self):
+        ir = """
+        graph():
+          %12 : float = prim::Constant[value=inf]()
+          return (%12)
+        """
+        graph = torch._C.parse_ir(ir, True)
+        func = torch._C._create_function_from_graph("forward", graph)
+        ret = func()
+        self.assertTrue(ret == float("inf"))
+
     def test_parse_ir_single_minus_inf(self):
         ir = """
         graph():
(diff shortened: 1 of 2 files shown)


ezyang (approved):
suuuure why not

test/jit/test_misc.py
```diff
@@ -433,6 +433,54 @@ def test_parse_ir_single_element_tensor_negative(self):
         self.assertTrue(ret.numel() == 1)
         self.assertTrue(len(ret.size()) == 1)
 
+    @unittest.skipIf(os.environ.get("PYTORCH_TEST_WITH_DYNAMO") == "1", "Not supported with TorchDynamo")
```
ezyang: Use skipIfTorchDynamo instead

weinanliu: It's OK for new commit, thank you

ezyang (approved):

PR #163930, opened 2025-09-26
[DeviceMesh][ez] Extract the pg creation as a util function

fduwjj:
* #163367
* __->__ #163930
* #163928
* #163288
* #163212

This is just to extract common logic into a util function because we will use it many times for the following stack of Device Mesh refactoring.

### DIFF
--- a/torch/distributed/device_mesh.py
+++ b/torch/distributed/device_mesh.py
@@ -160,6 +160,13 @@ else:
                 )
                 if cur_rank in mesh_nd:
                     res_submesh = submesh
+            res_submesh = DeviceMesh._create_mesh_from_ranks(
+                device_mesh.device_type,
+                pg_ranks_by_dim,
+                cur_rank,
+                submesh_dim_names,
+                _init_backend=False,
+            )
 
             res_submesh._dim_group_names = slice_dim_group_name  # type: ignore[possibly-undefined, has-type]
             self.child_to_root_mapping[res_submesh] = device_mesh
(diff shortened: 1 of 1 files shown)


torch/distributed/device_mesh.py
```diff
@@ -856,6 +860,54 @@ def get_all_groups(self) -> list[ProcessGroup]:
...
+            device_type: str,
+            pg_ranks_by_dim: torch.Tensor,
+            cur_rank: int,
+            mesh_dim_names: tuple[str, ...],
+            backend_override: Optional[tuple[BackendConfig, ...]] = None,
+            _init_backend: bool = True,
+        ) -> "DeviceMesh":
+            """
+            Helper method to create a DeviceMesh from tensor `pg_ranks_by_dim`. This is due to
+            the constraint of ProcessGroup API that all ranks have to call the PG creation API.
```
fegin:
```suggestion
            the constraint of ProcessGroup API that all ranks have to call the PG creation API even if the rank is not in that PG.
```

fegin (approved):

torch/distributed/device_mesh.py
```diff
@@ -856,6 +860,55 @@ def get_all_groups(self) -> list[ProcessGroup]:
...
+            for mesh_nd in pg_ranks_by_dim:
+                mesh = DeviceMesh(
+                    device_type,
+                    mesh_nd,
+                    mesh_dim_names=mesh_dim_names,
+                    backend_override=backend_override,
+                    _init_backend=_init_backend,
+                )
+                if cur_rank in mesh_nd:
+                    res_mesh = mesh
```
lw, line 904: Thanks for splitting this out as I had asked! I didn't get to review this before it was landed, but there's still something I don't understand. I get that we need to call the "PG creation API" multiple times on each rank, even for the ranks that don't participate, but I don't get why we need to call the _DeviceMesh_ constructor multiple times!

Could we instead call the PG creation API directly and just invoke the DeviceMesh constructor once with the right PG?

PR #163936, opened 2025-09-26
[opaque obj] torch.compile support

angelayi:
* #167364
* #167363
* #167245
* #167231
* #167230
* __->__ #163936
* #163714
* #163284


torch/_dynamo/variables/builder.py
```diff
@@ -1446,27 +1447,32 @@ def build_key_value(i, k, v):
...
                 )
 
-            # This exists to allow a smoother transition.
-            # The implications are:
-            # The script objects won't be tracked as proxies.
-            # Methods on these objects won't show up in the graph.
-            # The original script object might be mutated.
-            if not hasattr(value, "__obj_flatten__"):
-                return self.wrap_user_defined(value)
+            if is_opaque_type(type(value)):
```
zou3519, line 1449: What happens if someone marks an int as an opaque type? or a torch.Tensor?

torch/_inductor/codegen/wrapper.py
```diff
@@ -2119,6 +2119,10 @@ def add_expr_input(name, val):
             output.writeline(f"{name} = {val}")
 
         def add_torchbind_input(name, value):
+            if value is None:
+                output.writeline(f"{name} = None")
+                return
```
zou3519, line 2127: what's this for?

torch/_inductor/lowering.py
```diff
@@ -2699,6 +2700,8 @@ def require_channels_last(_, *args, **kwargs):
 
 
 def constrain_to_fake_tensor(arg, fake_arg):
+    if isinstance(fake_arg, FakeScriptObject):
+        return arg
```
zou3519, line 2709: Why is this needed? What is `arg` when fake_arg is a FakeScriptObj?

test/test_opaque_obj_v2.py
```diff
@@ -292,6 +292,25 @@ def forward(self, arg0_1, arg1_1, arg2_1):
...
+                x = x + x
+                return x
+
+        mod = Model()
+        rng = RNGState(0)
+        x = torch.ones(2, 3)
+
+        res = torch.compile(mod, fullgraph=True)(rng, x)
+        self.assertFalse(torch.allclose(res, x * x + x))
+
```
zou3519, line 335: 1) can you do an assertexpectedinline on some graphs? preferably dynamo graph and aot graph?
2) we don't need to torch.compile a module. Can we do this test case without an nn.Module?
3) can you see what happens if we compile a function that captures the rng_state ? Dynamo should automatically lift the rng_state to being a graph input
4) what happens when we return the rng_state object? (i'd expect this to just work with no intervention from us)
5) calling a method on the rng_state object should induce a graph break, there should be a test here

zou3519: It might be good too to use an opaqueobject that is just a container around a tensor or an integer, so that we can check that compile didn't end up baking in some value too

torch/_inductor/codegen/wrapper.py
angelayi: This code is for making sample inputs at the end of the inductor generated code for benchmarking(?). However since we don't pass around the original value of the opaque obj, here it is none. 

Even if we do have the original value of the opaque obj, we don't require that opaque objs are pickleable (unlike ScriptObjects), so we would probably have to do a workaround here.

torch/_inductor/lowering.py
angelayi: `arg` is an `IRNode`, like `TorchBindObject(name='arg1_1', value=<torch._library.fake_class_registry.FakeScriptObject object at 0x7f799445fce0>)`

my understanding is that this function adds runtime assertions for checking input tensor strides?

test/test_opaque_obj_v2.py
angelayi: updated! 

regarding #4, the graph doesn't actually contain the opaque obj in the output because dynamo doesn't detect that the object is mutated.

for #5, I added a better graph break message, but for TorchScriptObjectVariables graph breaks turn into [hard errors](https://github.com/pytorch/pytorch/blob/bdb37536befc51d1f67563869c06fa78fa941ee4/torch/_dynamo/variables/script_object.py#L43).

torch/_dynamo/variables/builder.py
zou3519: @angelayi?

torch/_dynamo/variables/script_object.py
```diff
@@ -80,6 +81,16 @@ def as_proxy(self) -> Proxy:
         "Dynamo cannot safely trace script object due to graph break."
     )
     def var_getattr(self, tx: "InstructionTranslator", name: str) -> VariableTracker:
+        if getattr(self.value, "script_class_name", "") == OpaqueTypeStr:
+            unimplemented(
+                gb_type="Attempted to access attributes/methods on an OpaqueObject",
```
zou3519, line 86: is it possible to get the type of the underlying opaqueobject? or is this info too difficult to get? doesn't need to happen in this PR

zou3519 (approved):

torch/_dynamo/variables/script_object.py
angelayi: yes we should be able to, but it would the string representation of the type

torch/_dynamo/variables/builder.py
angelayi: Oops sorry, I think based on the implementation of builder.py it would still still get treated as an int/torch.Tensor, since those casings are applied before the opaque object casing..

torch/_dynamo/variables/script_object.py
zou3519: string representation of the type is better, at least it is more specific than "OpaqueObject"

PR #163943, opened 2025-09-26
[BE] Remove 32 bit system check

PaliC:
Removing the 32 bit system memory check as we longer support 32 bit systems.


PaliC:
* (to be filled)

albanD:
@malfet do we have 32-bit systems left?
This check was never triggered in any CI we have.

malfet (changes requested):
This feels weird: you are removing : you are removing check_sizes function for 32-bit systems, but are saying they are still used

If we really want to get rid of this check, just add `static_assert(sizeof(void *) == 8);` to this file

PR #163944, opened 2025-09-26
[Docs] Add standard-imghdr for PyTorch Doc

fffrog:
* __->__ #163944

As the title stated.

Python [Pep-0594](https://peps.python.org/pep-0594) have removed imghdr from python standard libaries, the older version of sphinx don`t add it as installation dependencies, so we need to add it to requirement as an temporary dependencies.


albanD (approved):
FYI @svekars

svekars (approved):

PR #163947, opened 2025-09-26
[1/N] Apply UP035 rule in tests

cyyever:
Apply UP035 `ruff` rule in tests, but some tests for `fx` and `dynamo` are excluded in case the old typing is the test target.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @ezyang @EikanWang @jgong5 @wenzhe-nrv @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela

### DIFF
--- a/test/distributed/_composable/fsdp/test_fully_shard_comm.py
+++ b/test/distributed/_composable/fsdp/test_fully_shard_comm.py
@@ -6,7 +6,8 @@ import itertools
 import os
 import tempfile
 import unittest
-from typing import Callable, Optional, Union
+from collections.abc import Callable
+from typing import Optional, Union
 from unittest.mock import MagicMock
 
 import torch
--- a/test/distributed/_composable/fsdp/test_fully_shard_overlap.py
+++ b/test/distributed/_composable/fsdp/test_fully_shard_overlap.py
@@ -3,7 +3,7 @@
 import copy
 import functools
 import unittest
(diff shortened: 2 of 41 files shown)


test/dynamo/cpython/3_13/typinganndata/ann_module.py
```diff
@@ -50,7 +49,7 @@ class S(metaclass = Meta):
 
 def foo(x: int = 10):
     def bar(y: List[str]):
-        x: str = 'yes'
+        pass
```
albanD: Isn't the intent of the test to check this?

cyyever: yes, this is an incident change that must be reverted..

Skylion007:
Some of these tests are explicitly to check that tracing through typing import Tuple and tuple typing work, if I recall it's some of the files that already import / use both.

ezyang (approved):

cyyever:
@Skylion007 I have the same concern, let me revert the changes of fx and dynamo files.

PR #154170, opened 2025-05-23
[cuBLASLt][cuBLAS] Support 2D bias and `beta != 1.0` in cuBLASLt

eqy:
Fixes https://github.com/pytorch/pytorch/issues/153590

cc @ptrblck @msaroufim @jerryzh168

PR #163955, opened 2025-09-26
[CUDA][cuBLAS] addmm -- some refactoring for easier navigation between the Lt and non-Lt paths

nikitaved:
As per title. Additionally, some Lt selection conditions are revisited, and some redundancy removed (especially in the ROCm vs non-ROCm paths).

### DIFF
--- a/aten/src/ATen/native/cuda/Blas.cpp
+++ b/aten/src/ATen/native/cuda/Blas.cpp
@@ -272,28 +272,110 @@ cuda::blas::GEMMAndBiasActivationEpilogue activation_to_gemm_and_blas_arg(Activa
   }
 }
 
-static bool getDisableAddmmCudaLt() {
-    static const auto env_value = c10::utils::get_env("DISABLE_ADDMM_CUDA_LT");
-    if (env_value == "1") {
-      return true;
-    }
-    return false;
+/*
+ * Checks whether DISABLE_ADDMM_CUDA_LT is set.
+ * Additionally, for ROCM we test whether the architecture supports the Lt.
+ */
+static bool isGloballyDisabledAddmmCudaLt(const at::Device& device) {
+  // When hipBLASLt is not supported on the architecture, return true
+  #ifdef USE_ROCM
+  static const std::vector<std::string> archs = {
+        "gfx90a", "gfx942",
+    #if ROCM_VERSION >= 60300
+        "gfx1100", "gfx1101", "gfx1200", "gfx1201", "gfx908",
+    #endif
+    #if ROCM_VERSION >= 70000
+        "gfx950", "gfx1150", "gfx1151"
+    #endif
+  };
+  const auto is_hipblas_lt_arch_supported = at::detail::getCUDAHooks().isGPUArch(archs, device.index());
+  if (!is_hipblas_lt_arch_supported) {
+    return true;
+  }
+  #endif
+
+  // Check whether it is disabled in the env
+  static const auto is_addmm_cuda_lt_disabled = c10::utils::get_env("DISABLE_ADDMM_CUDA_LT");
+  if (is_addmm_cuda_lt_disabled == "1") {
+    return true;
+  }
+
+  return false;
 }
 
-#ifdef USE_ROCM
-static bool isSupportedHipLtROCmArch(int index) {
-    static const std::vector<std::string> archs = {
-        "gfx90a", "gfx942",
-#if ROCM_VERSION >= 60300
-        "gfx1100", "gfx1101", "gfx1200", "gfx1201", "gfx908",
-#endif
-#if ROCM_VERSION >= 70000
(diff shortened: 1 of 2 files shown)


nikitaved:
ROCM is fixed.

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -343,105 +484,49 @@ Tensor& addmm_out_cuda_impl(Tensor& result, const Tensor& self, const Tensor& ma
     "expected mat1 and mat2 to have the same dtype, but got: ", mat1.dtype(), " != ", mat2.dtype()
   )
 
+  if (result.is_same(self)) {
+    TORCH_CHECK(result.dim() == 2, "tensors must be 2-D");
+    TORCH_CHECK(self.sizes()[0] == mat1.sizes()[0], "self dim 0 must match mat1 dim 0");
+    TORCH_CHECK(self.sizes()[1] == mat2.sizes()[1], "self dim 1 must match mat2 dim 1");
+  }
```
nikitaved, line 497: This was directly taken from the previous impl. Maybe we can benefit from changing these error messages.

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -270,28 +270,106 @@ cuda::blas::GEMMAndBiasActivationEpilogue activation_to_gemm_and_blas_arg(Activa
...
+  if (result.is_same(self)) {
+    return false;
+  }
+
+  #if defined(USE_ROCM) && ROCM_VERSION == 60400
+  // hipblaslt TT fp32 regression on ROCm 6.4, cannot use
+  const auto args = cublasCommonArgs(mat1, mat2, result);
+  if (args.transa == 't' && args.transb == 't') {
+    return false;
+  }
```
nikitaved, line 323: Taken from the previous impl. It is potentially a pretty heavy check (with memory allocation and copy), so needs a revisit for perf reasons.

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -270,28 +270,105 @@ cuda::blas::GEMMAndBiasActivationEpilogue activation_to_gemm_and_blas_arg(Activa
...
+
+  const auto mat1_sizes = mat1.sizes();
+  const auto mat2_sizes = mat2.sizes();
+  #if defined(CUDA_VERSION) || defined(USE_ROCM)
+  const auto scalar_type = mat1.scalar_type();
+  return (beta.toComplexDouble() == 1.0
+    // self.dim() == 1 && result.dim() == 2 && self.sizes()[0] == mat2_sizes[1]
+    // is to use lt interface only when self is bias.
+    && self.dim() == 1 && self.sizes()[0] == mat2_sizes[1] && self.is_contiguous()
+    && result.dim() == 2 && result.is_contiguous()
```
eqy, line 334: There is some hope that we would eventually be able to resolve this https://github.com/pytorch/pytorch/pull/154170/

ngimel: What's preventing #154170 from being landed?

ngimel (approved):
So this is only for addmm, right? Do we have adequate testing for these code paths? I suspect H100 and especially B200 are not tested for this in our CI.

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -333,7 +416,70 @@ static void launchTunableGemmAndBias(cublasCommonArgs &args, const Scalar& alpha
   }
 }
 
+template <typename scalar_t, typename res_scalar_t = scalar_t>
+bool launchGemmAndBiasCudaLt(
```
ngimel:
```suggestion
bool launchGemmAndBiasCublasLt(
```

aten/src/ATen/native/cuda/Blas.cpp
eqy: complaints about internal Meta CI breaking but have not followed up on those in a while

nikitaved:
@ngimel , yes, these are just for addmm. I did test locally with H100 -- the grep over `addmm` in `test_linalg.py` is very comprehensive (different layouts, transposition flags, alpha/beta values), but I also tested with `test_matmul_cublas.py` and tests for linear for good measure. I will update regarding B200 once I get my hands on it.

aten/src/ATen/native/cuda/Blas.cpp
jeffdaily: we're already building rocm 6.4.2 and rocm 7 now, so this check is rarely exercised today, if that helps.

eqy (approved):

eqy:
scaled_mm failures should be unrelated and are being addressed in : https://github.com/pytorch/pytorch/pull/165747

PR #163959, opened 2025-09-26
Helper to augment graph with additional deps

eellison:
* #163960
* __->__ #163959
* #163754
* #163215

In comm-compute overlap we will have a graph with: 

```
def foo(...):
     ag = all_gather(...)
     hiding_compute = mm(...)
     wait(ag)
```

There is no explicit dependency between the hiding compute and the collectives, but we want to add implicit dependencies from wait->hiding_compute, and from hiding_compute->all_gather to preserve overlap. 

Additionally, while bucketing, we will merge collective starts and collective waits together. In this case, we will want to treat the two nodes as a single subgraph - each node in the merged set will have the union of all deps in the set. 

This pr adds `AugmentedGraphHelper` that adds the apis, and allows querying for dependency with this augmented graph.


test/inductor/test_augmented_graph_helper.py
```diff
@@ -0,0 +1,341 @@
...
+            ]:
+                self.nodes[node.name] = node
+
+        # Get all nodes and create tracker
+        self.all_nodes = list(self.graph.nodes)
+        self.tracker = AugmentedGraphHelper(self.graph)
+
+    def get_deps(self, node):
+        """Helper to get dependencies for a node."""
+        return list(node.args) if hasattr(node, "args") else []
```
v0i0:
```suggestion
        return list(getattr(node, "args", []))
```

torch/_inductor/augmented_graph_helper.py
```diff
@@ -0,0 +1,113 @@
...
+        """
+        Merge nodes into a single set set.
+        """
+        if len(nodes) <= 1:
+            return
+        n1 = nodes[0]
+        for n2 in nodes[1:]:
+            self.merge_to_set(n1, n2)
+
+    def unmerge_node(self, node: fx.Node) -> None:
```
v0i0, line 44: the fact that this function exists doesn't quite sit right with me. it feels like we'd want to be able to undo previous merges, not extract single nodes?

v0i0 (approved):

test/inductor/test_augmented_graph_helper.py
eellison: i'll blame claude here haha..

torch/_inductor/augmented_graph_helper.py
eellison: the use case is that to do a collective merge, we need to merge a node start to an existing set of starts, then, attempt to merge the node wait with existing waits. If the merging of waits is unsuccessful we undo the most recent adding of the start to existing starts.

I can assert that the new node is a singleton in the callee if that helps ? I agree that what you're suggesting might be more general but I don't want to complicate this with not yet real use cases yet.

v0i0: i see, had a look at the next pr in the stack. the rhs node in merge is always a singleton, so this is always fine there. also seems like `merge_nodes` is never called? maybe `merge_to_set` could assert `new_node` is singleton?

eellison: sounds good

IvanKobzarev (approved):

PR #163960, opened 2025-09-26
Add Comm-Compute Preserving Bucketer

eellison:
* __->__ #163960
* #163959
* #163754
* #163215

tl;dr performs bucketing while preserving comm-compute overlap. 

In comm-compute overlap we will have a graph with: 

```
def foo(...):
     ag = all_gather(...)
     hiding_compute = mm(...)
     wait(ag)
```

There is no explicit dependency between the hiding compute and the collectives, but we want to add implicit dependencies from wait->hiding_compute, and from hiding_compute->all_gather to preserve overlap. 

Additionally, while bucketing, we will merge collective starts and collective waits together. In this case, we will want to treat the two nodes as a single subgraph - each node in the merged set will have the union of all deps in the set. 

We perform bucketing while augmenting the graph with these relationships. This can be done separably from comm-compute overlap, so long as the hiding compute relationships are passed in.

TODO: 
- need to instrument fx graph so inductor respects these relationships.
- the compile time of the bucketing search can be sped up significantly by limiting what portion of the graph we traverse through
- more memory aware handling


torch/_inductor/fx_passes/overlap_scheduling.py
```diff
@@ -309,6 +314,9 @@ def run(self) -> torch.fx.GraphModule:
                 self._force_oldest_wait()
                 continue
 
+            # if torch.distributed.distributed_c10d.get_rank() == 0:
```
IvanKobzarev: nit clean

torch/_inductor/fx_passes/overlap_preserving_bucketer.py
```diff
@@ -0,0 +1,255 @@
...
+                continue
+
+            # Initialize bucket with first collective
+            bucket_info = CollBucket(
+                collectives=[start_node],
+                total_bytes=self.collective_info[start_node].size_bytes,
+            )
+            processed.add(start_node)
+
+            for candidate in collective_group:
```
ruisizhang123, line 114: quick question: why we need to iterate over a `collective_group` twice here?

torch/_inductor/fx_passes/overlap_preserving_bucketer.py
```diff
@@ -0,0 +1,255 @@
...
+        if is_all_gather(bucket[0]):
+            new_nodes, replacements = merge_all_gather_bucket(
+                self.graph,
+                bucket,
+                wait_insertion_point=first_wait,
+                insert_before=next_node,
+                mode="custom_ops",
+            )
+        else:
+            assert is_reduce_scatter(bucket[0])
```
ruisizhang123, line 227: wonder if we need to add additional check here to ensure all the collective comms in `bucket` have the same comm_func, dtype, and pg_name?

ruisizhang123 (approved):
LGTM, thank you!

torch/_inductor/fx_passes/overlap_preserving_bucketer.py
eellison: this should already be handled by `bucket_key`

torch/_inductor/fx_passes/overlap_preserving_bucketer.py
eellison: We're trying every collective with every other collective (within the same bucket key). i can add a TODO about limiting O(n^2). there are a few compile time todos left.

v0i0 (approved):

torch/_inductor/fx_passes/bucketing.py
```diff
@@ -570,7 +570,7 @@ def process_collective_bucket(
     trace_args_fn: Callable[[list[torch.fx.Node]], tuple[Any, ...]],
     insert_before: Optional[torch.fx.Node] = None,
     wait_insertion_point: Optional[torch.fx.Node] = None,
-) -> dict[torch.fx.Node, torch.fx.Node]:
+) -> tuple[list[torch.fx.Node], dict[torch.fx.Node, torch.fx.Node]]:
```
ezyang, line 573: severe tuple blindness lol

IvanKobzarev (approved):

PR #163965, opened 2025-09-26
[ROCm][tunableop] Modified Online Tuning Mode to add Instant Logging

sarthaktandon9amd:
- Added instant logging in online tuning and made it the default mode so that tuned GEMMs are instantly written.
- Allows us to have saved tuning configs, in cases of OOM errors or crashes.
- Removed the `write_file` and `write_file_on_exit` APIs


test/test_linalg.py
```diff
@@ -4778,26 +4836,32 @@ def test_matmul_small_brute_force_tunableop(self, device, dtype):
...
             unique_id = self.id().split(".")[-1]
-            filename2 = f"{filename1}_tmp1.csv"
             filename3 = f"{filename1}_tmp2.csv"
             ordinal = torch.cuda.current_device()
             assert filename1 == f"tunableop_results_{unique_id}_{ordinal}.csv"
             assert len(torch.cuda.tunable.get_results()) > 0
 
-            assert torch.cuda.tunable.write_file()  # use default filename
-            assert torch.cuda.tunable.write_file(filename2)  # use custom, one-time filename
+            assert os.path.exists(filename1)
```
naromero77amd: This needs to be written to something like:
``
self.assertTrue(os.path.exists(filename1) 
``

test/test_linalg.py
```diff
@@ -4746,11 +4746,69 @@ def test_matmul_small_brute_force_3d_Nd(self, device, dtype):
...
+    #         with open(filename3) as file3:
+    #             file3_contents = file3.read()
+    #         assert file1_contents == file2_contents
+    #         assert file1_contents == file3_contents
+
+    #         # We need to reset the filename to the default value so we can properly
+    #         # clean up intermediate files
+    #         self._set_tunableop_defaults()
+
     @onlyCUDA
```
naromero77amd, line 4749: This UT was accidentally commented out. Please fix. I think you want to delete it since the updated UTs is the next one.

test/test_linalg.py
```diff
@@ -4778,26 +4836,32 @@ def test_matmul_small_brute_force_tunableop(self, device, dtype):
...
-            filename2 = f"{filename1}_tmp1.csv"
             filename3 = f"{filename1}_tmp2.csv"
             ordinal = torch.cuda.current_device()
             assert filename1 == f"tunableop_results_{unique_id}_{ordinal}.csv"
             assert len(torch.cuda.tunable.get_results()) > 0
 
-            assert torch.cuda.tunable.write_file()  # use default filename
-            assert torch.cuda.tunable.write_file(filename2)  # use custom, one-time filename
+            assert os.path.exists(filename1)
+
```
naromero77amd: For the `test_matmul_small_brute_force_tunableop` UT, anything passed Line 4845 is no longer needed IMO, since we got rid of write_file APIs.

test/test_linalg.py
```diff
@@ -5633,8 +5731,17 @@ def test_tf32_offline_tunableop(self, device, dtype):
...
+                self.assertTrue(os.path.exists(results_filename))
+
+                # self.assertTrue(torch.cuda.tunable.write_file())
+                with open(results_filename) as f:
+                    content = f.read()
+                    self.assertIn("Validator", content)
+                    # Count non-validator lines to verify results were written
+                    result_lines = [l for l in content.split('\n')
+                                if l and not l.startswith('Validator')]
+                    self.assertGreater(len(result_lines), 0)
```
naromero77amd: Same comment as before, just delete.

test/test_linalg.py
```diff
@@ -5863,7 +5971,17 @@ def test_mm_submatrix_offline_tunableop(self, device, dtype):
...
+            results_filename = torch.cuda.tunable.get_filename()
+            self.assertTrue(os.path.exists(results_filename))
+
+            with open(results_filename) as f:
+                content = f.read()
+                self.assertIn("Validator", content)
+                # Count non-validator lines to verify results were written
+                result_lines = [l for l in content.split('\n')
+                            if l and not l.startswith('Validator')]
+                self.assertGreater(len(result_lines), 0)
```
naromero77amd, line 5948: Same comment as before, just delete.

test/test_linalg.py
```diff
@@ -5432,7 +5520,17 @@ def test_gemm_bias_offline_tunableop(self, device, dtype):
...
```
naromero77amd: Same comment as before, just delete.

test/test_linalg.py
```diff
@@ -5007,7 +5084,17 @@ def test_scaled_gemm_offline_tunableop(self, device, dtype):
...
```
naromero77amd: Same comment as before, just delete this code.

test/test_linalg.py
```diff
@@ -4917,7 +4982,18 @@ def is_bmm_compatible(A, B):
...
```
naromero77amd: You don't need this extra segment of code to check tunableop_results0.csv.

Line 4999:
ok = self._compare_untuned_tuned_entries()

will catch any error if the results file was written incorrectly.

But please keep this part:
``
results_filename = torch.cuda.tunable.get_filename()
self.assertTrue(os.path.exists(results_filename))
``

naromero77amd (changes requested):

test/test_linalg.py
```diff
@@ -5007,7 +4996,16 @@ def test_scaled_gemm_offline_tunableop(self, device, dtype):
...
```
naromero77amd, line 5008: Not needed, get rid of these lines.

test/test_linalg.py
```diff
@@ -5432,7 +5431,16 @@ def test_gemm_bias_offline_tunableop(self, device, dtype):
...
```
naromero77amd, line 5443: Not needed, get rid of these lines.

test/test_linalg.py
```diff
@@ -5863,11 +5880,102 @@ def test_mm_submatrix_offline_tunableop(self, device, dtype):
...
+                            if l and not l.startswith('Validator')]
+                final_count = len(final_lines)
+            
+            self.assertGreater(final_count, initial_count)
+
+    @onlyCUDA
+    @skipCUDAIfNotRocm
+    @dtypes(torch.float32)
+    def test_matmul_empty_existing_file_tunableop(self, device, dtype):
+        import os
```
naromero77amd: Add some comments to clarify the scenario that is being tested here.

test/test_linalg.py
```diff
@@ -5863,11 +5880,102 @@ def test_mm_submatrix_offline_tunableop(self, device, dtype):
...
+        with self._tunableop_ctx():
+            torch.cuda.tunable.set_rotating_buffer_size(0)
+            results_filename = torch.cuda.tunable.get_filename()
+            
+            # Pre-create an empty results file
+            with open(results_filename, 'w') as f:
+                pass  # Empty file
+            
+            torch.cuda.tunable.enable(True)
+            torch.cuda.tunable.tuning_enable(True)
```
naromero77amd: These lines should already be part of `with self._tunableop_ctx()`

test/test_linalg.py
```diff
@@ -5863,11 +5880,102 @@ def test_mm_submatrix_offline_tunableop(self, device, dtype):
...
+            
+            make_arg = partial(make_tensor, device=device, dtype=dtype)
+            
+            x = make_arg((64, 64), noncontiguous=False)
+            y = make_arg((64, 64), noncontiguous=False)
+            self.check_single_matmul(x, y)
```
naromero77amd: Just do:
``
C = torch.matmul(A, B)
``
Use a unique A and B that is different from those used in a prior UT.

test/test_linalg.py
```diff
@@ -5863,11 +5880,102 @@ def test_mm_submatrix_offline_tunableop(self, device, dtype):
...
             ok = self._compare_untuned_tuned_entries()
             self.assertTrue(ok)
+    
+    @onlyCUDA
+    @skipCUDAIfNotRocm
+    @dtypes(torch.float32)
+    def test_ops_append_to_existing_file_tunableop(self, device, dtype):
+        import os
+        
+        with self._tunableop_ctx():
```
naromero77amd, line 5876: Inside with `self._tunableop_ctx(x):`

Cat a small tunableop_results0.csv file:

```
<Validator>
<single matmul result>
```

For example,
```
content = """First couple of lines would be the Validator
The one result line contain a matmul result.
"""

# Write it to a file
with open(<tunableop.get_filename()>, "w") as f:
    f.write(content)
```

test/test_linalg.py
```diff
@@ -5863,11 +5880,102 @@ def test_mm_submatrix_offline_tunableop(self, device, dtype):
...
+            torch.cuda.tunable.set_rotating_buffer_size(0)
+            torch.cuda.tunable.tuning_enable(True)
```
naromero77amd: Can be omitted tuning_enable is part of self._tunableop_ctx().

test/test_linalg.py
```diff
@@ -5863,11 +5880,102 @@ def test_mm_submatrix_offline_tunableop(self, device, dtype):
...
+
```
naromero77amd: Just do a single:
C = torch.matmul(A, B)

test/test_linalg.py
```diff
@@ -5863,11 +5880,102 @@ def test_mm_submatrix_offline_tunableop(self, device, dtype):
...
 
             # Compare Param Signature of untuned and tuned results
             ok = self._compare_untuned_tuned_entries()
             self.assertTrue(ok)
+    
+    @onlyCUDA
+    @skipCUDAIfNotRocm
+    @dtypes(torch.float32)
+    def test_ops_append_to_existing_file_tunableop(self, device, dtype):
+        import os
```
naromero77amd: Add the following comment:
```
Test that if an existing TunableOp results file already exists from a previous run, the default behavior is to append new results to the existing file if the Validator aligns.
```

naromero77amd (changes requested):
A few minor changes to the UTs.

naromero77amd (approved):

naromero77amd:
Added `suppress-bc-linter` label because we are removing the `write_file` and `write_file_on_exit` APIs since they are no longer needed.

sarthaktandon9amd:
Verifying that the new unit tests ran in this shard linux-jammy-rocm-py3.10 / test (default, 2, 6, linux.rocm.gpu.2):
```
test/test_linalg.py::TestLinalgCUDA::test_ops_append_to_existing_file_tunableop_cuda_float32, test/test_linalg.py::TestLinalgCUDA::test_matmul_empty_existing_file_tunableop_cuda_float32,
```

aten/src/ATen/cuda/tunable/Tunable.cpp
```diff
@@ -150,6 +166,78 @@ void TuningResultsManager::RecordUntuned( std::ofstream& untuned_file, const std
...
+  }
+
+  realtime_out_ = std::make_unique<std::ofstream>(filename, std::ios::out | std::ios::app);
+
+  if (!realtime_out_->good()) {
+    TORCH_WARN("TunableOp realtime append: failed to open '", filename,
+               "'; falling back to batch write on exit.");
+    realtime_out_.reset();
+    return;
+  }
```
naromero77amd, line 200: Do we still need this warning about falling back to batch write on exit?

naromero77amd (commented):
See my questions in the InitRealTimeAppend method -- the warning is a bit confusing since we don't really have a batch write on exit.

jeffdaily (approved):
LGTM, pending clean CI.

aten/src/ATen/cuda/tunable/Tunable.cpp
jeffdaily: What situations would cause a failure?  If a failure happens, shouldn't we warn?  Or will it be too verbose?

naromero77amd: My apologies for my vague comment. I think my point is that we don't have "batch write on exit" any more since we got rid of it. So do we need to update the warning message?

sarthaktandon9amd: Yes, the warning message was redundant since we removed batch_write_on_exit, however, we still need to have a warning in case the file fails to open for writing. I have updated the warning message.

PR #163966, opened 2025-09-26
Fix TMA transpose logic to handle 1D shapes + string differences

njriasan:
Fixes #163702.

This fixes 2 issues:
1. The value may inconsistently be a shape or string. This normalizes to handle both of these.
2. 1D shapes should not transpose data. This fixes the order of operations to prevent this.


torch/_inductor/codegen/triton.py
```diff
@@ -2546,13 +2546,24 @@ def codegen_block_ptr_store_line(self, name, indexing, block_ptr, value, other="
...
+                return tuple(
+                    symt.name if isinstance(symt, sympy.Symbol) else symt
+                    for symt in shape
                 )
+
+            final_shape = stringify_shape(indexing.final_shape)
+            forward_shape = stringify_shape(value.shape)
+            reverse_shape = stringify_shape(value.shape[::-1])
+            # Don't attempt to transpose for 1D shapes.
+            if final_shape != forward_shape:
```
eellison: What here represents that this is a 1D shape ? i dont follow

eellison (commented):
can you add a reprentative test ?

torch/_inductor/codegen/triton.py
njriasan: I can remove the comment. The issue was the in the old version I would check if the transposed order matches (which it does for 1D) and if so generate a `tl.trans`. The change here is to first just check if they match exactly. I'll remove the comment.

njriasan:
> can you add a reprentative test ?

Does this need a new test? This was fixing an existing test: `python test/inductor/test_torchinductor_strided_blocks.py TritonTensorDescriptorTestCUDA.test_broadcast_prefer_nd_tiling_False_x_size1_y_size1_cuda`. I can update that test to verify the 1D result isn't transposed if that's the check you want.

njriasan:
Okay there are actually more issues across the suite of `TritonTensorDescriptorTestCUDA` tests. I'll work on a more comprehensive solution.

eellison (approved):

PR #119750, opened 2024-02-13
[RelEng] Define `BUILD_BUNDLE_PTXAS`

malfet:
That would bundle PTXAS into a `bin` folder

When compiling for Triton, define `TRITION_PTXAS_PATH` if `ptxas` is bundled with PyTorch Needed to make PyTorch compiled against CUDA-11.8 usable with 11.8 driver, as Triton is bundled with latest (CUDA-12.3 at time of PyTorch-2.2 release) ptxas

Needs https://github.com/pytorch/builder/commit/5c814e2527b3f5797488bf57d9d5425e63dcc1ac to produce valid binary builds

Test plan: 
- Create dummy ptxas in `torch/bin` folder and observe `torch.compile` fail with backtrace in Triton module.
- Run following script (to be added to binary tests ) against CUDA-11.8 wheel:
```python
import torch
import triton

@torch.compile
def foo(x: torch.Tensor) -> torch.Tensor:
  return torch.sin(x) + torch.cos(x)

x=torch.rand(3, 3, device="cuda")
print(foo(x))
# And check that CUDA versions match
cuda_version = torch.version.cuda
ptxas_version = triton.backends.nvidia.compiler.get_ptxas_version().decode("ascii")
assert cuda_version in ptxas_version, f"CUDA version mismatch: torch build with {cuda_version}, but Triton uses ptxs {ptxas_version}"
```

Fixes https://github.com/pytorch/pytorch/issues/119054

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @peterbell10 @ipiszy @yf225 @chenyang78 @kadeng @muchulee8 @aakhundov @ColinPeppler

PR #163972, opened 2025-09-26
[CD][CUDA][Triton][PTXAS] Turn on  BUILD_BUNDLE_PTXAS=1 for CUDA13 X86 Wheel Build

nWEIdia:
Triton (release/3.5.x) by default ships CUDA12.8 ptxas. 
This PR tries to bundle a ptxas version for cuda13, so that it can help #163801 when users run on new devices like THOR and Spark. 

Fixes #163801 

Test Plan: 
1) Check binary size increase against nightly or v2.9RC 
2) Install the binary from into a working THOR and B200/H100 machine (reproduce the original issue first on THOR), then install the binary built from this PR and we expect the issue to be gone without any additional user setting. Testing on B200 is to ensure no regression.  

[Update: since the original ptxas issue seems to be an ARM specific issue - THOR device is on ARM], This PR seems not appropriate. 

Reference: https://github.com/pytorch/pytorch/pull/119750 and https://github.com/pytorch/builder/commit/5c814e2527b3f5797488bf57d9d5425e63dcc1ac 

cc @ptrblck @eqy @tinglvv @atalman @malfet

### DIFF
--- a/.ci/manywheel/build_cuda.sh
+++ b/.ci/manywheel/build_cuda.sh
@@ -121,8 +121,8 @@ if [[ $CUDA_VERSION == 12* || $CUDA_VERSION == 13* ]]; then
     # Compress the fatbin with -compress-mode=size for CUDA 13
     if [[ $CUDA_VERSION == 13* ]]; then
         export TORCH_NVCC_FLAGS="$TORCH_NVCC_FLAGS -compress-mode=size"
-	# Bundle ptxas into the cu13 wheel, see https://github.com/pytorch/pytorch/issues/163801
-	export BUILD_BUNDLE_PTXAS=1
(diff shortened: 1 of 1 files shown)


atalman (approved):

malfet (approved):
LGTM, but please update test plan
And propose followup PR for aarch64

nWEIdia:
> LGTM, but please update test plan And propose followup PR for aarch64

Thanks @malfet ! I updated the test plan and changed the title to reflect this is only x86. I will create a follow up for ARM CUDA13 wheel build, thanks for catching it!

nWEIdia:
Closing as the new ptxas that comes with cuda13 is mostly dealing with ARM based GPU addition. So having a cuda13 ptxas in x86 cuda binary does not have clear help. 
Prioritizing #163988 instead.

PR #163975, opened 2025-09-26
Bce with logits loss extreme

### LINKED ISSUE #102894: BCELoss and BCEWithLogitsLoss differ when one of the input logits is float("inf")
### 🐛 Describe the bug

Based on the docs, it seems like `BCEWithLogitsLoss` is supposed to be the more numerically table sibling of `BCELoss`. However, when one of the input logits to `BCEWithLogitsLoss` is `float("inf")` the result is `Nan` whereas `sigmoid(inf) = 1.0`, so it should probably be well defined. When I use `BCELoss(sigmoid(x), target)` it gives the desired result.

```
x = torch.tensor([float("inf"), 1.0])
y = torch.tensor([1.0, 0.0])

loss = F.binary_cross_entropy_with_logits(x, y)
print (loss)

loss = F.binary_cross_entropy(x.sigmoid(), y)
print (loss)

print(torch.nn.BCEWithLogitsLoss()(x, y))
print(torch.nn.BCELoss()(x.sigmoid(), y))
```

Output:

```
tensor(nan)
tensor(0.6566)
tensor(nan)
tensor(0.6566)
```

### Versions

1.12.1

cleonard530:
Fixes #102894 

Updated the Loss Function binary_cross_entropy_with_logits to handle extreme values in a similar manner as binary_cross_entropy.

BCELoss inputs $x\in[0, 1]$ and thresholds $ln\_x := \max(\log(x), -100)$ (similarly for $\log(1-x)$). 

BCEWithLogitsLoss, which uses $\log(\sigma(x))$ and $\log(1-\sigma(x))$ for any $x\in[-\infty, \infty]$, did not have any threshold before. This caused inconsistency between the two methods and without any threshold arbitrarily large loss values could cause major problems in training. These commits fix this issue by clamping 'input = at::clamp(input, -100, 100)' (Note: $\log(\sigma(x)) \approx x$ for $x \leq -100$ and $\log(1-\sigma(x))) \approx -x$ for $x\geq100$). 

This fix is tested in the C++ unit test FunctionalTest::BCEWithLogitsLoss in 'test/cpp/api/functional.cpp'. 

Note: BCELoss does not update gradient for extreme cases in backwards so I didn't update the BCELossWithLogits backwards either.

@janeyx99

### DIFF
--- a/aten/src/ATen/native/Loss.cpp
+++ b/aten/src/ATen/native/Loss.cpp
@@ -1,6 +1,5 @@
 #define TORCH_ASSERT_ONLY_METHOD_OPERATORS
 #include <ATen/core/Tensor.h>
-#include <ATen/ops/clamp.h>
 #include <ATen/core/Reduction.h>
 #include <ATen/Dispatch.h>
 #include <ATen/TensorIterator.h>
@@ -19,6 +18,7 @@
 #include <ATen/ops/binary_cross_entropy_backward_native.h>
 #include <ATen/ops/binary_cross_entropy_native.h>
 #include <ATen/ops/binary_cross_entropy_with_logits_native.h>
+#include <ATen/ops/clamp.h>
 #include <ATen/ops/clamp_min.h>
 #include <ATen/ops/cosine_embedding_loss_native.h>
 #include <ATen/ops/empty.h>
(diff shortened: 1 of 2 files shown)


aten/src/ATen/native/Loss.cpp
```diff
@@ -347,14 +348,24 @@ Tensor& binary_cross_entropy_backward_out_cpu(const Tensor& grad, const Tensor&
...
-  auto log_sigmoid_input = at::log_sigmoid(input);
+  // Binary cross entropy with logits tensor is defined by the equation:
+  // L = -w (p y ln(sigma(x)) + (1-y) ln(1-sigma(x)))
+  // Which can be simplified to
+  // L = -w (x (1-y) - ln(sigma(x))(1+y(p-1))
+  // => L = -w (x (1-y) - ln(sigma(x)), when p==1.
+
+  // // Threshold when ln(sigma(x)) or ln(1-sigma(x)) < -100 to be consistent with binary_cross_entropy
+  // // Note: ln(sigma(x)) ~= x if x < -100 and ln(1-sigma(x)) ~= -x for x > 100
+  Tensor input_clamped = at::clamp(input, -100, 100);
```
janeyx99, line 353: Is the 100 here chosen based on https://github.com/pytorch/pytorch/blob/148a605a8fdb3cab13bda7ef263828ab0870ea39/aten/src/ATen/native/Loss.cpp#L289-L292 ?

cc @mikaylagawarecki @jbschlosser  to confirm that this mathematical change is acceptable (I think it is, but i'm not sure about the underlying math for the loss functions enough to be super sure)

cleonard530: @janeyx99 Yes, the value of 100 was just chosen to be consistent with binary_cross_entropy_out. Note, I am putting the threshold on the input and not the log(sigma(input)) like binary_cross_entropy_out is, but for these extreme values the results almost identical ($|\ln(\sigma(-100)) - (-100)| \approx  e^{-100}$). Using the threshold on log(sigma(input)) and log(1-sigma(input)) would have required extra computations that binary_cross_entropy_with_logits can otherwise avoid.

cleonard530:
Hi @jbschlosser, just following up on this PR, please let me know if there’s anything I can adjust or clarify. Thanks!

PR #163566, opened 2025-09-22
Use _RecordFunctionFast for torch.autograd.profiler.record_function

azahed98:
This changes `record_function` to use _RecordFunctionFast, which is more efficient. This is important for Compiler runtime overhead in smaller graphs. Currently this profiler takes half of the overhead.

To measure impact on runtime overhead, here are results from [benchmarks/dynamo/microbenchmarks/overheads.py](https://github.com/pytorch/pytorch/blob/main/benchmarks/dynamo/microbenchmarks/overheads.py).

Before changes:
```
requires_grad=False
compiled 56.9us (warmup=19.3s)

requires_grad=True
compiled 101.6us (warmup=0.2s)

inference_mode()
compiled 53.4us (warmup=0.1s)
```

After changes:
```
requires_grad=False
compiled 27.2us (warmup=9.9s)

requires_grad=True
compiled 61.6us (warmup=0.1s)

inference_mode()
compiled 25.3us (warmup=0.1s)
```

PR #163976, opened 2025-09-26
[Inductor] CompiledFxGraph replace record_function with _RecordFunctionFast

azahed98:
This replaces `record_function` in `CompiledFxGraph.__call__` with `_RecordFunctionFast`, which is less expensive in scenarios that don't require tracing (such as runtime). This should help with more accurate profiles.

`TestExecutionTrace.test_execution_trace_with_pt2` in https://github.com/pytorch/pytorch/blob/main/test/profiler/test_execution_trace.py#L372 already checks that the `record_function` region is tracked during profiling.

Note: This is related to #163566, but this avoids the failures coming up with uses of `record_function` elsewhere.

### DIFF
--- a/torch/_inductor/output_code.py
+++ b/torch/_inductor/output_code.py
@@ -50,7 +50,6 @@ from torch._inductor.utils import (
     output_node,
     set_tracing_context_output_strides,
 )
-from torch.autograd.profiler import record_function
 from torch.utils._ordered_set import OrderedSet
 from torch.utils._python_dispatch import is_in_torch_dispatch_mode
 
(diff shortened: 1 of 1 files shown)


mlazos (approved):

torch/_inductor/output_code.py
```diff
@@ -609,7 +608,7 @@ def __call__(self, inputs: Sequence[Any]) -> Any:
         try:
             # Checking the profiler directly is faster than nullcontext
             if torch.autograd.profiler._is_profiler_enabled:
-                with record_function(
+                with torch._C._profiler._RecordFunctionFast(
```
sraikund16, line 632: you need to add a kwarg `scope` as done here https://github.com/pytorch/pytorch/pull/162661 so that the event will be seen as a user annotation instead of a function

sraikund16 (changes requested):
see comment

torch/_inductor/output_code.py
azahed98: Forgot again! Added the `keyword_values` argument with the scope.

sraikund16 (approved):

Issue #163847, opened 2025-09-25
DISABLED test/distributed/test_nvshmem.py::NVSHMEMAll2AllTest::test_all_to_all_vdev_2d_offset


IvanKobzarev:

IvanKobzarev:
cc @fegin

PR #163985, opened 2025-09-26
[CI] Re-enable test_all_to_all_vdev_2d_offset

kwen2501:
* __->__ #163985

Fixes https://github.com/pytorch/pytorch/issues/163847
Moving allocations upfront and collectives later. The hang goes away. 

My investigation indicates that the hang is inside the last call `torch.testing.assert_close(out_expected, out[:out_numel])`. Rank 3 calls into it, but never gets out. Don't know why yet. I will investigate more.


fegin (approved):
Thanks for the investigation

PR #163899, opened 2025-09-25
[BE] Refresh documentation for stable ABI / API

mikaylagawarecki:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #163991
* __->__ #163899

PR #163991, opened 2025-09-26
FC/BC policy for libtorch stable ABI

mikaylagawarecki:
* __->__ #163991
* #163899


docs/source/notes/libtorch_stable_abi.md
```diff
@@ -128,3 +128,26 @@ The above is relevant in two places:
...
+### Versioning and Forward/Backward compatibility guarantees
+
+We provide a `TORCH_ABI_VERSION` macro in `torch/headeronly/version.h` of the form
+
+```
+[ byte ][ byte ][ byte ][ byte ][ byte ][ byte ][ byte ][ byte ]
+[MAJ   ][ MIN  ][PATCH ][                 ABI TAG              ]
+```
+
+In the present phase of development, APIs in the C-shim will be versioned based on major.minor.patch release that they are first introduced in, with 2.10 being the first release where this will be enforced. The ABI tag is reserved for future use.
```
janeyx99, line 149:
```suggestion
In the present phase of development, APIs in the C-shim will be versioned based on the major.minor.patch release that they are first introduced in, with 2.10 being the first release where this will be enforced. The ABI tag is reserved for internal use.
```

no promises

docs/source/notes/libtorch_stable_abi.md
```diff
@@ -128,3 +128,26 @@ The above is relevant in two places:
...
+
+Extensions can select the minimum abi version to be compatible with using:
+
+```
+#define TORCH_TARGET_VERSION …
```
janeyx99: let's also give folks an example of what it could be (like the raw numbers) so they can copy/paste and adjust

docs/source/notes/libtorch_stable_abi.md
```diff
@@ -128,3 +128,26 @@ The above is relevant in two places:
...
+```
+
+before including any stable headers or by passing the equivalent `-D` option to the compiler. Otherwise, the default will be the current `TORCH_ABI_VERSION`.
+
+The above ensures that if a user defines `TORCH_TARGET_VERSION` as 2.9 and attempts to use a C shim API `foo` that was introduced in version 2.10 and  a compilation error will be raised. Similarly, C++ wrappers in `torch/csrc/stable` are compatible with older libtorch binaries up to the TORCH_FEATURE_VERSION they are exposed in and always forward compatible with newer libtorch binaries.
```
janeyx99:
```suggestion
The above ensures that if a user defines `TORCH_TARGET_VERSION` to be 0x0209000000000000 (2.9) and attempts to use a C shim API `foo` that was introduced in version 2.10, a compilation error will be raised. Similarly, the C++ wrapper APIs in `torch/csrc/stable` are compatible with older libtorch binaries up to the TORCH_FEATURE_VERSION in which they are exposed and always forward compatible with newer libtorch binaries.
```

>  always forward compatible with newer libtorch binaries
hm. maybe we shouldn't promise always.

docs/source/notes/libtorch_stable_abi.md
```diff
@@ -128,3 +128,26 @@ The above is relevant in two places:
...
+
+The forward and backward compatibility of StableIValue conversions will be handled appropriately and a user need not consider this when choosing `TORCH_TARGET_VERSION`. This means that if, for example, the memory layout of a `torch/headeronly` struct `T` was changed from version 2.9 to 2.10 and the user targets 2.9 when building, they can be assured that the StableIValue conversion will handle the conversion from the user extension `T` to libtorch `T` appropriately and vice versa.
```
janeyx99: Should we just not mention this then

docs/source/notes/libtorch_stable_abi.md
mikaylagawarecki: I think this might indeed by future rather than internal (I'm thinking we will use it if we want to allow users to depend on nightlies and bump the version more frequently than each release when the ABI becomes more mature).

janeyx99 (approved):
For aoti_torch_call_dispatcher_v2, I'm okay with keeping the docs here still but warning that we don't make any guarantees on the underlying ops changing.

docs/source/notes/libtorch_stable_abi.md
```diff
@@ -124,10 +124,36 @@ The above is relevant in two places:
...
+
+Extensions can select the minimum abi version to be compatible with using:
+
+```
+#define TORCH_TARGET_VERSION (((0ULL + major) << 56) | ((0ULL + minor) << 48))
+```
+
+before including any stable headers or by passing the equivalent `-D` option to the compiler. Otherwise, the default will be the current `TORCH_ABI_VERSION`.
+
+The above ensures that if a user defines `TORCH_TARGET_VERSION` to be 0x0209000000000000 (2.9) and attempts to use a C shim API `foo` that was introduced in version 2.10, a compilation error will be raised. Similarly, the C++ wrapper APIs in `torch/csrc/stable` are compatible with older libtorch binaries up to the TORCH_FEATURE_VERSION they are exposed in and forward compatible with newer libtorch binaries.
```
janeyx99: Wait do users know what TORCH_FEATURE_VERSION is? Should we say TORCH_TARGET_VERSION here?

janeyx99: Hm maybe i am confused,
```suggestion
The above ensures that if a user defines `TORCH_TARGET_VERSION` to be 0x0209000000000000 (2.9) and attempts to use a C shim API `foo` that was introduced in version 2.10, a compilation error will be raised. Similarly, the C++ wrapper APIs in `torch/csrc/stable` are compatible with older libtorch binaries up to the TORCH_ABI_VERSION in which they were introduced and forward compatible with newer libtorch binaries.
```
maybe it should be ABI?

mikaylagawarecki:
Build failure is unrelated to docs change

PR #163994, opened 2025-09-26
[inductor] Handle patterns where input/output nodes are the same

angelayi:
Fixes #ISSUE_NUMBER

### DIFF
--- a/torch/_inductor/pattern_matcher.py
+++ b/torch/_inductor/pattern_matcher.py
@@ -1127,6 +1127,10 @@ class GraphPatternEntry(PatternEntry):
 
 @dataclasses.dataclass
 class ReplacementPatternEntry(PatternEntry):
+    """
+    The replacement pattern for the graph
+    """
+
     normalize_args: Callable[..., list[Any]]
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


torch/_inductor/pattern_matcher.py
```diff
@@ -1245,13 +1250,21 @@ def replace(
...
+                    return node not in added_replacement_nodes
+
                 if old is None:
                     assert new is None
                     return
                 assert isinstance(old, torch.fx.Node)
                 if new is None:
-                    old.replace_all_uses_with(None)  # type: ignore[arg-type]
-                    graph.erase_node(old)
+                    old.replace_all_uses_with(
```
angelayi, line 1271: I'm not really sure what runs into this case.. seems like `replace_all_uses_with` doesn't even allow `None` inputs?

jansel: Sounds like a bug?

jansel (approved):

torch/export/unflatten.py
```diff
@@ -1221,6 +1221,8 @@ def copy_sym_call_function(self, x):
         return node
 
     def remap_input(self, x):
+        print(x)
```
mlazos: Remove?

mlazos (approved):

torch/export/unflatten.py
angelayi: oops

PR #163999, opened 2025-09-26
Migrate DeviceType to torch/headeronly

janeyx99:
* __->__ #163999


c10/core/DeviceType.h
```diff
@@ -1,100 +1,16 @@
...
-// ATen/core (which would require a lot more build system hacking.)
-// If you modify me, keep me synchronized with that file.
-
 #include <c10/macros/Export.h>
 
-#include <cstddef>
-#include <cstdint>
-#include <functional>
+// If you modified DeviceType in caffe2/proto/caffe2.proto, please also sync
+// your changes into torch/headeronly/core/DeviceType.h.
```
mikaylagawarecki, line 6: I'm not sure this is true anymore, DeviceType there looks really out of date https://www.internalfb.com/code/fbsource/fbcode/caffe2/caffe2/proto/caffe2.proto 😆

mikaylagawarecki (approved):
To check my understanding -- we don't need to handle stable ivalue conversions specially (as we do for scalartype) because we only expect to add to the end of this enum, is that right?

For StableIValue conversions, do we need to have nice error handling for the edge case where extension (using the latest header) has a new `kFooDevice` and libtorch doesn't have it yet?

c10/core/DeviceType.h
janeyx99: The reason we don't have to handle StableIValue conversions for this one is cuz you can't currently pass DeviceType across the boundary (as we're not using it in any shim APIs today). I don't think DeviceType will be passed through the StableIValue boundary until Device, and then we will have to work out the versioning.

janeyx99: And lololol a good BE thing to do is to remove all the outdated stuff there eventually. I'm going to not worry about it for this PR though.

janeyx99: To clarify, once we pass this through any shim APIs, I anticipate we'll do something similar to to ScalarType

PR #164002, opened 2025-09-26
[torchfuzz] ones over zero

bobrenjc93:
* #164003
* __->__ #164002
* #163890
* #163812
* #163743

reduces likelihood of divide by zero errors. long term we'll probably want to just fuzz these values entirely

### DIFF
--- a/tools/experimental/dynamic_shapes/torchfuzz/operators/constant.py
+++ b/tools/experimental/dynamic_shapes/torchfuzz/operators/constant.py
@@ -76,19 +76,20 @@ class ConstantOperator(Operator):
                 import torch
 
                 default_values = {
-                    torch.float16: 0.0,
-                    torch.float32: 0.0,
-                    torch.float64: 0.0,
-                    torch.bfloat16: 0.0,
-                    torch.int8: 0,
-                    torch.int16: 0,
-                    torch.int32: 0,
-                    torch.int64: 0,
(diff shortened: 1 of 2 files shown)


pianpwk (approved):

tools/experimental/dynamic_shapes/torchfuzz/tensor_fuzzer.py
```diff
@@ -373,7 +373,7 @@ def fuzz_tensor(
 
     # Handle empty tensor case
     if len(size) == 0:
-        return torch.zeros((), dtype=dtype), seed
```
laithsakka, line 376: would it make aense to fuzz a value ditto for consantts

tools/experimental/dynamic_shapes/torchfuzz/operators/constant.py
```diff
@@ -76,19 +76,20 @@ def codegen(
...
-                    torch.bfloat16: 0.0,
-                    torch.int8: 0,
-                    torch.int16: 0,
-                    torch.int32: 0,
-                    torch.int64: 0,
-                    torch.bool: False,
-                    torch.complex64: 0.0,
-                    torch.complex128: 0.0,
+                    torch.float16: 1.0,
+                    torch.float32: 1.0,
```
laithsakka, line 80: why do not we fuzz a value

PR #164008, opened 2025-09-26
lint: Only include files in pytorch

seemethere:
* #164012
* __->__ #164008

We were seeing instances of stdlib files in clang-tidy output so this
just essentially removes them from the things that lintrunner will
report up. Longer term fix here would be to just modify the clang-tidy
configuration in order to do the correct thing here but that requires a
bit more investigation as to why this is only happening in CI and is not
reproduceable locally.


malfet (commented):
Well, we should have those in output of clang-tidy to begin with...

ZainRizvi (approved):
Very odd that these are getting included in clangtidy output. 

Did you try seeing if lintrunner include/exclude rules could fix this?

Either ways, approving this PR since it's mitigating a really bad customer experience for now

Issue #164010, opened 2025-09-26
[doc infra] `pip install -r docs/requirements.txt` will fail in most contexts


filipviz:
### 🐛 Describe the bug

There are a few rough edges in `.ci/docker/requirements-docs.txt` (which `docs/requirements.txt` is a symlink of):
1. `pip install -r docs/requirements.txt` pulls `numpy 2.x`, which is incompatible with the pinned versions of `matplotlib` and `tensorboard`.
2. The pinned `matplotlib` versions mean it's *only* possible to install these requirements on Python 3.9/3.10:
```
matplotlib==3.5.3 ; python_version < "3.13"
matplotlib==3.6.3 ; python_version >= "3.13"
```
because matplotlib 3.5.3 only supports Python <=3.10, and matplotlib 3.6.3 only supports Python <=3.11.

I imagine there are some historical constraints I may not be aware of.
- If we *can't* upgrade the pinned `matplotlib` and `tensorboard` versions, we should at least pin `numpy<2` and make sure the library versions we're pinning are compatible with the Python versions they're pinned for.
- If we *can* upgrade `matplotlib` and `tensorboard`, we'd ideally upgrade to versions which support `numpy>2`, obviating the issue and possibly enabling the use of a single environment for development and building the docs.

Perhaps this should be a separate issue, but we should also update the dependencies in `tools/nightly.py`. At a glance, it's missing `expecttest` (causes errors while running tests). It includes `sphinx`, which makes me think it's intended to work for building the docs - but if that's true, it's missing other dependencies it'd need like `matplotlib`.

Also see #163789 - opening this as a separate issue as per @janeyx99's suggestion.

```bash
python3.10 -m venv
source venv/bin/activate
pip install -r docs/requirements.txt
python -c "import matplotlib.pyplot"  # fails with ImportError
```

### Versions

OS: macOS 26.0 (arm64)
Python version: 3.10.18 (main, Jun  3 2025, 18:23:41) [Clang 17.0.0 (clang-1700.3.19.1)] (64-bit runtime)
Python platform: macOS-26.0-arm64-arm-64bit
CPU: Apple M3 Pro

filipviz:
Another rough edge I just ran into in `tools/nightly.py` - `pull`'s argparse help page lists the `--prefix` flag, but it only seems to work with `checkout`:
```bash
python3 -V  # Python 3.13.7

python3.10 -m venv venv
source venv/bin/activate
python -V  # Python 3.10.18
[...]

PR #164011, opened 2025-09-26
[docs] update instructions for locally building docs

filipviz:
Updates the instructions for locally building the docs in `CONTRIBUTING.md`. In specific:
- Note that docs can only be built with Python 3.10. This may change soon; see #164010 for more context.
- Explain when/why users may need to install PyTorch via wheel, `tools/nightly.py`, or from source when building docs.
- Move instructions to build the docs as a PDF from the README to `CONTRIBUTING.md`.

Fixes #163789

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -473,7 +473,7 @@ In addition to the standard Google Style docstring formatting rules, the followi
 
 ### Building documentation
 
-Docs builds are tested with Python 3.10 and may fail with other Python versions. To build the documentation:
+Currently, the docs can only be built with Python 3.10. To build the documentation:
 
 1. Set up a Python 3.10 environment and install `pip`. We recommend using separate environments for PyTorch development and building the docs. For example:
 ```bash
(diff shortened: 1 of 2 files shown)


KunjShah01:
if its open can i work on it

CONTRIBUTING.md
```diff
@@ -476,41 +473,65 @@ In addition to the standard Google Style docstring formatting rules, the followi
...
+Or with `uv`:
 ```bash
-cd docs
-pip install -r requirements.txt
-# `katex` must also be available in your PATH.
-# You can either install katex globally if you have properly configured npm:
-# npm install -g katex
-# Or if you prefer an uncontaminated global executable environment or do not want to go through the node configuration:
-# npm install katex && export PATH="$PATH:$(pwd)/node_modules/.bin"
+uv venv -p 3.10 .venv-docs
```
svekars: I believe it's better to follow https://github.com/pytorch/pytorch#prerequisites and use conda where possible

CONTRIBUTING.md
```diff
@@ -476,41 +473,65 @@ In addition to the standard Google Style docstring formatting rules, the followi
...
-yarn global add katex
+# Installs KaTeX globally:
+npm install -g katex
+# If you would prefer to avoid the global installation, you can use:
+npm install katex && export PATH="$PATH:$(pwd)/node_modules/.bin"
 ```
-> If a specific version is required you can use for example `yarn global add katex@0.13.18`.
 
-3. Generate the documentation HTML files. The generated files will be in `docs/build/html`.
+Note that:
```
svekars: Admonitions are supported like this I think. 
```suggestion
> [!NOTE]
```

CONTRIBUTING.md
```diff
@@ -476,41 +473,65 @@ In addition to the standard Google Style docstring formatting rules, the followi
...
 
+5. Use `make` to build the docs. For example:
+```bash
+make html  # Generate HTML documentation in docs/build/html.
+make  # List available output formats.
+```
+
+Other output formats may have additional dependencies. For example, `pdflatex` must be available in order to build a PDF with `make latexpdf`.
+
+#### Building a PDF
```
svekars: there were instructions for this already. Why were they deleted?

CONTRIBUTING.md
filipviz: Deleted the `uv` example. For now, `conda` can be difficult - as far as I know, there isn't a straightforward way to use a `conda` env with `tools/nightly.py checkout` since the `--prefix` option only accepts a plain `venv`. We could recommend something like:
```bash
conda create -y -n docs python=3.10 pip
conda activate docs
./tools/nightly.py checkout
conda deactivate
source .venv-docs/bin/activate
```
But this seems circuitous. Hopefully we can simplify once the dependencies are updated - see #164010.

CONTRIBUTING.md
filipviz: They are! Turned the items in that list into admonitions - let me know if you'd like me to add them anywhere else.

CONTRIBUTING.md
filipviz: I moved them from the README to `CONTRIBUTING.md` because the workflow is somewhat niche. I also modified them, because:
- The original instructions stated that `texlive` was required. This is true, but `pdflatex` is the actual dependency (and the build will fail if one has `texlive` but not `pdflatex`).
- The `mactex` cask is quite heavy and not strictly necessary (`mactex-no-gui` is fine).

PR #164013, opened 2025-09-26
Fix unbacked replacement where LHS is purely backed expr and RHS is unbacked expr

ColinPeppler:
## Scenario
- If there's a `torch._check(backed_expr == unbacked_symbol)`
- then we should replace unbacked_symbol for backed_expr
- currently, we don't do that when generating inputs for autotune_at_compile_time

## Error traceback
```
$ python test/inductor/test_aot_inductor.py -k test_size_with_unbacked_add_expr_transitive
  ...
  File "/data/users/colinpeppler/pytorch/torch/_inductor/compile_fx.py", line 1696, in fx_codegen_and_compile
    return scheme.codegen_and_compile(gm, example_inputs, inputs_to_check, graph_kwargs)
  File "/data/users/colinpeppler/pytorch/torch/_inductor/compile_fx.py", line 1187, in codegen_and_compile
    dynamo_utils.preserve_rng_state(),
  File "/home/colinpeppler/.conda/envs/pytorch/lib/python3.12/contextlib.py", line 158, in __exit__
    self.gen.throw(value)
  File "/data/users/colinpeppler/pytorch/torch/_dynamo/utils.py", line 2236, in preserve_rng_state
    torch.cuda.set_rng_state(cuda_rng_state)  # type: ignore[possibly-undefined]
  File "/data/users/colinpeppler/pytorch/torch/cuda/random.py", line 79, in set_rng_state
    _lazy_call(cb)
  File "/data/users/colinpeppler/pytorch/torch/cuda/__init__.py", line 341, in _lazy_call
    callable()
  File "/data/users/colinpeppler/pytorch/torch/cuda/random.py", line 77, in cb
    default_generator.set_state(new_state)
torch.AcceleratorError: CUDA error: an illegal memory access was encountered
```

## Bad autotuning input generation
```
# assume unbacked_symint_fallback = 16
# we generate too small of an input (16)
buf11 = generate_example_value((16, 256), (256, 1), 'cuda:0', torch.float32, 0, (16, 256))
triton_poi_fused_ones_1.run(buf11, 4096, stream=stream0)

stream0 = get_raw_stream(0)
buf12 = generate_example_value((16, 256), (256, 1), 'cuda:0', torch.float32, 0, (16, 256))
buf13 = generate_example_value((16, 256), (256, 1), 'cuda:0', torch.float32, 0, (16, 256))
add_kernel_1.run(buf11, buf12, buf13, 4096, 16, 1, 1, stream=stream0)
del buf11, buf12

stream0 = get_raw_stream(0)
buf15 = generate_example_value((10500, 256), (256, 1), 'cuda:0', torch.float32, 0, (10500, 256))
triton_poi_fused_add_mul_2.run(buf2, buf13, buf15, 2688000, stream=stream0)
```

## Good autotuning input generation
```
# notice we generate with the proper size now (10500)
buf11 = generate_example_value((10500, 256), (256, 1), 'cuda:0', torch.float32, 0, (10500, 256))
triton_poi_fused_ones_1.run(buf11, 2688000, stream=stream0)

stream0 = get_raw_stream(0)
buf12 = generate_example_value((10500, 256), (256, 1), 'cuda:0', torch.float32, 0, (10500, 256))
buf13 = generate_example_value((10500, 256), (256, 1), 'cuda:0', torch.float32, 0, (10500, 256))
add_kernel_1.run(buf11, buf12, buf13, 2688000, 10500, 1, 1, stream=stream0)
del buf11, buf12

stream0 = get_raw_stream(0)
buf15 = generate_example_value((10500, 256), (256, 1), 'cuda:0', torch.float32, 0, (10500, 256))
triton_poi_fused_add_mul_2.run(buf2, buf13, buf15, 2688000, stream=stream0)
```

* __->__ #164013

### DIFF
--- a/torch/_inductor/sizevars.py
+++ b/torch/_inductor/sizevars.py
@@ -727,15 +727,31 @@ class SizeVarAllocator:
         if self.unbacked_replacements is not None:
             return self.unbacked_replacements
 
+        def all_symbols_are_unbacked(expr: Expr) -> bool:
+            return all(self.shape_env.is_unbacked_symint(s) for s in expr.free_symbols)
+
+        def all_symbols_are_backed(expr: Expr) -> bool:
+            return all(symbol_is_type(s, SymT.SIZE) for s in expr.free_symbols)
+
         def should_keep_src_dst(lhs: Expr, rhs: Expr):
             # assuming lhs is the expr to be replaced (src), rhs is the replacement (dst)
             # checking if we should keep them for the replacement rule or swap
+
+            if all_symbols_are_unbacked(lhs) and all_symbols_are_backed(rhs):
+                # prioritize replacing unbacked with backed expressions
+                # e.g. s0 + s1 == 3 * u0
+                return True
+            elif all_symbols_are_backed(lhs) and all_symbols_are_unbacked(rhs):
+                return False
             if lhs.has(rhs):
+                # handles cases where LHS is a sub-expression of the RHS.
+                # e.g. Max(2, u0) == s1 * Max(2, u0)
                 return True
             elif rhs.has(lhs):
                 return False
             else:
-                return lhs.compare(rhs) == 1  # see sympy.Basic.compare
+                # fallback to sympy.Basic.compare for a deterministic ordering
+                return lhs.compare(rhs) == 1
 
         self.unbacked_replacements = {}
         for assertions in self.shape_env.deferred_runtime_asserts.values():
--- a/torch/_inductor/sizevars.py
+++ b/torch/_inductor/sizevars.py
@@ -727,24 +727,18 @@ class SizeVarAllocator:
         if self.unbacked_replacements is not None:
             return self.unbacked_replacements
 
-        def all_symbols_are_unbacked(expr: Expr) -> bool:
-            return all(self.shape_env.is_unbacked_symint(s) for s in expr.free_symbols)
-
-        def all_symbols_are_backed(expr: Expr) -> bool:
-            return all(symbol_is_type(s, SymT.SIZE) for s in expr.free_symbols)
-
         def should_keep_src_dst(lhs: Expr, rhs: Expr):
             # assuming lhs is the expr to be replaced (src), rhs is the replacement (dst)
             # checking if we should keep them for the replacement rule or swap
 
-            if all_symbols_are_unbacked(lhs) and all_symbols_are_backed(rhs):
-                # prioritize replacing unbacked with backed expressions
-                # e.g. s0 + s1 == 3 * u0
+            if not has_free_unbacked_symbols(rhs):
+                # prioritize replacing unbacked exprs with backed expressions
+                # e.g. u0 + s3 ==> s0 + s1
                 return True
-            elif all_symbols_are_backed(lhs) and all_symbols_are_unbacked(rhs):
+            elif not has_free_unbacked_symbols(lhs):
                 return False
-            if lhs.has(rhs):
-                # handles cases where LHS is a sub-expression of the RHS.
+            elif lhs.has(rhs):
+                # handles cases where LHS is a sub-expression of the RHS
                 # e.g. Max(2, u0) == s1 * Max(2, u0)
                 return True
             elif rhs.has(lhs):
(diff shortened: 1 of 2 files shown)


torch/_inductor/sizevars.py
```diff
@@ -727,15 +727,31 @@ def _get_unbacked_replacements(self) -> dict[Expr, Expr]:
         if self.unbacked_replacements is not None:
             return self.unbacked_replacements
 
+        def all_symbols_are_unbacked(expr: Expr) -> bool:
+            return all(self.shape_env.is_unbacked_symint(s) for s in expr.free_symbols)
+
+        def all_symbols_are_backed(expr: Expr) -> bool:
+            return all(symbol_is_type(s, SymT.SIZE) for s in expr.free_symbols)
```
laithsakka: is expr.free_symbols iterstable or does it materialize all symbols?
make sure we dont do this mistake
https://github.com/pytorch/pytorch/pull/140027/files

torch/_inductor/sizevars.py
```diff
@@ -727,15 +727,31 @@ def _get_unbacked_replacements(self) -> dict[Expr, Expr]:
         if self.unbacked_replacements is not None:
             return self.unbacked_replacements
 
+        def all_symbols_are_unbacked(expr: Expr) -> bool:
+            return all(self.shape_env.is_unbacked_symint(s) for s in expr.free_symbols)
```
laithsakka: ditto like other place

torch/_inductor/sizevars.py
ColinPeppler: I'm using the `free_symbols` attribute stored on the `sympy.Expr` which stores the free symbols in a `Set`.

https://github.com/sympy/sympy/blob/6a47c4992f36cb298635b7240b662b4f5fad0a95/sympy/core/basic.py#L676-L723

torch/_inductor/sizevars.py
```diff
@@ -727,15 +727,31 @@ def _get_unbacked_replacements(self) -> dict[Expr, Expr]:
...
+
+        def all_symbols_are_backed(expr: Expr) -> bool:
+            return all(symbol_is_type(s, SymT.SIZE) for s in expr.free_symbols)
+
         def should_keep_src_dst(lhs: Expr, rhs: Expr):
             # assuming lhs is the expr to be replaced (src), rhs is the replacement (dst)
             # checking if we should keep them for the replacement rule or swap
+
+            if all_symbols_are_unbacked(lhs) and all_symbols_are_backed(rhs):
```
cp2923: n00b q, is all_symbols_are_backed(rhs) good enough? why do we need to check all_symbols_are_unbacked(lhs) here and all_symbols_are_unbacked(rhs) in line 744?

torch/_inductor/sizevars.py
```diff
@@ -727,15 +727,31 @@ def _get_unbacked_replacements(self) -> dict[Expr, Expr]:
...
             if lhs.has(rhs):
+                # handles cases where LHS is a sub-expression of the RHS.
+                # e.g. Max(2, u0) == s1 * Max(2, u0)
                 return True
             elif rhs.has(lhs):
                 return False
             else:
-                return lhs.compare(rhs) == 1  # see sympy.Basic.compare
+                # fallback to sympy.Basic.compare for a deterministic ordering
+                return lhs.compare(rhs) == 1
```
cp2923, line 748: I guess exsting issue come from: sympy.Basic.compare does not prefer backed symint over unbacked?

torch/_inductor/sizevars.py
laithsakka: shall we use an efficient API. like the one above?

cp2923 (approved):

torch/_inductor/sizevars.py
ColinPeppler: that's a good point, likely `all_symbols_are_backed(rhs)` is sufficient given deferred runtime assertions should always include an unbacked symbol today.

torch/_inductor/sizevars.py
ColinPeppler: it does prefer backed symbol over unbacked symbol, but not backed expression over unbacked symbol.

A symbol (backed or unbacked) is always preferred over an expression (e.g. sym1 + sym2)

torch/_inductor/sizevars.py
cp2923: maybe 'all_symbols_are_unbacked' -> 'some_symbols_are_backed'

laithsakka (approved):

PR #164016, opened 2025-09-26
[testing] upload test stats: Add info to the invoking file summary and some other changes

clee2000:
* Changes some internal logic for grouping so hopefully it's slightly less annoying write code for
* Changes the invoking file summary to just use file, which I think is correct most of the time
* Adds some fields to the file summary, like skips, errors, etc so I can reuse it for file report regression things

Output should be the same, maybe with slightly more fields since I got rid of some of the pops


huydhn (approved):
LGTM!

PR #164020, opened 2025-09-26
[Max Autotune][B200] Add addmm config to avoid test OOM

jananisriram:
Summary: Add a new `addmm` config that is small enough to not cause an OOM (out of memory error), since the configs for `blackwell_persistent_mm_configs`, which `addmm` used, are too large.

Test Plan: `test_max_autotune.py`

Differential Revision: D83378477

### DIFF
--- a/torch/_inductor/template_heuristics/triton.py
+++ b/torch/_inductor/template_heuristics/triton.py
@@ -315,6 +315,10 @@ class BaseConfigHeuristic(metaclass=BaseHeuristicSingleton):
             GemmConfig(128, 128, 128, 3, 8),
         ]
 
+        self.blackwell_persistent_addmm_configs: list[BaseConfig] = [
+            GemmConfig(256, 128, 64, 2, 4),
+        ]
+
         self.scaled_mm_configs: list[BaseConfig] = [
             GemmConfig(128, 256, 32, 3, 8),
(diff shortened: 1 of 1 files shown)


torch/_inductor/template_heuristics/triton.py
```diff
@@ -2094,7 +2098,8 @@ class CUDABlackwellAddmmPersistentTMATemplateConfigHeuristic(
 
     def __init__(self) -> None:
         super().__init__()
-        self.mm_configs = self.blackwell_persistent_mm_configs
+        # NOTE: to ensure that we pass tests, addmm needs a small config
+        self.mm_configs = self.blackwell_persistent_mm_configs + self.blackwell_persistent_addmm_configs
```
coconutruben: should we still keep the old ones in here for addmm then? or just make a smaller set?

coconutruben (approved):
LGTM you can take/reject the question but this looks safe and smart!

torch/_inductor/template_heuristics/triton.py
jananisriram: we probably want to use the same configs for addmm as we use for mm, so it doesn't make as much sense to me to duplicate the configs across both `mm_configs` and `addmm_configs`! i think adding the singular config here makes sense to avoid duplication.

njriasan (approved):

PR #164021, opened 2025-09-26
[Max Autotune][B200] Fix decompose_k test failure

jananisriram:
Summary:
Fix decompose_k test failure (`test_max_autotune_decompose_k `) in `test_max_autotune.py` on B200s by setting `torch._inductor.config` patches for variables `comprehensive_padding` and `shape_padding`. Initial failure was `AssertionError: False is not true : Could not find a split in {3, 9, 2187, 81, 243, 729, 27} in # AOT ID: ['6_forward']`. 

Refactor decompose_k test to follow patch semantics when setting all environment variables within a test.

Test Plan:
`test_max_autotune.py`:
```
buck2 test 'fbcode//mode/opt' fbcode//caffe2/test/inductor:max_autotune -c fbcode.nvcc_arch=b200a -c fbcode.enable_gpu_sections=true -c fbcode.platform010_cuda_version=12.8 -c fbcode.re_gpu_tests=False -- test_max_autotune_decompose_k
```

Differential Revision: D83390563

### DIFF
--- a/test/inductor/test_max_autotune.py
+++ b/test/inductor/test_max_autotune.py
@@ -1439,6 +1439,8 @@ class TestMaxAutotune(TestCase):
     @config.patch(
         max_autotune=True,
         max_autotune_gemm_backends="TRITON",
+        comprehensive_padding=False,
+        shape_padding=False,
     )
     def test_max_autotune_decompose_k(self, sizes, dtype, dynamic):
         fp16_red_setting = (


test/inductor/test_max_autotune.py
```diff
@@ -1438,17 +1439,14 @@ def f(x, y):
     @config.patch(
         max_autotune=True,
         max_autotune_gemm_backends="TRITON",
+        comprehensive_padding=False,
+        shape_padding=False,
+    )
+    @patch_matmul(
```
njriasan: Any reason you can't just reuse `config.patch` here? This feels like more infrastructure to maintain that isn't reused.

njriasan (commented):
See my comment on trying to reuse existing infrastructure.

test/inductor/test_max_autotune.py
jananisriram: afaik, `patch` only works for `config = torch._inductor.config` imports; it's not enabled for `torch.backends.cuda.matmul`

njriasan: Okay makes sense. I don't know enough about this part of the codebase to know if we want to unify eventually, but this seems safer for now.

njriasan (approved):

test/inductor/test_max_autotune.py
PaulZhang12: @jananisriram can we actually be explicit here in the test? Sorry for the added effort, but I don't think patch_matmul is super straightforward and would rather have the flags just explicitly set instead of using `patch_matmul` for just 1 test

torch/_inductor/template_heuristics/triton.py
```diff
@@ -315,6 +315,10 @@ def __init__(self) -> None:
             GemmConfig(128, 128, 128, 3, 8),
         ]
 
+        self.blackwell_persistent_addmm_configs: list[BaseConfig] = [
```
PaulZhang12, line 318: Can we add a todo that we need to tune these configs more? Are these for the persistent/tma templates? Why can't blackwell reuse the hopper ones?

test/inductor/test_max_autotune.py
jananisriram: fair enough, the reason for this decision was to make the modification of each flag safer (such that we don't miss resetting a flag accidentally). if any of the `torch.backends.cuda.matmul` flags will be used in the future, i think we should keep this!

jananisriram: let's remove it from here and figure out a clearer way to implement this patch for all tests in the future!

torch/_inductor/template_heuristics/triton.py
jananisriram: we needed a small enough config such that the test for addmm passes (otherwise it encountered an OOM). this modification shouldn't be in this pr and should be removed with the latest revision?

mlazos (approved):

eellison (approved):

PR #164031, opened 2025-09-27
[Replicate][Pipeline Parallelism] integration of new replicate function with pipeline parallelism

anshul-si:
**Summary:** In order to test numerics for replicate + pp, stage.py needs to be able to call replicate's backward manually as pipeline parallelism doesn't have this feature. 

**Test Case**
1.  pytest test/distributed/_composable/test_composability/test_pp_composability.py -k test_replicate_pp

* __->__ #164031
* #163897


torch/distributed/pipelining/stage.py
```diff
@@ -641,6 +642,30 @@ def perform_backward(
             else:
                 with self.submod.no_sync():  # type: ignore[operator]
                     result = perform_backward(backward_type)()
+
+        # If submod is a Replicate module
+        elif isinstance(self.submod, ReplicateModule):
```
weifengpy: I compared ReplicateModule and FSDPModule line by line, but seems they are almost identical. what's the core diffrerence?

anshul-si: The only difference is we need a way to distinguish modules that have been replicated vs fully-sharded so we know whether to get replicate state or fully shard state

H-Huang: Yeah, agree here. The only difference i see is from `fully_shard.state()` and `replicate.state()`. To reduce code duplication, i think you could define a helper method like `_handle_fsdp_module_backward` that takes in a lamba function.

I'm also not sure why `ReplicateModule` and `FSDPModule` are different modules if it seems like they support the same things. We should look into that first

test/distributed/_composable/test_composability/test_pp_composability.py
```diff
@@ -366,6 +368,238 @@ def apply_tp(
...
+
+            ref_pipeline_stage = PipelineStage(
+                ref_partial_model,
+                stage_idx,
+                pp_group.size(),
+                self.device,
+                group=pp_group,
+            )
+            ref_partial_models = [ref_pipeline_stage.submod]
+            ref_pipeline_schedule = ScheduleClass(
```
H-Huang, line 491: Your loss function is y.sum(), so I think you need `scale_grads=False`

test/distributed/_composable/test_composability/test_pp_composability.py
```diff
@@ -366,6 +368,238 @@ def apply_tp(
...
+                pipeline_schedule.step(inputs)
+                ref_pipeline_schedule.step(inputs)
+            elif is_last_stage:
+                losses = []
+                ref_losses = []
+                pipeline_schedule.step(target=labels, losses=losses)
+                ref_pipeline_schedule.step(target=labels, losses=ref_losses)
+
+                for loss, ref_loss in zip(losses, ref_losses):
+                    self.assertEqual(loss, ref_loss)
```
H-Huang, line 595: In a lot of our tests we compare single-device model grads with pipeline parallel model grads. This test just compares the loss of two pipelines. Might be out of scope to add this though, but maybe can add a TODO

torch/distributed/pipelining/stage.py
weifengpy: got you. is it possible to do this?

I don't mind unit test to be verbose. But I really hope the core code stay dry

```
elif isinstance(self.submod, FSDPModule):  # ReplicateModule is also a FSDPModule
      ...
      if isinstance(self.submod, ReplicateModule):
          state = replicate.state(self.submod)
```

H-Huang: oh yes, if ReplicateModule is also an FSDP Module, then I like what @weifengpy proposed we should do that!

test/distributed/_composable/test_composability/test_pp_composability.py
anshul-si: In this case, my only goals with this were to confirm that replicate + pipeline parallelism actually runs and that adding replicate to a pipeline doesn't affect it. I would prefer to make this a TODO that I can work on while running MAST tests in a few days

weifengpy (approved):
looking good on the fsdp part

H-Huang (approved):
Looks good!

PR #164035, opened 2025-09-27
[fx] Allow customization of submod name in split graph

### LINKED ISSUE #164030: HOP and pipelining both create submodules named `submod_i`
### 🐛 Describe the bug

Thus stepping on each other's toes.

For example:
HOP will translate the following code block:
```
with torch.no_grad():
        expert_counts.scatter_(1, topk_ids, 1)
        tokens_per_expert = expert_counts.sum(dim=0)
```
into:
```
submod_3 = self.submod_1
wrap_with_set_grad_enabled = torch.ops.higher_order.wrap_with_set_grad_enabled(False, submo
d_3, getitem_6);

class submod_1(torch.nn.Module):
         def forward(self, getitem_6: "i64[256, 6]"):
             scatter_: "i64[256, 64]" = torch.ops.aten.scatter_.value(new_zeros, 1, getitem_6, 1);  new_zeros = None
             sum_1: "i64[64]" = torch.ops.aten.sum.dim_IntList(scatter_, [0]);  scatter_ = None
```

pipelining, based on `split_module`, would also name its program partitions as `submod_i` -- hard coded in torch.fx.passes today:
https://github.com/pytorch/pytorch/blob/b48a3d0a38f69fbeeb4375e04ba45389cca12585/torch/fx/passes/split_module.py#L22

### Versions

main

kwen2501:
* #164078
* #164077
* #164037
* __->__ #164035
* #164045

Fixes #164030: HOP and pipelining both name things submod_i
by adding an optional argument `partition_affix` to `split_module` API.

### DIFF
--- a/torch/fx/passes/split_module.py
+++ b/torch/fx/passes/split_module.py
@@ -59,6 +59,7 @@ def split_module(
     keep_original_order: Optional[bool] = False,
     keep_original_node_name: Optional[bool] = False,
     keep_original_input_name: bool = True,
+    partition_affix: Optional[str] = None,
 ):
     """
     Creates subgraphs out of main graph
@@ -81,6 +82,7 @@ def split_module(
             have the same node names as the original graph.
         keep_original_input_name: bool: If the partitioned graphs should
             have the same input names as the original graph.
+        partition_affix: Optional[str]: If specified, the submodules' names will contain the affix, e.g. "submod_<affix>_<idx>".
 
     Returns:
         GraphModule: the module after split.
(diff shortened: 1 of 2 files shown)


torch/fx/passes/split_module.py
```diff
@@ -59,6 +59,7 @@ def split_module(
     keep_original_order: Optional[bool] = False,
     keep_original_node_name: Optional[bool] = False,
     keep_original_input_name: bool = True,
+    partition_affix: Optional[str] = None,
```
ezyang, line 63: Consider making this kwarg only

ezyang: Why is this "affix" and not "prefix"?

ezyang (approved):
suuure why not

torch/fx/passes/split_module.py
kwen2501: Because the eventual submod name will be `submod_abc_i`, where `abc` is the added argument here. I don't want to give user an illusion that their new submod starts with `abc`.

PR #164036, opened 2025-09-27
test(cuda): verify roundup_power2_divisions=1 rounds to next power-of-two

Aminsed:
### Motivation
`PYTORCH_CUDA_ALLOC_CONF` supports `roundup_power2_divisions`; value `1` should round to the next power-of-two. Currently it’s effectively ignored and default block rounding is used.

### Root cause
`round_size()` applies power-of-two rounding only for `divisions > 1`, and the helper asserts `divisions >= 2`. Thus, `divisions == 1` bypasses PoT logic.

### Approach
Honor `divisions == 1` in `round_size()` by rounding to the next power-of-two when `size > kMinBlockSize` and the size is not already a power-of-two. Keep existing behavior for `0` and `>= 2`. No parser changes.

### Tests
Added focused subprocess CUDA test `test/test_cuda_allocator_roundup.py` covering:
- `divisions == 1`, 514 MiB → ~1 GiB reserved
- `divisions == 1`, 1 GiB exact → unchanged
- `divisions == 0` → PoT rounding disabled
- `cudaMallocAsync` backend → setting ignored

### Backward compatibility
`divisions == 0` and `>= 2` unchanged; `cudaMallocAsync` unchanged. Behavior change only for `divisions == 1` and is opt-in via env.


test/test_cuda_allocator_roundup.py
```diff
@@ -0,0 +1,100 @@
...
+class TestCudaAllocatorRoundup(unittest.TestCase):
+    def test_roundup_divisions_one_next_power_of_two(self):
+        env = {
+            "PYTORCH_CUDA_ALLOC_CONF": "backend:native,roundup_power2_divisions:[>:1]",
+            # Request ~514MB allocation to test next-PoT rounding to 1GiB
+            "ALLOC_BYTES": str(514 * 1024 * 1024),
+        }
+        code, out, err = _run_subprocess_with_env(env)
+        # If CUDA not available in CI shard, accept skip via returncode 2
+        if code == 2:
```
ngimel: skips are handled by skip decorators

test/test_cuda_allocator_roundup.py
```diff
@@ -0,0 +1,100 @@
...
+
+
+def _run_subprocess_with_env(env):
+    cmd = [
+        sys.executable,
+        os.path.join(
+            os.path.dirname(__file__),
+            "..",
+            "tools",
+            "repro_allocator_roundup_divisions_one.py",
```
ngimel: this file doesn't exist

ngimel (commented):
You don't need to run tests in a subprocess, just do empty cache and make sure there are no large enough empty segments

test/test_cuda_allocator_roundup.py
Aminsed: fixed, tests now in test_allocator_settings under TestCuda class with skipIf decorator

test/test_cuda_allocator_roundup.py
Aminsed: fixed

test/test_cuda.py
```diff
@@ -4553,6 +4553,45 @@ def power2_div(size, div_factor):
             # not supported with the cudaMallocAsync backend
             self.assertEqual(pow2_div2_mem - start_mem, power2_div(nbytes_big, 2))
 
+        # test roundup_power2_divisions:1 rounds to next power of 2
```
eqy, line 4561: Not an issue introduced by this PR, but for anti-flakiness we probably want the @serialTest() decorator for this test as well

Aminsed: Fixed

PR #164037, opened 2025-09-27
[PP] Customize pipeline's submod name

kwen2501:
* #164078
* #164077
* __->__ #164037
* #164035
* #164045

Changing PP submodules' name from `submod_i` to `submod_pp_i` to distinguish from the submodule created by HOP.

### DIFF
--- a/torch/distributed/pipelining/_IR.py
+++ b/torch/distributed/pipelining/_IR.py
@@ -35,6 +35,16 @@ logger = logging.getLogger(__name__)
 # 2. Add parameter movement to split_module
 
 
+PP_SUBMOD_PREFIX = "submod_pp"
+
+
+def get_submod_name(stage_idx: int):
+    """Returns the name of the submod for a given stage index.
+    For example, "submod_pp_0", "submod_pp_1", etc.
+    """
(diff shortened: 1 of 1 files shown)


torch/distributed/pipelining/_IR.py
```diff
@@ -729,7 +741,7 @@ def split_callback(n: fx.Node):
 
         # TODO: what does split do with module invocations? does it move the modules
         # into the submodules?
-        split = split_module(traced, mod, split_callback)  # type: ignore[arg-type]
+        split = split_module(traced, mod, split_callback, partition_affix="pp")  # type: ignore[arg-type]
```
H-Huang, line 744: not sure if we have tests validating the submod name after splitting, might be good to have. Can do it in a followup PR

H-Huang (approved):
Looks good!

torch/distributed/pipelining/_IR.py
kwen2501: These are internal settings. And users are supposed to get submodules via:
`pipe.get_stage_module(i)`. 
If existing tests passed, we should be fine.

PR #164038, opened 2025-09-27
[DONT MERGE YET][dynamo][guards] Do not dict pop in framelocals dict creation

anijain2305:
* __->__ #164038

For this program

```
import torch

torch._dynamo.config.enable_cpp_symbolic_shape_guards = False

def fn(x):
    for _ in range(20000):
        x = x + 1
    return x

mod = torch.fx.symbolic_trace(fn)

print(mod)


opt_mod = torch.compile(mod, backend="eager", dynamic=True)

x = torch.randn(2, 2)
opt_mod(x)

```

It reduces the guard overhead from 1.8 ms to 72 us. The dict pop is not required as explained in the comment.

It can reduce to 46 us if we just remove the _dict.attr pop branch.


Lucaskabela (approved):

mlazos (approved):
really cool speedup!

williamwen42 (approved):

torch/csrc/dynamo/framelocals_mapping.cpp
```diff
@@ -145,16 +145,22 @@ FrameLocalsMapping::FrameLocalsMapping(FrameLocalsFrameType* frame)
...
     PyObject* value = _framelocals[i].ptr();
+    PyObject* name_ptr = PyTuple_GET_ITEM(framelocals_names_ptr, i);
     if (value == nullptr) {
-      _dict.attr("pop")(framelocals_names[i], py::none());
+      if (name_ptr != nullptr && seen.count(name_ptr) != 0) {
+        _dict.attr("pop")(framelocals_names[i], py::none());
+      }
     } else {
       _dict[framelocals_names[i]] = value;
+      seen.insert(name_ptr);
```
williamwen42, line 163: Actually, is there a guarantee that entries in framelocals_names with the same string value are the string object? If not, then the unordered_set should use Python string compare for its hash/equality? (We should also try to craft a test case that fails if this update is done incorrectly - I recall encountering failing tests when I made mistakes in the initial implementation, so I'm confused as to why the tests today don't seem to catch incorrect updates.)

anijain2305: Yeah, that was my concern as well. I don't know how to make that test case though.

anijain2305:
Not merging yet. @williamwen42  has concerns and he will take over this PR.

PR #164039, opened 2025-09-27
Enable dynamo `match` tests

cyyever:
Enable the commented tests for Python 3.10

### DIFF
--- a/test/dynamo/test_functions.py
+++ b/test/dynamo/test_functions.py
@@ -2302,30 +2302,29 @@ class FunctionTests(torch._dynamo.test_case.TestCase):
 
         return augment(x)
 
-    # # This is to test the new syntax for pattern matching
-    # # ("match ... case ...") added on python 3.10.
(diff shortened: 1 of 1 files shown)


test/dynamo/test_functions.py
```diff
@@ -2302,30 +2302,29 @@ def augment(x: torch.Tensor) -> torch.Tensor:
...
-    #     param = {"a": 0.5}
-    #     match param:
-    #         case {"a": param}:
-    #             return x * param
-    #         case {"b": param}:
-    #             return x / param
+    # This is to test the new syntax for pattern matching
+    # ("match ... case ...") added on python 3.10.
+    @make_test
+    def test_match_sequence(a):
```
guilhermeleobas, line 2308: I guess skipping the test if Python < 3.10 won't work, right? The entire test will fail with a parsing error.

cyyever: We have dropped Python <3.10 support.

PR #164044, opened 2025-09-27
Speed up FP precision lookup

### LINKED ISSUE #161822: [CUDA][cuBLAS] #125888 introduces measurable CPU overhead for matmuls

lakshayg:
This commit simplifies the precision lookup and setting logic
by reducing the number of branches and using a custom hash
function. Fixes #161822. The issue described in #163709 still
persists. This is meant as a short term fix.

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @voznesenskym @penguinwu @EikanWang @Guobing-Chen @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela

### DIFF
--- a/aten/src/ATen/Context.h
+++ b/aten/src/ATen/Context.h
@@ -25,6 +25,7 @@
 #include <c10/util/CallOnce.h>
 #include <c10/util/Exception.h>
 #include <c10/util/env.h>
+#include <c10/util/hash.h>
 #include <c10/util/irange.h>
 
 #include <cstdint>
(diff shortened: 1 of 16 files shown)


eqy (commented):
Thanks! What's the estimated performance difference between the old `float32Precision` and new version? IIRC this was the main issue as all matmuls (even ones that were not affected by TF32 setting) would call it

aten/src/ATen/Context.h
```diff
@@ -488,23 +489,36 @@ class TORCH_API Context {
...
-        {"conv", "tf32"},
-        {"rnn", "tf32"},
-        {"all", "none"}}},
+  // manually designed hash function
+  struct BackendOpHash {
+    size_t operator()(const std::pair<std::string, std::string>& p) const {
+      if (C10_UNLIKELY(p.first.empty() || p.second.empty())) {
+        return 0ull;
+      }
+      return 2 * static_cast<size_t>(p.first[0]) +
```
ngimel: this is super unsafe, what if at some point we have keys that start with the same letter? Can we switch to enums to avoid this? As was discussed in the previous issues, they, for some weird reason, want string on the python side, fine, but on c++ we should just have enums.

lakshayg: > this is super unsafe, what if at some point we have keys that start with the same letter

If that happens, we will get hash collisions which will reduce the performance but will not lead to incorrect behavior.

> ... but on c++ we should just have enums

I agree that using enums is the best solution here but if I make that change, it will require changing all calls to these functions in the C++ code. My intention was to make a small and local change that restores performance while we discuss the right overall design in #163709.

I am happy to change it to enums right away if you think that the discussion will take a long time to conclude.

ngimel: Yes we should change all the callsites in c++.

lakshayg: I have updated the PR.

ngimel (approved):

eqy (approved):

aten/src/ATen/Context.h
```diff
@@ -488,23 +495,33 @@ class TORCH_API Context {
...
-        {"all", "none"}}},
+  struct BackendOpHash {
+    size_t operator()(const std::pair<Float32Backend, Float32Op>& key) const {
+      size_t k1{static_cast<size_t>(key.first)};
+      size_t k2{static_cast<size_t>(key.second)};
+      return std::hash<size_t>{}((k1 << 32) | k2);
+    }
   };
 
+  std::unordered_map<
```
swolchok: this can be improved further by using nested arrays as in #164387

lakshayg: Yeah, I considered doing that but decided not to since the hastable approach is immune to the possibility of someone accidentally changing the enum order or forgetting to update the mapping when they add/remove backends.

I also think that this design needs to be revisited (#163709) and wanted to avoid potentially throw-away work.

It's a valid comment though. Feel free to take it up in a separate PR.

aten/src/ATen/Context.h
```diff
@@ -488,23 +495,33 @@ class TORCH_API Context {
...
-             : "tf32"},
-        {"conv", "tf32"},
-        {"rnn", "tf32"},
-        {"all", "none"}}},
+  struct BackendOpHash {
+    size_t operator()(const std::pair<Float32Backend, Float32Op>& key) const {
+      size_t k1{static_cast<size_t>(key.first)};
+      size_t k2{static_cast<size_t>(key.second)};
+      return std::hash<size_t>{}((k1 << 32) | k2);
+    }
```
ngimel: btw, don't need a custom hash, can use one from c10
```
using Key = std::pair<Float32Backend, Float32Op>;
using Map = std::unordered_map<Key, Float32Precision, c10::hash<Key>>;
```

yangw-dev:
@pytorchbot revert -c ghfirst -m "broke internal build In file included from xplat/caffe2/aten/src/ATen/DeviceAccelerator.cpp:1:
xplat/caffe2/aten/src/ATen/Context.h:502:38: error: shift count >= width of type [-Werror,-Wshift-count-overflow]
  502 |       return std::hash<size_t>{}((k1 << 32) | k2);"

aten/src/ATen/Context.h
lakshayg: Thanks! I will update the PR

ngimel: That should actually fix ` shift count >= width` error

PR #164045, opened 2025-09-27
[PP] Use default export mode (non-strict)

kwen2501:
* #164078
* #164077
* #164037
* #164035
* __->__ #164045

export's default mode has switched from strict to non-strict. We just follow suit in PP.


H-Huang (approved):
LGTM!

PR #164049, opened 2025-09-27
[CUDA] fix indexing on large tensor causing nvalid configuration argument

### LINKED ISSUE #164048: [CUDA] indexing on large tensor causes invalid configuration argument
### 🐛 Describe the bug

Reproducer:
```
import torch

mask = torch.randint(0, 20, (4, 87, 1056, 736), device="cuda")
to_apply = torch.tensor([True, False, False, True], device="cuda")
mask[to_apply]
```

### Versions

build on main from today

Isalia20:
Fixes #164048

### DIFF
--- a/aten/src/ATen/native/cuda/IndexKernel.cu
+++ b/aten/src/ATen/native/cuda/IndexKernel.cu
@@ -90,7 +90,6 @@ void gpu_index_kernel(TensorIteratorBase& iter, const IntArrayRef index_size, co
         const int max_grid_y = at::cuda::getCurrentDeviceProperties()->maxGridSize[1];
         // if it's an eligible grid we use the fast path, otherwise default to slower path
         if (blocks_per_slice_upper <= max_grid_y) {
-          if (iter.numel() == 0) return;
           at::native::vectorized_gather_kernel_launch<alignment, int64_t>(out_ptr, in_ptr, (int64_t*)iter.data_ptr(2), num_ind,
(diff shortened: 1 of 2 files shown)


aten/src/ATen/native/cuda/IndexKernel.cu
```diff
@@ -83,11 +84,18 @@ void gpu_index_kernel(TensorIteratorBase& iter, const IntArrayRef index_size, co
...
-        slice_size, ind_dim_size, inp_stride_bytes, out_stride_bytes, /*allow_neg_indices*/true);
-        return;
-      }
+        // avoid grid overflow in the fast kernel
+        const int64_t vec_chunks = ceil_div(slice_size, alignment);
+        const int64_t blocks_per_slice_upper = ceil_div(vec_chunks, (int64_t)launch_size_nd);
+        const int max_grid_y = at::cuda::getCurrentDeviceProperties()->maxGridSize[1];
+        // if it's an eligible grid we use the fast path, otherwise default to slower path
+        if (blocks_per_slice_upper <= max_grid_y) {
+          if (iter.numel() == 0) return;
```
eqy: Can we just delete this return if it's already at the beginning of the function?

eqy (approved):

aten/src/ATen/native/cuda/IndexKernel.cu
Isalia20: oh yeah, its checked above. Removed

PR #164054, opened 2025-09-28
[2/N] Apply ruff UP035 check in torch files

cyyever:
This is the result of applying the ruff `UP035` check.
`Callable` is imported from `collections.abc` instead of `typing`.
`TypeAlias` is also imported from `typing`.
This PR is the follow-up of #163947.
cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168

### DIFF
--- a/torch/utils/bundled_inputs.py
+++ b/torch/utils/bundled_inputs.py
@@ -1,8 +1,7 @@
 #!/usr/bin/env python3
 # mypy: allow-untyped-defs
 from typing import Any, TypeVar, Optional, NamedTuple, Union
-from collections.abc import Callable
-from collections.abc import Sequence
+from collections.abc import Callable, Sequence
 import textwrap
 import torch
 from torch._C import TupleType, ListType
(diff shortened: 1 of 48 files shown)


torch/utils/bundled_inputs.py
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 # mypy: allow-untyped-defs
-from typing import Any, TypeVar, Optional, NamedTuple, Union, Callable
+from typing import Any, TypeVar, Optional, NamedTuple, Union
+from collections.abc import Callable
 from collections.abc import Sequence
```
ezyang: merge these imports

ezyang (approved):

Skylion007 (approved):

torch/utils/bundled_inputs.py
cyyever: merged, thank you!

PR #164064, opened 2025-09-28
Enable bool dtype for meta lshift and rshift

wodesuck:
lshift and rshift allow one of argument to be bool, but not both. Meta op should follow the same rule.


soulitzer:
Thanks for the PR, could you add a test?

wodesuck:
@soulitzer Test added.

PR #164081, opened 2025-09-28
[dynamo] Special path for cloning of torch dispatch tensors

anijain2305:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #164084
* __->__ #164081

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela

PR #164084, opened 2025-09-28
[export] Skip the check instead of disable

anijain2305:
* __->__ #164084
* #164081

Its unclear why we had disable in the first place. With
install_free_tensors, we are tracing into this hook. A better way would
be to place the tracer without any hook. For now, disable the checking
while dynamo is tracing.


tugsbayasgalan (approved):

anijain2305:
Relanding with https://github.com/pytorch/pytorch/pull/164219

PR #164093, opened 2025-09-29
[MPSHooks] Release pending command encoder

malfet:
* __->__ #164093

Before returning a comand buffer, as subsequent calle are very likely to allocate their own encoder, which results in the following runtime error
```
 tryCoalescingPreviousComputeCommandEncoderWithConfig:nextEncoderClass:]:1090: failed assertion `A command encoder is already encoding to this command buffer'
```

Added regression test to `test_mps_extension`

Please note, that `torch::mps::get_command_buffer()` should be called with dispatch_queue held, both before and after this change, but many implementations skip that

Fixes https://github.com/pytorch/pytorch/issues/163721

### DIFF
--- a/test/cpp_extensions/mps_extension.mm
+++ b/test/cpp_extensions/mps_extension.mm
@@ -13,6 +13,11 @@ kernel void add_arrays(device const float* inA,
 {
     result[index] = inA[index] + inB[index];
 }
+
+kernel void add_one(device float* data,
+                    uint index [[thread_position_in_grid]]) {
+  data[index] += 1.0;
+}
 )MPS_ADD_ARRAYS");
 
 at::Tensor get_cpu_add_output(at::Tensor & cpu_input1, at::Tensor & cpu_input2) {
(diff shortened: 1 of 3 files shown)


atalman (approved):

test/cpp_extensions/mps_extension.mm
```diff
@@ -50,7 +55,31 @@ kernel void add_arrays(device const float* inA,
   return mps_output;
 }
 
+void mps_add_one_new_encoder(const at::Tensor& input) {
+  using namespace at::native::mps;
+  TORCH_CHECK(input.is_mps());
```
Skylion007, line 60: Use TORCH_CHECK_VALUE for these two checks

Skylion007 (approved):

test/cpp_extensions/mps_extension.mm
malfet: It does not matter, as this is an internal test, I'm unsure why they contain those checks to begin with, as they are tests no examples

PR #164095, opened 2025-09-29
Fix silent incorrectness for bmm/baddmm out_dtype overload

PaulZhang12:
* __->__ #164095

Add input checks like meta functions for standard ops in `ATen/native/LinearAlgebra.cpp` for the `out_dtype` variants. Fixes silent incorrectness in https://github.com/pytorch/pytorch/issues/163816

### DIFF
--- a/aten/src/ATen/native/cuda/Blas.cpp
+++ b/aten/src/ATen/native/cuda/Blas.cpp
@@ -15,6 +15,7 @@
 #include <ATen/cuda/tunable/Tunable.h>
 #include <ATen/cuda/tunable/TunableGemm.h>
 #include <ATen/native/Resize.h>
+#include <ATen/native/LinearAlgebra.h>
 #include <c10/util/MaybeOwned.h>
 #include <ATen/native/GroupedMMUtils.h>
 #include <ATen/native/cuda/RowwiseScaledMM.h>
@@ -784,6 +785,8 @@ const Tensor& baddbmm_out_cuda_impl(const Tensor& result, const Tensor& self, co
 }
 
 } // anonymous namespace
+
+#define MM
 
 TORCH_IMPL_FUNC(addmm_out_cuda)(const Tensor& self, const Tensor& mat1, const Tensor& mat2, const Scalar& beta, const Scalar& alpha, const Tensor& result) {
   // NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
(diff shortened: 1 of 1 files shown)


ngimel (commented):
Add a test?

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1811,6 +1811,38 @@ std::optional<c10::ScalarType> out_dtype) {
   return out;
 }
 
+static void baddbmm_bmm_out_dtype_checks(const Tensor& batch1, const Tensor& batch2, const Scalar& beta, const Scalar& alpha, const at::ScalarType out_dtype, bool is_bmm, const std::optional<Tensor>& self_baddbmm = std::nullopt) {
```
malfet, line 1814: Is this really specific to CUDA or could it be shared with CPU implementation as well? (i.e. move it to header or define in `ATen/native/Blas.cpp`

ngimel: `dtype` overload is supported only on cuda, `.default` overload params are checked via structured bindings

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1811,6 +1811,38 @@ std::optional<c10::ScalarType> out_dtype) {
...
+  // ref ATen/native/LinearAlgebra.cpp common_checks_baddbmm_bmm
+  TORCH_CHECK(batch1.dim() == 3, "batch1 must be a 3D tensor");
```
Skylion007, line 1816: Many of these should be TORCH_CHECK_VALUE

PaulZhang12: I am following convention in Blas.cpp, where we only have `TORCH_CHECK`

PaulZhang12:
> Add a test?

@ngimel added tests, let me know if there are any other concerns

test/test_matmul_cuda.py
```diff
@@ -784,6 +784,54 @@ def expand(tensor):
...
+            c = torch.randn(B, M, N, device="cuda", dtype=torch.bfloat16)
+            extra_dim_b = a.clone().unsqueeze(0)
+        else:
+            a = torch.randn(M, K, device="cuda", dtype=torch.bfloat16)
+            mismatch_k_b = torch.randn(K + 1, N, device="cuda", dtype=torch.bfloat16)
+            c = torch.randn(M, N, device="cuda", dtype=torch.bfloat16)
+            extra_dim_b = a.clone().unsqueeze(0)
+
+        # Test mismatch K
+        with self.assertRaises(RuntimeError):
```
ngimel: use `assertRaisesRegex` to make sure you are catching the expected error

ngimel (approved):

ngimel:
Test failure is real, perhaps checks are too strict?

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1811,6 +1811,38 @@ std::optional<c10::ScalarType> out_dtype) {
...
+
+  TORCH_CHECK(out_dtype == batch1.scalar_type() ||
+    (out_dtype == at::ScalarType::Float && (batch1.scalar_type() == at::ScalarType::Half || batch1.scalar_type() == at::ScalarType::BFloat16)),
+    "out_dtype must be the same as input dtype or fp32 for fp16/bf16 inputs");
+
+  if (!is_bmm && self_baddbmm.has_value()) {
+    const auto& self = self_baddbmm.value();
+    TORCH_CHECK(self.dim() == 3, "self must be a 3D tensor");
+    TORCH_CHECK(self.sizes() == output_size, "self must have the same shape as the output");
+    TORCH_CHECK(self.scalar_type() == batch1.scalar_type(), "self and batch1 must have the same dtype");
```
PaulZhang12: Ahh @ngimel we shouldn't have this check here, as we always expect this to be false for out_dtype overload!

PR #164111, opened 2025-09-29
[PP] Let PP split BlockMask into micro-BlockMask

fegin:
* #163185
* #164500
* #162542
* __->__ #164111

BlockMask has batch dimension information. So PP has to split it as well just like all other tensors. All the tensors in BlockMask have the batch dimension, so we can just split it without too many issues. However, `mask_mod` requires the batch index as the input, which the value is going to be changed after the split. So we have to wrap it inside a closure to modify the batch index.


test/distributed/pipelining/test_microbatch.py
```diff
@@ -59,6 +61,82 @@ def test_split_and_merge(self):
...
+                return (seq_idx[b, q_idx] == seq_idx[b, kv_idx]) & (q_idx >= kv_idx)
+
+            return block_causal_mask
+
+        batch = list(range(30)) * 5
+        batch = torch.tensor(batch[: B * Q_LEN], device=device).reshape(B, Q_LEN)
+        q = torch.randn(B, H, Q_LEN, DIM, device=device)
+        k = torch.randn(B, H, KV_LEN, DIM, device=device)
+        v = torch.randn(B, H, KV_LEN, DIM, device=device)
+        # block_mask_fn = torch.compile(create_block_mask)
```
XilunWu: is compiled create_block_mask causing issue?

test/distributed/pipelining/test_microbatch.py
```diff
@@ -59,6 +61,82 @@ def test_split_and_merge(self):
...
+        block_mask_fn = create_block_mask
+        block_mask = block_mask_fn(
+            create_block_causal_mask(batch, 29),
+            B=B,
+            H=H,
+            Q_LEN=Q_LEN,
+            KV_LEN=KV_LEN,
+            device=device,
+        )
+        # flex_fn = torch.compile(flex_attention)
```
XilunWu: ditto

torch/distributed/pipelining/microbatch.py
```diff
@@ -132,112 +209,58 @@ def _shard_dict_of_args(
...
+        else:
             raise ValueError(
-                f"Argument value {arg} did not have the same number of "
-                f"values as as chunk spec {chunk_spec}"
+                f"Unsupported chunk spec: {spec} and value: {v} combination."
+            )
+    num_chunks = min(min(split_sizes), num_chunks)
+
+    flat_split_results = [[] for _ in range(num_chunks)]
+    for v, spec in zip(values, chunk_specs):
```
Skylion007: We are on Python 3.10 now officially, use strict=True kwarg for zip

test/distributed/pipelining/test_microbatch.py
```diff
@@ -59,6 +61,82 @@ def test_split_and_merge(self):
         torch.testing.assert_close(merged_kwargs, kwargs)
         print("Microbatch test passed")
 
+    def test_split_block_mask(self, device):
+        B = 4
+        H = 1
+        Q_LEN = 32
+        KV_LEN = 32
```
XilunWu: this size could be too small for testing, in the result `block_mask` there's only one block of size [128, 128].

test/distributed/pipelining/test_microbatch.py
```diff
@@ -59,6 +61,82 @@ def test_split_and_merge(self):
...
+            args_chunk_spec=None,
+            kwargs_chunk_spec=None,
+        )
+
+        for i in range(B):
+            q_chunk, k_chunk, v_chunk, block_mask_chunk = arg_split[i]
+            out_chunk = flex_fn(q_chunk, k_chunk, v_chunk, block_mask=block_mask_chunk)
+            self.assertEqual(q_chunk.squeeze(0), q[i])
+            self.assertEqual(k_chunk.squeeze(0), k[i])
+            self.assertEqual(v_chunk.squeeze(0), v[i])
```
XilunWu: would only hold true if `batch_size == world_size`.

XilunWu (commented):
reviewed the test. Reviewing the implementation now.

torch/distributed/pipelining/microbatch.py
XilunWu: thanks for the reminder ;-)

torch/distributed/pipelining/microbatch.py
```diff
@@ -115,6 +116,82 @@ class _Replicate:
...
+        ret.append(
+            BlockMask.from_kv_blocks(
+                kv_num_blocks=kv_num_blocks_chunks[chunk_idx],
+                kv_indices=kv_indices_chunks[chunk_idx],
+                full_kv_num_blocks=full_kv_num_blocks_chunks[chunk_idx],
+                full_kv_indices=full_kv_indices_chunks[chunk_idx],
+                BLOCK_SIZE=block_mask.BLOCK_SIZE,
+                mask_mod=create_mask_mod(batch_offset),
+                seq_lengths=block_mask.seq_lengths,
+            )
```
XilunWu, line 173: hey @drisspg , do you happen to know how performant this API call is compared to calling (compiled) `create_block_mask`? 

Chien-chin constructs the new `block_mask` in a different way than the one in `create_cp_block_mask` because the rewriting logic is much simpler than CP, but I still hope to unify these 2 functions (`split_block_mask` and `create_cp_block_mask`) since they're doing similar things.

torch/distributed/pipelining/microbatch.py
```diff
@@ -132,112 +209,58 @@ def _shard_dict_of_args(
...
+    ), f"args_dict.keys() = {list(args_dict.keys())} args_chunk_spec.keys() = {list(args_chunk_spec.keys())}"
+    assert args_chunk_spec is not None  # Should have been set by caller
+
+    values, tree_spec = tree_flatten(args_dict)
+    chunk_specs, _ = tree_flatten(args_chunk_spec)
+    assert len(values) == len(chunk_specs)
+
+    # Fist check and find the actual number of chunks
+    split_sizes = []
+    for v, spec in zip(values, chunk_specs):
```
XilunWu: as Aaron pointed out we should add `strict=True`

XilunWu (commented):
Overall LGTM. Let me know when you need a final review.

torch/distributed/pipelining/microbatch.py
```diff
@@ -132,112 +209,58 @@ def _shard_dict_of_args(
...
+            assert isinstance(spec, TensorChunkSpec)
+            assert spec.split_dim == 0
+            split_sizes.append(v.kv_num_blocks.size(0))
+        else:
             raise ValueError(
-                f"Argument value {arg} did not have the same number of "
-                f"values as as chunk spec {chunk_spec}"
+                f"Unsupported chunk spec: {spec} and value: {v} combination."
+            )
+    num_chunks = min(min(split_sizes), num_chunks)
```
H-Huang: nit: call it `result_num_chunks`?

torch/distributed/pipelining/microbatch.py
```diff
@@ -132,112 +209,58 @@ def _shard_dict_of_args(
...
+    split_sizes = []
+    for v, spec in zip(values, chunk_specs):
+        if spec is _Replicate:
+            split_sizes.append(num_chunks)
+        elif isinstance(v, torch.Tensor):
+            assert isinstance(spec, TensorChunkSpec)
+            split_sizes.append(v.size(spec.split_dim))
+        elif isinstance(v, BlockMask):
+            assert isinstance(spec, TensorChunkSpec)
+            assert spec.split_dim == 0
```
H-Huang: add assert error message

H-Huang (commented):
Overall changes look reasonable to me, will do another pass soon to understand the tests better

torch/distributed/pipelining/microbatch.py
fegin: From @drisspg `from_kv_blocks` is strictly faster.

torch/distributed/pipelining/microbatch.py
```diff
@@ -132,114 +238,58 @@ def _shard_dict_of_args(
...
-                        new_val[slice_indices] = chunk_tensor
-
-                        expanded_chunks.append(new_val)
-
-                        split_dim_idx += chunk_tensor.size(chunk_v.split_dim)
-
-                    sharded_arg_flat.append(expanded_chunks)
-                else:
-                    sharded_arg_flat.append(chunk_tensors)  # type: ignore[arg-type]
+        for _flat_split_result, _v_split in zip(flat_split_results, v_splits):
```
Skylion007: Should this be strict=True or strict=False?

test/distributed/pipelining/test_microbatch.py
```diff
@@ -59,6 +60,120 @@ def test_split_and_merge(self):
...
+                h: torch.Tensor,
+                q_idx: torch.Tensor,
+                kv_idx: torch.Tensor,
+            ):
+                return (seq_idx[b, q_idx] == seq_idx[b, kv_idx]) & (q_idx >= kv_idx)
+
+            return block_causal_mask
+
+        batch = list(range(DOC_LEN)) * 1024
+        batch = torch.tensor(batch[: B * SEQ_LEN], device=device).reshape(B, SEQ_LEN)
```
drisspg: nit: this is a little weird to me, I think at least do comment or  have some assert that doc_len * 1024 > b * seqlen

test/distributed/pipelining/test_microbatch.py
```diff
@@ -59,6 +60,120 @@ def test_split_and_merge(self):
...
+            B=B,
+            H=H,
+            Q_LEN=SEQ_LEN,
+            KV_LEN=SEQ_LEN,
+            device=device,
+        )
+        if device == "cuda":
+            flex_fn = torch.compile(flex_attention)
+        else:
+            # It's unclear why CPU + torch.compile + flex_attention can cause an issue.
```
drisspg, line 108: cpu flex code path in inductor is not as battle tested
and has poor training support AFAIK

test/distributed/pipelining/test_microbatch.py
```diff
@@ -59,6 +60,120 @@ def test_split_and_merge(self):
...
+        k_chunks = []
+        dk_chunks = []
+        v_chunks = []
+        dv_chunks = []
+        block_mask_chunks = []
+        out_chunks = []
+        for i in range(len(arg_split)):
+            q_chunk, k_chunk, v_chunk, block_mask_chunk = arg_split[i]
+            for chunk, chunks in zip(
+                (q_chunk, k_chunk, v_chunk), (q_chunks, k_chunks, v_chunks)
```
drisspg: nit this is a lil hard to understand, maybe q_chunk and q_total

torch/distributed/pipelining/microbatch.py
```diff
@@ -115,6 +116,111 @@ class _Replicate:
     pass
 
 
+def _split_block_mask(
+    block_mask: BlockMask,
+    spec: TensorChunkSpec,
```
drisspg: Is TensorChunkSpec a distributed API already or new?

torch/distributed/pipelining/microbatch.py
```diff
@@ -115,6 +116,111 @@ class _Replicate:
...
+    num_chunks: int,
+) -> list[BlockMask]:
+    """Given a block mask, and a chunking spec, split the block mask.
+
+    Args:
+        block_mask: Block mask to split
+        spec: Chunking spec
+        num_chunks: Number of chunks to split the block mask into
```
drisspg, line 127: maybe add a note somehwhere only batch dims support splitting?

torch/distributed/pipelining/microbatch.py
```diff
@@ -115,6 +116,111 @@ class _Replicate:
...
+    Returns:
+        chunk_block_masks: List of chunked block masks
+    """
+
+    assert block_mask.kv_num_blocks.size(0) >= num_chunks, (
+        "Block mask has fewer batch size than the number of chunks. "
+    )
+
+    kv_num_blocks_chunks = torch.tensor_split(
+        block_mask.kv_num_blocks, num_chunks, spec.split_dim
```
drisspg: needs to be dim 0 right?

torch/distributed/pipelining/microbatch.py
drisspg, line 127: Is num_chunks not embedded in the chunk spec?

torch/distributed/pipelining/microbatch.py
```diff
@@ -115,6 +116,111 @@ class _Replicate:
...
+    for chunk_idx in range(num_chunks):
+
+        def create_mask_mod(idx):
+            def mask_mod_wrapper(b, h, q_idx, kv_idx):
+                b_offset = torch.full_like(b, idx)
+                return block_mask.mask_mod(b + b_offset, h, q_idx, kv_idx)
+
+            return mask_mod_wrapper
+
+        # `from_kv_blocks` is strictly faster than `create_block_mask`.
```
drisspg: nit, I dont think you need to say this,  if you have the final tensors you dont need to go through and apply vmap + mask-mod to build them from scratch. So its more like create_block_mask is a factory and it ends up calling from_kv_blocks which is the the base constructor

torch/distributed/pipelining/microbatch.py
```diff
@@ -115,6 +116,111 @@ class _Replicate:
...
+        if block_mask.full_kv_indices is not None
+        else [None] * num_chunks
+    )
+
+    chunk_block_masks = []
+    batch_offset = 0
+    for chunk_idx in range(num_chunks):
+
+        def create_mask_mod(idx):
+            def mask_mod_wrapper(b, h, q_idx, kv_idx):
```
drisspg: nit: maybe call this a batch offset mask mod

test/distributed/pipelining/test_microbatch.py
H-Huang: can do this instead:

```
total_elements = B * SEQ_LEN
batch = torch.arange(total_elements, device=device) % DOC_LEN
batch = batch.reshape(B, SEQ_LEN)
```

torch/distributed/pipelining/microbatch.py
```diff
@@ -132,114 +238,58 @@ def _shard_dict_of_args(
...
-        chunk_spec = args_chunk_spec[arg_key]
-        assert chunk_spec is not None  # Should have been set by caller
-        chunk_spec_flat, _ = tree_flatten(chunk_spec)
-        if len(flat) != len(chunk_spec_flat):
+    assert args_chunk_spec is not None  # Should have been set by caller
+
+    values, tree_spec = tree_flatten(args_dict)
+    chunk_specs, _ = tree_flatten(args_chunk_spec)
+
+    # Fist check and find the actual number of chunks
```
H-Huang:
```suggestion
    # First check and find the actual number of chunks
```

torch/distributed/pipelining/microbatch.py
H-Huang: distributed API, it specifies which dimension of a tensor to split along when dividing it into chunks for microbatching (in pipeline parallelism).

torch/distributed/pipelining/microbatch.py
```diff
@@ -115,6 +116,111 @@ class _Replicate:
...
+def _split_block_mask(
+    block_mask: BlockMask,
+    spec: TensorChunkSpec,
+    num_chunks: int,
+) -> list[BlockMask]:
+    """Given a block mask, and a chunking spec, split the block mask.
+
+    Args:
+        block_mask: Block mask to split
+        spec: Chunking spec
```
H-Huang: is there a case where we would want to specify on non-0 dimension for block mask? If it's always 0, then maybe passing the spec is not necessary

H-Huang (approved):
Looks good! Thanks for cleaning up `_shard_dict_of_args()`, it is much more understandable now.

torch/distributed/pipelining/microbatch.py
fegin: No, chunk spec only specifies the chunk dim.

torch/distributed/pipelining/microbatch.py
fegin: sound good

torch/distributed/pipelining/microbatch.py
fegin: Right, I checked it beforehand. But maybe just like @H-Huang  mentioned, just remove chunk spec for this function at all.

torch/distributed/pipelining/microbatch.py
fegin: True.

torch/distributed/pipelining/microbatch.py
```diff
@@ -132,114 +234,60 @@ def _shard_dict_of_args(
...
+        v_splits: Sequence[Any] = []
+        if spec is _Replicate:
+            v_splits = [v] * result_num_chunks
+        elif isinstance(v, torch.Tensor):
+            v_splits = _split_tensor(v, spec, result_num_chunks)
+        elif isinstance(v, BlockMask):
+            v_splits = _split_block_mask(v, result_num_chunks)
+        else:
+            raise ValueError(
+                f"Unsupported chunk spec: {spec} and value: {v} combination."
```
carmocca, line 279: This raises for primitive type returns. For example if the model forward returns a boolean for whether to materialize logits

fegin: Sorry, I don't understand the comment. Does this raise cause any issue? @carmocca

carmocca: Well it now fails for primitive (e.g. boolean) return values, when they used to work.

I worked around it by explicitly marking it as a `_Replicate`
```py
        pp_schedule._kwargs_chunk_spec = {
            "a_tensor_return": TensorChunkSpec(0),
            "a_tensor_return": TensorChunkSpec(0),
            "a_boolean_return": _Replicate,  # type: ignore[dict-item]
        }
```

fegin: @carmocca I understand now. I believe my follow-up PR fixes it, https://github.com/pytorch/pytorch/pull/165306. Let me know if you still encounter the error with https://github.com/pytorch/pytorch/pull/165306.

PR #164114, opened 2025-09-29
Remove deprecated torch.nn.qat

cyyever:
This PR removes the deprecated `torch/nn/qat` folder.


jerryzh168:
hmmm,  we can't start deprecating these before we confirmed that it's safe I think, we still have internal users right now using all these things

jerryzh168 (changes requested):
not ready to deprecate yet, can do this until we have sufficient understanding of internal usage and confirm it's safe to remove

PR #164118, opened 2025-09-29
Fix/cuda stream backward 164094

juyterman1000:
Fixes #ISSUE_NUMBER
Let torch.cuda.set_stream() work within the backward method
Re-applied the stream guard after the function call
Keep  the stream enforcement for built-in functions


soulitzer (changes requested):
Not sure what problem this is trying to solve? The test should pass w/o this PR.

PR #164121, opened 2025-09-29
update the baseline data for the operator benchmark

### LINKED ISSUE #164089: Failing Operator Benchmarks
The operator benchmarks workflow is failing. Baseline numbers for improvements needs to be updated in `expected_ci_operator_benchmark_eager_float32_cpu.csv` to fix the issue

https://github.com/pytorch/pytorch/actions/runs/18070844712/job/51421106636

@LifengWang @huydhn

LifengWang:
Fixes #164089 

According to the results of the last five operator benchmark runs, we found that more than 600 models achieved more than a 30% improvement compared to the baseline. Therefore, we will update the operator benchmark baseline data.
We use the average results from the five runs as the new baseline for the five models.

cc @chuanqi129 @jainapurva @huydhn @desertfire 
[operator_benchmark_baseline_data.xlsx](https://github.com/user-attachments/files/22592822/operator_benchmark_baseline_data.xlsx)


huydhn:
I remember seeing a PR from @malfet renaming this file with an `ARCH` prefix https://github.com/pytorch/pytorch/pull/165584 so that we can add `aarch64` baseline too.  Plz do a quick rebase

huydhn (approved):
Stamped! There are some improvements I think, so plz update them before merging https://github.com/pytorch/pytorch/actions/runs/18585975250/job/52993121088#step:27:11391

huydhn:
> Yes, the issue looks unrelated. On a broader note, @huydhn should we have two different workflows for x86, and aarch64, as the POC for the two are different?

No need for that I think, we could set owner at the job level, e.g. https://github.com/pytorch/pytorch/blob/main/.github/workflows/periodic.yml#L127

PR #161155, opened 2025-08-21
[ZENDNN] Add ZenDNN as an optional third-party lib

naveenthangudu:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #164126
* __->__ #161155

----

- Add ZenDNN as a third-party library and link it to ATen
- Add ZenDNN(caffe2::zendnn) as a caffe2 dependency
- Introduce USE_ZENDNN CMake ENV option
    - default OFF
    - user-enableable, valid only on x86_64 hosts
- Define AT_ZENDNN_ENABLED macro to gate C/C++ code
- Add ZenDNN as deperdency to ATen library
- Add ZenDNN as a submodule
- Extend the CMake summary to report ZenDNN build status
- Add USE_ZENDNN  setting and its status to build settings string in __confg__.show()
- Add variable substitutions for bazel build.
- Modify zendnn inductor perf builds with "zen" keyword.
- USE_ZENDNN set to 1 for *zen* build enviroments.

Co-authored-by: Dinesh Mareedu <Dinesh.Mareedu@amd.com>
Co-authored-by: Aakar Dwivedi <aakar.dwivedi@amd.com>
 Co-authored-by: Arijit Mukhopadhyay<arijit.mukhopadhyay@amd.com>
Change-Id: I8fdd93e11384d3550557f163faf34a7b8f18a6a9

[RFC](https://github.com/pytorch/pytorch/issues/150296)

Issue #150296, opened 2025-03-31
[RFC] zentorch Integration


naveenthangudu:
# zentorch Integration

## Table of Contents
- [1. Authors](#1-authors)
- [2. Summary](#2-summary)
- [3. Highlights](#3-highlights)
- [4. Motivation](#4-motivation)
  - [4.1. Benchmarking Configuration](#41-benchmarking-configuration)
  - [4.2. Single Core Performance Summary](#42-single-core-performance-summary)
  - [4.3. Multi Core Performance Summary](#43-multi-core-performance-summary)
- [5. Proposal](#5-proposal)
  - [5.1. User Flow](#51-user-flow)
  - [5.2. Implementation](#52-implementation)
    - [5.2.1. ZenDNN Library](#521-zendnn-library)
    - [5.2.2. Graph Optimizations](#522-graph-optimizations)
    - [5.2.3. Plugging into Torch 2.x Flow](#523-plugging-into-torch-2x-flow)
- [6. Plan](#6-plan)
  - [6.1. Phase 1 - Optimizations of Recommender Systems for BF16 and FP32](#61-phase-1---optimizations-of-recommender-systems-for-bf16-and-fp32)
  - [6.2. Phase 2 - Migration to New ZenDNN Library API and Architecture](#62-phase-2---migration-to-new-zendnn-library-api-and-architecture)
  - [6.3. Phase 3 - Optimizations of Recommender Systems for INT8 Linear and woq_embedding_bag](#63-phase-3---optimizations-of-recommender-systems-for-int8-linear-and-woq-embedding_bag)
  - [6.4. Phase 4 - Optimizations of NLPs and Generative LLM Workloads](#64-phase-4---optimizations-of-nlps-and-generative-llm-workloads)

## 1. Authors
@naveenthangudu, @sudarshan, @charan, @amukho, @Dinesh-Mareedu and @avinash

## 2. Summary
This document proposes an approach for integrating the ZenDNN library and Zentorch optimizations into PyTorch. This integration will enable inference optimizations for deep learning workloads on AMD EPYC™ CPUs. It will provide AMD EPYC™-focused optimizations in both the AOT Inductor (Torch Export) and the Inductor (Torch Compile) paths.

The integration will be carried out in three phases for Recommender Systems targeting [Phase 1](#61-phase-1---optimizations-of-recommender-systems-for-bf16-and-fp32) in PT 2.8 and others in 2.9:
[...]

PR #164126, opened 2025-09-29
[ZENDNN] Add bmm support for eager mode

naveenthangudu:
* __->__ #164126
* #161155

-- Integrate zendnn_baddbmm in eager mode path
-- Enable support only on AMD CPUs with AVX512 ISA
-- Enable bmm for bf16 data type only
-- Set TORCHINDUCTOR_ENABLE_ZENDNN=1 to turn on the support

Change-Id: Ieca212075baa75646f251c16478ce1fad3594f01
Co-authored-by: Dinesh Mareedu <dinesh.mareedu@amd.com>

[RFC](https://github.com/pytorch/pytorch/issues/150296)


malfet (changes requested):
In principle it looks good to me, but I want to better understand why this integration needs to be done thru PyTorch rather than thru oneDNN, for example the way ARM does it with ACL

PR #164129, opened 2025-09-29
[CodeClean] Replace std::runtime_error with TORCH_CHECK

fffrog:
* #164130
* __->__ #164129

As the title stated.

**Changes**:
- torch/csrc/Module.cpp
- torch/csrc/utils.cpp
- torch/csrc/stable
- torch/lib/libshm


albanD (approved):
Thanks!

torch/csrc/stable/stableivalue_conversions.h
```diff
@@ -118,7 +119,8 @@ struct FromImpl<ScalarType> {
       case ScalarType::UInt64:
         return from(aoti_torch_dtype_uint64());
       default:
-        throw std::runtime_error(
+        TORCH_CHECK(
```
mikaylagawarecki, line 122: For future reference, we should **not** include files in c10/aten or use TORCH_CHECK in  torch/headeronly or torch/csrc/stable (instead we should use STD_TORCH_CHECK)

I'm opening a PR to fix this and will add better testing to prevent this

PR #164132, opened 2025-09-29
Improved support for autotuning in wrapper_fxir

nandesuka:
Summary:
- correct dtype propagation
- allow more more options to be passed to compiler

Test Plan: in follow up change

Differential Revision: D83367909

### DIFF
--- a/torch/_inductor/codegen/wrapper_fxir.py
+++ b/torch/_inductor/codegen/wrapper_fxir.py
@@ -952,6 +952,7 @@ class FxConverter:
                 return torch.empty_strided(
                     to_size_hint(fake.shape),
                     to_size_hint(fake.stride()),
+                    dtype=fake.dtype,
                     device=device,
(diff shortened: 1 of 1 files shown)


jansel (approved):

torch/_inductor/codegen/wrapper_fxir.py
```diff
@@ -952,6 +952,7 @@ def to_size_hint(arg: Any) -> Any:
                 return torch.empty_strided(
                     to_size_hint(fake.shape),
                     to_size_hint(fake.stride()),
+                    dtype=fake.dtype,
```
blaine-rister, line 955: Whoops

blaine-rister (commented):
The core logic LGTM, but do we have any OSS tests for this functionality? There is an existing autotuning test which mocks `tuner.run`. That might be a good way to check the dtype.

PR #164141, opened 2025-09-29
Add _scaled_mm_v2 API

slayton58:
* #164142
* __->__ #164141

Summary:

* Add new scaled-MM API to future-proof / clean-up existing code.
* Scaling is explicitly described rather than infer
* Swizzling of scaled must now be defined (vs. inferred)
* Adds API support for multi-level scaling
* Refactor dispatch logic to make it easier to add new implementations

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:
Signed-off-by: Simon Layton <simonlaytonmeta.com>


aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1659,9 +1659,895 @@ _scaled_mm_cuda(const Tensor& mat_a, const Tensor& mat_b,
...
+  NVFP4_NVFP4_SINGLE_SCALE = 8,
+};
+
+/**
+ * Convert passed int (enum) from python back into a
+ * strictly-typed enum
+ */
+template <class EnumType>
+std::vector<EnumType> convert_int_to_enum(ArrayRef<long>& v) {
+  std::vector<EnumType> converted{};
```
Skylion007: Call converted.reserve() to reserve the proper size

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1659,9 +1659,895 @@ _scaled_mm_cuda(const Tensor& mat_a, const Tensor& mat_b,
...
+  if (recipe_b[0] != ScalingType::BlockWise1x32) return false;
+  if (scales_b[0].scalar_type() != ScalarType::Float8_e8m0fnu) return false;
+
+  return true;
+}
+
+using acceptance_fn = std::function<bool(c10::ScalarType, std::vector<ScalingType>&, ArrayRef<Tensor>&, c10::ScalarType, std::vector<ScalingType>&, ArrayRef<Tensor>&)>;
+using namespace std::placeholders;
+
+std::vector<std::tuple<std::string, acceptance_fn, ScaledGemmImplementation>> scale_kernel_dispatch = {
```
Skylion007: Do we support dynamically adding things here? Why not an std::array?

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1659,9 +1659,895 @@ _scaled_mm_cuda(const Tensor& mat_a, const Tensor& mat_b,
...
+Tensor&
+_scaled_rowwise_rowwise(
+          const Tensor& mat_a, const Tensor& mat_b,
+          const Tensor& scale_a, const Tensor& scale_b,
+          const std::optional<Tensor>& bias,
+          const c10::ScalarType out_dtype,
+          bool use_fast_accum,
+          Tensor& out) {
+  // Restrictions:
+  // A, B are FP8, scales are fp32, shape M/N for A/B
```
Skylion007, line 2104: Some of these should be TORCH_CHECK_VALUE or raise NotImplementedError

aten/src/ATen/native/cuda/Blas.cpp
slayton58: Done

aten/src/ATen/native/cuda/Blas.cpp
slayton58: Done

aten/src/ATen/native/cuda/Blas.cpp
slayton58: What's the guidance for `TORCH_CHECK` vs. `TORCH_CHECK_VALUE` vs. raising `NotImplementedError`? From a quick look, only `TORCH_CHECK` is used in this particular file to this point

Skylion007: @slayton58 TORCH_CHECK raise RuntimeError, TORCH_CHECK_VALUE raises ValueError, I don't know if we have a pre-exisitng macro for raising NotImplementedError. RuntimeError's are under-specified, it could be a transient error, a user error, etc. ValueError denotes it's a user error and the args to the specified function are invalid. NotImplementedError means something could be supported in upstream PyTorch, but we haven't implemented it yet.

Most of this file should probably be changed to ValueError at some point, but that can be a subsequent PR

danielvegamyhre:
@slayton58 I like the explicit instead of inferred scaling type.

Question, why do the scaling recipe strings only repeat for fp8 types but not for MX types ? e.g., 

- `ScaledGemmImplementation::TENSORWISE_TENSORWISE` => "tensorwise_tensorwise"  
- `ScaledGemmImplementation::MXFP8_MXFP8` => `"mxfp8"`

aten/src/ATen/native/native_functions.yaml
```diff
@@ -7155,13 +7155,24 @@
...
   variants: function
   dispatch:
     CPU: _scaled_mm_out_cpu
     CUDA: _scaled_mm_out_cuda
   tags: needs_exact_strides
 
+- func: _scaled_mm_v2(Tensor self, Tensor mat2, Tensor[] scale_a, int[] recipe_a, int[] swizzle_a, Tensor[] scale_b, int[] recipe_b, int[] swizzle_b, Tensor? bias, ScalarType? out_dtype, int[] contraction_dim=[], bool use_fast_accum=False) -> Tensor
+  variants: function
+  dispatch:
+    CUDA: _scaled_mm_cuda_v2
```
drisspg, line 7171: We should think about how this will look for other backends, e.g. do we ever think that CPU will support some subset of recipes? I dont think that changes anything just wanted to note

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1565,7 +1565,7 @@ _scaled_mm_out_cuda(const Tensor& mat1, const Tensor& mat2,
         args.scale_result_ptr,
         args.result_ld,
         out_dtype_,
-        use_fast_accum);
+        false /*use_fast_accum*/);
```
drisspg, line 1569: intentional?

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1679,9 +1679,895 @@ _scaled_mm_cuda(const Tensor& mat_a, const Tensor& mat_b,
...
+  NVFP4_NVFP4 = 7,
+  NVFP4_NVFP4_SINGLE_SCALE = 8,
+};
+
+/**
+ * Convert passed int (enum) from python back into a
+ * strictly-typed enum
+ */
+template <class EnumType>
+std::vector<EnumType> convert_int_to_enum(ArrayRef<long>& v) {
```
drisspg: Might be worth pybinding the type, I think I did this w/ SDPBackend to make sure things stay consistent

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1679,9 +1679,895 @@ _scaled_mm_cuda(const Tensor& mat_a, const Tensor& mat_b,
...
+  auto swizzle_b_enum = convert_int_to_enum<SwizzleType>(swizzle_b);
+
+  // at this point we can start working out what we want to be doing
+  // Try to do as few steps as possible.
+  // NOTE: support is deliberately sparse, can explicitly enumerate all combinations allowed.
+  // Do this via a list of defined (name, acceptance, concrete_impl) tuples.
+  bool found_impl = false;
+  ScaledGemmImplementation gemm_impl = ScaledGemmImplementation::NONE;
+
+  for (const auto& fn_entry : scale_kernel_dispatch) {
```
drisspg, line 2493: should it be a scan down finding the first matching ? or should it just be a direct map from recipe pair to impl?

aten/src/ATen/native/cuda/Blas.cpp
slayton58: I am 100% willing to do this, I don't like the approach this second - an example would be amazing. I think I've conflated no enums through native_functions with no enums shared at all between python/C++ in pytorch..

aten/src/ATen/native/native_functions.yaml
slayton58: CPU can/will certainly support some subset (or superset) of recipes - I'm honestly not sure what is / will be supported though. A CPU backend wouldn't be hard to add (or a shim to dispatch to the existing `_scaled_mm_cpu` backend.

aten/src/ATen/native/cuda/Blas.cpp
slayton58: The big problem I have with a map -> impl, is that different implementations can (and imo really should) have different signatures - nvfp4xnvfp4 needs global scales, swizzles passed in addition for instance. I think this might be work-around-able with `std::bind` to present a unified API, but that's also messy..

aten/src/ATen/native/cuda/Blas.cpp
slayton58: Nope, fixed.

aten/src/ATen/native/cuda/Blas.cpp
drisspg: Spoke offline but for provenance: https://github.com/pytorch/pytorch/blob/11f5f656867089dac1fa1e64e34c9966578fbddd/torch/csrc/Module.cpp#L2392

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1679,9 +1680,895 @@ _scaled_mm_cuda(const Tensor& mat_a, const Tensor& mat_b,
...
+  if (scales_b[0].scalar_type() != ScalarType::Float8_e4m3fn) return false;
+
+  return true;
+}
+
+/**
+ * Both inputs must be fp8
+ * A, B must only have 1 scale each, A: {Blockwise_1x128 (float), B: {Blockwise_128x128 (float)
+ */
+bool check_deepseek_recipe(ScalingType expected_recipe_a,
```
drisspg, line 1845: nit lets change the name

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1679,9 +1680,895 @@ _scaled_mm_cuda(const Tensor& mat_a, const Tensor& mat_b,
...
+  if (recipe_b[0] != ScalingType::BlockWise1x32) return false;
+  if (scales_b[0].scalar_type() != ScalarType::Float8_e8m0fnu) return false;
+
+  return true;
+}
+
+using acceptance_fn = std::function<bool(c10::ScalarType, std::vector<ScalingType>&, ArrayRef<Tensor>&, c10::ScalarType, std::vector<ScalingType>&, ArrayRef<Tensor>&)>;
+using namespace std::placeholders;
+
+std::array<std::tuple<std::string, acceptance_fn, ScaledGemmImplementation>, 8> scale_kernel_dispatch = {{
```
drisspg, line 1904: I think I wrote a utility called https://github.com/pytorch/pytorch/blob/5f15110499a9256f64cce600aedd96f7f02446fa/c10/util/Array.h#L15

which helps w/ seg faults as we expand this list, dont ask my how I know 😂

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1679,9 +1680,895 @@ _scaled_mm_cuda(const Tensor& mat_a, const Tensor& mat_b,
...
+  { "block_128x128_1x128", std::bind(check_deepseek_recipe, ScalingType::BlockWise128x128, ScalingType::BlockWise1x128, _1, _2, _3, _4, _5, _6),
+    ScaledGemmImplementation::BLOCK_128x128_1x128},
+  { "block_1x128_1x128", std::bind(check_deepseek_recipe, ScalingType::BlockWise1x128, ScalingType::BlockWise1x128, _1, _2, _3, _4, _5, _6),
+    ScaledGemmImplementation::BLOCK_1x128_1x128},
+  { "nvfp4", check_nvfp4_recipe, ScaledGemmImplementation::NVFP4_NVFP4},
+  { "nvfp4_single_scale", check_nvfp4_recipe_single_scale, ScaledGemmImplementation::NVFP4_NVFP4_SINGLE_SCALE },
+  { "mxfp8", check_mxfp8_recipe, ScaledGemmImplementation::MXFP8_MXFP8}}};
+
+Tensor&
+_cutlass_scaled_gemm(
```
drisspg, line 1918: Nit: name is weird since it goes to alot more than just cutlass

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1679,9 +1680,895 @@ _scaled_mm_cuda(const Tensor& mat_a, const Tensor& mat_b,
...
+          const bool use_fast_accum,
+          Tensor& out) {
+  cublasCommonArgs args(mat1, mat2, out, scale_a, scale_b, std::nullopt, scaling_choice_a, scaling_choice_b);
+  const auto out_dtype_ = args.result->scalar_type();
+  TORCH_CHECK(args.transa == 't' && args.transb == 'n', "Only multiplication of row-major and column-major matrices is supported by cuBLASLt");
+
+#ifdef USE_ROCM
+  auto tuning_ctx = at::cuda::tunable::getTuningContext();
+  if (tuning_ctx->IsTunableOpEnabled()) {
+#define TUNABLE_DISPATCH(BLASOP_A, BLASOP_B)                            \
```
drisspg, line 1932: I would love the AMD path to be a lil more modular. E.g. one if def that is like USE_ROCM -> go to rocm path, otherwise go to cuda path. It makes it alot easier to grok

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1679,9 +1680,895 @@ _scaled_mm_cuda(const Tensor& mat_a, const Tensor& mat_b,
...
+  TORCH_CHECK_VALUE(scale_b.numel() == mat_b.size(1) && scale_b.scalar_type() == kFloat, "scale_b must have ", mat_b.size(1), " Float elements, got ", scale_b.numel())
+
+  TORCH_CHECK_VALUE(scale_a.stride(1) == 1, "expected scale_a.stride(1) to be 1, but got ", scale_a.stride(1));
+  TORCH_CHECK_VALUE(scale_b.stride(1) == 1, "expected scale_b.stride(1) to be 1, but got ", scale_b.stride(1));
+
+  auto scaling_choice_a = ScalingType::RowWise;
+  auto scaling_choice_b = ScalingType::RowWise;
+  //
+  // NVIDIA's cuBLAS only started supporting row-wise scaling in version 12.9,
+  // and only for compute capability 9.0+. In other cases we use CUTLASS.
```
drisspg, line 2118: Is rowwise also not supported on sm100 like the 128x recipes?

aten/src/ATen/native/cuda/Blas.cpp
drisspg: Yeah that makes sense, but to confirm every accpet_fn matches against the enum right? so there is no ambiguity?

aten/src/ATen/native/cuda/Blas.cpp
slayton58: From deepseek? What would you prefer?

drisspg (approved):
Comment soup but LGTM

aten/src/ATen/native/cuda/Blas.cpp
slayton58: 9.0+ = >= 9.0, so yes, it should be on Blackwell too (I think my CC nomenclature here is different to yours :D )

aten/src/ATen/native/cuda/Blas.cpp
drisspg: maybe just the 128 by recipes similar to the  other ones

aten/src/ATen/native/cuda/Blas.cpp
drisspg: Ohh I just meant, that didn't the 128x recipes ONLY work on sm90 and not blackwell? Not sure if cublas did the same thing for the rowwise recipes

PR #164142, opened 2025-09-29
Add scaled_mm python API, test

slayton58:
* __->__ #164142
* #164141

Summary:

* Add `torch.nn.functional.scaled_mm` as an abstraction around the C++
  methods
* Wraps `torch._scaled_mm_v2` API by default, but user can force use of
  the older `torch._scaled_mm` interface.
* Scaled MM tests now run on the new API

Test Plan:

`pytest test/test_scaled_matmul_cuda.py`

Reviewers:

Subscribers:

Tasks:

Tags:
Signed-off-by: Simon Layton <simonlaytonmeta.com>


docs/source/quantization.rst
```diff
@@ -3,6 +3,8 @@
 Quantization
 ============
 
+.. automodule:: torch.quantization.scaled_mm
```
vkuzo: nit: i'd probably not add stuff to `torch.quantization`, as it's deprecated and the name is ambiguous.  Can we have a more specific name for scaled_mm, or just keep it together with the non-scaled gemms in terms of naming / docs?

slayton58: I'm fine with this going wherever - It's in `torch.quantization` because there wasn't a clearly better place for it to go - it's not a functional version of a `torch.nn` op, so `torch.nn.functional` didn't seem like a good place

slayton58: Post some offline discussion, moved to `torch.nn.functional.scaled_mm`

docs/source/nn.functional.rst
```diff
@@ -218,3 +218,11 @@ DataParallel functions (multi-GPU, distributed)
     :nosignatures:
 
     torch.nn.parallel.data_parallel
+
+Low-Precision functions
+-----------------------
+.. autosummary::
+    :toctree: generated
```
drisspg, line 225: we need to document the enums and their meaning, still working down this review so might already be in here

test/test_fx.py
```diff
@@ -4657,6 +4657,7 @@ def tearDown(self):
         "adaptive_avg_pool3d": LEN_ERROR,
         "adaptive_max_pool2d_with_indices": LEN_ERROR,
         "adaptive_max_pool3d_with_indices": LEN_ERROR,
+        "scaled_mm": LEN_ERROR,
```
drisspg: OOC what does this signify?

torch/nn/functional.py
```diff
@@ -6509,3 +6509,121 @@ def multi_head_attention_forward(
...
+    swizzle_a: _SwizzleType | list[_SwizzleType] | None = None,
+    swizzle_b: _SwizzleType | list[_SwizzleType] | None = None,
+    bias: Optional[Tensor] = None,
+    output_dtype: Optional[torch.dtype] = torch.bfloat16,
+    contraction_dim: Optional[list[int]] = None,
+    use_fast_accum: bool = False,
+    **kwargs: _Any,
+) -> Tensor:
+    r"""
+    scaled_mm(mat_a, mat_b, scale_a, scale_recipe_a, scale_b, scale_recipe_b, swizzle_a, swizzle_b, bias, output_dtype,
```
drisspg, line 6538: This is needed for the doc builds right?

test/test_fx.py
slayton58: `len(...)` is non-traceable by default (errors out during the test), and `scaled_mm` uses `len` for some list processing - this prevents the test from running.

torch/nn/functional.py
```diff
@@ -6509,3 +6509,121 @@ def multi_head_attention_forward(
...
+        bias: optional bias term to be added to the output
+        output_dtype: dtype used for the output tensor
+        contraction_dim: describe which dimensions are :math:`K` in the matmul.
+        use_fast_accum: enable/disable tensor-core fast accumulation (Hopper-GPUs only)
+    """
+    use_deprecated_api = kwargs.pop("use_deprecated_scaled_mm", False)
+    if len(kwargs) > 0:
+        raise RuntimeError("kwargs contains unexpected entries, ", kwargs.keys())
+
+    if use_deprecated_api:
```
drisspg: OOC why this?

slayton58: The `deprecated_api` path? It allows for a back-compat path to isolate any differences from the implementations. I found it incredibly useful for debugging, I guess it could be removed if desired..

test/test_fx.py
drisspg: Does this mean that this new op won't be compilable ?

slayton58: I guess so - I see 2 immediate ways to fix this:
* remove the error-checking for the deprecated fallback path and just pass the first scale etc. from the passed list, and rely on erroring out in C++ with an invalid scaling recipe
* Remove the deprecated fallback path entirely

docs/source/nn.functional.rst
slayton58: Done in an updated commit

drisspg (approved):
Looks good in a follow up we should have some more robust composability testing, the immediate things that come to mind is writing meta_registrations.py entry adding some compile tests and then making sure we (for now) fallback in inductor and then ultimately rewire the lowerings.

cc @eellison  is the best way to get most of this testing still through common_methods_invocations.py?

PR #164144, opened 2025-09-29
Fix truediv numerics between eager and compile

PaulZhang12:
* __->__ #164144

Addresses numeric differences between eager and compile in https://github.com/pytorch/pytorch/issues/141753


test/inductor/test_cuda_repro.py
```diff
@@ -2507,6 +2507,26 @@ def forward(q_proj, k_proj, attn_mask):
...
+        @torch.compile
+        def compiled_divide(x, y):
+            return x / y
+            
+        for y_dtype in [torch.float16, torch.bfloat16, torch.float32, torch.float64]:
+            for x_dtype in [torch.float16, torch.bfloat16, torch.float32, torch.float64]:
+                y = torch.tensor([y], dtype=y_dtype, device="cuda")
+                x = torch.tensor([x], dtype=x_dtype, device="cuda")
+                
+                torch._dynamo.reset()
```
eellison, line 2602: nit: would it be more accurate to `self.assertEqual(eager_div, compiled_div, atol=0.0, rtol=0.0)` ?

torch/_inductor/codegen/triton.py
```diff
@@ -1053,7 +1053,10 @@ def abs(x):
 
     @staticmethod
     def truediv(x, y):
-        out = f"({x} / {y})"
+        if x.dtype == torch.float32 and y.dtype == torch.float32:
```
eellison: Just for my own understanding - do you know where in pytorch cuda `div_rn` is dispatched to ? maybe this is just default `x / y` in cuda ?

eellison (dismissed):

test/inductor/test_cuda_repro.py
PaulZhang12: Here we compare the exact decimal values themselves from Python's decimal module: https://docs.python.org/3/library/decimal.html, so I don't think this is needed?

torch/_inductor/codegen/triton.py
PaulZhang12: I am not sure.. I can do a followup and check

PaulZhang12: https://www.internalfb.com/code/fbsource/[5291bf823ebee20df47c5b01645639a1f93da5ab]/fbcode/caffe2/caffe2/operators/elementwise_div_op.h?lines=21 I believe it's here? Lots of different layers to go through

ngimel: it's here https://github.com/pytorch/pytorch/blob/e4cdc31227de4ec20d8d30e03c2a396a19a9ec39/aten/src/ATen/native/cuda/BinaryInternal.h#L21, plain `x/y`, but you can't say which ptx instruction nvcc chooses to lower it to (called from https://github.com/pytorch/pytorch/blob/57a49018b10fb9d1c290d82a4fa367539185f509/aten/src/ATen/native/cuda/BinaryDivTrueKernel.cu#L52)

PaulZhang12:
@ngimel the test case in this pr exposes the numerical differences, even with converting both to a tensor and doing tensor / tensor in Inductor. Am I misunderstanding your point?

jansel (dismissed):

ngimel (dismissed):

torch/_inductor/codegen/triton.py
```diff
@@ -1062,7 +1062,13 @@ def abs(x):
...
     @staticmethod
     def truediv(x, y):
-        out = f"({x} / {y})"
+        x_dtype = getattr(x, "dtype", None)
+        y_dtype = getattr(y, "dtype", None)
+
+        if x_dtype == torch.float32 and y_dtype == torch.float32:
+            out = f"triton.language.div_rn({x}, {y})"
+        else:
+            out = f"({x} / {y})"
```
ngimel, line 1086: can you leave a comment that `x /y` in triton is lowered to `dif.full` which is approximate, instead of `div_rn`

bobrenjc93 (approved):

test/test_torchfuzz_repros.py
```diff
@@ -20,157 +20,6 @@ def setUp(self):
         torch._dynamo.config.capture_dynamic_output_shape_ops = True
         torch._inductor.config.emulate_precision_casts = True
 
-    @pytest.mark.xfail(reason="Issue #164428")
```
malfet, line 23: Why are you removing the entire test? Just remove xfail and it should pass, shouldn't it?

PaulZhang12: This test now passes due to this change, no longer a need as stated by @bobrenjc93

PR #164155, opened 2025-09-29
Add attention benchmarking numbers to pytorch operator microbenchmarks

jainapurva:
This pull request introduces a standardized YAML-based configuration system for transformer attention benchmarks, making it easier to run and manage comprehensive performance tests. It adds example configs, and a wrapper script to convert YAML configs into CLI arguments for the benchmark runner. 

#### Next Steps:
CI Enablement: This change would further lead to running the attention ops in CI for regression tracking.

#### Developer flow: (Run locally)
`python score_mod.py --config configs/config_test.yaml`


benchmarks/transformer/configs/config_comprehensive.yaml
```diff
@@ -0,0 +1,29 @@
+# Comprehensive benchmark configuration for PyTorch transformer benchmarks
+# Usage: python score_mod.py --config config_comprehensive.yaml
```
drisspg: lets just keep the basicyaml config, we can add there

benchmarks/transformer/score_mod.py
```diff
@@ -1077,59 +1135,365 @@ def generate_experiment_configs(
...
+        python score_mod.py --print-config > my_config.json
+
+        # Use a config file
+        python score_mod.py --config my_config.json
+
+        # Override config with CLI args
+        python score_mod.py --config my_config.json -dtype float16 --max-autotune
+
+        # Pure CLI usage
+        python score_mod.py -b 4 8 -s 1024 2048 -mods causal alibi --backend efficient
```
jbschlosser: should we have an example with a yaml config to indicate that it's supported?

benchmarks/transformer/score_mod.py
```diff
@@ -1077,59 +1135,365 @@ def generate_experiment_configs(
...
+                    dynamic=dynamic,
                 ),
             )
         )
 
-    print_results(results, args.save_path)
+        experiment_count += 1
+        # Periodic memory cleanup every 10 experiments
+        if experiment_count % 10 == 0:
+            cleanup_memory()
```
jbschlosser, line 1490: what's the reason behind manually doing memory cleanup? I expect this to be handled automatically, is this not the case?

benchmarks/transformer/score_mod.py
```diff
@@ -1199,8 +1563,113 @@ def heads_input_type(s):
...
+                        config[key] = value.strip('"\'')
+                    else:
+                        # Try to parse as number
+                        try:
+                            config[key] = int(value)
+                        except ValueError:
+                            try:
+                                config[key] = float(value)
+                            except ValueError:
+                                config[key] = value
```
jbschlosser: nit: I suggest factoring out YAML config parsing into a separate helper for readability

benchmarks/transformer/score_mod.py
```diff
@@ -1199,8 +1563,113 @@ def heads_input_type(s):
...
+                    if key_normalized == "nh" and isinstance(value, list):
+                        setattr(args, key_normalized, [heads_input_type(h) if isinstance(h, str) else h for h in value])
+                    else:
+                        setattr(args, key_normalized, value)
+
+    # Convert dtype string to torch dtype (only if it's still a string)
+    if isinstance(args.dtype, str):
+        args.dtype = getattr(torch, args.dtype)
+
+    main(
```
jbschlosser: is not possible to just do `main(**vars(args))` here instead of spelling these all out?

benchmarks/transformer/score_mod.py
```diff
@@ -1199,8 +1563,113 @@ def heads_input_type(s):
...
+                arg_value = getattr(args, key_normalized)
+                default_value = parser.get_default(key_normalized)
+
+                # If the arg value equals default, use config file value
+                if arg_value == default_value:
+                    # Handle special case for 'nh' which needs conversion
+                    if key_normalized == "nh" and isinstance(value, list):
+                        setattr(args, key_normalized, [heads_input_type(h) if isinstance(h, str) else h for h in value])
+                    else:
+                        setattr(args, key_normalized, value)
```
jbschlosser: nit: I think this logic for overriding config file settings with explicitly-set CLI args could be simpler.. here's an example of one other way to handle this: https://stackoverflow.com/a/58609044

benchmarks/transformer/score_mod.py
```diff
@@ -1077,59 +1135,365 @@ def generate_experiment_configs(
...
+    if isinstance(dtype, str):
+        dtype = getattr(torch, dtype)
+
+    # Parse head configurations (if not already parsed)
+    nh_parsed = []
+    for h in nh:
+        if isinstance(h, tuple):
+            nh_parsed.append(h)
+        else:
+            nh_parsed.append(heads_input_type(h))
```
jbschlosser: since head config is done here, does it need to be done during arg parsing?

benchmarks/transformer/score_mod.py
```diff
@@ -1077,59 +1135,365 @@ def generate_experiment_configs(
...
+            model: ModelInfo
+            metric: MetricInfo
+
+        # Add record for forward latency
+        record_fwd_latency = BenchmarkRecord(
+            benchmark=BenchmarkInfo(
+                name=benchmark_name,
+                mode=mode,
+                dtype=dtype,
+                extra_info={
```
jbschlosser: this feels kinda verbose for what's going on - is there any way to reduce the copy / paste duplication throughout? e.g. we can at least precalculate `extra_info` once and use that for any records we need

benchmarks/transformer/score_mod.py
```diff
@@ -821,13 +882,9 @@ def generate_FA_callable(
...
-        try:
-            from flash_attn import flash_attn_func, flash_attn_varlen_func
-        except ImportError:
-            print(
-                "Flash attention 2 is not installed. Please install it to run fav2 backend. "
-            )
-            raise
+        # Use PyTorch's built-in Flash Attention via SDPA
+        flash_attn_func = None  # We'll define this below using SDPA
+        flash_attn_varlen_func = None
```
jbschlosser: sorry I don't understand, why not define these here?

benchmarks/transformer/score_mod.py
```diff
@@ -33,6 +35,61 @@
...
 
+def cleanup_memory():
+    """Aggressively free GPU memory"""
+    torch.cuda.empty_cache()
+    gc.collect()
+    if torch.cuda.is_available():
+        torch.cuda.synchronize()
+
+
+def safe_backend(backend_name=None):
```
jbschlosser: decorator looks solid, nice idea :)

jbschlosser (commented):
Nice work! I've got a few stylistic comments but nothing major

benchmarks/transformer/score_mod.py
jainapurva: In the config_basic.yaml, there's a cmd to run it using score_mod, I'll add the same sample command here.

benchmarks/transformer/score_mod.py
jainapurva: Thanks, this makes code much simpler.

benchmarks/transformer/score_mod.py
jainapurva: Thanks, nice catch. Missed removing it after testing

benchmarks/transformer/score_mod.py
jainapurva: I'm just doing it to be more thorough.

benchmarks/transformer/score_mod.py
jainapurva: Moved to new file config_utils.py https://github.com/pytorch/pytorch/blob/2bcd29cc480b2c1cc39836d2e36ec25e8f3f3237/benchmarks/transformer/config_utils.py

benchmarks/transformer/score_mod.py
jbschlosser: alrighty, I don't have a strong opinion against it so I won't hold up the review, but I do generally prefer introducing logic that isn't strictly necessary to keep complexity lower, up to you :)

jbschlosser (approved):
awesome!

PR #164161, opened 2025-09-29
[Fix] Completely remove stride normalization on DLPack Tensor

### LINKED ISSUE #163274: [DLPack] Further restrict the normalization condition for DLPack Export

Kathryn-cat:
A followup on PR #163282
Fixes #163274

### DIFF
--- a/aten/src/ATen/DLConvertor.cpp
+++ b/aten/src/ATen/DLConvertor.cpp
@@ -389,37 +389,16 @@ void fillVersion<DLManagedTensorVersioned>(
 // constructed out of ATen tensor
 template <class T>
 T* toDLPackImpl(const Tensor& src) {
-  auto view = src;
-
-  // Detect whether there is need to normalize the strides
-  // Background: gh-83069
-  //
-  // However, normalizing strides can come at a high-cost
-  // to slow down toDLPack conversion 3x, so we
-  // only normalize if needed.
-  //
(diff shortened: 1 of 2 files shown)


ngimel (approved):

Kathryn-cat:
@eqy please help me merging this PR

eqy (approved):

eqy (approved):

PR #164162, opened 2025-09-29
Exporting aten.sdpa with cuda under fake mode on a cuda-less machine

yiming0416:
Summary:
As titled.

sdpa will select backend based on hardware check, and it fails when exporting with cuda under fake mode on a cuda-less machine.

We guard `at::cuda::is_available()` check before `at::cuda::getCurrentDeviceProperties()` and give warnings.

Test Plan: buck2 run mode/dev-nosan caffe2/test:test_export -- -r nn_functional_scaled_dot_product_attention

Differential Revision: D83496154


SherlockNoMad (approved):
thanks

PR #164165, opened 2025-09-29
[CMake] Fix `USE_FBGEMM_GENAI` option

malfet:
* __->__ #164165

----

- `cmake_dependent_option` condition should be `USE_ROCM OR (USE_CUDA AND NOT MSVC)` (similar to the one for flash attention)
- Default settings should be user overridable, i.e. even if one builds for SM_10, they should be able to pass `USE_FBGEMM_GENAI=0` and skip the build


cmake/Summary.cmake
```diff
@@ -138,6 +138,7 @@ function(caffe2_print_configuration_summary)
   message(STATUS "  USE_EIGEN_FOR_BLAS    : ${CAFFE2_USE_EIGEN_FOR_BLAS}")
   message(STATUS "  USE_EIGEN_FOR_SPARSE  : ${USE_EIGEN_SPARSE}")
   message(STATUS "  USE_FBGEMM            : ${USE_FBGEMM}")
+  message(STATUS "  USE_FBGEMM_GENAI      : ${USE_FBGEMM_GENAI}")
```
Skylion007, line 141: Why remove the printing here?

CMakeLists.txt
```diff
@@ -888,23 +888,28 @@ cmake_dependent_option(
   "(USE_CUDA AND NOT MSVC) OR USE_ROCM"
   OFF)
 
+
+IF(USE_ROCM AND "gfx942" IN_LIST PYTORCH_ROCM_ARCH)
+  message(WARNING "Setting USE_FBGEMM_GENAI for gfx942 to ON by default, doing ROCM build")
+  set(USE_FBGEMM_GENAI_DEFAULT ON)
+elseif(USE_CUDA AND "$ENV{TORCH_CUDA_ARCH_LIST}" MATCHES "10.0" AND CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL 12.8 AND NOT WIN32)
```
Skylion007, line 895: Only for datacenter Blackwell? Do we not need it for 11.0 or 12.0 arches too?

Skylion007 (approved):
Looks good, but I'm confused by we are removing useful debugging print info from the Summary CMake

CMakeLists.txt
malfet: This check is insane to be frank, and yes, it should probably be enabled for more GPU architectures, but at least it makes the logic user-overridable, so one can build for SM_12 if they choose to

Skylion007: Yeah, this is going to be a footgun for consumer blackwell soon though

cmake/Summary.cmake
malfet: I'm adding the print statement, don't I?

CMakeLists.txt
malfet: I'm not sure if it'll work on consumer Blackwells, as it compiles for sm_10a, which is not directly runnable/translatable for sm_12

danielvegamyhre: The mxfp8 grouped gemm in fbgemm must be built with sm100a and cuda 12.8+, which is why this check exists the way it does. Furthermore, torch nightly builds only set TORCH_CUDA_ARCH_LIST=10.0, never 10.0a due to lack of portability, so we have to [add sm100a to the build targets for fbgemm](https://github.com/pytorch/pytorch/blob/349c960970f4e29eff0d37a9b3c1ca5ed86a121a/aten/src/ATen/CMakeLists.txt#L270).

I am not sure if sm120 supports the tcgen* PTX instructions needed for the mxfp8 grouped gemm kernel?

PR #164167, opened 2025-09-29
fix copy_ for scalar in inductor

parsshar-RH:
Fixes #158437 

### Summary

- TorchInductor was not properly handling scalar copy operations `(tensor.copy_(scalar_value))`
- Ensured scalar sources are converted to appropriate tensor representations with correct dtype and device

### Impact

- Enables compilation of models using ` tensor.copy_(scalar) `patterns
- module: inductor

### DIFF
--- a/torch/_inductor/lowering.py
+++ b/torch/_inductor/lowering.py
@@ -2955,6 +2955,8 @@ make_fallback(aten.repeat_interleave.Tensor, override_decomp=True)
 # For example, fp16.copy_(fp32) should **not** promote the first input's dtype.
 @register_lowering(aten.copy, type_promotion_kind=None)
 def copy(self, src, non_blocking=False):
+    if not isinstance(src, (TensorBox, ir.IRNode)):
+        src = tensor(src, dtype=self.get_dtype(), device=self.get_device())
     x = src
     if self.get_device() != src.get_device():
         x = to_device(x, self.get_device())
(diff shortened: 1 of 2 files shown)


torch/_inductor/lowering.py
```diff
@@ -2955,6 +2955,8 @@ def is_aligned(x):
 # For example, fp16.copy_(fp32) should **not** promote the first input's dtype.
 @register_lowering(aten.copy, type_promotion_kind=None)
 def copy(self, src, non_blocking=False):
+    if not isinstance(src, (TensorBox, ir.IRNode)):
```
shunting314: nit: TensorBox is a subclass of ir.IRNode.

Doing 
```
if not isinstance(src, ir.IRNode):
```

is enough

shunting314 (approved):

torch/_inductor/lowering.py
parsshar-RH: @shunting314 

Thanks for the review
I have added commit with the changes as suggested 

Thanks

shunting314:
@parsshar-RH  I would suggest repro the failure on your dev environment to debug more.

My guess is maybe the tensor is so small that some unrolling is happening. I would try if the test pass with a larger tensor size.

parsshar-RH:
Hi @shunting314 

I have fixed the tests
Please review

Thanks

shunting314:
LG. let's wait for the CI result before merging.

parsshar-RH:
Hi @shunting314 

Looks like all the CI checks are passed

parsshar-RH:
@shunting314 

Looks like some unrelated test is failing.
Please have a look

Thanks

parsshar-RH:
@shunting314 

Looks like there is some CI issue

PR #126573, opened 2024-05-17
aot_autograd: copy metadata from fw to bw nodes

vkuzo:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #126698
* __->__ #126573

Summary:

Uses the `seq_nr` field (introduced to aot_autograd nodes in
https://github.com/pytorch/pytorch/pull/103129) to map the aot_autograd
fx bw nodes to the corresponding fw nodes, and copy the metadata over.

I am trusting the `seq_nr` mapping in the linked PR here. I did
some validation with a toy LLaMa 3 8b training run and the mapping seemed
correct.

I am also trusting that the forward is single threaded, since `seq_nr` is thread local.  If this isn't always true, we'll need to also plumb `thread_id` through the same machinery which is populating `seq_nr`.

I'd like to use this data in a future PR to make inductor kernels easily
attributable to the nn.Module path in modeling land, to make it easier
to do performance debugging.

Test Plan:

```
// 1. unit test
python test/dynamo/test_aot_autograd.py -k test_aot_sequence_nr

// 2. manual test
// run LLaMa 3 8B fw + bw with torch.compile, print out the inductor graphs
// seen in `torch/_inductor/utils.py::get_kernel_metadata`, they seemed
// right to me.
```

Reviewers:

Subscribers:

Tasks:

Tags:

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames

PR #103129, opened 2023-06-07
Add sequence_nr to aot_autograd to map forward ops to their corresponding backward ops

alexmsettle:
Fixes #102375 

Sequence_nr increments in the forward pass and decrements in the backward pass.  Backward ops with the same sequence_nr as a forward op represent the backward implementation for the op.  The long term goal is to make this information available to the profiler so users can observe which ops are fused by the inductor openai triton kernels.

Added a test for this feature **test/dynamo/test_aot_autograd.py::AotAutogradFallbackTests::test_aot_sequence_nr**.  The test case uses **aot_export_module()** to create a joint fwd/bwd fx graph.  Then it walks all the nodes in fx graph using fx_graph.graph.nodes.   The seq_nr of each node is recorded in node.meta.  During the fwd pass the seq_nr increments and it decrements during the bwd pass.  This allows the user to map forward ops to their corresponding bwd ops which is useful for performance analysis.  

Expected output from the test case.

 SeqNr|OrigAten|SrcFn
0|aten.convolution.default|l__self___conv1
0|aten.add.Tensor|l__self___bn1
1|aten._native_batch_norm_legit_functional.default|l__self___bn1
2|aten.relu.default|l__self___relu1
3|aten.add.Tensor|add
4|aten.view.default|flatten
5|aten.t.default|l__self___fc1
6|aten.unsqueeze.default|l__self___fc1
7|aten.mm.default|l__self___fc1
8|aten.squeeze.dim|l__self___fc1
9|aten.add.Tensor|l__self___fc1
10|aten.sub.Tensor|l__self___loss_fn
11|aten.abs.default|l__self___loss_fn
12|aten.mean.default|l__self___loss_fn
12|aten.ones_like.default|
12|aten.expand.default|
12|aten.div.Scalar|
11|aten.sgn.default|
11|aten.mul.Tensor|
8|aten.unsqueeze.default|
7|aten.t.default|
7|aten.mm.default|
7|aten.t.default|
7|aten.t.default|
7|aten.mm.default|
6|aten.squeeze.dim|
5|aten.t.default|
4|aten.view.default|
2|aten.threshold_backward.default|
1|aten.native_batch_norm_backward.default|
0|aten.convolution_backward.default|
0|aten.add.Tensor|

      

[...]

PR #164174, opened 2025-09-29
Support propagating custom meta field to backward graph nodes

yushangdi:
# Propagate custom meta data to backward

Support propagating the user annotation tags to backward graph, by extending the `copy_fwd_metadata_to_bw_nodes` utils (recommended by @xmfan , thanks!). 

Example annotation API (added in https://github.com/pytorch/pytorch/pull/163673):

```
class M(torch.nn.Module):
    def forward(self, x):
            with fx_traceback.annotate({"fdsp_bucket": 0}):
                x = x + 1
            x = x - 2
            with fx_traceback.annotate({"cuda_stream": 2, "fsdp_bucket": 1}):
                x = x * 2
        x = x / 3
        return x
```

Assumptions (some inherited from https://github.com/pytorch/pytorch/pull/126573):

- I am trusting the seq_nr mapping introduced to aot_autograd nodes in https://github.com/pytorch/pytorch/pull/103129
- I am also trusting that the forward is single threaded, since seq_nr is thread local.  If this isn't always true, we'll need to also plumb thread_id through the same machinery which is populating seq_nr.
- **(This is changed in this PR!) I assume all backward graph nodes has "is_backward" for 'partitioner_tag', and all other nodes are forward graph nodes**.  If we don't run export before `aot_export_join_with_descriptors`, then none of the nodes has "nn_module_stack" in node meta. If we do run export first, then we don't need this change. 
- I copy "custom" node meta from forward to backward graph nodes. 

Question:
- Is it a good idea to copy all "custom" node meta? Or should we create a dedicated key in custom node meta to be copied? @SherlockNoMad 
- Do we expect people to run export before using `aot_export_join_with_descriptors`?
- Can we assume the following for graph produced by `aot_export_join_with_descriptors`? "all backward graph nodes has "is_backward" for 'partitioner_tag', and all other nodes are forward graph nodes". Maybe this is a question for @ezyang 

```
python test/functorch/test_aot_joint_with_descriptors.py -k test_preserve_
python test/export/test_export.py -k preserve_anno
python test/distributed/tensor/test_dtensor_export.py 
```

### DIFF
--- a/test/distributed/tensor/test_dtensor_export.py
+++ b/test/distributed/tensor/test_dtensor_export.py
@@ -5,6 +5,7 @@ import unittest
 
 import torch
 import torch.distributed as dist
+import torch.fx.traceback as fx_traceback
 from torch._dynamo.functional_export import _dynamo_graph_capture_for_export
 from torch._functorch.aot_autograd import aot_export_joint_with_descriptors
 from torch._functorch.partitioners import min_cut_rematerialization_partition
@@ -35,6 +36,18 @@ class SimpleModel(torch.nn.Module):
 
     def forward(self, input):
         return self.mlp_1(self.mlp_0(input))
+
+
+class SimpleModelAnnotated(torch.nn.Module):
+    def __init__(self, device):
+        super().__init__()
+        self.mlp_0 = MLPModule(device)
+        self.mlp_1 = MLPModule(device)
+
+    def forward(self, input):
+        with fx_traceback.annotate({"pp_stage": 0}):
+            x = self.mlp_0(input)
+        return self.mlp_1(x)
 
 
 def strict_export_and_aot_export_joint_with_descriptors(model, inputs):
@@ -90,7 +103,7 @@ class DTensorExportTest(TestCase):
         )
         self.device_type = "cuda"
 
-    def _run_test(self, export_fn):
+    def _run_test(self, export_fn, test_annotation=False):
         dp_degree = 2
         tp_degree = self.world_size // dp_degree
 
@@ -101,7 +114,11 @@ class DTensorExportTest(TestCase):
             mesh_dim_names=["dp", "tp"],
         )
 
-        model = SimpleModel(self.device_type)
+        model = None
+        if test_annotation:
+            model = SimpleModelAnnotated(self.device_type)
+        else:
+            model = SimpleModel(self.device_type)
         parallelize_plan = {
             "mlp_0.net1": ColwiseParallel(),
             "mlp_0.net2": RowwiseParallel(),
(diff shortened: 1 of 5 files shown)


SherlockNoMad:
Interesting Approach! 

- I don't know how seq_nr is generated, is this reliable? Also, it seems to have the granularity of torch-level ops? (which is fine btw)
- I like the idea of having the mapping between forward and backward node by explicit IDs. This saves propagating stringfied metadata through autograd engine, which cumbersome and might have some overhead. 
- I am fined with copying all "custom" node meta to bw node, as the starting point. If ppl need more fine grain control, we can change to selective copying. 
- Do we expect people to run export before using aot_export_join_with_descriptors?
As mentioned, this is undecided. We are leaning towards fullgraph_capture + aot_export_join_with_descriptors. 
But I do encourage you to proactively adding the provenance tracking info to "aot_export_join_with_descriptors alone workflow", if bandwidth allows.

xmfan:
Note that with this approach, AC nodes should have the fw tags

xmfan (approved):
since this is load bearing, please add a few tests covering DTensor style usage

test/functorch/test_aot_joint_with_descriptors.py
```diff
@@ -792,18 +792,10 @@ def forward(self, x):
...
-                            torch.ops.prims.broadcast_in_dim.default,
-                            torch.ops.prims.add.default,
-                        )
-                        # TODO: add annotation to backward graph nodes
-                        and node.meta.get("partitioner_tag") != "is_backward"
-                    ):
+                    if node.target not in (
+                        torch.ops.prims.sub.default,
+                        torch.ops.aten.sub.default,
+                    ) and node.op not in ("placeholder", "output"):
```
xmfan: why did this PR change the number of nodes that should have the annotation? i'd suggest checking the forward and backward individually if the opset don't match

test/functorch/test_aot_joint_with_descriptors.py
SherlockNoMad: let's use decompositions={} to match real case

I don't think prims ops should show up in joint graph.

test/functorch/test_aot_joint_with_descriptors.py
SherlockNoMad: export followed by aot_export_joint_with_descriptors would produce a joint graph with missing backward region. 

this might be the cause.

test/functorch/test_aot_joint_with_descriptors.py
SherlockNoMad: change to _dynamo_graph_capture_for_export 


see https://github.com/pytorch/pytorch/pull/163609 for reference.

SherlockNoMad (approved):
lgtm. 

Need more test cases, you can do it in next diff. 

In particular:

- Use DTensor + parallelized module, see https://github.com/pytorch/pytorch/pull/163609 for example. 

- Verify collectives invoked from DTensor redistribution have the tags as expected

test/functorch/test_aot_joint_with_descriptors.py
yushangdi: hmm, unfortunately `_dynamo_graph_capture_for_export` doesn't work with `fx_traceback.annotate` (no annotation in metadata produced). I'm trying to figure out why...

SherlockNoMad: likely the ctx manager is not setup `with fx_traceback.preserve_node_meta()`?

yushangdi: this is fixed now.

test/functorch/test_aot_joint_with_descriptors.py
yushangdi: before this PR, there is no annotation on the backward nodes, which is not good. Now we have annotation for the nodes in backward

test/distributed/tensor/test_dtensor_export.py
```diff
@@ -131,18 +148,129 @@ def _run_test(self, export_fn):
...
-        "export_fn",
+        "export_fn, test_annotation",
         [
-            graph_capture_and_aot_export_joint_with_descriptors,
-            aot_export_joint_with_descriptors_alone,
+            (graph_capture_and_aot_export_joint_with_descriptors, False),
+            (aot_export_joint_with_descriptors_alone, True),
         ],
     )
     def test_export_parallelize_module_with_dtensor_input(
```
SherlockNoMad, line 268: can you test annotation in a separate test?

SherlockNoMad (approved):

test/distributed/tensor/test_dtensor_export.py
yushangdi: sure, added a new `test_annotate_aot_export_joint_with_descriptors_alone` now.

PR #164177, opened 2025-09-29
Add initial suppressions for pyrefly

maggiemoss:
Adds suppressions to pyrefly will typecheck clean: https://github.com/pytorch/pytorch/issues/163283

Test plan:
`python3 scripts/lintrunner.py`
`pyrefly check`

--- 

Pyrefly check before: https://gist.github.com/maggiemoss/3a0aa0b6cdda0e449cd5743d5fce2c60
After:

```
 INFO Checking project configured at `/Users/maggiemoss/python_projects/pytorch/pyrefly.toml`
 INFO 0 errors (1,063 ignored)
```

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @mcarilli @ptrblck @leslie-fang-intel


Lucaskabela (changes requested):
Hi Maggie thanks for this change! I am really interested to see how this project can improve our type coverage and subsequent correctness/developer speed :) 

One general note on the PR is I would suggest updating the testplan - my understanding is these changes are to make `pyrefly check` pass - can you update the testplan to show this command and the output before/after? Doesn't have to be the full output but the final lines with errors before/after would be helpful, thanks!

test/test_bundled_inputs.py
```diff
@@ -58,18 +58,24 @@ def forward(self, arg):
...
+        self.assertTrue(
+            loaded(*inflated[0]) is inflated[0][0]
+        )  # pyrefly: ignore  # missing-attribute
 
         for idx, inp in enumerate(inflated):
-            self.assertIsInstance(inp, tuple)
+            self.assertIsInstance(inp, tuple)  # pyrefly: ignore  # missing-attribute
             self.assertEqual(len(inp), 1)
-            self.assertIsInstance(inp[0], torch.Tensor)
+            self.assertIsInstance(
```
Lucaskabela: Any thoughts as to why pyrefly is failing to find these testcase asserts but is able to manage `assertEqual`?

maggiemoss: I will dig in and put a fix into Pyrefly if needed, thanks for flagging :)

torch/nested/_internal/ops.py
```diff
@@ -29,11 +29,14 @@ def _outer_to_inner_dim(ndim, dim, ragged_dim, canonicalize=False):
...
         dim = canonicalize_dims(ndim, dim)
 
-    assert dim >= 0 and dim < ndim
+    assert dim >= 0 and dim < ndim  # pyrefly: ignore  # unsupported-operation
 
     # Map dim=0 (AKA batch dim) -> packed dim i.e. outer ragged dim - 1.
     # For other dims, subtract 1 to convert to inner space.
-    return ragged_dim - 1 if dim == 0 else dim - 1
+    return (
+        # pyrefly: ignore  # unsupported-operation
```
Lucaskabela, line 37: I noticed sometimes we wrap like this (two ignores wrapping a statement).  Is this related to the line breaks?

test/test_bundled_inputs.py
Lucaskabela: Sounds good - that doesn't need to be done in this PR necessarily but as long as we have some issue linked here we can follow up on

torch/nested/_internal/ops.py
maggiemoss: Nice catch! Removed the extra ones :)

Lucaskabela (approved):
Thanks for the changes - lets link an issue to investigate from Pyrefly side, as well as fix the linter error then we should be good to land

Lucaskabela:
Thanks for the work! Please cc me for future reviews - I am definitely interested in championing this work for the repo :)

PR #164183, opened 2025-09-29
[CI] Push `viable/strict/${time}` tags

malfet:
* __->__ #164183

Every time viable strict is updated


.github/workflows/update-viablestrict.yml
```diff
@@ -48,4 +48,6 @@ jobs:
             echo "{\"sha\": \"${LATEST_SHA}\", \"repository\":\"pytorch/pytorch\", \"timestamp\": ${TIME}}" > "/tmp/${LATEST_SHA}.json"
             pip install awscli==1.29.40
             aws s3 cp "/tmp/${LATEST_SHA}.json" "s3://ossci-raw-job-status/stable_pushes/pytorch/pytorch/${LATEST_SHA}.json"
+            # Push new viable/strict tag
+            git push origin "${LATEST_SHA}:refs/tags/viable/strict/${TIME}"
```
jathu, line 53: nit: perhaps we should create another step instead of coupling it with this aws step. Suppose someone decides to remove this step in the future

seemethere (approved):

PR #164184, opened 2025-09-29
Fix TestExportOpInfo

SherlockNoMad:
Fixes https://github.com/pytorch/pytorch/issues/163699

* __->__ #164184


yiming0416 (approved):
Just curious, what bug would `cuda:0` cause (as commented in the original code)?

tugsbayasgalan (approved):

SherlockNoMad:
> Just curious, what bug would `cuda:0` cause (as commented in the original code)?

No, I was just trying to a bit more adventurous to use cuda:1, hoping to expose some bugs. 
but I didn't notice any. 

One thing I do notice is that, you can only do up to cuda:255, IIRC, code device idx is uint8.

PR #164196, opened 2025-09-30
[export] Better state_dict and constant dedup in torch.export.save

yiming0416:
Summary:

Previously, weight deduplication was done by simply grouping tensors with their untyped storage and saving the first tensor in the group.

A more rigorous approach would be to find a complete tensor that covers the storage and store that tensor. This is particularly important for GPU weights because when saving to raw bytes, we move the weight to CPU first, and if the weight being saved is not a complete one, it will lose the storage information during the copy to CPU.

In this diff, we reuse code in `_package_weights.py` for better weights and constants deduplication in `torch.export.save`.

Test Plan: buck2 run mode/dev-nosan caffe2/test:test_export -- -r test_weight_sharing_gpu

Differential Revision: D83523690

### DIFF
--- a/test/export/test_serialize.py
+++ b/test/export/test_serialize.py
@@ -14,7 +14,8 @@ from collections import namedtuple
 from pathlib import Path
 from typing import NamedTuple
 
-from torch.testing._internal.inductor_utils import HAS_GPU
+from torch.testing._internal.inductor_utils import GPU_TYPE, HAS_GPU
+from torch.testing._internal.triton_utils import requires_gpu
 
 
 if HAS_GPU:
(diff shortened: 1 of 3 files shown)


test/export/test_serialize.py
```diff
@@ -924,6 +925,27 @@ def forward(self, x):
...
+                return self.linear(x) + self.c1 + self.c2
+
+        m = M().to(GPU_TYPE)
+        sample_inputs = (torch.randn(2, 4, device=GPU_TYPE),)
+        ep = torch.export.export(m, sample_inputs)
+        buffer = io.BytesIO()
+        save(ep, buffer)
+        buffer.seek(0)
+        loaded_ep = load(buffer)
+        self.assertEqual(m(*sample_inputs), loaded_ep.module()(*sample_inputs))
```
yushangdi, line 976: nit: I think this test would pass even if c1 and c2 are saved as two separate tensors? Might be good to also check that c1 is a slice of c2 in the loaded ep.

torch/export/pt2_archive/_package_weights.py
```diff
@@ -1,39 +1,58 @@
...
 from torch.utils._ordered_set import OrderedSet
 
 
 def _end_ptr(tensor: torch.Tensor) -> int:
     if tensor.nelement():
-        stop = tensor.view(-1)[-1].data_ptr() + tensor.element_size()
+        if tensor.is_contiguous():
+            return tensor.view(-1)[-1].data_ptr() + tensor.element_size()
+        else:
+            stop = tensor.reshape(-1)[-1].data_ptr() + tensor.element_size()
```
yushangdi: I think `reshape` may copy the tensor when it's not possible to return a view. In that case, the returned _end_ptr is not really the real _end_ptr of the original tensor.

yushangdi: IIRC, `self.end` is only used to determine if a tensor is complete. If a tensor is not contiguous, I think we can just say it's not complete without looking at the end_ptr?

test/export/test_serialize.py
yiming0416: @yushangdi  good point. Checking "c1" and "c2" have the same untyped_storage after loading

torch/export/pt2_archive/_package_weights.py
yiming0416: @yushangdi  good catch! now always return False in `is_complete()` when the tensor is not contiguous

torch/export/pt2_archive/_package.py
```diff
@@ -361,58 +362,104 @@ def _get_raw_tensor_bytes(value: torch.Tensor) -> bytes:
...
-    idx = archive_writer.count_prefix(os.path.join(WEIGHTS_DIR, WEIGHT_FILENAME_PREFIX))
+    pickled_weights: list[tuple[str, torch.Tensor]] = []
+    raw_weights: dict[str, tuple[torch.Tensor, TensorProperties]] = {}
+
+    # Categorize weights
     for weight_fqn, weight_tensor in exported_program.state_dict.items():
         assert isinstance(weight_tensor, torch.Tensor), (
             "only torch.Tensor is allowed in state_dict"
         )
+        if _should_use_pickle(weight_tensor):
```
angelayi, line 452: can you share the code between saving weights and constants?

yiming0416: @angelayi yep updated!

angelayi (approved):

PR #164198, opened 2025-09-30
distributed/serialization: support zero sized tensors

d4l3k:
Fixes
```
[4] ValueError: both buffer length (0) and count (-1) must not be 0
```

Test plan:

```
pytest test/distributed/test_serialization.py
```

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @pragupta @msaroufim @dcci


amirafzali (approved):
thanks for the quick fix!

PR #164205, opened 2025-09-30
DebugMode add ignore_compile_internals

### LINKED ISSUE #164143: DebugMode silently disables torch.compile
### 🐛 Describe the bug

Self explanatory. 

```
V0929 08:06:25.203000 1331468 /data/users/ezyang/b/pytorch/torch/_dynamo/convert_frame.py:1968] skipping: forward (reason: non-infra torch dispatch mode present, this is not supported today in torch.compile, file: /data/users/ezyang/b/autoparallel/autoparallel/_testing/models/llama3.py)
```

At minimum we should error when this occurs, but it would be better for DebugMode to work when you have a backend like aot_eager.

cc @chauhang @penguinwu @SherlockNoMad @anijain2305 

### Versions

main

SherlockNoMad:
Fixes #164143

* __->__ #164205

### DIFF
--- a/test/distributed/tensor/debug/test_debug_mode.py
+++ b/test/distributed/tensor/debug/test_debug_mode.py
@@ -258,6 +258,16 @@ class TestDTensorDebugMode(TestCase):
         # Verify that cond operations are captured in debug mode
         self.assertIn("torch.ops.higher_order.cond", debug_mode.debug_string())
 
+    def test_compile(self):
+        @torch.compile
+        def f(x):
+            return x.sin().cos()
+
+        x = torch.randn(8)
+        with DebugMode() as debug_mode:
+            f(x)
+        self.assertEqual(len(debug_mode.debug_string()), 0)
+
 
 instantiate_parametrized_tests(TestDTensorDebugMode)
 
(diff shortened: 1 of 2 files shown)


test/distributed/tensor/debug/test_debug_mode.py
```diff
@@ -258,6 +258,16 @@ def test_debug_mode_higher_order_cond(self):
...
 
+    def test_compile(self):
+        @torch.compile
+        def f(x):
+            return x.sin().cos()
+
+        x = torch.randn(8)
+        with DebugMode() as debug_mode:
+            f(x)
+        self.assertEqual(len(debug_mode.debug_string()), 0)
```
albanD, line 272: Feels suspicious that this is empty, but maybe it's the cpp codegen that doesn't generate anything visible to us.

albanD (approved):
Ok incremental improvement, we should investigate more though

test/distributed/tensor/debug/test_debug_mode.py
ezyang: Let's do a test with aot_eager, which should have a sin/cos call.

PR #164212, opened 2025-09-30
[Inductor] Enable Custom op Autotune Decompositions and Parameter Tuning

tianrengao:
This PR introduces CustomOp autotuning. It allows user to provide a CustomOpConfig:
(1) to register (optional) multiple decomposition implementations for custom operations and 
(2) to register parameter tuning knobs and values they want to tune for the decompositions
so that inductor automatically select the best-performing variant through Inductor's autotune benchmarking.

Example:
```python
 register_custom_op_autotuning(
            custom_op=my_attention_op,
            configs=[
                CustomOpConfig(attention_impl, head_dim=32, method='chunked'),
                CustomOpConfig(attention_impl, head_dim=64, method='tiled'),
                CustomOpConfig(head_dim=128), # no decompositions
            ],
            input_gen_fns={
                "query": lambda fake: torch.randn_like(fake, device='cuda'),
                "key": lambda fake: torch.randn_like(fake, device='cuda'),
                "value": lambda fake: torch.randn_like(fake, device='cuda'),
            }
    )
```

**CustomOpConfig**: Each CustomOpConfig defines exactly one autotuning variant with specific parameter values and optional decomposition implementation with PyTorch aten ops. Users can register their own tuning knobs and optional decomposition functions for the same custom operation. The system automatically benchmarks all variants to select the best performing. If no decomposition is provided in the config, the CustomOp's default implementation will be used.

**Custom Input Generation**: Users can provide custom input generators via an optional `input_gen_fns` to control how synthetic inputs are created during benchmarking. This enables more realistic performance testing by generating inputs that match expected data distributions and characteristics for each tensor argument.

**More Examples with autotune logs:**:
1. Allow user to register customOp decompositions with tuning parameters for autotuning. Example usage:
```python
from torch._inductor.kernel.custom_op import CustomOpConfig, register_custom_op_autotuning

def decompose_k_implementation(a: torch.Tensor, b: torch.Tensor, k_splits: int = 4) -> torch.Tensor:
    """Matrix multiply with k-way decomposition."""
         # Implementation...with k_splits

@torch.library.custom_op("my_lib::decompose_k", mutates_args=())
def test_decompose_k_op(
        a: torch.Tensor, b: torch.Tensor, k_splits: int
    ) -> torch.Tensor:
        return decompose_k_implementation(a, b, k_splits)

# Register autotuning with different k_splits values
register_custom_op_autotuning(
    custom_op=test_decompose_k_op,
    configs=[
        CustomOpConfig(decompose_k_implementation, k_splits=2),
        CustomOpConfig(decompose_k_implementation, k_splits=32),
        CustomOpConfig(decompose_k_implementation, k_splits=64), 
        CustomOpConfig(k_splits=128), # can make decomposition optional, then use default impl test_decompose_k_op
        CustomOpConfig(k_splits=256)
    ],
    input_gen_fns={
        "a": lambda fake: torch.randn_like(fake, device='cuda') * 0.1,
        "b": lambda fake: torch.randn_like(fake, device='cuda') * 0.1,
    }
)
```

Example result: 
```
{"num_choices": 6, "num_triton_choices": 0, "best_kernel": "test_decompose_k_autotuned_fallback_default", "best_time": 0.09980800002813339}
AUTOTUNE test_decompose_k_autotuned(256x65536, 65536x1024)
strides: [65536, 1], [1024, 1]
dtypes: torch.float16, torch.float16
  test_decompose_k_autotuned_fallback_default 0.0998 ms 100.0% 
  test_decompose_k_autotuned_decompose_k_implementation_k_splits_2_0 0.1096 ms 91.0% CustomOp decompose_k_implementation_k_splits_2
  test_decompose_k_autotuned_decompose_k_implementation_k_splits_32_1 0.1277 ms 78.2% CustomOp decompose_k_implementation_k_splits_32
  test_decompose_k_autotuned_decompose_k_implementation_k_splits_64_2 0.1454 ms 68.6% CustomOp decompose_k_implementation_k_splits_64
  test_decompose_k_autotuned_decompose_k_implementation_k_splits_128_3 0.1536 ms 65.0% CustomOp decompose_k_implementation_k_splits_128
  test_decompose_k_autotuned_decompose_k_implementation_k_splits_256_4 0.2084 ms 47.9% CustomOp decompose_k_implementation_k_splits_256
```

2. Allow user to tune parameter knob by passing the parameter and values in the CustomOpConfig. 
**Example**
```python
def mlp_variants(input_tensor, gate_weight, up_weight, down_weight, method):
    """MLP implementation with different computational approaches."""
    if method == 0:
        # Standard separate matmuls
        # ... implementation
    elif method == 1:
        # Batched approach with torch.mm
        # ... implementation
    elif method == 2:
        # Fused weights approach
        # ... implementation

@torch.library.custom_op("my_lib::mlp_op", mutates_args=())
        def mlp_op(
            input_tensor: torch.Tensor,
            gate_weight: torch.Tensor,
            up_weight: torch.Tensor,
            down_weight: torch.Tensor,
            method: int,
        ) -> torch.Tensor:
            return mlp_variants(
                input_tensor, gate_weight, up_weight, down_weight, method=method
            )

register_custom_op_autotuning(
    custom_op=mlp_op,
    configs=[
        CustomOpConfig(method=0),
        CustomOpConfig(method=1),
        CustomOpConfig(method=2),
        # method=0 is the default fallback in the original op
    ],
    input_gen_fns={
        "input_tensor": lambda fake: torch.randn_like(fake, device='cuda') * 0.1,
        "gate_weight": lambda fake: torch.randn_like(fake, device='cuda') * 0.05,
        # ... other input generators
    }
)

```

Example result: 
```
AUTOTUNE test_mlp_autotuned(4x32x512, 512x1024, 512x1024, 1024x256)
  test_mlp_autotuned_mlp_variants_method_2 0.0181 ms 100.0% CustomOp mlp_variants_method_2
  test_mlp_autotuned_mlp_variants_method_1 0.0185 ms 97.8% CustomOp mlp_variants_method_1
  test_mlp_autotuned_mlp_default_fallback_method_0 0.0198 ms 91.4% CustomOp fallback
```

### Test Suite (`test/inductor/test_custom_op_autotune.py`)

*   **RMSNorm autotuning**: Tests different RMSNorm implementations with dynamic input shapes
*   **MLP autotuning**: Tests different MLP decomposition and tuning "method" parameter
*   **DecomposeK**: Tests different k_splits values for matrix multiplication decomposition with k dim split
*   **Multi-parameter tuning**: Tests configs with multiple tuning parameters (scale_mode, chunk_size)

### Next Step:
- Enable Max-autotune with user passed in max-autotune config. https://github.com/pytorch/pytorch/pull/165526/files 
- Support inline epilogue fusion for selected best customop decomposition with surrounding elementwise ops. https://github.com/pytorch/pytorch/pull/165952/files 
- Support customop autotune considering fusion with multiTemplateBuffer. WIP


torch/_inductor/kernel/custom_op.py
```diff
@@ -0,0 +1,194 @@
...
+from torch._inductor.ir import Buffer, FixedLayout, Layout, TensorBox
+from torch._inductor.lowering import register_lowering
+from torch._inductor.select_algorithm import autotune_select_algorithm
+from torch.fx.experimental.proxy_tensor import make_fx
+
+
+__all__ = ["autotune_custom_op", "register_custom_op_lowering"]
+
+
+class CustomOpTemplate(SubgraphTemplate):
```
eellison: What about this needs to be different than Subgraph Template ?

torch/_inductor/kernel/custom_op.py
```diff
@@ -0,0 +1,194 @@
...
+            # Extract layout from actual output tensor
+            if isinstance(output, torch.Tensor):
+                return FixedLayout(
+                    device=output.device,
+                    dtype=output.dtype,
+                    size=output.shape,
+                    stride=output.stride(),
+                )
+
+        except Exception:
```
eellison: When do we expect an exception ? if we do expect an exception, can we more narrowly try catch it ?

torch/_inductor/kernel/custom_op.py
```diff
@@ -0,0 +1,194 @@
...
+        name: Operation name for identification
+        decompositions: Single decomposition function or list of alternative implementations
+        inputs: Input tensors/nodes
+        kwargs: Additional arguments for decomposition functions
+        layout: Output layout (inferred if None)
+
+    Returns:
+        Optimized implementation result
+    """
+    if kwargs is None:
```
eellison: The default implementation of the operator should also be used

torch/_inductor/kernel/custom_op.py
```diff
@@ -0,0 +1,194 @@
...
+                make_fx_graph=traced_fn,
+            )
+            choices.append(choice)
+
+        return choices
+
+
+def autotune_custom_op(
+    name: str,
+    decompositions: Union[Callable[..., Any], list[Callable[..., Any]]],
```
eellison: Can this always be a list ? do we need it to be a union of callable, list?

torch/_inductor/kernel/custom_op.py
```diff
@@ -0,0 +1,194 @@
...
+                "rmsnorm",
+                rmsnorm_decomposition,
+                [input_tensor, weight],
+                {"eps": eps},
+                layout=layout,
+            )
+    """
+
+    def decorator(fn: Callable[..., Any]) -> Callable[..., Any]:
+        return register_lowering(custom_op, type_promotion_kind=None)(fn)
```
eellison: We should not expose register_lowering. I'd rather we just expose apis for registering autotuning decompositions

torch/_inductor/select_algorithm.py
```diff
@@ -3663,6 +3679,27 @@ def clear_preprocessing_fns(self, clear_defaults: bool = False):
...
             self._register_default_preprocessing_fns()
 
+    def _can_fallback_to_first_choice(
+        self, name: str, choices: list[ChoiceCaller]
+    ) -> bool:
+        """
+        Check if this is a custom op autotuning case that can fallback to the first choice.
+
+        Custom ops with multiple decompositions can safely fallback to the first decomposition
+        when all autotuning choices fail (e.g., on CPU-only environments).
```
eellison: When would cpu-only envs cause autotuning choices to fail?

torch/_inductor/kernel/custom_op.py
```diff
@@ -0,0 +1,194 @@
...
+"""
+Custom operation autotuning support for PyTorch Inductor.
+
+This module extends Inductor's autotuning capabilities to support arbitrary custom
+operations. Users can define ATen-based decompositions for their custom ops and
+automatically generate optimized implementations through Inductor's autotuning system.
+
+Example:
+    @register_custom_op_lowering(custom_ops.rmsnorm.default)
+    def tuned_rmsnorm(input_tensor, weight, eps=1e-6, *, layout=None):
```
eellison: `register_lowering` is for inductor ir, which we do not want to release as stable. can we just register decompositions instead ?

test/inductor/test_custom_op_autotune.py
```diff
@@ -0,0 +1,474 @@
...
+    Uses the tanh approximation which may be faster on some hardware.
+    """
+    gate_proj = torch.matmul(input_tensor, gate_weight)
+    up_proj = torch.matmul(input_tensor, up_weight)
+    gated = gelu_tanh_approximation(gate_proj) * up_proj
+    output = torch.matmul(gated, down_weight)
+    return output
+
+
+class TestCustomOpAutoTune(TestCase):
```
eellison, line 24: nit: would love if these tests were a bit more concise.

eellison (commented):
good first start ! 

Would be great to consolidate a bit of the logic with existing subgraph choice.
Still needs to be handled:
- stride output logic validation
- register fn for input generation

torch/_inductor/kernel/custom_op.py
tianrengao: Reused SubgraphTemplate with updated member functions

torch/_inductor/kernel/custom_op.py
tianrengao: Decompositions is always a list. Assertion added

torch/_inductor/kernel/custom_op.py
tianrengao: Removed register_lowering and used lowering dictionary instead in a wrapper of decorator.

torch/_inductor/select_algorithm.py
tianrengao: Removed this change. Cpu tests should behave similar as GPU tests. If all choices failed, it falls back to default implementation of the custom op

test/inductor/test_custom_op_autotune.py
tianrengao: Updated the test cases to cover a RMSNorm, a MLP, and a emulated DecomposeK tests . Each test contains various implementations, and asserts the numerical results are equivalent. 
The run log should show autotuning benchmark including a default implementation of the custom op.

torch/_inductor/kernel/custom_op.py
tianrengao: Sure, included default implementation with an ExternKernel Choice caller

tianrengao:
> good first start !
> 
> Would be great to consolidate a bit of the logic with existing subgraph choice. Still needs to be handled:
> 
> * stride output logic validation
> * register fn for input generation

Thanks for the review! @eellison 
I added the stride output logic validation and register fn for input generation. Also the code structure is adjusted a bit to improve readability. I also resolved comments and updated tests. Now the default implementation of customop is also used as a choice for benchmark.

torch/_inductor/select_algorithm.py
```diff
@@ -3483,6 +3482,11 @@ def log_results(
         dtypes = ", ".join([str(n.get_dtype()) for n in input_nodes])
         if config.autotune_num_choices_displayed == 0:
             return
+
+        # If no timings (e.g., after fallback), skip detailed logging
+        if not timings:
```
tianrengao: nit: Not needed. I will disgard this change later

test/inductor/test_custom_op_autotune.py
```diff
@@ -0,0 +1,455 @@
...
+
+        lib_name, op_name = test_op_name.split("::")
+        op_object = getattr(getattr(torch.ops, lib_name), op_name)
+
+        decompositions = [
+            rmsnorm_decomposition1,
+            rmsnorm_decomposition2,
+        ]
+
+        input_gen_fns = {
```
eellison: Inductor ir is an internal implementation detail. This applies for the input gen fns, decompositions etc. we want to only expose aten. 

So this function should take in fake tensors, and the user can return real ones

tianrengao: Removed input_gen_fns from user's arguments and moved it to internal custom_op.py which registers input_gen_fns with random inputs generated from fake tensors. Thanks for pointing it out!

torch/_inductor/kernel/custom_op.py
```diff
@@ -0,0 +1,175 @@
...
+        Decorator function for registering the autotuning function
+
+    Example:
+        @register_custom_op_autotuning(my_ops.rmsnorm.default)
+        def tuned_rmsnorm(input_tensor, weight, eps=1e-6):
+            return autotune_custom_op(
+                "rmsnorm",
+                [rmsnorm_impl1, rmsnorm_impl2, rmsnorm_impl3],
+                [input_tensor, weight],
+                {"eps": eps},
```
zou3519: why do we need to pass in args kwargs if they're just going to be the same as what is passed to the op?

test/inductor/test_custom_op_autotune.py
```diff
@@ -0,0 +1,433 @@
...
+
+        @register_custom_op_autotuning(op_object.default)
+        def _(input_tensor, weight, eps: float = 1e-8, default_impl=None):
+            return autotune_custom_op(
+                name="test_rmsnorm_autotuned",
+                decompositions=decompositions,
+                inputs=[input_tensor, weight],
+                kwargs={"eps": eps},
+                default_impl=default_impl,
+            )
```
zou3519: In your PR body, the decorator is named `register_custom_op_lowering`. That made sense to me: we are specifying in the lowering that inductor has an option to do autotuning.

If the decorator is named `register_custom_op_autotuning`, I don't see why we also have to call a function called `autotune_custom_op`. What else can one write in this function?

test/inductor/test_custom_op_autotune.py
```diff
@@ -0,0 +1,433 @@
...
+from torch.testing._internal.common_utils import skipIfXpu
+from torch.testing._internal.inductor_utils import HAS_GPU
+
+
+torch.set_float32_matmul_precision("high")
+
+
+class TestCustomOpAutoTune(TestCase):
+    """Test custom operation autotuning functionality."""
+
```
zou3519, line 26: One of the use cases (I'm still looking for the PR) is that someone has a custom operator that takes in a flag. The flag is some hyperparameter: it's the "algorithm" to use. The user would like to specify that inductor should change the flag (to 0, 1, or 2).

This seems a bit difficult to do with the current UX. When someone is invoking the custom operator they need to make a decision on what the flag should be.

triton's autotuning decorator does allow for this case. If you are autotuning an input to a triton kernel then you can invoke the triton kernel without said input.

test/inductor/test_custom_op_autotune.py
tianrengao: Thanks, I agree that `autotune_custom_op` seems redundant as an API for users if we already have a decorator. The main purpose was to expose an argument of decomposition list for the users. Let me think about if I can unify them into a one-stop decorator, either with a list of decompositions or allowing calling `register_custom_op_lowering` multiple times for multiple implementations, and hide the `autotune_custom_op`.

test/inductor/test_custom_op_autotune.py
```diff
@@ -0,0 +1,433 @@
...
+This module tests the custom op autotuning system, which allows users to provide
+multiple decomposition implementations of custom operations and automatically
+select the best performing one through Inductor's autotuning system.
+
+The tests cover:
+1. Custom op registration and autotuning for different operations (RMSNorm, MLP, Attention)
+2. Numerical equivalence between different decomposition implementations
+3. End-to-end compilation and performance validation
+4. Fallback behavior when decompositions fail
+"""
```
zou3519, line 8: High level question: there's roughly two features that were asked.
1) Autotune a custom operator vs its decomposition into native pytorch ops. vLLM actually has a lot of these. For example, there's an RMSNorm kernel, and there is a decomposition into native pytorch ops. vLLM by default selects the decomposition during compile. But in some cases we've decided to use the custom operator instead of its decomposition and we don't use vLLM's CustomOp machinery. It would be good to provide a way to say "hey, here's a decomp, use it during autotuning only."

https://github.com/vllm-project/vllm/blob/0d37450eb7e38ac82df35d4e0f21d4254435049d/vllm/model_executor/layers/layernorm.py#L151

2) Autotune hyperparameters for a custom operator. `myop(tensor x, int method)`. Method = 0, 1, 2, decides different algorithms for the execution. A user wants to autotune the `method`.

Are both of these use cases meant to be covered by your API? Or just the first one?

I think our past discussions were that this one API should cover both use cases but correct me if I'm wrong. @eellison

tianrengao: I see, thanks for the valuable comments. We should cover both use cases. This PR has implemented it but the current API somehow doesn't fit exactly to what you described. 

For 1, I think the current design should have already covered it, but the API is a bit ugly.

For 2, so far the PR allows users to register `decompositions` = [`myop(tensor x, int method=0)`, `myop(tensor x, int method=1)` and `myop(tensor x, int method=2)`] and inductor will autotune them. If you look at the decomposek test in the PR, it covers this case, but in an awkward way. 
To match your description, it seems we need an additional wrapper. I should modify the current API to take in a customop and a list of their tuning knob values(`method=[0,1,2]`), then the inductor internally map the method values to create multiple decompositions, and reuse the current stuff.

test/inductor/test_custom_op_autotune.py
tianrengao: Ah I see what you expected. So we need a proper decorator API, allowing user to register a list of flag values, and automatically tune them. 
Just to clarify, will there be multiple flags? say my_op(tensor x, flag1:[0,1,2], flag2:[a, b, c])? It would be good if you could share me the triton's autotuning decorator as a reference. 

Update: this has been implemented. The user can provide multiple tuning knobs and inductor benchmark all possible variants for those parameter values and return the fastest one

test/inductor/test_custom_op_autotune.py
```diff
@@ -0,0 +1,433 @@
...
+        self.device = "cuda" if HAS_GPU else "cpu"
+        self.dtype = torch.float16 if self.device == "cuda" else torch.float32
+
+    def simple_randn_gen(self, x):
+        """Simple random normal generator for input_gen_fns - reusable across all tests."""
+        from torch._inductor.virtualized import V
+
+        return torch.randn(
+            [V.graph.sizevars.size_hint(i) for i in x.get_size()],
+            dtype=x.get_dtype(),
```
eellison: This looks like Inductor Buffer - we should pass in FakeTensors.

youkaichao (commented):
High-level questions:

> Users can register several decomposition functions for the same custom operation

by "decomposition functions", do you mean the decomposition functions will still be traced by Dynamo/Inductor? what variants do you use to benchmark? what variants are targeted here? are they simple python code that just run as-is, or still go through Dynamo/Inductor optimizations (and at which level?)

Do you think we should give users the control over how to benchmark each variants? e.g.  for some communication ops, we might need to do a sync before measuring anything.

Does PyTorch provide autotuning cache? And what are the cache keys and cache values?

test/inductor/test_custom_op_autotune.py
tianrengao: Updated the PR to realize both 2 targets. Now autotuning hyperparameters is possible via an optional tuning_knob argument

test/inductor/test_custom_op_autotune.py
tianrengao: Updated the func to use fake tensors as input

torch/_inductor/kernel/custom_op.py
tianrengao: Removed args and kwargs. Now the API usage is just asking for decompositions and optional tuning_knob of the custom_op.
```
register_custom_op_autotuning(
    custom_op=torch.ops.mylib.attention.default,
    decompositions=[attention_variants],
    tuning_knob={"method": [0, 1, 2], "use_fp32": [True, False]}  
)
```

test/inductor/test_custom_op_autotune.py
```diff
@@ -0,0 +1,613 @@
...
+
+        @test_decompose_k_op.register_fake
+        def _(a: torch.Tensor, b: torch.Tensor, k_splits: int = 4):
+            return torch.empty(a.shape[0], b.shape[1], device=a.device, dtype=a.dtype)
+
+        lib_name, op_name = test_op_name.split("::")
+        op_object = getattr(getattr(torch.ops, lib_name), op_name)
+
+        # Use parameter tuning to test different k_splits values
+        register_custom_op_autotuning(
```
eellison, line 377: Would it make sense to have a knob for whether or not max-autotune is enabled ? potentially there are some cases where we'd want to opt in even if it weren't enabled globally, or vice versa. cc @zou3519

tianrengao: It makes sense to me to add a max-autotune knob, for example in this MM split-k case.
Should the user provide the max-autotune configurations themselves?

test/inductor/test_custom_op_autotune.py
```diff
@@ -0,0 +1,613 @@
...
+            mlp_decomposition2,
+            mlp_decomposition3,
+        ]
+
+        register_custom_op_autotuning(
+            op_object.default,
+            decompositions=decompositions,
+            name="test_mlp_autotuned",
+            input_gen_fns={
+                0: lambda fake_tensor: torch.randn_like(fake_tensor, device=self.device)
```
eellison: this is a non-intuitive api to me. Would it make sense to pass the argname instead ?

cc @zou3519 for thoughts

torch/_inductor/codegen/subgraph.py
```diff
@@ -207,3 +212,113 @@ def generate(  # type: ignore[override]
...
+        kwargs = kwargs or {}
+
+        # Infer layouts and ensure stride consistency for fair autotuning comparison
+        layouts = [
+            self._infer_custom_op_layout(input_nodes, [decomp], kwargs, default_impl)
+            for decomp in decompositions
+        ]
+
+        self._validate_stride_consistency(name, decompositions, layouts)
+        layout = layouts[0]  # All layouts have equivalent stride now
```
eellison: this assumes there is a single output. Can we assert that, or make this more robust?

torch/_inductor/codegen/subgraph.py
```diff
@@ -207,3 +212,113 @@ def generate(  # type: ignore[override]
...
+        choices = []
+        for decomp in decompositions:
+            # Create make_fx_graph function for this decomposition
+            def make_fx_graph(*args: Any, decomp: Callable[..., Any] = decomp) -> Any:
+                import functools
+
+                from torch.fx.experimental.proxy_tensor import make_fx
+
+                # Ensure kwargs is not None for unpacking
+                decomp_kwargs = kwargs if kwargs is not None else {}
```
eellison: Don't we want to functools.partial all of the non tensor args ? What if a tensor is in kwargs ?

torch/_inductor/codegen/subgraph.py
```diff
@@ -207,3 +212,113 @@ def generate(  # type: ignore[override]
...
+        import functools
+
+        from torch._inductor.ir import FixedLayout, ir_node_to_tensor
+        from torch._inductor.virtualized import V
+
+        # Use default_impl if available, otherwise use first decomposition
+        impl = default_impl if default_impl is not None else decompositions[0]
+
+        with V.fake_mode:
+            example_inputs = [ir_node_to_tensor(inp) for inp in input_nodes]
```
eellison: `ir_node_to_tensor` is going to guard - that's probably not what we want.

torch/_inductor/kernel/custom_op.py
```diff
@@ -0,0 +1,345 @@
...
+        Tuple of (tensor_inputs_list, non_tensor_kwargs)
+    """
+    tensor_inputs = []
+    non_tensor_kwargs = {}
+
+    for arg in args:
+        if isinstance(arg, (TensorBox, Buffer)) or (
+            hasattr(arg, "dtype") and hasattr(arg, "shape")
+        ):
+            tensor_inputs.append(arg)
```
eellison: what happens to non tensor arg inputs ? they just get dropped

torch/_inductor/kernel/custom_op.py
```diff
@@ -0,0 +1,345 @@
...
+
+    for key, value in kwargs.items():
+        if isinstance(value, (TensorBox, Buffer)) or (
+            hasattr(value, "dtype") and hasattr(value, "shape")
```
eellison: when are these `hasattr(value, "dtype") and hasattr(value, "shape")` necessary ? can we delete them?

torch/_inductor/kernel/custom_op.py
```diff
@@ -0,0 +1,345 @@
...
+        inputs: List of input IR nodes from compilation
+        user_input_gen_fns: User-provided input generation functions
+
+    Returns:
+        Dict mapping indices to internal input generation functions
+    """
+    internal_input_gen_fns = {}
+
+    with V.fake_mode:
+        fake_inputs = [ir_node_to_tensor(inp) for inp in inputs]
```
eellison: similarly, this is going to guard on dynamic shapes, thats probably not what we want. or we want to opt out of dynamic shape inputs.

eellison (commented):
another round of comments - cc @zou3519 , @BoyuanFeng please take a look

torch/_inductor/codegen/subgraph.py
tianrengao: The kwargs passed in SubgraphTemplate have been filtered and only contains non-tensor args. 
Processed here https://github.com/pytorch/pytorch/pull/164212/files#diff-a8f767b3bf3b50f0398a6599a9517c4a117255c5c061555883d974e233215bd0R40-R54

And being passed in from here:
https://github.com/pytorch/pytorch/pull/164212/files#diff-a8f767b3bf3b50f0398a6599a9517c4a117255c5c061555883d974e233215bd0R343

youkaichao:
some clarifications:

there're two concepts of "custom ops" in vLLM:

1. pytorch's custom op, registered by `direct_register_custom_op` . this is a real custom op in pytorch world, and we can call `torch.ops.vllm.xxx` to call the underlying function.
2. vLLM's custom op module, a `CustomOp` class that inherits `nn.Module`. It's mostly for dispatching the forward function to different platform implementations. Essentially it is still a `nn.Module` and is stateful.

the custom op autotune looks attractive. it should target the first kind of custom op iiuc. maybe we can augment vLLM's `direct_register_custom_op` with the proposed `register_custom_op_autotuning`. one caveat is we need to split the native torch implementation https://github.com/vllm-project/vllm/blob/44c85556216bb24ab5ccf410dbb2bcfc25c85dd5/vllm/model_executor/layers/layernorm.py#L198 into `rms_norm` and `fused_add_rms_norm`. Maybe it's not a big deal, we can ignore the `forward_native` function and just write another native `rms_norm` and `fused_add_rms_norm` implementation for inductor to compile.

> Flow: Python functions → make_fx → FX graph → Inductor compilation (no fusion, Triton kernels) → benchmark competition

It's not clear what would happen if we have some graph transforms that need to recognize the rmsnorm pattern. cc @ProExpertProg Luka knows more about this. I'm mainly concerned about how this custom op tuning interacts with the way vLLM uses inductor.

another potential usage is to tune the allreduce implementation. see https://github.com/vllm-project/vllm/blob/17838e50efc9e686fffc901dfed811b15b7d614a/vllm/distributed/device_communicators/cuda_communicator.py#L127 to check how many allreduce implementations we have. I really want to simplify them (that's why I mention distributed communication). however, I don't think people will always turn on autotune, so we need to keep all these code path anyway 😢 

btw, one more caveat: some operations use different code path for cudagraph capture and non-cudagraph capture, and that can be a problem for custom op autotune. but maybe we can leave them alone now, too complicated to consider.

eellison:
> It's not clear what would happen if we have some graph transforms that need to recognize the rmsnorm pattern. cc @ProExpertProg Luka knows more about this. I'm mainly concerned about how this custom op tuning interacts with the way vLLM uses inductor.

Relatedly - I'm not sure how much of the inductor pass pipeline we want to apply to these decompositions. Potentially this is something that can be opted into by the user. I think a default of no passes probably makes sense. If you're writing a decomp, this is already narrowly scoped to the point where you should be able to write it as you want it to end up..

eellison:
@youkaichao we've also seen instances where vllm custom ops were slower than the corresponding inductor generated decomposition. in any case, the decomposition can be optional.

test/inductor/test_custom_op_autotune.py
zou3519: @tianrengao did you do anything in response to this comment? I see the PR description hasn't been updated

test/inductor/test_custom_op_autotune.py
zou3519: Is the suggestion that a user should be able to specify configs to autotune when we're under max-autotune? If so, then the implication for this PR is that the custom op autotuning runs by default (without max-autotune)?

torch/_inductor/kernel/custom_op.py
```diff
@@ -0,0 +1,351 @@
...
+            decompositions=[standard_attention, flash_attention],
+            tuning_knob={
+                "head_dim": [32, 64, 128],
+                "method": ["chunked", "tiled"]
+            },
+            input_gen_fns={
+                0: lambda fake: torch.randn_like(fake, device='cuda') * 0.1,
+                1: lambda fake: torch.randn_like(fake, device='cuda') * 0.1,
+                2: lambda fake: torch.randn_like(fake, device='cuda') * 0.1,
+            }
```
zou3519, line 361: What is {0, 1, 2} ?

torch/_inductor/kernel/custom_op.py
```diff
@@ -0,0 +1,351 @@
...
+
+    Example:
+        # Multiple decompositions with parameter tuning
+        register_custom_op_autotuning(
+            torch.ops.mylib.attention.default,
+            decompositions=[standard_attention, flash_attention],
+            tuning_knob={
+                "head_dim": [32, 64, 128],
+                "method": ["chunked", "tiled"]
+            },
```
zou3519: I feel like this API is trying to do too much.

> tuning_knob

Instead of tuning the cartesian product, it seems like it would be better to provide an explicit list of all of the configs, like how triton.autotune does it. Right now there is no way to autotune something that is not the explicit cartesian product of all the configs.

> decompositions=[standard_attention, flash_attention],
tuning_knob={

Now I'm lost -- do we autotune the cartesian product of all of the tuning_knobs and the decompositions? e.g. is we standard_attention with head_dim=32 a tuning option?

zou3519: Here's a strawman:
```
register_custom_op_autotuning(
  torch.ops.mylib.attention.default,
  configs = (
    Config(head_dim=32, method="chunked"),
    Config(head_dim=64, method="chunked"),
    Decomposition(standard_attention),
  )
)
```

torch/_inductor/kernel/custom_op.py
```diff
@@ -0,0 +1,351 @@
...
+    When tuning_knob is provided, creates variants of each decomposition
+    with all combinations of parameter values (Cartesian product).
+
+    Args:
+        custom_op: Custom operation to register (e.g., torch.ops.mylib.myop.default)
+        decompositions: Implementation functions to benchmark
+        name: Operation name for identification (default: "{op_name}_autotuned")
+        input_gen_fns: Custom input generators for benchmarking
+        tuning_knob: Optional parameter tuning dict. Supports multiple parameters.
+                    Creates all combinations of parameter values.
```
zou3519: @tianrengao 

Let's say I specify tuning_knob={"head_dim": [32, 64, 128]}.
Does this mean I still need to pass a `head_dim` to my custom operator?

In triton autotuning, when you do this, I think the triton kernel doesn't need to be passed a head_dim anymore. Maybe that is too much magic though

@eellison for a second opinion

zou3519: it's going to be very hard to omit an input to a custom operator, so I think we should just make it so that the autotuning configs are "in addition" to the way the user has invoked their custom operator

torch/_inductor/kernel/custom_op.py
tianrengao: Good point! Yes, so far we have been tuning a Cartesian product of all the configurations. This approach is excessive, and it would be better to follow the method used in triton.autotune by allowing only an explicit list of configurations, as shown in the example.

torch/_inductor/kernel/custom_op.py
tianrengao: Sorry this is a deprecated docstring. The correct behavior in code today is mapping from _argument name string_ to the _input_gen_fn_. I will update this part.

torch/_inductor/kernel/custom_op.py
tianrengao: Updated the API to be 
```        
register_custom_op_autotuning(
            torch.ops.mylib.attention.default,
            configs=[
                CustomOpConfig(attention_impl, head_dim=32, method='chunked'),
                CustomOpConfig(attention_impl, head_dim=64, method='tiled'),
                CustomOpConfig(fallback_impl),  # No params
            ],
            input_gen_fns={
                "query": lambda fake: torch.randn_like(fake, device='cuda'),
                "key": lambda fake: torch.randn_like(fake, device='cuda'),
                "value": lambda fake: torch.randn_like(fake, device='cuda'),
            }
        )
```

So it provides fine-grained control for users to combine a set of param values and decompositions.

test/inductor/test_custom_op_autotune.py
tianrengao: Yes, updated the PR description.

test/inductor/test_custom_op_autotune.py
tianrengao: @zou3519
Custom op autotuning runs by default for the parameters registered via `CustomOpConfig`, even if max-autotune is not enabled globally.

To enable max-autotune, users can specify a `"max_autotune_config"` parameter with detailed values in `CustomOpConfig`. I'm considering implementing this in the next PR.

Does this approach sound reasonable to you?

torch/_inductor/kernel/custom_op.py
tianrengao: Followed the method that autotune configs are "in addition" to what user has invoked in their custom op signature.

torch/_inductor/kernel/custom_op.py
```diff
@@ -0,0 +1,417 @@
...
+
+    Examples:
+        register_custom_op_autotuning(
+            torch.ops.mylib.attention.default,
+            configs=[
+                CustomOpConfig(attention_impl, head_dim=32, method='chunked'),
+                CustomOpConfig(attention_impl, head_dim=64, method='tiled'),
+                CustomOpConfig(fallback_impl),  # No params
+            ],
+            input_gen_fns={
```
zou3519, line 357: It's not clear to me what these are for - add a comment? Do we use this to generate one or multiple example inputs? If we generate multiple example inputs, what exactly is auto tuning benchmarking? (Best config with mean performance?)

test/inductor/test_custom_op_autotune.py
```diff
@@ -0,0 +1,542 @@
...
+            max_autotune_gemm_backends=autotune_backends,
+            fx_graph_cache=False,
+            benchmark_kernel=True,
+        ):
+            compiled_result = test_model(*inputs)
+
+        self.assertEqual(
+            compiled_result.shape, expected.shape, f"{test_name} shape mismatch"
+        )
+        torch.testing.assert_close(
```
zou3519, line 54: Why not self.assertEqual? That's how we test the other stuff in PyTorch

test/inductor/test_custom_op_autotune.py
```diff
@@ -0,0 +1,542 @@
...
+                f"{op_name} {name} produced non-finite values",
+            )
+
+        # Verify numerical equivalence
+        reference_name, reference_result = next(iter(results.items()))
+        for name, result in results.items():
+            if name != reference_name:
+                rtol = 1e-1 if "Approximated" in name else 1e-2
+                atol = 1e-1 if "Approximated" in name else 1e-2
+                torch.testing.assert_close(
```
zou3519, line 82: Ditto

test/inductor/test_custom_op_autotune.py
```diff
@@ -0,0 +1,542 @@
...
+                device=input_tensor.device,
+                dtype=input_tensor.dtype,
+            )
+
+        lib_name, op_name = test_op_name.split("::")
+        op_object = getattr(getattr(torch.ops, lib_name), op_name)
+
+        # Use explicit configs with method parameter as tuning knob
+        register_custom_op_autotuning(
+            op_object.default,
```
zou3519: We should just be able to pass the function that is decorated by the custom_op decorator (test_mlp_op) No need to do that in this pr though.

zou3519 (approved):
Thank you, I like the new explicit api.

torch/_inductor/kernel/custom_op.py
tianrengao: This is mentioned in the PR description.
The input_gen_fns is used for simulating user's real data distribution during benchmarking. During compile time, we use random data by default for benchmarking. This API allow users to provide fake tensors as inputs and generates ideal distribution. This not generating multiple example inputs. This function is used to reduce any gap between the real data and benchmark data.

test/inductor/test_custom_op_autotune.py
tianrengao: Sure. Will change to assertEqual later

PR #164225, opened 2025-09-30
Missing lambda in torch._check

xadupre:


Skylion007 (approved):
Huh why didn't type checking detect this.

PR #164227, opened 2025-09-30
`canUse32BitIndexMath` : simplify the logic

nikitaved:
As per title. We can answer this question in a single pass, with fewer ops.

### DIFF
--- a/aten/src/ATen/native/IndexingUtils.cpp
+++ b/aten/src/ATen/native/IndexingUtils.cpp
@@ -4,31 +4,22 @@
 namespace at::native {
 
 bool canUse32BitIndexMath(const TensorBase& t, int64_t max_elem) {
-  auto elements = t.sym_numel();
-  if (elements >= max_elem) {
-    return false;
-  }
-  if (elements == 0) {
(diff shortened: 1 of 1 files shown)


aten/src/ATen/native/IndexingUtils.cpp
```diff
@@ -4,31 +4,22 @@
...
-    return max_elem > 0;
-  }
-
+  const auto strides = t.sym_strides();
+  const auto sizes = t.sym_sizes();
   c10::SymInt offset = 0;
-  auto linearId = elements - 1;
 
   // NOTE: Assumes all strides are positive, which is true for now
   // NOLINTNEXTLINE(bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions)
```
nikitaved, line 12: Not quite sure we need this lint here anymore...

albanD (commented):
The early conditions are for performance reasons I assume. Is the claim that this doesn't make any difference?

Also the linearId logic is weird but I assume was added there for a good reason, is that reason not valid anymore?

PR #164228, opened 2025-09-30
Fix CSEPass handling of grad mode and in-place node

Eldalie:
### Motivation

CSEPass did not correctly handle gradient mode and in-place operations, which could lead to incorrect optimization behavior. Additionally, its initialization API was not flexible enough for the intended use.

### Changes

- Handle gradient mode properly
- Handle in-place nodes
- Fix typing annotations
- Add tests for in-place operations and gradient mode
- Update the initialization of CSEPass to take a callable as input, but since it is marked as is_backward_compatible=False, this should not be an issue.

### Testing

- Added dedicated tests covering in-place operations and gradient mode

### DIFF
--- a/test/fx/test_cse_pass.py
+++ b/test/fx/test_cse_pass.py
@@ -1,16 +1,17 @@
 # Owner(s): ["oncall: fx"]
-
+import operator
 import random
+from copy import deepcopy
 
 import torch
-from torch.fx import symbolic_trace
+from torch._ops import OpOverload
+from torch.fx import Node, symbolic_trace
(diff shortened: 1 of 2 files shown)


ezyang:
PR desc looks like Codex

ezyang:
This is timely though, we were discussing making our passes support internal mutation

Eldalie:
> PR desc looks like Codex

@ezyang 
Haha no, it's good old ChatGPT.

I fixed the test issue by adding a __all__ and restoring get_CSE_banned_ops. However, I am not sure why the BC-linter fails, since is_backward_compatible=False.

PR #164252, opened 2025-09-30
Update compute elementwise output strides semantics

morrison-turnansky:
Continuation of work from https://github.com/pytorch/pytorch/pull/161400 and https://github.com/pytorch/pytorch/pull/163017. 

Updating stride semantics for ```clone_meta``` and underlying function, ```compute_elementwise_output_strides```. 

### DIFF
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -915,9 +915,9 @@ def forward(self, x_1):
             )
         )
 
-    def test_prims_non_overlapping_and_dense(self):
+    def test_is_non_overlapping_and_dense_or_false(self):
         shape_env = ShapeEnv()
-        cf = torch._prims_common.is_non_overlapping_and_dense
+        cf = torch._prims_common.is_non_overlapping_and_dense_or_false
 
         # backed case
         a0 = create_symint(shape_env, 5)
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -915,7 +915,7 @@ def forward(self, x_1):
             )
         )
 
-    def test_is_non_overlapping_and_dense_or_false(self):
+    def test_prims_is_non_overlapping_and_dense_or_false(self):
         shape_env = ShapeEnv()
         cf = torch._prims_common.is_non_overlapping_and_dense_or_false
 
--- a/torch/_prims/__init__.py
+++ b/torch/_prims/__init__.py
@@ -693,12 +693,8 @@ def _clone_meta(
     else:
         # Match eager behavior by preserving strides for non_overlapping_and_dense tensors
         # If not, eager clone creates contiguous strides
-        computed_stride = None
-        if utils.is_non_overlapping_and_dense(input):
-            computed_stride = input.stride()
-        else:
-            computed_stride = utils.compute_elementwise_output_strides(input)
-
+        computed_stride = utils.compute_elementwise_output_strides(input)
+    
         return torch.empty_strided(
             input.shape,
             computed_stride,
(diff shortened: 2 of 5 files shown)


morrison-turnansky:
@Lucaskabela I thought it would be cleaner to start on new PR. 

The above fixes the case when one tensor is given. The reason this works is that ```torch.empty_like``` uses the c++ implementation of infer_dense_strides under the hood, but this requires some extra allocations. 

As an alternative fix to avoid this extra cost, there is a python implementation  of the above function, namely ```torch._inductor.kernel.flex.infer_dense_strides```. In order to use it, I think we should/would have to move it out of _inductor. Let me know what you think.

torch/_prims/__init__.py
```diff
@@ -694,12 +695,7 @@ def _clone_meta(
     else:
         # Match eager behavior by preserving strides for non_overlapping_and_dense tensors
         # If not, eager clone creates contiguous strides
-        computed_stride = None
-        if utils.is_non_overlapping_and_dense(input):
-            computed_stride = input.stride()
```
Lucaskabela, line 698: Seems removing this line is causing the unit test for this bug to fail :/

torch/_prims_common/__init__.py
```diff
@@ -510,7 +510,7 @@ def __lt__(self, other):
     return check_contiguous_sizes_strides(sizes, strides, false_if_dde=True)
 
 
-def is_non_overlapping_and_dense(a: Tensor) -> bool:
+def is_non_overlapping_and_dense_or_false(a: Tensor) -> bool:
```
Lucaskabela, line 521: Thanks for propagating this naming

Lucaskabela (changes requested):
Seems this change still has some error to work through

Lucaskabela:
I took a look at this function - I think we can refactor it out to leverage it here

torch/_prims/__init__.py
morrison-turnansky: @Lucaskabela which failing test are you referring to?

The only failing test I saw in the CI was  
test/test_dynamic_shapes.py: test_prims_non_overlapping_and_dense. 

I had missed renaming one of the function calls. Updated and pushed change.

morrison-turnansky:
@Lucaskabela Circling back on this. I had fixed the failing test in the CI pass. If there are other changes that you want can you elaborate? Unclear what the intent of this comment was. 

> I took a look at this function - I think we can refactor it out to leverage it here

test/test_dynamic_shapes.py
```diff
@@ -897,9 +897,9 @@ def test_non_overlapping_and_dense_unbacked(self):
             )
         )
```
Lucaskabela, line 917: I think @zou3519 had some extra test cases for us to consider here; if not, I still think it would be worth adding an additional test case or two to make sure we are doing the right thing here

Lucaskabela (commented):
Thanks for the changes! I think we should add test cases that this change is covering our previous change is not, otherwise LGTM

morrison-turnansky:
@Lucaskabela Enjoy the PTO!

test/test_dynamic_shapes.py
morrison-turnansky: @Lucaskabela Following up. Any addtional action?

morrison-turnansky:
CI is passing. Please let me know if you want any additional changes, and/or resolve requested changes if you found my updates were sufficient. Thank you. Trying to tie this up before  I go on PTO for end of year holidays.

Lucaskabela (approved):
Thanks for the change, I think this is generally good now :)

PR #164253, opened 2025-09-30
[DTensor] Allow redistribute to Partial if src matches

SherlockNoMad:
* __->__ #164253

### DIFF
--- a/torch/distributed/tensor/_api.py
+++ b/torch/distributed/tensor/_api.py
@@ -530,7 +530,8 @@ class DTensor(torch.Tensor):
         for i, placement in enumerate(placements):
             if placement.is_partial() and self.placements[i] != placement:
                 raise RuntimeError(
-                    f"Can not redistribute from {self.placements[i]} to {placement}, redistributing to Partial is for internal use only!"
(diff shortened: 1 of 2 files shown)


zpcore (approved):
Good catch, LGTM!

torch/distributed/tensor/_api.py
```diff
@@ -528,9 +528,10 @@ def redistribute(
 
         placements = list(placements)
         for i, placement in enumerate(placements):
-            if placement.is_partial():
+            if placement.is_partial() and self.placements[i] != placement:
```
tianyu-l, line 531: Does this code make sure the redistribute fails if both src and tgt are partial, but they different classes/subclasses of Partial?
e.g. when src is `Partial("sum")`, tgt is `Partial("avg")` or `_MaskPartial`
https://github.com/pytorch/pytorch/blob/d2c5f231f6ab62e523f49cee652367baefcbd9ce/torch/distributed/tensor/_ops/_embedding_ops.py#L70

Could you add some tests for such cases?

SherlockNoMad: already included, see last test case

PR #164263, opened 2025-09-30
[BE][Easy]: Add prims common TypeGuard

Skylion007:
Slightly improves typing by adding a TypeGuard.

### DIFF
--- a/torch/_prims_common/__init__.py
+++ b/torch/_prims_common/__init__.py
@@ -19,7 +19,7 @@ from typing import (
     TypeVar,
     Union,
 )
-from typing_extensions import deprecated, TypeAlias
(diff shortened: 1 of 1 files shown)


albanD (commented):
The condition is actually more than that?

Will let @lolpack give an opinion on this one.

samwgoldman:
Even though the actual condition is more specific, TypeGuard is a suitably type. As long as `is_cpu_scalar_tensor(x)` implies `isinstance(x, TensorLike)`, it's safe. As you note, `not is_cpu_scalar_tensor(x)` does _not_ imply `not isinstance(x, TensorLike)`, but `TypeGuard`'s implication is not bidirectional.

albanD (approved):
Ok then!

PR #164265, opened 2025-09-30
[c10d] Integrate NCCL new `ncclAllToAll` into PT

fduwjj:
* __->__ #164265

NCCL 2.28.3 now supports `ncclAlltoAll`, this PR aims at integrating it. (release note: https://docs.nvidia.com/deeplearning/nccl/release-notes/rel_2-28-3.html#rel_2-28-3)


Skylion007 (dismissed):

torch/csrc/cuda/nccl.cpp
```diff
@@ -850,6 +850,8 @@ void all2all_single_equal_split(
   // operations issued as a part of the collective (e.g. alltoall) vs those
   // inside traditional p2p operations.
   NCCL_CHECK(ncclAllToAll(sendbuff, recvbuff, count, type, comm, stream));
+#elif ((NCCL_MAJOR > 2) || ((NCCL_MAJOR == 2) && (NCCL_MINOR >= 28)))
```
Skylion007: Wait, why isn't NCCL_ALLTOALL_SUPPORTED defined on this version?

Skylion007: Actually does this macro come from? Maybe that is the issue? I don't see it defined anywhere in NCCL or in PyTorch, is it FBCode specific?

fduwjj: it is something defined previously within Meta unfortunately.. And now NCCL supports this, things get interesting.

Skylion007: @fduwjj Do the APIs match? If so, can we just set the condition of ifdef and use NCCLX macro to switch between FB nccl and OSS NCCL?

Skylion007: Wait, yeah the API is the same so we can just define it ourselves in OSS and reuse the logic...

Skylion007 (approved):

torch/csrc/cuda/nccl.cpp
fduwjj: the API is different unfortunately...

Skylion007: Can we at least add a comment to explain it's from FBCODE lol?

Skylion007: @fduwjj Ah, one is lower case...
   
```NCCL_CHECK(ncclAllToAll(sendbuff, recvbuff, count, type, comm, stream));```
  vs 
```NCCL_CHECK(ncclAlltoAll(sendbuff, recvbuff, count, type, comm, stream));```

Skylion007 (approved):

torch/csrc/cuda/nccl.cpp
kwen2501: Can we use `NCCL_VERSION_CODE >= NCCL_VERSION(2, 28, 0)` for short?

kwen2501 (approved):

torch/csrc/cuda/nccl.cpp
fduwjj: no this macro is not available in this file..

kwen2501: @fduwjj the macro is defined by NCCL -- does not need to be in torch.

PR #164267, opened 2025-09-30
[c10d] Integrate ncclScatter into PT

fduwjj:
* __->__ #164267
* #164270
* #164265
* #164276


torch/csrc/cuda/nccl.cpp
```diff
@@ -25,6 +25,10 @@
 #define NCCL_HAS_COMM_NONBLOCKING 1
 #endif
 
+#if (NCCL_MAJOR > 2) || (NCCL_MAJOR == 2) && (NCCL_MINOR >= 28)
+#define NCCL_HAS_COMM_NONBLOCKING
```
Skylion007: Uh, this is defined on line 25, isn't it?

torch/csrc/cuda/nccl.cpp
```diff
@@ -1158,6 +1162,39 @@ void scatter(
...
+          i);
+      TORCH_CHECK(
+          src.is_cuda() && src.get_device() == outputs.get_device(),
+          "device mismatch at i=",
+          i);
+      flat.narrow(0, i * count, count).copy_(src, /*non_blocking=*/true);
+    }
+    // Make sure packing copies are visible to NCCL on the same stream
+    // (copy_ on the same stream is already ordered; no extra sync needed)
+    send_ptr = flat.data_ptr();
```
Skylion007: Same data_ptr feedback

torch/csrc/cuda/nccl.cpp
```diff
@@ -1158,6 +1162,39 @@ void scatter(
...
+      const at::Tensor& src = inputs[i].contiguous().view({count});
+      TORCH_CHECK(
+          src.scalar_type() == outputs.scalar_type(),
+          "dtype mismatch at i=",
+          i);
+      TORCH_CHECK(
+          src.is_cuda() && src.get_device() == outputs.get_device(),
+          "device mismatch at i=",
+          i);
+      flat.narrow(0, i * count, count).copy_(src, /*non_blocking=*/true);
```
Skylion007: Is there a layout trick we can to pack the src tensor into copy_ with the right striding here without a possible intermediate copy?

torch/csrc/cuda/nccl.cpp
fduwjj: opps rebase error.

torch/csrc/cuda/nccl.cpp
fduwjj: any suggestions on how to do that? I do want to get rid of this intermediate copy if possible

torch/csrc/cuda/nccl.cpp
```diff
@@ -1158,6 +1158,39 @@ void scatter(
...
```
Skylion007: Need to double check if the invariant holds, but I think you can exploit the fact that flat is already contiguous and do:
```suggestion
      const at::Tensor& src = inputs[i];
      TORCH_CHECK_VALUE(
          src.scalar_type() == outputs.scalar_type(),
          "dtype mismatch at i=", i);
      TORCH_CHECK_VALUE(
          src.is_cuda() && src.get_device() == outputs.get_device(),
          "device mismatch at i=", i);
      TORCH_CHECK_VALUE(
          src.numel() == count,
          "numel mismatch at i=", i, " (got ", src.numel(), ", expected ", count, ")");
    
      // Take the 1D slot [i*count:(i+1)*count], reshape it to src's shape,
      // then copy. This avoids making src contiguous.
      auto slot = flat.narrow(0, i * count, count).view(src.sizes()); // could even do sym_sizes here to be extra fancy
      slot.copy_(src, /*non_blocking=*/true);

File: torch/csrc/cuda/nccl.cpp (hunk 3 above)
Comment (Skylion007) [reply torch/csrc/cuda/nccl.cpp]: Pasted below, TORCH_CHECK_VALUE should be used here since it's input invariants being violated

File: torch/csrc/cuda/nccl.cpp [hunk 5]
```diff
@@ -1158,6 +1158,39 @@ void scatter(
...
   NCCL_CHECK_TIMEOUT(ncclCommUserRank(comm, &cur_rank), _comm);
 #endif
+#if ((NCCL_MAJOR > 2) || ((NCCL_MAJOR == 2) && (NCCL_MINOR >= 28)))
+  const void* send_ptr = nullptr;
+  auto type = to_nccl_data_type(outputs);
+  int64_t count = outputs.numel();
+  at::Tensor flat; // keep alive until after NCCL call
+  if (cur_rank == root) {
+    TORCH_CHECK(
+        (int)inputs.size() == numranks,
```
Skylion007:
```suggestion
    TORCH_CHECK_VALUE(
        static_cast<int>(inputs.size()) == numranks,
```

torch/csrc/cuda/nccl.cpp
Skylion007: You can exploit the fact that the empty Tensor can be forced to be contiguous on initialization
```suggestion
  flat = at::empty({numranks * count}, outputs.options(), c10::MemoryFormat::Contiguous);
  // Pack each inputs[i] into its slot
  
  for (int i = 0; i < numranks; ++i) {
    const at::Tensor& src = inputs[i];
    TORCH_CHECK_VALUE(
        src.scalar_type() == outputs.scalar_type(),
        "dtype mismatch at i=", i);
    TORCH_CHECK_VALUE(
        src.is_cuda() && src.get_device() == outputs.get_device(),
        "device mismatch at i=", i);
    TORCH_CHECK_VALUE(
        src.numel() == count,
        "numel mismatch at i=", i, " (got ", src.numel(), ", expected ", count, ")");
  
    // Take the 1D slot [i*count:(i+1)*count], reshape it to src's shape,
    // then copy. This avoids making src contiguous.
    auto slot = flat.narrow(0, i * count, count).view(src.sizes()); // could even do sym_sizes here to be extra fancy
    slot.copy_(src, /*non_blocking=*/true);

File: torch/csrc/cuda/nccl.cpp [hunk 6]
```diff
@@ -1158,6 +1158,39 @@ void scatter(
...
+        "root must provide inputs.size()==numranks");
+    // Allocate one flat buffer [world_size * count]
+    flat = at::empty({numranks * count}, outputs.options());
```
Skylion007:
```suggestion
    flat = at::empty({numranks * count},
                 outputs.options(),
                 c10::MemoryFormat::Contiguous);
```

Skylion007 (commented):
You can do this to assert the flat tensor is contiguous

kwen2501:
The current API accepts a list of tensors as input:
```
void scatter(
    const std::vector<at::Tensor>& inputs,
    at::Tensor& outputs,
    ...
)
```
Better to keep its impl grouped send/recv's to avoid an extra copy of the input tensors into a flattened buffer.

kwen2501:
The potential new API would accept a single tensor as input:
```
void _scatter_tensor(
    const at::Tensor& input,
    at::Tensor output,
    ...
)
```

kwen2501:
@fduwjj It looks like all tests passed. which means, we are indeed in lack of a test that covers this case:
> this list form API is supposed to support a case where outputs on different ranks have different lengths

Skylion007:
> @Skylion007 the Python surface would need a different API anyway.

I mean you could do if isinstance dispatch, but yeah I see your point.

PR #164269, opened 2025-09-30
Add TensorImplWeakRef for weak references to TensorImpl

### LINKED ISSUE #145253: PyObject preservation does not prevent weakrefs being cleared by Python garbage collector
### 🐛 Describe the bug

If a dead cycle holds the last reference to a resurrectable Tensor, gc will clear its weakrefs. One implication of this interaction is that entries can spuriously vanish from WeakTensorKeyDictionary, which is used in various places across the code base.

Repro:
```python
import gc; gc.disable()
a = torch.tensor(1.)
param = torch.tensor(2., requires_grad=True)
param.grad = a
b = [a]
b.append(b) # create a cycle

def callback(x):
    print("callback called!")

a_ref = weakref.ref(a, callback)

del a, b
gc.collect()
print("done collecting")
assert a_ref() is None
```

https://github.com/pytorch/pytorch/issues/75932 is a good read for more background; this issue is the same as that one except we also check weakrefs. The TLDR:

Today, PyObject preservation logic already has special handling for interaction with Python's garbage collector in two ways:
1) [In the original PR to add PyObject preservation](https://github.com/pytorch/pytorch/pull/56017),  @ezyang and @albanD foresaw that if gc detected a  unreachable cycle that included a ressurectable Tensor, `tp_clear` may be called on it, preventing a clean resurrection. The fix here is to tell `tp_traverse` to not traverse its members in the case we are resurrectable (thus preventing any cycle involving the resurrectable Tensor from being detected).
2) Later @Chillee [investigated](https://github.com/pytorch/pytorch/issues/75932) a separate case where the ressurectable Tensor is not directly part of the cycle, but is instead kept alive by a reference from a dead cycle. The `tp_traverse` fix did not address this because `tp_clear` will be called on all unreachable objects whether they are directly part of the cycle or not. The resolution here is to bail from `tp_clear` if the Tensor is resurrectable.

The `tp_clear` fix, however,  does not prevent weakrefs from already being cleared because gc clears weakrefs before calling `tp_clear`!

https://github.com/python/cpython/blob/e65a1eb93ae35f9fbab1508606e3fbc89123629f/Python/gc.c#L1725-L1744
https://github.com/python/cpython/blob/main/InternalDocs/garbage_collector.md#destroying-unreachable-objects

### Versions

main

cc @albanD

ethanclawsie:
- Introduce Python-exposed C++ type TensorImplWeakRef to track TensorImpl lifetime
- Add tensor_weakref.h/cpp implementing TensorImplWeakRef
- Register TensorImplWeakRef in Module.cpp
- Update _guards.py for TensorImplWeakRef string formatting
- Add tensor_weakref.cpp to build_variables.bzl for build integration

Fixes https://github.com/pytorch/pytorch/issues/145253

### DIFF
--- a/torch/_guards.py
+++ b/torch/_guards.py
@@ -26,6 +26,7 @@ from typing import (
 )
 
 import torch
+from torch._C import TensorImplWeakRef
 from torch.utils import _pytree as pytree
 from torch.utils._backport_slots import dataclass_slots
 from torch.utils._traceback import CapturedTraceback, format_frame
@@ -329,7 +330,7 @@ class Guard:
             obj_weakref = weakref.ref(obj)
             str(obj_weakref)  # raise error: KeyError: '__name__'
         """
-        if isinstance(obj_weakref, weakref.ReferenceType):
+        if isinstance(obj_weakref, (weakref.ReferenceType, TensorImplWeakRef)):
             obj = obj_weakref()
             if obj is not None:
                 return f"<weakref at {hex(id(obj_weakref))}; to '{obj.__class__.__name__}' at {hex(id(obj))}>"
--- a/torch/csrc/utils/tensor_weakref.cpp
+++ b/torch/csrc/utils/tensor_weakref.cpp
@@ -1,0 +1,158 @@
+#include <ATen/core/TensorBody.h>
+#include <Python.h>
+#include <c10/core/TensorImpl.h>
+#include <torch/csrc/autograd/python_variable.h>
+#include <torch/csrc/utils/tensor_weakref.h>
+
+typedef struct {
+  PyObject_HEAD
+  c10::weak_intrusive_ptr<c10::TensorImpl> weak;
+  Py_hash_t hash;
+} THPTensorImplWeakRef;
+
+static PyObject* THPTensorImplWeakRef_call(
+    THPTensorImplWeakRef* self,
+    PyObject*,
+    PyObject*) {
+  auto strong = self->weak.lock();
+  if (!strong) {
+    Py_RETURN_NONE;
+  }
+  at::Tensor t(std::move(strong));
+  return THPVariable_Wrap(t);
+}
+
+static Py_hash_t THPTensorImplWeakRef_hash(THPTensorImplWeakRef* self) {
+  if (self->hash != -1) {
+    return self->hash;
+  }
+  auto strong = self->weak.lock();
+  if (!strong) {
+    PyErr_SetString(PyExc_TypeError, "weak object has gone away");
+    return -1;
+  }
+  at::Tensor t(strong);
+  PyObject* py_tensor = THPVariable_Wrap(t);
+  if (!py_tensor) {
+    return -1;
+  }
+  self->hash = PyObject_Hash(py_tensor);
+  Py_DECREF(py_tensor);
+  return self->hash;
+}
+
+static PyObject* THPTensorImplWeakRef_richcmp(
+    PyObject* a,
+    PyObject* b,
+    int op) {
+  if (op != Py_EQ && op != Py_NE) {
+    Py_RETURN_NOTIMPLEMENTED;
+  }
+  auto* wa = reinterpret_cast<THPTensorImplWeakRef*>(a);
+  auto* wb = reinterpret_cast<THPTensorImplWeakRef*>(b);
+  auto aptr = wa->weak.lock();
+  auto bptr = wb->weak.lock();
+  if (!aptr || !bptr) {
+    int res = (a == b);
+    if (op == Py_NE)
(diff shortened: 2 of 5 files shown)


torch/csrc/utils/tensor_weakref.cpp
```diff
@@ -0,0 +1,124 @@
...
+  }
+}
+
+static int THPTensorImplWeakRef_bool(PyObject* self_) {
+  THPTensorImplWeakRef* self = reinterpret_cast<THPTensorImplWeakRef*>(self_);
+  c10::intrusive_ptr<c10::TensorImpl> strong = self->weak.lock();
+  return strong ? 1 : 0;
+}
+
+static PyType_Slot THPTensorImplWeakRef_slots[] = {
```
Skylion007, line 131:
```suggestion
constexpr static PyType_Slot THPTensorImplWeakRef_slots[] = {
```
making it constexpr allows the compiler to do static bounds checking

ethanclawsie: It decays to const PyType_Slot*, but PyType_Spec.slots expects a PyType_Slot* (non-const).

torch/csrc/utils/tensor_weakref.cpp
```diff
@@ -0,0 +1,124 @@
+#include "torch/csrc/utils/tensor_weakref.h"
+#include <ATen/core/TensorBody.h>
+#include <Python.h>
+#include <c10/core/TensorImpl.h>
+#include <torch/csrc/autograd/python_variable.h>
+
+typedef struct {
+  PyObject_HEAD
+  c10::TensorImpl* addr;
+  c10::weak_intrusive_ptr<c10::TensorImpl> weak;
```
ezyang, line 9: Why do you need two pointers here 🤔

torch/_guards.py
```diff
@@ -32,6 +32,11 @@
 from torch.utils.weak import WeakTensorKeyDictionary
 
 
+try:
+    from torch._C import TensorImplWeakRef
+except Exception:
+    TensorImplWeakRef = None
```
ezyang: LLM generated smell

torch/csrc/utils/tensor_weakref.cpp
swolchok: > making it constexpr allows the compiler to do static bounds checking

surely the size of a non-const array can't change either?

torch/csrc/utils/tensor_weakref.cpp
```diff
@@ -0,0 +1,124 @@
...
+    PyObject* args,
+    PyObject* kwargs) {
+  PyObject* py_tensor = nullptr;
+  if (!PyArg_ParseTuple(args, "O", &py_tensor))
+    return nullptr;
+  if (!THPVariable_Check(py_tensor)) {
+    PyErr_SetString(PyExc_TypeError, "expected a Tensor");
+    return nullptr;
+  }
+  at::Tensor t = THPVariable_Unpack(py_tensor);
```
swolchok: THPVariable_Unpack returns `const at::Tensor&` and py_tensor isn't going anywhere; why create a new Tensor?

torch/csrc/utils/tensor_weakref.cpp
```diff
@@ -0,0 +1,124 @@
...
+    PyObject* a,
+    PyObject* b,
+    int op) {
+  if (Py_TYPE(a) != Py_TYPE(b))
+    Py_RETURN_NOTIMPLEMENTED;
+  auto* wa = reinterpret_cast<THPTensorImplWeakRef*>(a);
+  auto* wb = reinterpret_cast<THPTensorImplWeakRef*>(b);
+  bool eq = (wa->addr == wb->addr);
+  if (op == Py_EQ)
+    return PyBool_FromLong(eq);
```
swolchok: please just put the braces in everywhere and always

torch/_guards.py
ethanclawsie: I agree this check is redundant as if there was issues with the import it would have gotten caught before this. I will remove this.

torch/csrc/utils/tensor_weakref.cpp
ethanclawsie: Good point, I will add a const.

torch/csrc/utils/tensor_weakref.cpp
ethanclawsie: I will remove the self->addr and modify the comparison and hashing logic to match CPython's weakref implementation.

ezyang:
This is likely obsoleted by https://github.com/pytorch/pytorch/pull/166342

PR #164273, opened 2025-09-30
[Inductor] Mutable custom op pattern matching and safety checker

karthickai:
* __->__ #164273

TL;DR
TorchInductor now supports pattern matching mutable custom ops directly by unwrapping auto_functionalized wrappers and inserting explicit dependency edges. This enables stable fusion patterns across PyTorch versions.

Problem:
 vLLM has mutable custom ops such as (`rms_norm`, `static_scaled_fp8_quant`) that require pattern matching for [fusion passes](https://github.com/vllm-project/vllm/blob/824a3f403f8cbb5b7984d33ca4741f67928c1d61/vllm/compilation/fusion.py#L122-L131). Currently they pattern match against `auto_functionalized(mutable_op)` wrappers, but vLLM is upgrading to `auto_functionalized_v2` (soon v3) with incompatible semantics that break existing patterns.
`auto_functionalized_v2` decomposes to: view + clone + functional_op + copy_. The specific view operations vary based on which inputs are mutated, making it difficult to write stable patterns that match view+op combinations.

Why current pattern matcher not support the raw custom mutating op ?
Consider this mutable op sequence:

```python
foo_inplace(x)    # Mutates tensor x
bar_out(x, out)   # Uses mutated x, produces out
```

FX Graph Representation:
```python
%x = placeholder()
%out = placeholder()
%foo_result = call_function(foo_inplace, (%x,))
%bar_result = call_function(bar_out, (%x, %out))  # Missing dependency!
```
There is no explicit edge from `foo_inplace` to `bar_out` even though `bar_out` depends on `foo_inplace` mutation. Without explicit edges, pattern matchers cannot reliably detect op sequences or ensure correct execution order.

High level idea: Wrap mutable op with control_deps op to preserve mutation ordering.
For each mutation:
1. Identifies all mutated args and their storage aliases
2. Finds views that need regeneration after mutation
3. Creates a subgraph that performs the mutation and regenerates views
4. Wraps the mutation with control_deps(deps_tuple, subgraph, *args)
5. Threads downstream users to use outputs from control_deps

We use a `control_deps` that explicitly captures mutation dependencies:
`control_deps(deps_tuple, subgraph, *args) -> tuple`

Parameters:
`deps_tuple`: tuple of tensors this operation depends on (for ordering)
`subgraph`: actual op to execute
`*args`: args to pass to the operation

Returns:
tuple containing: `(mutated_arg1, mutated_arg2, ..., view1, view2, ...)`
first N elements: mutated arguments (in the order they were mutated)
remaining elements: any views/aliases that need regeneration after mutation

Example:

Custom ops definitions
```python
@torch.library.custom_op("mylib::foo_inplace", mutates_args={"x"})
def foo_inplace(x: torch.Tensor) -> None:
    x.add_(1)

@torch.library.custom_op("mylib::bar_out", mutates_args={"out"})
def bar_out(x: torch.Tensor, out: torch.Tensor) -> None:
    out.copy_(x + 2)

@torch.library.custom_op("mylib::foobar_out", mutates_args={"x", "out"})
def foobar_out(x: torch.Tensor, out: torch.Tensor) -> None:
    x.add_(1)
    out.copy_(x + 2)

# pattern registration
def pattern(x, out):
    foo_inplace(x)
    bar_out(x, out)
    return x, out

def replacement(x, out):
    foobar_out(x, out)
    return x, out
```

Pattern graph after add_control_dep_edges (used for matching)
```python
    %arg0_1 : [num_users=1] = placeholder[target=arg0_1]
    %subgraph_foo_inplace_default : [num_users=1] = get_attr[target=subgraph_foo_inplace_default]
    %foo_inplace_default : [num_users=2] = call_function[target=torch.ops.higher_order.control_deps](args = ((%arg0_1,), %subgraph_foo_inplace_default, %arg0_1), kwargs = {})
    %full_default : [num_users=1] = call_function[target=torch.ops.aten.full.default](args = ([3], 0), kwargs = {dtype: torch.float32, layout: torch.strided, device: cuda:0, pin_memory: False})
    %subgraph_bar_out_default : [num_users=1] = get_attr[target=subgraph_bar_out_default]
    %bar_out_default : [num_users=1] = call_function[target=torch.ops.higher_order.control_deps](args = ((%full_default,), %subgraph_bar_out_default, %foo_inplace_default, %full_default), kwargs = {})
    %copy_ : [num_users=0] = call_function[target=torch.ops.aten.copy_.default](args = (%foo_inplace_default, %foo_inplace_default), kwargs = {})
    return (bar_out_default,)
```

after pattern matching
```python
    %arg0_1 : [num_users=2] = placeholder[target=arg0_1]
    %full_default : [num_users=2] = call_function[target=torch.ops.aten.full.default](args = ([3], 0), kwargs = {dtype: torch.float32, layout: torch.strided, device: cuda:0, pin_memory: False})
    %foobar_out_default : [num_users=0] = call_function[target=torch.ops.mylib.foobar_out.default](args = (%arg0_1, %full_default), kwargs = {})
    %copy_ : [num_users=0] = call_function[target=torch.ops.aten.copy_.default](args = (%arg0_1, %arg0_1), kwargs = {})
    return (full_default,)
```

Case 2: multi-mutable arg
```python
  @torch.library.custom_op("mylib::mutate_three", mutates_args={"x", "y", "z"})
  def mutate_three(x: torch.Tensor, y: torch.Tensor, z: torch.Tensor) -> None:
      x.add_(1)
      y.copy_(x)
      z.copy_(y + 1)

  @torch.library.custom_op("mylib::fused_three", mutates_args={"x", "y", "z"})
  def fused_three(x: torch.Tensor, y: torch.Tensor, z: torch.Tensor) -> None:
      x.add_(2)
      y.copy_(x + 1)
      z.copy_(y + 1)

  def pattern(x, y, z):
      mutate_three(x, y, z)
      return x, y, z

  def replacement(x, y, z):
      fused_three(x, y, z)
      return x, y, z
```

user graph
```python
  @torch.compile(fullgraph=True, backend=custom_backend_three)
  def fn(x, y, z):
      mutate_three(x, y, z)
      mutate_three(x, y, z)
      return x, y, z
```

after `decompose_auto_functionalized`
```python
graph():
    %arg0_1 : [num_users=3] = placeholder[target=arg0_1]
    %arg1_1 : [num_users=3] = placeholder[target=arg1_1]
    %arg2_1 : [num_users=3] = placeholder[target=arg2_1]
    %mutate_three_default_1 : [num_users=0] = call_function[target=torch.ops.mylib.mutate_three.default](args = (%arg0_1, %arg1_1, %arg2_1), kwargs = {})
    %mutate_three_default : [num_users=0] = call_function[target=torch.ops.mylib.mutate_three.default](args = (%arg0_1, %arg1_1, %arg2_1), kwargs = {})
    %copy_ : [num_users=0] = call_function[target=torch.ops.aten.copy_.default](args = (%arg0_1, %arg0_1), kwargs = {})
    %copy__1 : [num_users=0] = call_function[target=torch.ops.aten.copy_.default](args = (%arg1_1, %arg1_1), kwargs = {})
    %copy__2 : [num_users=0] = call_function[target=torch.ops.aten.copy_.default](args = (%arg2_1, %arg2_1), kwargs = {})
    return ()
```

after `add_control_dep_edges` (For pattern matching)
```python
graph():
    %arg0_1 : [num_users=3] = placeholder[target=arg0_1]
    %arg1_1 : [num_users=3] = placeholder[target=arg1_1]
    %arg2_1 : [num_users=3] = placeholder[target=arg2_1]
    %subgraph_mutate_three_default_1 : [num_users=1] = get_attr[target=subgraph_mutate_three_default_1]
    %mutate_three_default_1 : [num_users=3] = call_function[target=torch.ops.higher_order.control_deps](args = ((%arg0_1, %arg1_1, %arg2_1), %subgraph_mutate_three_default_1, %arg0_1, %arg1_1, %arg2_1), kwargs = {})
    %getitem_10 : [num_users=1] = call_function[target=operator.getitem](args = (%mutate_three_default_1, 2), kwargs = {})
    %getitem_9 : [num_users=1] = call_function[target=operator.getitem](args = (%mutate_three_default_1, 1), kwargs = {})
    %getitem_8 : [num_users=1] = call_function[target=operator.getitem](args = (%mutate_three_default_1, 0), kwargs = {})
    %subgraph_mutate_three_default : [num_users=1] = get_attr[target=subgraph_mutate_three_default]
    %mutate_three_default : [num_users=3] = call_function[target=torch.ops.higher_order.control_deps](args = ((%arg0_1, %arg1_1, %arg2_1), %subgraph_mutate_three_default, %getitem_8, %getitem_9, %getitem_10), kwargs = {})
    %getitem_13 : [num_users=1] = call_function[target=operator.getitem](args = (%mutate_three_default, 2), kwargs = {})
    %getitem_12 : [num_users=1] = call_function[target=operator.getitem](args = (%mutate_three_default, 1), kwargs = {})
    %getitem_11 : [num_users=1] = call_function[target=operator.getitem](args = (%mutate_three_default, 0), kwargs = {})
    %subgraph_copy_ : [num_users=1] = get_attr[target=subgraph_copy_]
    %copy_ : [num_users=0] = call_function[target=torch.ops.higher_order.control_deps](args = ((%arg0_1,), %subgraph_copy_, %getitem_11, %getitem_11), kwargs = {})
    %subgraph_copy__1 : [num_users=1] = get_attr[target=subgraph_copy__1]
    %copy__1 : [num_users=0] = call_function[target=torch.ops.higher_order.control_deps](args = ((%arg1_1,), %subgraph_copy__1, %getitem_12, %getitem_12), kwargs = {})
    %subgraph_copy__2 : [num_users=1] = get_attr[target=subgraph_copy__2]
    %copy__2 : [num_users=0] = call_function[target=torch.ops.higher_order.control_deps](args = ((%arg2_1,), %subgraph_copy__2, %getitem_13, %getitem_13), kwargs = {})
    return ()
```

after pattern matching
```python
graph():
    %arg0_1 : [num_users=2] = placeholder[target=arg0_1]
    %arg1_1 : [num_users=2] = placeholder[target=arg1_1]
    %arg2_1 : [num_users=2] = placeholder[target=arg2_1]
    %fused_three_default_1 : [num_users=0] = call_function[target=torch.ops.mylib.fused_three.default](args = (%arg0_1, %arg1_1, %arg2_1), kwargs = {})
    %fused_three_default : [num_users=0] = call_function[target=torch.ops.mylib.fused_three.default](args = (%arg0_1, %arg1_1, %arg2_1), kwargs = {})
    return ()
```

The `PatternSafetyChecker` validates pattern replacements to prevent violations and is opt-in via a flag when creating the `PatternMatcherPass(check_aliasing_safety=True)`.

Invariant 1: Escaping mutation invariance

The mutation state of escaping tensors (inputs and outputs) must remain unchanged. The escaping tensors are graph inputs (placeholders),  graph outputs and any tensor visible to the caller. If a pattern changes whether an escaping tensor gets mutated, the caller will see different behavior.

Example
```python
@torch.library.custom_op("mylib::add_one_inplace", mutates_args={"x"})
def add_one_inplace(x: torch.Tensor) -> None:
    x.add_(1)

def pattern(x):
    return x + 1

def replacement(x):
    add_one_inplace(x)  # new mutation on escaping input
    return x

@torch.compile(fullgraph=True, backend=custom_backend)
def func(x):
    y = x + 1  # pattern converts to x.add_(1)
    z = x + 2  # subsequent use of escaping input x
    return y, z
```

This is blocked because `x` is an escaping input original code: `x` is not mutated, `z = x + 2` after replacement: `x` is mutated, `z = (x+1) + 2`

Invariant 2: Subsequent use safety

Cannot introduce mutations on tensors that have subsequent uses (unless those uses already expected the mutated value). Mutating a tensor changes all future references to it through the same storage.

```python
@torch.library.custom_op("mylib::add_value_inplace", mutates_args={"x"})
def add_value_inplace(x: torch.Tensor, value: float) -> None:
    x.add_(value)

def pattern(x):
    return x + 5

def replacement(x):
    add_value_inplace(x, 5.0)
    return x

@torch.compile(fullgraph=True, backend=custom_backend)
def func(x):
    intermediate = x + 0  # creates new storage, non-escaping intermediate
    result1 = intermediate + 5  # pattern target on intermediate
    result2 = intermediate * 2  # subsequent use of intermediate
    return result1, result2

```

This is blocked because`intermediate` is used after the pattern (in `result2 = intermediate * 2`) original: `result2 = (x+0) * 2` after replacement with inplace: `result2 = (x+0+5) * 2` wrong.

below example is allowed

```python
def func(x):
    intermediate = x + 0
    result1 = intermediate + 5  # Pattern target, no subsequent use
    return result1
```

Invariant 3: Escaping Aliasing Invariance

Aliasing relationships between escaping tensors must be preserved. Three sub-rules are:

1.  cannot remove existing aliasing - if two escaping tensors aliased before, they must still alias
2.  cannot create new aliasing between two globally escaping tensors - don't make input alias output
3.  cannot create new aliasing that interacts with external mutations - new aliases could expose mutations

Aliasing affects memory semantics. If caller expects two separate tensors but we make them alias, mutations to one affect the other.

Example  - removes aliasing:

```python
# Pattern: x.clone()  ->  Replacement: x

def func(x):
    return x.clone()  # Creates new storage (no aliasing)
```

This is blocked because original: output does not alias input (separate storage) after replacement: output does alias input (same storage). 

Example -  creates aliasing with external mutation:

```python
# Pattern: y = x + 0  ->  Replacement: y = x

def func(x):
    b = x + 0  # Pattern target - creates new storage
    b.add_(3)  # External mutation (not in pattern)
    z = x + 4
    return z
```

This is blocked because original: `b` has separate storage, `b.add_(3)` doesn't affect `x` after replacement: `b` aliases `x`, so `b.add_(3)` mutates `x` result: `z = (x+3) + 4` instead of `x + 4`

Block multi-pass optimization that would create unsafe aliasing: Sequential pattern passes can create unsafe combinations even when each pass alone seems safe. For example: Pass 1's no-op removal is safe by itself, but it creates aliasing that makes Pass 2 unsafe. The safety checker blocks Pass 2 when it detects the mutation would affect the escaping input.

```python
# Pass 1: No-op removal
def pattern_a(x):
    y = x + 0
    return y

def replacement_a(x):
    y = x
    return y

# Pass 2: Inplace conversion
def pattern_b(y):
    z = y + 1
    return z

def replacement_b(y):
    z = y.add_(1)
    return z

patterns = PatternMatcherPass(check_aliasing_safety=True)

# Register both patterns
register_replacement(pattern_a, replacement_a, ...)  # No-op removal
register_replacement(pattern_b, replacement_b, ...)  # Inplace conversion

def fn(x):
    b = x + 0  # Pass 1 target
    c = b + 1  # Pass 2 target (after Pass 1)
    z = x + 4  # Subsequent use
    return c, z
```

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @mlazos @chenyang78 @choijon5


karthickai:
After discussing with @ProExpertProg, I reordered post_grad_mutable_custom_post_pass:
  1. post_grad_custom_pre_pass (before mutable pipeline - functional)
  2. post_grad_mutable_custom_post_pass (during mutable pipeline) - newly added
  3. post_grad_custom_post_pass (after mutable pipeline - functional)

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
...
+
+
+@torch.library.custom_op(
+    "pattern_matcher::op_for_dependencies", mutates_args=OrderedSet(["t"])
+)
+def op_for_dependencies(t: torch.Tensor) -> torch.Tensor:
+    """
+    A no-op function that serves to create dependencies in the FX graph.
+    """
+    return t
```
zou3519: custom operators generally aren't allowed to return outputs that are aliases of the inputs. Furthermore, even for built-in view operators, the operators are generally not allowed to return the input tensor directly. 

Can this just be a clone_preserving_strides(t)? I expect that we end up removing this custom operator eventually, so the extra clone shouldn't affect performance.

If we want to go the alias route, we should return t.alias() instead of t directly -- returning the tensor directly has also led to issues for other things.

eellison: or reason, to not just use the hop ?

karthickai: both options will work. Using t.clone() or t.alias() will not impact the existing flow. hop would be a good idea, but I would need to change the place where I detect the DEP_OP.

karthickai: update: I've changed it to t.clone()

zou3519: > or reason, to not just use the hop ?

you're suggesting using the control_deps HOP?

torch/_inductor/pattern_matcher.py
```diff
@@ -1223,7 +1223,16 @@ def run_node(self, node: torch.fx.Node) -> Any:
...
             ]
-            last_node = min(indices, key=operator.itemgetter(0))[1]
+            has_op_for_dep = any(
+                n.op == "call_function" and n.target == DEP_OP
+                for n in output_nodes
+                if isinstance(n, torch.fx.Node)
+            )
+            if has_op_for_dep:
+                # Todo: alternate approach graph walk and find proper index instead of max
+                last_node = max(indices, key=operator.itemgetter(0))[1]
```
mlazos: I still feel weird about doing max and min here, I feel like it's possible to end up with a malformed grapph without topological sorting ..

torch/_inductor/fx_passes/post_grad.py
```diff
@@ -1279,8 +1300,27 @@ def decomp(*flat_args):
             mode = args[0]
             return auto_functionalized_dense(mode, only_clone_these_tensors, **kwargs)
 
-        # pyrefly: ignore  # bad-argument-type
-        match.replace_by_example(decomp, flat_args, run_functional_passes=False)
+        if preserve_meta:
+            from torch._inductor.pattern_matcher import is_mutation_op
+
+            original_metadata = match.nodes[0].meta.copy()
+            nodes_before = OrderedSet([node.name for node in graph.nodes])
```
eellison: Not great that we need to do O(n) on every pattern replacement now. do we even need `mutation_region_id` if we have this mutable pattern matching?

torch/_inductor/fx_passes/post_grad.py
```diff
@@ -1246,14 +1266,15 @@ def decomp(*flat_args):
         raise AssertionError("triton_kernel_wrapper_functional was not removed")
 
 
-def decompose_auto_functionalized(graph):
+def decompose_auto_functionalized(graph, preserve_meta=False):
```
eellison: `preserve_meta` is a misleading name. we do preserve other meta. it's specifically related to mutation_region_id

torch/_inductor/fx_passes/post_grad.py
```diff
@@ -1325,8 +1365,27 @@ def decomp(*flat_args):
...
-        match.replace_by_example(decomp, flat_args, run_functional_passes=False)
+        if preserve_meta:
+            from torch._inductor.pattern_matcher import is_mutation_op
+
+            original_metadata = match.nodes[0].meta.copy()
+            nodes_before = OrderedSet([node.name for node in graph.nodes])
+
+            match.replace_by_example(decomp, flat_args, run_functional_passes=False)
+
+            for node in graph.nodes:
```
eellison: same comments here

torch/_inductor/fx_passes/post_grad.py
```diff
@@ -181,6 +181,26 @@ def post_grad_passes(gm: torch.fx.GraphModule, is_inference: bool):
...
+            decompose_auto_functionalized(gm.graph, preserve_meta=True)
+            add_implict_edges(gm)
+            gm.graph.lint()
+            mutable_custom_op_pass = True
+        GraphTransformObserver(
+            gm, "post_grad_mutable_custom_post_pass"
+        ).apply_graph_pass(post_grad_mutable_custom_post_pass)
+        if mutable_custom_op_pass:
+            remove_implict_edges(gm.graph)
+            gm = make_autofunctionalize(gm)
```
eellison: What are we doing here? we should not need to re-trace the entire graph. this is going to add a significant amount of compile time, in addition to destroying all of the existing meta. Some nodes are not make-fx traceable at this point in the graph as well.

torch/_inductor/fx_passes/post_grad.py
```diff
@@ -181,6 +181,26 @@ def post_grad_passes(gm: torch.fx.GraphModule, is_inference: bool):
...
         )
 
+    if post_grad_mutable_custom_post_pass := config.post_grad_mutable_custom_post_pass:
+        from torch._inductor.pattern_matcher import (
+            add_implict_edges,
+            check_mutable_custom_op,
+            remove_implict_edges,
+        )
+
+        mutable_custom_op_pass = False
```
eellison: Just because there are not existing mutable ops, does not mean that the user does not want to introduce them.

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
         assert isinstance(node.target, str)
         return _get_attr(node.graph.owning_module, node.target).__class__
     return node.target
+
+
+@torch.library.custom_op(
+    "pattern_matcher::op_for_dependencies", mutates_args=OrderedSet(["t"])
+)
+def op_for_dependencies(t: torch.Tensor) -> torch.Tensor:
```
eellison: can you say how this is different than the existing op for deps? and why we need both?

test/inductor/test_pattern_matcher.py
```diff
@@ -1799,6 +1799,208 @@ def f(x):
         self.assertEqual(len(sigmoid_nodes), 1)
         self.assertTrue("original_aten" in sigmoid_nodes[0].meta)
 
+    def test_mutable_op_nonview_inputs_register_replacement(self):
```
eellison: Mutable pattern matching extends beyond just mutating patterns. It also means that we need to make non mutable patterns safe wrt/mutation. I think this is basically just passes that introduce aliasing relationships where they did not exist previously. Previously, all mutation was on inputs of the graph, so it was not an issue. Now, post mutable ops, I think we need to be more careful for patterns that might do this. 

Can you add the following tests:
 
- we have a pattern that creates an aliasing relationship between input and output, but we fail to match, because adding the pattern would add an illegal mutation.

See
```
def foo(x):
     y = x.clone()
     y.add_(2)
     return y
```

Where the pattern would be to remove the clone. This is illegal because `x` escapes the graph.

Similarly

```
def foo():
     x = torch.rand([20])
     y = x.clone()
     y.add_(2)
     return x + y
```

This is illegal because x has a subsequent use in the graph.

torch/_inductor/pattern_matcher.py
eellison: why we are doing this ? i'm not sure i follow.

torch/_inductor/pattern_matcher.py
```diff
@@ -1223,7 +1223,16 @@ def run_node(self, node: torch.fx.Node) -> Any:
...
+            has_op_for_dep = any(
+                n.op == "call_function" and n.target == DEP_OP
+                for n in output_nodes
+                if isinstance(n, torch.fx.Node)
+            )
+            if has_op_for_dep:
+                # Todo: alternate approach graph walk and find proper index instead of max
+                last_node = max(indices, key=operator.itemgetter(0))[1]
+            else:
+                last_node = min(indices, key=operator.itemgetter(0))[1]
```
eellison: aren't these branches the same?

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
...
+
+    from .fx_passes.memory_estimator import GraphAliasTracker, StorageKey
+
+    # Ensure FakeTensor metadata exists for placeholders so alias tracking works
+    if not any(
+        node.meta.get("val") is not None
+        for node in gm.graph.nodes
+        if node.op == "placeholder"
+    ):
+        FakeTensorProp(gm).propagate()
```
eellison: You can use FakeTensorUpdator for this. we shouldn't need to re run the whole graph, this is slow.

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
...
+
+    nodes = list(gm.graph.nodes)
+    alias_info = GraphAliasTracker(nodes)
+
+    # Build mapping from storage -> nodes that produce that storage as outputs
+    storage_to_nodes: dict[StorageKey, OrderedSet[torch.fx.Node]] = {}
```
eellison: Isn't this `node_to_fresh_allocations` or `node_to_fresh_allocations` ? which already exist in the class ?

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
...
+    for n in nodes:
+        for sk in alias_info.node_to_output_storages.get(n, OrderedSet()):
+            storage_to_nodes.setdefault(sk, OrderedSet()).add(n)
+
+    def mutated_arg_nodes(node: torch.fx.Node) -> list[torch.fx.Node]:
+        if not is_mutation_op(node) or not hasattr(node.target, "_schema"):
+            return []
+
+        mutable_names, _ = get_mutable_args_from_schema(node.target._schema)
```
eellison: cc @zou3519 take a look ?

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
...
+        for idx, arg in enumerate(node.target._schema.arguments):
+            if arg.name in mutable_names:
+                val = (
+                    node.args[idx]
+                    if idx < len(node.args)
+                    else node.kwargs.get(arg.name)
+                )
+                if isinstance(val, torch.fx.Node):
+                    mutated.append(val)
+        return mutated
```
eellison: This seems very similar to code that already exists. can we refactor this ?

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
...
+
+    def topo_index() -> dict[torch.fx.Node, int]:
```
eellison: you may need to stable topological sort before this. the grpah is not guaranteed to be ordered at all points.

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
...
+
+        # For each mutated arg node, collect all storages it outputs
+        storage_to_mutated: dict[StorageKey, torch.fx.Node] = {}
+        for mutated in mutated_nodes:
+            for sk in alias_info.node_to_output_storages.get(mutated, OrderedSet()):
+                storage_to_mutated.setdefault(sk, mutated)
+        if not storage_to_mutated:
+            continue
+
+        order = topo_index()
```
eellison: we are calling topo_index (O(n)) in each node in the graph.. O(n^2)

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
...
+                    else node.kwargs.get(arg.name)
+                )
+                if isinstance(val, torch.fx.Node):
+                    mutated.append(val)
+        return mutated
+
+    def topo_index() -> dict[torch.fx.Node, int]:
+        return {n: i for i, n in enumerate(gm.graph.nodes)}
+
+    for node in list(gm.graph.nodes):
```
eellison, line 2566: nit: add comment we're mutating graph? thus list

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
...
+                if alias is dep_node:
+                    continue
+                for user in list(alias.users):
+                    if user in (node, dep_node):
+                        continue
+                    if order.get(user, -1) <= order[dep_node]:
+                        continue
+                    user.replace_input_with(alias, dep_node)
+
+    gm.recompile()
```
eellison: why do we need this?

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
...
+        if node.op == "call_function" and node.target == DEP_OP:
+            continue
+        mutated_nodes = mutated_arg_nodes(node)
+        if not mutated_nodes:
+            continue
+
+        # For each mutated arg node, collect all storages it outputs
+        storage_to_mutated: dict[StorageKey, torch.fx.Node] = {}
+        for mutated in mutated_nodes:
+            for sk in alias_info.node_to_output_storages.get(mutated, OrderedSet()):
```
eellison: nit: we should not need to construct an OrderedSet here if it is not in `node_to_output_storages`.

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
...
+
+    for node in list(gm.graph.nodes):
+        if node.op == "call_function" and node.target == DEP_OP:
+            continue
+        mutated_nodes = mutated_arg_nodes(node)
+        if not mutated_nodes:
+            continue
+
+        # For each mutated arg node, collect all storages it outputs
+        storage_to_mutated: dict[StorageKey, torch.fx.Node] = {}
```
eellison: do we have the invariant that all mutable ops will return the tensor it mutates? should we assert this ? Otherwise we are just skipping which could be incorrect

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
...
+        if not storage_to_mutated:
+            continue
+
+        order = topo_index()
+        insert_after: torch.fx.Node = node
+        for storage_key, mutated_tensor in storage_to_mutated.items():
+            dep_kwargs = {"writer_token": node}
+            with gm.graph.inserting_after(insert_after):
+                dep_node = gm.graph.call_function(
+                    DEP_OP, args=(mutated_tensor,), kwargs=dep_kwargs
```
eellison: Should other alias of the mutated tensor be an input as well ? what is preventing something from introducing a use on an alias, such that it reorders mutation?

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
...
+                )
+            insert_after = dep_node
+            if "val" in mutated_tensor.meta:
+                dep_node.meta["val"] = mutated_tensor.meta["val"]
+            dep_node.meta["mutated_of"] = node
+
+            storage_to_nodes.setdefault(storage_key, OrderedSet()).add(dep_node)
+
+            # Redirect later users of any alias of this storage to depend on dep_node
+            order = topo_index()
```
eellison: same thing

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
...
+            for alias in list(storage_to_nodes.get(storage_key, OrderedSet())):
```
eellison: nit: no need for duplicate `list`, and we shouldn't construct an OrderedSet on every iteration

torch/_inductor/pattern_matcher.py
```diff
@@ -2331,3 +2350,153 @@ def extract_target(node: torch.fx.Node) -> torch.fx.node.Target:
...
+
+def remove_implict_edges(graph: torch.fx.Graph) -> None:
+    """Remove implicit edges added by add_implict_edges"""
+    nodes_to_remove = []
+
+    for node in graph.nodes:
+        if (
+            node.op == "call_function"
+            and hasattr(node.target, "__name__")
+            and "op_for_dependencies" in str(node.target)
```
eellison: u should be able to check `node.target == ` directly

eellison (commented):
good first start, have some questions. would you mind printing out the graph/pattern at various stages, just for reviewers understanding?

test/inductor/test_pattern_matcher.py
karthickai: Thank you for your detailed feedback. After moved the mutable custom op pattern matching pass after the reinplace pass it restricted illegal mutatation, reinplace marks which tensors need cloning `only_clone_these_bases` and passes this info to `auto_functionalize_dense_`. which creates explicit clone nodes, this restricts the illegal mutation.

example search pattern (with clone):
```python
def f(x):
    y = x.clone()
    add_two_inplace(y)
    return y
```

replace pattern (without clone):
```python
def replacement(x):
    add_two_inplace(x)
    return x
```

Graph transformations:

1. registered pattern (`search_fn` after `add_implict_edges`):
``` python
graph():
    %x_1 : [num_users=1] = placeholder[target=x_1]
    %clone : [num_users=2] = call_function[target=torch.ops.aten.clone.default](args = (%x_1,), kwargs = {})
    %add_two_inplace : [num_users=1] = call_function[target=torch.ops.mylib.add_two_inplace.default](args = (%clone,), kwargs = {})
    %op_for_dependencies : [num_users=1] = call_function[target=torch.ops.pattern_matcher.op_for_dependencies](args = (%clone,), kwargs = {writer_token: %add_two_inplace})
    return op_for_dependencies
```

2. user function `f` before mutable custom op pass (after `reinplace`):
``` python
graph():
    %arg0_1 : [num_users=1] = placeholder[target=arg0_1]
    %auto_functionalized_v2 : [num_users=1] = call_function[target=torch.ops.higher_order.auto_functionalized_v2](args = (mylib.add_two_inplace.default,), kwargs = {_x_base_index: 0, _all_bases: [%arg0_1]})
    %getitem_1 : [num_users=1] = call_function[target=operator.getitem](args = (%auto_functionalized_v2, 1), kwargs = {})
    return (getitem_1,)
```

3. after `decompose_auto_functionalized` (unwraps `auto_functionalized` -> clone + mutation):
``` python
graph():
    %arg0_1 : [num_users=1] = placeholder[target=arg0_1]
    %as_strided_default : [num_users=1] = call_function[target=torch.ops.aten.as_strided.default](args = (%arg0_1, [3], [1], 0), kwargs = {})
    %clone_default : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%as_strided_default,), kwargs = {})
    %as_strided_default_1 : [num_users=2] = call_function[target=torch.ops.aten.as_strided.default](args = (%clone_default, [3], [1], 0), kwargs = {})
    %add_two_inplace_default : [num_users=0] = call_function[target=torch.ops.mylib.add_two_inplace.default](args = (%as_strided_default_1,), kwargs = {})
    return (as_strided_default_1,)
```

4. after `add_implict_edges` (inserts `DEP_OP` for ordering):
``` python
graph():
    %arg0_1 : [num_users=1] = placeholder[target=arg0_1]
    %as_strided_default : [num_users=1] = call_function[target=torch.ops.aten.as_strided.default](args = (%arg0_1, [3], [1], 0), kwargs = {})
    %clone_default : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%as_strided_default,), kwargs = {})
    %as_strided_default_1 : [num_users=2] = call_function[target=torch.ops.aten.as_strided.default](args = (%clone_default, [3], [1], 0), kwargs = {})
    %add_two_inplace_default : [num_users=1] = call_function[target=torch.ops.mylib.add_two_inplace.default](args = (%as_strided_default_1,), kwargs = {})
    %op_for_dependencies : [num_users=1] = call_function[target=torch.ops.pattern_matcher.op_for_dependencies](args = (%as_strided_default_1,), kwargs = {writer_token: %add_two_inplace_default})
    return (op_for_dependencies,)
```

the registered pattern expects clone -> add_two_inplace -> `DEP_OP,` but the user graph has as_strided(arg0_1) -> clone -> as_strided -> add_two_inplace -> `DEP_OP` due the structural mismatch, pattern matcher fails to match this pattern.

karthickai: I've added the test cases for above scenario `test_mutable_custom_op_pattern_matcher_unsafe_clone_removal`

torch/_inductor/pattern_matcher.py
karthickai: When a pattern has multiple output nodes, we must determine where to insert the replacement graph. The original code used `min` for non DEP_OP cases, but this is incorrect.

Using `min` causes "use before def" errors because the replacement graph requires all its outputs to be available as inputs. If we insert at `min` (the earliest output), later outputs may not yet exist.

Example malformed graph when using `min`:
```python
graph():
    %arg0_1 : [num_users=1] = placeholder[target=arg0_1]
    %arg1_1 : [num_users=1] = placeholder[target=arg1_1]
    %as_strided_default_2 : [num_users=1] = call_function[target=torch.ops.aten.as_strided.default](args = (%arg0_1, [3], [1], 0), kwargs = {})
    %clone_default_1 : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%as_strided_default_2,), kwargs = {})
    %as_strided_default_3 : [num_users=1] = call_function[target=torch.ops.aten.as_strided.default](args = (%clone_default_1, [3], [1], 0), kwargs = {})
    %foobar_out_default : [num_users=0] = call_function[target=torch.ops.mylib.foobar_out.default](args = (%as_strided_default_3, %as_strided_default_1), kwargs = {})
    %as_strided_default : [num_users=1] = call_function[target=torch.ops.aten.as_strided.default](args = (%arg1_1, [3], [1], 0), kwargs = {})
    %clone_default : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%as_strided_default,), kwargs = {})
    %as_strided_default_1 : [num_users=2] = call_function[target=torch.ops.aten.as_strided.default](args = (%clone_default, [3], [1], 0), kwargs = {})
    return (as_strided_default_1,)
```
In this example, `foobar_out` instered btw `as_strided_default_1` and `as_strided_default_3`. This violates the requirement that all nodes must be defined before use.

using `max`
```python
graph():
    %arg0_1 : [num_users=1] = placeholder[target=arg0_1]
    %arg1_1 : [num_users=1] = placeholder[target=arg1_1]
    %as_strided_default_2 : [num_users=1] = call_function[target=torch.ops.aten.as_strided.default](args = (%arg0_1, [3], [1], 0), kwargs = {})
    %clone_default_1 : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%as_strided_default_2,), kwargs = {})
    %as_strided_default_3 : [num_users=1] = call_function[target=torch.ops.aten.as_strided.default](args = (%clone_default_1, [3], [1], 0), kwargs = {})
    %as_strided_default : [num_users=1] = call_function[target=torch.ops.aten.as_strided.default](args = (%arg1_1, [3], [1], 0), kwargs = {})
    %clone_default : [num_users=1] = call_function[target=torch.ops.aten.clone.default](args = (%as_strided_default,), kwargs = {})
    %as_strided_default_1 : [num_users=2] = call_function[target=torch.ops.aten.as_strided.default](args = (%clone_default, [3], [1], 0), kwargs = {})
    %foobar_out_default : [num_users=0] = call_function[target=torch.ops.mylib.foobar_out.default](args = (%as_strided_default_3, %as_strided_default_1), kwargs = {})
    return (as_strided_default_1,)
```

I added `stable_topological_sort` before computing indices and simplified to always use max for multiple outputs and added comments explaining the logic.

torch/_inductor/fx_passes/post_grad.py
karthickai: `mutation_region_id` is mandatory for the pattern matcher. The `compute_mutation_region_ids` function assigns region IDs, and if `mutation_region_id` is different between nodes, the pattern won't match.

In our case, `mutation_region_id` is always zero because `auto_functionalized` ops inherit zero from previous passes. I added the `mutation_region_id `assignment logic in  `add_implict_edges` itself, so we don't need `preserve_meta` anymore.

Because we're adding `add_implict_edges` with `mutation_region_id=0` for all custom mutable ops, this won't impact us for pattern matching.

torch/_inductor/fx_passes/post_grad.py
karthickai: intially I tried use preserve all meta then realized only `mutation_region_id` is mandatory for pattern matcher. As mentioned above I removed the entire preserve_meta logic from `decompose_auto_functionalized`

torch/_inductor/fx_passes/post_grad.py
karthickai: as mentioned above I removed the `preserve_meta` logic from `decompose_auto_functionalized `

torch/_inductor/fx_passes/post_grad.py
karthickai: `make_autofunctionalize` is completely not needed. Initially, this pass ran before the `reinplace` pass, which is why I created `make_autofunctionalize`, now I've removed the `make_autofunctionalize` call and moved our pass to run after the `reinplace` pass. https://github.com/pytorch/pytorch/blob/a14ec2a79c60c6005e7094e74e798793d0d31ca8/torch/_inductor/fx_passes/post_grad.py#L287C1-L309C1

torch/_inductor/fx_passes/post_grad.py
karthickai: thanks for pointing out, I've removed the condition and always decompose and add edges, even if no existing mutable ops

torch/_inductor/pattern_matcher.py
karthickai: no, this is not the same branch. Previously, only `min` was used. I introduced `max` for `DEP_OP` and created a branch to preserve the existing flow. As mentioned above, I removed this branch and made `max` the default.

torch/_inductor/pattern_matcher.py
karthickai: I've added a comment explaining how it differs from the existing control_deps.

torch/_inductor/pattern_matcher.py
karthickai: thank you, I've updated with ` FakeTensorUpdater`

torch/_inductor/pattern_matcher.py
karthickai: yes, it was used two places in `auto_functionalize.py` so I've created a shared helper function `normalize_args_kwargs_by_schema` an refactored all occurrences to use it.

torch/_inductor/pattern_matcher.py
karthickai: I've added the `stable_topological_sort` before generating the index

torch/_inductor/pattern_matcher.py
karthickai: thanks for pointing this out. I've changed it from O(n^2) to O(n). The fix: call `stable_topological_sort()` once upfront, build the index mapping once, then update incrementally using fractional indices as we insert `DEP_OP` nodes.

torch/_inductor/pattern_matcher.py
karthickai: I've added this comment # Use list() to create a snapshot since we'll be mutating the graph by inserting

torch/_inductor/pattern_matcher.py
karthickai: yes it is not needed, I've removed it.

torch/_inductor/pattern_matcher.py
karthickai: I've fixed it.

torch/_inductor/pattern_matcher.py
karthickai: thanks, I've removed the `OrderedSet()` allocation.

torch/_inductor/pattern_matcher.py
karthickai: as mentioned above, I've fixed the topo_index call from O(n^2) to O(n)

torch/_inductor/pattern_matcher.py
karthickai: I've removed the OrderedSet() allocation.

torch/_inductor/pattern_matcher.py
karthickai: mutable custom ops don't return the mutated tensor, they return None. Looking at the graphs, all mutable ops have `num_users=0`. Since we can't track mutations through the return value, I added an assert to ensure we can track them through their mutated input args. If it's a mutation op, it must have identifiable mutated args. This catches cases where a mutable op's schema is inconsistent or missing mutation info.

torch/_inductor/pattern_matcher.py
karthickai: existing `node_to_fresh_allocations` maps nodes to the fresh storages they allocate. However, we need the reverse mapping from storage to all nodes that produce it as output. So I added `storage_to_producers` to the `GraphAliasTracker` class

torch/_inductor/pattern_matcher.py
karthickai: ideally yes, `DEP_OP` should depend on all aliases, but custom_op doesn't support variable-length args. Instead, at insertion time we redirect all users of all aliases to depend on the `DEP_OP` node.

eellison: Why do we need DEP_OP and control_deps. custom_ops do support lists of tnesors

karthickai: thank you, I've missed that, I will rewrite the `DEP_OP`

karthickai: I attempted to implement your suggestion to pass all aliases to `DEP_OP`  changed signature to `def op_for_dependencies(aliases: list[torch.Tensor])`and modified `add_implict_edges` to collect all aliases and pass them to `DEP_OP`. Pattern matching failed. I found that `gen_register_replacement` `fx_to_pattern(..., ignore_types=(list, ...))` lists are converted to `Ignored()` during pattern creation, preventing matching. I removed `list` from `ignore_types` pattern matching still failed. Initially suspected non-deterministic list order causing mismatches so added topological sorting, it didn't worked out. 

Finally I found out the fundamental issue, pattern registration traces a simple graph that produces args eg. `args: ([x_1],)`  - list length 1 user graphs go through `decompose_auto_functionalized` which creates multiple aliases (clone, as_strided, etc.), pattern matcher requires exact structural match including list lengths. Current single-tensor approach with user redirection provides equivalent ordering guarantees (all users of all aliases are redirected through `DEP_OP`). if you have ideas for making list-based matching work with unordered semantics, I'm happy to explore that

registered pattern
```python
graph():
    %x_1 : [num_users=2] = placeholder[target=x_1]
    %out_1 : [num_users=2] = placeholder[target=out_1]
    %foo_inplace : [num_users=1] = call_function[target=torch.ops.mylib.foo_inplace.default](args = (%x_1,), kwargs = {})
    %op_for_dependencies : [num_users=2] = call_function[target=torch.ops.pattern_matcher.op_for_dependencies](args = ([%x_1],), kwargs = {writer_token: %foo_inplace})
    %bar_out : [num_users=1] = call_function[target=torch.ops.mylib.bar_out.default](args = (%op_for_dependencies, %out_1), kwargs = {})
    %op_for_dependencies_1 : [num_users=1] = call_function[target=torch.ops.pattern_matcher.op_for_dependencies](args = ([%out_1],), kwargs = {writer_token: %bar_out})
    return (op_for_dependencies, op_for_dependencies_1)
```

user graph after decompose_auto_functionalized and add_implict_edges
```python
graph():
    %arg0_1 : [num_users=1] = placeholder[target=arg0_1]
    %arg1_1 : [num_users=1] = placeholder[target=arg1_1]
    %as_strided_default_2 : [num_users=1] = call_function[target=torch.ops.aten.as_strided.default](args = (%arg0_1, [3], [1], 0), kwargs = {})
    %clone_default_1 : [num_users=2] = call_function[target=torch.ops.aten.clone.default](args = (%as_strided_default_2,), kwargs = {})
    %as_strided_default_3 : [num_users=2] = call_function[target=torch.ops.aten.as_strided.default](args = (%clone_default_1, [3], [1], 0), kwargs = {})
    %foo_inplace_default : [num_users=1] = call_function[target=torch.ops.mylib.foo_inplace.default](args = (%as_strided_default_3,), kwargs = {})
    %op_for_dependencies : [num_users=1] = call_function[target=torch.ops.pattern_matcher.op_for_dependencies](args = ([%clone_default_1, %as_strided_default_3],), kwargs = {writer_token: %foo_inplace_default})
    %as_strided_default : [num_users=1] = call_function[target=torch.ops.aten.as_strided.default](args = (%arg1_1, [3], [1], 0), kwargs = {})
    %clone_default : [num_users=2] = call_function[target=torch.ops.aten.clone.default](args = (%as_strided_default,), kwargs = {})
    %as_strided_default_1 : [num_users=2] = call_function[target=torch.ops.aten.as_strided.default](args = (%clone_default, [3], [1], 0), kwargs = {})
    %bar_out_default : [num_users=1] = call_function[target=torch.ops.mylib.bar_out.default](args = (%op_for_dependencies, %as_strided_default_1), kwargs = {})
    %op_for_dependencies_1 : [num_users=1] = call_function[target=torch.ops.pattern_matcher.op_for_dependencies](args = ([%clone_default, %as_strided_default_1],), kwargs = {writer_token: %bar_out_default})
    return (op_for_dependencies_1,)
```

ProExpertProg:
Please see vllm-project/vllm#27072 for planned changes to the use of custom pattern matcher passes that might slightly modify the needed location for the mutable pattern matcher

mlazos (approved):
Karthick and I discussed offline in our 1:1: This looks good, but hold off from merging for now due to the ongoing discussion w/ VLLM

PR #164280, opened 2025-09-30
Add failing bitwise equivalence UT for aot_eager on rms_norm

ezyang:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.12.0) (oldest at bottom):
* #164289
* __->__ #164280

Signed-off-by: Edward Z. Yang <ezyang@meta.com>

PR #164289, opened 2025-09-30
Remove CompositeImplicitAutograd from _fused_rms_norm

ezyang:
* __->__ #164289
* #164280

The failure of bitwise equivalence is because aot_eager decomposes rms_norm. The decomposition occurs from Functionalize, because CompositeImplicitAutograd registrations will fill the Functionalize entry, which performs a decomposition (even though we had an explicit autograd formula and can go to the fused implementation directly).

This PR takes the approach of just removing the CompositeImplicitAutograd from the entry entirely, since it is not being used by out-of-tree backends (it was just a technical trick when fused RMS norm originally landed). An alternative approach would be to ensure we don't register CompositeImplicitAutograd to Functionalize, but this is tricky, as to register here or not needs to be done on a per-backend basis (since some backends DO decompose and we want to decompose in functionalize too), but we cannot do this because Functionalize is not a per-backend key.

Some more history: https://github.com/pytorch/pytorch/pull/103275

### DIFF
--- a/aten/src/ATen/native/native_functions.yaml
+++ b/aten/src/ATen/native/native_functions.yaml
@@ -3334,6 +3334,8 @@
     MPS: _fused_rms_norm_mps
     CompositeImplicitAutograd: rms_norm_composite
 
+- func: _rms_norm_composite(Tensor input, int[] normalized_shape, Tensor? weight, float? eps) -> (Tensor, Tensor)
+
 - func: _fused_rms_norm_backward(Tensor grad_out, Tensor input, int[] normalized_shape, Tensor rstd, Tensor? weight, bool[2] output_mask) -> (Tensor, Tensor)
   dispatch:
     CUDA: _fused_rms_norm_backward_cuda
--- a/aten/src/ATen/native/native_functions.yaml
+++ b/aten/src/ATen/native/native_functions.yaml
@@ -3332,7 +3332,7 @@
   dispatch:
     CUDA: _fused_rms_norm_cuda
     MPS: _fused_rms_norm_mps
-    CompositeImplicitAutograd: rms_norm_composite
+    CompositeExplicitAutograd: rms_norm_composite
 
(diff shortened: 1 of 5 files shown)


torch/_decomp/decompositions.py
```diff
@@ -1747,6 +1747,11 @@ def native_layer_norm_backward_out(
     return grad_input
 
 
+@register_decomposition(aten._fused_rms_norm.default)
+def _fused_rms_norm(input, normalized_shape, weight, float):
+    return torch.ops.aten.rms_norm_composite(input, normalized_shape, weight, float)
```
ezyang: Without this, Inductor cannot compile fused_rms_norm.

ngimel (approved):

Skylion007 (approved):

aten/src/ATen/native/native_functions.yaml
```diff
@@ -3332,7 +3332,9 @@
   dispatch:
     CUDA: _fused_rms_norm_cuda
     MPS: _fused_rms_norm_mps
-    CompositeImplicitAutograd: rms_norm_composite
+    CompositeExplicitAutograd: rms_norm_composite
+
+- func: _rms_norm_composite(Tensor input, int[] normalized_shape, Tensor? weight, float? eps) -> (Tensor, Tensor)
```
albanD: I would argue we don't need this one.
The only reason we kept CompositeImplicitAutograd above is to avoid churn for PT2 world when adding the new version.
But if we are ok with that, then just moving to CompositeExplicitAutograd is enough and we don't need to add this one.

Or do you add this just so you can call it from the decomp?

ezyang: this is dead now, cuz i ported the decomp to python LOL

albanD (approved):
Looks great! Thanks!

ezyang:
the tests should be easy to fix but I'm holding while I check on functionalization

PR #164308, opened 2025-09-30
[dynamo][sac] Support functools partial context_fn for sac

anijain2305:
* #164314
* __->__ #164308

Fixes https://github.com/pytorch/pytorch/issues/164300

### DIFF
--- a/torch/_dynamo/trace_rules.py
+++ b/torch/_dynamo/trace_rules.py
@@ -3450,6 +3450,7 @@ MOD_INLINELIST = [
     "torch.utils._foreach_utils",
     "torch.utils._python_dispatch",
     "torch.utils._pytree",
+    "torch.utils.checkpoint",
     "torch.utils.hooks",
 ]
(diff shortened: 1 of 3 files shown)


Lucaskabela (approved):
Generally LGTM, the test "failure" is actually because test succeeds now though so probably needs to be updated

torch/_dynamo/trace_rules.py
```diff
@@ -3450,6 +3450,7 @@ def _module_dir(m: types.ModuleType) -> Optional[str]:
     "torch.utils._foreach_utils",
     "torch.utils._python_dispatch",
     "torch.utils._pytree",
+    "torch.utils.checkpoint",
```
soulitzer: noob question: why do we also need to do this as part of the fix?

anijain2305: We don't need to ... just removed

soulitzer (approved):
Thanks!

PR #164309, opened 2025-09-30
Add python bindings for NCCL CTA policies

lakshayg:
NCCLConfig can now be constructed with non-default [cta policies][1]

```python
import torch
from torch.distributed import ProcessGroupNCCL as nccl

config = nccl.NCCLConfig()
config.cta_policy = nccl.NCCL_CTA_POLICY_ZERO  # NCCL version >= 2.28
```

[1]: https://docs.nvidia.com/deeplearning/nccl/archives/nccl_2283/user-guide/docs/api/flags.html#nccl-communicator-cta-policy-flags

### DIFF
--- a/torch/csrc/distributed/c10d/init.cpp
+++ b/torch/csrc/distributed/c10d/init.cpp
@@ -3366,6 +3366,22 @@ options :class:`~torch.distributed.ProcessGroupNCCL.Options`).
             return ::c10d::getNcclVersionTuple();
           });
 
+#ifdef NCCL_HAS_CTA_POLICY
+  processGroupNCCL
+#ifdef NCCL_CTA_POLICY_ZERO
+      // NCCL_CTA_POLICY_ZERO is available in NCCL versions >= 2.28
+      .def_property_readonly_static(
+          "NCCL_CTA_POLICY_ZERO",
+          [](py::object&) { return NCCL_CTA_POLICY_ZERO; })
+#endif // NCCL_CTA_POLICY_ZERO
+      .def_property_readonly_static(
+          "NCCL_CTA_POLICY_DEFAULT",
+          [](py::object&) { return NCCL_CTA_POLICY_DEFAULT; })
+      .def_property_readonly_static(
+          "NCCL_CTA_POLICY_EFFICIENCY",
+          [](py::object&) { return NCCL_CTA_POLICY_EFFICIENCY; });
+#endif // NCCL_HAS_CTA_POLICY
+
   module.def(
       "_get_intra_node_comm_usage_counter",
       &::c10d::intra_node_comm::getIntraNodeCommUsageCounter);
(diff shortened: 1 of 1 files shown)


torch/csrc/distributed/c10d/init.cpp
```diff
@@ -3366,6 +3366,22 @@ options :class:`~torch.distributed.ProcessGroupNCCL.Options`).
             return ::c10d::getNcclVersionTuple();
           });
 
+#ifdef NCCL_HAS_CTA_POLICY
+  processGroupNCCL
```
Skylion007: Sure but should this be an enum wrapper?

eqy: Looks like a typo?

Skylion007 (dismissed):

torch/csrc/distributed/c10d/init.cpp
lakshayg: These are defined in NCCL as preprocessor macros. I did not wrap it into an enum to avoid giving the false impression that this was an enum.

eqy: ah nvm, I see it's a nested ifdef, could the following condition be moved higher then?

torch/csrc/distributed/c10d/init.cpp
```diff
@@ -3366,6 +3366,20 @@ options :class:`~torch.distributed.ProcessGroupNCCL.Options`).
...
           });
 
+#ifdef NCCL_HAS_CTA_POLICY
+  processGroupNCCL.def_property_readonly_static(
+      "NCCL_CTA_POLICY_DEFAULT",
+      [](const py::object&) { return NCCL_CTA_POLICY_DEFAULT; });
+  processGroupNCCL.def_property_readonly_static(
+      "NCCL_CTA_POLICY_EFFICIENCY",
+      [](const py::object&) { return NCCL_CTA_POLICY_EFFICIENCY; });
+#ifdef NCCL_CTA_POLICY_ZERO // requires NCCL version >= 2.28
```
Skylion007, line 3376: Where is this defined? I don't see NCCL_CTA_POLICY_ZERO defined anywhere in PyTorch or the NCCL github. Might as well set it conditionally based on NCCL version

lakshayg: https://github.com/NVIDIA/nccl/blob/8d26308e6aba7f1667b24a861b5dc73f0f2e1f40/src/nccl.h.in#L66

eqy (approved):

torch/csrc/distributed/c10d/init.cpp
lakshayg: @Skylion007 Can this comment be resolved or would you prefer I change it to use NCCL version?

PR #164315, opened 2025-09-30
Add stack trace to kineto trace for aot_eager

yushangdi:
Current approach:

- we add a "stack_trace" metadata to the profiler at the end of each Interpreter run via `torch.autograd._add_metadata_json`.  The key is node name and the value is the stack trace
- After the stack trace is produced, we modify attach the stack trace information to it.

We can't directly modify the profiler output (instead of exporting to json and then loading it back),  because the profiler result is stored as a list of `KinetoEvent` and I don't see any method exposed to modify `KinetoEvent` in python. 

How to use it:

```
import torch
import torch.profiler
import json
from torch.fx.traceback import populate_stack_traces_to_kineto_trace

def fn(x):
    x = torch.relu(x) 
    return x + 2

compiled_fn = torch.compile(fn, backend="aot_eager")
x = torch.randn(10, 10, requires_grad=True)


with torch.profiler.profile(
    activities=[torch.profiler.ProfilerActivity.CPU],
    schedule=torch.profiler.schedule(wait=1, warmup=1, active=3, repeat=1),
) as prof:
    for step in range(6):
        y = compiled_fn(x)

        loss = y.sum()

        loss.backward()

        x.grad.zero_()

        prof.step()


prof.export_chrome_trace("trace.json")
populate_stack_traces_to_kineto_trace("trace.json")

```

### DIFF
--- a/torch/fx/traceback.py
+++ b/torch/fx/traceback.py
@@ -5,6 +5,7 @@ import traceback
 from contextlib import contextmanager
 from enum import Enum
 from typing import Any, Optional, Union
+import json 
 
 from torch._utils_internal import signpost_event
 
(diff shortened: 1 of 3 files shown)


torch/fx/traceback.py
```diff
@@ -396,3 +397,49 @@ def get_graph_provenance_json(graph: Graph) -> dict[str, Any]:
...
+        # Check if this is a user_annotation event
+        if event.get("cat") == "user_annotation":
+            event_name = event.get("name")
+            
+            # If the event name matches a node in node_stack_traces, attach the stack trace
+            if event_name in all_stack_traces:
+                event["args"]["stack_trace"] = all_stack_traces[event_name]
+    
+    if update_file:
+        json.dump(trace_data, open(file_name, 'w'))
```
ezyang, line 444: This will lose all newlines if you dump like this

ezyang:
@swolchok realized that because we're using fx Interpreter in aot_eager, this is a big slowdown for eager perf. So in "real" usage we must NOT use the interpreter. But that means this strategy will not work for accurate Kineto information. And you will also need to figure out how to do this *efficiently* without slowing things down too much (most importantly, when profiling is not enabled, but you also don't want to slow down too much when profiling is on either, to avoid CPU overhead distortion.)

SherlockNoMad:
> @swolchok realized that because we're using fx Interpreter in aot_eager, this is a big slowdown for eager perf. So in "real" usage we must NOT use the interpreter. But that means this strategy will not work for accurate Kineto information. And you will also need to figure out how to do this _efficiently_ without slowing things down too much (most importantly, when profiling is not enabled, but you also don't want to slow down too much when profiling is on either, to avoid CPU overhead distortion.)

@ezyang

yushangdi:
@ezyang Ok, so if can go with the codegen eager run, I have an alternative implementation here: https://github.com/pytorch/pytorch/pull/164640. There is a 50-100 us overhead of each aten call from my very preliminary experiment. Do you think that's too much overhead?

ezyang:
50us is a lot. But I see you are using the record function op; what about the fast binding instead?

PR #164318, opened 2025-10-01
[Inductor][Triton][FP8] Refactor scaled_mm template to accept scaling mode

jananisriram:
Summary: Refactor `scaled_mm` Inductor template to support template choice based on scaling mode. This modification sets up the infrastructure for adding new templates based on new scaling modes, such as deepseek-style scaling (a follow-up diff), as new scaling modes (deepseek, block, group) scale before the accumulation (as opposed to per-tensor and per-row scaling, which apply scaling after accumulation). This modification also further enables Inductor to infer a scaling type based on the shape of the scaling tensors, which makes existing infrastructure more extensible to new scaling modes.

Test Plan:
```
TORCHINDUCTOR_CACHE_DIR=~/personal/cache_dir_inductor CUDA_LAUNCH_BLOCKING=1 TORCH_USE_CUDA_DSA=1 TRITON_PRINT_AUTOTUNING=1 TRITON_ALWAYS_COMPILE=1 TORCH_LOGS=+inductor TORCHINDUCTOR_FORCE_DISABLE_CACHES=1 ENABLE_PERSISTENT_TMA_MATMUL=1 TORCHINDUCTOR_MAX_AUTOTUNE_GEMM=1 buck2 run mode/{opt,inplace} pytorch/tritonbench:run -- --op fp8_gemm --only torch_fp8_gemm,pt2_fp8_gemm --metrics tflops,accuracy --m 256 --n 768 --k 512 --output="/home/jananisriram/personal/random_bench.csv" --scaling_rowwise --atol=20 --rtol=2 2>&1 | tee ~/personal/random.log
```

Differential Revision: D83591083

### DIFF
--- a/torch/_inductor/kernel/mm.py
+++ b/torch/_inductor/kernel/mm.py
@@ -16,6 +16,7 @@ from torch._inductor.codegen.cpp_gemm_template import CppGemmTemplate
 from torch._inductor.remote_gemm_autotune_cache import gen_best_config
 from torch._inductor.virtualized import ops, V
 from torch.fx.experimental.proxy_tensor import make_fx
+from torch.nn.functional import ScalingType  # type: ignore[attr-defined]
 from torch.torch_version import TorchVersion
 
 from .. import config as inductor_config
@@ -372,15 +373,11 @@ persistent_tma_mm_template = TritonTemplate(
 
 load_scales = r"""
 @triton.jit
-def load_scales(a_scale_ptr, b_scale_ptr, SCALING_ROWWISE: tl.constexpr):
-    if SCALING_ROWWISE:
-        # For row-wise scaling, we'll return the pointers
-        return a_scale_ptr, b_scale_ptr
+def load_scales(scale_ptr, SCALE_RECIPE: tl.constexpr):
+    if SCALE_RECIPE == 0:
+        return tl.load(scale_ptr)  # For tensor-wise scaling, we'll load the scalar values
     else:
-        # For per-tensor scaling, we'll load the scalar values
-        a_scale = tl.load(a_scale_ptr)
-        b_scale = tl.load(b_scale_ptr)
-        return a_scale, b_scale
+        return scale_ptr  # For all other scaling recipes, we'll return the pointers
 """
 
 
@@ -390,7 +387,8 @@ def apply_scaling(
     accumulator,
     a_scale,
     b_scale,
-    SCALING_ROWWISE: tl.constexpr,
+    SCALE_RECIPE_A: tl.constexpr,
+    SCALE_RECIPE_B: tl.constexpr,
     offs_cm,
     offs_cn,
     M,
@@ -398,7 +396,7 @@ def apply_scaling(
     stride_a_scale_m,
     stride_b_scale_n,
 ):
-    if SCALING_ROWWISE:
+    if SCALE_RECIPE_A == 1 and SCALE_RECIPE_B == 1:  # (ScalingType.RowWise, ScalingType.RowWise)
         # For row-wise scaling, we need to load the scales for each row/column
         a_scales = tl.load(
             a_scale + (offs_cm * stride_a_scale_m),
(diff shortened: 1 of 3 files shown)


torch/_inductor/kernel/mm.py
```diff
@@ -421,7 +421,7 @@ def apply_scaling(
     stride_bk = {{stride("B", 0)}}
     stride_bn = {{stride("B", 1)}}
 
-    if SCALING_ROWWISE:
+    if SCALING_MODE == 1:
```
PaulZhang12: What is `SCALING_MODE == 1` indicate here?

torch/_inductor/kernel/mm.py
```diff
@@ -1341,8 +1341,27 @@ def tuned_scaled_mm(
...
+
+            def _is_scalar_like(sz: Any) -> bool:
+                return (len(sz) == 0) or all(
+                    V.graph.sizevars.statically_known_equals(d, 1) for d in sz
+                )
+
+            # Supported scaling modes (from aten/src/ATen/native/cuda/Blas.cpp::1163):
+            # TensorWise, RowWise
+            # TODO (jananisriram): support BlockWise1x16, BlockWise1x32, BlockWise1x128, BlockWise128x128
+            if _is_scalar_like(scale_a_size) and _is_scalar_like(scale_b_size):
```
PaulZhang12: Now that we will have different scaling modes, can we make this an enum?

torch/_inductor/kernel/mm.py
jananisriram: `SCALING_MODE` is a new template arg that replaces `SCALING_ROWWISE`; 0 --> tensor, 1 --> row, and up in the diff stack, 2 --> deepseek, etc.

I think it would make sense to make this clear as a comment somewhere, perhaps in `tuned_scaled_mm` (where it's calculated)? alternatively, making an enum (as discussed below) would be even clearer

torch/_inductor/kernel/mm.py
jananisriram: yeah this makes sense to me to avoid confusion

torch/_inductor/kernel/mm.py
```diff
@@ -378,15 +380,15 @@ def apply_scaling(
...
+    SCALING_MODE: tl.constexpr,
     offs_cm,
     offs_cn,
     M,
     N,
     stride_a_scale_m,
     stride_b_scale_n,
 ):
-    if SCALING_ROWWISE:
+    if SCALING_MODE == 1:
```
drisspg: can you pass Int enums into triton?

drisspg (approved):

torch/_inductor/kernel/mm.py
jananisriram: afaik you can't (which is why you have to manually check == 1 here)

jananisriram: just checked and it doesn't seem like we can!

torch/_inductor/kernel/mm.py
```diff
@@ -1253,6 +1256,45 @@ def tuned_sparse_semi_structured_mm(
     )
 
 
+# Supported scaling types (from aten/src/ATen/native/cuda/Blas.cpp::1163)
+# TODO (jananisriram): support BlockWise1x16, BlockWise1x32, BlockWise1x128, BlockWise128x128
+class ScalingType(IntEnum):
```
slayton58: Left this comment accidentally on a later review :( 

The same enum used in ATen is now available through `torch.nn.functional.ScalingType` -- it should be a drop-in change for your enum, and would keep us all on the same source-of-truth.

slayton58 (approved):
Looks good, minor nit on enum

torch/_inductor/kernel/mm.py
jananisriram: thanks, this is amazing!

torch/_inductor/kernel/mm.py
```diff
@@ -16,6 +16,7 @@
 from torch._inductor.remote_gemm_autotune_cache import gen_best_config
 from torch._inductor.virtualized import V
 from torch.fx.experimental.proxy_tensor import make_fx
+from torch.nn.functional import ScalingType
```
PaulZhang12: I dont think this import is right

jananisriram: this should (?) be fixed with a rebase, trying that now

PR #164319, opened 2025-10-01
[vllm hash update] update the pinned vllm hash

pytorchupdatebot:
This PR is auto-generated nightly by [this action](https://github.com/pytorch/pytorch/blob/main/.github/workflows/nightly.yml).
Update the pinned vllm hash.


huydhn:
This current commit is botched because of https://github.com/vllm-project/vllm/pull/25896, for example https://github.com/pytorch/pytorch/actions/runs/18209207310/job/51852649134?pr=164319, need to update the pinned commit again to include the fix at https://github.com/vllm-project/vllm/pull/26138

PR #164324, opened 2025-10-01
[atomically_apply_size_hint] Make unbacked replacements reconciles to a single expr

ColinPeppler:
## Problem
Okay there's limitations with today's `atomically_apply_size_hint` though it works for most observed failures we've seen so far. However, it's easy to come up with an edge case.

Suppose you encounter this setup.
```
a: [s0 + u0]
b: [s1 + u1]
c: [u2 + u3]
d: [u100]
```

Today, we use a few heuristics to specify the LHS and RHS for replacements.

https://github.com/pytorch/pytorch/blob/10d2734d9b9aee767dcecaf2175d9f231810f4be/torch/_inductor/sizevars.py#L730-L759

It's possible to end up with these replacement rules. Notice how there's no replacement for `s1 + u1` and `u2 + u3` :( That's because today picking the LHS and RHS matters a lot, and `s1 + u1` & `u2 + u3` happened to end up on the RHS.
```
s0 + u0 => s1 + u1
s0 + u0 => u2 + u3         # overrides previous replacement; each expr only gets one replacement
s0 + u0 => u100            # overrides previous replacement; ditto
```

I believe what we really want is this: everybody gets a replacement! And they all should (eventually) settle at the same canonical expr (i.e. `u100`) when running the replacement several times.
```
s1 + u1 ==> s0 + u0
u2 + u3 ==> s0 + u0
s0 + u0 ==> u100
```

We can just short-cut this by using the canonical expr as the replacement.
```
s1 + u1 ==> u100
u2 + u3 ==> u100
s0 + u0 ==> u100
```

## Implementation

I offer one way to deal with this:
1. assure every expression has one canonical replacement (i.e. `u100`)
2. if two expressions are equal (inferred from `deferred_runtime_asserts`), then they must have the same canonical replacement

 We can implement the above with union find.
* Whenever you see `Eq(lhs, rhs)` then do `union(lhs, rhs)`.
* Whenever you want to find the canonical replacement for a given expr then do `find(expr)`.
* When picking the canonical replacement we can use a few heuristics like (1) prefer a fully backed expr, (2) replacing with sub-expressions, and whatever we'd like.

* __->__ #164324

Differential Revision: [D84549260](https://our.internmc.facebook.com/intern/diff/D84549260)


laithsakka:
I need to chat about this with you to understand it.

test/inductor/test_aot_inductor.py
```diff
@@ -1661,6 +1661,82 @@ def forward(self, x, y):
...
+    @config.patch({"unbacked_symint_fallback": 12})
+    @config.patch({"triton.autotune_at_compile_time": None})
+    def test_unbacked_expr_replacements(self, shift_k, use_static_size):
+        """
+        Test parameters
+        - shift_k: Validates that torch._check assertion order doesn't affect
+        results by shifting the order of torch._checks
+        - use_static_size: Tests torch._check compatibility between unbacked
+        symbolic expressions and static shapes
+        """
```
ColinPeppler, line 1674: see this gist for example torch._check loggings for these parameters
https://gist.github.com/ColinPeppler/6a28d199b5740441e82a81d840ddbfa7

torch/_inductor/sizevars.py
```diff
@@ -739,47 +740,110 @@ def _get_unbacked_replacements(self) -> dict[Expr, Expr]:
...
         Then it can substitute s0 + u0 for u1.
         """
+
         if self.unbacked_replacements is not None:
             return self.unbacked_replacements
 
-        def should_keep_src_dst(lhs: Expr, rhs: Expr):
-            # assuming lhs is the expr to be replaced (src), rhs is the replacement (dst)
-            # checking if we should keep them for the replacement rule or swap
+        class UnionFind:
```
laithsakka: can you add a comment describing what this is doing you can use AI to generate it.

laithsakka (approved):

laithsakka (approved):

Issue #162858, opened 2025-09-12
[dynamo] Add most recent bytecode/Python stack context to graph break developer error context


williamwen42:
Log the last k (k = 50?) bytecode instructions traced, along with the state of the symbolic Python stack.

i.e. print the last k outputs of `TORCH_LOGS=trace_bytecode` when logging a graph break.

For example,
```python
import torch


@torch.compile(backend="eager")
def fn(x):
    y = x + 1
    z = x + y
    torch._dynamo.graph_break()
    return z


fn(torch.ones(3))
```

Would have these instructions logged into the graph break message:
```
Graph break in user code at /data/users/williamwen/pytorch/playground3.py:8
Graph Break Reason: Call to `torch._dynamo.graph_break()`
  Explanation: User-inserted graph break. Message: None
  Hint: Remove the `torch._dynamo.graph_break()` call.

  Developer debug context: Called `torch._dynamo.graph_break()` with args `[]`, kwargs `{}`
  Most recent 50 bytecode instructions traced:
    RESUME 0 []
    LOAD_FAST x []
    LOAD_CONST 1 [LazyVariableTracker()]
    BINARY_OP 0 [LazyVariableTracker(), ConstantVariable(int: 1)]
    STORE_FAST y [TensorVariable()]
    LOAD_FAST x []
    LOAD_FAST y [TensorVariable()]
    BINARY_OP 0 [TensorVariable(), TensorVariable()]
    STORE_FAST z [TensorVariable()]
    LOAD_GLOBAL torch []
    LOAD_ATTR _dynamo [LazyVariableTracker()]
    LOAD_ATTR graph_break [LazyVariableTracker()]
    CALL 0 [LazyVariableTracker(), NullVariable]

 For more details about this graph break, please visit: https://meta-pytorch.github.io/compile-graph-break-site/gb/gb0025.html
User code traceback:
  File "/data/users/williamwen/pytorch/playground3.py", line 12, in <module>
    fn(torch.ones(3))
  File "/data/users/williamwen/pytorch/playground3.py", line 8, in fn
    torch._dynamo.graph_break()
```

Including bytecode source attribution to the logged VariableTrackers (https://github.com/pytorch/pytorch/issues/162857) would make this debug information even more helpful.

[...]

PR #164326, opened 2025-10-01
[Dynamo] Adding latest bytecode for user calling graph break

fxdawnn:
#162858 The issue described the feature implemented.

This adds to the existing graph break log with the latest 20 (or viable user frame) bytecode instructions. The scenario is when the graph_break happens without errors. It happens during the case when user calling torch._dynamo.graph_break().

Meanwhile, in the testing, one can find that the generated frame based on step() is not deterministic as sometimes it reached the maximum amount, sometimes it generated the less than that. The bytecode generation is python version dependent. Thus, the testing plan excludes the bytecode output but generated the total bytecode line count.

This is a helpful process to understand bytecode transformation, symbolic convert, and convert frame. It is a helpful task to provide hands-on experience with dynamo workflow.

### DIFF
--- a/test/dynamo/test_error_messages.py
+++ b/test/dynamo/test_error_messages.py
@@ -773,6 +773,8 @@ from user code:
             return f
 
         def post_munge(s):
+            s = re.sub(r"TRACE.*\n", "", s, flags=re.MULTILINE)
+            s = re.sub(r"\nTRACE.*", "", s)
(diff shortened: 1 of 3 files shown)


fxdawnn:
/easycla re-run this again after signing up

fxdawnn:
easycla

fxdawnn:
\EasyCLA

PR #164329, opened 2025-10-01
[PP]: Optimize memory by early releasing stage inputs' gradients

mingyuanw-mt:
Seems that we can release input activations' gradients early in `stage_backward()` in PP, which helps to reduce the peak memory.  
  
I tested this using `1F1B` and `Interleaved1F1B` PP strategy (for simplicity, I use 4 decoder layers of llama3, set PP size to 2 and set num_microbatches to 128)  based on torchtitan
run command using torchtitan: 
```bash
CUDA_VISIBLE_DEVICES=4,5 LOG_RANK=0,1 NGPU=2 CONFIG_FILE=./torchtitan/models/llama3/train_configs/llama3_8b.toml ./run_train.sh --metrics.log_freq 1  --training.seq_len 8192 --training.steps 10 --parallelism.data_parallel_shard_degree 1 --activation_checkpoint.mode full --model.tokenizer_path /workspace/torchtitan-v0.1.0/torchtitan/torchtitan/datasets/tokenizer/original/tokenizer.model --tr
aining.dataset wikipedia  --parallelism.pipeline_parallel_degree 2  --training.local_batch_size 128 --parallelism.pipeline_parallel_microbatch_size 1 --training.dataset_path /workspace/wikipedia_subset --training.seed 42 --parallelism.pipeline_parallel_schedule 1F1B
```  
### before fix
b8e281cce1dac15e827c216e7d83f402

### after fix
70d5ceba311a8398d041189bf8897cfc
  
after fix, the memory usage on rank1, i.e., non first stages saving 6.9GB compare to before fix. the memory usage on rank0 remains unchanged (rank0 represents stage0)
  
### before fix
a28b7f9704b9234870619c43194e8a72

### after fix
2d8d6d956b72885186f8c7059146c41a
  
after fix, the memory usage on rank1 saving 14.57GB (rank1 holds layer1 and layer3) and rank0 saving 7.5GB (rank0 holds layer0 and layer2)  
  
## Memory snapshot results
also, I have dumped the memory snapshot to observe the memory under the 1F1B PP strategy.  

### before fix
6fd4e4ba82b8bacf9ca6edee4f3d5581
  
we can see the memory is increasing as pp step_microbatches running. (the lifetime of input activation's gradient, i.e., the output of `FusedRMSNormBackward`  lasts too long)

### after fix
2e415f25af6750d06e5e647683b212b9
  
after fix, we got more steady memory usage during training. (the input activation's gradient will be released or return allocator soon)

### DIFF
--- a/torch/distributed/pipelining/_backward.py
+++ b/torch/distributed/pipelining/_backward.py
@@ -366,6 +366,13 @@ def stage_backward(
         for val in input_values:
             if isinstance(val, torch.Tensor):
                 grad_inputs.append(val.grad)
+                # Since gradients that will pass back to previous stages do not require gradient accumulation,
+                # by decrementing the gradients' reference count at this point, the memory of gradients will be
+                # returned to the allocator as soon as the next micro batch's get_bwd_send_ops comes and current
+                # asynchronous send completes.
+                # This prevents the gradients from persisting in GPU memory for the entire duration of step_microbatches
+                # until clear_runtime_states() is called.
+                val.grad = None
             else:
                 grad_inputs.append(None)
 


torch/distributed/pipelining/_backward.py
```diff
@@ -366,6 +366,13 @@ def extract_tensors_with_grads(
         for val in input_values:
             if isinstance(val, torch.Tensor):
                 grad_inputs.append(val.grad)
+                # Since gradients that will pass back to previous stages do not require gradient accumulation,
+                # by decrementing the gradients' reference count at this point, the memory of gradients will be
+                # returned to the allocator as soon as the next micro batch's get_bwd_send_ops comes and current
+                # asynchronous send completes.
+                # This prevents the gradients from persisting in GPU memory for the entire duration of step_microbatches
+                # until clear_runtime_states() is called.
+                val.grad = None
```
Skylion007, line 375: You can do this, but we should think about the control flow about when we delay the gradients calculation of the variables to the end of the pipeline stage. i am not familiar with PyTorch's pipeling code path as I am with Megatron so this case might already be accounted for though.

mingyuanw-mt: > You can do this, but we should think about the control flow about when we delay the gradients calculation of the variables to the end of the pipeline stage. i am not familiar with PyTorch's pipeling code path as I am with Megatron so this case might already be accounted for though.

Thank you for raising the point about delaying the gradients calculation. IIUC, in PyTorch’s implementation, the gradients of stage inputs are also saved in `self.backward_state` or `self.bwd_cache`  which are then sent to other rank via  `get_bwd_send_ops`, and I haven't come across any part of the PyTorch implementation that explicitly delays gradient calculation. So it should be safe to set `val.grad = None`  to free memory without affecting the pipeline flow currently.  
  
@kwen2501 @H-Huang  could you please help to take a look ?

H-Huang, line 375: Very solid work and investigation! By chance, do you have any ideas to test memory is released as expected via a unit test? Want to make sure we don't regress in the future, otherwise I am happy to land this

mingyuanw-mt: Thanks, I'll try to add a unit test soon

H-Huang: Thanks! You can follow the template for a unit test here: https://github.com/pytorch/pytorch/blob/0d39ecb2ce8556e85343d8da0c87450192c2fdf8/test/distributed/pipelining/test_schedule_multiproc.py#L749-L812

Run with `python test/distributed/pipelining/test_schedule_multiproc.py -k test_zero_bubble_with_model_kwargs`

Going to accept your PR and land it for now

H-Huang (approved):

PR #164332, opened 2025-10-01
Add scaffolding for StableIValue FC/BC (no PoC)

mikaylagawarecki:
1. Add `extension_build_version` and `is_internal` to `FromImpl`/`ToImpl` (this will be useful for future if we need to break the BC of any type) #163832 has the PoC of how we would actually use this system
2. Add `aoti_torch_library_impl_v2` that takes in an additional `extension_build_version` argument, updates callsite in `torch/csrc/stable/library.h` to always pass `TORCH_ABI_VERSION` for this argument
3. Add `extension_build_version` to `from_ivalue` and `to_ivalue` and update all callsites
4. Add a private `_from` and `_to` that pass `is_internal=True` to `FromImpl`/`ToImpl`, making it easier to reason about what is being called from libtorch-land / extension-land

**Note: This PR does not include a linter that tells the user to update from/to if changing the ABI of a type in headeronly, which I intend to do in https://github.com/pytorch/pytorch/pull/163998**

* #165285
* #165284
* #165283
* __->__ #164332
* #163683
* #166373
* #164356


janeyx99:
For both this PR and the scaffolding PR before, we should have tests for the use cases we expect people to do (feel free for these to be contrived, but these are the tests we should have in CI to prevent longterm regressions):

A. User builds on 2.10 targeting version 2.9 + runs on 2.9  -> should work amidst new ops/SIV changes
B. User builds on 2.10 targeting version 2.9 + runs on 2.10  -> should work amidst new ops/SIV changes
C. User builds on 2.10 targeting version 2.10 + runs on 2.10 -> should work
D. User builds on 2.10 targeting version 2.10 + runs on 2.9 -> not guaranteed to work, we can have a test case that it fails smoothly during runtime 
E. User builds on 2.9 targeting version 2.9 + runs on 2.9 -> should work
F. User builds on 2.9 targeting version 2.9 + runs on 2.10 -> should work amidst new ops/SIV changes
G. User builds on 2.9 targeting version 2.10 -> build time error?
H. User builds on 2.10 targeting 2.9 but uses something in 2.10 -> build time error

These tests would make it easier to review the current code/hopefully make us feel confident we're doing the right thing. Might be cases I missed (if we had three versions, we could get spicy with building on X+1, targeting X, running on X+2).

janeyx99 (approved):

Issue #131911, opened 2024-07-26
Guard failing with `.item()` indexing


mitkotak:
### 🐛 Describe the bug

```python
import torch

torch._dynamo.config.capture_scalar_outputs = True

def my_arithmetic(a, b):
    wrk = torch.zeros(a.size(0))
    for i in range(a.size(0)):
        idx = b[i].item()
        torch._check_is_size(idx)
        wrk[idx] += 1

    return wrk

model = torch.compile(my_arithmetic, fullgraph=True, disable=False)
my_a = torch.randn([9])
my_b = torch.ones(9, dtype=torch.int32)
model(my_a, my_b)
```

```bash
E0726 12:48:36.644000 140715036880896 torch/fx/experimental/recording.py:281] [0/0] failed while running evaluate_expr(*(u0 >= 9, None), **{'fx_node': None})
E0726 12:48:36.645000 140715036880896 torch/_subclasses/fake_tensor.py:1761] [0/0] failed while attempting to run meta for aten.select.int
E0726 12:48:36.645000 140715036880896 torch/_subclasses/fake_tensor.py:1761] [0/0] Traceback (most recent call last):
E0726 12:48:36.645000 140715036880896 torch/_subclasses/fake_tensor.py:1761] [0/0]   File "/home/mkotak/atomic_architects/projects/e3nn_simple_network/.venv/lib/python3.10/site-packages/torch/_subclasses/fake_tensor.py", line 1757, in _dispatch_impl
E0726 12:48:36.645000 140715036880896 torch/_subclasses/fake_tensor.py:1761] [0/0]     r = func(*args, **kwargs)
E0726 12:48:36.645000 140715036880896 torch/_subclasses/fake_tensor.py:1761] [0/0]   File "/home/mkotak/atomic_architects/projects/e3nn_simple_network/.venv/lib/python3.10/site-packages/torch/_ops.py", line 667, in __call__
E0726 12:48:36.645000 140715036880896 torch/_subclasses/fake_tensor.py:1761] [0/0]     return self_._op(*args, **kwargs)
E0726 12:48:36.645000 140715036880896 torch/_subclasses/fake_tensor.py:1761] [0/0]   File "/home/mkotak/atomic_architects/projects/e3nn_simple_network/.venv/lib/python3.10/site-packages/torch/_meta_registrations.py", line 4788, in meta_select
[...]

Issue #114293, opened 2023-11-21
Recompilation triggered at each step of the loop involving array indexing


yaroslavvb:
### 🐛 Describe the bug

Running the following in today's nightly (`2.2.0.dev20231121+cu121`) I see recompilation issued at each step of the loop. The shapes are static, is there a way to avoid recompilation? (parent [discussion](https://discuss.pytorch.org/t/compiling-for-loop-makes-it-run-25x-slower-than-uncompiled-version/191488))

```
import torch
torch._logging.set_logs(recompiles=True)

@torch.compile
def accumulate(X0, start, end):
    result = X0[start]
    for i in range(start+1, end):
        result += X0[i]
    return result


X0 = torch.randn(100, 100)

for i in range(10):
    print(i)
    accumulate(X0, i, i+10)
```

[colab](https://colab.research.google.com/drive/11OGjZmy8zeZpT0m0irkVCxzPY0st2dGD#scrollTo=8KmAKFH6Px9Q)

cc @chauhang @penguinwu @ezyang @bobrenjc93 @msaroufim @wconstab @bdhirsh @anijain2305 @zou3519

jon-chuang:
Hmm, could you show `TORCH_LOGS=recompiles`?

yaroslavvb:
```
0
[2023-11-21 19:02:13,073] torch._dynamo.guards.__recompiles: [DEBUG] Recompiling function accumulate in <ipython-input-6-4764417fa0ae>:6
[2023-11-21 19:02:13,073] torch._dynamo.guards.__recompiles: [DEBUG]     triggered by the following guard failure(s):
[2023-11-21 19:02:13,073] torch._dynamo.guards.__recompiles: [DEBUG]     - L['end'] == 10                                                # for i in range(start+1, end):  # <ipython-input-6-4764417fa0ae>:9 in accumulate
1
[2023-11-21 19:02:15,412] torch._dynamo.guards.__recompiles: [DEBUG] Recompiling function accumulate in <ipython-input-6-4764417fa0ae>:6
[2023-11-21 19:02:15,412] torch._dynamo.guards.__recompiles: [DEBUG]     triggered by the following guard failure(s):
[2023-11-21 19:02:15,412] torch._dynamo.guards.__recompiles: [DEBUG]     - L['start'] == 1                                               # result = X0[start]  # <ipython-input-6-4764417fa0ae>:8 in accumulate
[...]

PR #164341, opened 2025-10-01
Fix issues with generalized_scatter  and setitem allocated unbacked symbols.

laithsakka:
* #164948
* __->__ #164341

Three fixes:
1. When doing t[u0] +=1  if u0 is unbacked we could allocate a new unbacked symbol during the the indexing of t[u0] (when we fake trace setitem), namely because meta_select does allocate a new unbacked symbol for the storage offset when we do not know if u0>=0 or u0<0.  but the output size/stride of setitem(), does not depend on that new symbol. it's self consumed in setitem so we shall ignore it. 

2. Also when we trace through generalized_scatter the applications of the views could allocate unbacked symints
but those do not effect final output, we also shall ignore them.

3.Before accessing strides in lowering we shall materialize. 

Address  https://github.com/pytorch/pytorch/issues/114293 and https://github.com/pytorch/pytorch/issues/131911


torch/_inductor/lowering.py
```diff
@@ -3141,8 +3142,14 @@ def select_scatter(x, src, dim: int, index: int):
     assert x.get_dtype() == src.get_dtype()
     x_loader = x.make_loader()
     dim = _validate_dim(x, dim, 0)
-    if V.graph.sizevars.evaluate_expr(sympy.Lt(index, 0)):
+    if V.graph.sizevars.guard_or_false(sympy.Lt(index, 0)):
         index = index + x.get_size()[dim]
+    elif V.graph.sizevars.guard_or_false(sympy.Ge(index, 0)):
+        pass
+    else:
+        # unabacked index
```
laithsakka: unabacked->unbacked.

torch/_dynamo/variables/tensor.py
```diff
@@ -1096,7 +1097,18 @@ def method___setitem__(self, key, value):
             #   value.requires_grad is True => self.has_grad_fn becomes True
 
             # Not sure if __setitem__ can ever save activations, disabling just in case
-            with torch._dynamo.utils._disable_saved_tensors_hooks_during_tracing():
+            # with torch._dynamo.utils._disable_saved_tensors_hooks_during_tracing():
```
laithsakka: remove this line

torch/_dynamo/variables/tensor.py
```diff
@@ -1096,7 +1097,20 @@ def method___setitem__(self, key, value):
             #   value.requires_grad is True => self.has_grad_fn becomes True
 
             # Not sure if __setitem__ can ever save activations, disabling just in case
-            with torch._dynamo.utils._disable_saved_tensors_hooks_during_tracing():
+
+            # Ignore fresh unbacked symbols that arises from patterns like this:
+            #   idx = b[i].item()
+            #   wrk[idx] += 1
```
laithsakka: maybe update the comment to explain that wrk[idx]  could allocate an unbacked symint 
but the shape/size is not changed since it its self updating. 
so we just ignore it.

torch/_dynamo/variables/tensor.py
```diff
@@ -1096,7 +1097,18 @@ def method___setitem__(self, key, value):
...
-            with torch._dynamo.utils._disable_saved_tensors_hooks_during_tracing():
+
+            # Ignore fresh unbacked symbols that arises from patterns the could arise from the
+            # internal indexing (selection) in, since the whole op output size/stride is not
+            # dependent on the indexing.
+            #   t[idx] += 1
+
+            with (
+                torch._dynamo.utils._disable_saved_tensors_hooks_during_tracing(),
+                tx.fake_mode.shape_env.ignore_fresh_unbacked_symbols()
```
ezyang, line 1122: idk, this seems kind of suspicious

laithsakka: ok let me explain this in details
When we trace   wrk[idx] += 1 this is what happen : 
```

TRACE LOAD_FAST wrk [RangeIteratorVariable()]
TRACE LOAD_FAST idx [RangeIteratorVariable(), TensorVariable()]
TRACE DUP_TOP_TWO None [RangeIteratorVariable(), TensorVariable(), SymNodeVariable()]
TRACE BINARY_SUBSCR None [RangeIteratorVariable(), TensorVariable(), SymNodeVariable(), TensorVariable(), SymNodeVariable()]      
create u1 for the select operation since input to is unbacked and use it in the storage offset of  output of 

TRACE BINARY_SUBSCR  TRACE LOAD_CONST 1 [RangeIteratorVariable(), TensorVariable(), SymNodeVariable(), TensorVariable()] 
TRACE INPLACE_ADD None [RangeIteratorVariable(), TensorVariable(), SymNodeVariable(), TensorVariable(), ConstantVariable(int: 1)]
 .. 
TRACE STORE_SUBSCR None [RangeIteratorVariable(), TensorVariable(), TensorVariable(), SymNodeVariable()]      
```
then we reach this line
```get_fake_value(proxy.node, tx, allow_non_graph_fake=False)```     where 
 
```proxy.node = setitem.```
 during this meta_select select is called on the write position, 
 maybe internally it mimics a[x] = a[x]+1? but the out of the setitem does not care about 
 u2 allocated in  that select. 
 
 let me know if you still think its fishy

Comment (laithsakka) [reply torch/_dynamo/variables/tensor.py:1122]: another thing we can do is figure out how to write a meta for setitem, this did not work 
```
@register_meta(torch.ops.aten.setitem)
def ss(self, a, v, c):
    ..
```
or push the ignore 
inside fake tensor dispatch for all setitem calls on fake tensors?? sounds risky.

Comment (laithsakka) [reply torch/_dynamo/variables/tensor.py:1122]: fix the comment to better explain

File: test/test_dynamic_shapes.py [hunk 5]
```diff
@@ -4093,6 +4093,38 @@ def func(x):
...
+        b = torch.ones(9, dtype=torch.int32)
+        compiled(a, b)
+        self.assertEqual(compiled(a, b), my_arithmetic(a, b))
+
+    @torch._dynamo.config.patch("capture_scalar_outputs", True)
+    def test_unbacked_item_set_item2(self):
+        def accumulate(X0, start):
+            start = start.item()
+            N = 3
+            result = X0[start].clone()
```
laithsakka: no need for the clone

ezyang:
I don't understand the pr description for (1). What does it mean that you get a new unbacked symbol if x is unbacked?

torch/_dynamo/variables/tensor.py
ezyang: So, an approach that would be less fishy to me would be to add a special case in the setitem Dynamo desugaring to specifically directly desugar it into aten ops that don't have this bad behavior. So in other words, the problem is the operator.setitem above; don't put that into the graph, put something that correctly allocates unbacked symbols when it needs to and doesn't otherwise.

torch/_dynamo/variables/tensor.py
```diff
@@ -1096,7 +1097,19 @@ def method___setitem__(self, key, value):
             #   value.requires_grad is True => self.has_grad_fn becomes True
 
             # Not sure if __setitem__ can ever save activations, disabling just in case
-            with torch._dynamo.utils._disable_saved_tensors_hooks_during_tracing():
+
+            # Ignore fresh unbacked symbols that could arise from the internal indexing (selection),
+            # that happen in code like t[idx] += 1 when idx is unabacked. Namely the selection
```
ezyang: unbacked

torch/_inductor/fx_passes/reinplace.py
```diff
@@ -78,7 +80,15 @@ def _inplace_generalized_scatter(
...
-        tmp = view.target(tmp, *fake_args, **fake_kwargs)
+        # slice and select can allocate new unbacked symints, but those won't be reflected
+        # in the output of this function, hence shall be ignored.
+        fake_mode = detect_fake_mode(fake_args)
+        with (
+            fake_mode.shape_env.ignore_fresh_unbacked_symbols()
+            if fake_mode and fake_mode.shape_env
+            else nullcontext()
+        ):
+            tmp = view.target(tmp, *fake_args, **fake_kwargs)
```
ezyang, line 91: I'm also confused here. You are going to have view.target inside of the graph now right? So tmp DOES have the unbacked symbol, and you should resolve unbacked symbols here (instead of suppressing them)

ezyang (changes requested):
vc me if you don't agree

torch/_inductor/fx_passes/reinplace.py
laithsakka: this is Not the function that does the de-composition , 
this is the function used to trace the  _generalized_scatter.


for the function that does the decomposition we need to do what you said right which i am doing on the PR that is right on top of this one on the stack.
https://github.com/pytorch/pytorch/pull/164948

for this function we self consume the views but return return inp on line 98 which would not have a reference to the new symbols in its example input.

torch/_dynamo/variables/tensor.py
laithsakka: mm so i see your point 
we originally did not even trace through setitem then this was added for when value is tensor to to propagate some properties
 
https://github.com/pytorch/pytorch/pull/161036

i.e according to the comment.
```
            #   value.requires_grad is True => self.requires_grad becomes True
            #   value.requires_grad is True => self.has_grad_fn becomes True
```

do you require this as part of this PR or just a suggestion for better approach?
basically we want to update the tensor properties but not require the trace through the set item that does this
```
  copy_to(
          impl::applySelect(
              self, 0, index.integer(), 0, self_device, self_sizes),
          value);
      return;
```

laithsakka: one thing could be , index any element in the input tensor during tracing we do not care which element since all we need is to propagate the properties.... mhmm still sounds complex lol.

wonder if there is a function that say propagate properties from x to y that we can just call here let me ask AI

torch/_dynamo/variables/tensor.py
bobrenjc93: s/unabacked/unbacked

torch/_inductor/lowering.py
```diff
@@ -1946,6 +1946,7 @@ def select(x, dim, idx):
             # Additionally, we want to avoid accidental unbacked unsqueeze semantics. To resolve this,
             # we use as_strided instead.
             # Removing this branch will cause test_unbacked_select_index_with_check to fail.
+            x.realize()
```
bobrenjc93, line 1961: add a comment?

torch/_inductor/lowering.py
```diff
@@ -1968,7 +1969,7 @@ def select(x, dim, idx):
     assert unbacked_bindings is not None
     assert len(unbacked_bindings) == 1, unbacked_bindings
     unbacked_offset_sym, _ = next(iter(unbacked_bindings.items()))
-
+    x.realize()
```
bobrenjc93, line 1986: ditto

bobrenjc93 (approved):

laithsakka:
address comments and rebase

PR #159323, opened 2025-07-29
[CI][CUDA] Add periodic b200 distributed job

nWEIdia:
1. Run distributed job with B200 runner, periodically. 
2. discovered generic distributed test issue that certain unit test hard-coded ranks, calling for require_exact_world_size(world_size) API instead of require_world_size(world_size). 

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @ezyang @msaroufim @dcci @ptrblck @eqy @tinglvv @huydhn @atalman @ZainRizvi @malfet

PR #164345, opened 2025-10-01
[CI][CUDA] Fix distributed tests for b200 

Aidyn-A:
This PR fixes the tests that were encountered in #159323.
Namely it fixes #162746 and #162745.

### DIFF
--- a/test/distributed/pipelining/test_stage.py
+++ b/test/distributed/pipelining/test_stage.py
@@ -30,7 +30,7 @@ from torch.utils._pytree import tree_map_only
 
 d_hid = 512
 batch_size = 256
-chunks = 4
+chunks = 8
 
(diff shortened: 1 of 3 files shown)


test/distributed/pipelining/test_stage.py
```diff
@@ -30,7 +30,7 @@
 
 d_hid = 512
 batch_size = 256
-chunks = 4
+chunks = 8
```
eqy, line 33: Is this corresponding to the number of devices that are expected? Should this be parameterized based on that?

Or is it OK as long as chunks >= number of accelerators?

eqy (approved):

test/distributed/pipelining/test_stage.py
Aidyn-A: Right, it it OK as long as `chunks >= world_size`.

PR #164350, opened 2025-10-01
Move AT_FORALL_... macros and ScalarTypeToCPPTypeT to headeronly

pearu:
* #164405
* #164354
* __->__ #164350


janeyx99 (commented):
Looks fine. What's the test plan?

torch/headeronly/core/ScalarType.h
```diff
@@ -82,6 +133,142 @@ struct dummy_int1_7_t {};
...
+  _(at::Float8_e4m3fn, Float8_e4m3fn)     \
+  _(at::Float8_e5m2fnuz, Float8_e5m2fnuz) \
+  _(at::Float8_e4m3fnuz, Float8_e4m3fnuz) \
+  _(at::Float8_e8m0fnu, Float8_e8m0fnu)
+
+#define AT_FORALL_COMPLEX_TYPES(_)     \
+  _(c10::complex<float>, ComplexFloat) \
+  _(c10::complex<double>, ComplexDouble)
+
+namespace c10 {
```
janeyx99: namespace c10 can still live on line 21, right? in the original file, these macros are defined within the namespace

pearu: Defining a CPP macro inside or outside of a namespace has no importance (`#define` is a preprocessor directive that is not visible to compiler after the preprocessing step) and there is no need to extend the `namespace c10` statement to capture macro definitions if there is no effect.

janeyx99: I'd like to simply keep the history as unchanged as possible except for the core thing that we're doing.

pearu: The namespace c10 is now extended to capture the macros.

While I think moving the comment `[dtype Macros note]` to headeronly ScalarType.h makes sense as it is closer to macro definitions (hence BE), this movement is unnecessary if the aim is to minimize history changes. Would you like the original comment location restored?

test/cpp/aoti_abi_check/test_scalartype.cpp
```diff
@@ -0,0 +1,55 @@
...
+    M(__VA_ARGS__ DEFINE_CHECK);                         \
+    EXPECT_EQ(count, EXPECTEDCOUNT);                     \
+  }
+
+TEST_FORALL(AT_FORALL_SCALAR_TYPES_WITH_COMPLEX_EXCEPT_COMPLEX_HALF_F8NZ, 14)
+TEST_FORALL(AT_FORALL_SCALAR_TYPES_WITH_COMPLEX, 18)
+TEST_FORALL(AT_FORALL_SCALAR_TYPES_WITH_COMPLEX_AND_QINTS, 46)
+TEST_FORALL(AT_FORALL_INT_TYPES, 5)
+TEST_FORALL(AT_FORALL_SCALAR_TYPES, 7)
+TEST_FORALL(AT_FORALL_SCALAR_TYPES_AND, 8, Bool, )
```
janeyx99, line 37: nit: why do we have the `, )`s at the end?

janeyx99 (approved):
thank u!

test/cpp/aoti_abi_check/test_scalartype.cpp
pearu: Notice `M(__VA_ARGS__ DEFINE_CHECK);` in TEST_FORALL that will expand to
```
AT_FORALL_SCALAR_TYPES_AND(Bool, DEFINE_CHECK);
                               ^--- this comma corresponds to the comma in `, )` 
```

PR #162351, opened 2025-09-07
[BE]: Update NCCL to 2.28.3

Skylion007:
@eqy New NCCL has some a bunch of bugfixes for features including reducing the number SMs needed by NVLINK collectives as well as some very useful new APIs for SymmetricMemory.  Also allows FP8 support for non-reductive operations on pre-sm90 devices.

PR #164352, opened 2025-10-01
Revert nccl upgrade back to 2.27.5

albanD:
Revert https://github.com/pytorch/pytorch/pull/162351 as it breaks H100


atalman (approved):

malfet (approved):

nWEIdia:
If I am not mistaken, the cu13 was using 2.27.7, not 2.27.5. So this is not strictly a revert but rather 
revert and downgrade NCCL from 2.27.7 to 2.27.5 in cu13. 
See https://github.com/pytorch/pytorch/pull/162351

PR #164359, opened 2025-10-01
Make mutation on DTensor.local_tensor work by making it a real operator

ezyang:
* __->__ #164359
* #165072

This is similar to things that were done for the NJT workstream.

I can't make it work for grad placements without making placements
representable in JIT schema though.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @aditvenk


ezyang:
@albanD is doing this properly

PR #164361, opened 2025-10-01
Fix vllm build issue

yangw-dev:
Fixes #ISSUE_NUMBER
unstable https://github.com/pytorch/pytorch/issues/164362

### DIFF
--- a/.github/ci_configs/vllm/Dockerfile.tmp_vllm
+++ b/.github/ci_configs/vllm/Dockerfile.tmp_vllm
@@ -353,6 +353,7 @@ RUN --mount=type=cache,target=/root/.cache/ccache \
     git clone --depth 1 --recursive --shallow-submodules \
         --branch ${FLASHINFER_GIT_REF} \
         ${FLASHINFER_GIT_REPO} flashinfer \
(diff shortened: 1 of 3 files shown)


.github/ci_configs/vllm/Dockerfile.tmp_vllm
```diff
@@ -338,20 +338,23 @@ ARG torch_cuda_arch_list='8.0;8.9;9.0a;10.0a;12.0'
...
 # see issue: https://github.com/flashinfer-ai/flashinfer/issues/738
 
-RUN pip install build==1.3.0
-RUN pip freeze | grep -E 'setuptools|packaging|build'
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv pip install  --system build==1.3.0
+RUN uv pip freeze | grep -E 'setuptools|packaging|build'
 
 ENV TORCH_CUDA_ARCH_LIST=${torch_cuda_arch_list}
+ENV FLASHINFER_CUDA_ARCH_LIST=${torch_cuda_arch_list}
```
huydhn: Btw, there is a new script now to build FlashInfer from vLLM https://github.com/vllm-project/vllm/pull/25782 called `tools/flashinfer-build.sh`.  You could try to see if that script works for us here (later).  It could simplify the logic here a bit I think

yangw-dev: i see, i think that one did similar thing we do here
sounds good

huydhn (approved):

PR #164387, opened 2025-10-01
Improve efficiency of at::float32Precision APIs

swolchok:
* __->__ #164387

I saw this taking up 1% total time when profiling training in torchtitan with (requires Python 3.13+ for PYTHON_PERF_JIT_SUPPORT):

```
MASTER_PORT=3000 MASTER_ADDR=0.0.0.0 RANK=0 LOCAL_RANK=0 WORLD_SIZE=1 PYTHON_PERF_JIT_SUPPORT=1 perf record -m 1024  -D11500 -k 1 -g --call-graph dwarf python -m torchtitan.train --job.config-file ./torchtitan/models/llama3/tra
in_configs/debug_model.toml --compile.enable --training.steps 30 --compile.backend=aot_eager
```

It's a straightforward enough fix (should've used enums all along, IMO) so I went ahead and fixed it.

### DIFF
--- a/aten/src/ATen/Context.cpp
+++ b/aten/src/ATen/Context.cpp
@@ -77,7 +77,7 @@ std::pair<detail::Float32Backend, detail::Float32Op> check_fp32_prec_backend_and
   return std::make_pair(result_backend, detail::Float32Op(op_it - operators.begin()));
 }
 
-constexpr std::array<std::string_view, 4> float32_precision_strs = {"ieee", "tf32", "bf16", "none"};
+constexpr std::array<std::string_view, 4> float32_precision_strs = {"none", "ieee", "tf32", "bf16"};
 
 std::string_view fp32_prec_to_str(detail::Float32Precision prec) {
   return float32_precision_strs.at(static_cast<int>(prec));
--- a/aten/src/ATen/Context.cpp
+++ b/aten/src/ATen/Context.cpp
@@ -10,6 +10,7 @@
 #include <cctype>
 #include <stdexcept>
 #include <string>
+#include <string_view>
 
 #include <ATen/cpu/FlushDenormal.h>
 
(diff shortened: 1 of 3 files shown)


aten/src/ATen/Context.cpp
```diff
@@ -40,50 +41,87 @@ namespace {
...
+  // With so few backends, it seems better to just check them
+  // explicitly rather than loop over an array. If adding more,
+  // consider going back to an array like for operators.
+  detail::Float32Backend result_backend;
+  if (backend == "cuda") {
+    result_backend = detail::Float32Backend::CUDA;
+  } else if (backend == "mkldnn") {
+    result_backend = detail::Float32Backend::MKLDNN;
+  } else {
+    TORCH_CHECK(
```
Skylion007:
```suggestion
    TORCH_CHECK_VALUE(

File: aten/src/ATen/Context.cpp [hunk 2]
```diff
@@ -40,50 +41,87 @@ namespace {
                 ->conv
                 ->rnn
 */
-const std::map<std::string, std::vector<std::string>> _fp32_precisions = {
-    {"generic", {{"ieee", "tf32", "bf16", "none"}}},
-    {"mkldnn", {{"ieee", "tf32", "bf16", "none"}}},
-    {"cuda", {{"ieee", "tf32", "none"}}}};
+const std::array<std::pair<detail::Float32Backend, std::array<detail::Float32Precision, 4>>, 3> _fp32_precisions = {{
```
Skylion007: If this doesn't work, leave a constexpr ifdef comment for C++20 etc.
```suggestion
constexpr std::array<std::pair<detail::Float32Backend, std::array<detail::Float32Precision, 4>>, 3> _fp32_precisions = {{

File: aten/src/ATen/Context.cpp [hunk 3]
```diff
@@ -40,50 +41,87 @@ namespace {
...
+  } else if (backend == "mkldnn") {
+    result_backend = detail::Float32Backend::MKLDNN;
+  } else {
+    TORCH_CHECK(
+        false,
+        "Invalid backend: ",
+        backend);
   }
+  const auto op_it = std::find(operators.begin(), operators.end(), op);
+  TORCH_CHECK(
```
Skylion007:
```suggestion
  TORCH_CHECK_VALUE(

File: aten/src/ATen/Context.cpp [hunk 4]
```diff
@@ -40,50 +41,87 @@ namespace {
...
+  static constexpr std::array<std::string_view, 4> operators = {"conv", "matmul", "rnn", "all"};
   if (backend == "generic") {
     TORCH_CHECK(op == "all", "Invalid operation for generic backend: ", op);
+    return std::make_pair(detail::Float32Backend::GENERIC, detail::Float32Op::ALL);
   }
+  // With so few backends, it seems better to just check them
+  // explicitly rather than loop over an array. If adding more,
+  // consider going back to an array like for operators.
+  detail::Float32Backend result_backend;
+  if (backend == "cuda") {
```
Skylion007, line 62: nit: use a switch statement here?

aten/src/ATen/Context.h
```diff
@@ -488,22 +495,32 @@ class TORCH_API Context {
...
-             ? "none"
-             : "tf32"},
-        {"conv", "tf32"},
-        {"rnn", "tf32"},
-        {"all", "none"}}},
-  };
+  // Yes, it's odd that the inner array has extra length here for the
+  // generic backend, but it seems better than using std::vector.
+  // Map backend->op->precision setting.
+  std::array<
```
Skylion007: Shouldn't these be constexpr or at least const?

aten/src/ATen/Context.h
```diff
@@ -488,22 +495,32 @@ class TORCH_API Context {
...
+      std::pair<
+          detail::Float32Backend,
+          std::
+              array<std::pair<detail::Float32Op, detail::Float32Precision>, 4>>,
+      3>
```
benjaminglass1: Tempted to say that an `unordered_map` might be a better data structure to contain this, since you end up having to do `find_if` calls for keys in this structure.

aten/src/ATen/Context.cpp
benjaminglass1: We should probably gloss over this with a `TORCH_CHECK_FAIL` macro at some point for clarity.

aten/src/ATen/Context.cpp
benjaminglass1: @Skylion007 IIRC switch doesn't work with strings?

Skylion007: Whoops you are right

aten/src/ATen/Context.h
Skylion007: Normally yes, but unordered_map is a pretty inefficient data struct for small number of elements unfortunately so this is a valid and unfortunate micropt.

benjaminglass1: Ugh, valid. Feedback withdrawn!

Skylion007: @benjaminglass1 It looks like we always index based on the first value of this enum though, could we map the enums to a static offset (like using an int enum) for constant time checking though? That might work, be faster, and more readable, if a bit hacky.

benjaminglass1: Honestly, based on the amount of `find_if` here...that might be a good idea. I don't love it, but what we have is also a little hacky, so springing for some more readability would be helpful.

aten/src/ATen/Context.cpp
Skylion007:
```cpp
// enum_str_map.hpp (C++17)
#pragma once
#include <array>
#include <cstdint>
#include <optional>
#include <ostream>
#include <string_view>

// ---------- tiny constexpr FNV-1a 64-bit ----------
constexpr uint64_t fnv1a(std::string_view s) {
    uint64_t h = 1469598103934665603ull;
    for (unsigned char c : s) { h ^= c; h *= 1099511628211ull; }
    return h;
}
template <size_t N>
constexpr uint64_t fnv1a_lit(const char (&s)[N]) { // excludes trailing '\0'
    return fnv1a(std::string_view{s, N - 1});
}

// ---------- Your enum & traits in one place ----------
enum class Color : uint8_t { Red, Green, Blue, _count };

struct ColorTraits {
    static constexpr std::array<std::string_view, static_cast<size_t>(Color::_count)> names{
        "Red", "Green", "Blue"
    };
    // Optional: allow lowercase, aliases, etc.
};

// ---------- to_string ----------
constexpr std::string_view to_string(Color e) {
    const auto idx = static_cast<size_t>(e);
    return (idx < ColorTraits::names.size()) ? ColorTraits::names[idx] : std::string_view{"<invalid>"};
}

// ---------- from_string (switch on hash + collision check) ----------
inline std::optional<Color> color_from_string(std::string_view sv) {
    switch (fnv1a(sv)) {
        case fnv1a_lit("Red"):   if (sv == "Red")   return Color::Red;   break;
        case fnv1a_lit("Green"): if (sv == "Green") return Color::Green; break;
        case fnv1a_lit("Blue"):  if (sv == "Blue")  return Color::Blue;  break;
        // You can add lowercase aliases as additional cases if desired:
        // case fnv1a_lit("red"): if (sv == "red") return Color::Red; break;
        default: break;
    }
    return std::nullopt;
}

// ---------- stream helpers ----------
inline std::ostream& operator<<(std::ostream& os, Color e) {
    return os << to_string(e);
}

// Optional: input stream >> (simple, not robust to errors)
#include <istream>
inline std::istream& operator>>(std::istream& is, Color& out) {
    std::string token;
    if (is >> token) {
        if (auto v = color_from_string(token)) out = *v;
        else is.setstate(std::ios::failbit);
    }
    return is;
}
```
Was thinking of this constexpr hash hacking, but maybe premature optimization. Might be a better way to go between str to enum and back though than current code

aten/src/ATen/Context.h
Skylion007: Posted a snippit above which is fast, the actual inline code is readable, but the enum code becomes a little bit less so. It's very fast and efficient and abuses constexpr though.

https://github.com/pytorch/pytorch/pull/164387#discussion_r2395759778

aten/src/ATen/Context.cpp
swolchok: I expect comparison of small string views to be 1) faster than attempting to hash 2) not worth benchmarking either way

aten/src/ATen/Context.h
swolchok: this one is mutable, it's part of the Context state

aten/src/ATen/Context.h
swolchok: if we had a real enum-keyed map abstraction it would simplify this code further. I just wasn't willing to spend even more time than I already spent on this PR developing one.

swolchok: OK, after staring at this slightly more, I'm willing to get rid of the enum "keys" in Context::fp32_precision; agree that's silly. Went through multiple incremental refinements to get where the PR is at the moment and just stopped too early, whoops

benjaminglass1 (approved):

PR #164389, opened 2025-10-01
torch.cond on DTensor triggers an internal assert, add xfail for this.

ezyang:
* __->__ #164389


albanD (approved):

ydwu4:
I'm looking into adding a sharding propagation rule for cond (and other control flow operators), any suggestions on where I should start with? i'm now just trying to prototyping a custom handler following other operators starting with asserting two branches return the same sharding.

PR #164399, opened 2025-10-01
Change default device to current acclerator

drisspg:
* __->__ #164399

cc @albanD @guangyey @EikanWang


drisspg:
cc @albanD

drisspg:
what does torch.accelerator.current_accelerator() return on a device w/ no accelerator is it not cpu? That being said in retrospect there should have been no default arg for this func

torch/nn/attention/flex_attention.py
```diff
@@ -36,6 +36,8 @@
 from torch.nn.attention._utils import _validate_sdpa_input
 from torch.utils._pytree import tree_map_only
 
+DEFAULT_DEVICE = torch.accelerator.current_accelerator()
```
albanD: Since this can change post-initial import for out of tree backends, I would suggest setting the arg to `None` and querying this in your function.

drisspg: sounds good

albanD (commented):
SGTM !

albanD (approved):

PR #164404, opened 2025-10-01
[Inductor][Triton][FP8] Support deepseek-style scaling in Inductor

jananisriram:
Summary:
Support deepseek-style scaling in Inductor Triton for FP8 GEMMs. DeepSeek-style scaling is a colloquial term for a fine-grained mixed precision framework using FP8 to train [Deepseek-V3](https://arxiv.org/pdf/2412.19437), DeepSeek AI's recent MoE (Mixture of Experts) model. DeepSeek-style scaling effectively extends the dynamic range of FP8 by mitigating dequantization overhead under increased-precision accumulation, which is key to achieving more accurate FP8 GEMM results.

DeepSeek-style scaling on matmul `A @ B` leverages two different types of scaling strategies to preserve a balance between numerical stability and training efficiency:
- Activations (input tensor `A`): tile-wise (1x128 across shape `(M, K)`)
- Weights (input tensor `B`): block-wise (128x128 across shape `(N, K)`)

This diff enables Inductor users to replicate past successes with deepseek-style scaling and achieve higher numerical stability while increasing training efficiency.

NOTE: Block-wise 128x128 scaling is only supported in CUDA 12.9+; therefore, deepseek-style scaling is currently unsupported in `fbcode` (CUDA 12.4). Use OSS PyTorch to run deepseek-style scaling.

NOTE: Accuracy for FP8 is unstable, even with high tolerances, which is why TritonBench benchmarks are unlikely to be accurate against a `torch` implementation.

Test Plan:
In OSS PyTorch, run
```
TORCHINDUCTOR_CACHE_DIR=~/personal/cache_dir_inductor CUDA_LAUNCH_BLOCKING=1 TORCH_USE_CUDA_DSA=1 TRITON_PRINT_AUTOTUNING=1 TRITON_ALWAYS_COMPILE=1 TORCH_LOGS=+inductor TORCHINDUCTOR_FORCE_DISABLE_CACHES=1 ENABLE_PERSISTENT_TMA_MATMUL=1 TORCHINDUCTOR_MAX_AUTOTUNE_GEMM=1 python run.py --op fp8_gemm --only torch_fp8_gemm,pt2_fp8_gemm --metrics tflops,accuracy --m 4096 --n 768 --k 512 --output="{output_dir}/deepseek_bench.csv" --scaling_deepseek --atol=1e-2 --rtol=0.5 2>&1 | tee ~/personal/deepseek_style/deepseek_bench.log
```

Differential Revision: D83609850

### DIFF
--- a/torch/_inductor/kernel/mm.py
+++ b/torch/_inductor/kernel/mm.py
@@ -42,6 +42,7 @@ from ..select_algorithm import (
 )
 from ..utils import (
     _use_cutlass_for_op,
+    ceildiv,
     use_aten_gemm_kernels,
     use_ck_gemm_template,
     use_ck_tile_gemm_template,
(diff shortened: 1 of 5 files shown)


torch/_inductor/kernel/mm.py
```diff
@@ -558,10 +559,172 @@ def apply_scaling(
...
+    num_pid_n = tl.cdiv(N, BLOCK_N)
+    k_tiles = tl.cdiv(K, BLOCK_K)
+    num_tiles = num_pid_m * num_pid_n
+
+    {%- if TMA_EXPERIMENTAL_API %}
+    workspace_base = ws_ptr + start_pid * 2 * TMA_SIZE
+    a_desc_ptr = workspace_base
+    b_desc_ptr = workspace_base + TMA_SIZE
+
+    triton.language.extra.cuda.experimental_device_tensormap_create2d(
```
drisspg: do we need experimental tma and the tensor descriptor? isn't the later the way to do this?

jananisriram: i left this in because the other scaled_mm device_tma template also supports it, but if we don't want to support the experimental tma api, we can definitely remove this! @drisspg are we interested in supporting it here?

drisspg: I think that is planning to be removed AFAIK, and that the tensor descriptor will be the main path forward so we should remove

test/inductor/test_max_autotune.py
```diff
@@ -799,6 +799,9 @@ def addmm(x, a, b):
     @unittest.skipIf(
         not has_triton_tma_device(), "Need device-side TMA support in Triton"
     )
+    @unittest.skipIf(
+        has_datacenter_blackwell_tma_device(), "B200 doesn't support sm carveout"
+    )
```
drisspg, line 814: dirty commit?

torch/_inductor/kernel/mm.py
```diff
@@ -407,7 +408,7 @@ def apply_scaling(
 """
 
 
-device_tma = r"""
+scaled_mm_device_tma_tensor_row = r"""
```
drisspg: Is this only handling row scaling?

drisspg: maybe `row` and `tensor` right?

torch/_inductor/kernel/mm.py
```diff
@@ -558,10 +559,136 @@ def apply_scaling(
...
-scaled_mm_device_tma_template = TritonTemplate(
-    name="scaled_mm_device_tma",
+scaled_mm_device_tma_tensor_row_template = TritonTemplate(
+    name="scaled_mm_device_tma_tensor_row",
     grid=persistent_mm_grid,
-    source=device_tma + load_scales + apply_scaling,
+    source=scaled_mm_device_tma_tensor_row + load_scales + apply_scaling,
+)
+
+scaled_mm_device_tma_deepseek_block_group = r"""
```
drisspg: can we call this groupwise_blockwsie not deepseek?

torch/_inductor/kernel/mm.py
drisspg: actually lets call this epilogue scaling

torch/_inductor/kernel/mm.py
```diff
@@ -558,10 +559,136 @@ def apply_scaling(
 """
 
 
-scaled_mm_device_tma_template = TritonTemplate(
-    name="scaled_mm_device_tma",
+scaled_mm_device_tma_tensor_row_template = TritonTemplate(
```
drisspg: can we call this something main_loop scaling

torch/_inductor/kernel/mm.py
jananisriram: yeah this gets renamed later to blockwise scaling (encompassing deepseek and tile-wise). i should change the name in this pr though

torch/_inductor/kernel/mm.py
```diff
@@ -558,10 +559,136 @@ def apply_scaling(
...
+
+        scale_a_expanded = scale_a_tile[:, :, None]  # [BLOCK_M, tl.cdiv(K, BLOCK_K), 1]
+        scale_a_expanded = tl.broadcast_to(
+            scale_a_expanded,
+            (BLOCK_M, (BLOCK_K + 128 - 1) // 128, MIN_BLOCK_K_128)
+        )  # [BLOCK_M, tl.cdiv(K, BLOCK_K), MIN_BLOCK_K_128]
+        scale_a_expanded = scale_a_expanded.reshape(
+            BLOCK_M,
+            ((BLOCK_K + 128 - 1) // 128) * MIN_BLOCK_K_128
+        )  # [BLOCK_M, MIN_BLOCK_K_128]
```
drisspg: it might be easier to maintain,  to have the same base template but to `mixin` this main loop scaling.  You can do this w/ jinja templates pretty easily, and essentially produce a new template w/ the mainloop scaling set

torch/_inductor/kernel/mm.py
```diff
@@ -1338,8 +1471,60 @@ def tuned_scaled_mm(
...
-            templates_to_use.append(scaled_mm_device_tma_template)
-            kwarg_overrides[scaled_mm_device_tma_template.uid] = overriders
+            scale_a_size, scale_b_size = scale_a_real.shape, scale_b_real.shape
+
+            def _is_scalar_like(sz: Any) -> bool:
+                return (len(sz) == 0) or all(
+                    V.graph.sizevars.statically_known_equals(d, 1) for d in sz
+                )
+
+            def _ceil_div(v: int, d: int) -> int:
```
drisspg: we gotta have one of these lying around already lol

torch/_inductor/kernel/mm.py
```diff
@@ -1338,8 +1471,60 @@ def tuned_scaled_mm(
...
+                return (v + d - 1) // d
+
+            def _is_blockwise_1x128(
+                t: torch.Tensor, scale_t_size: torch.Tensor
+            ) -> torch.Tensor:
+                return scale_t_size[0] == t.shape[0] and scale_t_size[1] == _ceil_div(
```
drisspg: did we fuzz w/ dynamic shapes to make sure we dont get any symint errors

torch/_inductor/kernel/mm.py
```diff
@@ -1253,6 +1380,12 @@ def tuned_sparse_semi_structured_mm(
     )
 
 
+class ScalingMode(IntEnum):
+    TENSOR = 0
+    ROW = 1
+    DEEPSEEK = 2
```
drisspg: name change, should align w/ aten

drisspg (commented):
Overall looks good a couple comments,

cc @slayton58

drisspg (commented):
Also, tests?

torch/_inductor/kernel/mm.py
jananisriram: also aten right now doesn't support group-wise (it's more so tile-wise), so block-wise seems to be more of an encompassing name

drisspg: Yeah honestly naming is hard, I have used
PerTensor -> 1 scale
PerRow -> [M,K] Tensor  = M scales
PerGroup -> [M, K] Tensor = [M, K//group] scales with a Block_size [ 1, group]
PerBlock -> [M, K] Tensor = [M//group_a, K//group_b] scales with a Block_size [ group_a, group_b]

So group is 1d spanning block is 2d spanning

slayton58: I must confess that "group" scales are a new name to me :D 

ATen is using `BlockWise{N}x{M}` for both 1d and 2d blocks, so we really should be consistent with that.

drisspg: ahh okay yeah lets stick with that sorry

jananisriram: aten doesn't suggest any names for the [pairs](https://www.internalfb.com/code/fbsource/[ee620a7a67ff3d13d44d28983a0976424d2354e2]/fbcode/caffe2/aten/src/ATen/native/cuda/Blas.cpp?lines=1276) of `(Blockwise{M}x{K}, Blockwise{N}x{K})`, i.e. deepseek-style scaling would be `(Blockwise1x128, Blockwise128x128)`, while tile-wise scaling would be `(Blockwise1x128, Blockwise1x128)`. any suggestions on naming conventions (conforming with aten) for these pairs?

torch/_inductor/kernel/mm.py
jananisriram: yep i checked that dynamic shapes also works!

torch/_inductor/kernel/mm.py
jananisriram: thanks for the feedback! as discussed offline, created a better design for the jinja templates so that it's easy to add in new scaling modes

torch/_inductor/kernel/mm.py
```diff
@@ -1253,6 +1469,80 @@ def tuned_sparse_semi_structured_mm(
     )
 
 
+# Supported scaling types (from aten/src/ATen/native/cuda/Blas.cpp::1163)
+# TODO (jananisriram): support BlockWise1x16, BlockWise1x32, BlockWise1x128, BlockWise128x128
+class ScalingType(IntEnum):
```
slayton58: Enums for scaling now exposed via. `torch.nn.functional.ScalingType`, and consistent between ATen and frontend - can we use those instead (Should be drop-in I think) and all use the same single source-of-truth?

torch/_inductor/kernel/mm.py
```diff
@@ -558,10 +562,222 @@ def apply_scaling(
...
+
+            offs_am = pid_m * BLOCK_M
+            offs_bn = pid_n * BLOCK_N
+
+        offs_k = ki * BLOCK_K
+
+        a = tl.load_tensor_descriptor(a_desc, [offs_am, offs_k])
+        b = tl.load_tensor_descriptor(b_desc, [offs_bn, offs_k])
+
+        m_blocks = tl.cdiv(M, 128)
```
slayton58: Maybe a noob question, but is the constant 128 here due to the gemm tiles, or the scaling tiles? If the latter, it should probably be templatized to more easily handle MX/NV-style blocksizes.

torch/_inductor/template_heuristics/triton.py
```diff
@@ -2127,19 +2120,74 @@ def _filter_configs(self, configs: list[BaseConfig]) -> list[BaseConfig]:
...
+            kernel_inputs,
+            op_name,
+        ):
+            # Add scaling-specific options for main loop scaling variants
+
+            # Inductor templates require compile-time constants passed in as tl.constexpr values.
+            # In cases in which the block size (BLOCK_*) is smaller than the tile size (128, 32, 16),
+            # scales must be broadcasted to BLOCK_* (rather than to a tile_sizextile_size chunk).
+
+            # TODO (jananisriram): find a way to calculate MIN_BLOCK_TILE_* using
```
slayton58: Is the problem here that you can't pass `_get_template_configs_impl` with additional arguments to denote 128/32/16?

slayton58 (commented):
Looks good, couple of minor comments

torch/_inductor/kernel/mm.py
jananisriram: yep it's almost drop-in (with the exception of the enum order, which does affect the templates a little)!

torch/_inductor/kernel/mm.py
slayton58: @jananisriram I missed this, but you can look at `enum class ScaledGemmImplementation` [HERE](https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/Blas.cpp#L1691) as an idea of what I did

torch/_inductor/kernel/mm.py
jananisriram: it's due to the scaling tiles, and in an ideal world (follow-up PR hopefully), we're able to templatize this to pass in one of 128/32/16. i'm currently looking into whether we can actually get the correct scaling tile values into `triton.py` to simplify this a bit.

torch/_inductor/kernel/mm.py
jananisriram: btw, we do get this mypy lint error: `Module "torch.nn.functional" has no attribute "ScalingType"` even though functionality is correct. is that ok to bypass?

torch/_inductor/kernel/mm.py
jananisriram: update: we found a way to get the tile sizes into the kernel, so this won't be hard-coded into the relevant templates (`1xTILESIZE_scaling`) anymore!

torch/_inductor/template_heuristics/triton.py
jananisriram: here, the problem is that when the tuned block size for a kernel is smaller than the scaling mode's tile size itself, our kernel needs to use the tuned block size rather than the scaling mode's tile size to calculate its shape

slayton58 (approved):

PR #164354, opened 2025-10-01
Remove workaround to old CUDA bug

pearu:
As in the title.

A check for https://github.com/pytorch/pytorch/issues/164348 to see if the workaround can be removed.

Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #164405
* __->__ #164354
* #164350

cc @ptrblck @msaroufim @eqy @jerryzh168 @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10

PR #164405, opened 2025-10-01
Move toString(ScalarType) and ScalarType ostream operator to headeronly

pearu:
* __->__ #164405
* #164354
* #164350


torch/headeronly/core/ScalarType.h
```diff
@@ -267,6 +267,25 @@ enum class ScalarType : int8_t {
...
+
+  switch (t) {
+    AT_FORALL_SCALAR_TYPES_WITH_COMPLEX_AND_QINTS(DEFINE_CASE)
+    default:
+      return "UNKNOWN_SCALAR";
+  }
+#undef DEFINE_CASE
+}
+
+inline std::ostream& operator<<(
```
janeyx99, line 301: We should expose what we want to use in torch::headeronly (for toString and this)

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
```diff
@@ -486,6 +486,37 @@ STABLE_TORCH_LIBRARY_IMPL(libtorch_agnostic, CompositeExplicitAutograd, m) {
...
+  int64_t sizes[] = {static_cast<int64_t>(s.size())};
+  int64_t strides[] = {1};
+  aoti_torch_create_tensor_from_blob(s.data(), 1, sizes, strides, 0, aoti_torch_dtype_int8(), aoti_torch_device_type_cpu(), 0, &ath);
+  // cloning the Tensor instance because ath does not own the data
+  // when creating a tensor from blob:
+  return torch::stable::clone(Tensor(ath));
+}
+
+Tensor ScalarType2string(Tensor t) {
+  return string2Tensor(c10::toString(t.scalar_type()));
```
janeyx99: use torch::headeronly here

janeyx99 (commented):
I'm looking for tests in aoti_abi_check like mentioned before, because those test that using these APIs does not require linking torch. See example in https://github.com/pytorch/pytorch/pull/163999.

In general, I'm not convinced we need a test that is a bit roundabout like string2Tensor. We can just go through the ScalarTypes and call toString on them and make sure they're as expected.

torch/headeronly/core/ScalarType.h
pearu: We want `operator<<` so that ScalarType values can be used within `STD_TORCH_CHECK/TORCH_CHECK`.

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
pearu: resolving as ScalarType2string is removed.

torch/headeronly/core/ScalarType.h
pearu: Exposed to torch::headeronly.

Skylion007 (approved):

janeyx99 (commented):
LGTM! Can you add these 2 APIs to torch/header_only_apis.txt as well

torch/headeronly/core/ScalarType.h
```diff
@@ -263,11 +263,32 @@ enum class ScalarType : int8_t {
...
+
 } // namespace c10
 
 namespace torch::headeronly {
 using c10::dummy_int1_7_t;
 using c10::dummy_uint1_7_t;
 using c10::NumScalarTypes;
 using c10::ScalarType;
+using c10::toString;
+using c10::operator<<;
```
janeyx99, line 318: If we're going to expose << in torch::headeronly, can you move the tests to this PR then? Or wait til that PR to expose this API.

pearu: I have created tests for `<<` in this PR.

janeyx99 (approved):

Issue #160317, opened 2025-08-11
Memory leak when using mark_dirty in Python custom autograd.Function


maoyuanpeng:
### 🐛 Describe the bug

I'm observing a memory leak when using a custom torch.autograd.Function with an in-place operation marked by ctx.mark_dirty(...). The following minimal repro highlights the issue:

```python
import torch

class ReluOps(torch.autograd.Function):
    @staticmethod
    def forward(ctx, x):
        x.relu_()
        ctx.mark_dirty(x)
        ctx.save_for_backward(x)
        return x

    @staticmethod
    def backward(ctx, grad):
        x = ctx.saved_tensors
        return (grad * x[0])

def run():
    for i in range(100):
        x = torch.rand((100, 100, 1000), requires_grad=True)
        z = x + 1.0
        z_view = z[0]
        ReluOps.apply(z_view)

if __name__ == "__main__":
    run()

```

Memory usage increases over iterations, and is not released between invocations.

The cycle looks like this:
CopySlices -> PyNode -> THPFunction -> SavedVariable -> AsStridedBackward -> CopySlices

### Versions
torch master
run above code

cc @ezyang @gchanan @zou3519 @kadeng @msaroufim @albanD @gqchen @nikitaved @soulitzer @Varal7 @xmfan @chauhang @penguinwu

vishalgoyal316:
I would like to work on this issue.

PR #164407, opened 2025-10-01
Fix custom autograd Function memory leak when saving mutated view

soulitzer:
* __->__ #164407

Fixes https://github.com/pytorch/pytorch/issues/160317


torch/csrc/autograd/python_function.cpp
```diff
@@ -812,17 +812,29 @@ static void _get_tensors_to_save(
...
+      output_impls.insert(tensor.unsafeGetTensorImpl());
+    }
+  }
+
   for (const auto& opt_tensor : tensors_to_save) {
     if (!opt_tensor.has_value()) {
       self->saved_variables.emplace_back();
     } else {
-      bool is_output = opt_tensor.value().grad_fn().get() == cdata_ptr.get();
+      bool is_output = output_impls.count(opt_tensor.value().unsafeGetTensorImpl()) > 0;
```
soulitzer: It's okay to not check whether grad_fn (e.g. to avoid unnecessarily marking non-float/complex outputs, outputs marked non-differentiable as is_output) because SavedVariable categorizes these tensors as is_leaf=true and goes through the save_original path.

torch/csrc/autograd/python_function.cpp
```diff
@@ -812,17 +812,29 @@ static void _get_tensors_to_save(
...
+
+  std::unordered_set<at::TensorImpl*> output_impls{};
+  for (const auto i : c10::irange(num_outputs)) {
+    PyObject* obj = PyTuple_GET_ITEM(outputs, i);
+    if (THPVariable_Check(obj)) {
+      const auto& tensor = THPVariable_Unpack(obj);
+      output_impls.insert(tensor.unsafeGetTensorImpl());
+    }
+  }
+
```
soulitzer, line 833: We also construct the same set for checking whether mark_dirty'd tensors are actually outputs, but its hard to deduplicate.
- It's not possible to create the set prior to processing the outputs because the outputs will change (e.g. in the forward is same tensor as output case) 
- One way to do it would be to return the set created in process_backward_ad and return it all the way up all the layers. Let me know if this seems better.

torch/csrc/autograd/python_function.cpp
```diff
@@ -812,17 +812,29 @@ static void _get_tensors_to_save(
...
+    THPFunction* self,
+    PyObject* outputs,
+    int64_t num_outputs) {
   if (tensors_to_save.size() == 0)
     return;
   size_t num_saved = tensors_to_save.size();
   self->saved_variables.clear();
   self->saved_variables.reserve(num_saved);
+
+  std::unordered_set<at::TensorImpl*> output_impls{};
```
Skylion007, line 824:
```suggestion
  std::unordered_set<at::TensorImpl*> output_impls{};
  output_impls.reserve(num_outputs);
```

test/test_autograd.py
```diff
@@ -8902,6 +8902,28 @@ def test_named_tensor_for_complex_views(self):
...
+            def backward(ctx, grad):
+                pass
+
+        def scope():
+            x = torch.tensor(1., requires_grad=True).clone()
+            x = x.view_as(x)
+            y = Test.apply(x)
+            return weakref.ref(x)
+
+        with disable_gc():
```
albanD: If this is a real memory leak, whether you put the gc here or not, it wouldn't change anything!

albanD (approved):
nit on removing the gc from test, but sgtm!

PR #164409, opened 2025-10-01
[ROCm] Improve non stride-one backwards indexing for small index sets

doru1004:
This patch fixes a performance problem which occurs when a small set of indices is used and there are practically no duplicates.

### DIFF
--- a/aten/src/ATen/native/cuda/Indexing.cu
+++ b/aten/src/ATen/native/cuda/Indexing.cu
@@ -59,7 +59,7 @@ constexpr uint64_t getDefaultMaxThreadsPerBlock() {
 #ifdef USE_ROCM
 #define SKIP_SORTED_INDICES 32
 template <typename scalar_t, int SZ>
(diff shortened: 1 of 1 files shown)


aten/src/ATen/native/cuda/Indexing.cu
```diff
@@ -780,11 +782,72 @@ void index_put_with_sort_kernel(Tensor & self, const c10::List<std::optional<Ten
             kBool,
             kBFloat16);
         } else {
+#ifdef USE_ROCM
```
jerrymannil, line 785: Can we do this instead to reduce code bloat?

```
#ifdef USE_ROCM
     if (num_indices >= 200000) {
         ....................
         ....................
         ....................
         return;
     }
#endif

# Use common path (indexing_backward_kernel) for  NV & ROCM with num_indices < 200000
```

jerrymannil (approved):

jeffdaily (approved):

PR #164412, opened 2025-10-01
[PyTorch / Sigrid GPU] Fixes in pinned stats collection and add new ODS pinned memory stats

banitag1:
We do some fixes in pinned memory allocation stats collection and better differentiate between active vs allocated bytes.
Reviewed By: bbus, sayitmemory

Differential Revision: D83162346


mradmila (approved):
I only reviewed the changes wrt to the HostStats and how we collect those.

PR #164416, opened 2025-10-01
[MPS] Support large tensors in `torch.cat`

### LINKED ISSUE #164415: Support printing large tensors in MPS
```
>>> import torch
>>> a = torch.ones((2, 1<<31), dtype=torch.int8, device='mps')
>>> print(a)
...
RuntimeError: MPSGaph does not support tensor dims larger than INT_MAX
```

The proposed solution is to add an alternate MPS implementation of `torch.cat` to use when the input is too large to use the MPSGraph implementation.

kurtamohler:
* __->__ #164416

Fixes #164415


test/test_mps.py
```diff
@@ -3637,6 +3640,70 @@ def rotate_subset(data, dim):
...
 
+    # Skip if a test needs more memory than the system has.
+    def _skip_if_exceeds_total_memory(self, required_memory):
+        if total_memory < required_memory:
+            self.skipTest(
+                f"Needs {required_memory / (1024**3):0.01f} GiB RAM, "
+                f"but only {total_memory / (1024**3):0.01f} GiB is available.")
+
+    @parametrize("dtype", MPS_DTYPES)
+    def test_cat_large_tensor(self, dtype):
```
kurtamohler, line 3651: Note that on the machine I'm using, it takes about 1 second to run this for one of the dtypes, so maybe we shouldn't test all dtypes?

aten/src/ATen/native/mps/operations/Shape.mm
```diff
@@ -249,6 +329,12 @@ static void check_shape_except_dim(const Tensor& first, const Tensor& second, in
     return;
   }
 
+  has_large_tensor |= isTooLargeForMPSGraph(out);
+
+  if (has_large_tensor) {
```
kurtamohler, line 333: I wanted to check whether the alternate implementation works correctly for smaller sizes as well, so I tried temporarily changing this condition to always be true. I ran `python -m pytest test/test_mps.py -k output_match_cat` and it passed.

But it might be a good idea to cover this in CI too. To do that, we could add a non-public python api (either a global flag or a non-public function) that forces calling the alternate impl even if the tensors are small. Then we could add a test in `test_mps.py` that runs the opinfo cases for `cat` using the alternate impl.

But idk, maybe that's overkill. Let me know if it seems like something we'd want to do

test/test_mps.py
kulinseth: Its an important test , I think 1s should be okay

kulinseth (commented):
LGTM

aten/src/ATen/native/mps/operations/Shape.mm
malfet: Is there a reason not to use it unconditionally? I suspect MPSGraph construction overhead for small tensors is probably significant, and perf for medium sized tensors should be the same.

aten/src/ATen/native/mps/operations/Shape.mm
```diff
@@ -57,6 +68,71 @@ static void check_shape_except_dim(const Tensor& first, const Tensor& second, in
...
+    if (input.numel() == 0) {
+      continue;
+    }
+
+    auto& input_params = input_params_list[input_idx];
+
+    input_params.cat_dim_offset = cat_dim_offset;
+    cat_dim_offset += input.size(dimension);
+
+    // Metal can only launch up to MAX_INT threads at one time. If the input has
```
malfet, line 100: Why not launch multiple kernels with offset?

aten/src/ATen/native/mps/kernels/Shape.metal
```diff
@@ -0,0 +1,85 @@
...
+REGISTER_CAT_LARGE_OP_ALL_INPUT_TYPES(half);
+REGISTER_CAT_LARGE_OP_ALL_INPUT_TYPES(bfloat);
+REGISTER_CAT_LARGE_OP_ALL_INPUT_TYPES(int);
+REGISTER_CAT_LARGE_OP_ALL_INPUT_TYPES(long);
+REGISTER_CAT_LARGE_OP_ALL_INPUT_TYPES(short);
+REGISTER_CAT_LARGE_OP_ALL_INPUT_TYPES(char);
+REGISTER_CAT_LARGE_OP_ALL_INPUT_TYPES(uchar);
+REGISTER_CAT_LARGE_OP_ALL_INPUT_TYPES(bool);
+
+REGISTER_CAT_LARGE_OP(float2, float2);
```
malfet, line 81: Why not half2 (for `torch.chalf`)?

aten/src/ATen/native/mps/kernels/Shape.metal
```diff
@@ -0,0 +1,85 @@
...
+      device T_out * output [[buffer(1)]],                           \
+      constant CatLargeSharedParams<> & shared_params [[buffer(2)]], \
+      constant CatLargeInputParams<> & input_params [[buffer(3)]],   \
+      uint tid [[thread_position_in_grid]]);
+
+#define REGISTER_CAT_LARGE_OP_ALL_INPUT_TYPES(T_out) \
+  REGISTER_CAT_LARGE_OP(float, T_out);               \
+  REGISTER_CAT_LARGE_OP(half, T_out);                \
+  REGISTER_CAT_LARGE_OP(bfloat, T_out);              \
+  REGISTER_CAT_LARGE_OP(int, T_out);                 \
```
malfet, line 58: MPS support unsigned types, so why not add variants for uint, ushort and ulong

malfet (approved):
Probably looks fine to me, though I think it would be good to have an implementation that  is more perf-aware and could completely replace MPSGraph.
I guess to achieve that one needs to have fast-path kernel variants for storage-dense tensors and may be just one flavor that supports type-casts (by doing if condition rather than have all possible permutations of the kernels, see example of the copy kernel)

aten/src/ATen/native/mps/operations/Shape.mm
kurtamohler: Sounds good, I'll make a follow-up PR that replaces the MPSGraph impl

aten/src/ATen/native/mps/operations/Shape.mm
kurtamohler: Oh yeah that is simpler, I'll do that

PR #164417, opened 2025-10-01
[doc infra] update docs dependencies

### LINKED ISSUE #164010: [doc infra] `pip install -r docs/requirements.txt` will fail in most contexts

filipviz:
1. Updates `.ci/docker/requirements-docs.txt`, which now uses a single set of dependencies which work for Python 3.10-3.13 (inclusive).
2. Removes `standard-imghdr` (no longer necessary with Sphinx 7.2.6).
3. Adds `ninja`, which is needed for some notebook examples which use `torch.utils.cpp_extension`.

Fixes #164010

### DIFF
--- a/.ci/docker/requirements-docs.txt
+++ b/.ci/docker/requirements-docs.txt
@@ -24,4 +24,4 @@ beautifulsoup4==4.14.2
 python-etcd==0.4.5
 
 # For notebook examples which use torch.utils.cpp_extension
-ninja==1.13.0  
+ninja==1.13.0
--- a/.ci/docker/requirements-docs.txt
+++ b/.ci/docker/requirements-docs.txt
@@ -1,68 +1,27 @@
-sphinx==5.3.0
-#Description: This is used to generate PyTorch docs
-#Pinned versions: 5.3.0
-
-standard-imghdr==3.13.0; python_version >= "3.13"
-#Description: This is needed by Sphinx, so it needs to be added here.
-# The reasons are as follows:
-# 1) This module has been removed from the Python standard library since Python 3.13(https://peps.python.org/pep-0594/#imghdr);
-# 2) The current version of Sphinx (5.3.0) is not compatible with Python 3.13.
-# Once Sphinx is upgraded to a version compatible with Python 3.13 or later, we can remove this dependency.
-
 -e git+https://github.com/pytorch/pytorch_sphinx_theme.git@71e55749be14ceb56e7f8211a9fb649866b87ad4#egg=pytorch_sphinx_theme2
-# TODO: sphinxcontrib.katex 0.9.0 adds a local KaTeX server to speed up pre-rendering
-# but it doesn't seem to work and hangs around idly. The initial thought that it is probably
-# something related to Docker setup. We can investigate this later.
 
-sphinxcontrib.katex==0.8.6
-#Description: This is used to generate PyTorch docs
-#Pinned versions: 0.8.6
-
-sphinxext-opengraph==0.9.1
-#Description: This is used to generate PyTorch docs
-#Pinned versions: 0.9.1
-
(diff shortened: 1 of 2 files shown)


filipviz:
Open questions:
- I removed `sphinxext-opengraph` because it isn't enabled in `docs/source/conf.py`. Do we want the docs to have og metadata? If so, I can reintroduce and enable it.
- `sphinxcontrib-mermaid` is enabled, but there aren't any mermaid diagrams in the docs. Should we remove it, or keep it around in case we add mermaid diagrams in the future?
- Could we remove the `-e` flag for `pytorch_sphinx_theme2`, or is it important for theme maintainers? I ask because using `-e` on a git URL breaks compatibility with some tools (such as `uv`). cc @svekars

Notes:
- Some notebook examples use `torch.utils.cpp_extension.load_inline`. We should document that the user needs to install a C++ compiler - I can update #164011 with this.
- Because `matplotlib` and `tensorboard` are upgraded, `numpy>2` is now supported and a separate `pip install numpy<2` is no longer necessary.
- I replaced `bs4` with the full `beautifulsoup4`.
- `docs/source/_templates/classtemplate.rst` was using `.. autoclass:: {{ name }}`. In older Sphinx versions, `name` added the package prefix when generating autosummary directives. This was technically a bug, and was [changed in Sphinx 7.4.6](https://www.sphinx-doc.org/en/master/changes/7.4.html#id1). To avoid breaking pages like `torch.Tensor.view`, I replaced `name` with `fullname`.

filipviz:
I was mistaken - the Sphinx 7.4.6 change couldn't be the root cause since I only upgraded to Sphinx 7.2.6 here. In any case, the `name` -> `fullname` change is necessary and the functionality was changed at some point after Sphinx 5.3.0.

filipviz:
Hi @malfet - let me know if you'd like any changes to this PR or have questions for me.

PR #164422, opened 2025-10-02
[dynamo][logging] Add most recent bytecode to graph break with torch._dynamo.graph_break() and verbose

fxdawnn:
https://github.com/pytorch/pytorch/issues/162858 The issue described the feature implemented.

This adds to the existing graph break log with the latest 20 (or viable user frame) bytecode instructions. The scenario is when the graph_break happens without errors. It happens during the case when user calling torch._dynamo.graph_break().

Meanwhile, in the testing, one can find that the generated frame based on step() is not deterministic as sometimes it reached the maximum amount, sometimes it generated the less than that. The bytecode generation is python version dependent. Thus, the testing plan excludes the bytecode output but generated the total bytecode line count.

This is a helpful process to understand bytecode transformation, symbolic convert, and convert frame. It is a helpful task to provide hands-on experience with dynamo workflow.

### DIFF
--- a/test/dynamo/test_error_messages.py
+++ b/test/dynamo/test_error_messages.py
@@ -1096,6 +1096,7 @@ Most recent bytecode instructions traced (max 20):
 
         pattern = r"TRACE.*"
         s = munge_exc(records[-1].getMessage(), skip=0)
+        breakpoint()
         matches = re.findall(pattern, s)
         self.assertIn(len(matches), [13, 20])
         # TODO: Checking inconsistent logging output
--- a/test/dynamo/test_error_messages.py
+++ b/test/dynamo/test_error_messages.py
@@ -1085,48 +1085,39 @@ Most recent bytecode instructions traced (max 20):
     @torch._dynamo.config.patch(verbose=True)
     @make_logging_test(graph_breaks=True)
     def test_variable_tracker_bytecode_to_graph_break_fullgraph(self, records):
-        @torch.compile(backend="eager", fullgraph=True)
+        # def test_bytecode_graphbreak_fullgraph(self, x):
         def fn(x):
             y = x + 1
             z = x + y
             torch._dynamo.graph_break()
             return z
 
-        fn(torch.ones(3))
+        # assertRaises suppresses the traceback, so manually catch
+        e = None
+        try:
+            torch.compile(fn, backend="eager", fullgraph=True)(torch.ones(3))
+        except Exception as exn:
+            e = exn
 
-        pattern = r"TRACE.*"
-        s = munge_exc(records[-1].getMessage(), skip=0)
-        breakpoint()
-        matches = re.findall(pattern, s)
-        self.assertIn(len(matches), [13, 20])
-        # TODO: Checking inconsistent logging output
-
-        def post_munge(s):
-            s = re.sub(r"TRACE.*\n", "", s, flags=re.MULTILINE)
-            return re.sub(r"TRACE.*", "", s)
-
+        self.assertIsNotNone(e)
+        msg = "".join(traceback.format_exception(type(e), e, e.__traceback__))
         self.assertExpectedInline(
-            post_munge(s),
-            """\
-Graph break in user code at test_error_messages.py:N
-Graph Break Reason: Call to `torch._dynamo.graph_break()`
+            munge_exc(msg, suppress_suffix=True, skip=0),
+            """Traceback (most recent call last):
+  File "test_error_messages.py", line N, in test_variable_tracker_bytecode_to_graph_break_fullgraph
+    torch.compile(fn, backend="eager", fullgraph=True)(torch.ones(3))
+torch._dynamo.exc.Unsupported: Call to `torch._dynamo.graph_break()`
   Explanation: User-inserted graph break. Message: None
   Hint: Remove the `torch._dynamo.graph_break()` call.
 
   Developer debug context: Called `torch._dynamo.graph_break()` with args `[]`, kwargs `{}`
 
  For more details about this graph break, please visit: https://meta-pytorch.github.io/compile-graph-break-site/gb/gb0025.html
-User code traceback:
-  File "test_error_messages.py", line N, in test_variable_tracker_bytecode_to_graph_break
-    fn(torch.ones(3))
 
-========== most recent `torch.compile` tracing attempt started here ==========
-
-  File "test_error_messages.py", line N, in fn
+from user code:
+   File "test_error_messages.py", line N, in fn
     torch._dynamo.graph_break()
 
-NOTE: the most recent `torch.compile` tracing attempt might not be where you applied `torch.compile`! This is due to how graph breaks are implemented - the optimized code object returned by Dynamo will call another Dynamo-generated resume function and tracing is re-enabled by calling the resume function as a normal Python function, which Dynamo intercepts as a top-level frame.
-Most recent bytecode instructions traced (max 20):
 """,
         )
 
@@ -1143,8 +1134,8 @@ Most recent bytecode instructions traced (max 20):
         fn(torch.ones(3))
 
         pattern = r"TRACE.*"
-        s = munge_exc(records[-1].getMessage(), skip=0)
-        breakpoint()
+        s = munge_exc(records[0].getMessage(), skip=0)
+        # breakpoint()
         matches = re.findall(pattern, s)
         self.assertIn(len(matches), [13, 20])
         # TODO: Checking inconsistent logging output
--- a/test/dynamo/test_error_messages.py
+++ b/test/dynamo/test_error_messages.py
@@ -772,13 +772,15 @@ from user code:
             torch._dynamo.graph_break()
             return f
 
-        torch.compile(fn, backend="eager")()
-
         def post_munge(s):
             return re.sub(r"0x[0-9A-Fa-f]+", "0xmem_addr", s)
 
+        torch.compile(fn, backend="eager")()
+
         self.assertExpectedInline(
-            munge_exc(records[0].getMessage(), suppress_suffix=True, skip=0),
+            post_munge(
+                munge_exc(records[0].getMessage(), suppress_suffix=True, skip=0)
+            ),
             """\
 Graph break in user code at test_error_messages.py:N
 Graph Break Reason: Call to `torch._dynamo.graph_break()`
@@ -1013,7 +1015,7 @@ Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especiall
             "<Internal traceback>\n",
             msg,
         )
-        msg = re.sub(r"TRACE.*\n", "", msg, flags=re.MULTILINE)
+
         self.assertExpectedInline(
             msg,
             """\
@@ -1169,6 +1171,7 @@ TRACE CALL 0 [NullVariable, LazyVariableTracker()]""",
         matches = re.findall(pattern, s)
         self.assertEqual((len(matches) > 10), True)
         self.assertEqual((len(matches) <= 20), True)
+        self.assertIn("Most recent bytecode instructions traced (max 20):", s)
 
     @torch._dynamo.config.patch(verbose=True)
     @make_logging_test(graph_breaks=True)
--- a/test/dynamo/test_error_messages.py
+++ b/test/dynamo/test_error_messages.py
@@ -1140,17 +1140,17 @@ User code traceback:
 NOTE: the most recent `torch.compile` tracing attempt might not be where you applied `torch.compile`! This is due to how graph breaks are implemented - the optimized code object returned by Dynamo will call another Dynamo-generated resume function and tracing is re-enabled by calling the resume function as a normal Python function, which Dynamo intercepts as a top-level frame.
 Most recent bytecode instructions traced (max 20):
 TRACE RESUME 0 []
-TRACE LOAD_FAST x []
+TRACE LOAD_FAST 'x' []
 TRACE LOAD_CONST 1 [LazyVariableTracker()]
 TRACE BINARY_OP 0 [LazyVariableTracker(), ConstantVariable(int: 1)]
-TRACE STORE_FAST y [TensorVariable()]
-TRACE LOAD_FAST x []
-TRACE LOAD_FAST y [TensorVariable()]
+TRACE STORE_FAST 'y' [TensorVariable()]
+TRACE LOAD_FAST 'x' []
+TRACE LOAD_FAST 'y' [TensorVariable()]
 TRACE BINARY_OP 0 [TensorVariable(), TensorVariable()]
-TRACE STORE_FAST z [TensorVariable()]
-TRACE LOAD_GLOBAL torch []
-TRACE LOAD_ATTR _dynamo [LazyVariableTracker()]
-TRACE LOAD_ATTR graph_break [LazyVariableTracker()]
+TRACE STORE_FAST 'z' [TensorVariable()]
+TRACE LOAD_GLOBAL 'torch' []
+TRACE LOAD_ATTR '_dynamo' [LazyVariableTracker()]
+TRACE LOAD_ATTR 'graph_break' [LazyVariableTracker()]
 TRACE CALL 0 [NullVariable, LazyVariableTracker()]""",
         )
 
--- a/test/dynamo/test_error_messages.py
+++ b/test/dynamo/test_error_messages.py
@@ -1085,7 +1085,6 @@ Most recent bytecode instructions traced (max 20):
     @torch._dynamo.config.patch(verbose=True)
     @make_logging_test(graph_breaks=True)
     def test_variable_tracker_bytecode_to_graph_break_fullgraph(self, records):
-        # def test_bytecode_graphbreak_fullgraph(self, x):
         def fn(x):
             y = x + 1
             z = x + y
@@ -1190,36 +1189,6 @@ TRACE CALL 0 [NullVariable, LazyVariableTracker()]""",
         # breakpoint()
         matches = re.findall(pattern, s)
         self.assertEqual(len(matches), 13)
-        # TODO: Checking inconsistent logging output
-
-        def post_munge(s):
-            s = re.sub(r"TRACE.*\n", "", s, flags=re.MULTILINE)
-            return re.sub(r"TRACE.*", "", s)
-
-        self.assertExpectedInline(
-            post_munge(s),
-            """\
-Graph break in user code at test_error_messages.py:N
-Graph Break Reason: Call to `torch._dynamo.graph_break()`
-  Explanation: User-inserted graph break. Message: None
-  Hint: Remove the `torch._dynamo.graph_break()` call.
-
-  Developer debug context: Called `torch._dynamo.graph_break()` with args `[]`, kwargs `{}`
-
- For more details about this graph break, please visit: https://meta-pytorch.github.io/compile-graph-break-site/gb/gb0025.html
-User code traceback:
-  File "test_error_messages.py", line N, in test_variable_tracker_bytecode_to_graph_break
-    fn(torch.ones(3))
-
-========== most recent `torch.compile` tracing attempt started here ==========
-
-  File "test_error_messages.py", line N, in fn
-    torch._dynamo.graph_break()
-
-NOTE: the most recent `torch.compile` tracing attempt might not be where you applied `torch.compile`! This is due to how graph breaks are implemented - the optimized code object returned by Dynamo will call another Dynamo-generated resume function and tracing is re-enabled by calling the resume function as a normal Python function, which Dynamo intercepts as a top-level frame.
-Most recent bytecode instructions traced (max 20):
-""",
-        )
 
     @torch._dynamo.config.patch(verbose=True)
     @make_logging_test(graph_breaks=True)
--- a/test/dynamo/test_error_messages.py
+++ b/test/dynamo/test_error_messages.py
@@ -772,17 +772,13 @@ from user code:
             torch._dynamo.graph_break()
             return f
 
-        def post_munge(s):
-            s = re.sub(r"TRACE.*\n", "", s, flags=re.MULTILINE)
-            s = re.sub(r"\nTRACE.*", "", s)
-            return re.sub(r"0x[0-9A-Fa-f]+", "0xmem_addr", s)
-
         torch.compile(fn, backend="eager")()
 
+        def post_munge(s):
+            return re.sub(r"0x[0-9A-Fa-f]+", "0xmem_addr", s)
+
         self.assertExpectedInline(
-            post_munge(
-                munge_exc(records[0].getMessage(), suppress_suffix=True, skip=0)
-            ),
+            munge_exc(records[0].getMessage(), suppress_suffix=True, skip=0),
             """\
 Graph break in user code at test_error_messages.py:N
 Graph Break Reason: Call to `torch._dynamo.graph_break()`
@@ -797,7 +793,7 @@ User code traceback:
     torch.compile(fn, backend="eager")()
   File "test_error_messages.py", line N, in fn
     torch._dynamo.graph_break()
-Most recent bytecode instructions traced (max 20):""",
+""",
         )
 
         self.assertExpectedInline(
@@ -1054,12 +1050,8 @@ from user code:
 
         torch.compile(fn, backend="eager")(torch.randn(3))
 
-        def post_munge(s):
-            s = re.sub(r"TRACE.*\n", "", s, flags=re.MULTILINE)
-            return re.sub(r"TRACE.*", "", s)
-
         self.assertExpectedInline(
-            post_munge(munge_exc(records[-1].getMessage(), skip=0)),
+            munge_exc(records[-1].getMessage(), skip=0),
             """\
 Graph break in user code at test_error_messages.py:N
 Graph Break Reason: Call to `torch._dynamo.graph_break()`
@@ -1078,13 +1070,12 @@ User code traceback:
     hn(x + 1)
   File "test_error_messages.py", line N, in hn
     torch._dynamo.graph_break()  # 1
-Most recent bytecode instructions traced (max 20):
 """,
         )
 
     @torch._dynamo.config.patch(verbose=True)
     @make_logging_test(graph_breaks=True)
-    def test_variable_tracker_bytecode_to_graph_break_fullgraph(self, records):
+    def test_latest_bytecode_to_graph_break_fullgraph(self, records):
         def fn(x):
             y = x + 1
             z = x + y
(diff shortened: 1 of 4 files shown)


torch/_dynamo/symbolic_convert.py
```diff
@@ -1341,6 +1351,9 @@ def step(self) -> bool:
                 "TRACE %s %s %s", inst.opname, inst.argval, self.stack
             )
 
+        # Store the latest 20 bytecode execution for the process
+        latest_bytecode_queue.append(f"TRACE {inst.opname} {inst.argval} {self.stack}")
```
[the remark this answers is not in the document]
williamwen42: Please follow this guidance.

williamwen42: I think it would be helpful to include trace_source lines from the last 20 traced bytecodes too - this can make it more easy to tell where the bytecode is coming from.

torch/_dynamo/symbolic_convert.py
```diff
@@ -607,6 +612,10 @@ def log_graph_break(
         # This log line MUST contain the string "Graph break in user code",
         # This log line is exercised from
         #   python test/dynamo/test_exc.py -k test_graph_break_log
+        if latest_bytecode_log:
```
williamwen42: Need some more opinions from @anijain2305 @StrongerXi @Lucaskabela - would this be too spammy? Should we only output this when TORCHDYNAMO_VERBOSE=1?

test/dynamo/test_exc.py
```diff
@@ -170,11 +171,13 @@ def fn001(x):
         torch.compile(fn001, backend="eager")(torch.randn(1))
 
         record = self.getRecord(records, "Graph break in user code")
+        msg = re.sub(r"TRACE.*\n", "", record.getMessage(), flags=re.MULTILINE)
+        # msg =
```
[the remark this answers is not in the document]
williamwen42: Please follow this guidance

test/dynamo/test_error_messages.py
```diff
@@ -1072,6 +1078,54 @@ def hn(x):
...
+            z = x + y
+            torch._dynamo.graph_break()
+            return z
+
+        fn(torch.ones(3))
+
+        pattern = r"TRACE.*"
+        s = munge_exc(records[-1].getMessage(), skip=0)
+        matches = re.findall(pattern, s)
+        self.assertIn(len(matches), [13, 20])
```
williamwen42: Why 13 or 20?

test/dynamo/test_error_messages.py
```diff
@@ -1072,6 +1078,54 @@ def hn(x):
...
+    fn(torch.ones(3))
+
+========== most recent `torch.compile` tracing attempt started here ==========
+
+  File "test_error_messages.py", line N, in fn
+    torch._dynamo.graph_break()
+
+NOTE: the most recent `torch.compile` tracing attempt might not be where you applied `torch.compile`! This is due to how graph breaks are implemented - the optimized code object returned by Dynamo will call another Dynamo-generated resume function and tracing is re-enabled by calling the resume function as a normal Python function, which Dynamo intercepts as a top-level frame.
+Most recent bytecode instructions traced (max 20):
 """,
```
williamwen42: I think it would be good to test for the exact bytecodes logged for a particular python version - e.g. 3.12, just so the format as a whole is written somewhere in the file.

williamwen42 (commented):
Overall looks good! Just have a few comments.

torch/_dynamo/symbolic_convert.py
fxdawnn: That sounds like a good follow-up PR!

torch/_dynamo/symbolic_convert.py
fxdawnn: I believe the current setup is that it's only triggered when the graph_break is called by user.

torch/_dynamo/symbolic_convert.py
```diff
@@ -208,6 +209,9 @@
 compare_op_handlers["not in"] = lambda tx, args, _: handle_not(
     tx, [handle_contains(tx, [*reversed(args)], {})], {}
 )
+latest_bytecode_queue: deque[str] = deque(
```
williamwen42: This global is causing your 13/20 issue I believe. Can you move this queue to be a member of InstructionTranslator?

fxdawnn: Great catch! Thank you!

torch/_dynamo/symbolic_convert.py
fxdawnn: For example, TEST TEST_ERROR_MESSAGE.test_data_dependent_branching_gb(self, records) test_graph_break_traceback_collapsed_resume_frames and test_reconstruction_failure_gb shows that it's only the scenario when the user call torch._dynamo.graph_break(). It does not matter that TORCHDYNAMO_VERBOSE=1

But this is slightly different than the original github issue #162858
 which only cares if user call torch._dynamo.graph_break(), so I'll keep it as it is?

torch/_dynamo/symbolic_convert.py
fxdawnn: I don't think it was resolved but I've changed the queue to be a member of instrctionTranslator.

test/dynamo/test_error_messages.py
fxdawnn: Not sure this can be done in test file to control the output -- but I've attached the log that was generated.
Screenshot 2025-10-03 at 12 57 04 PM

torch/_dynamo/symbolic_convert.py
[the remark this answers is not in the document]
fxdawnn: It's wrong.... i tried and it broke the file. So I'm keeping the original form for now?

test/dynamo/test_error_messages.py
fxdawnn: I've added to #todo for this to be explored. But currently, the @make_logging_test(graph_breaks=True) run through the tests twice and each time leads to a different number of log within the same python environment. This does not resolved after the queue is move to a local InstructorTranslator local environment. 

I've attached the file for pdb printing the log output when it's going through test twice. Thanks for helping with setting up the testing environment yesterday. Run: python test/dynamo/test_error_messages.py ErrorMessagesTest.test_variable_tracker_bytecode_to_graph_break and the test would generate two different logs.
Screenshot 2025-10-03 at 1 11 06 PM
Screenshot 2025-10-03 at 1 11 27 PM

test/dynamo/test_error_messages.py
williamwen42: you can do an assertExpectedInline with `if sys.version_info >= (3, 12) and sys.version_info < (3, 13)`

torch/_dynamo/symbolic_convert.py
williamwen42: #162858 is asking for support for all graph breaks, not just `torch._dynamo.graph_break()` (but `torch._dynamo.graph_break()` isn't really handled that specially by Dynamo anyway, so if it works for `torch._dynamo.graph_break()`, it should work for all graph breaks).

In order to only log this when `TORCHDYNAMO_VERBOSE=1`, you should guard this logging by checking `config.verbose`.

StrongerXi: > would this be too spammy? Should we only output this when TORCHDYNAMO_VERBOSE=1?

If it always shows up in user-facing graph break error (tlparse or just regular error from `fullgraph=True`, then yeah I think it's too spammy.

torch/_dynamo/symbolic_convert.py
```diff
@@ -1174,6 +1181,8 @@ class InstructionTranslatorBase(
     parent: Optional[InstructionTranslatorBase]
     debug_locals: list[tuple[VariableTracker, list[VariableTracker]]]
     package: Optional[CompilePackage]
+    latest_bytecode_queue: deque[str] = deque(maxlen=20)
```
williamwen42: This is improperly initialized - every instance of InstructionTranslatorBase will refer to the same deque - this is why you see the 13, 20 problem.

test/dynamo/test_error_messages.py
```diff
@@ -1072,6 +1078,55 @@ def hn(x):
     hn(x + 1)
   File "test_error_messages.py", line N, in hn
     torch._dynamo.graph_break()  # 1
+Most recent bytecode instructions traced (max 20):
+""",
+        )
+
+    @torch._dynamo.config.patch(verbose=True)
+    @make_logging_test(graph_breaks=True)
+    def test_variable_tracker_bytecode_to_graph_break(self, records):
```
williamwen42: Could you also test for `torch.compile(backend="eager", fullgraph=True)`

fxdawnn: Would you elaborate the output behavior of log once the scenario is set to torch.compile(backend="eager", fullgraph=True)? Thanks!

fxdawnn: Will be added for the next PR.

torch/_dynamo/symbolic_convert.py
fxdawnn: Sounds good! I think we're changing the graph breaks log to the torch._dynamo.graph_break(). We could discuss more and add the follow-up on generalize it to a broader scenario.

torch/_dynamo/symbolic_convert.py
fxdawnn: Fixed! Thank you!

test/dynamo/test_error_messages.py
fxdawnn: Tested. Thank you!

torch/_dynamo/symbolic_convert.py
fxdawnn: Fixed. Thank you!

test/dynamo/test_error_messages.py
fxdawnn: Fixed! Added test. Thank you!

torch/_dynamo/symbolic_convert.py
fxdawnn: Resolved. Only logged when verbose is true.

test/dynamo/test_exc.py
```diff
@@ -170,11 +171,12 @@ def fn001(x):
         torch.compile(fn001, backend="eager")(torch.randn(1))
 
         record = self.getRecord(records, "Graph break in user code")
+        msg = re.sub(r"TRACE.*\n", "", record.getMessage(), flags=re.MULTILINE)
```
williamwen42: Are these changes still needed if we only log bytecode when config.verbose is set?

test/dynamo/test_error_messages.py
```diff
@@ -1072,6 +1078,146 @@ def hn(x):
...
+ For more details about this graph break, please visit: https://meta-pytorch.github.io/compile-graph-break-site/gb/gb0025.html
+
+from user code:
+   File "test_error_messages.py", line N, in fn
+    torch._dynamo.graph_break()
+
+""",
+        )
+
+    @skipIfNotPy312
```
williamwen42: This makes the test run in python >= 3.12 - we need to make sure this doesn't run on 3.13+ (because bytecode will be different)

test/dynamo/test_error_messages.py
```diff
@@ -773,6 +773,8 @@ def fn():
             return f
 
         def post_munge(s):
+            s = re.sub(r"TRACE.*\n", "", s, flags=re.MULTILINE)
```
williamwen42: Is this still needed? (also other similar subs below where config.verbose is not set)

test/dynamo/test_error_messages.py
```diff
@@ -1072,6 +1078,146 @@ def hn(x):
...
   File "test_error_messages.py", line N, in hn
     torch._dynamo.graph_break()  # 1
+Most recent bytecode instructions traced (max 20):
+""",
+        )
+
+    @torch._dynamo.config.patch(verbose=True)
+    @make_logging_test(graph_breaks=True)
+    def test_variable_tracker_bytecode_to_graph_break_fullgraph(self, records):
+        # def test_bytecode_graphbreak_fullgraph(self, x):
```
williamwen42: delete

test/dynamo/test_error_messages.py
```diff
@@ -1072,6 +1078,146 @@ def hn(x):
...
+        def fn(x):
+            y = x + 1
+            z = x + y
+            torch._dynamo.graph_break()
+            return z
+
+        # assertRaises suppresses the traceback, so manually catch
```
williamwen42: Can you use `assertExpectedInlineMunged` here?

torch/_dynamo/symbolic_convert.py
```diff
@@ -4655,6 +4670,7 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.generated_items = []
         self.generator_exhausted = False
         self.is_generator_from_ctx_manager = False
+        self.latest_bytecode_queue = Queue
```
williamwen42: Fix (also need to set `latest_bytecode_queue` of `InliningInstructionTranslator` as well?)

torch/_dynamo/symbolic_convert.py
williamwen42: Don't need to set this value if you're setting it again in the __init__

torch/_dynamo/symbolic_convert.py
```diff
@@ -1174,6 +1181,8 @@ class InstructionTranslatorBase(
     parent: Optional[InstructionTranslatorBase]
     debug_locals: list[tuple[VariableTracker, list[VariableTracker]]]
     package: Optional[CompilePackage]
+    latest_bytecode_queue: deque[str] = deque(maxlen=20)
+    # Store the latest bytecode before graph_break() call by user
```
williamwen42, line 1195: Move this comment down to the __init__ (for consistency)

test/dynamo/test_error_messages.py
```diff
@@ -1072,6 +1078,146 @@ def hn(x):
...
+  Explanation: User-inserted graph break. Message: None
+  Hint: Remove the `torch._dynamo.graph_break()` call.
+
+  Developer debug context: Called `torch._dynamo.graph_break()` with args `[]`, kwargs `{}`
+
+ For more details about this graph break, please visit: https://meta-pytorch.github.io/compile-graph-break-site/gb/gb0025.html
+
+from user code:
+   File "test_error_messages.py", line N, in fn
+    torch._dynamo.graph_break()
```
williamwen42, line 1101: are you checking for bytecode here? or this will be a followup? if it's a followup, can you add this test there instead?

test/dynamo/test_error_messages.py
```diff
@@ -1072,6 +1078,146 @@ def hn(x):
...
+TRACE STORE_FAST z [TensorVariable()]
+TRACE LOAD_GLOBAL torch []
+TRACE LOAD_ATTR _dynamo [LazyVariableTracker()]
+TRACE LOAD_ATTR graph_break [LazyVariableTracker()]
+TRACE CALL 0 [NullVariable, LazyVariableTracker()]""",
+        )
+
+    @torch._dynamo.config.patch(verbose=True)
+    @make_logging_test(graph_breaks=True)
+    def test_variable_tracker_bytecode_to_graph_break(self, records):
```
williamwen42: the test name is a little confusing - can we rename to something like `test_most_recent_bytecode_logging_on_graph_break` (and similarly with the other test names)

test/dynamo/test_error_messages.py
```diff
@@ -1072,6 +1078,146 @@ def hn(x):
...
+========== most recent `torch.compile` tracing attempt started here ==========
+
+  File "test_error_messages.py", line N, in fn
+    torch._dynamo.graph_break()
+
+NOTE: the most recent `torch.compile` tracing attempt might not be where you applied `torch.compile`! This is due to how graph breaks are implemented - the optimized code object returned by Dynamo will call another Dynamo-generated resume function and tracing is re-enabled by calling the resume function as a normal Python function, which Dynamo intercepts as a top-level frame.
+Most recent bytecode instructions traced (max 20):
+TRACE RESUME 0 []
+TRACE LOAD_FAST x []
+TRACE LOAD_CONST 1 [LazyVariableTracker()]
```
williamwen42, line 1144: I have some concerns that Dynamo might change stack representation output (e.g. with your improved lazy VT logging change) - but this doesn't seem like a super big deal. Either we need to update this test when the representation changes (which could be annoying), or use regex to clear the contents of the stack (use your judgment here).

test/dynamo/test_error_messages.py
```diff
@@ -1072,6 +1078,146 @@ def hn(x):
...
+            return z
+
+        fn(torch.ones(3))
+
+        pattern = r"TRACE.*"
+        s = munge_exc(records[0].getMessage(), skip=0)
+        # breakpoint()
+        matches = re.findall(pattern, s)
+        self.assertEqual(len(matches), 13)
+        # TODO: Checking inconsistent logging output
```
williamwen42: remove

test/dynamo/test_exc.py
fxdawnn: I think this is for the purpose of testing the number of lines of output. But it is set to be when config.verbose is set scenario.

test/dynamo/test_error_messages.py
fxdawnn: yes, this was already at the end of the file.

test/dynamo/test_error_messages.py
fxdawnn: Would this be risky with new version of python once they're available?

test/dynamo/test_error_messages.py
fxdawnn: fixed.

test/dynamo/test_error_messages.py
fxdawnn: I'm hoping to merge this back before working on that one. But you're right. This will have to be update once the lazyvariabletracker is updated.

test/dynamo/test_error_messages.py
fxdawnn: updated.

test/dynamo/test_error_messages.py
fxdawnn: fixed

test/dynamo/test_error_messages.py
fxdawnn: removed

test/dynamo/test_exc.py
fxdawnn: you're right. i removed.

test/dynamo/test_error_messages.py
williamwen42: This is not risky - we just need to keep testing it on some currently supported Python version since your changes aren't Python version specific. I hope we can remember to do this once 3.12 gets deprecated 😅

test/dynamo/test_error_messages.py
williamwen42: Oh, I'm referring to the code below that does the try/except and the whole `msg = "".join(traceback.format_exception(type(e), e, e.__traceback__))` - I think that's only necessary because the test using it needed to look at the actual traceback contents. Your PR doesn't change the traceback contents, so I don't think you need to do this? Just a clean `assertExpectedInlineMunged` call like most tests in this file.

fxdawnn: sounds good. Fix!

test/dynamo/test_error_messages.py
fxdawnn: Gotcha! I update to only 3.12 now.

torch/_dynamo/testing.py
```diff
@@ -505,6 +505,12 @@ def skipIfNotPy312(fn: Callable[_P, _T]) -> Callable[_P, _T]:
     return unittest.skip("Requires Python 3.12+")(fn)
 
 
+def skipIfOnlyNotPy312(fn: Callable[_P, _T]) -> Callable[_P, _T]:
+    if sys.version_info == (3, 12):
```
williamwen42: This is not correct since `sys.version_info` is a tuple of more than 2 elements (e.g. `(3, 12, 5)`) - I think you can probably just use `@unittest.skipIf` in test_error_messages instead of defining a new decorator since we don't anticipate many 3.12 only tests.

torch/_dynamo/symbolic_convert.py
```diff
@@ -1353,7 +1353,7 @@ def step(self) -> bool:
         # Store the latest 20 bytecode execution for the process
         if isinstance(inst.argval, bytes):
             self.latest_bytecode_queue.append(
-                f"TRACE {inst.opname} {inst.argval.decode('utf-8')} {self.stack}"
+                f"TRACE {inst.opname} {inst.argval.decode('utf-8', errors='ignore')} {self.stack}"
```
williamwen42: hmm, why is a utf decode needed here, but not above?

test/dynamo/test_error_messages.py
```diff
@@ -772,15 +772,13 @@ def fn():
...
         def post_munge(s):
             return re.sub(r"0x[0-9A-Fa-f]+", "0xmem_addr", s)
 
-        torch.compile(fn, backend="eager")()
-
         self.assertExpectedInline(
-            post_munge(
-                munge_exc(records[0].getMessage(), suppress_suffix=True, skip=0)
-            ),
+            munge_exc(records[0].getMessage(), suppress_suffix=True, skip=0),
```
williamwen42: Why was this changed?

test/dynamo/test_error_messages.py
```diff
@@ -1015,6 +1013,7 @@ def gn():
             "<Internal traceback>\n",
             msg,
         )
+        msg = re.sub(r"TRACE.*\n", "", msg, flags=re.MULTILINE)
```
williamwen42: Not needed anymore?

test/dynamo/test_error_messages.py
```diff
@@ -1075,6 +1073,103 @@ def hn(x):
...
+            torch._dynamo.graph_break()
+            return z
+
+        fn(torch.ones(3))
+
+        pattern = r"TRACE.*"
+        s = munge_exc(records[0].getMessage(), skip=0)
+        matches = re.findall(pattern, s)
+        self.assertEqual((len(matches) > 10), True)
+        self.assertEqual((len(matches) <= 20), True)
```
williamwen42, line 1173: Let's also check for `Most recent bytecode instructions traced (max 20):`?

torch/_dynamo/symbolic_convert.py
fxdawnn: i do not know. CI failing in one instance.

test/dynamo/test_error_messages.py
fxdawnn: What do you mean? You asked for these changes??

williamwen42: As in why was the post_munge call removed when it was previously there? (and less important, but the torch.compile call was also moved)

test/dynamo/test_error_messages.py
fxdawnn: yes. fixed.

test/dynamo/test_error_messages.py
fxdawnn: fixed

test/dynamo/test_error_messages.py
fxdawnn: my bad. Must accidentally changed these lines.

torch/_dynamo/symbolic_convert.py
fxdawnn: changed it back.

torch/_dynamo/symbolic_convert.py
```diff
@@ -1341,6 +1350,11 @@ def step(self) -> bool:
                 "TRACE %s %s %s", inst.opname, inst.argval, self.stack
             )
 
+        # Store the latest 20 bytecode execution for the process
+        self.latest_bytecode_queue.append(
+            f"TRACE {inst.opname} {str(inst.argval)} {self.stack}"
```
williamwen42: I see the error - we should only append if config.verbose is set - that should make the dynamo_wrapped errors go away. Is the `TRACE %s` above also susceptible to this error? Perhaps we should fix both sites with a decode if necessary?

fxdawnn: ok i found another way to fix! Thanks!

fxdawnn: I tried to use decode earlier but that led to more errors so trying repr() now.

torch/_dynamo/symbolic_convert.py
```diff
@@ -1341,6 +1350,11 @@ def step(self) -> bool:
                 "TRACE %s %s %s", inst.opname, inst.argval, self.stack
```
williamwen42, line 1360: Can we also do `repr(inst.argval)` here as well to prevent future breaks?

williamwen42 (approved):
Looks good! Thanks for the work!

torch/_dynamo/symbolic_convert.py
fxdawnn: Great idea! Will follow up with a smaller PR for this.

mlazos (approved):

torch/_dynamo/symbolic_convert.py
```diff
@@ -1351,6 +1360,17 @@ def step(self) -> bool:
...
             )
 
+        # Store the latest 20 bytecode execution for the process,
+        # Used repr for byte processing and limiting the length to 2048
+        try:
+            stack_repr = repr(self.stack)
```
guilhermeleobas, line 1366: I believe this PR made dynamo considerably slower. `test_ziplongest` is almost 2x slower when compared to https://github.com/pytorch/pytorch/commit/fa919feab6a

```
                              Comparison Results
┏━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┳━━━━━━━━━━━━━━━┳━━━━━━━━━━┳━━━━━━━━━━━━━━━━━━┓
┃ Benchmark                    ┃ Baseline Time ┃ New Time ┃ Speedup/Slowdown ┃
┡━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╇━━━━━━━━━━━━━━━╇━━━━━━━━━━╇━━━━━━━━━━━━━━━━━━┩
│ TestBasicOps.test_ziplongest │ 45.5060s      │ 80.6820s │            1.77x │
└──────────────────────────────┴───────────────┴──────────┴──────────────────┘
```

Could this try/except be wrapped in a `if config.verbose` conditional?

williamwen42: cc @fxdawnn maybe we could also mirror the stack with its contents' string representation instead of repr'ing the stack each time (we would probably need to do this if we want to log this information to tlparse every time)

fxdawnn: Using the str representation caused some CI failure earlier. Should we still revert to none? I doubt if this is the cause since this is only triggered when users are calling ``torch._dynamo.graph_break()`` in their code. But this doesn't seem to be the case?

fxdawnn: Right. Will update the bytecode queue soon.

fxdawnn: See PR https://github.com/pytorch/pytorch/pull/165926

PR #164427, opened 2025-10-02
[inductor] Add flag to run post-grad graph

angelayi:
Adding an inductor config and envvar `TORCHINDUCTOR_RUN_WITH_POST_GRAD` so that when specified, instead of running the code with inductor's generated python code, we just run the post-grad graph eagerly. This is helpful when debugging VLLM to rule out any issues due to inductor's codegen (it's a little complicated to specify backend="aot_eager_..." in the context of VLLM).

### DIFF
--- a/torch/_inductor/compile_fx.py
+++ b/torch/_inductor/compile_fx.py
@@ -13,7 +13,7 @@ import sys
 import time
 import warnings
 from abc import ABC, abstractmethod
-from collections import defaultdict
+from collections import Counter, defaultdict
 from contextlib import AbstractContextManager
 from dataclasses import dataclass
 from inspect import currentframe
@@ -640,6 +640,52 @@ def is_tf32_warning_applicable(gm: GraphModule) -> bool:
             ):
                 return True
     return False
+
+
+def create_post_grad_runnable(
+    gm: torch.fx.GraphModule,
+    *,
+    static_input_idxs: Sequence[int],
+    fx_kwargs: _CompileFxKwargs,
+    inputs_to_check: Sequence[int],
+    runnable_graph_str: str,
+    inductor_post_grad_graph_str: str,
+) -> CompiledFxGraph:
+    """
+    Creates a CompiledFxGraph given a graph module. This is used to directly run
+    with the post-grad graph.
+    """
+
+    def compiled_fn(args):  # type: ignore[no-untyped-def]
+        return torch.fx.Interpreter(gm).boxed_run(args)
+
+    compiled_fn._boxed_call = True  # type: ignore[attr-defined]
+
+    dummy_metrics_deltas = metrics.CachedMetricsDeltas(
+        generated_kernel_count=0,
+        generated_cpp_vec_kernel_count=0,
+        ir_nodes_pre_fusion=0,
+        cpp_to_dtype_count=0,
+        num_bytes_accessed=0,
+        num_matches_for_scatter_upon_const_tensor=0,
+    )
+
+    return CompiledFxGraph(
+        current_callable=compiled_fn,
+        graph=GraphLowering(gm),  # type: ignore[arg-type]
+        gm=gm,
+        output_strides=[],
+        disabled_cudagraphs_reason="Running with post-grad graph",
+        metrics_deltas=dummy_metrics_deltas,
+        counter_deltas=Counter(),
+        cudagraphs=BoxedBool(False),
+        example_inputs=[],
+        static_input_idxs=static_input_idxs,
+        fx_kwargs=fx_kwargs,
+        inputs_to_check=inputs_to_check,
+        runnable_graph_str=runnable_graph_str,
+        inductor_post_grad_graph_str=inductor_post_grad_graph_str,
+    )
 
 
 def maybe_disable_comprehensive_padding(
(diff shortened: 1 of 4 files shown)


torch/_inductor/config.py
```diff
@@ -1078,6 +1078,14 @@ def decide_compile_threads() -> int:
 # Lookup table for overriding autotune configs based on hash of Triton source code
 autotune_lookup_table: dict[str, dict[str, Any]] = {}
 
+# This is used for debugging purposes in JIT Inductor where you just want to
+# check the graph after post-grad passes. Instead of running with Inductor's
+# generated python code + generated triton kernels, we will just run with the
+# post-grad graph.
+run_with_post_grad_graph: bool = (
```
jansel: Is this only useful for debugging?  Maybe we should give this a name like `debug_return_post_grad_graph` so users don't enable it by mistake.

Does this need to be part of inductor? Couldn't you construct this directly by calling aot_autograd directly?  For example, a different backend similar to `aot_eager`.

If you can't specify a backend, you could monkey patch out inductor's compile.

jansel (changes requested):

torch/_inductor/config.py
angelayi: Ah I think by constructing this directly with aot_autograd, I would have to mimic the same pre_grad/joint_graph/post_grad behavior that inductor has, and there might be some divergence in the end. By directly putting it in inductor I can just exit inductor early.

torch/_inductor/fx_passes/post_grad.py
```diff
@@ -816,9 +816,8 @@ def scatter_upon_const_tensor(
 
     # Check if inputs are tensors instead of inductor IR nodes
     if isinstance(selector, torch.Tensor):
-        # Return a fake tensor with the proper shape that this operator is intended to return
-        device = selector.device if hasattr(selector, "device") else torch.device("cpu")
-        return torch.empty(shape, dtype=dtype, device=device)
+        background_tensor = torch.full(shape, background_val, dtype=dtype)
+        return torch.scatter(background_tensor, dim, selector, val)
```
angelayi, line 821: the previous implementation caused a numerical issue in https://github.com/pytorch/torchtitan/pull/1785

ruisizhang123: I think we would still need a device here to avoid device mismatch error when running on GPU

```python
device = selector.device if hasattr(selector, "device") else torch.device("cpu")
background_tensor = torch.full(shape, background_val, dtype=dtype, device=device)
return torch.scatter(background_tensor, dim, selector, val)
```

torch/_inductor/fx_passes/post_grad.py
```diff
@@ -816,9 +816,9 @@ def scatter_upon_const_tensor(
 
     # Check if inputs are tensors instead of inductor IR nodes
     if isinstance(selector, torch.Tensor):
-        # Return a fake tensor with the proper shape that this operator is intended to return
         device = selector.device if hasattr(selector, "device") else torch.device("cpu")
-        return torch.empty(shape, dtype=dtype, device=device)
+        background_tensor = torch.full(shape, background_val, dtype=dtype, device=device)
```
shunting314, line 820: Curious why you see a selector as a torch.Tensor rather than IRNode?

shunting314 (commented):
Is the behavior very similar to fallback all ops? @eellison

torch/_inductor/fx_passes/post_grad.py
angelayi: Ah this is because I'm directly returning the post-grad graph and running it, so inputs to this function would be tensors not inductor ir

shunting314: I'm a bit worried that similar thing may need to be done for all lowering patterns. They are not supposed to work on torch.Tensor when they are first time created.

A safer way to do what you want is probably fallback all ops which can be done by the bisector Elias build previously

angelayi: What's the config to specify to fallback all ops?

angelayi: got it from Elias -- `TORCH_BISECT_BACKEND=inductor TORCH_BISECT_SUBSYSTEM=lowerings TORCH_BISECT_MAX=-1`

angelayi:
closing - since some of the post-grad passes have inductor-lowering specific logic, it's not necessarily safe to directly return the post grad pass. Instead we can disable inductor's fusion logic by telling inductor to fallback on all operators. To do this, we can set the flags `TORCH_BISECT_BACKEND=inductor TORCH_BISECT_SUBSYSTEM=lowerings TORCH_BISECT_MAX=-1`

torch/_inductor/compile_fx.py
```diff
@@ -642,6 +642,52 @@ def is_tf32_warning_applicable(gm: GraphModule) -> bool:
...
+    runnable_graph_str: str,
+    inductor_post_grad_graph_str: str,
+) -> CompiledFxGraph:
+    """
+    Creates a CompiledFxGraph given a graph module. This is used to directly run
+    with the post-grad graph.
+    """
+
+    def compiled_fn(args):  # type: ignore[no-untyped-def]
+        return torch.fx.Interpreter(gm).boxed_run(args)
```
anijain2305, line 660: For performance issue - try replacing this line with

```
return gm(list(args))
```

angelayi: I think we also tried this but @ruisizhang123 didnt notice an improvement

PR #164435, opened 2025-10-02
[Max Autotune][B200] Skip carveout tests

jananisriram:
Summary: Skip sm `carveout` tests on B200, as carveout is currently unsupported.

Test Plan:
```
buck2 test 'fbcode//mode/opt' fbcode//caffe2/test/inductor:max_autotune -c fbcode.nvcc_arch=b200a -c fbcode.enable_gpu_sections=true -c fbcode.platform010_cuda_version=12.8 -c fbcode.re_gpu_tests=False -- test_honor_sm_carveout_with_triton_tma
```

Differential Revision: D83395610

### DIFF
--- a/test/inductor/test_max_autotune.py
+++ b/test/inductor/test_max_autotune.py
@@ -799,6 +799,9 @@ class TestMaxAutotune(TestCase):
     @unittest.skipIf(
         not has_triton_tma_device(), "Need device-side TMA support in Triton"
     )
+    @unittest.skipIf(
+        has_datacenter_blackwell_tma_device(), "B200 doesn't support sm carveout"
+    )
     @parametrize("carveout", (None, 0, 27))
     @parametrize("op", ("mm", "scaled_mm"))
     def test_honor_sm_carveout_with_triton_tma(self, carveout, op: str):


torch/testing/_internal/common_utils.py
```diff
@@ -1405,6 +1405,24 @@ def printErrors(self) -> None:
...
 
+def get_nvidia_gpu_model() -> str:
+    """
+    Retrieves the model of the NVIDIA GPU being used.
+    Will return the name of the first GPU listed.
+    Returns:
+        str: The model of the NVIDIA GPU or empty str if not found.
+    """
+    try:
+        model = subprocess.check_output(
```
njriasan: You should just be able to use the existing information that doesn't depend on subprocess. See the blackwell tests in test_max_autotune.py.

torch/testing/_internal/common_utils.py
```diff
@@ -1405,6 +1405,24 @@ def printErrors(self) -> None:
...
+            ["nvidia-smi", "--query-gpu=name", "--format=csv,noheader,nounits"]
+        )
+        return model.decode().strip().split("\n")[0]
+    except OSError:
+        log.warning("nvidia-smi not found. Returning empty str.")
+        return ""
+
+IS_B200 = torch.version.cuda is not None and get_nvidia_gpu_model() == "NVIDIA B200"
```
njriasan: This also wouldn't be correct to slight variations in the model name.

torch/testing/_internal/common_utils.py
jananisriram: makes sense, so we just want to check for `has_datacenter_blackwell_tma_device()`?

eellison (approved):

PR #164439, opened 2025-10-02
Move file selection for ruff from .lintrunner.toml to pyproject.toml

cyyever:
`test/dynamo/cpython/**` contains various valid and invalid python files for testing compilation. We should keep them untouched from linters.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -123,6 +123,27 @@ standard_library = ["typing_extensions"]
 line-length = 88
 src = ["caffe2", "torch", "torchgen", "functorch", "test"]
 
+include = [
+    "**/*.py",
+    "**/*.pyi",
+    "**/*.ipynb",
+]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -268,6 +268,10 @@ select = [
     "YTT",
 ]
 
+exclude = [
+  "test/dynamo/cpython/3_13/**"
```
Skylion007: All CPython right?

cyyever: yes, but currently `cpython` contains only `3_13`

cyyever: fixed

albanD (commented):
We don't run these in CI? Why is this not failing in CI?

albanD (commented):
Can we remove them from .lintrunner.toml?

PR #164463, opened 2025-10-02
Stop building nativert in OSS

malfet:
* __->__ #164463


albanD (approved):
Sounds ok to me, please confirm with @SherlockNoMad if there is anything else we should know about here

Skylion007 (approved):

Issue #73604, opened 2022-03-01
Support views in custom autograd functions


ezyang:
### 🐛 Describe the bug

Currently this fails:

```
        class Alias(Function):
            @staticmethod
            def forward(ctx, input):
                return input[:]
            
            @staticmethod
            def backward(ctx, grad):
                return grad
    
        base = torch.tensor([1.0], requires_grad=True)
        z = base * 1
        x = Alias.apply(z)
        z.mul_(-1)
        x.sum().backward()
        self.assertEqual(base.grad, torch.tensor([-1.0]))
```

with

```
    x.sum().backward()
RuntimeError: Output 0 of AliasBackward is a view and its base or another view of its base has been modified inplace. This view was created inside a custom Function (or because an input was returned as-is) and the autograd logic to handle view+inplace would override the custom backward associated with the custom Function, leading to incorrect gradients. This behavior is forbidden. You can fix this by cloning the output of the custom Function.
```

But we can actually support this today after XLA added the concept of view function. We just stash the custom autograd function in the view function and then replay it (lazily) when the base got mutated. Gradients would be preserved in this case.

Lack of support for this is causing this test to fail in NegativeTensor: https://github.com/albanD/subclass_zoo/blob/722bb0b503430a0781771cc75f6ac1cc0a1b48f1/negative_tensor.py#L130-L141 (and unfortunately, we don't even get a good error message for this case!)

### Versions

master

cc @albanD @zou3519 @gqchen @pearu @nikitaved @soulitzer @Lezcano @Varal7 @msaroufim @ezyang

albanD:
Unfortunately this is not really true :( 
The problem is that the user custom Function can have arbitrary side effect and so, contrary to our view ops, it is not safe to replay it.

ezyang:
OK... so let's add a decorator that says that a custom function is pure/referentially transparent, and then enable it in this case :)

albanD:
Sure we could do that.
Not sure what the final API would look like but I agree that if the user guarantees that this is ok to do, then we can do this.

myron:
any updates on this? thank you

albanD:
No update I'm afraid :/
But if you're interested in looking at this, I would be happy to help you here!
I think we can add a function like `ctx.set_is_pure_view_function(True)`. And when this is enabled, we can actually trigger the view logic as proposed by Ed.
For the internal change, that would be about making sure we track these as view_fn on the `ViewInfo` and not just metadata.

[...]

PR #164467, opened 2025-10-02
Add pure view support in autograd Function

albanD:
* #164359
* #164520
* #164576
* __->__ #164467

Fix https://github.com/pytorch/pytorch/issues/73604


torch/csrc/autograd/custom_function.cpp
```diff
@@ -448,21 +450,29 @@ optional_variable_list _wrap_outputs(
...
   std::unordered_map<at::TensorImpl*, size_t> inputs_mapping;
   inputs_mapping.reserve(input_vars.size());
   for (const auto i : c10::irange(input_vars.size())) {
     inputs_mapping.emplace(input_vars[i].unsafeGetTensorImpl(), i);
   }
 
+  // Limit pure views to 1-1 mapping as it is unclear if it is even
+  // possible to have a pure view for N-1 or 1-N.
+  TORCH_CHECK(
+      !pure_view || (input_vars.size() == 1 && raw_outputs.size() == 1),
```
soulitzer, line 464: I guess there's no limitation on non-Tensor input and outputs

torch/csrc/autograd/python_function.h
```diff
@@ -109,6 +109,10 @@ struct THPFunction {
   // Default is true.
   bool materialize_grads;
 
+  // boolean indication whether the function is a "pure view", meaning that
```
soulitzer:
```suggestion
  // boolean indicating whether the function is a "pure view", meaning that
```

torch/csrc/autograd/custom_function.cpp
albanD: No, you can capture these from the view_func replay

test/test_autograd.py
```diff
@@ -8279,71 +8282,99 @@ def backward(ctx, grad):
...
+                            return tmp.sum()
+
+                        a = torch.ones(2, dtype=dtype, requires_grad=True)
+                        b = torch.ones(2, dtype=dtype, requires_grad=True)
+
+                        err_msg = fn_id_to_inplace_on_view_err_msg[fn_id][
+                            int(pure_view)
+                        ]
+
+                        print(pure_view, fn_id, inplace, output_is_a_view)
```
soulitzer: debug print

ezyang (approved):
One thing I am puzzled by, is if I have something goes from DTensor -> Tensor (it's a view) and I mark it as a pure view, it seems kind of hard to believe that this will "just work".

test/test_autograd.py
```diff
@@ -8279,71 +8282,99 @@ def backward(ctx, grad):
...
+                        if (
+                            (pure_view and fn_id == "one_output")
+                            or (pure_view and fn_id == "view_of_temp" and not inplace)
+                            or (not pure_view and (not inplace or not output_is_a_view))
```
soulitzer: There are so many of these conditions now haha, its probably fundamental complexity though

soulitzer: Although if this condition and the condition for the error case are complements, could be good to either reuse or add an assertion?

soulitzer (approved):
Nice!

PR #164468, opened 2025-10-02
[nccl][pytorch] add functions to check nccl dynamic header

jushg:
Summary: For NCCL header only implementation, we need functions to check at runtime whether the NCCL we are compiling with is header only. NCCL with header only have NCCL DYNAMIC header

Differential Revision: D83476166


torch/csrc/distributed/c10d/init.cpp
```diff
@@ -3364,6 +3364,12 @@ options :class:`~torch.distributed.ProcessGroupNCCL.Options`).
               })
           .def_static("get_runtime_nccl_version", [] {
             return ::c10d::getNcclVersionTuple();
+          })
+          .def_static("is_nccl_dynamic_used", [] {
+            #ifdef NCCL_DYNAMIC
```
kwen2501, line 3369: What does `NCCL_DYNAMIC` mean?
Would there be more PRs after this one? If so, what is the overall goal?

jushg: @kwen2501 NCCL DYNAMIC is the header-only nccl we have for fbsource (i.e. build with attribute(weak) and does not include implementation, which will be load at runtime). 

We need to expose the C++ header definition of nccl to python to differ between this and static NCCL at runtime (hence this PR). The other changes are fbsource-internal linked via diff stack.

I can also try making  this as a separate module not touching OSS pytorch if that's more desirable

jushg: @kwen2501 yup got a separate module to check this, so I think i can abandon this PR

PR #164473, opened 2025-10-02
Fix mesh.get_local_rank when it is > 1d

fmassa:
Previously, we would not take the arguments passed by get_local_rank into account. This means that we wouldn't be able to trace this call if we had a device_mesh > 1d


xmfan (approved):
thank you! i should do a pass of that entire file

xmfan:
The CI error is asking you to update the graph break counts of some models, `vision_maskrcnn` now has 2 less graph breaks per iteration

Skylion007 (approved):

PR #164482, opened 2025-10-02
Fix type hints in PrepareModuleInput and PrepareModuleInputOutput

### LINKED ISSUE #161646: Incorrect type listed for PrepareModuleInput causing confusion and linter errors
### 📚 The doc issue

If you look at the docs for [PrepareModuleInput](https://docs.pytorch.org/docs/2.8/distributed.tensor.parallel.html#torch.distributed.tensor.parallel.PrepareModuleInput), as well as the type hints in the actual PyTorch codebase, you'll notice that the type for `input_layouts` and `desired_input_layouts` are specified as `Optional[Union[Placement, tuple[Optional[Placement]]]]`.

The issue is in the `tuple[Optional[Placement]]` section. Unlike lists, type hints for tuples regulate the number of elements passed in as well as the type of the elements. From the examples section, it is obvious that the intent of this argument is for it to be a tuple containing any number of `Optional[Placement]` values, but `tuple[Optional[Placement]]` means "a tuple containing exactly one `Optional[Placement]`. 

This is both confusing to read, because it directly contradicts the example code given, and irritating for people using type checkers, since it correctly notices that passing `(Shard(0), None)` is incorrect for `tuple[Optional[Placement]]`. 

### Suggest a potential alternative/fix

The correct type for these arguments would be `Optional[Union[Placement, tuple[Optional[Placement], ...]]]`, meaning a tuple with any number of `Optional[Placement]` elements.

RohitRathore1:
Fixes #161646 

### DIFF
--- a/torch/distributed/tensor/parallel/style.py
+++ b/torch/distributed/tensor/parallel/style.py
@@ -473,7 +473,9 @@ class PrepareModuleInput(ParallelStyle):
     def __init__(
         self,
         *,
-        input_layouts: Optional[Union[Placement, tuple[Optional[Placement], ...]]] = None,
+        input_layouts: Optional[
+            Union[Placement, tuple[Optional[Placement], ...]]
+        ] = None,
         desired_input_layouts: Optional[
             Union[Placement, tuple[Optional[Placement], ...]]
         ] = None,
@@ -634,8 +636,8 @@ class PrepareModuleOutput(ParallelStyle):
     def __init__(
         self,
         *,
-        output_layouts: Union[Placement, tuple[Placement]],
-        desired_output_layouts: Union[Placement, tuple[Placement]],
+        output_layouts: Union[Placement, tuple[Optional[Placement], ...]],
+        desired_output_layouts: Union[Placement, tuple[Placement, ...]],
         use_local_output: bool = True,
     ):
         self.output_layouts = (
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):

Skylion007 (changes requested):
Actually output layout needs to be fixed here too, right?

torch/distributed/tensor/parallel/style.py
```diff
@@ -764,9 +764,9 @@ class PrepareModuleInputOutput(ParallelStyle):
     def __init__(
         self,
         *,
-        input_layouts: Optional[Union[Placement, tuple[Optional[Placement]]]] = None,
+        input_layouts: Optional[Union[Placement, tuple[Optional[Placement], ...]]] = None,
         desired_input_layouts: Optional[
-            Union[Placement, tuple[Optional[Placement]]]
+            Union[Placement, tuple[Optional[Placement], ...]]
         ] = None,
         input_kwarg_layouts: Optional[dict[str, Placement]] = None,
```
Skylion007, line 775: Fix output_layout in this call signature too.

RohitRathore1: @Skylion007 I addressed your feedback, please check again. Thanks!

RohitRathore1:
> Actually output layout needs to be fixed here, right?

You are right, `output_layouts` can contain `None` values (`if out_layout is not None`) and the docstring also explicitly says: `If some outputs are not torch.Tensor or no need to convert to DTensors, None need to be specified as a placeholder.` Thanks!

torch/distributed/tensor/parallel/style.py
```diff
@@ -634,8 +636,8 @@ class PrepareModuleOutput(ParallelStyle):
     def __init__(
         self,
         *,
-        output_layouts: Union[Placement, tuple[Placement]],
-        desired_output_layouts: Union[Placement, tuple[Placement]],
+        output_layouts: Union[Placement, tuple[Optional[Placement], ...]],
```
Skylion007, line 639: Hmm, making this Optional is also desired?

Skylion007 (approved):

torch/distributed/tensor/parallel/style.py
RohitRathore1: Yes, making output_layouts Optional is desired because the docstring explicitly documents that None can be used as a placeholder

PR #164499, opened 2025-10-02
[multi-kernel] base tensor sizes for shape cache key

pianpwk:
to match shape key in https://github.com/pytorch/pytorch/blob/3ca09d65f1bdf83142dc9fe47976227ae4a88e7b/torch/_inductor/select_algorithm.py#L3571


ColinPeppler:
I was thinking what happens if we have a base tensor that is dynamic, but we take a view with static shapes, will this cause `_kernel_has_dynamic_shapes` to return True when it shouldn't have?

ColinPeppler (approved):

PR #164500, opened 2025-10-02
[CP] Replace context_parallel context manager with functional APIs

fegin:
* #165039
* #163185
* __->__ #164500

`context_parallel()` being a context manager has annoyed users. Now that we plan to redesign CP's UX to explicitly ask users to:

1. Wrap the attention op into an `nn.Module`
2. Lift any buffers that are not sequence agnostic to input

We can replace `context_parallel()` with two functional APIs: `_context_parallel_shard` and `_enable_context_parallel_dispatcher`.


test/distributed/tensor/test_attention.py
```diff
@@ -15,9 +15,11 @@
 from torch.distributed.tensor.debug import CommDebugMode
 from torch.distributed.tensor.experimental._attention import (
     _CausalBehavior,
+    _context_parallel_shard,
```
XilunWu, line 18: I suggest we make it public just like `context_parallel_unshard`.

test/distributed/tensor/test_attention.py
```diff
@@ -105,10 +107,7 @@ def test_ring_attention_sdpa(self) -> None:
                 "load_balance": [True, False],
                 "rotater": [_RotateMethod.ALL_TO_ALL, _RotateMethod.ALL_GATHER],
                 "test_forward_only": [True, False],
-                "dispatch_mode": [
-                    _DispatchMode.MONKEY_PATCH,
-                    _DispatchMode.MODULE_WRAPPER,
-                ],
+                "new_api": [True, False],
```
XilunWu: suggestion: rename it to "use_context"

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -998,14 +998,22 @@ def _context_parallel_dispatcher(
...
+def _disable_context_parallel_dispatcher_impl() -> None:
+    if _dispatch_mode == _DispatchMode.MONKEY_PATCH:
         _restore_function(F.scaled_dot_product_attention, F)
     elif _dispatch_mode == _DispatchMode.MODULE_WRAPPER:
-        with _enable_cp_dtensor_dispatcher():
-            yield
+        pass
     else:
-        raise NotImplementedError("torch dispatch mode is not supported yet.")
+        raise NotImplementedError
```
XilunWu: `raise ValueError(f"Unknown dispatch mode: {_dispatch_mode}")` like in `_enable_context_parallel_dispatcher_impl()`.

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1257,6 +1285,61 @@ def sdpa_output_fn(
         return tuple(new_outputs)
 
 
+def _context_parallel_shard(
+    mesh: DeviceMesh,
+    buffers: list[torch.Tensor | BlockMask],
+    seq_dims: list[int],
+    load_balance_indices: Optional[torch.Tensor] = None,
+) -> list[torch.Tensor | BlockMask]:
```
XilunWu, line 1360: It's interesting to see the concept of sharding a `BlockMask` object since it's mainly a collection of sparsity tensor and sharding them is equivalent to sharding a `BlockMask` itself. But this may not be straightforward to users. Some documentation should help, can be a TODO for now.

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1051,22 +1059,42 @@ def _generate_round_robin_indices(
...
+            sharded_buffer = _create_cp_block_mask(
+                mask_mod=buffer.mask_mod,
+                B=buffer.kv_num_blocks.shape[0],
+                H=buffer.kv_num_blocks.shape[1],
+                Q_LEN=buffer.seq_lengths[0],
+                KV_LEN=buffer.seq_lengths[1],
+                device_mesh=mesh,
+            )
+        else:
+            raise ValueError(f"Unknown buffer type: {type(buffer)}")
```
XilunWu, line 1103: TODO: this function is too similar to `_context_parallel_shard` in terms of function signature. We can refactor them later.

XilunWu (approved):
Some nits, stamp to unblock us and we can come back later to address them.

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1051,22 +1059,42 @@ def _generate_round_robin_indices(
...
+        if isinstance(buffer, torch.Tensor):
+            if load_balance_indices is not None:
+                buffer = torch.index_select(
+                    buffer, dim=seq_dim, index=load_balance_indices
+                )
+            # use DTensor to shard the buffer on sequence dimension, retain the local tensor
+            sharded_buffer = distribute_tensor(
+                buffer, mesh, [Shard(seq_dim)], src_data_rank=None
+            ).to_local()
+        elif isinstance(buffer, BlockMask):
```
tianyu-l, line 1092: If this doesn't support `load_balance_indices` for now, shall we raise `NotImplementedError`

test/distributed/tensor/test_attention.py
```diff
@@ -127,66 +126,73 @@ def _ring_attention_sdpa(
...
             cp_plan = _ContextParallel(
                 seq_dim=seq_dim,
                 attention_type=_ContextParallel.AttentionType.SDPA,
             )
             attention = SDPAWrapper(compiled=compiled, backend=backend)
             attention = parallelize_module(attention, mesh, cp_plan)
+            cp_q, cp_k, cp_v = _context_parallel_shard(
+                mesh, (cp_q, cp_k, cp_v), (seq_dim,) * 3
+            )
+            _enable_context_parallel_dispatcher(seq_dim, mesh)
```
tianyu-l: It looks for now this is only for sdpa but not FlexAttention.

Can we put them in `sdpa_cp.sdpa_input_fn`, `sdpa_cp.sdpa_output_fn`? It's also safer that way.

tianyu-l (dismissed):
I somehow feel we don't need `_enable_context_parallel_dispatcher` as user-facing API.

torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1257,6 +1285,61 @@ def sdpa_output_fn(
         return tuple(new_outputs)
 
 
+def _context_parallel_shard(
+    mesh: DeviceMesh,
+    buffers: list[torch.Tensor | BlockMask],
+    seq_dims: list[int],
+    load_balance_indices: Optional[torch.Tensor] = None,
```
XilunWu: note: as we discussed, this argument will be `load_balancer: Optional[_LoadBalancer] = None`. Then in the function body we add `load_balance_indices = load_balancer.generate_indices() if load_balancer is not None else None`

test/distributed/tensor/test_attention.py
fegin: I think the problem is that we cannot put _disable_context_parallel_dispatcher in `sdpa_output_fn` because we want to wait until the backward so that we can disable it. iirc, if we do something in the backward hook, it may cause graph break? I'm not sure.

torch/distributed/tensor/experimental/_attention.py
fegin: `_context_parallel_shard` will be the public API while `_context_parallel_buffers` is the internal implementation that is used by both legacy `context_parallel` and `_context_parallel_shard`.

test/distributed/tensor/test_attention.py
fegin: Yes, I'll publish some day, but now that we are still iterating the implementation, I want to make it private.

torch/distributed/tensor/experimental/_attention.py
fegin: ye, I think unless users know what BlockMask is composed with, it may not be easy for users to understand what tensors to pass. We also need to rewrite some attribute like seq_len.

test/distributed/tensor/test_attention.py
fegin: Discussed offline, keep it for now. We need to think about a better way to integrate DTensor with SDPA

PR #162326, opened 2025-09-06
RFC varlen_api op

drisspg:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #162326

```Py
import torch
import torch.nn as nn
import torch.nn.functional as F
from torch.nn.attention import varlen_attn
from rich import print

# Set seed for reproducibility
torch.manual_seed(42)
if torch.cuda.is_available():
    torch.cuda.manual_seed(42)

class AttentionBlock(nn.Module):
    def __init__(self, embed_dim: int, num_heads: int, device: torch.device, dtype: torch.dtype):
        super().__init__()
        self.embed_dim = embed_dim
        self.num_heads = num_heads
        self.head_dim = embed_dim // num_heads
        
        self.qkv_proj = nn.Linear(embed_dim, 3 * embed_dim, bias=False, device=device, dtype=dtype)
        self.out_proj = nn.Linear(embed_dim, embed_dim, bias=False, device=device, dtype=dtype)
    
    def forward_varlen(self, x_packed: torch.Tensor, cu_seq: torch.Tensor, max_len: int, is_causal: bool = False):
        qkv = self.qkv_proj(x_packed)
        q, k, v = qkv.chunk(3, dim=-1)
        
        q = q.view(-1, self.num_heads, self.head_dim)
        k = k.view(-1, self.num_heads, self.head_dim)
        v = v.view(-1, self.num_heads, self.head_dim)
        
        attn_out = varlen_attn(q, k, v, cu_seq, cu_seq, max_len, max_len, is_causal=is_causal)
        attn_out = attn_out.view(-1, self.embed_dim)
        
        return self.out_proj(attn_out)
    
    def forward_sdpa(self, x_padded: torch.Tensor, is_causal: bool = False):
        batch_size, seq_len, _ = x_padded.shape
        
        qkv = self.qkv_proj(x_padded)
        q, k, v = qkv.chunk(3, dim=-1)
        
        q = q.view(batch_size, seq_len, self.num_heads, self.head_dim).transpose(1, 2)
        k = k.view(batch_size, seq_len, self.num_heads, self.head_dim).transpose(1, 2)
        v = v.view(batch_size, seq_len, self.num_heads, self.head_dim).transpose(1, 2)
        
        attn_out = F.scaled_dot_product_attention(q, k, v, is_causal=is_causal)
        attn_out = attn_out.transpose(1, 2).contiguous().view(batch_size, seq_len, self.embed_dim)
        
[...]

PR #164502, opened 2025-10-02
varlen api

liangel-02:
**Summary** 

Today, the only way to have variable sequence length support in PyTorch attention is through nested tensors [here](https://docs.pytorch.org/tutorials/intermediate/scaled_dot_product_attention_tutorial.html#nestedtensor-and-dense-tensor-support). We also want to add an explicit lower-level API that provides variable sequence length support without padding/masking in SDPA. 

This PR builds out `varlen_attn`, the public API that users can call for the forward method, and `_varlen_attn`, the private API that calls into the Flash Attention/cuDNN backend.

**Benchmarking**

To benchmark, we compare runtime and TFLOPs against the current SDPA approach with padding.

Settings: 

- 1 H100 machine
- `batch_size=8`, `max_seq_len=2048`, `embed_dim=1024`, `num_heads=16`
- dtype `torch.bfloat16`
- `is_causal=False`
- for variable length, we set sequences to be random multiples of 64 up to `max_seq_len`
- 100 runs 

|        | Variable Length API | SDPA     |
|--------|--------------------|----------|
| Runtime | 0.21750560760498047 ms       | 0.43171775817871094 ms  |
| TFLOPs | 231.812         | 320.840  |

The sparsity is 0.453 which we can see matches the speedup we get from Varlen (approx 50%). TFLOPs remains around the same, with SDPA slightly larger due to potential higher overhead and total flops scaling with sequence length.

**Testing**

Run `python test/test_varlen_attention.py` for unit tests where we verify basic functionality and confirm numerical match between varlen outputs vs SDPA.

**Next steps** 

Next steps from this PR (higher in the stack) include registering the private API `_varlen_attn` as a custom op, implementing backward support, and enabling cuDNN with correct numerics. 

* #164950
* #164504
* #164503
* __->__ #164502

(This stack builds on top of #162326)


torch/nn/attention/varlen.py
```diff
@@ -0,0 +1,147 @@
...
+
+    This function is similar to scaled_dot_product_attention but optimized for
+    variable-length sequences using cumulative sequence position tensors.
+
+    Args:
+        query (Tensor): Query tensor
+        key (Tensor): Key tensor
+        value (Tensor): Value tensor
+        cu_seq_q (Tensor): Cumulative sequence positions for queries
+        cu_seq_k (Tensor): Cumulative sequence positions for keys/values
```
jbschlosser: nit: might be nice to have an example to explain how these are expected to be formatted. For example, what shape is expected? should both 0 and the max sequence length be included? etc.

drisspg: Yeah I think adding the shape legend similar to what we did for SDPA would be great

torch/nn/attention/varlen.py
```diff
@@ -0,0 +1,147 @@
...
+    - query shape: (total, num_heads, head_dim)
+    - logsumexp shape: (num_heads, total_q)
+    """
+    # Output has same shape as query
+    output = torch.empty_like(query)
+
+    # For varlen path: logsumexp shape is (num_heads, total_q)
+    total_q = query.size(0)
+    num_heads = query.size(1)
+    logsumexp = torch.empty(
```
drisspg, line 116: double check the meta impl in meta_registrations.py to ensure this correct

torch/nn/attention/varlen.py
```diff
@@ -0,0 +1,147 @@
...
+def varlen_attn(
+    query: torch.Tensor,
+    key: torch.Tensor,
+    value: torch.Tensor,
+    cu_seq_q: torch.Tensor,
+    cu_seq_k: torch.Tensor,
+    max_q: int,
+    max_k: int,
+    is_causal: bool = False,
+    return_lse: bool = False,
```
drisspg: Also I recently changed to how we do this in flex-attention lets use the same way to do this

torch/nn/attention/varlen.py
```diff
@@ -0,0 +1,185 @@
...
+
+# import failures when I try to register as custom op
+# @torch.library.custom_op("torch_nn_attention::_varlen_attn", mutates_args={})
+def _varlen_attn(
+    query: torch.Tensor,
+    key: torch.Tensor,
+    value: torch.Tensor,
+    cu_seq_q: torch.Tensor,
+    cu_seq_k: torch.Tensor,
+    max_q: int,
```
drisspg, line 44: This is annoying because we no longer need this host metadata for in FAv3.


Do we need this for CuDNN @eqy ?

eqy: Is this the max batch size for batch q and k?
In theory it's not needed if we just do something like round up to neariest power-of-2 etc

drisspg: So its the max( seqlen_q_i) for sequences in the batch and same for k

It was used for determining how much blocks need to be iterated , but in fav3 we just get this from the cu_seq_q and cu_seq_k

eqy: Bad typo on my part, yeah-sequence length--I think we can either just infer from other metadata or round on top of that. Don't see this as mandatory, it is only used as integration as part of the cache key do decide if we can go to an existing execution plan or if we exceeded the previous max and need a new one.

v0i0 (approved):

test/test_varlen_attention.py
```diff
@@ -0,0 +1,195 @@
+# Owner(s): ["module: nn"]
```
mikaylagawarecki:
```suggestion
# Owner(s): ["module: sdpa"]
```

torch/nn/attention/varlen.py
drisspg: Screenshot 2025-10-10 at 11 17 30 PM

My current TLDR is that:
* FAv2 needed because we do normal grid launch and for the ceildiv(seqlen, blockm) we use max_q as seqlen. Potentially overlaunch CTAs but they early exist
* FAv3 is much more involved, but appears to `always` choose the persistent launch for varlen. And there are a few branches that use max-seqlenq via `params.seqlen_q`, but I am not convinced that it any of these need to rely on this?
* Fav4/Cute doesn't have the maxq, there is a persistent schedule there is also a varlen schedule that isn't persistent but based off of the totalq

So TBH I am no sure about this arg


cc @tridao if I butchered any of this, I need to go to bed 😂 Will relook on monday

tridao: Right, max_seqlen_q is needed if one uses a simple scheduler. It's not needed if one uses a persistent scheduler, or a varlen scheduler (i.e. the scheduler does extra work to find out how many thread blocks it should launch). Typically persistent scheduler is faster, so FA3 goes with that.

torch/nn/attention/varlen.py
```diff
@@ -0,0 +1,193 @@
+"""
+Variable-length attention implementation using Flash Attention.
```
drisspg, line 2: nit lets remove the `using flash attention`

torch/nn/attention/varlen.py
```diff
@@ -0,0 +1,193 @@
...
+__all__ = ["varlen_attn", "AuxRequest"]
+
+
+@lru_cache(maxsize=8)
+def _should_use_cudnn(device_index: int) -> bool:
+    """Cache device capability check to avoid repeated CUDA calls."""
+    return False
+
+
+class AuxRequest(NamedTuple):
```
drisspg, line 26: add a lil doc block,

drisspg (approved):

PR #164503, opened 2025-10-02
register custom op

liangel-02:
**Summary**

Adding custom op registration for `_varlen_attn()`. 

**Test plan**

Added `test_custom_op_registration` that uses logging mode to verify that `_varlen_attn()` was called and tests with `torch.compile`. Verified that the test fails when registration line is commented out

* #164950
* #164504
* __->__ #164503
* #164502


jbschlosser (commented):
might have missed this but is the op tested in torch.compile?

test/test_varlen_attention.py
```diff
@@ -81,59 +120,116 @@ def create_variable_length_batch(shape: VarlenShape, device: torch.device, dtype
...
+        total_tokens = shape.batch_size * shape.max_seq_len
+        x_packed = torch.randn(
+            total_tokens, shape.embed_dim, device=device, dtype=dtype
+        )
+        cu_seq = torch.tensor(
+            [0, shape.max_seq_len, total_tokens], device=device, dtype=torch.int32
+        )
+
+        compiled_forward = torch.compile(
+            attention_block.forward_varlen, backend="eager"
```
jbschlosser: I'd specify `fullgraph=True` here to ensure we're not supporting this by graph breaking

PR #164510, opened 2025-10-02
[DeviceMesh] Move global state into class method

fduwjj:
* __->__ #164510

This is PR trying to move bookkeeping state maps from MeshEnv to DeviceMesh class members. The reason is that in general global variables are thread local and cause potential issue.

We will also need to do DTensor CPU overhead benchmark for this change.

3-5% CPU overhead in DTensor has been observed:

before: 
image

After:
image
 
running the benchmark mentioned here: https://github.com/pytorch/pytorch/issues/159169


torch/distributed/device_mesh.py
```diff
@@ -72,146 +72,21 @@ def _init_device_mesh_stub():
     class _MeshEnv(threading.local):
         def __init__(self) -> None:
             self.mesh_stack: list[DeviceMesh] = []
-            # TODO: Move the bookkeeping maps from _MeshEnv to DeviceMesh.
-            self.child_to_root_mapping: dict[DeviceMesh, DeviceMesh] = {}
+            # TODO: Move the mesh_dim_group_options maps from _MeshEnv to DeviceMesh.
             self.mesh_dim_group_options: dict[int, BackendConfig] = {}
```
lw: `mesh_dim_group_options` has been made redundant in https://github.com/pytorch/pytorch/pull/159371, but I forgot to add an explicit deprecation warning. We could either add a warning now, wait a couple of releases, and then remove it, or we could remove it right away counting on the fact that it's a private attribute/function and that no piece of code on GitHub appears to be using it: https://github.com/search?type=code&q=._set_mesh_dim_group_options%28&p=1.

torch/distributed/device_mesh.py
```diff
@@ -666,6 +420,19 @@ def _init_process_groups(
...
+            # If a mesh does not have a root mesh stored, it is a root mesh itself.
+            # A root mesh is not created through slicing.
+            return self._root_mesh if self._root_mesh else self
+
+        @root_mesh.setter
+        def root_mesh(self, mesh: Optional["DeviceMesh"]) -> None:
+            # you can add validation logic here if needed
+            if mesh is not None and not isinstance(mesh, DeviceMesh):
+                raise TypeError(f"Expected DeviceMesh or None, got {type(mesh)}")
+            self._root_mesh = mesh
```
lw: Could we make these attributes private? I'm still not convinced that the distinction between root and child mesh is really useful. At the end of the day, I believe, all we care about is which meshes belong to the same "universe" (i.e., which ones have the same shared state?).

While we discuss this and decide, I'd prefer if we avoid adding public API attributes which will limit us in the future.

torch/distributed/device_mesh.py
```diff
@@ -72,146 +72,21 @@ def _init_device_mesh_stub():
...
-            root_mesh = self.get_root_mesh(device_mesh)
-            child_mesh_dim_names = device_mesh.mesh_dim_names
-            if root_mesh and child_mesh_dim_names:
-                assert len(child_mesh_dim_names) == 1, (
-                    "The submesh can only be a 1D mesh."
-                )
-                child_mesh_dim_name = child_mesh_dim_names[0]
-                return self.get_mesh_dim_by_name(root_mesh, child_mesh_dim_name)
-            return None
+            # We keep this function for backward compatibility.
```
lw, line 86: Could we add a deprecation warning to it? This should allow us to remove it in 1-2 releases.

Unless we believe that there's only internal usages, in which case we can remove it much sooner.

torch/distributed/device_mesh.py
```diff
@@ -699,6 +466,7 @@ def __hash__(self):
                         self.device_type,
                         self.mesh_dim_names,
                         self._thread_id,
+                        self._root_mesh,
```
lw, line 464: Here and in the equality operator below, could we just use the `id(...)` of the root mesh?

torch/distributed/device_mesh.py
```diff
@@ -844,6 +607,245 @@ def get_all_groups(self) -> list[ProcessGroup]:
             """
             return [self.get_group(i) for i in range(self.mesh.ndim)]
 
+        def create_sub_mesh(
```
lw: Same here, could we keep these methods private for now? This is a refactoring so we don't have to change the public API, especially if it means future BC burden. We can make these public later if there's a compelling use case.

torch/distributed/device_mesh.py
```diff
@@ -844,6 +607,245 @@ def get_all_groups(self) -> list[ProcessGroup]:
...
+            res_submesh = DeviceMesh._create_mesh_from_ranks(
+                self.device_type,
+                pg_ranks_by_dim,
+                cur_rank,
+                submesh_dim_names,
+                _init_backend=False,
+                _layout=layout,
+            )
+            res_submesh._dim_group_names = slice_dim_group_name
+            res_submesh.root_mesh = root_mesh
```
lw: Any reason why we set it after-the-fact, rather than add a new argument to `_create_mesh_from_ranks` and pass it to it?

torch/distributed/device_mesh.py
```diff
@@ -844,6 +607,245 @@ def get_all_groups(self) -> list[ProcessGroup]:
...
+                root_mesh.device_type,
+                pg_ranks_by_dim.flatten(
+                    start_dim=1
+                ),  # this is needed for flatten non-contiguous mesh dims.
+                cur_rank,
+                (mesh_dim_name,),
+                (backend_override,),
+                _layout=self._layout.coalesce(),
+            )
+            res_flattened_mesh.root_mesh = root_mesh
```
lw: Same here

torch/distributed/device_mesh.py
```diff
@@ -844,6 +607,245 @@ def get_all_groups(self) -> list[ProcessGroup]:
...
+                    f"The mesh_dim_names of submesh and flattened mesh are {invalid_dim_names}. "
+                    f"Please specify another valid mesh_dim_name.",
+                )
+
+            flattened_mesh_layout = self._layout.coalesce()
+            # Quick return if the flatten mesh has been created before.
+            if (
+                root_mesh._flatten_mapping
+                and mesh_dim_name in root_mesh._flatten_mapping
+            ):
```
lw: This check can now be simplified:

```suggestion
            if mesh_dim_name in root_mesh._flatten_mapping:
```

lw (approved):
This is great! I like this a lot! Thanks!

torch/distributed/device_mesh.py
fegin: This should be only used by Meta internal. 
Except for the Meta own repos (e.g., TorchTitan, Shampoo), other repos are basically fork from the PyTorch. We never claim this is supported publicly except for the internal users. Maybe this is okay to remove sooner, after another release?

But it worths to check.

torch/distributed/device_mesh.py
fegin: Agree this. I also have concerns on other public methods. Please see the flatten one.

torch/distributed/device_mesh.py
```diff
@@ -844,6 +607,245 @@ def get_all_groups(self) -> list[ProcessGroup]:
...
+            )
+            res_submesh._dim_group_names = slice_dim_group_name
+            res_submesh.root_mesh = root_mesh
+            return res_submesh
+
+        def create_flatten_mesh(
+            self,
+            mesh_dim_name: Optional[str] = None,
+            backend_override: BackendConfig = (None, None),
+        ) -> "DeviceMesh":
```
fegin, line 649: Why is this a public method? We also have a `_flatten`. This is confusing. Also, if we make this a public method, what is the semantic? Is the flattened mesh able to be access from the root mesh? If a public method is provided and users start to try it, it will be difficult to change the semantic.

fegin (changes requested):
These methods should be private.

torch/distributed/device_mesh.py
fduwjj: yes making it private and will have another PR for the idea for universe.

torch/distributed/device_mesh.py
fduwjj: Removed in https://github.com/pytorch/pytorch/pull/164750.

torch/distributed/device_mesh.py
fduwjj: sure, let me add a warning first.

torch/distributed/device_mesh.py
fduwjj: id(...) is the fast path, if the CPU overhead is not too big, I think we can still keep the old way. Once we have the universe thing, we can do id based comparison.

torch/distributed/device_mesh.py
```diff
@@ -72,145 +72,23 @@ def _init_device_mesh_stub():
     class _MeshEnv(threading.local):
         def __init__(self) -> None:
             self.mesh_stack: list[DeviceMesh] = []
```
lw, line 74: This is great! This is the only attribute that _needs_ to be in this object (since it's modified via a context manager), it's so good that we managed to remove all other ones!

torch/distributed/device_mesh.py
```diff
@@ -72,145 +72,23 @@ def _init_device_mesh_stub():
...
-            if root_mesh and child_mesh_dim_names:
-                assert len(child_mesh_dim_names) == 1, (
-                    "The submesh can only be a 1D mesh."
-                )
-                child_mesh_dim_name = child_mesh_dim_names[0]
-                return self.get_mesh_dim_by_name(root_mesh, child_mesh_dim_name)
-            return None
+            # We keep this function for backward compatibility.
+            warnings.warn(
+                "This _get_all_submeshes API will be deprecated soon."
```
lw: The warning message contains the incorrect method name

lw (approved):
Great! Thanks for addressing the comments!

fegin (approved):

torch/distributed/device_mesh.py
fduwjj: opps

torch/distributed/device_mesh.py
fduwjj: 3-5% of CPU overhead regression was observed which should be fine, because this enables us for cleaner device mesh implementation.

PR #156703, opened 2025-06-24
Stop parsing command line arguments every time common_utils is imported.

AnthonyBarbier:
Last PR in the series to re-submit https://github.com/pytorch/pytorch/pull/134592 as smaller PRs:

https://github.com/pytorch/pytorch/pull/154612
https://github.com/pytorch/pytorch/pull/154628
https://github.com/pytorch/pytorch/pull/154715
https://github.com/pytorch/pytorch/pull/154716
https://github.com/pytorch/pytorch/pull/154725
https://github.com/pytorch/pytorch/pull/154728

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci

PR #164516, opened 2025-10-02
[testing] Fix internal testing not being able to import some tests

clee2000:
I think the internal runner imports modules and files directly, so it doesn't run the `if __name__ == "__main__"` and the command line parsing there that would set the default values for some global variables

See https://github.com/pytorch/pytorch/pull/156703#issuecomment-3362135688 for context

Testing:
ran locally on devserver to see if tests could be collected

### DIFF
--- a/torch/testing/_internal/common_utils.py
+++ b/torch/testing/_internal/common_utils.py
@@ -877,7 +877,7 @@ def cppProfilingFlagsToProfilingMode():
     else:
         return ProfilingMode.LEGACY
 
-# Set default value to for the internal test runner which imports files/modules
+# Set default value for the internal test runner which imports files/modules
 # directly
 GRAPH_EXECUTOR : ProfilingMode = cppProfilingFlagsToProfilingMode()
 
--- a/torch/testing/_internal/common_utils.py
+++ b/torch/testing/_internal/common_utils.py
@@ -115,7 +115,6 @@ CI_FUNCTORCH_ROOT = ""
 CI_PT_ROOT = ""
 CI_TEST_PREFIX = ""
 DISABLED_TESTS_FILE = ""
-GRAPH_EXECUTOR : Optional[ProfilingMode] = None
 LOG_SUFFIX = ""
 PYTEST_SINGLE_TEST = ""
 REPEAT_COUNT = 0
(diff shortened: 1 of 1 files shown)


malfet (approved):

torch/testing/_internal/common_utils.py
```diff
@@ -878,6 +877,10 @@ def cppProfilingFlagsToProfilingMode():
     else:
         return ProfilingMode.LEGACY
 
+# Set default value for the internal test runner which imports files/modules
+# directly
+GRAPH_EXECUTOR : ProfilingMode = cppProfilingFlagsToProfilingMode()
```
seemethere, line 882: Are there any negatives to this?

huydhn (approved):

torch/testing/_internal/common_utils.py
AnthonyBarbier: Yes, this is not a fix at all: this is just restarting the mess fixed by https://github.com/pytorch/pytorch/pull/156703

The problem was that a simple `import common_utils` used to have a lot of side effects because it would implicitly parse sys.argv and set a whole bunch of global states, but for example if the import is done in a subprocess (e.g distributed tests) where the `sys.argv` are not set, then the global states are different (And also the real command line arguments are ignored which might surprise the user).

So the other PR made the command line parsing explicit and added an assert to make sure the global state is initialised before it's being used.
Setting a default value to GRAPH_EXECUTOR (Which is used at tests collection to parameterise some tests) means some tests might read the default value, then if the command line gets parsed, some others will read a different value.
The proper way of fixing this is to make sure the command line arguments are parsed before the tests are collected which is done [here](https://github.com/pytorch/pytorch/blob/ddf8de28c25944a58e739ba9996b06753e4199cc/test/conftest.py#L96) for pytest) and [here](https://github.com/pytorch/pytorch/blob/ddf8de28c25944a58e739ba9996b06753e4199cc/test/test_jit.py#L6-L11) for unittest

What does your CI do? I thought it was using pytest?

AnthonyBarbier (changes requested):
If your CI imports the tests as modules just get it to call [parse_cmd_line_args()](https://github.com/pytorch/pytorch/blob/ddf8de28c25944a58e739ba9996b06753e4199cc/test/conftest.py#L96) once before the tests collection

PR #164517, opened 2025-10-02
multimem reduce

kwen2501:
* __->__ #164517

Modified `multimem_one_shot_all_reduce_out` function to accept a `root` argument, making it a `multimem_reduce` op.

The original `multimem_one_shot_all_reduce` op becomes a caller of the `multimem_reduce`, with each rank providing its own rank id as root.

### DIFF
--- a/test/distributed/test_symmetric_memory.py
+++ b/test/distributed/test_symmetric_memory.py
@@ -918,6 +918,37 @@ class SymmMemCollectiveTest(MultiProcContinuousTest):
             gathered_inps.sum(dim=0), res, rtol=1e-03, atol=1e-05
         )
 
+    @skip_if_lt_x_gpu(4)
+    @requires_multicast_support()
+    @parametrize("dtype", [torch.float, torch.bfloat16])
+    @parametrize("size_bytes", [4, 8192, 8196])
+    # https://github.com/pytorch/pytorch/issues/164015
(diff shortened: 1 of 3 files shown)


ngimel (approved):

shunting314:
Do we drop the support for cuda version before 12.3?

Since otherwise, something like:

```
+at::Tensor multimem_one_shot_reduce_out(
+    const at::Tensor& input,
+    std::string reduce_op,
+    int64_t root,
+    std::string group_name,
+    at::Tensor out) {
+  TORCH_CHECK(false, "multimem_one_shot_reduce_out: requires CUDA 12.3+.");
+  return input;
+}
```
need to be added after ```#elif defined(CUDART_VERSION) && CUDART_VERSION < 12030```
to be able to build

kwen2501:
@shunting314 The code you cited is merely host-side code, without fancy CUDA APIs. Which part is not buildable? Can you share your error log?

shunting314:
I don't have the log available. But it says multimem_one_shot_reduce_out is not defined since the definition is in a conditional block guarded by the CUDA 12.3 check.

kwen2501:
I see, thank you @shunting314

kwen2501:
Added the temporary workaround as suggested. 
Longer term, we should remove all those false impl's.

PR #164518, opened 2025-10-02
shrink_group implementation to expose ncclCommShrink API

### LINKED ISSUE #164529: [RFC] Implement shrink_group API to expose ncclCommShrink
### PyTorch Process Group Shrink API 

Authors: @brchang24 @spotluri @gbosilca 

**Summary** 

This document outlines proposed API changes to improve fault tolerance and flexibility in PyTorch Process Groups.  

**Motivation** 

Fault Tolerance support 

The API is designed to enhance fault tolerance capabilities in PyTorch by enabling exclusion of faulty ranks. 

With the new API, malfunctioning participants should be excluded to avoid hangings, as their reliability can't be guaranteed. This API is not limited to hard faults (where processes disappear due to hardware failures) but allows the application to mold the execution environment as needed (for correctness or performance reasons). 

Performance 

The existing abort \+ init method entails a fixed cost for full initialization, as illustrated in the chart below with rank shrink as an example.

Image

  
We also explored alternatives like split, but that approach requires the participation of all ranks, including malfunctioning ones, to form a new group.  

So, both above factors must be considered when designing the API. 

Proposed API changes 

To address these concerns above, we are proposing the following API changes: 

New API: shrink\_group() 

```python
shrink_group(ranks_to_exclude: List[int],
            Pg: Optional[ProcessGroup] = None,
            shrink_flags : int = NCCL_SHRINK_DEFAULT)

Shrink an existing distributed group. Only group members of the updated ProcessGroup need to enter this function. The excluded ranks do not need to call this function. The scope of this call is therefore collective across the processes that belong to the shrunk distributed group.

Args:
ranks_exclude (list[int]):  List of group ranks to be excluded in the updated ProcessGroup. This list must be consistent across all participating processes.
pg (ProcessGroup, optional): The process group to work on, If None, the default process group will be used.
shrink_flags (int): NCCL_SHRINK_DEFAULT (default)
                    NCCL_SHRINK_ABORT (attempt to terminate ongoing operations in the parent communicator before shrinking. 
```

Implementation 

PyTorch should directly use the shrink functionality if supported by the backend. 

**Use Cases** 

Rank Shrink 

When one rank is detected defective and needs to be excluded, the new shrink\_group() can be invoked as the below example: 

Example 

remove rank 2   
**shrink\_group**(\[2\], pg)   

Image

In the example above, only the original rank 0, 1, and 3 need to invoke shrink\_group(), not rank 2 to avoid potential hang. After the shrink, the original rank 2 is excluded, so, the original rank 3 will be shifted down to become rank 2 in the updated group, if no parameter key is passed in. 

Errors could be reported by the backend or PyTorch. However, it is up to the upper layer to decide whether to exclude a rank from the group. 

Ideally, the defective ranks should be excluded from all associated groups. However, it appears to be enforcing a policy on PyTorch. This design suggests delegating the decision to exclude ranks from a group or completely dissolve the group to the Upper Layer (UL). Meanwhile, PyTorch should verify that no subgroups use the rank before it is removed from the default group. 

**Things to Consider** 

**Group Rank recalculation** 

Shrink can lead to changes in the group's rank order. It will apply the default method to recalculate group ranks as detailed below.  

When shrinking, ranks will be shifted to close any gaps left by excluded ranks. 

\*\*Metrics \*\* 

What are the main metrics to measure the value of this feature? 

1. When one node/rank goes down completely   
   To compare with the existing solution. 

 

2. Performance comparison with the existing solutions   
1. Shrink 

Drawbacks 

Are there any reasons why we should not do this? Here we aim to evaluate risk and check ourselves.   
Please consider: 

* Is it a breaking change?   
  This change should be backward compatible.   
* Impact on UX   
  No   
* implementation cost, both in terms of code size and complexity   
  The assumption is that the pytorch cost should not be high, but the backend support cost might be high, especially for backends that do not have already support for fault management.   
* integration of this feature with other existing and planned features   
  PyTorch layer needs to integrate with the backend when it has the support. 

Alternatives 

What other designs have been considered? What is the impact of not doing this?   
   
Can be implemented using abort+init method.   
Need full init method which potentially requires broadcast for the NCCL bootstrap. 

Prior Art 

Discuss prior art (both good and bad) in relation to this proposal: 

* Does this feature exist in other libraries? What experience has their community had?   
  MPI has similar concepts with steady adoption in highly scalable HPC applications and runtimes.   
* What lessons can be learned from other implementations of this feature?   
  Such as NCCL shrink API   
* Published papers or great posts that discuss this   
  n/a 

How we teach this 

* There are some new concepts introduced after this RFC   
* exclude ranks   
     
* the PyTorch documentation needs to include the API changes   
  shrink\_group() 
  This is a new API and should be added.   
 
* How should this feature be taught to existing PyTorch users   
Including a section with examples would be ideal to demonstrate how shrink use cases function with the proposed new API changes.  

Unresolved questions 

* What parts of the design do you expect to resolve through the RFC process before this gets merged? 
No 

* What parts of the design do you expect to resolve through the implementation of this feature before stabilization?   
No   
 
* What related issues do you consider out of scope for this RFC that could be addressed in the future independently of the solution that comes out of this RFC?   
No 

Resolution 

Level of Support 

Choose one of the following: 

* 1: Overwhelming positive feedback.   
* 2: Positive feedback.   
* 3: Majority Acceptance, with conflicting Feedback.   
* 4: Acceptance, with Little Feedback.   
* 5: Unclear Resolution.   
* 6: RFC Rejected.   
* 7: RFC Rejected, with Conflicting Feedback. 

Additional Context 

Next Steps 

Tracking issue 

Exceptions 
 
no

### Alternatives

_No response_

### Additional context

_No response_

brchang24:
Closes #164529 

To expose the new [ncclCommShrink](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/comms.html#ncclcommshrink) API to PyTorch.

This is useful when you need to exclude certain GPUs or nodes from a collective operation, for example in fault tolerance scenarios or when dynamically adjusting resource utilization.

For more info:  [Shrinking a communicator](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/usage/communicators.html#shrinking-a-communicator)

### DIFF
--- a/test/distributed/test_c10d_nccl.py
+++ b/test/distributed/test_c10d_nccl.py
@@ -2,6 +2,7 @@
 
 import copy
 import json
+import logging
 import os
 import pickle
 import random
@@ -21,6 +22,7 @@ from unittest import mock, SkipTest
 import torch
 import torch.distributed as c10d
 import torch.distributed._functional_collectives as _functional_collectives
+from torch.distributed.distributed_c10d import SHRINK_ABORT as NCCL_SHRINK_ABORT
 
 
 if not c10d.is_available() or not c10d.is_nccl_available():
@@ -47,12 +49,15 @@ from torch._C._distributed_c10d import ErrorType, OpType, WorkResult
 from torch.nn.parallel import DistributedDataParallel
 from torch.testing._internal.common_cuda import _get_torch_rocm_version, TEST_MULTIGPU
 from torch.testing._internal.common_distributed import (
+    get_required_world_size,
     get_timeout,
     init_multigpu_helper,
     MultiProcessTestCase,
     requires_multicast_support,
     requires_nccl,
+    requires_nccl_shrink,
     requires_nccl_version,
+    requires_world_size,
     skip_if_lt_x_gpu,
     skip_if_rocm_multiprocess,
     sm_is_or_higher_than,
@@ -86,6 +91,53 @@ if TEST_WITH_DEV_DBG_ASAN:
 BFLOAT16_AVAILABLE = torch.cuda.is_available() and (
     torch.version.cuda is not None or torch.version.hip is not None
 )
+
+
+_start_time = time.time()
+_logger = logging.getLogger(__name__)
+
+
+def _ts():
+    return time.time() - _start_time
+
+
+def configure(level=logging.INFO, force=False):
+    try:
+        logging.basicConfig(
+            level=level,
+            format="%(asctime)s %(name)s %(levelname)s: %(message)s",
+            force=force,
+        )
+    except TypeError:
+        logging.basicConfig(
+            level=level, format="%(asctime)s %(name)s %(levelname)s: %(message)s"
+        )
+
+
+def log_test_info(rank, message):
+    _logger.info("[%7.3fs][Rank %s] %s", _ts(), rank, message)
+
+
+def log_test_success(rank, message):
+    _logger.info("[%7.3fs][Rank %s] ✅ %s", _ts(), rank, message)
+
+
+def log_test_validation(rank, message):
+    _logger.info("[%7.3fs][Rank %s] ✓ %s", _ts(), rank, message)
+
+
+def log_test_warning(rank, message):
+    _logger.warning("[%7.3fs][Rank %s] ⚠️ %s", _ts(), rank, message)
+
+
+def log_test_error(rank, message):
+    _logger.error("[%7.3fs][Rank %s] ✗ %s", _ts(), rank, message)
+
+
+_log_configure = configure
+
+
+_log_configure(level=logging.INFO, force=True)
 
 
 class RendezvousEnvTest(TestCase):
@@ -317,7 +369,7 @@ class ProcessGroupNCCLGroupTest(MultiProcessTestCase):
 
     @property
     def world_size(self):
-        return 2
+        return get_required_world_size(self, 2)
 
     @property
     def rank_to_GPU(self):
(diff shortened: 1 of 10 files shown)


torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
```diff
@@ -5803,6 +5803,113 @@ at::Tensor ProcessGroupNCCL::allocateTensor(
...
+#define NCCL_MIN_VERSION_FOR_SHRINK NCCL_VERSION(2, 27, 0)
+#endif
+  auto runtime_version = torch::cuda::nccl::version();
+  TORCH_CHECK(
+      runtime_version >= NCCL_MIN_VERSION_FOR_SHRINK,
+      "ProcessGroupNCCL::shrink_process_group requires NCCL version 2.27.0 or later. "
+      "Found version: ", runtime_version);
+
+  // Early validation with detailed error messages
+  TORCH_CHECK(!ranks_to_exclude.empty(),
```
Skylion007:
```suggestion
  TORCH_CHECK_VALUE(!ranks_to_exclude.empty(),

**Comment by syed-ahmed** [comment]
Related Issue: https://github.com/pytorch/pytorch/issues/155472

File: torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp [hunk 2]
```diff
@@ -5803,6 +5803,113 @@ at::Tensor ProcessGroupNCCL::allocateTensor(
...
+      "ranks_to_exclude cannot be empty");
+  TORCH_CHECK(static_cast<int>(ranks_to_exclude.size()) < size_,
```
Skylion007: Should be CHECK_VALUE here too

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
```diff
@@ -5803,6 +5803,113 @@ at::Tensor ProcessGroupNCCL::allocateTensor(
...
+      "Cannot exclude all ranks (", ranks_to_exclude.size(), " >= ", size_, ")");
+
+  // Validate ranks and convert to int efficiently
+  std::vector<int> int_ranks_to_exclude;
+  int_ranks_to_exclude.reserve(ranks_to_exclude.size());
+  for (int64_t rank : ranks_to_exclude) {
+    TORCH_CHECK(rank >= 0 && rank < size_,
```
Skylion007: Should be TORCH_CHECK_VALUE

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
```diff
@@ -5803,6 +5803,113 @@ at::Tensor ProcessGroupNCCL::allocateTensor(
...
+      // Batch the map operations
+      new_pg->devNCCLCommMap_[key] = shrunk_comm;
+      new_pg->ncclStreams_.emplace(key, new_stream);
+      new_pg->ncclEvents_.emplace(key, at::cuda::CUDAEvent(cudaEventDisableTiming));
+      new_pg->usedDeviceIdxs_.insert(new_pg_device.index());
+
+      // Register in global map if using allocator hooks (optimized check)
+      if (shouldAllCommunicatorsRegisterAllTensors()) {
+          std::lock_guard<std::mutex> map_lock(ncclCommMemPoolMapMutex);
+          ncclCommMemPoolMap.emplace(shrunk_comm, MemPoolSet{});
```
Skylion007: Shrunk come should be std::move into emplace here

Skylion007 (dismissed):

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
```diff
@@ -5803,6 +5803,113 @@ at::Tensor ProcessGroupNCCL::allocateTensor(
...
+  std::shared_ptr<NCCLComm> primary_comm = getNCCLComm(primary_key);
+  TORCH_CHECK(primary_comm,
+      "Primary NCCL communicator for device ", primary_device,
+      " (key: ", primary_key, ") is not initialized");
+
+  // Cache device index before shrink operation
+  at::DeviceIndex parent_device_index = primary_comm->getDeviceIndex();
+
+  // Perform shrink operation with optimized config
+  ncclConfig_t config = NCCL_CONFIG_INITIALIZER;
```
kwen2501: Why `NCCL_CONFIG_INITIALIZER` instead of the source comm's config?

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
```diff
@@ -5803,6 +5803,113 @@ at::Tensor ProcessGroupNCCL::allocateTensor(
...
+      new_pg->ncclStreams_.emplace(key, new_stream);
+      new_pg->ncclEvents_.emplace(key, at::cuda::CUDAEvent(cudaEventDisableTiming));
+      new_pg->usedDeviceIdxs_.insert(new_pg_device.index());
+
+      // Register in global map if using allocator hooks (optimized check)
+      if (shouldAllCommunicatorsRegisterAllTensors()) {
+          std::lock_guard<std::mutex> map_lock(ncclCommMemPoolMapMutex);
+          ncclCommMemPoolMap.emplace(shrunk_comm, MemPoolSet{});
+      }
+  }
```
kwen2501: Can this setup code be made more maintainable?

torch/csrc/distributed/c10d/init.cpp
```diff
@@ -3349,6 +3349,14 @@ options :class:`~torch.distributed.ProcessGroupNCCL.Options`).
               },
               py::arg("enable_nan_check"),
               py::call_guard<py::gil_scoped_release>())
+#ifdef NCCL_HAS_COMM_SHRINK
+          .def(
+              "shrink_process_group",
```
kwen2501: nit: `pg.shrink(...)` is shorter than `pg.shrink_process_group(...)`

test/distributed/logging_utils.py
```diff
@@ -0,0 +1,38 @@
+import logging
```
kwen2501: Perhaps rename this file without the "test_" prefix?
In some CI scripts, "test_" prefix is used to identify tests to run.

kwen2501 (commented):
Mostly look good to me. Left some minor comments.

test/distributed/logging_utils.py
brchang24: changed to logging_utils.py

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
brchang24: changed

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
brchang24: Changed

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
brchang24: added std::move

torch/csrc/distributed/c10d/init.cpp
brchang24: changed to shrink

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
brchang24: Sure, moved the setup code to a separate function.

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
brchang24: Good catch! Changed to nullptr to inherent from the source comm.

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
brchang24: changed

test/distributed/test_c10d_nccl.py
```diff
@@ -1255,6 +1277,477 @@ def test_set_process_group_desc(self):
         pg_2 = c10d.new_group([0, 1])
         self.assertEqual(pg_2.group_desc, "undefined")
 
+    @requires_nccl()
+    @skip_but_pass_in_sandcastle_if(int(os.environ.get("WORLD_SIZE", "1")) < 2, "Test requires world size >= 2")
+    @skip_if_lt_x_gpu(int(os.environ.get("WORLD_SIZE", "1")))
+    def test_shrink_group_basic(self):
+        """Test basic shrink_group functionality."""
```
syed-ahmed, line 1313: We need to verify if these tests actually ran in the upstream CI.

torch/csrc/distributed/c10d/NCCLUtils.cpp
```diff
@@ -259,6 +259,64 @@ std::shared_ptr<NCCLComm> NCCLComm::split(
...
+  } catch (const std::exception& e) {
+    // Fallback: if ncclCommUserRank fails, we can't determine the rank
+    LOG(ERROR) << "Failed to query NCCL-assigned rank: " << e.what();
+    throw;
+  }
+
+  // Child comm should be on the same device as parent comm
+  comm->deviceIndex_ = source->deviceIndex_;
+  if (config != nullptr) {
+    comm->nonBlocking_ = config->blocking == 0;
```
syed-ahmed, line 308: How about other configs like minCTAs, maxCTAs, cgaClusterSize etc.?

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
```diff
@@ -5807,6 +5803,110 @@ at::Tensor ProcessGroupNCCL::allocateTensor(
...
 
+#ifdef NCCL_HAS_COMM_SHRINK
+// Define the function only if NCCL supports shrink
+c10::intrusive_ptr<ProcessGroupNCCL> ProcessGroupNCCL::shrink(
+    const std::vector<int64_t>& ranks_to_exclude,
+    int shrink_flags) {
+
+  // Runtime version check with better error message
+#ifndef NCCL_MIN_VERSION_FOR_SHRINK
+#define NCCL_MIN_VERSION_FOR_SHRINK NCCL_VERSION(2, 27, 0)
```
syed-ahmed: nit: do you need to define this macro or is it possible to directly put `NCCL_VERSION(2, 27, 0)` in the runtime check?

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
```diff
@@ -5807,6 +5803,110 @@ at::Tensor ProcessGroupNCCL::allocateTensor(
...
+#endif
+  auto runtime_version = torch::cuda::nccl::version();
+  TORCH_CHECK(
+      runtime_version >= NCCL_MIN_VERSION_FOR_SHRINK,
+      "ProcessGroupNCCL::shrink requires NCCL version 2.27.0 or later. "
+      "Found version: ", runtime_version);
+
+  // Early validation with detailed error messages
+  TORCH_CHECK_VALUE(!ranks_to_exclude.empty(),
```
syed-ahmed: nit: is this check redundant since there is another one in `NCCLComm::shrink`? Do you anticipate using NCCLComm::shrink somewhere else and is that why the check needs to be in both of these functions?

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
```diff
@@ -5838,6 +5803,123 @@ at::Tensor ProcessGroupNCCL::allocateTensor(
...
+#endif
+  auto runtime_version = torch::cuda::nccl::version();
+  TORCH_CHECK(
+      runtime_version >= NCCL_MIN_VERSION_FOR_SHRINK,
+      "ProcessGroupNCCL::shrink requires NCCL version 2.27.0 or later. "
+      "Found version: ",
+      runtime_version);
+
+  // Early validation with detailed error messages
+  TORCH_CHECK_VALUE(
```
syed-ahmed, line 5859: nit: Is this check redundant since there is another one in NCCLComm::shrink? Do you anticipate using NCCLComm::shrink somewhere else and is that why both of these functions need this check?

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
```diff
@@ -5838,6 +5803,123 @@ at::Tensor ProcessGroupNCCL::allocateTensor(
...
+      primary_device,
+      " (key: ",
+      primary_key,
+      ") is not initialized");
+
+  // Cache device index before shrink operation
+  at::DeviceIndex parent_device_index = primary_comm->getDeviceIndex();
+
+  std::shared_ptr<NCCLComm> shrunk_comm = NCCLComm::shrink(
+      primary_comm.get(), int_ranks_to_exclude, nullptr, shrink_flags);
```
syed-ahmed: You read from the config in NCCL::shrink but you are passing a nullptr here. What's the plan for handling user defined ncclConfig_t?

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
```diff
@@ -5838,6 +5803,123 @@ at::Tensor ProcessGroupNCCL::allocateTensor(
...
+  int new_size = size_ - static_cast<int>(ranks_to_exclude.size());
+  int new_rank = shrunk_comm->rank_;
+
+  // Create new ProcessGroupNCCL with optimized options cloning
+  auto new_store = store_->clone();
+  auto new_opts =
+      ProcessGroupNCCL::Options::create(options_->is_high_priority_stream);
+  new_opts->timeout = options_->timeout;
+#ifdef NCCL_HAS_CONFIG
+  new_opts->config = options_->config;
```
syed-ahmed: Should this `options_->config` be passed in the NCCLComm::shrink above instead of the nullptr?

torch/distributed/distributed_c10d.py
```diff
@@ -5671,3 +5672,523 @@ def _get_process_group_name(pg: ProcessGroup) -> str:
...
+    Args:
+        ranks_to_exclude (List[int]): A list of ranks from the original
+            ``group`` to exclude from the new group.
+        group (ProcessGroup, optional): The process group to shrink. If ``None``,
+            the default process group is used. Defaults to ``None``.
+        shrink_flags (int, optional): Flags to control the shrinking behavior.
+            Can be ``NCCL_SHRINK_DEFAULT`` (default) or ``NCCL_SHRINK_ABORT``.
+            ``NCCL_SHRINK_ABORT`` will attempt to terminate ongoing operations
+            in the parent communicator before shrinking.
+            Defaults to ``NCCL_SHRINK_DEFAULT``.
```
kwen2501: This API surface is supposed to be vendor agnostic. Can you try naming the flags that way?

torch/csrc/distributed/c10d/init.cpp
```diff
@@ -3349,6 +3349,14 @@ options :class:`~torch.distributed.ProcessGroupNCCL.Options`).
...
               py::arg("enable_nan_check"),
               py::call_guard<py::gil_scoped_release>())
+#ifdef NCCL_HAS_COMM_SHRINK
+          .def(
+              "shrink",
+              &::c10d::ProcessGroupNCCL::shrink,
+              py::arg("ranks_to_exclude"),
+              py::arg("shrink_flags") = 0,
+              py::call_guard<py::gil_scoped_release>())
+#endif // NCCL_HAS_COMM_SHRINK
```
kwen2501: It may be ideal to expose this feature in `backend.hpp`.
For example:
https://github.com/pytorch/pytorch/blob/c9b2a0953017053aa3abdcfcfd0e0faa4845d0db/torch/csrc/distributed/c10d/Backend.hpp#L463-L468

torch/distributed/distributed_c10d.py
```diff
@@ -5671,3 +5672,523 @@ def _get_process_group_name(pg: ProcessGroup) -> str:
...
+            "Ensure PyTorch was compiled with NCCL support and NCCL libraries are installed."
+        )
+
+    # Validate NCCL version supports shrinking
+    nccl_version = torch.cuda.nccl.version()
+    if nccl_version < NCCL_MIN_VERSION_FOR_SHRINK:
+        raise RuntimeError(
+            f"ProcessGroupNCCL.shrink requires NCCL version >= {NCCL_MIN_VERSION_FOR_SHRINK}. "
+            f"Found NCCL version {nccl_version}. Please update NCCL to use shrinking functionality."
+        )
```
kwen2501: Should we move this type of checks to cpp? (so that a backend's support check is within the backend itself). For example:
https://github.com/pytorch/pytorch/blob/c9b2a0953017053aa3abdcfcfd0e0faa4845d0db/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp#L5783-L5790

torch/distributed/distributed_c10d.py
```diff
@@ -5671,3 +5672,523 @@ def _get_process_group_name(pg: ProcessGroup) -> str:
...
+    if hasattr(target_pg, "bound_device_id"):
+        bound_device_id = target_pg.bound_device_id
+
+    # Generate new group name for the shrunk group
+    remaining_ranks = [rank for rank in get_process_group_ranks(target_pg)]
+    new_group_name = _process_group_name(remaining_ranks, use_hashed_name=False)
+
+    return {
+        "backend_name": original_backend_name,
+        "store": original_store,
```
kwen2501, line 6009: How true is it that the original store can be reused?

test/distributed/test_c10d_nccl.py
```diff
@@ -87,6 +89,19 @@
     torch.version.cuda is not None or torch.version.hip is not None
 )
 
+sys.path.insert(0, os.path.dirname(os.path.abspath(__file__)))
```
kwen2501: what does this line do?

torch/csrc/distributed/c10d/NCCLUtils.cpp
brchang24: We pass the full ncclConfig_t to NCCL; NCCL applies fields like minCTAs, maxCTAs, cgaClusterSize, etc, and those are not used by the wrapper. We derive nonBlocking_ from config.blocking and use it in getNcclComm()/waitReady() to decide when/how to wait for initialization.

test/distributed/test_c10d_nccl.py
brchang24: With @eqy help, I realized all the tests are run but got skipped due to not enough GPUs. Now the test issue is identified, and the latest commit should enable the CI tests. The test now can expand the default 2 GPU to max GPUs the test requires.

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
brchang24: directly check with the latest commit.

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
brchang24: Unified all checks in the backend ProcessGroupNCCL layer and removed checks in PyTorch and wrapper layer.

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
brchang24: the checks should now only in ProcessGroupNCCL backend layer.

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
brchang24: Was thinking to simplify the config and only inherent from parent, but for flexibility, added the user conifg parameter for the user in the latest commit, and that seems not adding much complexity.

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
brchang24: Good catch. This should be fixed now with the latest commit.

torch/distributed/distributed_c10d.py
brchang24: Good point, I now have removed all vendor specific code, and comments as well.

torch/csrc/distributed/c10d/init.cpp
brchang24: Didn't realized the base class, was always thinking how to enable NCCL. Now the code should be more vendor agnostic

torch/distributed/distributed_c10d.py
brchang24: yes, now all checks should be only within backend

test/distributed/test_c10d_nccl.py
brchang24: It supposed to add this test file’s directory to the beginning of Python’s import search path so taking precedence over site-packages. This can make sure the local module overrides system-installed ones in case the same name.

But definitely can be removed if this is already the case.

torch/distributed/distributed_c10d.py
brchang24: Just like split, it reuses parent store, but it does clone on it, so I also followed split to clone the parent store with the latest commit.

test/distributed/test_c10d_nccl.py
syed-ahmed: I think what @kwen2501 is asking is why do you need this? Does the test not run without it? Usually I think inserting path into sys.path is a workaround/hack.

brchang24: No, it is not needed. Guess the search path should be added by either CI or the user who runs the test, but not he code itself. I can remove the line in the next commit.

brchang24: removed in the latest commit

test/distributed/test_c10d_nccl.py
brchang24: The latest CI is all green

Also confirmed the new Shrink tests are run and passed in the CI!

https://gha-artifacts.s3.amazonaws.com/pytorch/pytorch/18549341435/1/artifact/logs-test-distributed-3-3-lf.linux.g4dn.12xlarge.nvidia.gpu_52875791196.zip

brchang24:
The latest CI is passed, and all comments should be resolve. Please let me know if there is more opens.

syed-ahmed (dismissed):
LGTM.

torch/distributed/distributed_c10d.py
kwen2501: What I mean is that when a user needs to shrink a group, it is usually because some ranks failed. 
If the store server happens to live on the failed rank, then this store could be gone too.

kwen2501 (approved):

torch/distributed/distributed_c10d.py
brchang24: You're right. Resilient communication alone isn't enough for true resiliency; other components, like the storage, must also be resilient. Otherwise, it's just a best effort.

Additionally, "shrink" operations aren't always a response to failures. They can also occur due to demand fluctuations, management decisions (e.g., reducing ranks to save costs during low demand), or when ranks need to enter service mode for software updates, etc.

brchang24:
> @pytorchbot revert -m "Breaks lint" -c nosignal
> 
> See https://hud.pytorch.org/pytorch/pytorch/commit/4e643422f63a3cdd71bd141615f98de6bb54d15f for details. Not sure why this didn't fail on the PR though

I have seen this failure in one of the merge failure log, but not in PR CI and none of my local run can show this issue.

Just pushed another fix, hope it solved this issue. But a consistent CI will avoid this issue and can avoid extra CI runs.

syed-ahmed:
@brchang24 I don't think we can repro the internal failure. I think the error is happening because in test_c10d_nccl.py, you are importing logging_utils, but it needs to be in some module. I think the easiest way is to move your logging functions to `torch.testing._internal.common_distributed` (while making sure you are using the logger object from there instead of defining your own).

nWEIdia:
For future reference, this kind (NCCL) changes should apply distributed label to go through H100 or B200 distributed testing.

PR #164519, opened 2025-10-02
Reapply "C++-accessible Placements via pybind11 (#163030)"

swolchok:
* __->__ #164519

This makes Placement data representation available in C++ via pybind11. Reapply with fix for internal errors.

D83788896

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @kadeng @chauhang @amjames @Lucaskabela @jataylo @chenyang78


torch/csrc/distributed/Placement.h
```diff
@@ -0,0 +1,121 @@
...
+class Shard : public Placement {
+ public:
+  std::int64_t dim;
+  explicit Shard(std::int64_t dim_) : dim(dim_) {}
+
+  bool is_shard(std::optional<std::int64_t> dim_) const override {
+    return !dim_.has_value() || *dim_ == dim;
+  }
+
+  bool operator==(const Shard& rhs) const {
```
Skylion007, line 47: Nit: we should add a comment to default these methods in C++20

torch/csrc/distributed/python_placement.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+#include <pybind11/pybind11.h>
+#include <torch/csrc/distributed/Placement.h>
+#include <torch/csrc/utils/pybind.h>
+
+using namespace pybind11::literals;
+
+namespace torch::distributed {
+namespace {
+const auto placement_class_docstring =
```
Skylion007, line 11: Mark as constexpr

torch/distributed/tensor/placement_types.py
```diff
@@ -698,11 +646,6 @@ def _partition_value(
         num_chunks = mesh.size(mesh_dim=mesh_dim)
         return tensor / num_chunks
 
-    def __eq__(self, other: object) -> bool:
-        if not isinstance(other, Partial):
-            return False
-        return self.reduce_op == other.reduce_op
-
     def __hash__(self) -> int:
```
Skylion007, line 649: Should we rely on super method hashing here now?

torch/distributed/tensor/placement_types.py
```diff
@@ -718,11 +661,6 @@ def __str__(self) -> str:
         """
         return "P"
```
Skylion007, line 662: This is a super weird str method btw

torch/distributed/tensor/placement_types.py
```diff
@@ -349,11 +324,6 @@ def _to_new_shard_dim(
 
         return new_tensor
 
-    def __eq__(self, other: object) -> bool:
-        if not isinstance(other, Shard):
-            return False
-        return self.dim == other.dim
-
     def __hash__(self) -> int:
```
Skylion007, line 327: Same here, why not define repr and eq, and hash in C++ for consistency?

torch/csrc/distributed/python_placement.cpp
```diff
@@ -0,0 +1,104 @@
...
+      .def(
+          "__eq__",
+          [](const Partial& lhs, const Partial& rhs) { return lhs == rhs; },
+          py::is_operator())
+      .def(py::pickle(
+          [](const Partial& part) {
+            return py::dict("reduce_op"_a = part.reduce_op);
+          },
+          [](const py::dict& d) {
+            return Partial(py::cast<std::string>(d["reduce_op"]));
```
Skylion007, line 101: Not sure this cast is necessary if you add a py::init<std::optional<py::str>> declaration to the class

Skylion007 (approved):

torch/csrc/Module.cpp
```diff
@@ -2111,6 +2112,8 @@ PyObject* initModule() {
   THXPEvent_init(module);
 #endif
 
+  torch::distributed::initPlacementBindings(module);
```
Skylion007, line 2115: FYI, we should make another PR that does a find and replace here for module as that becomes a kwarg in more recent C++

torch/csrc/distributed/Placement.h
swolchok: eh, I think needing to look at that is implicit for all operator== everywhere

torch/distributed/tensor/placement_types.py
swolchok: hm, I agree that `__hash__` should be defined next to `__eq__`. I'll fix that. for `__repr__` I'm trying to force as little as possible to be done in C++, and it's fundamentally a Python thing.

torch/distributed/tensor/placement_types.py
swolchok: Why? I don't think there's an impetus for an algorithm change here.

torch/distributed/tensor/placement_types.py
swolchok: I agree but it's definitely out of scope

torch/csrc/Module.cpp
swolchok: agree in principle but out of scope for this PR

swolchok: also module is *not* a keyword: https://en.cppreference.com/w/cpp/identifier_with_special_meaning/module.html

ezyang (approved):

izaitsevfb:
@pytorchbot merge

verified the diff and pr are in sync (except the TARGETS file)

PR #164520, opened 2025-10-02
Add view support for library custom Function

albanD:
* #164359
* __->__ #164520


torch/csrc/autograd/autograd_not_implemented_fallback.cpp
```diff
@@ -478,6 +478,40 @@ torch::CppFunction autogradNotImplementedFallback() {
...
+        "ADInplaceOrView fallback view replay did not return a Tensor");
+    return result.toTensor();
+  }
+
+  std::unique_ptr<ViewFunc> clone_and_set(
+      std::optional<std::vector<c10::SymInt>> = std::nullopt,
+      std::optional<std::vector<at::Tensor>> = std::nullopt) const override {
+    return std::make_unique<GenericViewFunc>(
+        non_tensor_stack_, aliased_input_idx_val_, op_);
+  }
```
soulitzer, line 524: For the case with compile where the view is passed as an input (and there are things that need to be symintified), we may need to figure out how to swap the symints somehow

test/test_custom_ops.py
```diff
@@ -2550,6 +2551,56 @@ def sin_(x):
...
+    @skipIfTorchDynamo("Expected to fail due to no FakeTensor support; not a bug")
+    def test_subclass_accessor_view(self):
+        class MyTwoTensor(TwoTensor):
+            @classmethod
+            def __torch_dispatch__(cls, func, types, args, kwargs):
+                if func is torch.ops._torch_testing._two_tensor_accessor.default:
+                    self.assertIsInstance(args[0], MyTwoTensor)
+                    res = args[0].a
+                    # Always return a fresh Tensor!
+                    return res.view_as(res)
```
soulitzer, line 2607: No action necessary, but now that custom ops have official support for views, I wonder if there's things we can do to make the inference mode handling smoother. 

With this PR today, I guess wrapper subclasses still need special handling for inference mode for views (where ordinarily inference-ness being preserved via this line, but can be broken if the outer tensor must be rewrapped in a new subclass)

torch/csrc/autograd/autograd_not_implemented_fallback.cpp
albanD: tbh there was so much logic there I wasn't sure what to do with it.
Do we actually use this ever?

test/test_custom_ops.py
albanD: Do you have code sample that shows this concern?

soulitzer: Because LoggingTensor doesn't do this special inference-mode handling for views, doing
```python
import torch
from torch.testing._internal.logging_tensor import LoggingTensor

base = LoggingTensor(torch.rand(1, 2, 3)).requires_grad_(True)

with torch.inference_mode(True):
    base.view(-1)
```
Fails with
```
  File "/Users/jw3468/local/.tests/tst1.py", line 9412, in <module>
    base.view(-1)
RuntimeError: Cannot set version_counter for inference tensor
```
Example of how it is handled in nested tensor:
```python
def view_default(func, *args, **kwargs):
    ...
   with torch.inference_mode(inp.is_inference()):
        return NestedTensor(func(inp._values, inner_size), **extract_kwargs(inp))
 ```

File: torch/csrc/autograd/autograd_not_implemented_fallback.cpp (hunk 1 above)
Comment (soulitzer) [reply torch/csrc/autograd/autograd_not_implemented_fallback.cpp:524]: It's used for when we fakification views of subclasses. For non-subclasses, we go through the usual as_strided path.

https://github.com/pytorch/pytorch/blob/3c59351c6ea2fc29d346903e28e95c5f4d0ccdbb/torch/_subclasses/meta_utils.py#L1307

Comment (albanD) [reply torch/csrc/autograd/autograd_not_implemented_fallback.cpp:524]: We never capture any Tensor, so that shouldn't be a problem.
But I guess there are symints yes.

What would be a minimal test for that?

Comment (soulitzer) [reply torch/csrc/autograd/autograd_not_implemented_fallback.cpp:524]: Hmm found this test from the original [PR](https://github.com/pytorch/pytorch/pull/118404) that adds this cc @jbschlosser 
```python
with torch.autograd._force_original_view_tracking(True):
    base = torch.randn(3, 4, 5)
    view = base.select(1, 2)
    
    def symint_visitor_fn(x):
        # modify saved index
        return x + 1
    
    # ensure modifying state changes view replay
    new_base = torch.randn_like(base)
    new_view = view._view_func(new_base, symint_visitor_fn=symint_visitor_fn)
    self.assertEqual(new_view, new_base.select(1, 3))
  ```
Looking at the implementation of _view_func, it seems like the view author should tell us which are the integers that need to be swapped.

File: test/test_custom_ops.py (hunk 2 above)
Comment (albanD) [reply test/test_custom_ops.py:2607]: This is unrelated to the change at hand so will move that to a follow up

File: torch/csrc/autograd/autograd_not_implemented_fallback.cpp (hunk 1 above)
Comment (albanD) [reply torch/csrc/autograd/autograd_not_implemented_fallback.cpp:524]: From looking at this more, it is not trivial to handle this. So I added appropriate error checking for now and we can add support in a follow up if needed.

File: torch/testing/_internal/two_tensor.py [hunk 3]
```diff
@@ -9,7 +9,7 @@
 # A simple tensor subclass that holds two tensors internally, and runs every op on both tensors.
 class TwoTensor(torch.Tensor):
     @staticmethod
-    def __new__(cls, a, b, outer_size=None, outer_stride=None):
+    def __new__(cls, a, b, outer_size=None, outer_stride=None, *, requires_grad=None):
```
soulitzer, line 12: no action necessary, we could probably deduplicate so that the requires_grad of the inner tensor doesn't matter (requires refactoring all sites that use TwoTensor). I actually had a PR to do this lol, but need to pick it up again...

torch/csrc/autograd/autograd_not_implemented_fallback.cpp
```diff
@@ -608,13 +680,22 @@ static void autogradNotImplementedInplaceOrViewFallbackImpl(
       auto result = std::move(aliased_output);
       stack->at(stack->size() - num_returns + aliased_output_idx) = result;
     } else {
+      c10::IValue& aliased_output_iv =
+          (*stack)[stack->size() - num_returns + aliased_output_idx];
       TORCH_CHECK(aliased_output_iv.isTensor());
+      TORCH_CHECK(
+          num_returns == 1,
```
soulitzer, line 678: I guess this is more strict than the autograd.Function case because autograd.Function would allow multiple outputs as long as there is on a single tensor output?

torch/csrc/autograd/autograd_not_implemented_fallback.cpp
```diff
@@ -608,13 +680,22 @@ static void autogradNotImplementedInplaceOrViewFallbackImpl(
...
+          "ADInplaceOrView fallback only support single output view functions");
```
soulitzer, line 679: Looks like the TensorList case returns a different error. For consistency, should we just ban that entire case as well, and then be able to delete the erroring_view_func?

Or alternatively, we could have the  non-TensorList case also use the erroring_view_func when there's multiple outputs instead of erroring immediately.

torch/_library/utils.py
```diff
@@ -74,7 +75,7 @@ def is_builtin(op: OpOverload) -> bool:
     return op.namespace in {"aten", "prim", "prims"}
 
 
-def is_functional_schema(schema: Any) -> bool:
+def is_functional_schema(schema: Any, *, view_ok: bool = False) -> bool:
```
soulitzer: Just to keep this function simpler, maybe have a separate "is_valid_view" helper and everywhere we call
instead of is_functional_schema(schema, view_ok=True)
do 
is_functional_schema(schema) or is_valid_view(schema)

torch/_library/custom_ops.py
```diff
@@ -632,20 +634,27 @@ def fake_impl(*args, **kwargs):
...
                 for idx in mutated_idxs:
                     increment_version(args[idx])
                 for key in mutated_keys:
                     increment_version(kwargs[key])
-                with _C._AutoDispatchBelowADInplaceOrView():
-                    return self._opoverload.redispatch(
-                        keyset & _C._after_ADInplaceOrView_keyset, *args, **kwargs
-                    )
+                # Handle view + mutation that are in the schema
+                return original_kernel.call_boxed(keyset, *args, **kwargs)
```
soulitzer, line 657: If we're calling into the fallback now for the is_mutable case as well, being able to remove the whole branch would be a good simplification!
Is there something preventing us from doing that?

soulitzer (approved):
Cool! Added some minor comments

torch/csrc/autograd/autograd_not_implemented_fallback.cpp
albanD: Yes, this is only for simplicity of the code right now. We can extend

torch/csrc/autograd/autograd_not_implemented_fallback.cpp
albanD: I don't see any user of this code, so I took some liberty in changing the current behavior yes.
For now TensorList have the erroring_view_func and multi-output error eagerly.

I think it is fine for the limited usage we're looking at. But we can extend it easily if/when needed yes.

torch/_library/custom_ops.py
albanD: No we cannot :/ 
The inplace logic is based on user anotation and the schema is not actually made to reflect these inplace. So we do need this custom python logic to do the increment_version as the fallback won't do them.

We could refactor the whole thing to have accurate schema and rely on the fallback to do all incref though. But that is much more involved.

test/test_custom_ops.py
```diff
@@ -18,11 +18,11 @@
 import numpy as np
 import yaml
 
+import functorch
```
ezyang: lol

ezyang: You can just import make_fx directly from torch.fx.experimental.proxy_tensor

torch/testing/_internal/two_tensor.py
ezyang: Yes I deeply prefer this too

torch/csrc/autograd/autograd_not_implemented_fallback.cpp
```diff
@@ -478,6 +478,58 @@ torch::CppFunction autogradNotImplementedFallback() {
...
+        TORCH_CHECK(
+            ct->kind() != c10::TypeKind::SymIntType,
+            "Custom ops that are views do not support SymInt. Please file an issue if you need it.");
+      }
+    }
+  }
+
+  at::Tensor operator()(const at::Tensor& new_base) const override {
+    torch::jit::Stack local_stack = non_tensor_stack_;
+    auto hmmm = c10::IValue(new_base);
```
ezyang: hmmm?

torch/csrc/autograd/autograd_not_implemented_fallback.cpp
```diff
@@ -478,6 +478,58 @@ torch::CppFunction autogradNotImplementedFallback() {
       &autogradNotImplementedFallbackImpl>();
 }
 
+struct GenericViewFunc : public ViewFunc {
```
ezyang, line 481: So is this literally just "remember all non-Tensor arguments, and then rerun the function if you need to reapply the view"?

ezyang (approved):

torch/csrc/autograd/autograd_not_implemented_fallback.cpp
albanD: Yes.
With extra complexity if there are tensor/symint other arguments, which I bypass

test/test_custom_ops.py
albanD: Ah that's where it is now!
Thanks!

torch/_library/utils.py
albanD: Tried the refactor but I ended up having the same function as it is now. So I am reverting back to this one with a name change to make it easier to follow.

PR #164525, opened 2025-10-03
[inductor] tune some fields for reduction in CDT

shunting314:
* #164533
* #164532
* __->__ #164525
* #163589

With the change, in deterministic mode, we should still be able to tune XBLOCK for a reduction with coordinate descent tuning.

### DIFF
--- a/test/inductor/test_deterministic.py
+++ b/test/inductor/test_deterministic.py
@@ -1,5 +1,6 @@
 # Owner(s): ["module: inductor"]
 import contextlib
+import unittest
 
 import torch
 import torch._inductor.config as inductor_config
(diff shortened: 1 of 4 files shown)


jansel (approved):

eellison (approved):

v0i0 (commented):
I am not sure you can tune XBLOCK like this without affecting reduction order: [N8272384](https://www.internalfb.com/intern/anp/view/?id=8272384)

XBLOCK affects elems per thread, which may affect reduction order

htyu:
> When xblk==1, each thread load 2 elements : rnumel / (num-warps * 32)
When xblk==2, triton let each thread load4 elements to be fully vectorized. This can be done when xblk==2 since we now have 2x number of elements compared to xblk==1.

The thread mapping pattern is recurring when xblk is big enough. Triton tries to layout threads consecutively to favor memory coalescing and vectorization. Per hardware requirement each thread can load up to 16 bytes with one `ld.global.f32.v4` instruction. When `XBLOCK * RBLOCK` is big enough, Triton will lay out threads in a way that each thread access consecutive 16 bytes, thread by thread. This pattern repeats after a round of `num_warps * 32 *16` bytes. However when `XBLOCK * RBLOCK` is smaller than `num_warps * 32 *16`, Triton would assign each thread less than 16 bytes so every thread has some unique work to do. If it still assigned each thread 16 bytes, some threads would end up loading redundant data. The thread/bytes assignment logics is here: https://github.com/facebookexperimental/triton/blob/4f797b6dbbdb7ebea8fcc8238d4221b7da31abb9/lib/Dialect/TritonGPU/Transforms/Coalesce.cpp#L85.

PR #164526, opened 2025-10-03
Add an option to put store large mmap weights on disk

yushangdi:
As title

In windows, we cannot modify the .dll to append weights at the end, the windows .dll loader will complain it's not a valid .dll file. So we store the weight blob as a separete file.

1. We add the following API which allows passing in a pointer to the weight blob and get the size of the weight blob. 

```cpp
AOTI_API AOTIRuntimeError AOTInductorModelContainerGetConstantsBlobSize(
    AOTInductorModelContainerHandle container_handle,
    uint64_t* ret_size);

// Load weights from a single blob in weight_blob_ptr
AOTI_API AOTIRuntimeError AOTInductorModelUpdateConstantsFromBlob(
    AOTInductorModelContainerHandle container_handle,
    const uint8_t* weight_blob_ptr);
```

2. We also add a method in ModelContainerRunner to load the weight:

If the runner see that there is a `.blob` file in the package, if will mmap the .blob file and use the content to load the constants. 

3. We also add the `USE_MMAP_EXTERNAL` macro. When this macro is defined, the model expects to load the weights from external mmap'd weights. 

Test Plan:

```
buck run @mode/dev-nosan caffe2/test/inductor:test_aot_inductor -- -r test_large_mmaped_weights_on_disk
```

Also tested for windows-cross compilation with https://github.com/meta-pytorch/torchnative/blob/6542566585ab2cb4cd6c8e3873a6a02cfe51664b/demo/main_voxtral.cpp 

```
Loaded model.dll
audio_encoder loaded
C:\Users\shangdiy\source\repos\torchnative\demo\token_embedding\data\aotinductor\model\model.wrapper.so
Loaded model.dll
token_embedding loaded
C:\Users\shangdiy\source\repos\torchnative\demo\text_decoder\data\aotinductor\model\model.wrapper.so
Loaded model.dll
Loading weights from C:\Users\shangdiy\source\repos\torchnative\demo\text_decoder\data\aotinductor\model\model.wrapper_weights.blob
text_decoder loaded
Load latency (ms):
  audio_encoder: 1011.234
    archive extraction: 0.000
    .so loading: 1011.197
  token_embedding: 525.773
    archive extraction: 0.000
    .so loading: 525.704
  text_decoder: 3324.130
    archive extraction: 0.000
    .so loading: 3323.979
Run latency (ms):
  audio_encoder: 285.958
    audio_encoder output: dtype=bfloat16, shape=[1, 1125, 3072], numel=3456000
  token_embedding: 6.676
    token_embedding output: dtype=bfloat16, shape=[1, 1138, 3072], numel=3495936
  text_decoder: 576.519
    text_decoder output: dtype=bfloat16, shape=[1, 1138, 131072], numel=149159936
```

### DIFF
--- a/test/inductor/test_aot_inductor.py
+++ b/test/inductor/test_aot_inductor.py
@@ -701,6 +701,24 @@ class AOTInductorTestsTemplate:
         with config.patch({"aot_inductor.force_mmap_weights": True}):
             self.check_model(Model(), example_inputs)
 
+    def test_large_mmaped_weights_on_disk(self):
+        class Model(torch.nn.Module):
+            def __init__(self) -> None:
+                super().__init__()
+                self.linear = torch.nn.Linear(512, 250112)
+
+            def forward(self, x, y):
+                return x + self.linear(y)
+
+        example_inputs = (
+            torch.randn(1, 250112, device=self.device),
+            torch.randn(1, 512, device=self.device),
+        )
+        with config.patch(
+            {"aot_inductor.package_constants_on_disk_format": "binary_blob"}
+        ):
+            self.check_model(Model(), example_inputs)
+
     def test_with_offset(self):
         class Model(torch.nn.Module):
             def __init__(self, device):
@@ -5891,7 +5909,7 @@ class AOTInductorTestsTemplate:
                 {
                     "always_keep_tensor_constants": True,
                     "aot_inductor.package_constants_in_so": False,
-                    "aot_inductor.package_constants_on_disk": True,
+                    "aot_inductor.package_constants_on_disk_format": "pickle_weights",
                     "aot_inductor.package": True,
                 }
             ),
@@ -7430,6 +7448,14 @@ class TestAOTInductorConfig(TestCase):
             }
             with self.assertRaises(RuntimeError):
                 maybe_aoti_standalone_config(patches)
+
+    def test_compile_standalone_cross_compile_windows_package_format(self):
+        patches = {
+            "aot_inductor.cross_target_platform": "windows",
+            "aot_inductor.package_constants_in_so": True,
+        }
+        with self.assertRaises(RuntimeError):
+            maybe_aoti_standalone_config(patches)
 
 
 common_utils.instantiate_parametrized_tests(AOTInductorTestsTemplate)
--- a/torch/_inductor/codecache.py
+++ b/torch/_inductor/codecache.py
@@ -94,6 +94,7 @@ from torch._inductor.runtime.runtime_utils import cache_dir, default_cache_dir
 from torch._inductor.utils import (
     ALIGN_BYTES,
     clear_on_fresh_cache,
+    determine_aoti_mmap_flags,
     is_linux,
     is_windows,
 )
@@ -2175,7 +2176,10 @@ end
                 raw_bytes = bytes(raw_array.contents)
                 return raw_bytes if all_cuda else _pad_to_alignment(raw_bytes)
 
-            if config.aot_inductor.package_constants_in_so:
+            if (
+                config.aot_inductor.package_constants_in_so
+                or config.aot_inductor.package_constants_on_disk_format == "binary_blob"
+            ):
                 serialized_weights = b"".join(
                     _to_bytes(graph.get_original_value_of_constant(name), all_cuda)
                     for name in graph.constants.keys()
@@ -2184,7 +2188,7 @@ end
             else:
                 serialized_weights = b""
 
-            if config.aot_inductor.package_constants_on_disk:
+            if config.aot_inductor.package_constants_on_disk_format == "pickle_weights":
                 # We need to return a storage key here because the original value tensor might be a clone
                 weights_dict = Weights(
                     {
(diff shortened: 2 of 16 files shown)


test/inductor/test_aot_inductor.py
```diff
@@ -698,6 +698,22 @@ def forward(self, x, y):
...
+                self.linear = torch.nn.Linear(512, 250112)
+
+            def forward(self, x, y):
+                return x + self.linear(y)
+
+        example_inputs = (
+            torch.randn(1, 250112, device=self.device),
+            torch.randn(1, 512, device=self.device),
+        )
+        with config.patch({"aot_inductor.force_mmap_weights_on_disk": True}):
```
desertfire: I feel this config naming is not very user-friendly. Maybe something less implementation detail oriented, but more from user's perspective?

torch/_inductor/codegen/cpp_wrapper_cpu.py
```diff
@@ -780,7 +780,11 @@ def codegen_model_constructor(self):
                                        {num_constants},
                                        device_str,
                                        std::move(cubin_dir),
+#if defined(USE_MMAP_EXTERNAL)
+                                       false) {{
+#else
                                        {include_weights}) {{
```
desertfire, line 786: We should be able to tell include_weights is false at the codegen time?

torch/_inductor/utils.py
```diff
@@ -3612,6 +3612,54 @@ def force_patch_config(
     return config_patches
 
 
+def determine_aoti_mmap_flags(consts_size: int) -> tuple[bool, bool]:
```
desertfire, line 3646: For future work, we should really consolidate to use only one option: always store weights as a separate file.

torch/csrc/inductor/aoti_runner/model_container_runner.h
```diff
@@ -55,6 +55,7 @@ class TORCH_API AOTIModelContainerRunner {
       AOTInductorStreamHandle cuda_stream_handle = nullptr);
   void swap_constant_buffer();
   void free_inactive_constant_buffer();
+  void update_constant_buffer_from_blob(std::string weights_path);
```
desertfire: Use "const std::string&"

torch/_inductor/codecache.py
```diff
@@ -2285,7 +2301,15 @@ def _pad_to_alignment(raw_bytes: bytes) -> bytes:
             if not use_mmap_weights:
                 aot_constants = serialized_weights
                 magic_number = 0
+                if use_external_weights:
+                    aot_constants = struct.pack("q", consts_size)
+                    assert external_weights_path is not None
+                    # For external weights, write weights to separate file and embed minimal placeholder
+                    with open(external_weights_path, "wb") as f_weights:
+                        f_weights.write(serialized_weights)
```
desertfire, line 2311: This is fine for the current purpose, but I am debating with myself whether we should add more metadata to this file, some kind of header information, for potential future extension (e.g. versioning).

test/inductor/test_aot_inductor.py
yushangdi: Maybe I can change the existing `package_constants_on_disk: bool = False` into `package_constants_on_disk_format`? And the options are "pickle_weights" (we save each weight in pickle format) and "binary_blob" (we save all weights in a single binary blob).

torch/_inductor/codecache.py
yushangdi: yeah that makes sense. We can do that in a future PR.

torch/_inductor/codegen/cpp_wrapper_cpu.py
yushangdi: Not always. If we don't have the `force_mmap_weights_on_disk` flag on, we might still have to mmap weights on disk if the weights are too large. But we only know the size after we serialize it in `codecache.py`. 

Alternatively I can using some string matching technique to modify the include weights here in `codecache.py` but that seems hacky. 

Or I can compute the size of the weights here, but seems a little weird to serialize the weights twice.

desertfire: Fine to use a macro, but let's come up with a better name.

yushangdi: discussed offline, this `include_weights` mainly controls if `load_constants` run at model creation time. We'll keep the current impl.

test/inductor/test_aot_inductor.py
yushangdi: changed now.

torch/_inductor/config.py
```diff
@@ -1584,8 +1584,21 @@ class aot_inductor:
...
+    # Options:
+    # None:
+    #       Do not package weight separately on disk.
+    # "pickle_weights":
+    #       Each weight is pickled and stored separately in data/weights. We also store the
+    #       FQN names of each weight in a weights_config.json in each model's data/aot_inductor/model folder.
+    #       Can only be load back from python using torch._inductor.aoti_load_package API now.
+    # "binary_blob":
+    #       Stores all weights in a single binary blob in data/aot_inductor/model folder for each model.
+    #       When package_constants_in_so=True, cross_target_platform="windows" and we have >2GB weights,
```
desertfire: Can we force it to be always true when cross_target_platform="windows", despite the weight size? I think that will simplify things a bit.

desertfire (approved):
Minor followup. LGTM.

torch/_inductor/config.py
yushangdi: > Can we force it to be always true when cross_target_platform="windows", despite the weight size? I think that will simplify things a bit.

@JacobSzwejbka  does this work for you? If we make it always true, then windows will always have a separate weight.

yushangdi: ok, updated now. 

In the current version, using `package_constants_in_so` with cross-compilation will error out. So the only two options for cross compilation is 1) don't package any constant, or 2) package in binary_blob or pickle format (though the pickle format cannot be read in c++ now)

PR #164527, opened 2025-10-03
Add waitcounters for torch.compile subprocess pool

c00w:
Summary:
This ads waitcounter for whether or not the pool is running, as well as if we
are running jobs.

This also ads waitcounters for the first job within a pool. First job and running are working correctly. The job waitcounter seems to either be detecting a leak of a job, or is broken subtly.

Test Plan:
We've tested this internally and see valid ods metrics.

Note that we may be leaking jobs, or the job logic may not be handling an exception correctly.

Differential Revision: D83705931

### DIFF
--- a/torch/_inductor/compile_worker/subproc_pool.py
+++ b/torch/_inductor/compile_worker/subproc_pool.py
@@ -29,6 +29,7 @@ from torch._inductor.compile_worker.tracked_process_pool import (
 from torch._inductor.compile_worker.utils import _async_compile_initializer
 from torch._inductor.utils import get_ld_library_path, python_subprocess_env
 from torch._utils_internal import find_compile_subproc_binary
+from torch.monitor import _WaitCounter, _WaitCounterTracker
 
 
 log = logging.getLogger(__name__)
@@ -193,9 +194,26 @@ class SubprocPool:
 
         self.futures_lock = threading.Lock()
         self.pending_futures: dict[int, Future[Any]] = {}
+        # The pending waitcounter, is used to indicate the time when we have any specific job running.
+        self.pending_waitcounters: dict[int, Any] = {}
         self.job_id_count = itertools.count()
 
+        # The running waitcounter indicates the time when the SubProcPool object exists.
         self.running = True
+        self.running_waitcounter = _WaitCounter(
+            "pytorch.wait_counter.subproc_pool.running"
+        ).guard()
+        self.running_waitcounter.__enter__()
+
+        # The quiesce waitcounter indicates when the job is in a quiesced state.
+        self.quiesce_waitcounter: Optional[_WaitCounterTracker] = None
+
+        # Firstjob is used to capture the time from when the firstjob is queued, to when the first job is done.
+        self.firstjob = True
+        self.firstjob_id: Optional[int] = None
+        self.firstjob_waitcounter = _WaitCounter(
+            "pytorch.wait_counter.subproc_pool.first_job"
+        ).guard()
 
         # Start thread last to ensure all member variables are initialized
         # before any access.
@@ -212,6 +230,19 @@ class SubprocPool:
         with self.futures_lock:
             job_id = next(self.job_id_count)
             self.pending_futures[job_id] = future = Future()
+            self.pending_waitcounters[job_id] = _WaitCounter(
+                "pytorch.wait_counter.subproc_pool.job"
+            ).guard()
+            self.pending_waitcounters[job_id].__enter__()
+            if self.quiesce_waitcounter:
+                self.firstjob = True
+                self.quiesce_waitcounter.__exit__()
+                self.quiesce_waitcounter = None
+            # This can be entered from either quiesce wakeup, or from startup.
+            if self.firstjob:
+                self.firstjob_id = job_id
+                self.firstjob_waitcounter.__enter__()
+                self.firstjob = False
         future.set_running_or_notify_cancel()
         self._send(MsgHeader.JOB, job_id, job_data)
         return future
@@ -239,6 +270,7 @@ class SubprocPool:
                 if self.running:
                     log.warning("SubprocPool unclean exit")
                     self.running = False
+                    self.running_waitcounter.__exit__()
                 self.read_pipe.close()
                 # Cancel all the pending futures.
                 self.shutdown()
@@ -265,10 +297,21 @@ class SubprocPool:
                     self.pending_futures[job_id].set_exception(result)
                 else:
                     self.pending_futures[job_id].set_result(result)
+
+                self.pending_waitcounters[job_id].__exit__()
+                del self.pending_waitcounters[job_id]
+                if self.firstjob_id == job_id:
+                    self.firstjob_waitcounter.__exit__()
+
                 del self.pending_futures[job_id]
 
     def quiesce(self) -> None:
         self._send(MsgHeader.QUIESCE)
+        assert self.quiesce_waitcounter is None
+        self.quiesce_waitcounter = _WaitCounter(
+            "pytorch.wait_counter.subproc_pool.running"
+        ).guard()
+        self.quiesce_waitcounter.__enter__()
 
     def wakeup(self) -> None:
         self._send(MsgHeader.WAKEUP)
@@ -279,6 +322,7 @@ class SubprocPool:
                 if not self.running:
                     return
                 self.running = False
+                self.running_waitcounter.__exit__()
                 _send_msg(self.write_pipe, MsgHeader.SHUTDOWN)
                 self.write_pipe.close()
             self.process.wait(300)


torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -313,10 +338,13 @@ def __init__(
         self.running = True
 
     def main(self) -> None:
+        gc.disable()
+        gc.freeze()
         while True:
             msg_header, job_id, data = _recv_msg(self.read_pipe)
             if msg_header == MsgHeader.JOB:
                 self.submit(job_id, data)
+                gz.collect()
```
masnesral: This code looks very not tested ;)
What's going on with these gc-related changes? This looks like debugging code.

torch/_inductor/compile_worker/utils.py
```diff
@@ -27,7 +27,7 @@ def _async_compile_initializer(orig_ppid: int) -> None:
 
     def run() -> None:
         while True:
-            sleep(1)
+            sleep(60)
```
masnesral, line 30: This might bug OSS users. I guess I'm not sure. I think it means these subprocesses could stick around on the box for up to a minute after the main script exits? Could we consider something smaller? I agree 1s seems like overkill but 10s would already be 10x less frequent, for example ...

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -192,9 +194,13 @@ def __init__(
 
         self.futures_lock = threading.Lock()
         self.pending_futures: dict[int, Future[Any]] = {}
+        self.pending_waitcounters: dict[int, Any] = {}
         self.job_id_count = itertools.count()
 
         self.running = True
+        self.running_waitcounter =  _WaitCounter( "pytorch.wait_counter.subproc_pool.running").guard()
```
masnesral: What is "running" supposed to capture? Because I don't think the ProcessPoolExecutor inside the subproc is running until a wakeup call or the first just is submitted.

But as written, it seems to be capturing the interval when the pool object is initialized to when the process exits. But that just means we compiled at least one Triton kernel.

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -192,9 +194,13 @@ def __init__(
...
+        self.running_waitcounter.__enter__()
+        self.firstjob = True
```
masnesral, line 212: Can you also initialize the other "firstjob" related members here? Ideally a typechecker would complain that you could be accessing firstjob_id before it's initialized, but I guess Python.

Maybe you can also combine firstjob and firstjob_id into one variable. Like if firstjob_id is None, that means first job.

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -192,9 +194,13 @@ def __init__(
 
         self.futures_lock = threading.Lock()
         self.pending_futures: dict[int, Future[Any]] = {}
+        self.pending_waitcounters: dict[int, Any] = {}
```
masnesral, line 198: What is pending_waitcounters supposed to capture? The interval when the pool is doing work? If so I wonder if we could simplify and just have one waitcounter that entered/exited when there are no jobs (i.e., when len(pending_futures) == 0). Does that do the right thing?

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -263,20 +277,31 @@ def _read_thread(self) -> None:
...
+                self.pending_waitcounters[job_id].__exit__()
+                del self.pending_waitcounters[job_id]
+                if self.firstjob_id == job_id:
+                    self.firstjob_waitcounter.__exit__()
+
                 del self.pending_futures[job_id]
 
     def quiesce(self) -> None:
         self._send(MsgHeader.QUIESCE)
+        self.quiesce_waitcounter =  _WaitCounter( "pytorch.wait_counter.subproc_pool.running").guard()
```
masnesral: So you want to capture the interval the pool is shutdown. This probably needs to be initialized earlier too. Wakeup would be called before quiesce.

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -263,20 +277,31 @@ def _read_thread(self) -> None:
...
+        self.quiesce_waitcounter.__enter__()
 
     def wakeup(self) -> None:
         self._send(MsgHeader.WAKEUP)
+        self.quiesce_waitcounter.__exit__()
+        self.firstjob = True
```
masnesral: Unfortunately, this placement for firstjob=True isn't robust. I don't know if there are any codepaths that neglect to call wakeup, but it's is optional. If you submit to the pool before a wakeup, the pool will wake itself up. Probably set it to True in quiesce? If there's a race, I'm not sure that even works correctly though :/

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -237,6 +250,7 @@ def _read_thread(self) -> None:
                 if self.running:
                     log.warning("SubprocPool unclean exit")
                     self.running = False
+                    self.running_waitcounter.__exit__()
```
masnesral, line 273: Probably this can rely on the one in shutdown? See comment there.

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -263,20 +277,31 @@ def _read_thread(self) -> None:
...
+        self.quiesce_waitcounter.__exit__()
+        self.firstjob = True
 
     def shutdown(self) -> None:
         try:
             with self.write_lock:
                 if not self.running:
                     return
                 self.running = False
+                self.running_waitcounter.__exit__()
```
masnesral, line 325: I guess I don't know what this WC is capturing. Is it safe to exit waitcounters multiple times? If so, maybe you can move this above the check for self.running and call it unconditionally.

masnesral (changes requested):

torch/_inductor/compile_worker/utils.py
c00w: I think I forgot to hit the export button, and this had V1 code (which was a mess), fixed this file is no longer present.

torch/_inductor/compile_worker/subproc_pool.py
c00w: I think this was V1, which got exported with a lot more random stuff in it. This isn't present in v2, which was the verison which got tested (sorry for the confusion).

torch/_inductor/compile_worker/subproc_pool.py
c00w: Happy to rename this - Maybe created ? It's intended to be all time while the parent object exists and it's own internal running variable isn't set to false (i.e. we're not exiting).

torch/_inductor/compile_worker/subproc_pool.py
c00w: Done.

c00w: re: firstjob_id we actually have two states. The pool has started but not gotten any work (either from quiesce or from initial object creation). Then we arm it, and the next job we see, we record as the firstjob. Then when that job finishes, we mark it as a finished job.

torch/_inductor/compile_worker/subproc_pool.py
c00w: Yes - it's trying to capture the time when we have any work running. Note that waitcounters are "reentrant" so you'll get meaningful data out of them if we have a waitcouner per job (i.e. duration_avg is how long each job took, and call_count is the number of jobs we are processing per minute).

No strong objection to a inverted counter, but it's actually losing information we get with the pending waitcounters.

torch/_inductor/compile_worker/subproc_pool.py
c00w: Yeah I fixed this by adding slightly more state information here. Let me play around with waitcounters, to see if I can simplify it to a single waitcounter. (I.e. is it possible to check / okay to call __exit__ without calling enter).

torch/_inductor/compile_worker/subproc_pool.py
c00w: So we're tieing this waitcounter strictly to the running variable - In particular, running is only set to false in two spots, this code XOR the code below. Thus we're exiting in both spots.

torch/_inductor/compile_worker/subproc_pool.py
c00w: We're protecting this witht he self.running check above, we should never call __exit__ on this waitcounter multiple times.

torch/_inductor/compile_worker/subproc_pool.py
c00w: Ah didn't realize that - side question, why do we have the wakeup command then? Either way made this robust by removing it from the wakeup handling.

c00w:
Still working on fixing the typing issue, but should be reviewable.

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -193,9 +194,26 @@ def __init__(
...
 
+        # The running waitcounter indicates the time when the SubProcPool object exists.
         self.running = True
+        self.running_waitcounter = _WaitCounter(
+            "pytorch.wait_counter.subproc_pool.running"
+        ).guard()
+        self.running_waitcounter.__enter__()
+
+        # The quiesce waitcounter indicates when the job is in a quiesced state.
+        self.quiesce_waitcounter: Optional[_WaitCounterTracker] = None
```
masnesral, line 209: I guess this could be slightly confusing; maybe it's fine. The pool actually starts in a quiesced state in the sense that the subprocess's internal pool is not initialized yet (i.e., the 32 subprocs wouldn't get created yet). The internal pool gets created on a wakeup() -or- when the first job is submitted.

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -212,6 +230,19 @@ def submit(
         with self.futures_lock:
             job_id = next(self.job_id_count)
             self.pending_futures[job_id] = future = Future()
+            self.pending_waitcounters[job_id] = _WaitCounter(
+                "pytorch.wait_counter.subproc_pool.job"
+            ).guard()
+            self.pending_waitcounters[job_id].__enter__()
+            if self.quiesce_waitcounter:
```
masnesral, line 237: Tied to my comment above; we'd get out of the quiesce state on a wakeup too in the sense that the subprocs would get recreated. So depending on how you're thinking about what it means to be quiesced this might not be quite what you expect.

masnesral (approved):

torch/_inductor/compile_worker/subproc_pool.py
c00w: Yeah I think that's fine.

torch/_inductor/compile_worker/subproc_pool.py
c00w: Yeah ... Let me stack a PR to also log from wakeup this (since technically, you're not required to submit before wakeup, and vice versa).

PR #164531, opened 2025-10-03
Remove CUDA 11 branches for sparse code

cyyever:
This PR removes outdated CUDA version checks from sparse code in aten/src/ATen/cuda.


aten/src/ATen/cuda/CUDASparseDescriptors.cpp
```diff
@@ -12,8 +12,6 @@ cusparseStatus_t destroyConstDnMat(const cusparseDnMatDescr* dnMatDescr) {
   return cusparseDestroyDnMat(const_cast<cusparseDnMatDescr*>(dnMatDescr));
 }
 
-#if AT_USE_CUSPARSE_GENERIC_API() || AT_USE_HIPSPARSE_GENERIC_API()
```
eqy, line 15: might as well just remove these definitions then as they look even older?

cyyever: I just encountered HIP errors in removing them. So I would like to fix the current build errors before considering that.

cyyever: Removed

eqy (approved):
lgtm, may need eyes from hip/rocm side as I don't know details there

cyyever:
ROCm is not touched much, let's try it.

Issue #120105, opened 2024-02-16
Weird Artifact name: 'not_implemented' not registered being raised


thiagocrepaldi:
### 🐛 Describe the bug

@atalman is aware of this issue that only repros on Meta's internal tests
This currently blocks https://github.com/pytorch/pytorch/pull/108186, so a workaround is being created and this issue should track it

References:
* https://github.com/pytorch/pytorch/pull/108186#issuecomment-1929816055
* https://github.com/pytorch/pytorch/pull/108186#issuecomment-1930352099

### Versions

main

PR #164537, opened 2025-10-03
LocalTensor

dzmitry-huba:
A LocalTensor is a tensor subclass which simulates a tensor that is
distributed across SPMD ranks.  A LocalTensor might be size N, but in fact
there are world_size shards/replicas of it stored internally.  When you do a
plain PyTorch operation on it, we apply the operation to each shard; when you
do a collective, we do the mathematically equivalent operation on the local
shards.  A LocalTensor is associated with a list of ranks which specify
which ranks it holds local tensors for.

NB, this is NOT a DataParallel like abstraction where you can run operations
on multiple different GPUs. It is intended purely for *debugging* purposes,
the overhead is almost certainly too high to keep eight GPUs (even the C++
autograd needs multithreading to keep up!)  (It might potentially be possible
to trace through this with torch.compile and then compile it with CUDA graphs
but this is currently a non-goal.)

In order to handle MPMD, we provide a helper decorator that allows you to
run a function with no side effects for each LocalTensor shard and combine
results back into LocalTensor or LocalIntNode.

Note: This PR convert all DTensor ops and some DTensor tests to illustrate
intended usage and ensure conrrectness. In subsequent PR more tests will be
converted. DUring test conversion we aim to share as much as possible of
test logic between multi-process / multi-threaded and local tensor tests.
We would like to developers to be able to run both flavors of the tests.

Note: This work is based on the original proposal
by @ezyang (WIP PR https://github.com/pytorch/pytorch/pull/162753).
cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @ezyang @EikanWang @jgong5 @wenzhe-nrv @albanD


albanD (commented):
Skipping pr sanity check sounds good since we have extensive review by core maintainers on this one.

torch/fx/experimental/_constant_symnode.py
```diff
@@ -41,6 +41,9 @@ def __repr__(self) -> str:
     def _graph_repr(self) -> str:
         return self._str()
 
+    def add(self, other: Any) -> Any:
+        return other.add(self)
```
ezyang, line 45: Looks like you bypassed all the nested tensor failures successfully but we could also have this be on its own if needed

torch/distributed/_local_tensor/__init__.py
```diff
@@ -0,0 +1,652 @@
+from ast import Call
+
+
+"""
+A LocalTensor is a tensor subclass which simulates a tensor that is
+distributed across SPMD ranks.  A LocalTensor might be size N, but in fact
+there are world_size shards/replicas of it stored internally.  When you do a
+plain PyTorch operation on it, we apply the operation to each shard; when you
+do a collective, we the mathematically equivalent operation on the local
```
albanD:
```suggestion
do a collective, we do the mathematically equivalent operation on the local
```

torch/distributed/_local_tensor/__init__.py
```diff
@@ -0,0 +1,652 @@
...
+from torch.distributed import DeviceMesh
+from torch.distributed._functional_collectives import AsyncCollectiveTensor
+from torch.fx.experimental._constant_symnode import ConstantIntNode
+from torch.nested._internal.nested_int import NestedIntNode
+from torch.utils import _pytree as pytree
+from torch.utils._python_dispatch import return_and_correct_aliasing, TorchDispatchMode
+from torch.utils.checkpoint import get_device_states, set_device_states
+
+
+not_implemented_log = torch._logging.getArtifactLogger(__name__, "not_implemented")
```
albanD, line 67: Looks sus

torch/distributed/_local_tensor/__init__.py
```diff
@@ -0,0 +1,652 @@
...
+def _check_for_subclass(flat_args: Sequence[object]) -> bool:
+    return any(_check_for_subclass_arg(x) for x in flat_args)
+
+
+def _check_for_subclass_arg(x: object) -> bool:
+    return (
+        not isinstance(x, LocalTensor)
+        and isinstance(x, Tensor)
+        and type(x) is not Tensor
+        and type(x) is not torch.nn.Parameter
```
albanD:
```suggestion
        and type(x) not in (Tensor, torch.nn.Parameter, torch.nn.Buffer)
```

torch/distributed/_local_tensor/__init__.py
```diff
@@ -0,0 +1,652 @@
...
+    # Map from global rank to the local tensor
+    # This is a dict because in an MPMD situation, you will only have a subset
+    # of ranks that this tensor actually represents (the other ranks may be
+    # doing something different)
+    _local_tensors: dict[int, torch.Tensor]
+    _ranks: frozenset[int]  # precomputed for speed
+    __slots__ = ["_local_tensors"]
+
+    @staticmethod
+    @torch._disable_dynamo
```
albanD, line 288: What does this do?

torch/distributed/_local_tensor/__init__.py
```diff
@@ -0,0 +1,652 @@
...
+                    torch._C.DispatchKey.Negative
+                )
+            return extra_dispatch_keys
+
+        extra_dispatch_keys = get_extra_dispatch_keys(first_local_tensor)
+
+        for _rank, local_tensor in it:
+            assert dtype == local_tensor.dtype
+            assert layout == local_tensor.layout
+            assert extra_dispatch_keys == get_extra_dispatch_keys(local_tensor)
```
albanD: nit: proper error with messages

torch/distributed/_local_tensor/__init__.py
```diff
@@ -0,0 +1,652 @@
...
+
+        # This is horribly inefficient
+        flat_args, args_spec = pytree.tree_flatten((args, kwargs))
+        local_tensor = None
+        for arg in flat_args:
+            if isinstance(arg, LocalTensor):
+                local_tensor = arg
+                break
+
+        assert local_tensor is not None
```
albanD: Nice error message

torch/distributed/_local_tensor/__init__.py
```diff
@@ -0,0 +1,652 @@
...
+
+    def tolist(self) -> list[Any]:
+        return self.reconcile().tolist()
+
+    def reconcile(self) -> torch.Tensor:
+        # Force all tensors across ranks to be the same
+        it = iter(self._local_tensors.values())
+        t1 = next(it)
+        for t2 in it:
+            assert torch.equal(t1, t2)
```
albanD: real error message

albanD (commented):
Only nits, sounds ok to me.

torch/distributed/_local_tensor/__init__.py
dzmitry-huba: Thank you for reviewing the PR, agree on proper error messages in asserts, I'm still readying up the PR for review, will be ready soon

dzmitry-huba:
Rebasing

torch/distributed/_local_tensor/__init__.py
dzmitry-huba: This is related to https://github.com/pytorch/pytorch/issues/120105, several other Tensor subclasses do this work around.

test/distributed/tensor/test_dtensor_ops.py
```diff
@@ -81,6 +83,17 @@ def wrapped(fn):
     return wrapped
 
 
+def repurposeOps(op_db, base_test_name, derived_test_name):
```
ezyang: a little docblock here would be much appreciated!

test/distributed/tensor/test_dtensor_ops.py
```diff
@@ -683,17 +700,86 @@ def test_mean(self):
...
+    )
+    def test_dtensor_op_db(self, dtype, op):
+        self.run_opinfo_test(dtype, op)
+
+    def test_mean(self):
+        with LocalTensorMode(frozenset(range(0, self.world_size))):
+            self.run_mean()
+
+    def test_one_hot(self):
+        self.run_one_hot()
```
ezyang, line 809: This strategy is good for porting, but in the long term it is annoying as when you add a new test you do not automatically get the local test, and you have to remember to keep adding them here.

dzmitry-huba: Good point. I will work separately on automating this part.

torch/fx/experimental/_constant_symnode.py
dzmitry-huba: Can you please clarify here. While I resolved the issues with nested tensor this is indeed still needed, so it is kept in the implementation. Do you have something else in mind?

torch/csrc/distributed/c10d/Ops.cpp
```diff
@@ -15,7 +15,11 @@ TORCH_LIBRARY(c10d, m) {
   m.class_<Work>("Work")
       .def(torch::init<>())
       .def("wait", [](const c10::intrusive_ptr<Work>& self) { self->wait(); });
-  m.class_<ReduceOp>("ReduceOp").def(torch::init<>());
+  m.class_<ReduceOp>("ReduceOp")
+      .def(torch::init<>())
+      .def("op", [](const c10::intrusive_ptr<ReduceOp>& self) -> int64_t {
+        return self->op_;
+      });
```
ezyang, line 22: Someone asked me on Slack about this binding LOL

torch/distributed/_local_tensor/__init__.py
```diff
@@ -0,0 +1,749 @@
...
+NB, this is NOT a DataParallel like abstraction where you can run operations
+on multiple different GPUs. It is intended purely for *debugging* purposes,
+the overhead is almost certainly too high to keep eight GPUs (even the C++
+autograd needs multithreading to keep up!)  (It might potentially be possible
+to trace through this with torch.compile and then compile it with CUDA graphs
+but this is currently a non-goal.)
+
+In order to handle MPMD, we provide a helper decorator that allows you to
+run a function with no side effects for each LocalTensor shard and combine
+results back into LocalTensor or LocalIntNode.
```
ezyang: This isn't full MPMD though, it's a sort of limited divergence but all the ranks get back into synchrony by the time you get to LocalTensor/LocalIntNode. MPMD here would evoke PP which is not right.

torch/distributed/_local_tensor/__init__.py
```diff
@@ -0,0 +1,749 @@
...
+We need to be able to reverse engineer all of the collectives that don't
+involve the current local rank here to actually issue them.  This can be done
+two ways: (1) looking at the participating local ranks in the PG and computing
+the complement which specifies all the other collectives you have to run, or
+(2) retrieving the device mesh axis corresponding to the PG for this rank, and
+then running all the fibers for this.
+
+Our plan is to assume that the original list of ranks is arange and reverse engineer
+the Layout corresponding to the participating ranks, with respect to the global
+world size.
```
ezyang: This is one of my old comments but it's a little out of date for here 🤣

torch/distributed/_local_tensor/__init__.py
```diff
@@ -0,0 +1,749 @@
...
+        This class is primarily intended for debugging and simulating distributed tensor computations
+        on a single process.
+
+    """
+
+    # Map from global rank to the local tensor.
+    _local_tensors: dict[int, torch.Tensor]
+    # Precomputed for speed set of keys from the local tensor map.
+    _ranks: frozenset[int]
+    __slots__ = ["_local_tensors"]
```
ezyang: oops, let's not forget to put `_ranks` here too

torch/distributed/_local_tensor/__init__.py
ezyang: it was cargo culted from DTensor. Not sure why it is disabled there though

torch/distributed/_local_tensor/__init__.py
```diff
@@ -0,0 +1,749 @@
...
+        Creates a LocalTensor instance by mapping rank id to ids local shard.
+        """
+
+        with self.disable():
+            return LocalTensor({r: cb(r) for r in self.ranks})
+
+    def _patch_device_mesh(self) -> None:
+        assert self._old_get_coordinate is None
+        self._old_get_coordinate = DeviceMesh.get_coordinate  # type: ignore[assignment]
+        DeviceMesh.get_coordinate = _LocalDeviceMesh.get_coordinate  # type: ignore[method-assign]
```
ezyang, line 636: Not using unittest.mock.patch in the end?

torch/distributed/_local_tensor/__init__.py
```diff
@@ -0,0 +1,749 @@
...
+        return out  # type: ignore[return-value]
+
+
+def reconcile_args(args: Any, kwargs: dict[str, Any] | None = None) -> Any:
+    """
+    Reconciles arguments by converting any LocalTensor instances in the input
+    arguments to their underlying torch.Tensor representation.
+
+    This function is typically used to prepare arguments for functions that
+    expect standard torch.Tensor objects, by flattening the input arguments,
```
ezyang, line 675: it's actually not obvious to me what functions are going to take Tensor but not LocalTensor

torch/distributed/_local_tensor/_c10d.py
```diff
@@ -0,0 +1,710 @@
...
+    if iteration >= max_iterations:
+        raise RuntimeError(
+            f"Algorithm did not converge after {max_iterations} iterations"
+        )
+
+    # Convert to proper format for Layout
+    if len(shapes) == 1:
+        return Layout(shapes[0], strides[0])
+    else:
+        return Layout(tuple(shapes), tuple(strides))
```
ezyang: BTW, this code is LLM generated, it could probably use a touch up

torch/distributed/_local_tensor/_c10d.py
```diff
@@ -0,0 +1,710 @@
...
+
+
+# NB: lifted from https://github.com/pytorch/pytorch/pull/161016
+def _layout_to_indices(layout: Layout) -> list[int]:
+    return [
+        sum(c * s for c, s in zip(coord, flatten(layout.stride)))
+        for coord in product(*(range(s) for s in flatten(layout.shape)))
+    ]
```
ezyang: BTW, this PR was merged so it probably should be possible to use the actual code there

torch/distributed/tensor/_dispatch.py
```diff
@@ -276,14 +275,16 @@ def default_tensor(spec: DTensorSpec) -> torch.Tensor:
...
+                # The output of the equal op is a bool, by converting it into a
+                # a single value tensor, we can use all-reduce with min reduce op
+                # to simulate logical and.
+                assert local_results is None or isinstance(local_results, bool)
+                r = torch.tensor(
+                    int(local_results) if local_results is not None else 1,
+                    device=mesh.device_type,
+                )
+                dist.all_reduce(r, op=dist.ReduceOp.MIN)
+                local_results = bool(r.item())
```
ezyang, line 289: BTW this is a substantive change outside of LocalTensor, so it potentially could be landed on its own for more safety.

torch/distributed/tensor/_utils.py
```diff
@@ -170,6 +170,11 @@ def _compute_local_shape_and_global_offset(
 
             local_shape[shard_dim] = shard_size
 
+            zero_global_offset = global_shape[shard_dim]
+            if shard_offset is not None and isinstance(shard_offset, torch.SymInt):
+                zero_global_offset = torch.SymInt(zero_global_offset)
+
+            # TODO: HANDLE THIS
```
ezyang: The TODO is done now I guess?

torch/distributed/tensor/_utils.py
```diff
@@ -170,6 +170,11 @@ def _compute_local_shape_and_global_offset(
 
             local_shape[shard_dim] = shard_size
 
+            zero_global_offset = global_shape[shard_dim]
+            if shard_offset is not None and isinstance(shard_offset, torch.SymInt):
+                zero_global_offset = torch.SymInt(zero_global_offset)
```
ezyang, line 179: This is kind of weird, I wasn't expecting this

torch/distributed/tensor/placement_types.py
```diff
@@ -231,9 +249,47 @@ def _reduce_shard_tensor(
...
+    def _maybe_pad_tensor(
+        self,
+        local_tensor: torch.Tensor,
+        logical_dim_size: int,
+        num_chunks: int,
+    ) -> torch.Tensor:
+        is_padded = logical_dim_size % num_chunks != 0
+
+        if is_padded:
+            full_chunk_size = (logical_dim_size + num_chunks - 1) // num_chunks
```
ezyang, line 298: The duping here is kind of annoying but I guess it's not easily avoidable

torch/distributed/tensor/placement_types.py
```diff
@@ -231,9 +249,47 @@ def _reduce_shard_tensor(
         )
 
         if is_padded:
-            output = unpad_tensor(output, self.dim, pad_sizes[my_coordinate[mesh_dim]])  # type: ignore[possibly-undefined]
+            assert pad_sizes is not None
+            output = self._maybe_unpad_tensor_with_sizes(
+                output, pad_sizes, my_coordinate[mesh_dim]
+            )
```
ezyang, line 285: This site never called contiguous, intentional behavior change?

torch/distributed/tensor/placement_types.py
```diff
@@ -180,24 +195,26 @@ def _shard_tensor(
...
+            return Shard._select_shard(scatter_list, mesh_dim_local_rank)
 
         scatter_list, pad_sizes = self._split_tensor(
             tensor, num_chunks, with_padding=True, contiguous=True
         )
-        output = torch.empty_like(scatter_list[mesh_dim_local_rank])
+
+        it = iter(scatter_list)
+        first = next(it)
+        assert all(first.shape == v.shape for v in it)
```
ezyang, line 227: Substantive change, it's not obvious to me that this assert passes, if you audited split_tensor I guess it's cool, also something else safer to deploy on its own if it causes problems.

ezyang (approved):
Nice work, thank you! Please take care for the changes that change non-LocalTensor related behavior.

torch/distributed/_local_tensor/__init__.py
dzmitry-huba: Yes, decided that making patching inside of the LocalTensor mode makes it less burdensome on the developers. Another reason, is it to support temporary disabling LocalTensorMode - in which case DeviceMesh is expected to behave as usual. With mock.path it might be harder to do.

torch/distributed/_local_tensor/__init__.py
dzmitry-huba: This is primarily when needed to compare a single node op results with DTensor op result. One example is torch._C._nn.cross_entropy_loss op that calls into numel(). If the input provided into the op is LocalTensor directly then it fails, while attempting to call numel(). Therefore we reconcile LocalTensor before calling op to get single node result.

torch/distributed/tensor/_utils.py
dzmitry-huba: Yes, removing the comment.

torch/distributed/tensor/_utils.py
dzmitry-huba: Otherwise it is possible to pass incompatible arguments to the sym_ite operator, resulting in an error

torch/distributed/tensor/placement_types.py
dzmitry-huba: Good catch, this wasn't intentional. Updated call sites to match original behavior.

torch/distributed/_local_tensor/__init__.py
ezyang: OK. Sometimes it is necessary to do a call with sym_numel (or use one of our Python decompositions). Good to see you were able to get past it with reconcilation.

torch/distributed/tensor/placement_types.py
dzmitry-huba: In this particular case it is safe because when tensor is split shards are padded (see with_padding=True passed to the _split_tensor).

torch/distributed/_local_tensor/_c10d.py
dzmitry-huba: Updated in favor of using _MeshLayout

torch/distributed/tensor/_dispatch.py
dzmitry-huba: Moving this to a separate PR https://github.com/pytorch/pytorch/pull/164999

torch/distributed/_local_tensor/_c10d.py
dzmitry-huba: Updated to hopefully easier to understand implementation

skpark-rh:
@dzmitry-huba
Hi, I was having a conversation with @ezyang about LocalTensors at the PTC and I was referred to this PR.  I was wondering about specific areas and tests that I could contribute in this feature. If you have any ideas let me know. Thanks!

PR #164548, opened 2025-10-03
[cond] support dtensor x torch.cond

ydwu4:
* #164874
* #164867
* #164866
* #164865
* __->__ #164548
* #164778


torch/_higher_order_ops/cond.py
```diff
@@ -471,6 +471,15 @@ def _bound(s0, lower_bound: bool):
         mode.shape_env.constrain_symbol_range(merged_out.node.expr, *min_max(a, b))
         return merged_out
 
+    # FakeTensorMode dispatch is before tensor subclass dispatch
```
ydwu4: cc @bdhirsh this is the special casing I mentioned to you. Need to change the comment

torch/_higher_order_ops/cond.py
```diff
@@ -392,28 +392,36 @@ def inner(mode, pred, true_fn, false_fn, operands):
...
+        assert fake_mode is t.fake_mode, (
+            f"Got mixed fake modes {fake_mode}, {t.fake_mode}"
+        )
+
+    pytree.tree_map_only(FakeTensor, _find, inputs)
+    assert fake_mode is not None, "Cannot get fake mode from inputs."
+    return fake_mode
+
+
+@register_fake(cond_op)
```
ydwu4, line 379: Directly using @cond_op.py_impl(FakeTensorMode) causes us to see DTensor inputs.  register_fake redirects a fake dtensor input to fake_mode.__torch_dispatch__, which first re-dispatches to dtensor then fake, in that case fake tensor impl of cond will only see FakeTensor inputs.

torch/distributed/tensor/_ops/_higher_order_ops.py
```diff
@@ -0,0 +1,150 @@
...
+    true_output_spec = analyze_branch_output(true_fn, operands)
+    false_output_spec = analyze_branch_output(false_fn, operands)
+
+    merged_spec = merge_dtensor_specs(true_output_spec, false_output_spec)
+
+    local_pred, local_operands = pytree.tree_map_only(
+        DTensor, lambda x: x._local_tensor, (pred, operands)
+    )
+
+    local_result = cond_op(local_pred, true_fn, false_fn, local_operands)
```
zou3519, line 148: What happens in the follow situation?

Let's say the user has a custom.sum() operation. The semantics of custom.sum(x) is that the input is sharded, but the output is replicated.

Now, let's say the user does torch.cond, where both branches do custom.sum().

I think this would pass all of the checks (all of the output placements are the same). It's not correct to do custom.sum() on the local tensors and then add the replicate placement to them: this leads to each rank having a different local value for replicate.

Do we error or handling this situation somehow?

zou3519: from HOP sync: Instead we should pass in true_fn/false_fn that accept local tensors. the true_fn needs to reconstruct a DTensor with correct placements, call true_fn, and then unwrap back into local tensors. Ditto for false_fn.

also, maybe we can write a generic cond rule for "PT2-compatible tensor subclasses" that looks like the above.

```py
input = ... # DTensor
input_spec = ...
output_spec = true_fn(input).spec

def new_true_fn(x):
    x = to_dtensor(x, input_spec)
    y = true_fn(x)
    return y._local_tensor
    
local_out = cond(pred, new_true_fn, new_false_fn, input._local_tensor)
return to_dtensor(local_out, output_spec)
```

torch/distributed/tensor/_ops/_higher_order_ops.py
```diff
@@ -0,0 +1,150 @@
...
+
```
zou3519, line 149: From hop sync: Also, how does this work? Does speculate_subgraph trace a graph with DTensors or regular Tensors? Simon says that local_map ran into situations where speculate_subgraph doesn't work with tensor subclass inputs.

PR #164560, opened 2025-10-03
[BC breaking] Remove assert_allclose

cyyever:
`torch.testing.assert_allclose` was deprecated in favour of `torch.testing.assert_close`. This PR thus replaces usage of `torch.testing.assert_close` with `torch.testing.assert_allclose` and removes `torch.testing.assert_allclose`. Thus the PR breaks backward compatibility. 

Because `torch.testing.assert_allclose` is actually a wrapper of `torch.testing.assert_close`, the `kwargs` of the two functions overlaps largely as well as some of their default values, so it is relatively easy to move to the new way. One notable default value change is that `torch.testing.assert_allclose` uses `check_dtype=False` while `dtype` check is enabled by default in `torch.testing.assert_close`. Another change is that `eq_nan=True` in `torch.testing.assert_allclose` and it is `False` in `torch.testing.assert_close`

For example, before the PR we have
```
torch.testing.assert_allclose(a, b)
```

After the PR it should be changed to 
```
torch.testing.assert_close(a, b, check_dtype=False)
```
or simply to
```
torch.testing.assert_close(a, b)
```
to also check `dtype`.

cc @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @aditvenk @weifengpy @kapilsh @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @azahed98 @H-Huang @Lucaskabela @chenyang78

### DIFF
--- a/docs/source/testing.md
+++ b/docs/source/testing.md
@@ -15,7 +15,3 @@
 ```{eval-rst}
 .. autofunction:: make_tensor
 ```
-
-```{eval-rst}
-.. autofunction:: assert_allclose
-```
--- a/test/allowlist_for_publicAPI.json
+++ b/test/allowlist_for_publicAPI.json
@@ -995,7 +995,6 @@
     "all_types_and_complex",
     "all_types_and_complex_and",
     "all_types_and_half",
-    "assert_allclose",
     "assert_close",
     "complex_types",
     "double_types",
--- a/test/allowlist_for_publicAPI.json
+++ b/test/allowlist_for_publicAPI.json
@@ -996,7 +996,6 @@
     "all_types_and_complex",
     "all_types_and_complex_and",
     "all_types_and_half",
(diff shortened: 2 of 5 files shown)


janeyx99 (approved):
Please write a BC breaking note in the PR body similar to https://docs.google.com/document/d/14OmgGBr1w6gl1VO47GGGdwrIaUNr92DFhQbY_NEk8mQ/edit?tab=t.0#heading=h.a9htwgvvec1m stating that we're removing this function as it's been deprecated.

mlazos (approved):

PR #164564, opened 2025-10-03
[dynamo] Fix for einops 0.6.1 x torch.compile broken in pytorch nightlies

### LINKED ISSUE #157417: einops 0.6.1 x torch.compile broken in pytorch nightlies
```py
import torch
import torch.nn as nn
from einops import einsum, pack, rearrange, reduce, repeat, unpack

class TorchModuleWithOperations(nn.Module):
    def __init__(self) -> None:
        super().__init__()

    def forward(self, x_abc, suffix=""):
        a, b, c = x_abc.shape

        def suf(pattern):
            parts = pattern.split()
            return " ".join(
                [p if p[-1] not in "acd" else p + suffix for p in parts]
            )

        # patterns look a bit strange because names a, c, d will be modified on every run
        # by suf function
        x_abcd = repeat(x_abc, suf("a b c -> a b c 4"))
        x_abc = reduce(x_abcd, suf("a b c d -> a b c"), "min")
        x_abdc, ps = pack([x_abc] * (2 + len(suffix)), suf("a b * c"))
        x_array = unpack(
            rearrange(x_abdc, suf("a b d c -> (a b ) 1 c d")), ps, "ab one1 c *"
        )
        x1 = x_array[0] + len(x_array)
        x1 = rearrange(x1, suf("(a b ) 1 c -> a b c"), b=b)
        addition = einsum(x_abc, x_abcd, suf("a b c , a b c d -> d"))[0]
        return x1 + addition

original = TorchModuleWithOperations()
# Einops only interacts with Dynamo but we test backend="inductor" just in case
compiled = torch.compile(original, backend="eager", fullgraph=True)
for size in [10, 20, 40]:
    x = torch.rand([size, size + 1, size + 2])
    for suffix in ["", "suf1", "other_suffix"]:
        result1 = compiled(x, suffix)
        result2 = original(x.double(), suffix).float()
```
fails with 
```
TorchRuntimeError: Dynamo failed to run FX node with fake tensors: call_function <function repeat at 0x7f1e654a2ca0>(*(FakeTe
nsor(..., size=(s48, s86, s93)), 'a b c -> a b c 4'), **{}): got TypeError('unhashable type: non-nested SymInt')

from user code:
   File "<ipython-input-1-c92c344a3624>", line 20, in forward
    x_abcd = repeat(x_abc, suf("a b c -> a b c 4"))

Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For
even more developer context, set TORCH_LOGS="+dynamo"
```

cc @ezyang @gchanan @kadeng @msaroufim @chauhang @penguinwu @bobrenjc93 @voznesenskym @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @amjames @Lucaskabela @jataylo @chenyang78

jmaczan:
As mentioned in https://github.com/pytorch/pytorch/issues/157417#issuecomment-3348985391, it patches functions wrapped with lru_cache with unwrapped functions. The root cause is that lru_cache tries to get `__hash__` of SymInt, when it's not defined and throws TypeError, which is handled gracefully since einops 0.7.0rc2

Imo the ideal fix would be fix the SymInt itself and perhaps once I get better at PyTorch I'll do that. We rely on TypeError for einops >0.7.0 https://github.com/pytorch/pytorch/issues/157417#issuecomment-3348985391, but in fact if SymInt hash will work fine, the TypeError will never be thrown and the fallback that relies on TypeError won't be launched, so that seems to not break anything 

For now I think it's safe to just patch the problematic functions to unblock 0.6.1 einops x pytorch

Fixes #157417


jmaczan:
@zou3519 @ezyang @bdhirsh not sure who should I ping, you all were involved in the Issue so letting you know the fix is ready to review

test/dynamo/test_einops_patch.py
```diff
@@ -0,0 +1,255 @@
...
+                sys.modules.pop(module, None)
+
+        importlib.invalidate_caches()
+
+        super().tearDown()
+
+    def test_no_eager_import(self):
+        self.einops_patch_module._patch_einops()
+
+        assert "einops" not in sys.modules
```
williamwen42: Use `unittest` assert methods

.gitignore
```diff
@@ -395,3 +395,4 @@ android/pytorch_android_torchvision/.cxx
 CLAUDE.local.md
 /test_*.py
 /debug_*.py
+.venv/
```
williamwen42: Remove?

williamwen42 (commented):
lgtm aside from a few comments, deferring to @anijain2305 and @zou3519 since you have more context on the einops issue.

test/dynamo/test_einops_patch.py
jmaczan: Right, updated

.gitignore
jmaczan: Yes. these were added in a few PRs before your comment in another PR. Sorry! Removed

jmaczan (commented):
Adjusted code as per @williamwen42 comments - @anijain2305 @zou3519 @ezyang pls take a look and approve if ok

williamwen42 (dismissed):

zou3519 (changes requested):
I don't think we should be monkey-patching einops.

How are we getting a SymInt into a lru_cache?

jmaczan:
@zou3519 ping https://github.com/pytorch/pytorch/pull/164564#issuecomment-3417444306, I wonder what do you think about it. Is it wrong that we pass FakeTensor to einops, in the first place? Is it what we should focus on fixing, instead of monkey patch?

williamwen42:
We have an alternative fix for this

PR #164565, opened 2025-10-03
Move ScalarTypeToCPPType to headeronly and add tests for AT_FORALL_... macros

pearu:
* __->__ #164565
* #164405
* #164354
* #164350


torch/headeronly/core/ScalarType.h
```diff
@@ -25,18 +25,20 @@
...
+// and so in the end we ended up with some mish-mosh of some helper
+// macros, but mostly use sites making a call about what dtypes they
+// can or can't support.  So if you want to add a new dtype, the
+// preferred resolution is to find a dtype similar to what you want,
+// grep for it and edit all the sites you find this way.  If you need
+// to add a completely new kind of dtype, you're going to have to
+// laboriously audit all of the sites everywhere to figure out how it
+// should work.  Consulting some old PRs where we added new dtypes
+// (check history of this and c10/core/ScalarType.h files) can help
+// give you an idea where to start.
```
janeyx99, line 53: why was this change necessary--is this from a local autolinter?

test/cpp/aoti_abi_check/test_scalartype.cpp
janeyx99, line 1: This file should be added to the bottom PR that migrates all the macros. (So the testing and the migration are in the same PR). We should add a test in this PR for `ScalarTypeToCPPTypeT`.

We should add the macros to torch/header_only_apis.txt:
AT_FORALL_SCALAR_TYPES_WITH_COMPLEX_EXCEPT_COMPLEX_HALF_F8NZ
AT_FORALL_SCALAR_TYPES_WITH_COMPLEX
AT_FORALL_SCALAR_TYPES_WITH_COMPLEX_AND_QINTS
AT_FORALL_INT_TYPES
AT_FORALL_SCALAR_TYPES
AT_FORALL_SCALAR_TYPES_AND
AT_FORALL_SCALAR_TYPES_AND2
AT_FORALL_SCALAR_TYPES_AND3
AT_FORALL_SCALAR_TYPES_AND7

janeyx99: I mean in #164350 by the way. Let's be consistent with migration + tests.

test/cpp/aoti_abi_check/test_scalartype.cpp
```diff
@@ -2,6 +2,59 @@
 
 #include <torch/headeronly/core/ScalarType.h>
```
janeyx99, line 3: Also add a test in here for ScalarTypeToCPPTypeT + add ScalarTypeToCPPTypeT to torch/header_only_apis.txt

janeyx99: Oh, I see AT_FORALL_SCALAR_TYPES_WITH_COMPLEX_AND_QINTS tests ScalarTypeToCPPTypeT indirectly. In that case, you can just add a comment that it's indirectly tested already in header_only_apis.txt. I'll suggest a change.

torch/header_only_apis.txt
```diff
@@ -121,4 +121,13 @@ NumScalarTypes
 ScalarType
 toString
 <<
+AT_FORALL_SCALAR_TYPES_WITH_COMPLEX_EXCEPT_COMPLEX_HALF_F8NZ
```
janeyx99, line 125:
```suggestion
ScalarTypeToCPPTypeT  // Tested indirectly through AT_FORALL_SCALAR_TYPES_WITH_COMPLEX_AND_QINTS
AT_FORALL_SCALAR_TYPES_WITH_COMPLEX_EXCEPT_COMPLEX_HALF_F8NZ
```

pearu: Applied as a comment (lintrunner is not able to pick up indirectly tested features).

torch/headeronly/core/ScalarType.h
pearu: The actual change is from
```
(check history of this file)
```
to
```
(check history of this and c10/core/ScalarType.h files)
```
The rest is from local reformatting. I'll restore it...

test/cpp/aoti_abi_check/test_scalartype.cpp
pearu: The only issue with "the testing and the migration are in the same PR" is that toString is not yet available in #164350 that testing these macros use.

test/cpp/aoti_abi_check/test_scalartype.cpp
pearu: Thanks!

pearu:
https://github.com/pytorch/pytorch/pull/164565 has landed.

PR #164568, opened 2025-10-03
Add hop for additional control dependencies

eellison:
* #164738
* #164581
* #164569
* __->__ #164568

Adds [control_deps](https://en.wikipedia.org/wiki/Control_dependency) higher-order operator to enforce explicit scheduling dependencies in FX graphs. This prevents unwanted operation reordering/fusion by giving nodes additional dependencies, which we also respect in inductor by adding weakdeps on the additional dependencies.

This can be generally useful (such as for ordering collectives) but in this case I am using it so that fusions do not interfere with aten planned comm-compute overlap. 

There's definitely some similarity with the `with_effects` hop. Talked with @angelayi  - when @zou3519  is back we will figure out how we want to consolidate.

The implementation needs to be a subgraph (as opposed to `with_effects`) because inductor relies on `V.graph.current_node`. Changing the signature of the node with `with_effects`  breaks this, and additionally, also breaks striding constraints on the wrapped node - see this [TODO](https://github.com/pytorch/pytorch/blob/aed66248a01d309eb2ac1149b5f51310545b0783/torch/fx/experimental/proxy_tensor.py#L1246-L1249). By maintaining the node with its original calling structure in subgraph this all works.

Example transformation:

Before:
```
%add : [num_users=1] = call_function[target=torch.ops.aten.add.Tensor](args = (%arg0_1, 1), kwargs = {})
%mm : [num_users=1] = call_function[target=torch.ops.aten.mm.default](args = (%arg1_1, %arg1_1), kwargs = {})
%mul : [num_users=1] = call_function[target=torch.ops.aten.mul.Tensor](args = (%add, 2), kwargs = {})
```
After:
```
add: "f32[256, 256]" = torch.ops.aten.add.Tensor(arg0_1, 1)
mm: "f32[256, 256]" = torch.ops.higher_order.control_deps((add,), subgraph_mm, arg1_1, arg1_1)
mul: "f32[256, 256]" = torch.ops.higher_order.control_deps((mm,), subgraph_mul, add)
```

The mm operation now explicitly depends on add completing first, and mul depends on mm, with original operations preserved in subgraphs.


torch/_inductor/fx_passes/control_dependencies.py
```diff
@@ -0,0 +1,212 @@
...
+
+from typing import Any
+
+import torch.fx as fx
+from torch._higher_order_ops.utils import register_fake
+from torch._ops import HigherOrderOperator
+from torch.utils._ordered_set import OrderedSet
+
+
+class ControlDeps(HigherOrderOperator):
```
ezyang, line 19: Can this live with the other HOPs? Put it in torch/_higher_order_ops/

torch/_inductor/fx_passes/control_dependencies.py
```diff
@@ -0,0 +1,212 @@
...
+        Args:
+            additional_deps: Tuple of tensors that must be computed first
+            subgraph: GraphModule containing the exact operation to execute
+            *args: Arguments to pass to the subgraph
+        """
+        if not isinstance(additional_deps, (tuple, list)):
+            raise TypeError(
+                f"additional_deps must be tuple/list, got {type(additional_deps).__name__}"
+            )
+        breakpoint()
```
ezyang: brakkponnnt

eellison: oop

torch/_inductor/fx_passes/control_dependencies.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+control_deps = ControlDeps()
+
+
+# Register fake implementation for tracing
+@register_fake(control_deps)
+def _(additional_deps, subgraph, *args, **kwargs):
+    """Fake tensor implementation - execute the subgraph."""
+    return subgraph(*args, **kwargs)
```
ezyang, line 62: Do we have to restore the fake mode before redispatching? You will need to check what happens with a factory function with control dep.

torch/_inductor/fx_passes/control_dependencies.py
eellison: can we defer on this when we actually have discussions on consolidation and such with @zou3519

torch/_inductor/scheduler.py
```diff
@@ -2681,6 +2681,10 @@ def add_user(
                             )
                             add_user(other_name, node, is_weak=True)
 
+                for add_dep in V.graph.additional_buffer_deps[buf.get_name()]:
+                    add_user(add_dep, node, is_weak=True)
+                    node.add_fake_dep(WeakDep(add_dep, node.get_name()))
```
ezyang, line 2686: You turn these into weak deps. That means they don't impede DCE. Is it important that FX level DCE is also able to DCE past control deps?

torch/_inductor/lowering.py
```diff
@@ -7226,6 +7226,59 @@ def invoke_subgraph(subgraph_fn: ir.Subgraph, identifier: str, *operands):
...
+
+    The control_deps_op HOP makes dependencies explicit in the graph. During lowering:
+    1. Realize all additional dependencies to ensure they're computed
+    2. Execute the target operation normally
+    3. Track the dependencies for the scheduler
+    """
+    # Realize all additional dependencies
+    dep_names = []
+    for dep in additional_deps:
+        if hasattr(dep, "realize"):
```
ezyang: nothing to fix, but it annoys me to no end that this hasattr test is idiomatic in inductor for this specifically. why can't we consistently have realize defined on all deps lol

torch/_inductor/lowering.py
```diff
@@ -7226,6 +7226,59 @@ def invoke_subgraph(subgraph_fn: ir.Subgraph, identifier: str, *operands):
...
+            dep_names.append(dep.get_name())
+
+    original_args = V.graph.current_node.args
+    arg_offset = 2  # first two args (additional_deps, subgraph)
+    assert len(args) + arg_offset == len(original_args)
+
+    output = None
+
+    assert len(subgraph_fn.graph_module.graph.find_nodes(op="placeholder")) == len(args)
+    for i, node in enumerate(subgraph_fn.graph_module.graph.nodes):
```
ezyang, line 7259: Are the node names in the subgraph guaranteed not to conflict with the parent graph's nodes? I don't know anywhere where we enforce this.

torch/_inductor/scheduler.py
eellison: Yes, that is intentional. they are just additional dependencies.

torch/_inductor/lowering.py
eellison: there might be a better way of doing.. let me clean up

torch/_inductor/lowering.py
```diff
@@ -7226,6 +7226,59 @@ def invoke_subgraph(subgraph_fn: ir.Subgraph, identifier: str, *operands):
...
+            V.graph.env[node] = V.graph.run_node(node)
+
+    assert output is not None and additional_deps
+    output_list = output if isinstance(output, (list, tuple)) else [output]
+
+    for out in output_list:
+        if not hasattr(out, "realize"):
+            continue
+
+        out.realize()
```
ezyang, line 7277: how come?

torch/_inductor/lowering.py
eellison: we're not looking at node names, we're looking at node objects themself. does not matter.

torch/_inductor/scheduler.py
eellison: for now - could consider a non-dce version with consolidation.

torch/_inductor/lowering.py
eellison: they need to be realized for us to add the dependency.

torch/_inductor/fx_passes/control_dependencies.py
```diff
@@ -0,0 +1,211 @@
...
+        """
+        if not isinstance(additional_deps, (tuple, list)):
+            raise TypeError(
+                f"additional_deps must be tuple/list, got {type(additional_deps).__name__}"
+            )
+        if not (isinstance(subgraph, fx.GraphModule) or callable(subgraph)):
+            raise TypeError(
+                f"subgraph must be GraphModule or callable, got {type(subgraph).__name__}"
+            )
+        return super().__call__(additional_deps, subgraph, *args, **kwargs)
```
ezyang, line 52: I have a dumb, noobie question about HOPs. How exactly does mutation inside the HOP work here? I know for many HOPs we have to go through great lengths to assert there's no mutation in the HOP. For this case where you are just wrapping a single op it seems feasible to handle a mutable op. But how exactly does this work?

eellison: This is only being inserted in a post grad pass. so it does not need to handle mutation and other things as `invoke_subgraph` does.

torch/_inductor/fx_passes/control_dependencies.py
```diff
@@ -0,0 +1,211 @@
...
+    """
+    if not additional_deps_map:
+        return
+
+    # Track replacements so we can update dependencies
+    replacements: dict[fx.Node, fx.Node] = {}
+
+    # Process each node that needs additional dependencies
+    for dependent_node, dep_nodes in additional_deps_map.items():
+        assert dependent_node.op == "call_function"
```
ezyang: nit: print the node op when assert fails. call_function only means you can't use this prior to AOTAutograd, it would seem.

torch/_inductor/fx_passes/control_dependencies.py
```diff
@@ -0,0 +1,211 @@
...
+        dependent_node.replace_all_uses_with(ordered_node)
+
+        # Remove the original node from the graph
+        graph.erase_node(dependent_node)
+
+        # Now rename the ordered node to the original name
+        ordered_node.name = original_name  # PRESERVE ORIGINAL NAME
+
+        # Track the replacement for future dependencies
+        replacements[dependent_node] = ordered_node
```
ezyang, line 162: What if I have a replacement from n1 -> n2, and then later I do n2 -> n3, how do I ensure n1 -> n3? Or is this impossible?

torch/_inductor/fx_passes/control_dependencies.py
```diff
@@ -0,0 +1,211 @@
...
+        original_kwargs = dependent_node.kwargs
+        original_meta = dependent_node.meta.copy()
+
+        updated_dep_nodes = [replacements.get(dep, dep) for dep in dep_nodes]
+
+        # Create a subgraph that preserves the exact original operation
+        subgraph_module = _create_subgraph_for_node(graph, dependent_node)
+
+        subgraph_attr_name = f"subgraph_{original_name}"
+        setattr(graph.owning_module, subgraph_attr_name, subgraph_module)
```
ezyang, line 119: can you assert the subgraph name doesn't already exist. But it would be better to have some way of just allocating a fresh name here

ezyang (approved):

torch/_inductor/fx_passes/control_dependencies.py
eellison: per - `invoke_subgraph` - no

torch/_inductor/fx_passes/control_dependencies.py
eellison: any node such only have a single set of additional deps, so it cant be replaced twice.

IvanKobzarev (approved):

eellison:
@IvanKobzarev since we're only adding these deps for nodes that are hidden, it shouldn't affect those fusions, which are targeting exposed collectives.

torch/_inductor/fx_passes/control_dependencies.py
```diff
@@ -0,0 +1,226 @@
...
+
+        original_name = dependent_node.name
+        original_args = dependent_node.args
+        original_kwargs = dependent_node.kwargs
+        original_meta = dependent_node.meta.copy()
+
+        updated_dep_nodes = [replacements.get(dep, dep) for dep in dep_nodes]
+
+        # Create a subgraph that preserves the exact original operation
+        subgraph_module = _create_subgraph_for_node(graph, dependent_node)
```
bdhirsh, line 114: on the topic of "better aot_eager perf": I imagine that we probably want one of these outcomes in the situation where we are running bucketing/reordering and only using the aot_eager backend:

(1) we have the reordering pass not bother inserting these HOPs/subgraph if we know we are not going to inductor

(2) or we insert the dependencies unconditionally, but if we do we may need to make sure that the aot_eager runtime code doesn't do anything too slow with them? no action needed, mostly thinking out loud (maybe calling a subgraph in the fx interpreter is fast enough already, or if not maybe we make the `control_deps` HOP have a faster eager impl)

eellison: 1 - we have a config for this, we'd set it false for aot_eager.

2 - if it turns out we do need to insert these for another reordering pass, we'd probably want to inline it at the end of invocation. (Similarly, we'd potentially want to inline invoke_subgraph)

torch/_inductor/fx_passes/control_dependencies.py
zou3519: This should live in torch/_higher_order_ops -- it's easier to keep track of HOPs that way.

zou3519:
Discussed a bit with @bdhirsh, there are some things that might be problems. NB: I did not read this PR carefully, it might already handle these cases.
1) the HOP should not return the inputs directly as outputs. Instead it should return aliases. This is because bad things happen when you make_fx any operation that returns the input directly as an output (we used to have some clobbering happen for inplace operations, I'm not sure how we avoided it).
2) there are passes in inductor that analyze views and aliases. Those need to be updated to understand that this HOP returns aliases. An example is the FakeTensorUpdator: FakeTensorUpdator doesn't support HOPs yet, but assuming it does, it needs to understand that this new HOP returns views. The view information is important to prevent silent incorrectness in reinplacing (this has bit us a couple of times).

eellison:
@zou3519 yes, correct (today at least)

PR #164570, opened 2025-10-03
Allow at::native::offset_t to be offset using `operator+=`

bernhardmgruber:
This will be required by CCCL 3.1.

### DIFF
--- a/aten/src/ATen/native/cuda/SortStable.cu
+++ b/aten/src/ATen/native/cuda/SortStable.cu
@@ -21,9 +21,15 @@ namespace {
 struct offset_t {
   int stride;
   int begin;
-  __device__ int operator[](int i) {
+  __device__ int operator[](int i) const {
     return stride * (begin + i);
   }
+#if CCCL_VERSION >= 3001000
+  __device__ offset_t& operator+=(int i) {
+    begin += i;
+    return *this;
+  }
+#endif
 };
 // Segmented sort by full sort algorithm:.
 // Say we are sorting a (2, 3) tensor. We have in flattened form:


aten/src/ATen/native/cuda/SortStable.cu
```diff
@@ -21,9 +21,13 @@ namespace {
 struct offset_t {
   int stride;
   int begin;
-  __device__ int operator[](int i) {
+  __device__ int operator[](int i) const {
```
bernhardmgruber, line 24: Drive-by fix. Iterator subscript operators should be `const`.

aten/src/ATen/native/cuda/SortStable.cu
```diff
@@ -21,9 +21,15 @@ namespace {
...
     return stride * (begin + i);
   }
+#if CCCL_VERSION >= 3001000
```
Skylion007, line 27: Any reason to not use it always be on for older versions of CCCL too? Dont see the harm

bernhardmgruber: I did and I got a warning that `operator+=` was unused. The preprocessor check suppresses that warning.

Skylion007: Ah because internal only linkages, guess it's fine in this case then since CCCL is the only thing that would call that method.

Skylion007 (approved):

aten/src/ATen/native/cuda/SortStable.cu
bernhardmgruber: I think so, yes!

eqy (approved):

aten/src/ATen/native/cuda/SortStable.cu
```diff
@@ -21,9 +21,15 @@ namespace {
...
+  __device__ offset_t& operator+=(int i) {
```
Artem-B, line 28: This may need to be `__host__ __device__`. In some builds, the operator may end up being called from a `CUB_RUNTIME_FUNCTION` which may be `__host__` for some compilations. E.g. from here: https://github.com/NVIDIA/cccl/blob/4872515b07a7e21c0bdc8b7968d217c34c37857a/cub/cub/device/dispatch/dispatch_radix_sort.cuh#L1425

At the very least, it fails for my build with CCCL 3.1.4 and clang.
```
third_party/py/torch/aten/src/ATen/native/cuda/SortStable.cu:28:24: note: candidate function not viable: call to __device__ function from __host__ function
   28 |   __device__ offset_t& operator+=(int i) {
      |                        ^
In file included from third_party/py/torch/aten/src/ATen/native/cuda/SortStable.cu:9:
In file included from third_party/py/torch/aten/src/ATen/cuda/cub.cuh:18:
In file included from blaze-out/k8-fastbuild-cuda/bin/third_party/gpus/cccl/_virtual_includes/_staged_headers/third_party/gpus/cccl/cub/cub.cuh:70:
In file included from blaze-out/k8-fastbuild-cuda/bin/third_party/gpus/cccl/_virtual_includes/_staged_headers/third_party/gpus/cccl/cub/device/device_radix_sort.cuh:46:
blaze-out/k8-fastbuild-cuda/bin/third_party/gpus/cccl/_virtual_includes/_staged_headers/third_party/gpus/cccl/cub/device/dispatch/dispatch_radix_sort.cuh:1425:32: error: no viable overloaded '+='
 1425 |         end_offsets_current_it += num_current_segments;
      |         ~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~
```

Making it `__host__ __device__` appears to be sufficient to let the build finish.

Artem-B: @bernhardmgruber : ^^^ FYI.

bernhardmgruber: > [@bernhardmgruber](https://github.com/bernhardmgruber) : ^^^ FYI.

Thx. Does that mean you (or somebody else) is going to fix this instead of me? :)

PR #144370, opened 2025-01-08
[RELAND] Generalize at::manual_seed for all accelerators

guangyey:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #144370

# Additional Context
This is a reland PR originated from eeb57394f93d720bca498c3fa9d167fc7b9cca46

cc @albanD @EikanWang

PR #164571, opened 2025-10-03
[MPS] Make deviceCount() implementation consistent with Python to fix at::manual_seed()

blawrence-ont:
In #144370 it was assumed that MPS would have at least 1 device, but that wasn't the case since MPSHooks wasn't overriding it.

This meant that `at::manual_seed()` wouldn't have any effect on MPS and using it, for example in tests, didn't make `at::rand()` deterministic.

Digging a bit deeper we see that `torch.mps.device_count()` returns 1 if MPS is available:

https://github.com/pytorch/pytorch/blob/aed66248a01d309eb2ac1149b5f51310545b0783/torch/mps/__init__.py#L27-L29

Mimicking that behaviour in C++ land fixes the determinism issue and makes the 2 APIs consistent.

### DIFF
--- a/aten/src/ATen/mps/MPSHooks.h
+++ b/aten/src/ATen/mps/MPSHooks.h
@@ -66,6 +66,9 @@ struct MPSHooks : public at::MPSHooksInterface {
     // When MPS is available, it is always in use for the one device.
     return true;
   }
+  DeviceIndex deviceCount() const override {
+    return static_cast<int>(hasMPS() && isAvailable());
+  }
 };
 
 } // namespace at::mps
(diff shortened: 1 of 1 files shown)


malfet (approved):

aten/src/ATen/mps/MPSHooks.h
```diff
@@ -66,6 +66,9 @@ struct MPSHooks : public at::MPSHooksInterface {
     // When MPS is available, it is always in use for the one device.
     return true;
   }
+  DeviceIndex deviceCount() const override {
+    return static_cast<int>(hasMPS() && isAvailable());
```
malfet:
```suggestion
    return hasMPS() ? 1 : 0;
```

PR #164573, opened 2025-10-03
Use codegen for the boxed interpreters

ezyang:
* __->__ #164573

Authored with claude code.  The arg parsing is kind of horrible, open
to more suggestions.


torch/fx/graph.py
```diff
@@ -28,7 +28,7 @@
 from .node import _get_qualified_name, _type_repr, Argument, Node, Target
 
 
-__all__ = ["PythonCode", "CodeGen", "Graph"]
+__all__ = ["PythonCode", "CodeGen", "Graph", "_BoxedCodeGen"]
```
albanD: You can't put private API in there.
You can just remove this as it doesn't do anything.

torch/fx/graph.py
```diff
@@ -848,6 +848,59 @@ def _tensor_annotation(t: torch.Tensor) -> str:
...
+        """
+        Process boxed inputs. Expects a list of arguments.
+        """
+        return args_list
+
+    def process_outputs(self, out: Any) -> Any:
+        """
+        Process outputs - just return as-is for boxed calling convention.
+        """
+        return out
```
albanD: dead code

albanD (approved):
Small nit,
I think the input unpacking will be terrible one way or the other haha

Sounds good if perf is actually where you need it to be!

torch/fx/graph.py
```diff
@@ -848,6 +848,59 @@ def _tensor_annotation(t: torch.Tensor) -> str:
...
+        Generate function definition for boxed calling convention.
+
+        Instead of taking individual arguments, the generated function takes
+        a single 'args_list' parameter, extracts placeholder values from it,
+        and clears the list.
+        """
+        # Generate the function signature with args_list parameter
+        fn_def = f"def {self._func_name}(self, args_list){maybe_return_annotation}:"
+
+        if len(free_vars) > 0:
```
swolchok: empty containers are falsey, non-empty containers are truthy

(why don't people ever rely on that anymore?)

torch/fx/graph.py
ezyang: I BLAME CLAUDE sorry for not reviewing carefully enough lol

albanD (approved):

jansel (approved):

ezyang:
test fail is real

PR #164576, opened 2025-10-03
Make custom op alias check consistent

albanD:
* #164359
* #164520
* __->__ #164576


torch/_library/utils.py
```diff
@@ -341,13 +341,13 @@ def check_aliasing_constraint(name, prev, result, get_module=lambda: "???"):
...
     custom operators' outputs must not alias any inputs or other outputs.
     """
-    storages = {id(t.untyped_storage()) for t in prev if isinstance(t, torch.Tensor)}
+    storages = {t.untyped_storage()._cdata for t in prev if isinstance(t, torch.Tensor)}
     tuple_result = result
     if not isinstance(result, tuple):
         tuple_result = (result,)
     for tensor in iter_tensors(tuple_result, {}):
-        key = id(tensor.untyped_storage())
-        if id(tensor.untyped_storage()) in storages:
```
soulitzer, line 350: Curious what cases these can diverge, e.g. 
Can these diverge if storages have PyObject preservation?

soulitzer (approved):

torch/_library/utils.py
albanD: Well, if storage doesn't have pyobject preservation, that would be plain wrong all the time as we would keep getting ids from temporary objects.
But in theory yes, there should be a one to one mapping between PyObject and StorageImpl. But I wanted to enforce it as a sanity check.

zou3519 (commented):
late review but this lgtm, thanks for catching it

PR #164569, opened 2025-10-03
respect aten planned overlap in inductor

eellison:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #164738
* #164581
* __->__ #164569
* #164568

Now that we have a hop to add implicit deps - use those deps for comm/compute overlap.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben

PR #164581, opened 2025-10-03
Limit path search within range

eellison:
* #164738
* __->__ #164581
* #164569
* #164568

When we are looking if two nodes are dependent, limit path search within the bounds of their node idxs. 

### DIFF
--- a/torch/_inductor/fx_passes/overlap_preserving_bucketer.py
+++ b/torch/_inductor/fx_passes/overlap_preserving_bucketer.py
@@ -1,4 +1,5 @@
 from collections import defaultdict
+from dataclasses import dataclass, field
 from typing import Optional
 
 import torch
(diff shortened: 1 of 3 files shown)


torch/_inductor/fx_passes/overlap_preserving_bucketer.py
```diff
@@ -11,10 +12,58 @@
     is_reduce_scatter_tensor as is_reduce_scatter,
     is_wait_tensor,
 )
-from torch._inductor.fx_passes.overlap_scheduling import CollBucket, CollectiveInfo
+from torch._inductor.fx_passes.overlap_scheduling import CollectiveInfo
 from torch.utils._ordered_set import OrderedSet
 
 
+@dataclass
```
Skylion007: Nit: might want to make slots=true here since the class looks potentially performance sensitive/ in hot paths and only modifies declared instance variables

ezyang (approved):
High level approach seems fine, I didn't do a fine grained code review

PR #164159, opened 2025-09-29
[cutlass-4][take 2] upgrade to cutlass 4.2.1

henrylhtsang:
Test Plan: Sandcastle

Differential Revision: D83492704

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben

PR #164589, opened 2025-10-03
[BE][cutlass backend] BE changes post cutlass_cppgen name change

henrylhtsang:
Differential Revision: D83809105

Handle reviews from https://github.com/pytorch/pytorch/pull/164159

### DIFF
--- a/torch/_inductor/codegen/cuda/cutlass_utils.py
+++ b/torch/_inductor/codegen/cuda/cutlass_utils.py
@@ -9,6 +9,7 @@ import time
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Optional
+from typing_extensions import TypeIs
 
 import sympy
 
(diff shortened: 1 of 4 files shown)


torch/_inductor/codegen/cuda/cutlass_utils.py
```diff
@@ -40,20 +40,17 @@ def move_cutlass_compiled_cache() -> None:
...
         return
 
     try:
-        filename = os.path.basename(python_cutlass.CACHE_FILE)
-        shutil.move(python_cutlass.CACHE_FILE, os.path.join(cache_dir(), filename))
+        filename = os.path.basename(cutlass_cppgen.CACHE_FILE)
+        shutil.move(cutlass_cppgen.CACHE_FILE, os.path.join(cache_dir(), filename))
         log.debug("Moved CUTLASS compiled cache file to %s", cache_dir())
     except OSError as e:
         log.warning("Failed to move CUTLASS compiled cache file: %s", str(e))
```
Skylion007:
```suggestion
        log.warning("Failed to move CUTLASS compiled cache file: %s", e)
```
'%s' does an implicit str cast anyway

torch/_inductor/codegen/cuda/cutlass_utils.py
```diff
@@ -422,7 +419,7 @@ def get_max_alignment(inductor_layout: Layout) -> int:
     size = inductor_layout.size
     offset = inductor_layout.offset
 
-    def is_static_int(number):
+    def is_static_int(number: object) -> bool:
```
Skylion007:
```suggestion
    def is_static_int(number: object) -> TypeIs[int | sympy.Integer]:

Skylion007: import from typing_extensions if not already imported

henrylhtsang: actually I don't feel like I understand why we should return TypeIs

Skylion007 (approved):

PR #164591, opened 2025-10-03
Add knobs in FR dump by watchdog (stacktrace and only active collectives) and trigger FR even on any exceptions

sbak5:
This PR includes a couple of changes to extend FlightRecorder dump by PyTorch watchdog

- New knobs to control FR dump as suggested in the public documentation even for watchdog 
(TORCH_INCLUDE_STACK_TRACE, TORCH_INCLUDE_ONLY_ACTIVE)
- Trigger the flight recorder dump on exceptions which could be triggered by any CUDA / host side error
  (TORCH_NCCL_EXTRA_DUMP_ON_EXEC)
-> Can be used as a snapshot of the workload progress for post-mortem analysis

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @EikanWang @jgong5 @wenzhe-nrv @sanchitintel @jeffdaily @sunway513 @jithunnair-amd @pruthvistony @ROCmSupport @dllehr-amd @jataylo @hongxiayang @naromero77amd @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela

### DIFF
--- a/torch/csrc/distributed/c10d/FlightRecorder.hpp
+++ b/torch/csrc/distributed/c10d/FlightRecorder.hpp
@@ -256,6 +256,16 @@ struct FlightRecorder {
       bool onlyActive);
 };
 
+// Whether to include stack trace in the Flight Recorder trace (default true)
+static std::vector<std::string> TORCH_INCLUDE_STACK_TRACE = {
+    "TORCH_INCLUDE_STACK_TRACE"};
+
+// Whether to include only active collectives in the Flight Recorder trace
+// (default false)
+static std::vector<std::string> TORCH_INCLUDE_ONLY_ACTIVE = {
+    "TORCH_INCLUDE_ONLY_ACTIVE"};
+
+
 // Dumps the fr traces and additional information about the Process
 // Group.
 TORCH_API std::string dump_fr_trace(
--- a/torch/csrc/distributed/c10d/FlightRecorder.hpp
+++ b/torch/csrc/distributed/c10d/FlightRecorder.hpp
@@ -265,7 +265,6 @@ static std::vector<std::string> TORCH_INCLUDE_STACK_TRACE = {
 static std::vector<std::string> TORCH_INCLUDE_ONLY_ACTIVE = {
     "TORCH_INCLUDE_ONLY_ACTIVE"};
 
-
 // Dumps the fr traces and additional information about the Process
 // Group.
 TORCH_API std::string dump_fr_trace(
--- a/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
+++ b/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
@@ -1432,14 +1432,17 @@ void ProcessGroupNCCL::dumpExtraDebuggingInfo() {
   bool dumpExtraOnExec_ = getCvarBool(TORCH_NCCL_EXTRA_DUMP_ON_EXEC, false);
   if (dumpExtraOnExec_) {
     bool should_dump_local = false;
-    bool succeded = shouldDump_.compare_exchange_strong(should_dump_local, true, std::memory_order_release, std::memory_order_acquire);
+    bool succeded = shouldDump_.compare_exchange_strong(
+        should_dump_local,
+        true,
+        std::memory_order_release,
+        std::memory_order_acquire);
     if (succeded) {
-      LOG(INFO) << logPrefix()
-                << "Sending extra dumping signal";
+      LOG(INFO) << logPrefix() << "Sending extra dumping signal";
       broadcastDumpSignal();
-      // When this routine is called, exception is captured so 
-      // dumping by default_pg is not guaranteed due to early termination of process
-      // So we call dumping manually here 
+      // When this routine is called, exception is captured so
+      // dumping by default_pg is not guaranteed due to early termination of
+      // process So we call dumping manually here
       bool onlyActive = getCvarBool(TORCH_NCCL_INCLUDE_ONLY_ACTIVE, false);
       // Stacktrace is not included at the moment to prevent deadlock due to GIL
       dumpDebuggingInfo(false, onlyActive);
@@ -1452,7 +1455,7 @@ void ProcessGroupNCCL::abort() {
   // This will log counter for how long the abort actually takes.
   STATIC_SCOPED_WAIT_COUNTER(pytorch.ProcessGroupNCCL__abort);
 
-  dumpExtraDebuggingInfo(); 
+  dumpExtraDebuggingInfo();
   // Don't join threads here since the purpose of this method is to abort all
   // communicators and signal the threads to exit. Joining on the threads could
   // potentially block and hence avoid it in this method.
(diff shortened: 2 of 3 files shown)


sbak5:
@fduwjj This PR includes changes discussed about the knobs and extra dumping on failures.

fduwjj (commented):
Thanks for contributing to FR code and thanks for your feedback. Overall the change looks reasonable, but I am a little bit concerned that adding more dumps might interfere with existing timeout dump, especially it would override the existing dump file. So can we set the default value of `TORCH_NCCL_EXTRA_DUMP_ON_EXEC` to be false, so you can directly turn on it on your end and we can gradually roll it out on the Meta side?

torch/csrc/distributed/c10d/ProcessGroupNCCL.hpp
```diff
@@ -126,6 +126,18 @@ static std::vector<std::string> TORCH_NCCL_COORD_CHECK_MILSEC = {
 static std::vector<std::string> TORCH_NCCL_LOG_CPP_STACK_ON_UNCLEAN_SHUTDOWN = {
     "TORCH_NCCL_LOG_CPP_STACK_ON_UNCLEAN_SHUTDOWN"};
 
+// Whether to include stack trace in the Flight Recorder trace (default true)
+static std::vector<std::string> TORCH_NCCL_INCLUDE_STACK_TRACE = {
+    "TORCH_NCCL_INCLUDE_STACK_TRACE"};
+
+// Whether to include only active collectives in the Flight Recorder trace (default false)
+static std::vector<std::string> TORCH_NCCL_INCLUDE_ONLY_ACTIVE = {
+    "TORCH_NCCL_INCLUDE_ONLY_ACTIVE"};
```
fduwjj: Also since FR now is a separate module and is generic, can we not make it NCCL specific? WDYT? I mean just `TORCH_INCLUDE_STACK_TRACE`. Ideally these two ENV values need to be defined in torch/csrc/distributed/c10d/FlightRecorder.hpp.

torch/csrc/distributed/c10d/ProcessGroupNCCL.hpp
```diff
@@ -126,6 +126,18 @@ static std::vector<std::string> TORCH_NCCL_COORD_CHECK_MILSEC = {
...
+
+// Whether to include only active collectives in the Flight Recorder trace (default false)
+static std::vector<std::string> TORCH_NCCL_EXTRA_DUMP_ON_EXEC = {
+    "TORCH_NCCL_EXTRA_DUMP_ON_EXEC"};
```
fduwjj, line 132: dump logic now is still PGNCCL specific, so I am ok with keeping this as it is.

sbak5:
> Thanks for contributing to FR code and thanks for your feedback. Overall the change looks reasonable, but I am a little bit concerned that adding more dumps might interfere with existing timeout dump, especially it would override the existing dump file. So can we set the default value of `TORCH_NCCL_EXTRA_DUMP_ON_EXEC` to be false, so you can directly turn on it on your end and we can gradually roll it out on the Meta side?

It doesn't trigger duplicate dumping due to `compare_exchange` (only extra dumping is triggered when this atomic CAS is sucessful) but I agree with you setting the var to false. Pushed a change.

torch/csrc/distributed/c10d/ProcessGroupNCCL.hpp
sbak5: Thanks for the suggestion. I think that should happen in a separate PR which may include some refactoring of other env vars in `ProcessGroupNCCL.hpp` to `FlightRecorder.hpp`

sbak5: Moved the env vars to `FlightRecorder.hpp`

torch/csrc/distributed/c10d/FlightRecorder.hpp
```diff
@@ -256,6 +256,16 @@ struct FlightRecorder {
       bool onlyActive);
 };
 
+// Whether to include stack trace in the Flight Recorder trace (default true)
+static std::vector<std::string> TORCH_INCLUDE_STACK_TRACE = {
```
fduwjj, line 260: thanks!!

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
```diff
@@ -1869,7 +1896,8 @@ void ProcessGroupNCCL::HeartbeatMonitor::runLoop() {
   if (checkDumpSignal && shouldDump_.load()) {
     // Store debug info to storage if no other thread does it. (By default to
     // local disk)
-    bool dumpStackTrace = true;
+    bool dumpStackTrace = getCvarBool(TORCH_INCLUDE_STACK_TRACE, true);
+    bool onlyActive = getCvarBool(TORCH_INCLUDE_ONLY_ACTIVE, false);
```
fduwjj, line 1900: nit: not a blocking, usually inside PGNCCL, we will first check the ENV value and store it as a member, so that we don't check it every time when we call this function. Unless, you plan to dynamically change the value of this ENV.

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
```diff
@@ -1426,17 +1426,41 @@ bool ProcessGroupNCCL::abortComms(
   return true;
 }
 
+void ProcessGroupNCCL::dumpExtraDebuggingInfo() {
+  // This extra dump is intended to capture the current snapshot of collectives
+  // When this process group is terminated for some exception out of NCCL
+  bool dumpExtraOnExec_ = getCvarBool(TORCH_NCCL_EXTRA_DUMP_ON_EXEC, false);
+  if (dumpExtraOnExec_) {
+    bool should_dump_local = false;
+    bool succeded = shouldDump_.compare_exchange_strong(
```
fduwjj: Well, when you have multiple PGs, compare_exchange_strong only works on one object but the dump (writer) is global (it is a singleton because FR itself is singleton). I know the current design is a bit hacky, and we do want to consolidate the watchdog thread and monitor thread to be per class. But it is a different story from this PR, that is why I am a bit worried about having this might override the more completed dump.

fduwjj (approved):
Overall, this looks reasonable to me and I am ok to merge this change for now. But like I mention in the comment, there could be potential race condition in the dump for the same rank which you need to be careful about.

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
sbak5: This is intended to happen only on the first PG observing exception so any other PG, which also sees the same CUDA exceptions, doesn't try to dump it. In the current default path, dumping happens only on PG 0. 
the `shouldDump_` seems to be shared by all threads so this routine prevents any duplicate dumping on other process groups than the first PG.  All ranks try the CAS and one PG with succeded proceed to broadcast and dump. The PG 0 doesn't do dumping. 
the local variable is simply used for a reference value. What's actually updated is `shouldDump_`

PR #164594, opened 2025-10-03
torch.compile: Increase subprocess parent death check interval to lower cpu

c00w:
Summary:
This check is a good idea (we could potentially do it with prctl). However
we're seeing elevated rates of cpu usage in idle worker threads. This causes issues on production jobs, causing a large amount of spikeness in qps.

Test Plan:
Tested on a prod job with caches force disabled via
TORCH_COMPILE_FORCE_DISABLE_CACHES=1

Baseline
image

With this diff -
image

Differential Revision: D83803302

### DIFF
--- a/torch/_inductor/compile_worker/utils.py
+++ b/torch/_inductor/compile_worker/utils.py
@@ -27,7 +27,7 @@ def _async_compile_initializer(orig_ppid: int) -> None:
 
     def run() -> None:
         while True:
-            sleep(1)
+            sleep(60)
             if orig_ppid != os.getppid():
                 os.kill(os.getpid(), signal.SIGKILL)
 


torch/_inductor/compile_worker/utils.py
```diff
@@ -27,7 +27,7 @@ def _async_compile_initializer(orig_ppid: int) -> None:
 
     def run() -> None:
         while True:
-            sleep(1)
+            sleep(60)
```
masnesral, line 30: @c00w do you think there'd be any further advantage to introducing some random jitter too? Is there any negative affect from threads waking up at about the same time? Or the crux of it is that any time this thread wakes up it risks delaying some critical CPU operation so we just prefer that we lower the chances of it waking up at the wrong time?

masnesral (approved):

torch/_inductor/compile_worker/utils.py
c00w: I think it's the second (every time this runs, it could potentially hit a critical cpu operation, that causes a GPU op to get delayed and thus lag).

PR #164606, opened 2025-10-03
[fr] Re-order mismatch check in fr analysis script

fduwjj:
* __->__ #164606

In reality we found the current mismatch order does not match the actual error distribution, so we reorder it a bit as following:
1. We do collective type check first
2. Then size check (excluding all2all)
3. dtype check
4. state check


VieEeEw (approved):
Thanks @fduwjj !

PR #164619, opened 2025-10-03
[torchfuzz] check in some more xfail repros

bobrenjc93:
* __->__ #164619

### DIFF
--- a/test/test_torchfuzz_repros.py
+++ b/test/test_torchfuzz_repros.py
@@ -171,6 +171,106 @@ class TestFuzzerCompileIssues(TestCase):
         out_compiled.sum().backward()
         print("Compile Success! ✅")
 
+    @pytest.mark.xfail(reason="Issue #164484")
+    def test_fuzzer_issue_164484(self):
+        torch.manual_seed(9157)
+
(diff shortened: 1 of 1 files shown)


ezyang (approved):
Please consider running a minimizer (even a claude code minimizer) on the repros first!

jeffdaily:
@bobrenjc93 @ezyang  this PR adds the xfail test test_fuzzer_issue_163876 which isn't always failing for rocm CI.  What does that mean?

bobrenjc93:
@jeffdaily at a high level these tests track eager/compile divergence bugs the [fuzzer has discovered](https://github.com/pytorch/pytorch/issues?q=is%3Aissue%20state%3Aopen%20author%3Abobrenjc93). The fuzzer currently only runs in a H100 environment, so it's possible these cases won't fail on ROCM. I know in the past your team has been strongly opposed to marking test to skip rocm. Do you think it's reasonable to do that here? Or perhaps you have another suggestion?

Once the fuzzer is more stabilized, I'm happy to run it on ROCM as well, but the same problem will still stand (bugs that repro on nvidia but not rocm or vice versa).

PR #164623, opened 2025-10-03
[DTensor] Improve sharding propagation error msg in DTensor dispatch

### LINKED ISSUE #164543: [DTensor] Improve Sharding propagation error message
I got 
RuntimeError('Sharding propagation failed for Op(op=aten.embedding.default, args_schema=Spec(S(0) on (2048, 256)), Spec((Shard(dim=0), Replicate()) on (16, 2048)) @ mesh: (2, 4))')

This is hard to read. 

I want to see 
Sharding propagation failed for 
aten.embedding.default(dt: fp32[2048, 256][S(0)], dt: fp32[16,2048][S(0), R]) on mesh('dp_shard'=2, 'tp'=4)

yiming0416:
Fixes #164543

This PR improves the `__str__` method of DTensor's `OpSchema` to provide better readable error message when dispatch fails as the error message prints `{op_info.schema}`

example 1 `aten.embedding`
```
aten.embedding.default(Spec(f32[2048, 256](S(0))), Spec(i64[16, 2048](S(0)R))) on DeviceMesh((dp=2, tp=2), 'cuda', stride=(2, 1)))
```

example 2 `aten.mm`
```
aten.mm.default(Spec(f32[1024, 512](S(1))), Spec(f32[512, 256](S(0)))) on DeviceMesh((tp=4), 'cuda', stride=(1,)))
```

example 3 `aten._scaled_dot_product_flash_attention`
```
aten._scaled_dot_product_flash_attention.default(Spec(f16[8, 16, 128, 64](RS(1))), Spec(f16[8, 16, 128, 64](RS(1))), Spec(f16[8, 16, 128, 64](RS(1)))) on DeviceMesh((dp=2, tp=4), 'cuda', stride=(4, 1)))
```

Added test
```
python test/distributed/tensor/test_dtensor_ops.py -k test_embedding_error_msg
```

### DIFF
--- a/torch/distributed/tensor/_dtensor_spec.py
+++ b/torch/distributed/tensor/_dtensor_spec.py
@@ -9,6 +9,8 @@ from torch.distributed.tensor.placement_types import (
     Replicate,
     Shard,
 )
+from torch.utils._debug_mode import _stringify_shape
+from torch.utils._dtype_abbrs import dtype_abbrs
 
 
 class TensorMeta(NamedTuple):
@@ -106,14 +108,16 @@ class DTensorSpec:
         if len(self.placements) == 1:
             placement_str = str(self.placements[0])
         else:
-            placement_str = str(self.placements)
+            placement_str = f"{''.join(str(p) for p in self.placements)}"
 
         if self.tensor_meta is not None:
-            tensor_shape = str(tuple(self.tensor_meta.shape))
+            tensor_shape = _stringify_shape(self.tensor_meta.shape)
+            tensor_dtype = dtype_abbrs[self.tensor_meta.dtype]
         else:
             tensor_shape = "unknown shape"
+            tensor_dtype = "unknown dtype"
 
-        return f"Spec({placement_str} on {tensor_shape})"
+        return f"Spec({tensor_dtype}{tensor_shape}({placement_str}))"
 
     @property
     def shape(self) -> torch.Size:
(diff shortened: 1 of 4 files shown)


SherlockNoMad:
can you share a few example err msg in the summary?

+ folks to nitpick

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -106,14 +106,16 @@ def __str__(self) -> str:
         if len(self.placements) == 1:
             placement_str = str(self.placements[0])
         else:
-            placement_str = str(self.placements)
+            placement_str = f"({', '.join(str(p) for p in self.placements)})"
```
zpcore: I think we don't need ', ' here as most people are adapted with `S(0)R` instead of `S(0), R`. Just my personal opinion.

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -106,14 +108,16 @@ def __str__(self) -> str:
         if len(self.placements) == 1:
             placement_str = str(self.placements[0])
         else:
-            placement_str = str(self.placements)
+            placement_str = f"({''.join(str(p) for p in self.placements)})"
```
zpcore: nit: what about removing outer () in placement_str and move it to  
```
return f"Spec({tensor_dtype}{tensor_shape}({placement_str}))"
```

yiming0416: @zpcore good catch! updated

zpcore (approved):
LGTM!

PR #164624, opened 2025-10-04
Use PyObject_GetOptionalAttrString in PyObject_FastGetAttrString when available

swolchok:
* #164625
* __->__ #164624

Python 3.13 added PyObject_GetOptionalAttrString. I'm not 100% certain that it is strictly better than the old approach in all cases, but based on documentation/comments it seems to be meant for this type of use, and it's faster when I profile torchtitan training (which gets to the "check for the `__torch_function__` attr on some object" part of maybe_has_torch_function frequently enough to notice, but wastes a bunch of time generating exceptions that we then suppressed here).

### DIFF
--- a/torch/csrc/utils/python_strings.h
+++ b/torch/csrc/utils/python_strings.h
@@ -3,6 +3,7 @@
 #include <torch/csrc/python_headers.h>
 #include <torch/csrc/utils/object_ptr.h>
 #include <torch/csrc/utils/pybind.h>
+#include <torch/csrc/utils/python_compat.h>
 #include <stdexcept>
 #include <string>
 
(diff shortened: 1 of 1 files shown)


torch/csrc/utils/python_strings.h
```diff
@@ -98,6 +99,14 @@ inline void THPUtils_internStringInPlace(PyObject** obj) {
  */
 
 inline py::object PyObject_FastGetAttrString(PyObject* obj, const char* name) {
+#if IS_PYTHON_3_13_PLUS
+  PyObject* res = (PyObject*)nullptr;
```
Skylion007, line 103: Huh, so the cast from nulptr necessary? Probably a holdover from C to C++ copy paste but probably better to save for a file wide cleanup

Skylion007 (approved):

torch/csrc/utils/python_strings.h
swolchok: not sure, but didn't want to wait a full CI round-trip just to find out the old one was there for good reason

PR #164625, opened 2025-10-04
Save Python refcount bump on each arg in maybe_handle_torch_function

swolchok:
* __->__ #164625
* #164624

Pybind's API entails a small unnecessary overhead when working with args. (Similarly, we should probably be using vectorcall, but that's a bigger change for both us and pybind11.)

### DIFF
--- a/torch/csrc/jit/python/pybind_utils.cpp
+++ b/torch/csrc/jit/python/pybind_utils.cpp
@@ -874,10 +874,19 @@ std::optional<py::object> _maybe_handle_torch_function(
   std::vector<PyObject*> overloaded_args;
   const auto args_size = args.size();
   size_t total_arg_num = args_size + kwargs.size();
+  PyObject* const args_ptr = args.ptr();
   for (const auto i : c10::irange(args_size)) {
-    is_tensor_and_append_overloaded(args[i].ptr(), &overloaded_args);
+    // Because pybind object indexing is implemented generically for
+    // all objects, operator[] returns py::object instead of
+    // py::handle, so args[i].ptr() would cause a reference count
+    // round trip. This has enough overhead that I noticed it while
+    // profiling and came here to fix it. For tuples in particular,
+    // PyTuple_GET_ITEM returns a borrowed reference, so no counting
+    // overhead.
+    auto* const args_i_ptr = PyTuple_GET_ITEM(args_ptr, i);
+    is_tensor_and_append_overloaded(args_i_ptr, &overloaded_args);
     is_tensor_list_and_append_overloaded(
-        args[i].ptr(),
+        args_i_ptr,
         &overloaded_args,
         static_cast<int>(total_arg_num),
         false /* throw_error */);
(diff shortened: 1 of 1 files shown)


Skylion007:
We have vector call on the roadmap for pybind11 and would welcome the contribution upstream there.

torch/csrc/jit/python/pybind_utils.cpp
```diff
@@ -874,10 +874,19 @@ std::optional<py::object> _maybe_handle_torch_function(
...
   for (const auto i : c10::irange(args_size)) {
-    is_tensor_and_append_overloaded(args[i].ptr(), &overloaded_args);
+    // Because pybind object indexing is implemented generically for
+    // all objects, operator[] returns py::object instead of
+    // py::handle, so args[i].ptr() would cause a reference count
+    // round trip. This has enough overhead that I noticed it while
+    // profiling and came here to fix it. For tuples in particular,
+    // PyTuple_GET_ITEM returns a borrowed reference, so no counting
+    // overhead.
+    auto* const args_i_ptr = PyTuple_GET_ITEM(args_ptr, i);
```
albanD: Please use PyTuple_GetItem for proper bound checking.

torch/csrc/jit/python/pybind_utils.cpp
```diff
@@ -874,10 +874,19 @@ std::optional<py::object> _maybe_handle_torch_function(
...
   size_t total_arg_num = args_size + kwargs.size();
+  PyObject* const args_ptr = args.ptr();
   for (const auto i : c10::irange(args_size)) {
-    is_tensor_and_append_overloaded(args[i].ptr(), &overloaded_args);
+    // Because pybind object indexing is implemented generically for
+    // all objects, operator[] returns py::object instead of
+    // py::handle, so args[i].ptr() would cause a reference count
+    // round trip. This has enough overhead that I noticed it while
+    // profiling and came here to fix it. For tuples in particular,
+    // PyTuple_GET_ITEM returns a borrowed reference, so no counting
```
albanD: Please add a check that this is actually a tuple

albanD (approved):
Two small nits, bu SGTM

torch/csrc/jit/python/pybind_utils.cpp
swolchok: we know that the argument is in range because `i` is iterating over `args.size()`. Do you mean that we should ignore this and bounds check anyway?

torch/csrc/jit/python/pybind_utils.cpp
swolchok: is `static_assert(std::is_base_of_v<py::tuple, std::decay_t<decltype(args)>>)` good enough? (py::args inherits from py::tuple)

albanD: Sounds fair to me yes

torch/csrc/jit/python/pybind_utils.cpp
albanD: I'm mostly concerned that future changes could break this.
But if we consider this to be perf-critical, I'm ok with skipping the bound check.

albanD (approved):
Accepting with the extra assert.

torch/csrc/jit/python/pybind_utils.cpp
swolchok: > if we consider this to be perf-critical, I'm ok with skipping the bound check.

I don't think it's perf-critical, just noticeable. I'll change it to PyTuple_GetItem and leave a note that the bounds check could be skipped if anyone comes back and looks at the code because of perf.

PR #164629, opened 2025-10-04
[torch] Preserve buffer attributes in _apply

syleshfb:
Summary: Preserves buffer attributes in _apply so that attributes added to buffers are not lost when, for example, the model is moved to a different device.

Test Plan: `test_buffer_attribute_preservation`

Differential Revision: D83883919

### DIFF
--- a/torch/nn/modules/module.py
+++ b/torch/nn/modules/module.py
@@ -1021,7 +1021,12 @@ class Module:
 
         for key, buf in self._buffers.items():
             if buf is not None:
-                self._buffers[key] = fn(buf)
+                buf_applied = fn(buf)
+
+                if compute_should_use_set_data(buf, buf_applied):
+                    buf.data = buf_applied
+                else:
+                    self._buffers[key] = buf_applied
 
         return self
 
--- a/test/test_nn.py
+++ b/test/test_nn.py
@@ -13660,6 +13660,14 @@ if __name__ == '__main__':
         y = y.contiguous(memory_format=torch.contiguous_format)
         self.assertEqual(y, y_ref)
 
+    @onlyCUDA
+    def test_buffer_attribute_preservation(self):
+        model = torch.nn.Linear(1, 1)
+        model.register_buffer("test", torch.Tensor([0]))
+        model.test.test_attr = "test_attr"
+        model = model.to('cuda')
+        self.assertEqual(model.test.test_attr, "test_attr")
+
 
 class TestFunctionalPickle(TestCase):
 


torch/nn/modules/module.py
```diff
@@ -1023,6 +1023,16 @@ def compute_should_use_set_data(tensor, tensor_applied) -> bool:
             if buf is not None:
                 self._buffers[key] = fn(buf)
 
+                if torch.nn.parameter.is_lazy(self._buffers[key]):
```
mikaylagawarecki: why do we handle lazy buffers separately?

test/test_torch.py
```diff
@@ -429,6 +429,14 @@ def test_module_share_memory(self):
         _model_cuda = model.to('cuda')
         model.share_memory()
 
+    @onlyCUDA
```
mikaylagawarecki: this should live in test/test_nn.py

mikaylagawarecki (commented):
Could I clarify the rationale for this change please 

- If I understand correctly, the current behavior is that neither parameters nor buffers will preserve attributes. Why should we only preserve attributes for buffers and not parameters?
- What is the use case where you are running into this? Is it possible to manually repopulate the attributes you need to propagate? I would need to think about this more, but I imagine there might be cases where we don't want to copy all attributes.

torch/nn/modules/module.py
syleshfb: To avoid this:
```
ValueError: Attempted to use an uninitialized parameter in <function Tensor.__dir__ at 0x7fdc48be9a80>. This error happens when you are using a `LazyModule` or explicitly manipulating `torch.nn.parameter.UninitializedBuffer` objects. When using LazyModules Call `forward` with a dummy batch to initialize the parameters before calling torch functions
```

mikaylagawarecki:
Thanks for clarifying, 

I realized there are a subset of cases for parameters ([swap_tensors path](https://github.com/pytorch/pytorch/blob/2641ed4ff21b280384900af9df9c4775f1583dc1/torch/nn/modules/module.py#L972-L981)) and the ["else" path](https://github.com/pytorch/pytorch/blob/2641ed4ff21b280384900af9df9c4775f1583dc1/torch/nn/modules/module.py#L992-L996)) where the attributes of parameters do not get preserved) but I see what you mean that for plain tensors that are nn.Parameters, the default path would be to preserve the attributes

syleshfb:
> Thanks for clarifying,
> 
> I realized there are a subset of cases for parameters ([swap_tensors path](https://github.com/pytorch/pytorch/blob/2641ed4ff21b280384900af9df9c4775f1583dc1/torch/nn/modules/module.py#L972-L981)) and the ["else" path](https://github.com/pytorch/pytorch/blob/2641ed4ff21b280384900af9df9c4775f1583dc1/torch/nn/modules/module.py#L992-L996)) where the attributes of parameters do not get preserved) but I see what you mean that for plain tensors that are nn.Parameters, the default path would be to preserve the attributes

Ah, perhaps I should preserve the attributes in those two cases as well.

albanD (changes requested):
Copying python fields is very much something we never do anywhere and generally very un-expected.
The attributes are properties of the Tensors, and when the Tensor is swapped (whether its a parameter or buffer), these attributes not being preserved is ok.

If you want a Module-centric way to keep these attributes, I would suggest storing them on the Module itself.
But we should NOT copy such attributes as it is generally not ok in python to copy an object attributes onto another.

If you want to make a case for having a .data= version of the buffer update to match the parameter behavior more closely you can, but given that we're generally sunsetting all of these APIs, I don't expect any maintainer will be interested in that approach.

torch/nn/modules/module.py
```diff
@@ -1021,7 +1021,12 @@ def compute_should_use_set_data(tensor, tensor_applied) -> bool:
 
         for key, buf in self._buffers.items():
             if buf is not None:
-                self._buffers[key] = fn(buf)
+                buf_applied = fn(buf)
+
+                if compute_should_use_set_data(buf, buf_applied):
+                    buf.data = buf_applied
```
albanD, line 1027: This is very much BC-breaking.
If we were ok with this kind of changes, we would have changed the parameters version to not do .data a long time ago.

albanD (changes requested):

PR #164631, opened 2025-10-04
[dynamo, 3.14] fix misc. bugs to get most dynamo unittests passing locally in 3.14

williamwen42:
* #166355
* __->__ #164631

cc @EikanWang @jgong5 @wenzhe-nrv @sanchitintel @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela


test/dynamo/test_export.py
```diff
@@ -4613,6 +4614,7 @@ def forward(self, x):
 
 
 class ExportTestsDevice(torch._dynamo.test_case.TestCase):
+    @unittest.skipIf(sys.version_info >= (3, 14), "cpython slot assert fail")
```
Lucaskabela: Do we have more context why this fails in 3.14?

test/dynamo/test_fx_graph_runnable.py
```diff
@@ -136,7 +136,7 @@ def _exec_and_verify_payload(self):
             tmp.write(payload)
             tmp.flush()
             res = subprocess.run(
-                [sys.executable, tmp.name], capture_output=True, text=True, timeout=30
+                [sys.executable, tmp.name], capture_output=True, text=True, timeout=120
```
Lucaskabela: 2 mins doesn't seem too bad - do we need to cc CI team though as a heads up?

Lucaskabela (approved):
Generally looks good - seeking more context on the slots failure and wanted to check time adjustment before landing

test/dynamo/test_export.py
williamwen42: Will need to investigate further - will remove this skip for now.

test/dynamo/test_fx_graph_runnable.py
williamwen42: This might be because I'm running on debug cpython and it's slower than release version. I'll remove the timeout increase for now and see what happens on release once 3.14 CI is ready.

mlazos (approved):

PR #164637, opened 2025-10-04
Remove CUDA 11 workarounds for CUB_SUPPORTS_SCAN_BY_KEY and CUB_SUPPORTS_UNIQUE_BY_KEY

cyyever:
`CUB_SUPPORTS_SCAN_BY_KEY` and `CUB_SUPPORTS_UNIQUE_BY_KEY` are true since CUDA 12. This PR removes the old branches and source files.


aten/src/ATen/native/cuda/TensorTopK.cpp
```diff
@@ -19,7 +19,6 @@
 
 namespace at::native {
 
-// TODO: remove this when CUDA <11.6 is no longer supported
 void topk_out_with_sort(
```
janeyx99, line 22: should this be removed?

cyyever: ROCm still uses `topk_out_with_sort`, so it has to be kept.
```
  bool should_use_sort(const Tensor& self, int64_t dim) {    
  #if defined(USE_ROCM)    
    if (self.dtype() == kBool) return false; // Bool sort not supported in ROCm: https://github.com/pytorch/pytorch/issues/139972    
    return (self.numel() >= 10000 && self.numel() == self.size(dim)); // based on the experiments in https://github.com/pytorch/pytorch/pull/146387    
  #else    
    return false;    
  #endif    
  }    
```

ezyang (approved):

PR #164643, opened 2025-10-04
Bump tested Numba to 0.62.1

cyyever:
Try to bump tested Numba to 0.62.1 for all Python versions. If the changes are desired, we may need to discuss keep testing older versions of Numba.

### DIFF
--- a/.ci/docker/requirements-ci.txt
+++ b/.ci/docker/requirements-ci.txt
@@ -117,7 +117,7 @@ ninja==1.11.1.4
 #Pinned versions: 1.11.1.4
 #test that import: run_test.py, test_cpp_extensions_aot.py,test_determination.py
 
-numba==0.62.1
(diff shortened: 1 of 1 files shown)


cyyever:
@AlekseiNikiforovIBM Sadly, does redhat provide ways to update llvm? It is unrealistic to build llvm from source.

cyyever:
@albanD Note that the tested version of Python 3.10 is 0.55.2, which is a major version behind...

albanD:
@cyyever why is it a problem to be a major version behind? This environment is not meant to be the fastest with all the latest feature. But make sure that everything is working properly for our whole range of users. To that end, testing the oldest possible version is often what we do (for example for NumPy).

cyyever:
@albanD That makes sense, thank you for explaining that. Indeed it is quite important to cover all supported versions.

PR #162659, opened 2025-09-10
[BE] Make PyObjectSlot use a global PyInterpreter

PaliC:
This pr gets rid of the pyobj_interpreter_ variable from PyObjectSlot and saves a word in the process

Gonna ask for review from @huydhn as there are some changes to CI.

Testing: imported internally and the failed android build seems to work now!

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @aditew01 @EikanWang @voznesenskym @penguinwu @Guobing-Chen @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @Lucaskabela @ezyang @chenyang78

PR #164652, opened 2025-10-04
[2/N] Fix clang-tidy readability checks 

cyyever:
This PR applies clang-tidy readability checks to jit sources and all headers in the code base.
`readability-redundant-inline-specifier` is suppressed because it incurs too many changes. `readability-redundant-inline-specifier` is used to detect redundant inline specifiers on function and variable declarations. There are many in-class method definitions that are marked inline.

 
cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @EikanWang @jgong5 @wenzhe-nrv @sanchitintel @mingfeima @XiaobingSuper @ashokei @jingxu10 @jerryzh168 @voznesenskym @penguinwu @Guobing-Chen @zhuhaozhe @blzheng @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela @xmfan

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -66,6 +66,7 @@ readability-simplify-subscript-expr,
 readability-string-compare,
 -readability-redundant-access-specifiers,
 -readability-redundant-control-flow,
+-readability-redundant-inline-specifier,
 '
 HeaderFilterRegex: '^(aten/|c10/|torch/).*$'
 WarningsAsErrors: '*'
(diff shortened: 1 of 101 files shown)


Skylion007 (approved):
I can't find it now, but I saw some std::vector<std::vector that was default initialized while looking through this P (and then immediately filled with a forloop. Probably want to do a grep for that kind of line and fix those.

izaitsevfb:
@pytorchbot revert -m "need to revert due to a conflict with revert of https://github.com/pytorch/pytorch/pull/162659" -c ghfirst

Awfully sorry for the churn, but this is a fastest way to unblock the revert of #162659, please rebase and re-merge at your convenience.

PR #164656, opened 2025-10-04
Remove outdated CUB macros

cyyever:
This PR removes `CUB_SUPPORTS_NV_BFLOAT16` and `CUB_SUPPORTS_FUTURE_VALUE` because they are always true on CUDA >=12 installations with its CUB version. Their branches are also removed.


albanD (approved):
Thanks a lot!

eqy (approved):

jeffdaily (approved):

PR #164659, opened 2025-10-04
Remove unused operator<< in kernel_meta_info.cpp

cyyever:
This operator is unused.

### DIFF
--- a/torch/csrc/inductor/aoti_eager/kernel_meta_info.cpp
+++ b/torch/csrc/inductor/aoti_eager/kernel_meta_info.cpp
@@ -92,30 +92,6 @@ bool TensorMetadata::operator==(const TensorMetadata& other) const {
   }
 }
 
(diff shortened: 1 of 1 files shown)


torch/csrc/inductor/aoti_eager/kernel_meta_info.cpp
```diff
@@ -92,30 +92,6 @@ bool TensorMetadata::operator==(const TensorMetadata& other) const {
   }
 }
 
-std::ostream& operator<<(
```
Skylion007, line 95: Useful to have for debugging, is it not externally linked?

cyyever: No externally linked.

EikanWang: We can expose it:)

PR #164661, opened 2025-10-04
Simplify typing of datapipes

cyyever:
Fixes #ISSUE_NUMBER


torch/utils/data/datapipes/_typing.py
```diff
@@ -337,43 +326,43 @@ def _getitem_(self, params):
...
+        bases = (cls,) + cls.__bases__
 
-        return self.__class__(
+        return cls.__class__(
             name,
             bases,
             {"__init_subclass__": _dp_init_subclass, "__type_class__": True, "type": t},
         )
 
     # TODO: Fix isinstance bug
```
kinto0, line 352: what is the isinstance bug? does this change fix it?

PR #164669, opened 2025-10-04
Move AT_DISPATCH_SWITCH and AT_DISPATCH_CASE macros to headeronly.

pearu:
As in the title.

This PR defines a new API for headeronly Dispatch.h consisting of the following CPP definitions:
- `AT_DISPATCH_SWITCH_PRELUDE(DISPATCHNAME, ENUMTYPE)` - by default is a no-op
- `AT_DISPATCH_CASE_PRELUDE(ENUMTYPE)` - by default is a no-op
- `AT_DISPATCH_DEFAULT(DISPATCHNAME, ENUMTYPE)` - by default raises a runtime error about not implemented enum type

that the code including torch/headeronly/core/Dispatch.h may define.

Moving `AT_DISPATCH_CASE_QINT` and `AT_QINT_SUB_BYTE_PRIVATE_CASE_TYPE` to headeronly is postponed as we first need to move `toUnderlying` to headeronly.

* #165323
* __->__ #164669

cc @jbschlosser


aten/src/ATen/Dispatch.h
```diff
@@ -61,15 +61,17 @@ TORCH_API void record_kernel_function_dtype(std::string name);
...
+#define AT_DISPATCH_CASE_PRELUDE(ENUMTYPE) \
+  AT_PRIVATE_CHECK_SELECTIVE_BUILD(ENUMTYPE)
+#define AT_DISPATCH_SWITCH_PRELUDE(DISPATCHNAME, ENUMTYPE) \
+  RECORD_KERNEL_FUNCTION_DTYPE(DISPATCHNAME, ENUMTYPE)
+#define AT_DISPATCH_DEFAULT(DISPATCHNAME, ENUMTYPE) \
+  TORCH_CHECK_NOT_IMPLEMENTED(                      \
+      false, '"', DISPATCHNAME, "\" not implemented for '", ENUMTYPE, "'")
+/* Include of torch/headeronly/Dispatch.h must follow the definitions
+   of AT_DISPATCH_SWITCH_PRELUDE, AT_DISPATCH_CASE_PRELUDE, and
+   AT_DISPATCH_DEFAULT macros. */
```
janeyx99, line 73: explain why!

torch/headeronly/core/Dispatch.h
```diff
@@ -0,0 +1,152 @@
+#pragma once
+
+#include <torch/headeronly/core/ScalarType.h>
+
+/*
+  When defined, AT_DISPATCH_SWITCH_PRELUDE macro is inserted before
+  the dispatch switch statement.
```
janeyx99, line 7:
```suggestion
  the dispatch switch statement. It is intentionally empty
  for torch/headeronly whereas the original ATen implementation
  will call AT_PRIVATE_CHECK_SELECTIVE_BUILD(ENUMTYPE).
```
If we go with this approach, we are claiming that AT_PRIVATE_CHECK_SELECTIVE_BUILD(ENUMTYPE) is not necessary for custom ops libraries. Can you help me understand why audio doesn't care about it? If we anticipate that custom ops will want this functionality, then maybe we should be moving these macros to torch/csrc/stable instead. 

The reason I suggested migrating the other macro first is because it is definitely headeronly-able.

torch/headeronly/core/Dispatch.h
```diff
@@ -0,0 +1,152 @@
...
+*/
+#ifndef AT_DISPATCH_SWITCH_PRELUDE
+#define AT_DISPATCH_SWITCH_PRELUDE(DISPATCHNAME, ENUMTYPE)
+#endif
+
+/*
+  When defined, AT_DISPATCH_CASE_PRELUDE macro is prepended to the
+  case statement of the dispatch switch statement.
```
janeyx99, line 15: similar comment here, are we convinced that RECORD_KERNEL_FUNCTION_DTYPE(DISPATCHNAME, ENUMTYPE) is not of interest to other custom ops?

janeyx99 (commented):
Thanks for your detailed explanation for your approach. The reason I am hesitant about this design is because I'm not convinced RECORD_KERNEL_FUNCTION_DTYPE(DISPATCHNAME, ENUMTYPE) and AT_PRIVATE_CHECK_SELECTIVE_BUILD(ENUMTYPE) are not relevant for custom ops. cc @mikaylagawarecki  and @albanD for thoughts.

torch/headeronly/core/Dispatch.h
pearu: > Can you help me understand why audio doesn't care about it?

In short, if audio has not cared about it so far then considering that audio development has been frozen, then selective build of audio remains unsupported.

That said, this question is just not about selective build of audio and I guess other custom ops libraries may care about it, although, I don't know if such libraries exist (other than PyTorch) that support selective builds.

PR #164676, opened 2025-10-04
Prevent torch._check causing graph breaks

### LINKED ISSUE #163668: torch._check causing graph breaks
### 🐛 Describe the bug

Repro:
```
import torch

@torch.compile(fullgraph=True)
def f(x):
    torch._check(x.shape[0] > 3, lambda: f"{x.shape[0]} is not greater than 3")
    return x + 1

x = torch.randn(3, device="cuda")
torch._dynamo.maybe_mark_dynamic(x, 0)
print(f(x))
```
Error:
```python
NotImplementedError: NestedUserFunctionVariable()

  File "/home/willfeng/local/helion/repro_dynamo.py", line 10, in <module>
    print(f(x))
          ^^^^
  File "/home/willfeng/local/pytorch-nightly/torch/_dynamo/eval_frame.py", line 895, in compile_wrapper
    raise e.with_traceback(None) from e.__cause__  # User compiler error
torch._dynamo.exc.Unsupported: Failed to convert args/kwargs to proxy
  Explanation: Missing `as_proxy()` implementation for some arg/kwarg.


  Developer debug context: call_function args: SymNodeVariable() NestedUserFunctionVariable() 

 For more details about this graph break, please visit: https://meta-pytorch.github.io/compile-graph-break-site/gb/gb0055.html

from user code:
   File "/home/willfeng/local/helion/repro_dynamo.py", line 5, in f
    torch._check(x.shape[0] > 3, lambda: f"{x.shape[0]} is not greater than 3")

```

### Versions

PyTorch nightly

cc @chauhang @penguinwu @voznesenskym @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @amjames @Lucaskabela

jmaczan:
Handle `torch._check` in `TorchInGraphFunctionVariable.call_function`. Basically, it has two arguments - a predicate (bool) and a message (callable). If predicate is a constant, evaluate `torch._check`. If predicate is true, it just will compile and nothing happens. If predicate is false, `torch._check` will raise an exception. 

If predicate is not constant, we manually emit a proxy. I tried to build as_proxy() inside NestedUserFunctionVariable, but failed to, that's why I create it here. I try to extract message. If it's a function, I retrieve it. If not, set it to None. Maybe we could extract it if message is a closure, but not sure how

Fixes #163668


jmaczan:
Hi @StrongerXi @xmfan @yf225, the fix is ready for a review

jmaczan:
Btw I just recalled that there are more `check_` prefixed functions, I'm going to check if they cause graph breaks too, so I can patch them within this PR

torch/_dynamo/variables/torch.py
```diff
@@ -1431,6 +1411,54 @@ def patched_fn(*args, **kwargs):
                     args=list(map(ConstantVariable.create, exc.args)),
                 )
 
+        if self.value is getattr(torch, "_check", None):
```
StrongerXi: Please follow the pattern of other `def handle_xxx` methods, e.g., https://github.com/pytorch/pytorch/blob/361692cb544114b7377da8a24c1fbf16d314edc1/torch/_dynamo/variables/torch.py#L1069-L1075

torch/_dynamo/variables/torch.py
```diff
@@ -1431,6 +1411,54 @@ def patched_fn(*args, **kwargs):
...
+            # both args are variable trackers
+            # first arg is a predicate vt, second is optional message vt
+            if not args:
+                unimplemented_v2(
+                    gb_type="torch._check missing predicate",
+                    context="torch._check() was called without predicate",
+                    explanation="torch._check requires a predicate argument",
+                    hints=[*graph_break_hints.USER_ERROR],
+                )
```
StrongerXi: For now could you follow the `assert ...` pattern above? The reason is for consistency, and in the future we might come around and clean out the all the arity check handling systematically, to avoid introducing too many similar graph break messages.

You can treat this as us being a little hand-wavy at the error message and saying "compile" programs are _expected_ to be runnable in eager, which should be mostly true in practice since we are a jit compiler.

torch/_dynamo/variables/torch.py
```diff
@@ -1431,6 +1411,54 @@ def patched_fn(*args, **kwargs):
...
+                        )
+                    else:
+                        message = message_vt.as_python_constant()
+                except NotImplementedError:
+                    unimplemented_v2(
+                        gb_type="Can't extract message from torch._check",
+                        context="torch._check() has a second argument, which is expected to be a message",
+                        explanation="Trying to build a proxy of torch._check() message, but failed to",
+                        hints=[*graph_break_hints.SUPPORTABLE],
+                    )
```
StrongerXi: Given the previous comment, maybe this is not needed anymore?

torch/_dynamo/variables/torch.py
```diff
@@ -1431,6 +1411,54 @@ def patched_fn(*args, **kwargs):
...
+            message = None
+            if len(args) >= 2:
+                message_vt = args[1]
+                try:
+                    if isinstance(message_vt, NestedUserFunctionVariable):
+                        message = (
+                            None
+                            if message_vt.has_closure()
+                            else message_vt.get_function()
+                        )
```
StrongerXi: This is the most subtle part of this PR. I think we should actually graph break explicitly on `message_vt.has_closure()` and suggest compile doesn't support tracing through `torch._check` whose error message function captures external variables, e.g., `lambda: f"{x.shape[0]} is ...`.

torch/_dynamo/variables/torch.py
```diff
@@ -1431,6 +1411,54 @@ def patched_fn(*args, **kwargs):
...
+                        context="torch._check() has a second argument, which is expected to be a message",
+                        explanation="Trying to build a proxy of torch._check() message, but failed to",
+                        hints=[*graph_break_hints.SUPPORTABLE],
+                    )
+
+            predicate_vt = args[0]
+
+            if predicate_vt.is_python_constant():
+                self.value(predicate_vt.as_python_constant(), message)
+                return ConstantVariable.create(None)
```
StrongerXi: Do we actually need this explicit constant folding?

torch/_dynamo/variables/torch.py
```diff
@@ -1431,6 +1411,54 @@ def patched_fn(*args, **kwargs):
...
+
+            # emit missing proxy that was impossible to implement inside NestedUserFunctionVariable
+            # but needed to fix the issue https://github.com/pytorch/pytorch/issues/163668
```
StrongerXi: Let's move this to above `message = ...` since that's the fundamental fix.

StrongerXi (commented):
Dropped some nits, thanks!

torch/_dynamo/variables/torch.py
jmaczan: Not sure. I tried another approaches, but I hit another problems with building proxy and gave up. As I understand it, when predicate is constant, we want to evaluate right away and we don't want the `_check` inside a graph, because it would be a dead code (?), so constant folding is the only way I know to achieve it

torch/_dynamo/variables/torch.py
jmaczan: Done

torch/_dynamo/variables/torch.py
jmaczan: Sounds good, done for `args`, `has_closure` and VT

torch/_dynamo/variables/torch.py
jmaczan: It probably depends what's more important to the users and I don't know the answer - is it to have `_check`s in runtime regardless the presence of the message or is it to only have `_check`s in the runtime when message is exactly the way they defined it. And I have no opinion on that yet. 

Do we want to consult more people on breaking the graph break on closure in runtime vs removing the message?

For now I added an assertion for `has_closure`. If we ok with adding a custom message for this case, we can add it, I wasn't sure if we are.

torch/_dynamo/variables/torch.py
jmaczan: Removed

torch/_dynamo/variables/torch.py
jmaczan: Done, thanks!

jmaczan (commented):
I updated the code with @StrongerXi remarks. Thank you for the review! 

I'm still in doubt about graph break vs set to None for message with closure, when we have to add `_check` to the graph

test/dynamo/test_check.py
williamwen42, line 1: Maybe we could rename this to `test_torch.py` and also move a bunch of tests that test variables/torch.py related functionality. Otherwise, we can just move these after `test_torch_check` in `test_misc.py`. (I'd prefer not to create new test files unless the tested feature is substantial)

torch/_dynamo/variables/torch.py
```diff
@@ -1256,6 +1256,45 @@ def handle_set_default_device(
 
             return ConstantVariable.create(None)
 
+        @register(torch._check)
+        def handle_check(self, tx: "InstructionTranslator", *args, **kwargs):
+            # both args are variable trackers
+            # first arg is a predicate vt, second is optional message vt
+            assert args
```
williamwen42: Does improper arg format result in an `AssertionError` (or does dynamo swallow it)? If an assert failure is thrown, then I would say `raise_observed_exception` is preferred here over assert

torch/_dynamo/variables/torch.py
```diff
@@ -1256,6 +1256,45 @@ def handle_set_default_device(
...
+            message_graph = None
+            if len(args) >= 2:
+                message_vt = args[1]
+
+                if message_vt.is_python_constant():
+                    message_eager = message_vt.as_python_constant()
+                    message_graph = message_eager
+                else:
+                    assert isinstance(message_vt, NestedUserFunctionVariable)
+                    assert not message_vt.has_closure()
```
williamwen42: This should be a graph break (`unimplemented_v2`) here since this case shouldn't result in an error on eager.

williamwen42 (commented):
@StrongerXi I think we have some differing opinions on observed exception + graph break vs. assert

test/dynamo/test_check.py
jmaczan: I did it and it uncovered a regression when I ran `test_misc.py` tests ;d Still working on it

jmaczan (commented):
I moved tests to `test_misc` it and it uncovered a regression when I ran `test_misc.py` tests ;d Still working on it

torch/_dynamo/variables/torch.py
jmaczan: I got rid of that

torch/_dynamo/variables/torch.py
jmaczan: Done

jmaczan (commented):
@williamwen42 @StrongerXi  Updated it a bit - now there are graph breaks when there's a closure in the message or when message VT is not NestedUserFunctionVariable. I simplified code a bit too. I tested it manually for few different scenarios and I'm working on unit tests now. From what I saw in previous test runs there might be lots of tests to be changed in multiple existing tests

I'll move on with testing once we more or less land on implementation

jmaczan:
Ping @williamwen42 @StrongerXi please take a look at the updates and let me know if it's fine. I'll continue once you confirm

test/dynamo/test_check.py
williamwen42: Were you able to fix this?

torch/_dynamo/variables/torch.py
```diff
@@ -1256,6 +1256,55 @@ def handle_set_default_device(
...
+        @register(torch._check)
+        def handle_check(self, tx: "InstructionTranslator", *args, **kwargs):
+            message_eager = None
+            message_graph_proxy = None
+            if len(args) >= 2:
+                message_vt = args[1]
+                if message_vt.has_closure():
+                    unimplemented_v2(
+                        gb_type="Message in torch._check cannot have a closure",
+                        context="Message VT has a closure",
```
williamwen42:
```suggestion
                        context=str(message_vt)
```

torch/_dynamo/variables/torch.py
```diff
@@ -1256,6 +1256,55 @@ def handle_set_default_device(
...
+                        explanation=(
+                            "Trying to build a proxy of torch._check() message, but failed to"
+                            ", because it has a closure, which are unsupported for now."
+                            "If possible, remove any variables from message and try again."
+                        ),
+                        hints=[*graph_break_hints.SUPPORTABLE],
+                    )
+                if not isinstance(message_vt, NestedUserFunctionVariable):
+                    unimplemented_v2(
```
williamwen42, line 1357: I think we can combine the condition checks into 1 graph break call - i.e. check that `message_vt` is an instance of `NestedUserFunctionVariable` and has no closure.

Then the graph break callsite could look something like
```python
unimplemented_v2(
    gb_type="Can't extract message from torch._check()",
    context=str(message_vt),
    explanation="The second argument of torch._check() must be a function defined within the torch.compile region that does not reference a non-local variable."
    hints=[
        "Make sure the message function is defined in the torch.compile region.",
        "Remove any non-local variables, e.g. remove references to `x` in `lambda: f'{x} failed check'`",
        *graph_break_hints.SUPPORTABLE,
    ],
)

**Review by williamwen42 — COMMENTED** [review]
Can you also double check happens if the message callable refers to a global variable? This should result in a `NestedUserFunctionVariable` that doesn't have a closure.

File: test/dynamo/test_check.py
Comment (jmaczan) [reply test/dynamo/test_check.py:1]: Yes

File: torch/_dynamo/variables/torch.py (hunk 9 above)
Comment (jmaczan) [reply torch/_dynamo/variables/torch.py]: Thanks, updated

File: torch/_dynamo/variables/torch.py (hunk 10 above)
Comment (jmaczan) [reply torch/_dynamo/variables/torch.py:1357]: Sure, I have joined then with `or`

File: test/dynamo/test_check.py
Comment (jmaczan) [reply test/dynamo/test_check.py:1]: But I see some new tests failing, working on it now

**Review by jmaczan — COMMENTED** [review]
Updated according to your suggestions. Pls rerun CI

File: test/dynamo/test_check.py
Comment (jmaczan) [reply test/dynamo/test_check.py:1]: It seems to be fixed now for tests that recently failed. Pls rerun CI tests

**Review by williamwen42 — APPROVED** [review]
Were you able to check this?

> Can you also double check happens if the message callable refers to a global variable? This should result in a NestedUserFunctionVariable that doesn't have a closure.

**Comment by jmaczan** [comment]
@williamwen42 I had to missed it - now I've added tests for message with a global and it works just like you wrote (so no graph break) - when predicate is True, compile runs without errors. When predicate is false, there's the same error as if the message was None or lambda without any variable

Pls rerun CI

**Review by mlazos — APPROVED** [review]

**Comment by jmaczan** [comment]
@mlazos @williamwen42 I've resolved lint issues. Pls rerun CI

File: test/dynamo/test_misc.py [hunk 11]
```diff
@@ -1428,6 +1428,168 @@ def f(x):
...
+
+        @torch.compile(backend="eager", fullgraph=True)
+        def f(x):
+            torch._check(x.shape[0] > 3, lambda: f"{GLOBAL_INT} is not greater than 3")
+            return x + 1
+
+        x = torch.randn(3)
+        torch._dynamo.maybe_mark_dynamic(x, 0)
+
+        with self.assertRaisesRegex(RuntimeError, None):
```
williamwen42: No message printed here?

jmaczan: It's printed, I forgot to add it here. `assertRaisesRegex` accepts any message when `None` is set as `expected`, so that's why it worked. Test is fixed now.

jmaczan (commented):
Test updated @williamwen42

torch/_dynamo/variables/torch.py
```diff
@@ -1287,6 +1287,60 @@ def handle_set_default_device(
...
+                        context=str(message_vt),
+                        explanation=(
+                            "The second argument of torch._check() must be a function"
+                            "defined within the torch.compile region"
+                            "that does not reference a non-local variable."
+                        ),
+                        hints=[
+                            "Make sure the message function is defined in the torch.compile region.",
+                            "Remove any non-local variables, e.g. "
+                            "remove references to `x` in `lambda: f'{x} failed check'`",
```
williamwen42:
```suggestion
                            "Remove any closure variables, e.g. "
                            "remove references to closure variable `x` in `lambda: f'{x} failed check'`",
```

Perhaps we should be more precise here - global variables are actually supported, while closure variables are not.

I wonder if it's possible to suggest a workaround for closure issues by making the variable global, or if that would result in excessive recompilations.

jmaczan: Updated. I like the idea to suggest `global` as a workaround, if user needs the dynamic value in message really badly, but since it might need lots of testing to make sure we want to include it (recompilations as you mentioned etc), perhaps we can do it in a separate PR? @williamwen42

williamwen42: sgtm

jmaczan:
CI rerun pls

jmaczan:
@williamwen42 I put `torch._check` into `manual_torch_name_rule_map` and remove from `torch_non_c_binding_in_graph_functions`, as per this

https://github.com/pytorch/pytorch/blob/a74fe75c450277eb88a95c764e8b0a664a550a86/test/dynamo/test_trace_rules.py#L478-L486

https://github.com/pytorch/pytorch/actions/runs/19112446694/job/54614633353

Pls rerun merge

PR #164678, opened 2025-10-05
[dynamo] Support torch.fx.traceback.annotate

anijain2305:
* #164776
* __->__ #164678

Builds on top of https://github.com/pytorch/pytorch/pull/163673 and https://github.com/pytorch/pytorch/pull/164174. This will be used in the followup PRs to apply regional inductor compilation.

The existing implementation let Dynamo trace into the `torch.fx.traceback.annotate`, but thats not what we want. We want Dynamo to essentially run the torch.fx.traceback.annotate function in eager, so that every Fx node created in Dynamo Fx graph has the custom meta node.

What does not work?
* We still have to set the context manager `torch.fx.traceback.preserve_node_meta()` in the user code because CI was unhappy. This can be fixed but with some perseverance.
* This does not work with graph breaks yet. But we can solve that problem, if needed, in a separate PR.


test/dynamo/test_fx_annotate.py
```diff
@@ -0,0 +1,259 @@
...
+('placeholder', 'detach_5', {'compile_inductor': 0})
+('call_function', 'zeros', {'compile_inductor': 0})
+('call_function', 'detach', {'compile_inductor': 0})
+('call_function', 'detach_2', {'compile_inductor': 0})
+('call_function', 'detach_3', {'compile_inductor': 0})
+('get_attr', 'fw_graph0', {'compile_inductor': 0})
+[]
+('get_attr', 'joint_graph0', {'compile_inductor': 0})
+[]
+('get_attr', 'mask_graph0', {'compile_inductor': 0})
```
SherlockNoMad, line 260: hmm... what are these subgraphs? I suppose they are needed for flex_attention_backward. 
Does the inner nodes need annotation?

test/dynamo/test_fx_annotate.py
```diff
@@ -0,0 +1,259 @@
...
+('get_attr', 'mask_fn_0', {'compile_inductor': 0})
+[('placeholder', 'child', {'compile_inductor': 0}), ('placeholder', 'child_1', {'compile_inductor': 0}), ('placeholder', 'child_2', {'compile_inductor': 0}), ('placeholder', 'child_3', {'compile_inductor': 0}), ('call_function', 'ge', {'compile_inductor': 0}), ('output', 'output', {'compile_inductor': 0})]
+('call_function', 'flex_attention', {'compile_inductor': 0})
+('call_function', 'out', {'compile_inductor': 0})""",  # noqa: B950
+        )
+        self.assertExpectedInline(
+            str(fw_metadata),
+            """\
+('get_attr', 'sdpa_score0', {'compile_inductor': 0})
+[('placeholder', 'arg0_1', {'compile_inductor': 0}), ('placeholder', 'arg1_1', {'compile_inductor': 0}), ('placeholder', 'arg2_1', {'compile_inductor': 0}), ('placeholder', 'arg3_1', {'compile_inductor': 0}), ('placeholder', 'arg4_1', {'compile_inductor': 0}), ('call_function', 'mul', {'compile_inductor': 0}), ('output', 'output', {'compile_inductor': 0})]
```
SherlockNoMad, line 237: nice, so both the get_attr node and its inner subgraph nodes have the annotations! 

 @mlazos, this should also work for cuda stream HOP. Do you wanna add a test to try it out? 

cc @yushangdi

SherlockNoMad (approved):

test/dynamo/test_fx_annotate.py
anijain2305: These are flex attention subgraphs. It's ok for inner nodes to be annotated as well (happens automatically) so that during desugaring of a few hops (like ac), the inner nodes get the right metadata.

jansel (approved):

xmfan (approved):

PR #164690, opened 2025-10-05
Bump pinned transformers to 4.57.1

cyyever:
This PR bumps the transformers version to 4.57.1 for tests.


desertfire (approved):
Stamping, but you will need to update those baseline files.

PR #164718, opened 2025-10-06
FakeTensorMode shouldn't cache syms when tracing

aorenste:
Improve FakeTensor cache to handle SymNode and tracing properly.

For now, when we're proxy tracing just don't bother caching operations that contain SymNodes in the output. The problem is that the proxy tracer relies on SymNode identity and our cache doesn't preserve that. It can be fixed (and I left some notes in _validate_symbolic_output_for_caching() how) but it's not worth it for now.

If we aren't proxy tracing then caching is fine.

Thus these changes:

1. Our cache key needs to include whether we were actively tracing or not - this way if we create a cache entry when we weren't tracing and then we try to use it when we ARE tracing it gets rerun.

2. If there's a SymNode in the output then bypass tracing.

3. Some general cleanup of the output validation - we were unnecessarily doing it as a two-step process when it could just be a single step (it's still two parts internally but only a single outer try/except).

* #164869
* __->__ #164718
* #164717
* #165266


torch/_subclasses/fake_tensor.py
```diff
@@ -1517,7 +1517,7 @@ def _cached_dispatch_impl(
         output = self._dispatch_impl(func, types, args, kwargs)
 
         try:
-            self._validate_cache_key(func, args, kwargs)
+            entry = self._make_cache_entry(state, key, func, args, kwargs, output)
```
aorenste, line 1541: This is just clean-up. Originally we validated the cache key before running the operation but now that we compute the output first there's no reason for this to be a separate check from the _make_cache_entry() below.

torch/_subclasses/fake_tensor.py
```diff
@@ -1574,6 +1565,11 @@ def _cache_key(
             # Disallowing dynamic shapes can introduce a DynamicOutputShapeException
             # where it wasn't seen on a previous instance of the same op.
             self.shape_env.settings if self.shape_env else None,
+            # ProxyTorchDispatchMode needs to track how SymNodes are constructed
+            # so we need to handle things a little different depending on
+            # whether we're tracing or not.
+            torch.fx.experimental.proxy_tensor.get_proxy_torch_dispatch_mode_sym_tracing()
```
aorenste, line 1571: This is the actual fix

torch/_subclasses/fake_tensor.py
```diff
@@ -3013,17 +3009,67 @@ def from_tensor(
...
+       build them.
+    """
+    from torch.fx.experimental.symbolic_shapes import _iterate_exprs, _iterate_nodes
+
+    # Check for SymNode types in PROXY mode - this should bypass caching
+    # regardless of whether symbols are known or not
+    proxy_mode = (
+        torch.fx.experimental.proxy_tensor.get_proxy_torch_dispatch_mode_sym_tracing()
+    )
+    if proxy_mode is not None:
```
aorenste, line 3064: This is part 2 of the fix.

torch/_subclasses/fake_tensor.py
```diff
@@ -1595,6 +1585,10 @@ def _cache_key(
             # Disallowing dynamic shapes can introduce a DynamicOutputShapeException
             # where it wasn't seen on a previous instance of the same op.
             self.shape_env.settings if self.shape_env else None,
+            # ProxyTorchDispatchMode needs to track how SymNodes are constructed
+            # so we need to handle things a little different depending on
+            # whether we're tracing or not.
+            torch.fx.experimental.proxy_tensor.is_sym_tracing(),
```
aorenste: This is part 1 of the actual fix

torch/_subclasses/fake_tensor.py
```diff
@@ -3070,17 +3064,65 @@ def from_tensor(
...
+       strides.
+
+    4. We could use the proxy to track "how" the SymNodes were computed and when
+       using the cache we could "replay" them properly to teach the proxy how to
+       build them.
+    """
+    from torch.fx.experimental.symbolic_shapes import _iterate_exprs, _iterate_nodes
+
+    is_sym_tracing = torch.fx.experimental.proxy_tensor.is_sym_tracing()
+    if is_sym_tracing:
```
aorenste: This is part 2 of the actual fix

bobrenjc93 (approved):
lgtm, please fix lint before landing.

PR #164729, opened 2025-10-06
Fix inplace ops on Partial DTensors to preserve aliasing semantics

RohitRathore1:
Fixes #163374. 

Here is the output from reproducible code:

```
W1006 09:09:26.329000 2457 /home/fedora/github/pytorch/torch/distributed/run.py:811] 
W1006 09:09:26.329000 2457 /home/fedora/github/pytorch/torch/distributed/run.py:811] *****************************************
W1006 09:09:26.329000 2457 /home/fedora/github/pytorch/torch/distributed/run.py:811] Setting OMP_NUM_THREADS environment variable for each process to be 1 in default, to avoid your system being overloaded, please further tune the variable for optimal performance in your application as needed. 
W1006 09:09:26.329000 2457 /home/fedora/github/pytorch/torch/distributed/run.py:811] *****************************************
  aten::clamp_(dt: f32[][R], None, 2)
    redistribute_input(0, [P] -> [R])
      redistribute_input(t: f32[], [P] -> [R])
        _c10d_functional::all_reduce(t: f32[], sum, 0)
        _c10d_functional::wait_tensor(t: f32[])
    aten::clamp_(t: f32[], None, 2)
    aten::view(t: f32[], [])
(Replicate(),)
tensor(2., device='cuda:0')
```

The behavior is now matching what you were expecting in issue #163374:

Expected behavior (from the issue):
  1. Placement should change from Partial(sum) to Replicate()
  2. Value should be tensor(2.) instead of tensor(144.)

  Actual output from this build:
  1. (Replicate(),) - placement is correct
  2. tensor(2., device='cuda:0') - value is correct

so the inplace operation now properly redistributes the partial DTensor to replicate before performing the clamp snd maintains the correct aliasing semantics. It also produces the expected clamped value. 

cc: @SherlockNoMad @ezyang @janeyx99 @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci

### DIFF
--- a/test/distributed/tensor/test_pointwise_ops.py
+++ b/test/distributed/tensor/test_pointwise_ops.py
@@ -331,6 +331,24 @@ class DistElementwiseOpsTest(DTensorOpTestBase):
         self.assertEqual(z.placements, (Replicate(),))
         self.assertEqual(z.to_local(), input)
 
+    def test_inplace_op_partial_to_replicate(self):
+        # test that in-place operations correctly update the spec when converting
+        # from Partial to Replicate placement (issue #163374)
+        device_mesh = self.build_device_mesh()
+
+        tensor = torch.ones(8, 8, device=self.device_type)
+        in_dtensor = distribute_tensor(tensor, device_mesh, [Shard(0)])
+        partial_dt = in_dtensor.sum()
+
+        self.assertTrue(partial_dt.placements[0].is_partial())
+        self.assertTrue(partial_dt.placements[0].is_partial("sum"))
+        out = partial_dt.clamp_(max=10)
+        self.assertEqual(out.placements, (Replicate(),))
+        self.assertEqual(partial_dt.placements, (Replicate(),))
+
+        full = out.full_tensor()
+        self.assertEqual(full.item(), 10.0)
+        self.assertEqual(out.to_local().item(), 10.0)
 
 if __name__ == "__main__":
     run_tests()
--- a/test/distributed/tensor/test_pointwise_ops.py
+++ b/test/distributed/tensor/test_pointwise_ops.py
@@ -343,10 +343,10 @@ class DistElementwiseOpsTest(DTensorOpTestBase):
 
         self.assertTrue(partial_dt.placements[0].is_partial())
 
-        # Inplace ops that require redistribution (Partial -> Replicate) should error
+        # Inplace ops that require placement changes (Partial -> Replicate) should error
         with self.assertRaisesRegex(
             RuntimeError,
-            "in-place operations that require redistribution are not supported"
+            "in-place operations that require placement changes are not supported",
         ):
             partial_dt.clamp_(max=10)
 
--- a/test/distributed/tensor/test_pointwise_ops.py
+++ b/test/distributed/tensor/test_pointwise_ops.py
@@ -336,12 +336,12 @@ class DistElementwiseOpsTest(DTensorOpTestBase):
         # from Partial to Replicate placement (issue #163374)
         device_mesh = self.build_device_mesh()
 
-        tensor = torch.ones(8, 8, device=self.device_type)
-        in_dtensor = distribute_tensor(tensor, device_mesh, [Shard(0)])
-        partial_dt = in_dtensor.sum()
+        input_tensor = torch.tensor(64.0, device=self.device_type)
+        partial_dt = DTensor.from_local(
+            input_tensor, device_mesh, placements=(Partial(),)
+        )
 
         self.assertTrue(partial_dt.placements[0].is_partial())
-        self.assertTrue(partial_dt.placements[0].is_partial("sum"))
         out = partial_dt.clamp_(max=10)
         self.assertEqual(out.placements, (Replicate(),))
         self.assertEqual(partial_dt.placements, (Replicate(),))
@@ -349,6 +349,7 @@ class DistElementwiseOpsTest(DTensorOpTestBase):
         full = out.full_tensor()
         self.assertEqual(full.item(), 10.0)
         self.assertEqual(out.to_local().item(), 10.0)
+
 
 if __name__ == "__main__":
     run_tests()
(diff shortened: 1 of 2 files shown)


test/distributed/tensor/test_pointwise_ops.py
```diff
@@ -331,6 +331,24 @@ def test_mul_partial(self):
...
+    def test_inplace_op_partial_to_replicate(self):
+        # test that in-place operations correctly update the spec when converting
+        # from Partial to Replicate placement (issue #163374)
+        device_mesh = self.build_device_mesh()
+
+        tensor = torch.ones(8, 8, device=self.device_type)
+        in_dtensor = distribute_tensor(tensor, device_mesh, [Shard(0)])
+        partial_dt = in_dtensor.sum()
+
+        self.assertTrue(partial_dt.placements[0].is_partial())
```
fduwjj, line 344: with L344, I guess this is not needed?

test/distributed/tensor/test_pointwise_ops.py
```diff
@@ -331,6 +331,24 @@ def test_mul_partial(self):
...
         self.assertEqual(z.to_local(), input)
 
+    def test_inplace_op_partial_to_replicate(self):
+        # test that in-place operations correctly update the spec when converting
+        # from Partial to Replicate placement (issue #163374)
+        device_mesh = self.build_device_mesh()
+
+        tensor = torch.ones(8, 8, device=self.device_type)
+        in_dtensor = distribute_tensor(tensor, device_mesh, [Shard(0)])
+        partial_dt = in_dtensor.sum()
```
zpcore: Normally we build Partial tensor in this way:
```
input_tensor = torch.randn(8, 6, device=self.device)
full_tensor = DTensor.from_local(
            input_tensor, mesh, placements=(Partial()))
        )
```
In this way `full_tensor` will be in Partial()

RohitRathore1: Thanks for the suggestion. I am going to update the test to use `DTensor.from_local()` with `placements=(Partial(),)` as the standard construction pattern. So this will make the test more direct and cleaner.

test/distributed/tensor/test_pointwise_ops.py
RohitRathore1: Aah, good catch! I am going to remove the redundant type, thanks!

torch/distributed/tensor/_dispatch.py
```diff
@@ -288,15 +288,22 @@ def default_tensor(spec: DTensorSpec) -> torch.Tensor:
...
+                output_spec = output_sharding.output_spec
+                assert isinstance(output_spec, DTensorSpec)
+                assert isinstance(args[0], dtensor.DTensor)
+                # update the spec for all inplace ops to handle placement changes
+                # (e.g., Partial -> Replicate conversion)
+                args[0]._spec = output_spec
+                # for in-place ops, we also need to ensure the local tensor is updated
+                # if redistribution occurred. The local_results contains the modified
+                # local tensor after the in-place operation.
+                if participating and output_sharding.needs_redistribute:
```
zpcore: I am not sure if `participating` condition is used correctly here. cc @ezyang

RohitRathore1: Umm, I added the `participating` check because non-participating ranks receive dummy tensors which shouldn't need updating and the redistribution logic only runs for participating ranks. According to you should non-participating ranks also update their `_local_tensor` to maintain consistency? I'd appreciate @ezyang's input on whether the condition should be:
  - `if participating and output_sharding.needs_redistribute:` or,
  - `if output_sharding.needs_redistribute:` (remove participating check)

ezyang (changes requested):
I don't think this is right. The problem is if there is a view into the partial tensor; just overwriting the local tensor is not enough to cause the views to update. In fact, I tend to think you are just toast if you need a redistribute on the tensor being inplace'd. I'd look more positively on something that just errors when this happens.

ezyang:
sure

torch/distributed/tensor/_dispatch.py
```diff
@@ -288,15 +288,30 @@ def default_tensor(spec: DTensorSpec) -> torch.Tensor:
...
+                        f"{op_call}: in-place operations that require redistribution "
+                        f"are not supported. The operation would change placement from "
+                        f"{args[0]._spec.placements} to {output_spec.placements}, "
+                        f"which requires redistribution and breaks aliasing semantics. "
+                        f"Please use the out-of-place version of this operation instead."
+                    )
+
+                # update the spec for all inplace ops to handle placement changes
+                # that don't require redistribution
+                args[0]._spec = output_spec
```
ezyang: I think it's still OK (and in fact preferred) for the squeeze_ logic to live inside the squeeze_ branch. Most inplace ops do NOT change the tensor meta. So because no redistribution could have occurred, the input cannot have a placement change.

ezyang:
Just a small tweak please, thanks

torch/distributed/tensor/_dispatch.py
RohitRathore1: Thanks @ezyang! You're absolutely right. I'will move the `args[0]._spec = output_spec` inside the `squeeze_.dim` branch.

RohitRathore1:
> Just a small tweak please, thanks

Done, thanks!

ezyang:
need to update tests

RohitRathore1:
> need to update tests

@ezyang you are referring to failing test or do I need to update existing test?

ezyang:
your ci failed, fix it

RohitRathore1:
@ezyang I've fixed the CI failure. The issue was that the placement change check was happening before the `squeeze_.dim` special case, so I moved it into the else block to exempt `squeeze_.dim`. I also removed the spec update for other inplace ops as you mentioned since most inplace ops don't change tensor meta and no redistribution occurs when placements match, the spec update isn't needed. Let me know if there's anything else that needs to be addressed.

ezyang (approved):

PR #164731, opened 2025-10-06
Add regression test for get_root_mesh with multiple independent meshes

RohitRathore1:
Fixes #163330 

I tried to reproduce the bug with my 4-GPU setup (the original issue used 8 GPUs). I created several different test scenarios, trying to trigger the bug by:
- creating two different device meshes
- slicing them in various ways
- checking if get_root_mesh() would get confused

but the bug didn't show up! Everything worked correctly in `2.10`. I found that there was a massive refactoring of the `DeviceMesh` code (PR #163213) that landed on October 2nd. That PR completely rewrote how `DeviceMesh` tracks relationships between parent meshes and submeshes using. It seems like this refactoring fixed the bug! But I added a regression test to make sure it doesn't come back. The test (`test_get_root_mesh_multiple_independent_meshes`) does exactly what the bug report described:
  - creates two independent meshes
  - slices them both
  - verifies that each submesh correctly points back to its real parent
  - makes sure submeshes from mesh1 don't incorrectly claim mesh2 as their parent

cc: @ezyang @zhiqi-0 

### DIFF
--- a/test/distributed/test_device_mesh.py
+++ b/test/distributed/test_device_mesh.py
@@ -350,6 +350,35 @@ class DeviceMeshTest(DTensorTestBase):
         # Fake pg only have BackendType as BackendType::CUSTOM.
         self.assertEqual(mesh.get_group(1)._get_backend_name(), "custom")
 
+    @with_comms
+    def test_get_root_mesh_multiple_independent_meshes(self):
+        # regression test for issue #163330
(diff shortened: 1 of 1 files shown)


fduwjj (approved):
lgtm, can you move this test after this test `test_get_root_mesh`? That way all get_root_mesh tests are grouped together? Thanks

PR #164741, opened 2025-10-06
Enable mimalloc on non-Windows platforms and make default for AArch64 builds

murste01:
This change removes the Windows requirement for mimalloc builds, and makes mimalloc the default c10 system allocator for AArch64 builds. This significantly improves the performance of AArch64 builds of PyTorch as large allocations are better cached by mimalloc than glibc.

**Updated Results**

Torchbench FP32 eager Inference, 16 threads:
mimalloc-v2-fp32-diff

Torchbench BF16 eager Inference, 16 threads:
mimalloc-v2-bf16-diff

cc @malfet @snadampal @milpuz01 @aditew01 @nikhil-arm @fadara01

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -388,9 +388,9 @@ cmake_dependent_option(USE_PRIORITIZED_TEXT_FOR_LD "Use prioritized text linker 
 
 option(USE_MIMALLOC "Use mimalloc" OFF)
 # Enable third party mimalloc library to improve memory allocation performance
-# on Windows.
+# on Windows and AArch64.
 option(USE_MIMALLOC_ON_MKL "Use mimalloc on MKL" OFF)
-if(WIN32)
+if(WIN32 OR CPU_AARCH64)
   set(USE_MIMALLOC ON)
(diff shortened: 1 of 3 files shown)


fadara01 (dismissed):
Amazing work, thank you!

murste01:
I’ve found that the original results I posted above are skewed by some outliers in the glibc results, so I’ve run some more benchmarks and expanded my scope to include bf16.

The results are still good for this PR (~28.6% mean improvement for fp32, ~19.8% mean improvement for bf16), but there are a small number of relatively minor regressions.

I still think this PR is worth merging - perf gains are good and in the past we’ve seen huge amounts of page-faults caused by poor caching of large allocations by glibc, which has generated a lot of variability in our benchmarks, and which should be massively improved by this patch.

aditew01 (dismissed):

malfet (dismissed):

aditew01 (approved):

fadara01 (approved):

malfet (approved):

PR #164743, opened 2025-10-06
[Caffe2] Add NEON transpose routine

Nicoshev:
Summary: We are adding a custom NEON-based matrix transpose procedure, enabled for NEON and SVE128 translation layers.

Test Plan:
Ported the code to fbgemm and ran:

buck2 test mode/opt //deeplearning/fbgemm:TransposeTest

Differential Revision: D83978975


malfet (commented):
If I understand correctly, this is a performance change, isn't it?

If that's the case, can you please share a benchmark?

PR #164747, opened 2025-10-06
Make c7i the default for _linux-build.yml

zxiiro:
Use linux.c7i.2xlarge as the default runner for the _linux-build.yml workflow. In testing we found that switching from c5 - c7i grants a 15-20% faster build times despite c7i costing 5% more. This should reduce costs of jobs using _linux-build.yml.

Relates to pytorch/test-infra#7175.

### DIFF
--- a/.github/workflows/_linux-build.yml
+++ b/.github/workflows/_linux-build.yml
@@ -37,7 +37,7 @@ on:
       runner:
         required: false
         type: string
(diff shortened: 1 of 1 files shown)


atalman (approved):
lgtm. Looks like all failures are not related to this PR. I don't see any involvement with ``linux.c7i.2xlarge``

PR #164748, opened 2025-10-06
Add pyrefly suppressions

maggiemoss:
Adds suppressions to pyrefly will typecheck clean: https://github.com/pytorch/pytorch/issues/163283

Test plan:
dmypy restart && python3 scripts/lintrunner.py -a
pyrefly check

step 1: delete lines in the pyrefly.toml file from the `project-excludes` field
step 2: run pyrefly check
step 3: add suppressions, clean up unused suppressions
before: https://gist.github.com/maggiemoss/4b3bf2037014e116bc00706a16aef199

after:

0 errors (4,263 ignored)

cc @EikanWang @jgong5 @wenzhe-nrv @sanchitintel @ezyang


torch/ao/nn/intrinsic/quantized/dynamic/modules/linear_relu.py
```diff
@@ -30,6 +30,7 @@ class LinearReLU(nnqd.Linear):
         torch.Size([128, 30])
     """
 
+    # pyrefly: ignore  # bad-override
     _FLOAT_MODULE = nni.LinearReLU
```
Skylion007, line 34: ,This is a legitimate error we should fix probably

oulgen (approved):

torch/ao/nn/quantizable/modules/rnn.py
```diff
@@ -454,6 +455,7 @@ def __init__(
 
         if (
             not isinstance(dropout, numbers.Number)
+            # pyrefly: ignore  # unsupported-operation
             or not 0 <= dropout <= 1
```
Skylion007, line 459: if not is causing issues, does pyrefly have a precedence bug?

torch/distributions/studentT.py
```diff
@@ -31,6 +31,7 @@ class StudentT(Distribution):
         scale (float or Tensor): scale of the distribution
     """
 
+    # pyrefly: ignore  # bad-override
```
Skylion007, line 34: Is Pyrefly incorrectly assuming that all subclasses should have a dictionary of the same literal type as the parent class? Seems like a common failure case if so.

Skylion007 (commented):
We should try to distinguish between legitimate typing bugs and pyrefly bugs (like handing subclass of legacy dynamically created namedtuple.

torch/ao/nn/intrinsic/quantized/dynamic/modules/linear_relu.py
maggiemoss: My plan is to first silence the errors and get pyrefly running, and then come back and fix ones like this. There are a few places we've identified where adding additional annotations or small refactors (i.e. checking for `None`) can reduce type errors in some files.

torch/distributions/studentT.py
maggiemoss: A simplified example also raises an error in pyright, but not in MyPy. I'll send the example to the team and see which implementation they agree with. Thanks for flagging :)

maggiemoss:
I've been running `benchmarks/dynamo/ci_expected_accuracy/update_expected.py` with what I think are the right environment variables and SHA but getting an unauthorized error.

PR #164750, opened 2025-10-06
[DeviceMesh] Remove private _set_mesh_dim_group_options API

fduwjj:
* #164954
* __->__ #164750

We allow passing in PG option via https://github.com/pytorch/pytorch/pull/159371 and we did a clean up of Meta internal usage of `_set_mesh_dim_group_options`, since this a private API, we don't have any bc guarantee, we want to directly remove so that people use the new behavior from now on.

Also since we now allow passing pg in both DeviceMesh constructor and flatten API, so that we also want to get rid of the global pg option override variable.


lw (approved):
Awesome! Thanks!

torch/distributed/device_mesh.py
```diff
@@ -583,18 +578,7 @@ def _init_process_groups(
 
                     # Respect dim group options specified via _MeshEnv.set_dim_group_options().
```
lw: Nit: update this comment

And check whether there are other ones?

fduwjj: Done.

fegin (approved):

PR #164757, opened 2025-10-06
[SymmMem] Multi-root tile reduction

kwen2501:

Perform multiple tile reductions concurrently, with each tile reduced to a separate root.

- The number of concurrent reductions can be smaller than world size, i.e. roots can be a subset of all ranks. But all ranks are still required to call into this API.

- Currently supports NVLink SHARP scope only.

### DIFF
--- a/torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
+++ b/torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
@@ -978,6 +978,89 @@ void tile_reduce(
   });
 }
 
+/* Multi-tile Communication */
+
+void multi_root_tile_reduce(
+    at::ArrayRef<at::Tensor> in_tiles,
+    at::Tensor& out_tile,
+    at::ArrayRef<int64_t> roots,
+    std::string group_name,
+    std::string reduce_op) {
+  /* Perform multiple tile reductions concurrently, with each tile reduced to a separate root.
+   Args:
+     - `in_tiles` is a list of input tensors.
+     - `out_tile` is the output tensor.
+     - `roots` is a list of root ranks corresponding to each input tile, in the same order. A rank cannot be a root more than once.
+     - `group_name` is the name of the group to use for the collective operation.
+     - `reduce_op` is the reduction operation to perform. Currently only "sum" is supported.
+   */
+  TORCH_CHECK(reduce_op == "sum", "tile_reduce: only sum is supported for now");
+  TORCH_CHECK(out_tile.dtype() == at::kFloat, "Only float is supported");
+  TORCH_CHECK(out_tile.dim() == 2, "Only 2D tensors are supported");
+  TORCH_CHECK(roots.size() == in_tiles.size(), "Number of roots must match number of tiles");
+
+  // Get device and stream
+  auto device = out_tile.device();
+  c10::cuda::CUDAGuard guard(device);
+  auto stream = at::cuda::getCurrentCUDAStream();
+
+  // Rendezvous all tensors, and find the tile "I" need to reduce
+  auto hdl = c10d::symmetric_memory::rendezvous(out_tile, group_name);
+  int rank = hdl->get_rank();
+  int world_size = hdl->get_world_size();
+  int i = 0, my_tile_idx = 0, root = world_size;
+  // Note: if there is no tile for the current rank, my_tile_idx will remain
+  // initial value 0, and root will remain `world_size`. This is OK. In
+  // `nvshmemx::tile_sum_reduce_block`, this rank would skip the reduction
+  // operation, but would still participate in the barrier.
+  for (auto& in_tile : in_tiles) {
+    TORCH_CHECK(in_tile.dtype() == at::kFloat, "Only float is supported");
+    c10d::symmetric_memory::rendezvous(in_tile, group_name);
+    if (roots[i] == rank) {
+      TORCH_CHECK(root == world_size, "Each rank can only be a root once");
+      my_tile_idx = i;
+      root = rank;
+    }
+    i++;
+  }
+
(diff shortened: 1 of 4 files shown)


weifengpy (approved):
looking good on UX part

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
```diff
@@ -978,6 +978,89 @@ void tile_reduce(
...
+  for (auto& in_tile : in_tiles) {
+    TORCH_CHECK(in_tile.dtype() == at::kFloat, "Only float is supported");
+    c10d::symmetric_memory::rendezvous(in_tile, group_name);
+    if (roots[i] == rank) {
+      TORCH_CHECK(root == world_size, "Each rank can only be a root once");
+      my_tile_idx = i;
+      root = rank;
+    }
+    i++;
+  }
```
fegin, line 1033: Should we check that `root != world_size` here?

fegin (approved):

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
kwen2501: This implementation uses `root == world_size` to indicate that current rank does not need to reduce any tile. (Yet it still calls into this API to fulfill the collective requirement). You can see the "Note" above.

ngimel: we don't have a test for it though (`root==world_size`), do we?

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
```diff
@@ -978,6 +978,89 @@ void tile_reduce(
...
+  /* Perform multiple tile reductions concurrently, with each tile reduced to a separate root.
+   Args:
+     - `in_tiles` is a list of input tensors.
+     - `out_tile` is the output tensor.
+     - `roots` is a list of root ranks corresponding to each input tile, in the same order. A rank cannot be a root more than once.
+     - `group_name` is the name of the group to use for the collective operation.
+     - `reduce_op` is the reduction operation to perform. Currently only "sum" is supported.
+   */
+  TORCH_CHECK(reduce_op == "sum", "tile_reduce: only sum is supported for now");
+  TORCH_CHECK(out_tile.dtype() == at::kFloat, "Only float is supported");
```
ngimel: can you support at least BFloat16 also?

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
```diff
@@ -978,6 +978,89 @@ void tile_reduce(
...
+  int world_size = hdl->get_world_size();
+  int i = 0, my_tile_idx = 0, root = world_size;
+  // Note: if there is no tile for the current rank, my_tile_idx will remain
+  // initial value 0, and root will remain `world_size`. This is OK. In
+  // `nvshmemx::tile_sum_reduce_block`, this rank would skip the reduction
+  // operation, but would still participate in the barrier.
+  for (auto& in_tile : in_tiles) {
+    TORCH_CHECK(in_tile.dtype() == at::kFloat, "Only float is supported");
+    c10d::symmetric_memory::rendezvous(in_tile, group_name);
+    if (roots[i] == rank) {
```
ngimel, line 1027: we should check that `roots[i]` is valid (>=0 and < world_size)

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
```diff
@@ -978,6 +978,89 @@ void tile_reduce(
...
+      root = rank;
+    }
+    i++;
+  }
+
+  // Ideally 16 bytes per thread
+  int nblocks = at::ceil_div(
+      out_tile.numel() * out_tile.element_size(),
+      (int64_t)THREADS_PER_BLOCK * 16);
+  nblocks = std::min(nblocks, 16);
```
ngimel: why limit at 16? I think for cuda backend we limit at 24 at least, maybe we even need more for blackwell

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
kwen2501: In `test_multi_root_tile_reduce`, when `root_ratio` is 2, we will exercise this case.
`root_ratio=2` means only half of the ranks are root, the rest of ranks will provide `root==world_size` here to skip the reduction.

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
kwen2501: Added now.

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
kwen2501: Added in base PR #162243

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
kwen2501: Set it to 24 now.

nWEIdia:
In retrospective, it would have been great if the test could be guarded with: 
"Currently supports NVLink SHARP scope only." 
As we found numerical mismatches on platforms without NVLink (A100 or H100)

ngimel:
Were those large precision errors, as in grossly incorrect results, or some accuracy mismatch? If the latter, feel free to submit PRs guarding the tests, if the former, can we disable those ops on hardware that doesn't support it?

PR #164766, opened 2025-10-06
Add split-K control to cuBLAS reduced-precision settings

ngimel:
## Summary
- add a CuBLASReductionOption enum so the CUDA context can track reduced-precision and split-K options
- extend the Python bindings, backend helpers, and docs to accept an optional allow_splitk argument for fp16/bf16 matmul controls
- update cuBLAS/cuBLASLt call sites plus dynamo guards and tests to respect the new combinations

## Testing
- python test/test_cuda.py TestCuda.test_cublas_allow_fp16_reduced_precision_reduction_get_set -v *(fails: ModuleNotFoundError: No module named 'psutil')*

------
https://chatgpt.com/codex/tasks/task_e_68e404623178832f8a3e1d34e1e175da

### DIFF
--- a/torch/backends/cuda/__init__.py
+++ b/torch/backends/cuda/__init__.py
@@ -139,13 +139,9 @@ class cuBLASModule:
             return value, True
         if isinstance(value, (list, tuple)):
             if not value:
-                raise TypeError(
-                    f"{attr_name} expects at least one boolean argument"
-                )
+                raise TypeError(f"{attr_name} expects at least one boolean argument")
             if len(value) > 2:
-                raise TypeError(
-                    f"{attr_name} expects at most two boolean arguments"
-                )
+                raise TypeError(f"{attr_name} expects at most two boolean arguments")
             allow_reduced_precision = _ensure_bool(value[0], "allow_reduced_precision")
             if len(value) == 1:
                 return allow_reduced_precision, True
--- a/torch/backends/cuda/__init__.py
+++ b/torch/backends/cuda/__init__.py
@@ -1,6 +1,6 @@
 # mypy: allow-untyped-defs
 import contextlib
-from typing import Union
+from typing import Any, Union
 from typing_extensions import deprecated
 
 import torch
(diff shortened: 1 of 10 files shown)


torch/csrc/Module.cpp
```diff
@@ -1224,48 +1224,84 @@ static PyObject* THPModule_allowTF32CuBLAS(
...
-  if (at::globalContext().allowFP16ReductionCuBLAS()) {
-    Py_RETURN_TRUE;
-  }
-  Py_RETURN_FALSE;
+  auto option = at::globalContext().allowFP16ReductionCuBLAS();
+  bool allow_reduced_precision =
+      option == at::CuBLASReductionOption::AllowReducedPrecisionWithSplitK;
+  bool allow_splitk = option !=
+      at::CuBLASReductionOption::DisallowReducedPrecisionDisallowSplitK;
+  return Py_BuildValue("(pp)", allow_reduced_precision, allow_splitk);
```
eqy: maybe it wants something like `Py_BuildValue("(OO)", allow_reduced_precision ? Py_True : Py_False, allow_splitk ? Py_True : Py_False);`

malfet (approved):
LGTM, though I guess @albanD would be a bit unhappy that it does not follow accelerator generic abstraction

torch/backends/cuda/__init__.py
```diff
@@ -126,13 +126,54 @@ def __setattr__(self, name, value):
...
+        def _ensure_bool(obj: Any, which: str) -> bool:
+            if isinstance(obj, bool):
+                return obj
+            raise TypeError(
+                f"{attr_name} expects a bool for {which}, but got {type(obj)!r}"
+            )
+
+        if isinstance(value, bool):
+            return value, True
+        if isinstance(value, (list, tuple)):
```
albanD, line 140: nit: the doc says tuple-only

torch/backends/cuda/__init__.py
```diff
@@ -126,13 +126,54 @@ def __setattr__(self, name, value):
...
+    def _parse_reduction_setting(value: Any, attr_name: str) -> tuple[bool, bool]:
+        def _ensure_bool(obj: Any, which: str) -> bool:
+            if isinstance(obj, bool):
+                return obj
+            raise TypeError(
+                f"{attr_name} expects a bool for {which}, but got {type(obj)!r}"
+            )
+
+        if isinstance(value, bool):
+            return value, True
```
albanD, line 139: This should query the current split_k value and not force override it to True?

albanD (approved):
That sounds good to me, this is quite specific for now, so it's ok to have a oneoff API and we can extend it later as needed.

torch/backends/cuda/__init__.py
ngimel: If we want to preserve current behavior we should force-override it to True

albanD: I see, ok

PR #164767, opened 2025-10-06
Updated binary_cross_entropy_with_logits and BCEWithLogitsLoss docume…

### LINKED ISSUE #68780: Incorrect documentation for `BCEWithLogitsLoss` `weight`?
## 📚 Documentation

The documentation for the `weight` parameter for the `BCEWithLogitsLoss` class is:

        weight (Tensor, optional): a manual rescaling weight given to the loss
            of each batch element. If given, has to be a Tensor of size `nbatch`.

This implies a vector of N

However the documentation for `F.binary_cross_entropy_with_logits` is different, it states

        weight (Tensor, optional): a manual rescaling weight
            if provided it's repeated to match input tensor shape

And `BCEWithLogitsLoss` simply forwards it's arguments to `F.binary_cross_entropy_with_logits`, so clearly "has to be a Tensor of size `nbatch` is incorrect".

In my tests `weights` can be 1xC, 1x1, Nx1 or NxC (which I guess is what the `F.binary_cross_entropy_with_logits`  docstring is implying). But it's not totally clear and I think the `BCEWithLogitsLoss` is incorrect.

cc @svekars @brycebortree @albanD @mruberry @jbschlosser @walterddr @mikaylagawarecki @brianjo @kshitij12345

cleonard530:
Fixes #68780

Updated binary_cross_entropy_with_logits and BCEWithLogitsLoss documentation to show that the weight supports shape broadcasting with respect to the input tensor's shape.

@albanD

### DIFF
--- a/torch/nn/functional.py
+++ b/torch/nn/functional.py
@@ -3590,8 +3590,8 @@ def binary_cross_entropy_with_logits(
     Args:
         input: Tensor of arbitrary shape as unnormalized scores (often referred to as logits).
         target: Tensor of the same shape as input with values between 0 and 1
-        weight (Tensor, optional): a manual rescaling weight
-            if provided it's repeated to match input tensor shape
+        weight (Tensor, optional): The dimension of weight supports :ref:`broadcasting to a common shape <broadcasting-semantics>`
+            with respect to the input (and target) shape.
         size_average (bool, optional): Deprecated (see :attr:`reduction`).
         reduce (bool, optional): Deprecated (see :attr:`reduction`).
         reduction (str, optional): Specifies the reduction to apply to the output:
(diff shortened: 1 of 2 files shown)


torch/nn/functional.py
```diff
@@ -3588,8 +3588,8 @@ def binary_cross_entropy_with_logits(
     Args:
         input: Tensor of arbitrary shape as unnormalized scores (often referred to as logits).
         target: Tensor of the same shape as input with values between 0 and 1
-        weight (Tensor, optional): a manual rescaling weight
-            if provided it's repeated to match input tensor shape
+        weight (Tensor, optional): The dimension of weight supports :ref:`broadcasting to a common shape <broadcasting-semantics>`
```
isuruf: Keep `a manual rescaling weight` here?

cleonard530: Yes, I should have left that in there. I will update this and then ping you later.

cleonard530:
Hi @mikaylagawarecki and @jbschlosser, just wanted to follow up on this PR. Please let me know if there are any edits I need to make. Thanks!

mikaylagawarecki (approved):
Thanks

PR #164769, opened 2025-10-06
Move aws OIDC credentials steps into setup-rocm.yml

jithunnair-amd:
The AWS ECR login step needs `id-token: write` permissions. We move the steps to get OIDC-based credentials from `_rocm-test.yml` to `setup-rocm.yml`. This lays the groundwork to enable access to AWS ECR in workflows in other repos such as torchtitan that use [linux_job_v2.yml](https://github.com/pytorch/test-infra/blob/main/.github/workflows/linux_job_v2.yml), which also uses [setup-rocm.yml](https://github.com/pytorch/test-infra/blob/335f4f80a0d7534a50ccc89414134b0cec8e2f3d/.github/workflows/linux_job_v2.yml#L168). 

Any caller workflows that eventually execute `setup-rocm` action will thus need to provide the `id-token: write` permission.

cc @jeffdaily @sunway513 @pruthvistony @ROCmSupport @dllehr-amd @jataylo @hongxiayang @naromero77amd


huydhn (approved):
LGTM!

PR #160843, opened 2025-08-17
[pytree] add `treespec_{leaf,tuple,dict}` functions for args_spec modification

XuehaiPan:
The goal of this PR is to provide a standard way to create simple treespec instances and hide the implementation details of the `PyTreeSpec` class.

Changes:

1. Add function `treespec_leaf()` to replace `LeafSpec()`.
2. Add function `treespec_tuple(...)` and `treespec_dict(...)` to create treespec for `tuple` / `dict` which is used for `*args` / `**kwargs`. This avoids direct modification to `treespec` instances that rely on the implementation details of the `PyTreeSpec` class.
3. Change `len(spec.children_specs)` to `spec.num_children`.
4. Change `isinstance(spec, LeafSpec)` to `spec.is_leaf()`.

------

Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.12.0) (oldest at bottom):
* __->__ #160843

cc @ezyang @SherlockNoMad @EikanWang @jgong5 @wenzhe-nrv @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela

Issue #152535, opened 2025-04-30
optree package status in PyTorch


zou3519:
## Motivation

"optree >= 0.13.0" is an optional dependency for PyTorch. There's not a good way to specify this in the package metadata, so we have code checks doing this. We ran into some issues with the PyTorch 2.7 release where the "optional dependency checking" code broken and we ended up hard requiring optree >= 0.13.0. Furthermore, I submitted two fixes for this, where the first fix didn't work and the second one actually worked.

We should fix this so that future releases (e.g. 2.8.0) do not run into the risk of this problem.

## Pitch

I see three high-level solutions:
1. We continue to have optree as an optional dependency. We would need to beef up our CI. In particular, we want to test three configurations: optree == 0.13.0, no optree, and optree < 0.13.0
2. We take the optree pypi package as a mandatory dependency.
3. We take optree as a required pytorch submodule and build it into PyTorch.

From discussions with @malfet, @seemethere, and @albanD it sounded like we preferred (3). The risk of doing (2) is that we get into dependency hell: if a third-party library pins optree and pytorch pins optree then we end up in a not-so-good-place.

Thoughts? Also cc @XuehaiPan @angelayi

seemethere:
I don't want to blow up our test matrix, if we want optree then we should just make it a mandatory dependency

NEREUScode:
@zou3519 The best way forward is to include `optree` as a built-in part of PyTorch using a submodule as you say in option 3  This avoids messy dependency issues

atalman:
Hi @XuehaiPan [RELEASE: Triton dependency for the release](https://github.com/pytorch/pytorch/blob/main/RELEASE.md#triton-dependency-for-the-release). Is basically option 2, since triton is mandatory dependency. 
For option 3 it has side effect of increasing binary size. Do we know what is the size increase we are talking about here ?

XuehaiPan:
FYI, the [Scientific Python Community](https://scientific-python.org) released a new post about Pytrees:

- Pytrees for Scientific Python: <https://blog.scientific-python.org/pytrees>

PR #137400, opened 2024-10-05
[pytree] Add public pytree module `torch.utils.pytree`

XuehaiPan:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.12.0) (oldest at bottom):
* #148180
* __->__ #137400
* #152624

Populate APIs from `torch.utils._pytree` (default) or `torch.utils._cxx_pytree` to a new public module `torch.utils.pytree`. There is an environment variable `PYTORCH_USE_CXX_PYTREE` (disabled by default) to control this.

Since the CXX pytree is now Dynamo traceable, the users can change the underlining pytree implementation by flipping the environment variable while using `torch.utils.pytree`.

Resolves #65761

- #65761

cc @zou3519 @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @kadeng @chauhang @amjames @Lucaskabela @jataylo @mlazos @chenyang78 @rec

Issue #65761, opened 2021-09-28
torch.utils._pytree -> stable


zou3519:
## 🚀 Feature

PyTorch should use a performant and stable version of _pytree.

## Motivation

`torch.utils._pytree` is something I cooked up when prototyping the initial non-composable version of vmap. Since then, it's found many legitimate use cases in PyTorch internals:
- functorch [transforms](https://github.com/facebookresearch/functorch/blob/6c026e965c50ab25a3a63f2625691c9a4a4a18ff/functorch/_src/eager_transforms.py#L13); [eager compilation](https://github.com/facebookresearch/functorch/blob/6c026e965c50ab25a3a63f2625691c9a4a4a18ff/functorch/_src/eager_compilation.py#L85)
- [fx](https://github.com/pytorch/pytorch/blob/master/torch/fx/_pytree.py)
- [writing torch_dispatch objects](https://github.com/pytorch/pytorch/blob/4752453d27c275df5f8efdea651ed9f08e366c7a/test/test_python_dispatch.py#L62)

The last use case means we want to expose torch.utils._pytree to users! (or provide a stable mechanism that users can use).

There is only really one main blocker to exposing torch.utils._pytree: performance.
- [tree_map is slow](https://github.com/pytorch/pytorch/issues/65753), it can probably be faster.
- Some of the eager compilation work is showing that torch.utils._pytree is slow, especially in the case where the input is a flat list.

## Pitch

torch.utils._pytree is heavily inspired by JAX's pytrees. JAX's pytrees are implemented in C++ (with bindings to Python) for performance reasons. We should either:
- delete our pytree and take a dependency on JAX (or figure out if they'd be interested in splitting off pytree into its own library)
- run our own implementation of pytree and improve its performance.

cc @albanD @Chillee @ezyang

Chillee:
Some benchmarks

TL;DR: Deepmind's Pytree implementation is often an order of magnitude+ slower than Jax's for unflattening. This is likely due to the fact that Deepmind's pytree provides an `unflatten_as` API while Jax uses a flatten spec. Our current implementation is way slower in almost all situations :)
```
empty list f_pt:          0.001885890961 ms
empty list f_jax:         0.000846385956 ms
empty list f_dm:          0.001115798950 ms
empty list f_u_pt:        0.002555847168 ms
empty list f_u_jax:       0.002150535583 ms
empty list f_u_dm:        0.005831718445 ms

single list f_pt:         0.003507137299 ms
[...]

PR #164771, opened 2025-10-06
Optimize the pytree utility module

kddnewton:
This commit attempts to optimize the speed of the pytree module, specifically the tree_is_leaf, tree_flatten, and tree_unflatten functions.

The main optimization is that TreeSpec no longer functions as a tree structure, but instead stores a flat list of nodes that are generated by a post-order traversal of the tree. This speeds up flattening a bit because we can use pointers into the leaf and nodes list to calculate the subtree sizes, and drastically speeds up unflattening because it becomes a linear loop through a list.

Most of the other changes in this commit follow from the structure change of the TreeSpec class. This especially includes the serialization functionality, which has a new schema. The various serialization functions have been updated to handle dumping and loading from either schema (as presumably the old schema will need to stick around for a while for migration purposes).

This commit purposefully tries to avoid changing anything outside of the _pytree.py file itself, which means there are plenty of places where additional code could be deleted/changed to make it make more sense. For example, the LeafSpec class no longer makes a ton of sense, and is a good candidate for deletion in the future.

The overall results include ~3x improvement for tree_flatten, ~2x improvement for tree_unflatten, and ~2x improvement for tree_map. Full results are below:

```
tree_flatten:
    _cxx_pytree          5730.63 ips
    _pytree2             2439.90 ips  (-57.42%)
    _pytree              850.49 ips  (-85.16%)
tree_unflatten:
    _cxx_pytree          11241.22 ips
    _pytree2             4035.69 ips  (-64.10%)
    _pytree              1862.81 ips  (-83.43%)
tree_map:
    _cxx_pytree          3049.11 ips
    _pytree2             1100.40 ips  (-63.91%)
    _pytree              520.17 ips  (-82.94%)
tree_leaves:
    _cxx_pytree          5736.86 ips
    _pytree2             1736.03 ips  (-69.74%)
    _pytree              1525.42 ips  (-73.41%)
```

Fixes #ISSUE_NUMBER

cc @zou3519 @XuehaiPan @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @kadeng @chauhang @amjames @Lucaskabela @jataylo @chenyang78

### DIFF
--- a/torch/utils/_pytree.py
+++ b/torch/utils/_pytree.py
@@ -20,6 +20,7 @@ import functools
 import importlib
 import importlib.metadata
 import json
+import sys
 import threading
 import types
 import warnings
@@ -92,7 +93,7 @@ U = TypeVar("U")
 R = TypeVar("R")
 
 
-DEFAULT_TREESPEC_SERIALIZATION_PROTOCOL = 1
+DEFAULT_TREESPEC_SERIALIZATION_PROTOCOL = 2
 NO_SERIALIZED_TYPE_NAME_FOUND = "NO_SERIALIZED_TYPE_NAME_FOUND"
 
 
@@ -107,18 +108,20 @@ class KeyEntry(Protocol):
 
 
 class EnumEncoder(json.JSONEncoder):
-    def default(self, obj: object) -> Union[str, dict[str, Any]]:
-        if isinstance(obj, Enum):
+    def default(self, o: object) -> Union[str, dict[str, Any]]:
+        if isinstance(o, Enum):
             return {
                 "__enum__": True,
-                "fqn": f"{obj.__class__.__module__}:{obj.__class__.__qualname__}",
-                "name": obj.name,
+                "fqn": f"{o.__class__.__module__}:{o.__class__.__qualname__}",
+                "name": o.name,
             }
-        return cast(str, super().default(obj))
+        return cast(str, super().default(o))
 
 
 Context = Any
 PyTree = Any
+PyTreeType = type[Any]
+
 FlattenFunc = Callable[[PyTree], tuple[list[Any], Context]]
 UnflattenFunc = Callable[[Iterable[Any], Context], PyTree]
 DumpableContext = Any  # Any json dumpable text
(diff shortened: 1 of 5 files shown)


torch/utils/_pytree.py
```diff
@@ -1048,215 +1082,224 @@ def _is_leaf(tree: PyTree, is_leaf: Optional[Callable[[PyTree], bool]] = None) -
...
-                    )
 
-                if both_standard_dict:
-                    # dictionary types are compatible with each other
-                    dict_context = (
-                        treespec.context
-                        if treespec.type is not defaultdict
-                        # ignore mismatch of `default_factory` for defaultdict
-                        else treespec.context[1]
+@final
```
Skylion007: Why can these not be final dataclasses with slots=True?
```
nodes : list[Node] = dataclasses.field(factory=list)
```
or something

Skylion007: It's good to use dataclasses so we don't ahve to manage slots especially for final classes

torch/utils/_pytree.py
```diff
@@ -1048,215 +1082,224 @@ def _is_leaf(tree: PyTree, is_leaf: Optional[Callable[[PyTree], bool]] = None) -
...
+    def children(self) -> Iterator[Self]:
+        """Returns a list of _WalkNode instances that are the children of this
+        node. This relies on the fact that the nodes in the spec are stored in
+        the order of a post-order traversal of the tree.
+        """
 
-    def __repr__(self, indent: int = 0) -> str:
-        return "*"
+        nodes = []
+        index = self.index - 1
```
angelayi: why is this -1?

torch/utils/_pytree.py
```diff
@@ -1048,215 +1082,224 @@ def _is_leaf(tree: PyTree, is_leaf: Optional[Callable[[PyTree], bool]] = None) -
...
+    else:
+        leaves.append(tree)
+        nodes.append(_LEAF)
+        return
+
+    children, context = flatten_fn(tree)
+    num_leaves = len(leaves)
+    num_nodes = len(nodes)
+
+    arity = -1
```
angelayi: why not just do `arity = len(children)`? then you can get rid of the `arity` in the for loop and the `arity + 1`?

torch/utils/_pytree.py
```diff
@@ -1267,34 +1310,109 @@ def tree_flatten(
...
+    return leaves, TreeSpec(nodes)
 
-        # Recursively flatten the children
-        subspecs = [helper(child, leaves) for child in children]
-        return TreeSpec(node_type, context, subspecs)
 
-    leaves: list[Any] = []
-    treespec = helper(tree, leaves)
-    return leaves, treespec
+def _tree_flatten_up_to(treespec: TreeSpec, tree: PyTree) -> list[PyTree]:
```
angelayi, line 1406: can you add a docstring here?

torch/utils/_pytree.py
kddnewton: I'll profile it momentarily, but IIRC making it a dataclass hurt the performance by at least 5%

torch/utils/_pytree.py
kddnewton: The last child of the current node in a post-order traversal is always going to be the node right before the current node. For example, with:

[a, [b, c], [d, e]]

the post-order traversal generated by the flatten function is:

[a, b, c, list1, d, e, list2, list0]

If you are looking for the children of list0, you will find list2 right before it, or the children if list2 you will find e right before it.

torch/utils/_pytree.py
kddnewton: I'll profile it momentarily, but it ended up being faster to do it this way. This also opens up the possibility for another optimization that I want to put in, which is that we should allow the flatten function to return an iterable, so that we don't have to reify the lists right when you flatten.

XuehaiPan (commented):
Although the `_pytree` module is a non-public module, it is already widely used in the PyTorch ecosystem (e.g., `torchvision`, third-party packages). This PR contains many BC-breaking changes. Maybe it should be implemented in a separate file instead of in-place modifying the original implementation. Otherwise, it cannot be merged.

Some context about the PyTree module:

- #65761

Making the C++ PyTree the default pytree impl:

- #152535
- #137400

Helper functions (can be added in this PR):

- #160843

torch/utils/_pytree.py
kddnewton: Yup!

Issue #164756, opened 2025-10-06
[PP] Unshard and reshard not working with FSDP


H-Huang:
We are seeing the FSDP allgathers not happening at the correct times when using `Interleaved1F1B`. The AMD team performed similar investigations (https://docs.google.com/document/d/1TbhOfPAIKoeDuVcKcMv9WNR2VgIrLxGopGuKqW6ga1s/edit?tab=t.0) and provided a fix that manually calls the `unshard()` and `reshard()` outside of the pipeline schedule. See https://github.com/AMD-AGI/torchtitan-amd/blob/dev/no_fsdp_bubble/torchtitan/distributed/amd_1f1b.py

kwen2501:
Thanks for the explanation. 
It feels like we are faced with "dy/dw and dy/di" issue but this time in forward!

PR #164775, opened 2025-10-06
[PP] Fix FSDP unshard/reshard

H-Huang:
* #164777
* __->__ #164775

First fix for https://github.com/pytorch/pytorch/issues/164756

In the pipeline IR we call `UNSHARD` and `RESHARD`,  but there is a bug because when we call `module.unshard()` these do not recursively call the FSDP modules, hence leading to sometime call allgather before the module forward. 

Since we want the pipeline IR to explicitly handle this, we can call `group.unshard` instead which ensures that all the modules are unsharded.


test/distributed/test_composability.py
```diff
@@ -375,8 +379,159 @@ def apply_dp(partial_model):
...
+        # Verify parameters are now unsharded
+        check_fsdp_unsharded_state(stage.submod, expected_unsharded=True)
+
 
 instantiate_parametrized_tests(ComposabilityTest)
 
+import fbvscode
+
+
+fbvscode.attach_debugger()
```
weifengpy: remove?

weifengpy (approved):

torch/distributed/pipelining/schedules.py
```diff
@@ -1961,18 +1964,27 @@ def _assert_unsharded(stage_idx: int):
...
-                                and stage_idx not in unshard_ops
+                                and stage_idx not in unshard_groups
                             ), f"Unsharding the same {stage_idx=} twice"
-                            unshard_ops[stage_idx] = stage.submod.unshard(async_op=True)  # type: ignore[operator]
+                            distributed_state = fully_shard.state(stage.submod)  # type: ignore[attr-defined]
+                            for state in distributed_state._state_ctx.all_states:
+                                if state._fsdp_param_group:
+                                    group = state._fsdp_param_group
+                                    group.unshard(async_op=True)
+                                    unshard_groups[stage_idx] = group
```
kwen2501: The intrusive level of this block seems a bit high.

kwen2501: We'd also need to evaluate two things:
- would every users want `all_states`? I assume that means all parameters of the submodule.
- is there any side effect of holding reference to the parameters here? (and why would we need to store the reference?) In FSDP's doc about `.unshard()`, I see that FSDP will automatically wait for the fired AG's in the module's pre-forward.

kwen2501:
```
        .. note:: If ``async_op=True``, then FSDP will wait on the pending
            unshard in the module's pre-forward for the user. The user only
            needs to call :meth:`wait` explicitly if the wait should happen
            before pre-forward.
```
Or, does this apply to `module.unshard()` only, instead of `group.unshard()`?

test/distributed/test_composability.py
wwwjn: Looks like these line cause the CI failing

H-Huang: oops, thank you 🙂

torch/distributed/pipelining/schedules.py
H-Huang: Good questions

> would every users want all_states? I assume that means all parameters of the submodule.

Since the default behavior is already "unshard after forward, dont reshard until all backwards finished". Then doing all the unshards in the beginning may allocate memory earlier, but shouldn't change peak memory. We still provide the ability for users to call unshard and reshard manually, but yeah the restriction is this now has to be all the parameters of the submodule.

> is there any side effect of holding reference to the parameters here? (and why would we need to store the reference?) In FSDP's doc about .unshard(), I see that FSDP will automatically wait for the fired AG's in the module's pre-forward.

We hold a reference to the parameters, but do delete this after we call forward. I store the reference since it is an async call and before forward computation I call `group.wait_for_unshard()`. Let me sync up with @weifengpy to make sure this is reasonable behavior.

fegin: > Since the default behavior is already "unshard after forward, dont reshard until all backwards finished".

Do we actually check if the FSDP is configured as `unshard after forward`?

weifengpy: we chatted offline about using public apis unshard(async=True) to avoid referencing to `FSDPParamGroup` for `comp_type == UNSHARD` and `comp_type == RESHARD`

```
handles = []
for submodule in model.modules():
    if not isinstance(submodule, FSDPModule):
        continue
    handle = submodule.unshard(async=True)
    handles.append(handle)
for handle in handles:
    handle.wait()
```

torch/distributed/pipelining/schedules.py
```diff
@@ -1884,13 +1884,14 @@ def _step_microbatches(
...
-        unshard_ops: dict[int, UnshardHandle] = {}
+        unshard_ops: dict[int, list[UnshardHandle]] = defaultdict(list)
         unsharded_stages = set()
 
         def _assert_unsharded(stage_idx: int):
             """If an unshard is active for `stage_idx`, wait() it and mark `stage_idx` unshared."""
             if stage_idx in unshard_ops:
-                unshard_ops[stage_idx].wait()
+                for op in unshard_ops[stage_idx]:
+                    op.wait()
```
weifengpy, line 1894: would it better to rename to `_wait_unshard` ? just because we actually call `op.wait()` inside? I usually treat _assert as an no-op sanity check

torch/distributed/pipelining/schedules.py
```diff
@@ -1884,13 +1884,14 @@ def _step_microbatches(
         send_ops: list[list[dist.Work]] = []
 
         # we track which stages are 'active' when used with FSDP, and wait on unshard ops before computing on stages
-        unshard_ops: dict[int, UnshardHandle] = {}
+        unshard_ops: dict[int, list[UnshardHandle]] = defaultdict(list)
```
weifengpy, line 1887: neat! they are looking cleaner!

torch/distributed/pipelining/schedules.py
kwen2501: That looks better, thank you!

> would every users want all_states? I assume that means all parameters of the submodule.

This question is not just about memory envelope, but also compute-comm overlap.
In the code above,
```
for handle in handles:
    handle.wait()
```
will pre-pay all AG tolls before any compute can happen.
I am still not sure if we (pipelining) should prescribe FSDP logic.

kwen2501: Just recording what we chatted offline yesterday:
making PP recv non-blocking and the received tensor an AsyncCollTensor (waited on only when it is used) is an alternative solution to touching FSDP logic -- it would also seem easier to control our own business than others' :)

PR #164777, opened 2025-10-06
[PP] Migrate other schedules to use PipelineScheduleRuntime

H-Huang:
* __->__ #164777
* #164775

Second fix for https://github.com/pytorch/pytorch/issues/164756

This has been a TODO to make the all schedules execute using the same runtime. Now after this change, schedules will use the same logic for `_PipelineScheduleRuntime` where it adds `UNSHARD` and `RESHARD` operations to the schedules which fixes the issue mentioned above. 

image

A test is failing after the conversion:
- Fixed a gradient scaling issue for dWeight


torch/distributed/pipelining/schedules.py
```diff
@@ -2081,11 +2081,19 @@ def _assert_unsharded(stage_idx: int):
...
+                        grad_scale_factor = (
+                            self._n_microbatches if self.scale_grads else 1
+                        )
                         stage.backward_weight_one_chunk(
                             mb_index,
                             last_backward=backward_counter[stage_idx]
                             == self._n_microbatches,
                         )
+                        if last_backward:
+                            stage.scale_grads(grad_scale_factor)
```
fegin:
```suggestion
                        stage.backward_weight_one_chunk(
                            mb_index, last_backward=last_backward,
                        )
                        if last_backward:
                            stage.scale_grads(grad_scale_factor)
```

fegin (approved):

torch/distributed/pipelining/schedules.py
H-Huang: Thank you!

wwwjn: n00b question: does this change solved the RS issue we were discussing offline?

PR #164783, opened 2025-10-06
Refactor memory estimator to use node storages, add test

eellison:
* #164945
* #164944
* __->__ #164783
* #164738

- Update the Memory Estimator to use node storages for analysis, which simplifies book keeping, as opposed to manually looking at operator schema. This will also allow me to reuse this component elsewhere.

- Factor out into separate class, so that this same logic can be used  in scheduling (node allocations / aliasing / uses)

- Adds Tests for correctness - right now only on fwd/bwd by itself, not with both.


torch/_inductor/fx_passes/memory_estimator.py
```diff
@@ -1,271 +1,221 @@
...
-        Args:
-        - nodes (list[fx.Node]): A list of nodes from the FX graph, ordered
-          in execution order.
-        - alias_info (AliasInfo): An instance of AliasInfo containing aliasing
-          relationships between nodes in the graph.
+    for node in itertools.chain(
+        graph.find_nodes(op="placeholder"), graph.find_nodes(op="get_attr")
+    ):
+        for storage_key in alias_info.get_fresh_allocations(node):
+            if storage_key.device.type != "cpu":
```
ruisizhang123: can we reuse `device_filter` instead of hardcode "cpu"?

torch/_inductor/fx_passes/memory_estimator.py
```diff
@@ -274,77 +224,61 @@ def get_fwd_bwd_interactions(
...
-        Returns:
-        - list[fx.Node]: A list of nodes in the output of the graph.
-        """
-        output_node = list(graph.nodes)[-1]
-        assert output_node.op == "output"
-        nodes_in_output: OrderedSet[fx.Node] = OrderedSet()
+    # Identify storages allocated by primal placeholder nodes
+    primal_storages: OrderedSet[StorageKey] = OrderedSet()
+    for node in fwd_graph.find_nodes(op="placeholder"):
+        if node.name.startswith("primals"):
```
ruisizhang123, line 250: curious: is it possible to check the input node names (which are primals) instead of hard-code "primals" in node.name strings here? Not sure if this hard-code would be very fragile to changes in graph inputs.

torch/_inductor/fx_passes/memory_estimator.py
```diff
@@ -1,271 +1,221 @@
...
+
+    def get_fresh_allocations(self, node: fx.Node) -> OrderedSet[StorageKey]:
+        """Get all fresh storage allocations by this node (not views/aliases)."""
+        return self.node_to_fresh_allocations[node]
+
+    def get_storage_uses(self, node: fx.Node) -> OrderedSet[StorageKey]:
+        """Get all storages that this node uses as inputs."""
+        return self.node_to_storage_uses[node]
+
+    def get_storages_last_used_at(
```
ruisizhang123: should be `get_storages_last_used`?

ruisizhang123 (approved):
LGTM, thanks!

eellison:
cc @karthickai could be relevant for you

PR #164776, opened 2025-10-06
[compile] Regional inductor compilation with fx.annotate

anijain2305:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #164776
* #165188

This PR introduces a way to compile a region of FX graph using `fx.traceback.annotate`.

### UX

1) In the user code, mark the region that you want to be compiled with inductor using `with fx_traceback.annotate({"compile_with_inductor": 0})`. As of now, we just rely on the string `compile_with_inductor` and ignore the integer. As the needs arise, we can update the logic.

Example

```
        def fn(x, y):
            sin = torch.sin(x)

            with fx_traceback.annotate({"compile_with_inductor": 0}):
                mul = sin * y
                add = mul + 1

            return torch.sin(add)
```

2) You have to instruct the compiler to use the annotations with `compile_fx_annotated_nodes_with_inductor` transformation. This is somewhat controversial, and a user might expect that just setting annotation is enough. But for now to control the blast radius, we need to explicitly do this. One such example is 

```

# Set the fw and bw compiler of aot_autograd to `compile_fx_annotated_nodes_with_inductor`
def aot_eager_regional_inductor():
    return aot_autograd(
        fw_compiler=compile_fx_annotated_nodes_with_inductor,
        bw_compiler=compile_fx_annotated_nodes_with_inductor,
    )

```

3) Fixable in short-term - You have to wrap the user code in `torch.fx.traceback.preserve_node_meta` to ensure that annotations are propagated to the compiler. This is fixable, just need to make CI happy.

### Implementation

1) Relies on `CapabilityBasedPartitioner` to "scoop" out regions based on annotations, and then create subgraphs in the main graph.
2) Call `torch._inductor.standalone_compile` on these subgraphs, and jam the returned callable into the FX graph at the place of call_module

Resulting graph looks something like this - search for `torch__inductor_standalone_compile_inner`

Forward graph
```
class GraphModule(torch.nn.Module):
    def forward(self, primals_1: "f32[10]", primals_2: "f32[10]"):
         # File: /data/users/anijain/pytorch2/test/dynamo/test_regional_inductor.py:64 in fn, code: sin = torch.sin(x)
        sin: "f32[10]" = torch.ops.aten.sin.default(primals_1)

        # No stacktrace found for following nodes
        inner = torch__inductor_standalone_compile_inner(sin, primals_2)

[...]

PR #164795, opened 2025-10-06
[annotate] Copy fwd to bwd metadata for subgraphs as well

anijain2305:
* #164776
* __->__ #164795
* #164772

The test is in the next PR. My older PR on dynamo annotate - https://github.com/pytorch/pytorch/pull/164678 is getting reverted due to unknown reasons, so difficult to add a test right now in this PR.  When I reland, I can add a test for this as well.

### DIFF
--- a/torch/_functorch/_aot_autograd/utils.py
+++ b/torch/_functorch/_aot_autograd/utils.py
@@ -404,17 +404,7 @@ def root_module_when_exporting_non_strict(flat_fn):
         return None
 
 
-def copy_fwd_metadata_to_bw_nodes(fx_g):
-    """
-    Input: `fx_g` which contains the joint fwd+bwd FX graph created by
-    aot_autograd.
-
-    This function walks the graph and copies over metadata from forward nodes
-    to backward nodes, using the `seq_nr` field as a one-to-many mapping
-    from forward node to backward node. This metadata is useful for performance
-    profiling and debugging.
-    """
-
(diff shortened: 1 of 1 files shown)


torch/_functorch/_aot_autograd/utils.py
```diff
@@ -459,6 +449,26 @@ def _is_backward_node_with_seq_nr(node):
...
+    from forward node to backward node. This metadata is useful for performance
+    profiling and debugging.
+    """
+
+    # Copy the metadata recursively - useful for HOPs
+    for node in fx_g.graph.nodes:
+        if node.op == "get_attr":
+            submod = getattr(fx_g, node.target)
+            if isinstance(submod, torch.fx.GraphModule):
+                copy_fwd_metadata_to_bw_nodes(submod)
```
yushangdi, line 468: do we know for sure that the backward nodes and the forward nodes are always in the same subgraph? If so this would work. In the current implementation, the bw nodes would not be able to find the fw nodes in other subgraphs.

yushangdi (approved):
LGTM, left a comment for potential future issue if bw ndoes are not in the same subgraph as fw nodes.

torch/_functorch/_aot_autograd/utils.py
yushangdi: implemented in https://github.com/pytorch/pytorch/pull/165202

PR #164797, opened 2025-10-06
[1/N] [DTensor device order] Device mesh util function to support device order placement

zpcore:
Improve some utils functions in device mesh in order to support device order placement

* #165197
* #165092
* #164902
* #164821
* #164820
* #164806
* __->__ #164797


torch/distributed/device_mesh.py
```diff
@@ -1092,6 +1114,18 @@ def _flatten(
                 self, mesh_dim_name, backend_override_tuple
             )
 
+        def get_mesh_dim_by_name(self, mesh_dim_name: str) -> int:
```
fduwjj: We have this function inside `_MeshEnv`, can you move it from the global variable into DeviceMesh class and make this a private API, similarly, we have some efforts to move most of everything out of global variable here: https://github.com/pytorch/pytorch/pull/164510

torch/distributed/device_mesh.py
```diff
@@ -1001,11 +1001,33 @@ def ndim(self) -> int:
         def shape(self) -> tuple[int, ...]:
             return tuple(self.mesh.shape)
 
-        def get_rank(self) -> int:
+        def get_rank(self, coordinate: Optional[list[int]] = None) -> int:
```
fduwjj: hmmm can we not mix it with this API? I think they are doing two different things, why are we doing this piggyback? Also, I am wondering if this function can be inside DTensor.

One more thing, since now device mesh is using layout, any chance you can directly use `crd2idx` inside torch/distributed/_pycute/int_tuple.py? That way you don't really need this code change.

fduwjj (commented):
I have some questions regarding the necessity of creating new API and the change to the existing API..

torch/distributed/device_mesh.py
zpcore: Oh, I didn't know we have `crd2idx`. Looks like it is serving the purpose. Let me revert this function.

torch/distributed/device_mesh.py
zpcore: Moved the function under DeviceMesh now.

PR #164801, opened 2025-10-07
[inductor] don't tune xblock for reduction

shunting314:
* #164905
* #164904
* #164532
* __->__ #164801

It turns out that tuning XBLOCK for a reduction can also change numerics ( https://github.com/pytorch/pytorch/pull/164525#pullrequestreview-3306235454 ). 

The PR skip tuning XBLOCK for a reduction. If we have multiple configs left with different XBLOCKs, the heuristic will pick the configs with second-largest XBLOCK. 

### DIFF
--- a/torch/_inductor/runtime/triton_heuristics.py
+++ b/torch/_inductor/runtime/triton_heuristics.py
@@ -17,7 +17,7 @@ import re
 import sys
 import threading
 import time
-from collections import defaultdict, namedtuple
+from collections import namedtuple
 from typing import (
     Any,
     Callable,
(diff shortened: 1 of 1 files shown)


jansel (approved):

mlazos (approved):

torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -2878,49 +2878,18 @@ def adapt_config_for_tiling(
...
-    and return the most promising group based on heuristics.
-
     Heuristics:
     - skip reduction configs with too small RBLOCK
     - skip reduction configs with XBLOCK==1 if we are confident it will not perform well
-    - pick the group with largest size: autotuning more configs may have more chance to give better perf
-    - if there is a tie, pick the group with second largest RBLOCK
-    - if there is still a tie, pick the group with second largest num_warps
+    - if there is a tie, pick the config with second largest RBLOCK
+    - if there is still a tie, pick the config with second largest num_warps
```
mlazos, line 2891: Why do we pick second largest in all of these cases? this for my understanding, everything else lgtm

shunting314: Good question. It all starts from collecting autotuning results for the huggingface benchmark suite and analyze them manually. It shows that second largest is a good heuristic to pick good config in general: the largest one may be too aggressive, thus I pick the second largest.

v0i0 (approved):

PR #164806, opened 2025-10-07
[2/N] [DTensor device order] Add shard_order attribute in DTensorSpec

zpcore:
Add `shard_order` field in DTensorSpec.

* #165197
* #165092
* #164902
* #164821
* #164820
* __->__ #164806
* #164797


torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -29,16 +60,97 @@ class DTensorSpec:
...
+    # devices. `shard_order` is a tuple of tuples of integers: in each inner
+    # tuple, the first element is the tensor dimension being sharded, and the
+    # remaining elements are the device mesh dimensions (in order) over which
+    # that tensor dimension is sharded.
+    #
+    # Example:
+    #   For a tensor of shape [8, 16] and a 3D device mesh, if dim 0 is sharded over
+    #   mesh dim 1, and dim 1 is sharded over mesh dim 0 and then mesh dim 2,
+    #   the shard_order would be: shard_order = ((0, 1), (1, 0, 2))
+    shard_order: TensorDimTuple = None  # type: ignore[assignment]
```
XilunWu: Optional[TensorDimTuple]

XilunWu (approved):
LGTM

torch/distributed/tensor/_dtensor_spec.py
zpcore: I have the same thought before, LOL. @ezyang  pointed out that since the class `__post_init__()` will automatically compute `shard_order`, this should not be Optional anymore: https://github.com/pytorch/pytorch/pull/160266#discussion_r2366372269

ezyang: yes, it is more important that read sites don't require Optional not handled type suppresions

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -83,6 +196,7 @@ def _check_equals(self, other: object, skip_shapes: bool = False) -> bool:
             isinstance(other, DTensorSpec)
             and self.mesh == other.mesh
             and self.placements == other.placements
+            and self.shard_order == other.shard_order
```
fduwjj, line 213: shall we do some CPU overhead benchmark for this change? Although it is not a blocker, it will give some ideas on the rough impact of this PR to DTensor CPU overhead.

zpcore: Good point! I will do overhead benchmarking after getting this feature upstream to unblock autoparalle first.

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -1,16 +1,48 @@
...
+#     - P(sum)[0]: partial reduction on mesh dimension 0
+#     - Clearly shows which tensor dims map to which mesh dims
+#
+#   When False (mesh-centric): "P(sum)S(1)[1]S(1)[0]R"
+#     - P(sum): mesh dimension 0 has partial reduction
+#     - S(1): mesh dimension 1 shards tensor dimension 1 (with order 1)
+#     - S(1): mesh dimension 2 shards tensor dimension 1 (with order 0)
+#     - R: Replicated on mesh dimension 3
+#     - Standard DTensor placement with concise format following mesh dimension order
+tensor_centric_format = True
```
ezyang: So why are you defaulting this to True? The old printing was according to False.

ezyang: I would also like to be persuaded why we should have a global flag toggling printing format. It seems confusing for users. For BC reasons we have to do mesh-oriented prints, and a more local and controlled way to do this, is to have dedicated stringification functions for the tensor centric format.

ezyang: My other hesitation is that you are using the SAME syntax for both formats. This means if I see something like S(1) without context, I cannot tell if it is tensor-centric or mesh-centric. Similarly for BC reasons, I think the old syntax (mesh-centric) wins, and you need to come up with a signifier that distinguishes the tensor centric syntax. One way to do this is to replace the numeric mesh dims with the string mesh names, in keeping with JAX's placement style.

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -29,16 +61,97 @@ class DTensorSpec:
...
             self.placements = tuple(self.placements)
-        self._hash: Optional[int] = None
+        if self.shard_order is None:
+            self.shard_order = DTensorSpec.compute_default_sparse_shard_order(
+                self.placements
+            )
+        self._hash: int | None = None
+
+    @staticmethod
+    def compute_default_sparse_shard_order(
```
ezyang: in what sense is this sparse? You return a tuple so you are dense in terms of tensor dims

ezyang: I think this is not actually returning a TensorDimTuple

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -29,16 +61,97 @@ class DTensorSpec:
...
+                    f"Shard dim {shard_dim} in placements {placements} must be normalized"
+                )
+                if shard_dim not in tensor_dim_to_mesh_dims:
+                    tensor_dim_to_mesh_dims[shard_dim] = []
+                tensor_dim_to_mesh_dims[shard_dim].append(mesh_dim)
+        # convert dict into the tuple of tuple that is hashable
+        default_sparse_shard_order = tuple(
+            tuple(item)
+            for item in (
+                [key] + value if isinstance(value, list) else [key, value]
```
ezyang: Why do this? What's so bad about having `(tensor_dim, (shard_dim,))` when only one thing is sharded on? Gives you a uniform representation.

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -1,16 +1,48 @@
...
+    _StridedShard,
     Partial,
     Placement,
     Replicate,
     Shard,
 )
 
 
+MeshDimTuple = tuple[int, ...]
+TensorDimTuple = tuple[MeshDimTuple, ...]
```
ezyang: I wouldn't name this TensorDimTuple. Because the value is set, you don't have a generic Tuple which has imbued a type on its index space. TensorDimTuple in my head is a parametric type ala `TensorDimTuple[V]`

torch/distributed/tensor/_dtensor_spec.py
wanchaol: 100% agree on it, we should not default on tensor_centric_format, it would make the sharding spec print loses Partial(sum)! Partial(sum) is critical to DTensor, so tensor_centric format should only be an option, not the default

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -29,16 +61,97 @@ class DTensorSpec:
...
+    # This order determines how tensor shards are mapped and distributed across
+    # devices. `shard_order` is a tuple of tuples of integers: in each inner
+    # tuple, the first element is the tensor dimension being sharded, and the
+    # remaining elements are the device mesh dimensions (in order) over which
+    # that tensor dimension is sharded.
+    #
+    # Example:
+    #   For a tensor of shape [8, 16] and a 3D device mesh, if dim 0 is sharded over
+    #   mesh dim 1, and dim 1 is sharded over mesh dim 0 and then mesh dim 2,
+    #   the shard_order would be: shard_order = ((0, 1), (1, 0, 2))
```
wanchaol: Is this `shard_order` representation be directly exposed to user? I don't feel like is very intuitive

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -29,16 +61,97 @@ class DTensorSpec:
...
+                placement = cast(Shard, placements[mesh_dim])
+                shard_dim = placement.dim
+                assert shard_dim >= 0, (
+                    f"Shard dim {shard_dim} in placements {placements} must be normalized"
+                )
+                if shard_dim not in tensor_dim_to_mesh_dims:
+                    tensor_dim_to_mesh_dims[shard_dim] = []
+                tensor_dim_to_mesh_dims[shard_dim].append(mesh_dim)
+        # convert dict into the tuple of tuple that is hashable
+        default_sparse_shard_order = tuple(
```
wanchaol: hmmm why not directly save the `tensor_dim_to_mesh_dims` as `shard_order` as you already computed this way 🤔

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -49,7 +162,7 @@ def __setattr__(self, attr: str, value: Any) -> None:
             # TODO: the TensorMetadata arises from
             # test/distributed/tensor/experimental/test_tp_transform.py::TensorParallelTest::test_tp_transform_e2e
             # but I actually can't reproduce it, maybe it is also a bug!
-            assert isinstance(value, (TensorMeta, TensorMetadata)), value
+            assert isinstance(value, TensorMeta | TensorMetadata), value
```
wanchaol, line 178: can we delete `test_tp_transform,py` and just unify on `TensorMeta`? iirc it was not used internally anyways

wanchaol (changes requested):
Some suggestions and questions

torch/distributed/tensor/_dtensor_spec.py
zpcore: The shard_order argument for funtion like `redistribute` will be in a dense tensor centric form. E.g., 

`shard_order = ((2, 0), (), (1), ())`, which shards a 4D tensor into the 3D mesh. Tensor dim 0 will be sharded on mesh dim 2 and 0; Tensor dim 2 will be sharded on mesh dim 1. Other tensor dims are replicated.

The API update will be shown in (6/N) PR. cc @fduwjj  and @fegin  as we discussed about this.

torch/distributed/tensor/_dtensor_spec.py
zpcore: Per offline discussion. We decide to remove tensor-centric format. For each placement, we will use the default str representation. For each shard, if one tensor dim is sharded on multiple mesh dims, we will append an order number.

torch/distributed/tensor/_dtensor_spec.py
zpcore: We need `shard_order` to be tuple in DTensorSpec to make it hashable.

torch/distributed/tensor/_dtensor_spec.py
wanchaol: For API to user, why not just use a similar format as `tensor_dim_to_mesh_dims` as you already have to computed it in that way in `_dtensor_spec.py`? I feel forcing user to specify the replicated dimensions as empty tuple are not quite necessary?

zpcore: I agree this is unnecessary to specify empty tuple for replicated dimensions. We have been discussed internally about which UX is more user friendly, and majority voted that we need to make it explicit from a PyTorch user perspective. That is to say, we need to specify for every tensor dim even though that inner tuple can be empty.

torch/distributed/tensor/_dtensor_spec.py
zpcore: Thanks for the info! I will address this in a different PR. I think the failure is from the autoparallel run.

torch/distributed/tensor/_dtensor_spec.py
zpcore: I have updated to use the NamedTuple now. I think this update should resolve all your comments about `TensorDimTuple`.

torch/distributed/tensor/_dtensor_spec.py
wanchaol: hmmm sure I am also an advocate for explicit over implicit APIs, but for this case why `dict[int, list[int]]` is not explicit compare to the UX proposed above? I think it is also every explicit, and when constructing the DTensorSpec we can explicitly check whether the tensor dim is sharded over multiple mesh dims, and if so and `shard_order` does not specify it, we explicitly error out to the user. From UX prospective I feel it is very explicit and simpler for end user to specify, I think simpler does not mean implicit. 

The problem why I feel it is not good to introduce another user facing representation like this `shard_order = ((2, 0), (), (1), ())`, is that you are essentially using THREE different kind of representations that are slightly different from each other, to represent the SAME information: 1. user facing shard order 2. tensor_dim_to_mesh_dims 3. shard_order in DTensorSpec. This just turn things into more complicated and error prone direction because we need to do conversions between these three representation (1 -> 2 -> 3), and everytime there needs a conversion, it adds a concept overhead and there could be bugs introduced and creates maintenance burden..

zpcore: OK, let's consolidate on using 2 representations: 
1. `dict[int, list[int | str]]` for user input. This will be used in `redistribute` and `distribute_tensor`;
2. `ShardOrder = tuple[ShardOrderEntry, ...]` (I just update in this PR), which is converted from user specified `dict[int, list[int]]` and is hashable for internal use like in DTensorSpec. 

The user interface will be show in 6/N PR. For this PR I have updated with newly defined type `ShardOrder`.  PTAL.

ezyang: @zpcore I want to add that in @tianyu-l's proposal, the user input can also take `dict[int, list[str]]`, where str is the mesh dim axis name

ezyang: @wanchaol re

> This just turn things into more complicated and error prone direction because we need to do conversions between these three representation (1 -> 2 -> 3), and everytime there needs a conversion, it adds a concept overhead and there could be bugs introduced and creates maintenance burden..

So, I think it is legitimate to complain that there are too many ways to specify shard order in the user API and we should accept less to reduce burden on people reading. But I disagree about the claim about internal implementation burden. No matter how many shard order representations we accept, as soon as they go into DTensor we should canonicalize them into a single, unified internal representation and use that uniformly everywhere on the inside. (And the internal representation has design pressure different from the frontend UX; for example, as @zpcore points out, whatever we use as internal rep needs to be hashable, which rules out anything involving dict.)

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -1,16 +1,54 @@
...
+# A tuple of ShardOrderEntry, one per sharded tensor dimension
+#
+# Example:
+#   shard_order = (
+#       ShardOrderEntry(tensor_dim=0, mesh_dims=(1,)),
+#       ShardOrderEntry(tensor_dim=2, mesh_dims=(0, 3)),
+#   )
+#   This means:
+#     - Tensor dimension 0 is sharded on mesh dimension 1
+#     - Tensor dimension 2 is sharded on mesh dimension 0 first, then mesh dimension 3
```
ezyang, line 51: Guaranteed to be sorted on tensor_dim?

ezyang: Also, guaranteed to NOT have a ShardOrderEntry when mesh_dims is empty, right?

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -1,16 +1,54 @@
...
+    Examples:
+        >>> # Tensor dim 1 sharded across mesh dim 2, then mesh dim 0
+        >>> ShardOrderEntry(tensor_dim=1, mesh_dims=(2, 0))
+
+        >>> # Tensor dim 0 sharded only on mesh dim 1
+        >>> ShardOrderEntry(tensor_dim=0, mesh_dims=(1,))
+    """
+
+    tensor_dim: int
+    mesh_dims: tuple[int, ...]
```
ezyang: This is guaranteed to be non-empty right?

torch/distributed/tensor/_dtensor_spec.py
ezyang: If this is too much CPU overhead we have much bigger problems lol (cc @swolchok)

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -103,18 +230,96 @@ def __str__(self) -> str:
...
+            )
+            if not is_increasing:
+                return False
+        return True
+
+    @staticmethod
+    def format_shard_order_str(
+        placements: tuple[Placement, ...],
+        shard_order: Optional[ShardOrder] = None,
+    ) -> str:
```
ezyang, line 262: OOC, why does this live on DTensorSpec? Is this just intended to be private?

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -103,18 +230,96 @@ def __str__(self) -> str:
...
+                                out_str += f"{placement}[{mesh_dims.index(mesh_dim)}]"
+                            else:
+                                # no need to show device order if the tensor dim is
+                                # only sharded in one mesh dim
+                                out_str += str(placement)
+                            break
+                else:
+                    out_str += str(placement)
+            else:
+                out_str += str(placement)
```
ezyang, line 323: note to self: str is the short rendering

torch/distributed/tensor/_dtensor_spec.py
zpcore: Yes, this is enforced in `_verify_shard_order`: 
1) tensor dim must be increasing in ShardOrder;
2) mesh_dims can not be None in ShardOrderEntry;
3) tensor dims are normalized.

Even when user manual overwrite tensor._spec.shard_order, this will still be triggered through `__setattr__`.

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -29,16 +67,103 @@ class DTensorSpec:
...
+                return ()
+            if isinstance(placements[mesh_dim], Shard):
+                placement = cast(Shard, placements[mesh_dim])
+                shard_dim = placement.dim
+                assert shard_dim >= 0, (
+                    f"Shard dim {shard_dim} in placements {placements} must be normalized"
+                )
+                if shard_dim not in tensor_dim_to_mesh_dims:
+                    tensor_dim_to_mesh_dims[shard_dim] = []
+                tensor_dim_to_mesh_dims[shard_dim].append(mesh_dim)
```
ezyang, line 118: just FYI (you don't have to change code), defaultdict is good here!

torch/distributed/tensor/_dtensor_spec.py
zpcore: The type definition doesn't enforce, but we have `_verify_shard_order ` as in the above comment.

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -29,16 +67,103 @@ class DTensorSpec:
...
+        Returns a ShardOrder where each ShardOrderEntry maps a tensor dimension
+        to the mesh dimensions it's sharded on, in left-to-right order.
+        """
+        # follow default left-to-right device order if shard_order is not specified
+        tensor_dim_to_mesh_dims: dict[int, list[int]] = {}
+        mesh_ndim = len(placements)
+        for mesh_dim in range(0, mesh_ndim):
+            # shard_order doesn't work with _StridedShard
+            if isinstance(placements[mesh_dim], _StridedShard):
+                return ()
```
ezyang, line 111: This is suspicious, because when something doesn't work it is far better to error rather than just silently return incorrect data.

torch/distributed/tensor/_dtensor_spec.py
zpcore: I want to preserve the original placement representation so just use `str()` here.

torch/distributed/tensor/_dtensor_spec.py
zpcore: I feel `format_shard_order_str` is related to the DTensorSpec. Move it out of the DTensorSpec class should also be fine.

torch/distributed/tensor/_dtensor_spec.py
zpcore: We can't error out here, or `_StridedShard` will always fail because `compute_default_shard_order` is executed in __post_init__.

torch/distributed/tensor/_dtensor_spec.py
wanchaol: Yeah I agree! For internal DTensorSpec it totally make sense to use different (unified) impl given the serialization consideration and we should use it everywhere else once it got canonicalized into DTensorSpec's internal named tuple representation. What I was mainly referring to is that supporting different shard order representation requires implementing different parsing logic from user API inputs to our internal representation, which seems not necessary given that a single user facing API would also reduce burden on ppl reading like you mentioned.

zpcore: Please check #165092 for the user API usage.

test/distributed/tensor/test_dtensor.py
```diff
@@ -1055,5 +1061,184 @@ def test_split_tensor_1D(self) -> None:
...
+                ),
+            ),
+            """S(2)S(1)S(0)""",
+        )
+        self.assertExpectedInline(
+            DTensorSpec.format_shard_order_str(
+                (Shard(1), Shard(1), Shard(1)),
+                (ShardOrderEntry(tensor_dim=1, mesh_dims=(2, 0, 1)),),
+            ),
+            """S(1)[1]S(1)[2]S(1)[0]""",
```
wanchaol, line 1090: hmm why this prints as 1, 2, 0 while the mesh dimes specified in ShardOrderEntry is (2, 0, 1)?

wanchaol (approved):
lgtm, one question about the test behavior

test/distributed/tensor/test_dtensor.py
zpcore: This is expected. `mesh_dims=(2, 0, 1)` means we shard the tensor dim 1 on mesh dim 2 first, then mesh dim 0, then 1. When reflected in **placements** it should be `S(1)[1]S(1)[2]S(1)[0]`, where you can see mesh dim 2 is `S(1)[0]`, means it is been sharded on first, and so on.

torch/distributed/tensor/_dtensor_spec.py
ezyang: I was hoping to suggest we simply just error when a StridedShard is on the same dimension with another Shard but that is how it is used... thinking of what I would suggest here...

torch/distributed/tensor/_dtensor_spec.py
ezyang: OK. This is just good to also put in the comments.

torch/distributed/tensor/_dtensor_spec.py
zpcore: Fixed!

torch/distributed/tensor/_dtensor_spec.py
zpcore: marked resolved after introducing `ShardOrder` type.

torch/distributed/tensor/_dtensor_spec.py
zpcore: marked resolved after introducing `ShardOrder` type.

torch/distributed/tensor/_dtensor_spec.py
zpcore: marked resolved after introducing `ShardOrder` type.

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -31,16 +70,101 @@ class DTensorSpec:
...
+            prev_tensor_dim = tensor_dim
+            total_shard += len(mesh_dims)
+            for mesh_dim in mesh_dims:
+                assert 0 <= mesh_dim < len(self.placements), (
+                    f"shard_order {shard_order} has invalid mesh dim {mesh_dims}"
+                )
+                assert self.placements[mesh_dim] == Shard(tensor_dim), (
+                    f"placement[{mesh_dim}] doesn't have a matching shard in shard_order"
+                )
+        assert total_shard == sum(1 for p in self.placements if isinstance(p, Shard))
```
ezyang, line 153: If a user can trigger this assert, it shouldn't be an assert, it should be regular error with an appropriate error message. It's hard to say because this is an argument to DTensorSpec and I don't know if we are going to do separate validation before we run DTensorSpec.

torch/distributed/tensor/_dtensor_spec.py
ezyang: This kind of thing is good to have in the comments themselves; even if they're enforced in `_verify_shard_order` the comment here is the first place someone is going to look at them.

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -31,16 +70,101 @@ class DTensorSpec:
...
+            # shard_order doesn't work with _StridedShard
+            if isinstance(placements[mesh_dim], _StridedShard):
+                return ()
+            if isinstance(placements[mesh_dim], Shard):
+                placement = cast(Shard, placements[mesh_dim])
+                shard_dim = placement.dim
+                assert shard_dim >= 0, (
+                    f"Shard dim {shard_dim} in placements {placements} must be normalized"
+                )
+                tensor_dim_to_mesh_dims[shard_dim].append(mesh_dim)
```
ezyang, line 118: It is suspicious that you only test for Shard and StridedShard. What about other placement types?

torch/distributed/tensor/_dtensor_spec.py
```diff
@@ -31,16 +70,101 @@ class DTensorSpec:
...
+            for key, value in sorted(tensor_dim_to_mesh_dims.items())
+            if value
+        )
+        return default_shard_order
+
+    def _verify_shard_order(self, shard_order: ShardOrder) -> None:
+        """Verify that the shard_order is valid and matches the placements."""
+        total_shard = 0
+        if any(isinstance(p, _StridedShard) for p in self.placements):
+            return
```
ezyang, line 132: Shouldn't you instead test that the shard order is exactly `()` per `compute_default_shard_order`?

PR #164808, opened 2025-10-07
outline various stages from aot stage2 compile

avikchaudhuri:
Splits the training and inference paths for aot stage2 compile.
1. Split `aot_stage2_autograd` into `_aot_stage2a_partition`, `_aot_stage2b_fw_compile` and `_aot_stage2b_bw_compile`, and rest.
2. Split `aot_stage2_inference` into `_aot_stage2b_inference_compile` and rest.
I'm leaving these as functions with underscore names since the I/O interfaces and the exact boundaries of these splits are somewhat in the air.

Differential Revision: D84028203

### DIFF
--- a/torch/_functorch/_aot_autograd/graph_compile.py
+++ b/torch/_functorch/_aot_autograd/graph_compile.py
@@ -84,6 +84,7 @@ from .schemas import (
     FlatFn,
     FxValue,
     MutationType,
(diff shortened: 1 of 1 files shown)


SherlockNoMad (approved):
signed off as this is refactor only, no behavior change on API.

PR #164809, opened 2025-10-07
[1/N] Don't return values in void functions

cyyever:
This PR fixes returning values in void C++ functions.


janeyx99 (approved):
ok

PR #164810, opened 2025-10-07
Add interface for passing in constant names using FQN names

yushangdi:
As title. We need an API to pass in the names using FQN names because this is the name that aligns with module state dict names

Will be tested together with https://github.com/pytorch/pytorch/pull/164811 when https://github.com/pytorch/pytorch/pull/164811 is ready to land.


desertfire (commented):
Is it possible to add a separate test instead of waiting for the other PR?

PR #164819, opened 2025-10-07
[user-streams] Fix stream graph output semantics

mlazos:
Preivously, we would stash a single stream value we constructed at trace time in a global and return the same value from repeated calls to the graph.

With this PR, we construct the stream value in advance, reference the constructed value in the graph via the lookup table, and if that value is returned as an output, read the value from the lookup table and return it (in bytecode, not as a graph output, since we don't support arbitrary stream outputs).

* #166472
* #166471
* #162905
* #164523
* #165356
* #165212
* #165211
* __->__ #164819


torch/_dynamo/bytecode_analysis.py
```diff
@@ -228,6 +228,7 @@ def exn_tab_jump(self, depth: int) -> None:
 
 def stacksize_analysis(instructions: list["Instruction"]) -> Union[int, float]:
     assert instructions
+    print(instructions)
```
mlazos: Remove

anijain2305 (approved):

PR #164820, opened 2025-10-07
[3/N] [DTensor device order] Make some placement type class method static

zpcore:
Some methods in `Placement` class can be exposed as static.

Those method should be useful w/o initializing the object. E.g., when we `distribute_tensor` from normal tensor, we may want:
```
local_tensor = Shard.shard_tensor(tensor_dim, local_tensor, device_mesh, mesh_dim,)
```

* #165197
* #165092
* #164902
* #164821
* __->__ #164820
* #164806
* #164797


XilunWu (approved):

fduwjj (approved):
No need to do it in this PR, but if we can add some comments to this file that will be really good. IIUC, most of collective calls happen in this `placement_types.py` so when users start to investigate DTensor collective issues, for example, integration with torchcomm, people know where to start.

wanchaol (changes requested):
Can you give some motivations on this change (in the PR summary)?

The consequence of this change is that it exposes these methods as public methods in the Placement, so we would need to maintain the BC of these APIs going forward and could not change them easily under the hood. I would not prefer to expose them unless there's a strong motivation

torch/distributed/tensor/placement_types.py
```diff
@@ -85,31 +86,47 @@ def _split_tensor(
...
-                shard = pad_tensor(shard, self.dim, pad_size)
+                pad_size = full_chunk_size - shard.size(dim)
+                shard = pad_tensor(shard, dim, pad_size)
                 pad_sizes.append(pad_size)
             if contiguous:
                 shard = shard.contiguous()
             shard_list.append(shard)
         return shard_list, pad_sizes
 
+    def _split_tensor(
```
wanchaol, line 114: don't have to be in this PR: maybe just merge these two methods to use the static method everywhere

wanchaol (approved):

PR #164836, opened 2025-10-07
[CUDA][Thor] Enable CUTLASS matmuls on Thor

Aidyn-A:
This PR enables special matmuls on Thor devices. This includes row-wise scaled matmul on `fp8` and group gemm on `bfloat16`.

### DIFF
--- a/aten/src/ATen/native/cuda/GroupMM.cu
+++ b/aten/src/ATen/native/cuda/GroupMM.cu
@@ -346,8 +346,9 @@ void dispatch_bf16_grouped_kernel_on_tile_size(
   bool small = (M <= 128 || N <= 128);
   cudaDeviceProp* properties = at::cuda::getCurrentDeviceProperties();
   const bool sm10x = properties != nullptr && properties->major == 10;
+  const bool sm11x = properties != nullptr && properties->major == 11;
 
-  if (sm10x) {
+  if (sm10x || sm11x) {
     if (small){
       bf16bf16_grouped_gemm_impl_sm90_sm100<
         cutlass::arch::Sm100,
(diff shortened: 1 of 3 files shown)


Skylion007 (dismissed):

aten/src/ATen/native/cuda/GroupMM.cu
```diff
@@ -346,8 +346,9 @@ void dispatch_bf16_grouped_kernel_on_tile_size(
   bool small = (M <= 128 || N <= 128);
   cudaDeviceProp* properties = at::cuda::getCurrentDeviceProperties();
   const bool sm10x = properties != nullptr && properties->major == 10;
+  const bool sm11x = properties != nullptr && properties->major == 11;
 
-  if (sm10x) {
+  if (sm10x || sm11x) {
```
Skylion007, line 351: Should this also be enabled on sm120x or nah?

aten/src/ATen/native/cuda/GroupMM.cu
```diff
@@ -346,8 +346,9 @@ void dispatch_bf16_grouped_kernel_on_tile_size(
...
     if (small){
       bf16bf16_grouped_gemm_impl_sm90_sm100<
         cutlass::arch::Sm100,
```
Skylion007, line 354: Wait, don't we need a seperate instationation here with 
```
        cutlass::arch::Sm110,
```
?

aten/src/ATen/native/cuda/GroupMM.cu
Aidyn-A: As far as I know, sm100 and sm110 are compatible, but sm120 is completely different from those two.

aten/src/ATen/native/cuda/GroupMM.cu
Aidyn-A: Nope, it does not exist in CUTLASS. The `cutlass::arch::Sm101` technically exist: https://github.com/NVIDIA/cutlass/blob/a2439551c765c5393aebe557ee75d3a0412d2211/include/cutlass/arch/arch.h#L104-L106
but it is it not used anywhere in CUTLASS. I was not able to compile anything with it.

ngimel (approved):

PR #164838, opened 2025-10-07
Fix grad_outputs type hints and parameter descriptions

### LINKED ISSUE #164298: Wrong type hint for `grad_outputs` in `torch.autograd.grad`
`grad_outputs` is type hinted as `Optional[_TensorOrTensors]`, and `_TensorOrTensors` is defined as `_TensorOrTensors: TypeAlias = Union[Tensor, Sequence[Tensor]]`.

However the documentation says:
> None values can be specified for scalar Tensors or ones that don't require grad.

I don't think it's possible to have any output that doesn't require grad (so this sentence may also need to be updated, if I understand correctly), but it's perfectly possible to have scalar outputs, and providing a `None` grad_output for each of them works just fine.

So I think the type hint should rather be `Optional[_TensorOrOptionalTensors]`, with `_TensorOrOptionalTensors` defined as `_TensorOrOptionalTensors: TypeAlias = Union[Tensor, Sequence[Optional[Tensor]]]`

This also applies to a few other places:
* The `grad_tensors` parameter of `torch.autograd.backward`
* The (deprecated) `grad_variables` parameter of `torch.autograd.backward`
* The `grad_outputs` parameter of `torch.autograd.gradgradcheck`
* The `tensors` parameter of `_tensor_or_tensors_to_tuple` (a helper function in `torch/autograd/__init__.py` that is only used on `grad_outputs` or `grad_tensors`).)

What do you think? I can try to make a PR for this if you agree.

ValerianRey:
* Add _TensorOrOptionalTensors type alias in types.py

* Change type hint to Optional[_TensorOrOptionalTensors] in the following:
  - grad_tensors and grad_variables in autograd.backward
  - grad_outputs in autograd.grad
  - tensors in autograd._tensor_or_tensors_to_tuple
  - grad_outputs in autograd.gradgradcheck

* Update parameter description to match the new type hint for grad_outputs in autograd.grad and in gradgradcheck. Other parameter descriptions were already ok.

Fixes #164298


soulitzer (approved):
Thanks!

PR #164848, opened 2025-10-07
[feature][dtensor] support nn.Module inputs containing DTensor param

RohitRathore1:
[feature][dtensor] support nn.Module inputs containing DTensor parameters in local_map

Fixes #163692

This patch adds support for `nn.Module` inputs containing DTensor parameters in the `local_map` function. Previously, users had to set `in_placements=None` for `nn.Module` arguments, which was not ergonomic. So now converts DTensor parameters in the module to local tensors before calling the user function, and supports redistribution and gradient placements like regular DTensor arguments.

```python
import torch
from torch.distributed.tensor.placement_types import Replicate, Shard
from torch.distributed.tensor import DTensor, distribute_tensor
from torch.distributed.tensor.experimental._func_map import local_map

world_size = 2
fake_store = FakeStore()
torch.distributed.init_process_group(
    "fake", store=fake_store, rank=0, world_size=world_size
)
mesh = torch.distributed.device_mesh.init_device_mesh("cuda", (world_size,))

print("Using Replicate placements")
print("=" * 60)

@local_map(
    out_placements=[Replicate()],
    in_placements=(
        [Replicate()],
        [Replicate()]
    ),
    redistribute_inputs=True,
    device_mesh=mesh,
)
def fn_replicate(linear, y):
    return linear(y)

linear = torch.nn.Linear(8, 16, bias=False).cuda()
linear.weight = torch.nn.Parameter(
    distribute_tensor(linear.weight, mesh, [Replicate()])
)

y = torch.randn(4, 8).cuda()
result = fn_replicate(linear, y)

print(f"Input shape: {y.shape}")
print(f"Weight shape: {linear.weight.shape}")
print(f"Output shape: {result.shape}")
print(f"Output is DTensor: {isinstance(result, DTensor)}")
if isinstance(result, DTensor):
    print(f"Output placements: {result.placements}")

print("\nbefore this fix, in_placements had to be None")
print("=" * 60)

@local_map(
    out_placements=[Replicate()],
    in_placements=(
        None,
        [Replicate()]
    ),
    redistribute_inputs=True,
    device_mesh=mesh,
)
def fn_old_way(linear, y):
    return linear(y)

linear2 = torch.nn.Linear(8, 16, bias=False).cuda()

result2 = fn_old_way(linear2, y)
print(f"Old way (in_placements=None) still works")
print(f"Output shape: {result2.shape}")

print("\n" + "=" * 60)
torch.distributed.destroy_process_group()
```

```
Using Replicate placements
============================================================
Input shape: torch.Size([4, 8])
Weight shape: torch.Size([16, 8])
Output shape: torch.Size([4, 16])
Output is DTensor: True
Output placements: (Replicate(),)

before this fix, in_placements had to be None
============================================================
Old way (in_placements=None) still works
Output shape: torch.Size([4, 16])
```                                                                                                                                                          

cc @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @aditvenk @xmfan @weifengpy @H-Huang

### DIFF
--- a/test/distributed/tensor/experimental/test_local_map.py
+++ b/test/distributed/tensor/experimental/test_local_map.py
@@ -494,5 +494,6 @@ class TestLocalMap(DTensorTestBase):
         Y_expected = torch.nn.functional.linear(X, weight)
         self.assertEqual(Y_dt.full_tensor(), Y_expected)
 
+
 if __name__ == "__main__":
     run_tests()
--- a/test/distributed/tensor/experimental/test_local_map.py
+++ b/test/distributed/tensor/experimental/test_local_map.py
@@ -426,6 +426,73 @@ class TestLocalMap(DTensorTestBase):
         # output lives in mesh_2d
         self.assertEqual(Y_dt.device_mesh, mesh_2d)
 
+    @with_comms
+    def test_local_map_with_nn_module(self):
+        """
+        test for issue #163692: local_map should support nn.Module inputs
+        containing DTensor parameters.
+        """
+        device_mesh = init_device_mesh(
+            device_type=self.device_type, mesh_shape=(self.world_size,)
+        )
+
+        W = torch.randn(16, 8, device=self.device_type, requires_grad=False)
+        W_dt = distribute_tensor(W, device_mesh, replicate)
+
+        X = torch.randn(4, 8, device=self.device_type, requires_grad=False)
+        X_dt = distribute_tensor(X, device_mesh, replicate)
+
+        def matmul_fn(weight, x):
+            # weight is (16, 8), x is (4, 8)
+            # result should be (4, 16)
+            return torch.mm(x, weight.t())
+
+        local_matmul = local_map(
+            matmul_fn,
+            out_placements=replicate,
+            in_placements=(replicate, replicate),
+            device_mesh=device_mesh,
+        )
+
+        Y_dt = local_matmul(W_dt, X_dt)
+
+        for placement in Y_dt.placements:
+            self.assertTrue(placement.is_replicate())
+
+        Y_expected = torch.mm(X, W.t())
+        self.assertEqual(Y_dt.full_tensor(), Y_expected)
+
+    @with_comms
+    def test_local_map_with_nn_module_actual(self):
+        device_mesh = init_device_mesh(
+            device_type=self.device_type, mesh_shape=(self.world_size,)
+        )
+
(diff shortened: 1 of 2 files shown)


ezyang (changes requested):
The heart is in the right place, but the inplace module modification is not

torch/distributed/tensor/experimental/_func_map.py
```diff
@@ -62,6 +62,12 @@ def local_map(
             ``redistribute_inputs`` is ``False``, an exception will be raised. Otherwise if
             ``redistribute_inputs`` is ``True``, the argument will be first redistributed to
             the required sharding placements before passing its local tensor to ``func``.
+
+            For :class:`torch.nn.Module` inputs containing :class:`DTensor` parameters,
+            the placements will be applied to all DTensor parameters in the module. The
```
zpcore: Shall we mention placements doesn't apply to submodule?

test/distributed/tensor/experimental/test_local_map.py
```diff
@@ -426,6 +426,74 @@ def test_multi_mesh_inputs(self):
         # output lives in mesh_2d
         self.assertEqual(Y_dt.device_mesh, mesh_2d)
 
+    @with_comms
+    def test_local_map_with_nn_module(self):
```
zpcore, line 430: This is no nn.Module arg in this test. Do we need this test?

zpcore (commented):
Overall LGTM! Thanks

Can you also test that the module parameters placements can be correctly restored and the grad placement?

test/distributed/tensor/experimental/test_local_map.py
RohitRathore1: thanks I have enhanced it and added `test_local_map_with_nn_module_grad_placements` to test that `in_grad_placements` works correctly with `nn.Module` inputs

zpcore:
> I'm confused. Are you saying that one nn.Module input corresponds to one placement? Can't an nn.module have more than one tensor on it, which could each have different placements?

I guess majority of the use cases will leave `in_placements` None for nn.Module :)

wconstab:
sorry- more specific question-
how important is it to preserve BC on the way module is passed to local_map?

if people think passing modules is useful but it is not being used yet, then we can have more freedom to design a good API.

PR #164855, opened 2025-10-07
Call internal log_compilation_event if it exists

masnesral:
* __->__ #164855

Summary: For internal conda on mast jobs, call the internal version of log_compilation_event if it exists.

Test Plan: Ran a simple test job that just calls the API: https://fburl.com/scuba/dynamo_compile/dqx8d10g

### DIFF
--- a/torch/_utils_internal.py
+++ b/torch/_utils_internal.py
@@ -13,6 +13,13 @@ import torch
 from torch._strobelight.compile_time_profiler import StrobelightCompileTimeProfiler
 
 
+try:
+    from pt2_internal import (  # type: ignore[import-not-found]
+        log_compilation_event as internal_log_compilation_event,
+    )
+except ImportError:
+    internal_log_compilation_event = None
+
 _T = TypeVar("_T")
 _P = ParamSpec("_P")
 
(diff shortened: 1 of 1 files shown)


torch/_utils_internal.py
```diff
@@ -13,6 +13,13 @@
 from torch._strobelight.compile_time_profiler import StrobelightCompileTimeProfiler
 
 
+try:
```
masnesral, line 16: I'm not proud of this. Lemme know if you know of a better way. The idea is that we use the internal logger automatically if it happens to be available in the conda env, i.e., without requiring users call some registration function or something.

c00w: Loose opinion of - make a register function, have the internal code call it. (provides a better api than, there is this magic import.

c00w (approved):

torch/_utils_internal.py
masnesral: Right - it's just not clear to me at this point if there's going to be a central place to put the registration or if we're going to have to chase a bunch individual usages to make sure they register. I'll land the hack and hope to improve once that plan is more clear.

PR #164857, opened 2025-10-07
[export] Fix weight sharing when there is no complete tensor

yiming0416:
Summary: As titled.

Test Plan: CI

Differential Revision: D84079625


yushangdi (approved):
LGTM. A more fundamental fix might be to get the complete underlying storage and not use the complete tensor.

PR #164861, opened 2025-10-07
[CI] Run cpp tests on windows in one run_tests call

clee2000:
The windows cpp tests take ~1 hour according to logs.  Each has run_test called on them individually, so I tried batching them together so it's just one run_test call for all of them.  I believe it now takes 30min.  I turned off TD since I don't think cpp tests are included in TD stuff.  

As always with batch, I'm not sure if the errorlevel/error surfacing stuff is correct

This code is written with a lot of help from chatgpu and copilot


huydhn (approved):
CI passes then LGTM!  i have no memory of this place :D

Issue #163343, opened 2025-09-19
Stable headers shouldn't use anonymous namespaces


lw:
### 🐛 Describe the bug

Some of the headers of the stable API have anonymous namespaces inside of them:
https://github.com/pytorch/pytorch/blob/264e7f68a095b5b2f10e7ca66a3bbb16f6e56e28/torch/csrc/stable/stableivalue_conversions.h#L11-L13

This is considered an [anti-pattern](https://clang.llvm.org/extra/clang-tidy/checks/misc/use-anonymous-namespace.html), and it has the side effect that the global `to` and `from` names are overloaded in every file that includes those headers.

I'm not sure whether I fully grasp the original reason for using an anonymous namespace, but I wonder whether an anonymous namespace _inside_ a named namespace would work. Something like this:
```cpp
namespace torch::stable {
namespace {
template <typename T>
StableIValue from(T val);
template <typename T>
T to(StableIValue val);
}
}
```

### Versions

PyTorch nightly from 2025-09-05

malfet:
@lw this is by design, so that if one has to change stable ABI implementation for some reason it will collide with older implementation of the same interface from another library.. But may be one can workaround with inline namespaces and bump the version every time something changes there

mikaylagawarecki:
> @mikaylagawarecki has been working on the versioning aspect with inline namespaces, which should resolve this.

fwiw, I do not think my work on versioning addresses this specific issue

janeyx99:
Oh, maybe I misunderstood. My thought was that we use anonymous namespaces to prevent collisions for when we change the API and two different libraries import different versions, and so when we add the versioned namespaces we would ensure that different versions of an API would be deduped.

Either way, I agree that we should:
- reach a better conclusion on whether we need the anonymous namespace at all -- i'm currently looking into whether this could be related https://github.com/pytorch/pytorch/issues/164742
- apply that decision consistently; today it is indeed weird that some of our stable APIs have this vs not.

PR #164882, opened 2025-10-07
Use inline instead of anon namespace for stableivalue from/to

janeyx99:
Fixes https://github.com/pytorch/pytorch/issues/163343.

After some consideration, I propose we remove the anonymous namespace around from/to in favor of:
1. Adding inline to the function implementations, assuming that they will not change in the near future
2. If we decide to change them, we will wrap the code in inline versioned namespaces such that the implementations within any versioned namespace will be guaranteed identical.

Note that:
- We eventually intend to abstract away usage of `from`/`to` (related: @lw's TORCH_BOX work)
- The from/to implementations are now powered through class template specializations, where adding a specialization does not change the from/to signatures.

I do plan to deprecate top-level from/to in favor of torch::stable::details::from/to consequently. This way we can stop polluting the global namespace.

* __->__ #164882


lw (approved):
Great! Thanks!

albanD (approved):

PR #164890, opened 2025-10-07
[Replicate][Test] tests that pp model grads are the same as single-device model grads

anshul-si:
**Summary:** Created a test so that we can verify that a model that has been pipelined + replicated has the same gradients as a reference model. To do this, I mapped the layers and their parameters in each partial model to the original full model and then compared the gradients. 
**Test Case**
1. pytest test/distributed/_composable/test_composability/test_pp_composability.py -k test_replicate_pp_grads

* __->__ #164890

### DIFF
--- a/test/distributed/_composable/test_composability/test_pp_composability.py
+++ b/test/distributed/_composable/test_composability/test_pp_composability.py
@@ -604,6 +604,284 @@ class ComposabilityTest(MultiProcessTestCase):
 
         torch.distributed.destroy_process_group()
 
+    @requires_accelerator_dist_backend(["nccl", "xccl"])
+    @skip_if_lt_x_gpu(8)
+    @skip_but_pass_in_sandcastle_if(
+        not TEST_MULTIGPU and not TEST_XPU, "Test requires 8+ GPUs"
+    )
+    @parametrize(
+        "ScheduleClass",
+        [
+            ScheduleGPipe,
+            Schedule1F1B,
+            ScheduleInterleaved1F1B,
+            ScheduleLoopedBFS,
+            ScheduleInterleavedZeroBubble,
+        ],
+    )
+    def test_replicate_pp_grads(self, ScheduleClass):
+        _device_raii = torch.device(device_type, self.device)
+        torch.accelerator.set_device_index(self.device)
+        store = torch.distributed.FileStore(self.file_name, self.world_size)
+        torch.distributed.init_process_group(
+            backend=backend,
+            store=store,
+            rank=self.rank,
+            world_size=self.world_size,
+        )
+        dim = 8
+        pp_size = 2
+        num_microbatches = 8
+        replicate_size = self.world_size // (pp_size)
+        device_mesh = init_device_mesh(
+            device_type,
+            mesh_shape=(replicate_size, 1, pp_size),
+            mesh_dim_names=("replicate", "shard", "pp"),
+        )
+        torch.manual_seed(42)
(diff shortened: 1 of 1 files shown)


test/distributed/_composable/test_composability/test_pp_composability.py
```diff
@@ -604,6 +604,284 @@ def apply_same_precision(partial_model):
 
         torch.distributed.destroy_process_group()
 
+    @requires_accelerator_dist_backend(["nccl", "xccl"])
+    @skip_if_lt_x_gpu(8)
```
H-Huang: this test works with just 4 GPUs

test/distributed/_composable/test_composability/test_pp_composability.py
```diff
@@ -604,6 +604,284 @@ def apply_same_precision(partial_model):
...
+    @skip_but_pass_in_sandcastle_if(
+        not TEST_MULTIGPU and not TEST_XPU, "Test requires 8+ GPUs"
```
H-Huang, line 609: same comment as above, edit message

test/distributed/_composable/test_composability/test_pp_composability.py
```diff
@@ -604,6 +604,284 @@ def apply_same_precision(partial_model):
...
+        [
+            ScheduleGPipe,
+            Schedule1F1B,
+            ScheduleInterleaved1F1B,
+            ScheduleLoopedBFS,
+            ScheduleInterleavedZeroBubble,
+        ],
+    )
+    def test_replicate_pp_grads(self, ScheduleClass):
+        _device_raii = torch.device(device_type, self.device)
```
H-Huang: confused how this is used? My guess based on the variable name is keep a reference to the device to keep it alive, but not sure why it is needed. If you can find out, i am curious!

test/distributed/_composable/test_composability/test_pp_composability.py
```diff
@@ -604,6 +604,284 @@ def apply_same_precision(partial_model):
...
+                            rtol=1e-4,
+                            atol=1e-5,
+                        )
+                    else:
+                        assert pipeline_parameter.grad is None
+
+        pipeline_model_parameter_dict = {}
+
+        # Attach to a schedule
+        if issubclass(ScheduleClass, PipelineScheduleSingle):
```
H-Huang, line 740: dang, this if-else needed for single stage schedules vs multistage schedules is so ugly, i should fix it 😭

test/distributed/_composable/test_composability/test_pp_composability.py
```diff
@@ -604,6 +604,284 @@ def apply_same_precision(partial_model):
...
+                # For non-last stages, still run ref model to generate gradients
+                simulate_stage_forward_backward(ref_model, inputs, None)
+                simulate_all_reduce_grads(ref_model, dp_group)
+
+            # Step optimizers
+            for optimizer in optimizers:
+                optimizer.step()
+            ref_optimizer.step()
+
+            # torch.distributed.breakpoint()
```
H-Huang: remove

H-Huang (approved):
LGTM, just some small comments / questions. Feel free to land it when you are ready!

test/distributed/_composable/test_composability/test_pp_composability.py
anshul-si: I believe its intended purpose was to do what you're saying. Turns out that its not needed in this instance so I will get rid of them.

PR #164895, opened 2025-10-07
Remove shared_ptr from MHAGraphCache

lakshayg:
This commit makes several cleanup changes to MHA.cpp, the main
one of which is removal of shared_ptr from MHAGraphCache as the
cache does not actually intend to share ownership. The changes are:

1. Remove shared_ptr from MHAGraphCache
2. Remove template arguments from MHAGraphCache
3. Remove unnecessary optional<shared_ptr<...>> vars
4. Change some functions with auto return type to the actual type

### DIFF
--- a/aten/src/ATen/native/cudnn/MHA.cpp
+++ b/aten/src/ATen/native/cudnn/MHA.cpp
@@ -341,16 +341,22 @@ struct MHACacheKeyWrapper : ParamsWrapper<MHAParams> {
   }
 };
 
-template <typename T, typename KeyType>
 struct MHAGraphCache {
-  std::unordered_map<KeyType, T, ParamsWrapperHash<KeyType>> engine_cache;
+  using KeyType = MHACacheKeyWrapper;
+  using ValueType = std::unique_ptr<fe::graph::Graph>;
+  using MapType =
+      std::unordered_map<KeyType, ValueType, ParamsWrapperHash<KeyType>>;
+  using iterator = typename MapType::iterator;
+  using const_iterator = typename MapType::const_iterator;
+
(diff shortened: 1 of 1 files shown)


eqy (approved):

aten/src/ATen/native/cudnn/MHA.cpp
```diff
@@ -363,15 +369,19 @@ struct MHAGraphCache {
...
-  void update(const KeyType& key, T& results) {
-    engine_cache.insert_or_assign(key, std::move(results));
+  const_iterator end() const {
+    return engine_cache.end();
+  }
+
+  template <typename... Args>
+  std::pair<iterator, bool> emplace(Args&&... args) {
+    return engine_cache.emplace(std::forward<Args>(args)...);
   }
```
Skylion007, line 385: Should this actually be try_emplace for added efficiency?

lakshayg: The only difference between the two is that this might construct the `std::pair<const Key, Value>` object even if it might end up not being inserted. There might be slight improvement to the efficiency however there are some mitigating factors here:

1. Constructing `std::pair<MHACacheKeyWrapper, unique_ptr<...>>` should not be too bad as none of the constructors looks expensive
2. ~The way this function is currently used (only called if element does not exist in the map), it is guaranteed that the element will need to be inserted so the pair has to be constructed.~

I don't have a strong opinion on emplace vs try_emplace so I am happy to change if you prefer.

lakshayg: I have updated the PR

PR #164900, opened 2025-10-08
[fx] Treat _functorch module as not torch for global name generation

anijain2305:
* #164776
* __->__ #164900
* #164678

Saw it while trying to jam `compile_fx` output back into the Fx graph.


bdhirsh (approved):

xmfan (approved):
pain

avikchaudhuri (approved):

jansel (changes requested):
Test failures?

torch/fx/graph.py
```diff
@@ -121,7 +121,7 @@ def _snake_case(s: str) -> str:
 
 # starts with torch but does not start with torch._dynamo. or torch._inductor.
 _torch_but_not_dynamo = re.compile(
-    r"^torch(?:\.(?!_dynamo\.|_inductor\.)[^.]+)*$"
+    r"^torch(?:\.(?!_dynamo\.|_inductor\.|_functorch\.)[^.]+)*$"
```
zou3519, line 124: What's the motivation?
I thought for torch._dynamo /torch._inductor things, those shouldn't show up in the FX graph.
It's actually desirable that torch._functorch things do show up in the FX graph.

If you have a good enough motivation then feel free to ship it, I think the difference is just cosmetic (@guilhermeleobas can also confirm). It's just going to take you a while to get through all of the expecttests :P

guilhermeleobas: We do have a few functions from the torch._functorch namespace that shows up in the graph.
https://github.com/pytorch/pytorch/blob/5a1fbf45ad727353e367740ecd8825ca7ee857e9/torch/_dynamo/trace_rules.py#L249-L250

PR #164902, opened 2025-10-08
[5/N][DTensor device order] Implement graph based redistribution algorithm

zpcore:
(Extract out the algorithm from https://github.com/pytorch/pytorch/pull/160266.)

Build a graph to search for the path from source placement to destination placement (with device order). Currently solution introduces too many all-gathers and missing the opportunity for all-to-all when redistribute, especially when we consider the device order.

### How to build the graph:
When operator of Shard, think of collective op as operation on a stack of device axis: 
- I, J are tensor dimensions;
- X, Y, Z, Y are ordered mesh dimensions.
image

Detailed collective op transition is implemented in `DTensorRedistributePlanner.get_next_state`.

### How to find the min cost path:
Assign weight to different type of collective ops and use Dijkstra to find the min cost path from the graph we build.

* #165197
* #165092
* __->__ #164902
* #164821
* #164820
* #164806
* #164797


zpcore:
I extracted out the main redistribution algorithm out into this PR. Without the `redistribute` API implemented, I have to hack with my own `redistribute` implementation in the test.

torch/distributed/tensor/_redistribute.py
```diff
@@ -169,96 +213,425 @@ def stringify_transform_infos(
...
+    ) -> list[_TransformInfo]:
+        """
+        Generate the transform infos from the source placements to the target placements.
+
+        To transform from source to target placement it might have multiple steps, i.e. it
+        might decompose Si -> Sj into Si -> R -> Sj.
+        This would detect if there're mis-aligned/nested shardings between src/dst placements.
+        E.g. Suppose the redistribution to perform is (Shard(0), Shard(0)) -> (Replicate(), Shard(0)),
+        in this case Shard(0) -> Shard(0) for mesh dimension 1 actually needs resharding, because in
+        the former is a nested-sharding of a tensor already already sharded dimension 0, whereras
```
ezyang: whereas

torch/distributed/tensor/_redistribute.py
```diff
@@ -363,10 +758,6 @@ def redistribute_local_tensor(
                 new_local_tensor = local_tensor
                 continue
 
-            logger.debug(
-                "redistribute from %s to %s on mesh dim %s", current, target, i
-            )
```
ezyang, line 368: ?

torch/distributed/tensor/_redistribute.py
```diff
@@ -169,96 +213,425 @@ def stringify_transform_infos(
...
+        full_tensor_shape: tuple[int, ...],
+    ) -> list[_TransformInfo]:
+        assert src_spec.shard_order is not None and dst_spec.shard_order is not None
+        src_state = self.DistState(src_spec.placements, src_spec.shard_order)
+        dst_state = self.DistState(dst_spec.placements, dst_spec.shard_order)
+        transform_infos: list[_TransformInfo] = []
+        state_path = self.find_min_cost_path(src_state, dst_state)
+        for cur_state, nxt_state in itertools.pairwise(state_path):
+            # find the mesh_dim that is different between cur_state and nxt_state
+            if cur_state.placements != nxt_state.placements:
```
ezyang, line 501: why can we generate no-op state transitions?

torch/distributed/tensor/_redistribute.py
```diff
@@ -169,96 +213,425 @@ def stringify_transform_infos(
...
+            new_placements[mesh_dim] = Partial()
+            dist_state = self.DistState(
+                self._to_tuple(new_placements), tensor_mesh_dim_tuple
+            )
+            all_next_state[dist_state] = self.chunk_cost
+
+        return all_next_state
+
+    # TODO(zpcore): if the dst_state contains special placement like
+    # `_MaskPartial`, we will never reach that state. Need to support this case.
```
ezyang, line 414: yes

ezyang (approved):

torch/distributed/tensor/_redistribute.py
zpcore: I feel the logger message is not that useful. It only tells for the mesh dim `i`, how its placement will be changed. Besides, we have DebugMode to print out the `TransformInfo` now.

torch/distributed/tensor/_redistribute.py
zpcore: Oh, I forgot to remove that "no-op" comment. Previously I assume that is a global device order that is applied to all mesh dims, but this is not valid anymore.

torch/distributed/tensor/_redistribute.py
zpcore: This will be handled in a separate PR. The idea is we can first transform the dest state that contains `_MaskPartial` to a normal placement and then conduct the graph search. In this way we won't expand the search space.

PR #164904, opened 2025-10-08
[inductor] verify determinism with inductor benchmark script

shunting314:
* #164905
* __->__ #164904

Verify the deterministic mode with torch.compile benchmark scripts.

Here is what my testing script does (pasted in the end):
- run a model in default mode, save it's result
- run the model again in default mode, but distort the benchmarking results. Compare it with the saved result.
- Do the above again in deterministic mode.

I tried to test a few modes
- BertForMaskedLM and GoogleFnet: I can repro the numeric change by distorting the benchnmark result in the default mode. The non-determinism is gone in the deterministic mode
- DistillGPT2: I can not repro the numeric change by distorting the benchmarking result in the default mode. It does not surprise me much. Reduction order change does not always cause numeric change.

```
model=GoogleFnet

export TORCHINDUCTOR_WRITE_ARE_DETERMINISTIC_ALGORITHMS_ENABLED=0
export TORCHINDUCTOR_FORCE_DISABLE_CACHES=1  # disable autotune cache
export TORCHINDUCTOR_FX_GRAPH_REMOTE_CACHE=0
export TORCHINDUCTOR_FX_GRAPH_CACHE=0
export TORCHINDUCTOR_CACHE_DIR=/tmp/torchinductor_shunting/
export TORCHINDUCTOR_BENCHMARK_KERNEL=1
export TORCHINDUCTOR_UNIQUE_KERNEL_NAMES=1
export INDUCTOR_TEST_DISABLE_FRESH_CACHE=1


# Non deterministic mode
# --float32 rather than --amp to make it easier to repro non-deterministic
echo "Save results for non-deterministic mode"
python benchmarks/dynamo/huggingface.py --backend inductor --float32 --accuracy --only $model --training --disable-cudagraphs --save-model-outputs-to=/tmp/saved-non-deterministic.pkl

echo "Compare results with distorted benchmarking in non-deterministic mode"
TORCHINDUCTOR_DISTORT_BENCHMARKING_RESULT=inverse python benchmarks/dynamo/huggingface.py --backend inductor --float32 --accuracy --only $model --training --disable-cudagraphs --compare-model-outputs-with=/tmp/saved-non-deterministic.pkl

echo "Save results for deterministic mode"
TORCHINDUCTOR_DETERMINISTIC=1 python benchmarks/dynamo/huggingface.py --backend inductor --float32 --accuracy --only $model --training --disable-cudagraphs --save-model-outputs-to=/tmp/saved-deterministic.pkl

echo "Compare results with distorted benchmarking in deterministic mode"
TORCHINDUCTOR_DETERMINISTIC=1 TORCHINDUCTOR_DISTORT_BENCHMARKING_RESULT=inverse python benchmarks/dynamo/huggingface.py --backend inductor --float32 --accuracy --only $model --training --disable-cudagraphs --compare-model-outputs-with=/tmp/saved-deterministic.pkl
```


jansel (approved):

torch/_inductor/compile_fx.py
```diff
@@ -2420,6 +2420,11 @@ def compile_fx(
                 ignore_shape_env=ignore_shape_env,
             )
 
+    if config.deterministic:
+        torch.backends.cudnn.deterministic = True
```
v0i0, line 2451: without knowing much about when this code executed - i feel like this should be reset somehow when we are done with it? or are we documenting that calling compile w/ deterministic=true will permanently change these?

torch/_inductor/runtime/benchmarking.py
```diff
@@ -23,6 +24,40 @@
 T = TypeVar("T")
 
 
+def may_distort_benchmarking_result(fn: Callable[..., Any]) -> Callable[..., Any]:
+    from torch._inductor import config
+
+    distort_method = config.test_configs.distort_benchmarking_result
```
v0i0: should we push this down into the `distort` function s.t. we can patch it easily?

v0i0 (approved):

torch/_inductor/compile_fx.py
shunting314: I think ideally when a user decides to enable deterministic mode, they should already set these flags accordingly. The settings here is just to make sure if a user didn't do that, we will do that.

Not reseting has the benefit that if some graph falls back to eager, they can still be run deterministically.

torch/_inductor/runtime/benchmarking.py
shunting314: yea

PR #164910, opened 2025-10-08
[cleanup] Migrate c10/util exceptions (registry, sparse_bitset, typecast, exception.h)

Nithurshen:
### Summary

This Pull Request addresses a portion of the "We should never throw vanilla C++ exceptions" umbrella issue (#148114) by migrating `throw std::runtime_error` to the recommended `TORCH_CHECK` or `C10_THROW_ERROR` macros within the `c10/util` component.

The goal is to ensure consistent, traceable, and Python-friendly exception handling across PyTorch's core C++ codebase.

### Changes Included

The following files were modified:

* `c10/util/Registry.h`: Replaced `throw std::runtime_error` in `Registry::Register` with `TORCH_CHECK(false, err_msg)`.
* `c10/util/sparse_bitset.h`: Replaced `throw std::runtime_error("Illegal empty element")` with `TORCH_CHECK(false, "Illegal empty element")` in `find_first` and `find_last`.
* `c10/util/Exception.h`: Replaced `throw std::runtime_error` fallback with `C10_THROW_ERROR(Error, ...)` within the `#ifdef STANDALONE_TORCH_HEADER` block.
* `c10/util/TypeCast.cpp`: Replaced `throw std::runtime_error` in `report_overflow` with `TORCH_CHECK(false, ...)` for consistent stack tracing.

### Next Steps

This work **partially addresses #148114**. I am planning to follow up with separate PRs for other unaddressed categories, such as `test/cpp` and `benchmarks/static_runtime`.

***

✅ **Tests:** All existing `c10_test` unit tests were run and passed successfully after the changes.

### DIFF
--- a/c10/util/Exception.h
+++ b/c10/util/Exception.h
@@ -545,7 +545,7 @@ namespace c10::detail {
 #ifdef STRIP_ERROR_MESSAGES
 #define TORCH_CHECK(cond, ...)                \
   if (C10_UNLIKELY_OR_CONST(!(cond))) {       \
-    throw std::runtime_error(TORCH_CHECK_MSG( \
+    C10_THROW_ERROR(Error, TORCH_CHECK_MSG( \
         cond,                                 \
         "",                                   \
         __func__,                             \
@@ -559,7 +559,7 @@ namespace c10::detail {
 #else
 #define TORCH_CHECK(cond, ...)                \
   if (C10_UNLIKELY_OR_CONST(!(cond))) {       \
-    throw std::runtime_error(TORCH_CHECK_MSG( \
+    C10_THROW_ERROR(Error, TORCH_CHECK_MSG( \
         cond,                                 \
         "",                                   \
         __func__,                             \
--- a/c10/util/Exception.h
+++ b/c10/util/Exception.h
@@ -545,7 +545,7 @@ namespace c10::detail {
 #ifdef STRIP_ERROR_MESSAGES
 #define TORCH_CHECK(cond, ...)                \
   if (C10_UNLIKELY_OR_CONST(!(cond))) {       \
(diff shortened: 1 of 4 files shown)


c10/util/Exception.h
albanD, line 540: @desertfire @janeyx99 is this still used? Should we kill it now that we have TORCH_STD_CHECK ?

albanD (commented):
Not sure about the change in Exception.h, but sgtm otherwise

c10/util/Registry.h
```diff
@@ -87,7 +87,7 @@ class Registry {
         if (terminate_) {
           std::exit(1);
         } else {
-          throw std::runtime_error(err_msg);
```
fffrog, line 90: missing the header file `c10/util/Exception.h`

c10/util/Exception.h
```diff
@@ -545,7 +545,7 @@ namespace c10::detail {
 #ifdef STRIP_ERROR_MESSAGES
 #define TORCH_CHECK(cond, ...)                \
   if (C10_UNLIKELY_OR_CONST(!(cond))) {       \
-    throw std::runtime_error(TORCH_CHECK_MSG( \
+    C10_THROW_ERROR(Error, TORCH_CHECK_MSG( \
```
albanD: Given we haven't heard back from the others.
Let's keep these two as is as the only valid exception in the codebase. so we can land the rest of this PR now. 
We can do a specific follow up to discuss the entries in this file later.

Nithurshen: What am I supposed to do from my side?

albanD: move the two throw in this file back to runtime_error for this PR.

Nithurshen: Done, rolled back to std runtime_error in the Exception.h file. lmk if there's anything else to do

PR #164917, opened 2025-10-08
Remove outdated skip conditons on tensor dtype tests

cyyever:
Enables PPC and s390x on float_to_int_conversion tests.


AlekseiNikiforovIBM:
ppc part looks good to me, but maybe @sandeepgupta12 can provide more input on that part.

PR #164921, opened 2025-10-08
fix default backend inference for custom distributed backend

quic-indrbane:
Current implementation uses `custom_backend = next(iter(Backend._plugins.keys()))` and it only makes sense when `Backend._plugins.keys()` contains exactly one custom backend. However, this invariant breaks with `import torch._dynamo` which indirectly adds the `fake` distributed backend used for testing.

This causes the default backend to not get set when:
- using a custom device + custom ccl backend
- using multiple backends config when calling `init_process_group` i.e. `cpu:gloo,custom_device:custom_backend`
- 'FAKE' comes before `custom_backend` name in the dict key order

### DIFF
--- a/torch/distributed/distributed_c10d.py
+++ b/torch/distributed/distributed_c10d.py
@@ -1,7 +1,6 @@
 # mypy: allow-untyped-defs
 """Distributed Collective Communication (c10d)."""
 
-import collections.abc
 import contextlib
 import ctypes
 import hashlib
@@ -14,7 +13,7 @@ import sys
 import time
 import warnings
 from collections import namedtuple
-from collections.abc import Callable
+from collections.abc import Callable, Iterable
 from datetime import timedelta
 from typing import Any, Optional, TYPE_CHECKING, Union
 from typing_extensions import deprecated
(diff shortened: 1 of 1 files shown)


torch/distributed/distributed_c10d.py
```diff
@@ -1957,8 +1956,9 @@ def _new_process_group_helper(
         if Backend.NCCL in backend_config.device_backend_map.values():
             pg._set_default_backend(ProcessGroup.BackendType.NCCL)
         elif Backend._plugins.keys():
-            custom_backend = next(iter(Backend._plugins.keys()))
-            if custom_backend in backend_config.device_backend_map.values():
```
weifengpy, line 1961: in the new implantation, what happens when custom_backend not in device_backend_map.values()

cc @d4l3k for custom backend change

quic-indrbane: In that case, default backend doesn't get set which matches the old behavior.
This change only fixes the case Backend._plugins contains multiple keys: some out-of-tree device backend and the FAKE backend added by torch upstream on _dynamo import.

PR #164928, opened 2025-10-08
Use `equal_nan=True` when doing gradcheck

hameerabbasi:
Uses `equal_nan=True` for all applicable invocations of `torch.allclose` in `torch/autograd/gradcheck.py`.

This is so doing `gradcheck` works in some cases where the gradient is undefined; e.g. `x = 0.0` for `torch.abs`.

Discovered in https://github.com/openteams-ai/pytorch-complex-tensor/pull/13.

cc @ezyang @albanD @gqchen @nikitaved @soulitzer @Varal7 @xmfan @amjames


albanD:
Our general strategy for this has been to only call gradcheck with inputs where we do know the finite difference will be valid.
Hiding NaN will not help that much for these as you might also get finite completely wrong values.

Issue #164725, opened 2025-10-06
[Fuzzer][Eager/Compile Divergence] DDE var_node_0 = var_node_1.item() # dtype=bool


bobrenjc93:
### 🐛 Describe the bug

```
import torch
torch._dynamo.config.capture_scalar_outputs = True
torch._dynamo.config.capture_dynamic_output_shape_ops = True

torch.manual_seed(1000030)

def fuzzed_program(sentinel):
    var_node_4 = torch.full((6,), True, dtype=torch.bool) # size=(6,), stride=(1,), dtype=bool, device=cuda
    var_node_3 = torch.reshape(var_node_4, [2, 3]) # size=(2, 3), stride=(3, 1), dtype=bool, device=cuda
    var_node_5 = torch.full((2, 3), False, dtype=torch.bool) # size=(2, 3), stride=(3, 1), dtype=bool, device=cuda
    _x_ms = torch.arange(max(1, 1), device=var_node_3.device).to(var_node_3.dtype)
    _mask_ms = torch.zeros_like(_x_ms, dtype=torch.bool)
    _mask_ms[:1] = True
    var_node_2 = torch.masked_select(_x_ms, _mask_ms) # size=(1,), stride=(1,), dtype=bool, device=cuda
    var_node_1 = torch.squeeze(var_node_2) # size=(), stride=(), dtype=bool, device=cuda
    var_node_0 = var_node_1.item() # dtype=bool
    # Ensure gradient computation by multiplying with sentinel and taking real part
    result = var_node_0 * sentinel
    if result.is_complex():
        result = result.real
    return result

# Sentinel tensor to ensure gradient computation
sentinel = torch.tensor(1.0, requires_grad=True)


args = () + (sentinel,)
result_original = fuzzed_program(*args)
print('✅ eager success')
compiled_program = torch.compile(fuzzed_program, fullgraph=True, dynamic=True)
result_compiled = compiled_program(*args)
print('✅ compile success')

```

```
   ✅ eager success                                                                                                                                                                                                                             
   Traceback (most recent call last):                                                                                                                                                                                                           
     File "/tmp/torchfuzz/fuzz_b3cba9ad.py", line 31, in <module>                                                                                                                                                                               
       result_compiled = compiled_program(*args)                                                                                                                                                                                                
     File "/home/bobren/local/a/pytorch/torch/_dynamo/eval_frame.py", line 895, in compile_wrapper                                                                                                                                              
[...]

Issue #164704, opened 2025-10-05
[Fuzzer][Eager/Compile Divergence] cpp_fused_mul_2(primals_3, buf10, u1) RuntimeError: expected int arg but got float


bobrenjc93:
### 🐛 Describe the bug

```
import torch
torch._dynamo.config.capture_scalar_outputs = True
torch._dynamo.config.capture_dynamic_output_shape_ops = True

torch.manual_seed(114503)

def fuzzed_program(arg_0, arg_1, sentinel):
    var_node_4 = torch.full((2, 3), 3, dtype=torch.int16) # size=(2, 3), stride=(3, 1), dtype=int16, device=cuda
    var_node_3 = torch.unique(var_node_4) # size=(1,), stride=(1,), dtype=int16, device=cuda
    var_node_2 = torch.squeeze(var_node_3) # size=(), stride=(), dtype=int16, device=cuda
    var_node_7 = arg_0 # size=(), stride=(), dtype=int16, device=cuda
    var_node_8 = arg_1 # size=(), stride=(), dtype=int16, device=cuda
    var_node_6 = torch.sub(var_node_7, var_node_8) # size=(), stride=(), dtype=int16, device=cuda
    var_node_10 = torch.full((1,), 3, dtype=torch.int16) # size=(1,), stride=(1,), dtype=int16, device=cuda
    var_node_9 = torch.squeeze(var_node_10) # size=(), stride=(), dtype=int16, device=cuda
    var_node_5 = torch.add(var_node_6, var_node_9) # size=(), stride=(), dtype=int16, device=cuda
    var_node_1 = torch.div(var_node_2, var_node_5) # size=(), stride=(), dtype=int16, device=cuda
    var_node_0 = var_node_1.item() # dtype=int16
    # Ensure gradient computation by multiplying with sentinel and taking real part
    result = var_node_0 * sentinel
    if result.is_complex():
        result = result.real
    return result

# Sentinel tensor to ensure gradient computation
sentinel = torch.tensor(1.0, requires_grad=True)

arg_0 = torch.as_strided(torch.randint(5, 30, (1,)).to(torch.int16), (), ())
arg_1 = torch.as_strided(torch.randint(5, 30, (1,)).to(torch.int16), (), ())

args = (arg_0, arg_1) + (sentinel,)
result_original = fuzzed_program(*args)
print('✅ eager success')
compiled_program = torch.compile(fuzzed_program, fullgraph=True, dynamic=True)
[...]

PR #164934, opened 2025-10-08
Fix Avoid DDE in item numel check

laithsakka:
* __->__ #164934

address https://github.com/pytorch/pytorch/issues/164725 and https://github.com/pytorch/pytorch/issues/164704

### DIFF
--- a/aten/src/ATen/native/Scalar.cpp
+++ b/aten/src/ATen/native/Scalar.cpp
@@ -1,6 +1,6 @@
 #define TORCH_ASSERT_ONLY_METHOD_OPERATORS
-#include <ATen/Dispatch_v2.h>
 #include <ATen/core/Tensor.h>
+#include <ATen/Dispatch_v2.h>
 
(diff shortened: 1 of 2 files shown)


ezyang (approved):
good catch!

aten/src/ATen/native/Scalar.cpp
```diff
@@ -15,7 +15,11 @@ namespace at::native {
 
 Scalar item(const Tensor& self) {
   auto numel = self.sym_numel();
-  TORCH_CHECK(numel == 1, "a Tensor with ", numel, " elements cannot be converted to Scalar");
+  TORCH_SYM_CHECK(
```
aorenste, line 18: As we discussed in-person (but possibly a different PR) we should comment TORCH_CHECK and TORCH_SYM_CHECK better to explain the difference.

aorenste (approved):

Skylion007 (approved):

PR #164938, opened 2025-10-08
[inductor] Expand use of generic benchmark function

kundaMwiza:
Use the more generic `Benchmarker.benchmark` function to allow benchmarking other devices that support the required functionality, for example prologue and epilogue fusion can be benchmarked for triton CPU.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @chauhang @penguinwu @voznesenskym @EikanWang @Guobing-Chen @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @aakhundov @coconutruben @Lucaskabela


kundaMwiza:
@nmacchioni Can I please get a review for this benchmarking update? Probably other reviewers are needed for the different parts of the codebase

torch/_inductor/wrapper_benchmark.py
```diff
@@ -137,7 +138,13 @@ def get_info_str(
                     f"  {get_info_str(ms, launcher.n_regs, launcher.n_spills, launcher.shared)} @ {launcher.config}"
                 )
         else:
-            ms = benchmarker.benchmark_gpu(lambda: kernel_mod.call(args), rep=40)
+            ms = benchmarker.benchmark(
+                lambda: kernel_mod.call(args),
+                fn_args=tuple(),
```
nmacchioni: Since `benchmarker.benchmark` is already going to wrap call in a lambda, why don't we just pass args as `fn_args = (args,),` and pass fn as `kernel_mod.call` so we're not adding an extra wrap

torch/_inductor/ir.py
```diff
@@ -4990,7 +4990,9 @@ def benchmark(self, *args: Any, out: torch.Tensor) -> float:
         }
         if config.profile_bandwidth_with_do_bench_using_profiling:
             return do_bench_using_profiling(lambda: algo(*args), **benchmark_configs)  # type: ignore[arg-type]
-        return benchmarker.benchmark(algo, args, {"out": out}, **benchmark_configs)
+        return benchmarker.benchmark(
+            algo, args, {"out": out}, device=None, **benchmark_configs
```
nmacchioni, line 5234: nit: device default is None so we don't have to specify it

torch/_inductor/codegen/triton_combo_kernel.py
```diff
@@ -923,7 +924,7 @@ def codegen_kernel_benchmark(self, num_gb: float) -> IndentedBuffer:
 
             result.writeline("args = get_args()")
             result.writeline(
-                "ms = benchmarker.benchmark_gpu(lambda: call(args), rep=40)"
+                f"ms = benchmarker.benchmark(lambda: call(args), fn_args=tuple(), fn_kwargs=dict(), device={device.type},rep=40)"
```
nmacchioni: same thing here with wrapping the call twice

torch/_inductor/codegen/triton.py
```diff
@@ -5343,13 +5351,20 @@ def store_cache():
...
+                    fn_kwargs=dict(),
+                    device=device,
                 )
-                ms_clone = benchmarker.benchmark_gpu(
-                    lambda: wrapped_jit_function.clone_args(*args)[0]
+                ms_clone = benchmarker.benchmark(
+                    lambda: wrapped_jit_function.clone_args(*args)[0],
+                    fn_args=tuple(),
+                    fn_kwargs=dict(),
+                    device=device,
```
nmacchioni, line 5992: same, unwrap call

torch/_inductor/codegen/triton.py
```diff
@@ -5157,24 +5157,32 @@ def load_cache():
...
                 )
                 # overhead of cloning args gives bias for fusing the kernel
                 # in the case of mutating/in-placeable second fusion
                 # TODO - would be better as a hook in triton do_bench that reset
                 # the input values between benchmarking
                 if len(wrapped_jit_function.mutated_arg_names) > 0:
-                    ms = ms - benchmarker.benchmark_gpu(
-                        lambda: wrapped_jit_function.clone_args(*args)
+                    ms = ms - benchmarker.benchmark(
+                        lambda: wrapped_jit_function.clone_args(*args),
```
nmacchioni, line 5813: here too

torch/_inductor/codegen/triton.py
```diff
@@ -5157,24 +5157,32 @@ def load_cache():
...
                 # skip benchmarking the kernel if there are register spills
                 ms = float("inf")
             else:
+                device = V.graph.get_current_device_or_throw()
                 # We have to clone the inplace updated arguments to avoid earlier calls
                 # generating out of range indices for later calls.
-                ms = benchmarker.benchmark_gpu(
-                    lambda: call(wrapped_jit_function.clone_args(*args)[0])
+                ms = benchmarker.benchmark(
+                    lambda: call(wrapped_jit_function.clone_args(*args)[0]),
```
nmacchioni, line 5804: and here

torch/_inductor/codegen/triton.py
```diff
@@ -4239,7 +4239,7 @@ def codegen_kernel_benchmark(self, num_gb: Optional[float]) -> IndentedBuffer:
 
             result.writeline("args = get_args()")
             result.writeline(
-                "ms = benchmarker.benchmark_gpu(lambda: call(args), rep=40)"
+                f"ms = benchmarker.benchmark(lambda: call(args), fn_args=tuple(), fn_kwargs=dict(), device={V.graph.get_current_device_or_throw().type}, rep=40)"  # noqa: B950 line too long
```
nmacchioni: unwrap call

torch/_inductor/codegen/subgraph.py
```diff
@@ -120,7 +120,11 @@ def benchmark(self, *args: list[Any], out: torch.Tensor) -> float:
                 bm_func([*sym_inputs, *args])
         if config.profile_bandwidth_with_do_bench_using_profiling:
             return do_bench_using_profiling(lambda: bm_func([*sym_inputs, *args]))
-        return benchmarker.benchmark_gpu(lambda: bm_func([*sym_inputs, *args]))
+        return benchmarker.benchmark(
+            lambda sym_inputs, args: bm_func([*sym_inputs, *args]),
+            fn_args=(sym_inputs, args),
```
nmacchioni: `bm_func`,
`fn_args=([*sym_inputs, *args],),`

torch/_inductor/codegen/multi_kernel.py
```diff
@@ -378,7 +379,15 @@ def inner():
             return inner
 
         return [
-            benchmarker.benchmark_gpu(wrap_fn(kernel, index), rep=40)
+            benchmarker.benchmark(
```
nmacchioni, line 384: sorry, why do we need if/else here? It seems both if/else would evaluate to the same output

nmacchioni (approved):
LGTM, just a couple small changes;

adding @eellison for review of select algorithm and scheduler changes, specifically enabling benchmarking for fusions/gemm autotuning of Triton CPU

torch/_inductor/ir.py
kundaMwiza: This is specified to bypass the linter error from unpacking benchmark_configs when device isn't specified:
```
Argument 4 to "benchmark" ... has incompatible type "**dict[str, int]"; expected "str | device | None
```

torch/_inductor/codegen/triton.py
kundaMwiza: This is a function composition so anyway a new function would have to be defined. I've changed `benchmark` to only wrap the function call if the user has provided arguments. In some cases it is clearer at the callsite what is going to be benchmarked

torch/_inductor/codegen/multi_kernel.py
kundaMwiza: I've changed it

torch/_inductor/codegen/triton.py
kundaMwiza: See response above

torch/_inductor/codegen/triton.py
kundaMwiza: Kept for clarity, see above

torch/_inductor/codegen/subgraph.py
```diff
@@ -120,7 +120,11 @@ def benchmark(self, *args: list[Any], out: torch.Tensor) -> float:
                 bm_func([*sym_inputs, *args])
         if config.profile_bandwidth_with_do_bench_using_profiling:
             return do_bench_using_profiling(lambda: bm_func([*sym_inputs, *args]))
-        return benchmarker.benchmark_gpu(lambda: bm_func([*sym_inputs, *args]))
+        return benchmarker.benchmark(
+            bm_func,
+            fn_args=([*sym_inputs, *args],),
+            fn_kwargs=dict(),
```
eellison: nit: can we update benchmark to:

```
fn_kwargs: Optional[dict[str, Any]] = None
...
fn_kwargs = {} if fn_kwargs is not None else fn_kwargs
```

and get rid of the `fn_kwargs=dict()` here ?

torch/_inductor/codegen/triton.py
```diff
@@ -5163,18 +5163,25 @@ def load_cache():
...
                 # overhead of cloning args gives bias for fusing the kernel
                 # in the case of mutating/in-placeable second fusion
                 # TODO - would be better as a hook in triton do_bench that reset
                 # the input values between benchmarking
                 if len(wrapped_jit_function.mutated_arg_names) > 0:
-                    ms = ms - benchmarker.benchmark_gpu(
-                        lambda: wrapped_jit_function.clone_args(*args)
+                    ms = ms - benchmarker.benchmark(
+                        lambda: wrapped_jit_function.clone_args(*args),
+                        fn_args=tuple(),
```
eellison: similarly, fn_args should be optional..

eellison (approved):
Looks good - one request would be to fill in default args for fn_args/fn_kwargs when they are not passed in.

torch/_inductor/ir.py
nmacchioni: Oh, that's unfortunate. Ok, np then

torch/_inductor/runtime/benchmarking.py
```diff
@@ -109,34 +140,52 @@ def benchmark(
         device-specific implementations, like `benchmark_cpu` and `benchmark_gpu`. Raises
         `ValueError(...)` if we can't safely infer the device type of `fn`; for example,
         if multiple device types are found in `fn_args` and `fn_kwargs`, or if no device
-        types are found.
+        types are found. To bypass device inference, provide the device to the `device`
+        parameter.
+
+        WARNING: if `fn` mutates `fn_args` or `fn_kwargs`, benchmarking may fail unexpectedly.
```
kundaMwiza, line 146: @nmacchioni Added some more changes, mainly for cases when the benchmarking function mutates the input arguments, like Runner.call(args) which clears the args, so the caller needs to pass a new list each time

kundaMwiza: @nmacchioni bump, CI is now green

kundaMwiza:
@nmacchioni The failures look unrelated to this PR to me

nmacchioni (approved):
The new changes LGTM, thanks!

PR #164941, opened 2025-10-08
Switch build jobs that use linux.12xlarge to c7i

zxiiro:
This PR updates build jobs that currently use linux.12xlarge to the c7i varient which should increase build times by 15% - 20% depending on the job and reduce costs of these jobs by 10% - 15%.

Relates to pytorch/test-infra#7175


huydhn (approved):
LGTM!

chuanqi129 (approved):

PR #164954, opened 2025-10-08
[DeviceMesh] Make all members of DeviceMesh private and add public access API

fduwjj:
* #164510
* __->__ #164954
* #164750

This is mostly mechanical change which make device mesh members all private and use a public property API instead. This is not a BC breaking change since the new API still guarantee BC.

cc @H-Huang @awgu @wanchaol @fegin @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @ezyang @gchanan


fegin (approved):
BC linter cannot figure out that this is not BC. You may want to suppress the BC linter

PR #164956, opened 2025-10-08
Move from/to to torch::stable::detail

janeyx99:
To not pollute the global namespace, we should move the `from`/`to` APIs into torch::stable::detail. We are also following our normal deprecation cycle and choosing to continue exposing the global `from`/`to` for the time being as people who onboard their extensions onto 2.9 would not be able to build with 2.10 otherwise.

Note that this means that within libtorch, we do not get the luxury of tacking on a `using torch::stable::detail::from` because then it leads to build time ambiguous calls --> both the global and namespace APIs are exposed, which one do I want? So that is why you see every local site is updated.

Note that the update is _not_ necessary from a custom op writer point of view. FA3 can continue to build on torch nightlies without changing any code. (Since this is a header change, this PR has no implication on runtime, a previously built FA3 ABI stable wheel will continue to work fine with newer torch versions after this PR.)

Once TORCH_BOX lands, we would be free to remove these global APIs when the deprecation cycle is up (April 2026) and encourage people to use TORCH_BOX and avoid from/to entirely.

* #164959
* __->__ #164956


lw (commented):
I'd prefer waiting until we make a call on https://github.com/pytorch/pytorch/pull/164882 before reviewing

torch/csrc/inductor/aoti_torch/shim_common.cpp
```diff
@@ -1413,28 +1413,28 @@ static StableIValue from_ivalue(
     case c10::TypeKind::TensorType: {
       AtenTensorHandle ath = torch::aot_inductor::new_tensor_handle(
           std::move(const_cast<at::Tensor&>(ivalue.toTensor())));
-      return from(ath);
+      return torch::stable::detail::from(ath);
```
malfet, line 1416: Just add `using namespace torch::stable::detail` on line 1411 and don't make any changes later on

torch/csrc/stable/ops.h
```diff
@@ -18,15 +18,15 @@ namespace torch::stable {
 inline torch::stable::Tensor empty_like(const torch::stable::Tensor& self) {
   const auto num_args = 6;
   std::array<StableIValue, num_args> stack{
-      from(self),
-      from(std::nullopt),
-      from(std::nullopt),
-      from(std::nullopt),
-      from(std::nullopt),
-      from(std::nullopt)};
+      torch::stable::detail::from(self),
```
malfet, line 21: Same as above, add `using namespace torch::stable::detail`

malfet (approved):

torch/csrc/inductor/aoti_torch/shim_common.cpp
janeyx99: If we want to keep the global from/to for BC, this leads to ambiguity. I wrote it in my PR desc: 
Note that this means that within libtorch, we do not get the luxury of tacking on a using torch::stable::detail::from because then it leads to build time ambiguous calls --> both the global and namespace APIs are exposed, which one do I want? So that is why you see every local site is updated.

PR #164958, opened 2025-10-08
[cuDNN][SDPA] Handle noncontig nested tensors in cuDNN SDPA

eqy:
Previously we hardcoded the assumption in cuDNN that the inputs would be dense which breaks when e.g., the user is chunking tensors yielding noncontig inputs

New test added to check this  when `TORCH_CUDNN_SDPA_NESTED_TENSOR_ENABLED=1` is set in `test/test_transformers.py`

One issue I noticed was that the old gating of nested tensor in `sdp_utils.cpp` seems to be a no-op? All of the inputs are reported as "dense" by the time that function is called in the nested tensor tests in `test/test_nestedtensor.py -k sdpa`

cc @csarofeen @ptrblck @xwang233 @cpuhrsch @jbschlosser @bhosmer @drisspg @soulitzer @davidberard98 @YuqingJ

### DIFF
--- a/aten/src/ATen/native/cudnn/MHA.cpp
+++ b/aten/src/ATen/native/cudnn/MHA.cpp
@@ -1540,9 +1540,10 @@ void run_cudnn_SDP_fprop_nestedtensor(
   }
   auto seqlen_q = at::diff(cum_seqlen_q, 1, 0);
   auto seqlen_kv = at::diff(cum_seqlen_kv, 1, 0);
-  auto rag_q_off = cum_seqlen_q.mul(h_q * d_qk);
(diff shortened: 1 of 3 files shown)


Skylion007 (approved):

aten/src/ATen/native/cudnn/MHA.cpp
```diff
@@ -1551,7 +1552,7 @@ void run_cudnn_SDP_fprop_nestedtensor(
       {SCALE, &scaling_factor},
       {O, o.data_ptr()},
```
Skylion007: nit: data_ptr is deprecated, please use mutable or const_data_ptr

drisspg (approved):

PR #164959, opened 2025-10-08
Add clang-tidy misc-definitions-in-headers check

janeyx99:
* __->__ #164959
* #164956


Skylion007 (approved):

torch/csrc/stable/.clang-tidy
```diff
@@ -5,5 +5,6 @@
 InheritParentConfig: true
 Checks: '
 google-global-names-in-headers,
+misc-definitions-in-headers,
```
Skylion007, line 8: Pretty sure this clang-tidy check does nothing unless any compilation targets on in the folder, but go ahead.

janeyx99: haha when I develop it shows underlines, which is better than not having underlines

mikaylagawarecki: yea I think it's good to add, and I'll test if turning on `warnings-as-errors` does anything in CI in https://github.com/pytorch/pytorch/pull/164963

mikaylagawarecki (approved):

PR #164964, opened 2025-10-08
Add view_as_real and view_as_complex for sparse tensors

### LINKED ISSUE #118153: Add support for sparse tensors with view_as_real

Currently, `aten::view_as_real` is not implemented for sparse tensors. A good reason to implement this is to enable us to support SparseAdam with complex dtypes.

### Alternatives

_No response_

### Additional context

_No response_

cc @alexsamardzic @nikitaved @pearu @cpuhrsch @amjames @bhosmer @jcaip @vincentqb @jbschlosser @albanD @crcrpar

cleonard530:
Fixes #118153

Updated torch.view_as_real and torch.view_as_complex to work with sparse matrices. 

@janeyx99

### DIFF
--- a/aten/src/ATen/native/native_functions.yaml
+++ b/aten/src/ATen/native/native_functions.yaml
@@ -424,6 +424,7 @@
   variants: function
   dispatch:
     CPU, CUDA, MPS, Meta: view_as_complex
+    SparseCPU, SparseCUDA: view_as_complex_sparse
 
 - func: sgn(Tensor self) -> Tensor
   variants: function, method
--- a/aten/src/ATen/native/native_functions.yaml
+++ b/aten/src/ATen/native/native_functions.yaml
@@ -418,13 +418,13 @@
   variants: function
   dispatch:
     CPU, CUDA, MPS, Meta: view_as_real
-    SparseCPU, SparseCUDA: view_as_real_sparse
+    SparseCPU, SparseCUDA, SparseMPS: view_as_real_sparse
 
 - func: view_as_complex(Tensor(a) self) -> Tensor(a)
   variants: function
   dispatch:
     CPU, CUDA, MPS, Meta: view_as_complex
-    SparseCPU, SparseCUDA: view_as_complex_sparse
+    SparseCPU, SparseCUDA, SparseMPS: view_as_complex_sparse
 
 - func: sgn(Tensor self) -> Tensor
   variants: function, method
--- a/aten/src/ATen/native/native_functions.yaml
+++ b/aten/src/ATen/native/native_functions.yaml
@@ -418,6 +418,7 @@
   variants: function
   dispatch:
     CPU, CUDA, MPS, Meta: view_as_real
+    SparseCPU, SparseCUDA: view_as_real_sparse   # , MPS, Meta: view_as_real
 
 - func: view_as_complex(Tensor(a) self) -> Tensor(a)
   variants: function
--- a/aten/src/ATen/native/native_functions.yaml
+++ b/aten/src/ATen/native/native_functions.yaml
@@ -418,7 +418,7 @@
   variants: function
   dispatch:
     CPU, CUDA, MPS, Meta: view_as_real
-    SparseCPU, SparseCUDA: view_as_real_sparse   # , MPS, Meta: view_as_real
+    SparseCPU, SparseCUDA: view_as_real_sparse
 
 - func: view_as_complex(Tensor(a) self) -> Tensor(a)
   variants: function
--- a/aten/src/ATen/native/sparse/SparseTensor.cpp
+++ b/aten/src/ATen/native/sparse/SparseTensor.cpp
@@ -903,7 +903,7 @@ Tensor view_as_real_sparse(const Tensor& self) {
   TORCH_CHECK(self.is_sparse() && self.is_complex(), "view_as_real_sparse is only supported for complex sparse tensors");
   TORCH_CHECK(!self.is_conj(), "view_as_real_sparse doesn't work on unresolved conjugated tensors.  To resolve the conjugate tensor so you can view it as real, use self.resolve_conj(); however, be warned that the resulting tensor will NOT alias the original.");
 
-  auto new_sizes = self.sizes().vec();
+  auto new_sizes = self.sym_sizes().vec();
   // last dimension will always have two elements containing the real and imag vals
   new_sizes.push_back(2);
 
@@ -911,7 +911,7 @@ Tensor view_as_real_sparse(const Tensor& self) {
   const auto float_type = c10::toRealValueType(self.scalar_type());
   auto options = self.options().dtype(float_type);
 
-  return at::_sparse_coo_tensor_with_dims_and_tensors(
+  return at::_sparse_coo_tensor_with_dims_and_tensors_symint(
       self.sparse_dim(),
       self.dense_dim() + 1,  // Add one dense dimension for real/imag
       new_sizes,
--- a/aten/src/ATen/native/sparse/SparseTensor.cpp
+++ b/aten/src/ATen/native/sparse/SparseTensor.cpp
@@ -66,6 +66,7 @@
 #include <ATen/ops/unique_dim.h>
 #include <ATen/ops/values_native.h>
 #include <ATen/ops/view_as_real.h>
+#include <ATen/ops/view_as_real_native.h>
 #include <ATen/ops/zeros.h>
 #include <ATen/ops/ones.h>
 #endif
--- a/aten/src/ATen/native/sparse/SparseTensor.cpp
+++ b/aten/src/ATen/native/sparse/SparseTensor.cpp
@@ -67,6 +67,8 @@
 #include <ATen/ops/values_native.h>
 #include <ATen/ops/view_as_real.h>
 #include <ATen/ops/view_as_real_native.h>
+#include <ATen/ops/view_as_complex.h>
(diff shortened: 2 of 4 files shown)


aten/src/ATen/native/sparse/SparseTensor.cpp
```diff
@@ -897,4 +899,28 @@ Tensor _pin_memory_sparse_coo(const Tensor& self, std::optional<Device> device)
       self.is_coalesced());
 }
 
+Tensor view_as_real_sparse(const Tensor& self) {
+  TORCH_CHECK(self.is_sparse(), "view_as_real_sparse is only supported for sparse tensors");
+  TORCH_CHECK(self.is_complex(), "view_as_real_sparse is only supported for complex sparse tensors");
```
janeyx99:
```suggestion
  TORCH_CHECK(self.is_sparse() && self.is_complex(), "view_as_real_sparse is only supported for complex sparse tensors");
```

janeyx99 (commented):
Thank you!!

The PR looks good to me, ccing @soulitzer  and @albanD  for the view portions.

aten/src/ATen/native/sparse/SparseTensor.cpp
cleonard530: Thanks, @janeyx99. Just updated the TORCH_CHECKs

aten/src/ATen/native/sparse/SparseTensor.cpp
```diff
@@ -897,4 +899,27 @@ Tensor _pin_memory_sparse_coo(const Tensor& self, std::optional<Device> device)
       self.is_coalesced());
 }
 
+Tensor view_as_real_sparse(const Tensor& self) {
+  TORCH_CHECK(self.is_sparse() && self.is_complex(), "view_as_real_sparse is only supported for complex sparse tensors");
+  TORCH_CHECK(!self.is_conj(), "view_as_real_sparse doesn't work on unresolved conjugated tensors.  To resolve the conjugate tensor so you can view it as real, use self.resolve_conj(); however, be warned that the resulting tensor will NOT alias the original.");
+
+  auto new_sizes = self.sizes().vec();
```
Skylion007: should new sizes be symbolize sizes?

cleonard530: Yes it should, my bad. Just updated it.

isuruf (approved):

test/test_sparse.py
```diff
@@ -5623,6 +5623,19 @@ def generic_constructor(*args, **kwargs):
                     RuntimeError, r"memory pinning of \w*indices \(=0\) must match memory pinning of values \(=1\)"):
                 generic_constructor(*args2, **kwargs)
 
+    @onlyNativeDeviceTypes
+    @dtypes(torch.complex32, torch.complex64, torch.complex128)
+    def test_view_as_real(self, device, dtype):
+        x = torch.tensor(2 + 3j, dtype=dtype, device=device)
```
pearu: I suggest using the following testing stategy:
```python
for xs in self.sample_inputs_sparse(torch.sparse_coo, device=device, dtype=dtype):
    x = xs.to_dense()
    res = torch.view_as_real(xs)
    self.assertEqual(res.layout, torch.sparse_coo)
    self.assertEqual(res._indices(), xs._indices())
    self.assertEqual(res.to_dense(), torch.view_as_real(x))
```
that will cover a variety of sparse tensor instances.

pearu (changes requested):
Overall, looks good! Thanks, @cleonard530!

I have a suggestion to add more samples to tests as the current sample with a single element is a degenerate one.

Do you plan implement sparse support for `view_as_complex` as well?
Btw, the approach used here is applicable to sparse compressed tensors as well.

test/test_sparse.py
cleonard530: Thank you @pearu, I will update the test.

cleonard530:
> Do you plan implement sparse support for `view_as_complex` as well? Btw, the approach used here is applicable to sparse compressed tensors as well.

Hi @pearu, I had previously opened an issue to add support for 'view_as_complex' as well (see #165781). However, after discussing with @ngimel in the associated PR (#166742) it was pointed out to me that the 'view' methods should be a cheap operation without complex data copying. I think this is unavoidable in view_as_complex since either the real or imaginary part of the tensor my be 0 and thus there is no **value** in the 'real view' of the tensor, but may need to part of the 'complex view'. 

I think this view_as_real implementation avoids that by using `auto real_values = at::view_as_real(self._values()); ` to create the 'real view' so the data is not copied. Let me know if I am wrong about this though and I will try to address it. 

Thanks!

cleonard530:
@pearu, that makes since. 

Do you think the view_as_complex should be added to this PR or should I reopen the Issue https://github.com/pytorch/pytorch/pull/166742 but only implement it under the conditions you stated (i.e. last dense_dim == 2 and keeping the indices the same).

test/test_sparse.py
```diff
@@ -5691,6 +5691,47 @@ def generic_constructor(*args, **kwargs):
...
+            self.assertEqual(res.layout, torch.sparse_coo)
+            self.assertEqual(res._indices(), xs._indices())
+            self.assertEqual(res.shape, xs.shape + (2,))
+            if dtype is torch.complex32 and torch.device(device).type == "cpu":
+                # ComplexHalf to_dense() is not supported on CPU.
+                # check res values manually
+                self.assertEqual(res._values()[..., 0], xs._values().real)
+                self.assertEqual(res._values()[..., 1], xs._values().imag)
+                continue
+            self.assertEqual(res.to_dense(), torch.view_as_real(xs.to_dense()))
```
pearu: Nit, testing `values` is always valid:
```suggestion
            self.assertEqual(res._values()[..., 0], xs._values().real)
            self.assertEqual(res._values()[..., 1], xs._values().imag)
            if not (dtype is torch.complex32 and torch.device(device).type == "cpu"):
                # ComplexHalf to_dense() is not supported on CPU.
                self.assertEqual(res.to_dense(), torch.view_as_real(xs.to_dense()))

            self.assertEqual(torch.view_as_complex(torch.view_as_real(xs)), xs)
```

test/test_sparse.py
```diff
@@ -5691,6 +5691,47 @@ def generic_constructor(*args, **kwargs):
...
+                self.assertEqual(res.to_dense(), torch.view_as_complex(xs.to_dense()))
+            except RuntimeError as e:
+                if xs.shape[-1] != 2 or xs.dense_dim() == 0:
+                    self.assertIn(
+                        "view_as_complex_sparse is only supported for sparse tensors with the last dim == 2 and dense_dim > 0.",
+                        str(e))
+                elif xs._values().stride()[-1] != 1:
+                    self.assertIn("Tensor must have a last dimension with stride 1", str(e))
+                else:
+                    raise
```
pearu, line 5723: Nit, minimize statements in the try-block:
```suggestion
            try:
                res = torch.view_as_complex(xs)
            except RuntimeError as e:
                if xs.shape[-1] != 2 or xs.dense_dim() == 0:
                    self.assertIn(
                        "view_as_complex_sparse is only supported for sparse tensors with the last dim == 2 and dense_dim > 0.",
                        str(e))
                elif xs._values().stride()[-1] != 1:
                    self.assertIn("Tensor must have a last dimension with stride 1", str(e))
                else:
                    raise
                continue
            self.assertEqual(res.layout, torch.sparse_coo)
            self.assertEqual(res._indices(), xs._indices())
            self.assertEqual(res.shape, xs.shape[:-1])
            self.assertEqual(res._values().real, xs._values()[..., 0])
            self.assertEqual(res._values().imag, xs._values()[..., 1])

            if not (dtype is torch.float16 and torch.device(device).type == "cpu"):
                # ComplexHalf to_dense() is not supported on CPU.
                self.assertEqual(res.to_dense(), torch.view_as_complex(xs.to_dense()))

            self.assertEqual(torch.view_as_real(torch.view_as_complex(xs)), xs)
```

pearu (changes requested):
Looks good!

I have two nits in tests that include adding invariant checks about `view_as_real` and `view_as_complex` relation.

cleonard530:
@pearu, sorry, I don't know how I missed that review comment. I made the updated the test file now.

pearu (approved):
LGTM! Thanks, @cleonard530!

aten/src/ATen/native/native_functions.yaml
```diff
@@ -418,11 +418,13 @@
   variants: function
   dispatch:
     CPU, CUDA, MPS, Meta: view_as_real
+    SparseCPU, SparseCUDA: view_as_real_sparse
```
pearu:
```suggestion
    SparseCPU, SparseCUDA, SparseMPS: view_as_real_sparse
```

aten/src/ATen/native/native_functions.yaml
```diff
@@ -418,11 +418,13 @@
...
 
 - func: view_as_complex(Tensor(a) self) -> Tensor(a)
   variants: function
   dispatch:
     CPU, CUDA, MPS, Meta: view_as_complex
+    SparseCPU, SparseCUDA: view_as_complex_sparse
```
pearu:
```suggestion
    SparseCPU, SparseCUDA, SparseMPS: view_as_complex_sparse
```

pearu (changes requested):
@cleonard530 , here is a suggestion to fix macos CI failures.

pearu (approved):
LGTM!

PR #164965, opened 2025-10-08
TORCH_CHECK(false, ...) -> TORCH_FAIL(..)

laithsakka:
* __->__ #164965

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @gujinghui @PenghuiCheng @jianyuh @min-jean-cho @yanbing-j @Guobing-Chen @Xia-Weiwen @snadampal @mcarilli @ptrblck @leslie-fang-intel @EikanWang @voznesenskym @penguinwu @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela

### DIFF
--- a/aten/src/ATen/core/ivalue.cpp
+++ b/aten/src/ATen/core/ivalue.cpp
@@ -757,7 +757,7 @@ IValueComparator getLessThanComparator(const IValue& v) {
     torch::jit::Function* lt_func =
         checkObjectSortSchema(v.type()->expect<ClassType>(), why_not);
     if (!lt_func) {
-      TORCH_CHECK(false, why_not.str());
+     TORCH_FAIL(why_not.str());
     }
 
     return [lt_func](const IValue& a, const IValue& b) {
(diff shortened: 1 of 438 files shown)


aorenste (approved):
nice

albanD (changes requested):
I don't think we should do this for a few reasons:
- It is a lot of churn in the codebase for very little benefits.
- It is another new error macro we need to know about. For example for extensions and headeronly
- It doesn't align well with the nicer macros TORCH_CHECK_VALUE/TYPE/etc that convert to the appropriate python errors.

Would you be able to share some of the motivation on why you want this?

aten/src/ATen/core/ivalue.cpp
```diff
@@ -1053,7 +1053,7 @@ c10::intrusive_ptr<ivalue::Object> ivalue::Object::deepcopy(
       }
       err << ". Please define serialization methods via def_pickle() for "
             "this class.";
-      TORCH_CHECK(false, err.str());
+     TORCH_FAIL(err.str());
```
Skylion007: What's wthe indent changes here.

laithsakka: addressed

PR #164978, opened 2025-10-08
[inductor][choices] lookup table choices 1/3

coconutruben:
* #164980
* #164979
* __->__ #164978

\# why

- enable users to control which choices get used on which inputs
- reduce lowering time, and pin kernel selection, by selecting
  them for the inputs

\# what

- a new InductorChoices subclass that implements a lookup table
- a README explaining the usage
- corresponding testing

- currently only supports templates that go through
  `V.choices.get_template_configs`

\# testing

```
python3 -bb -m pytest test/inductor/test_lookup_table.py -v
```

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @mlazos

Differential Revision: [D85685743](https://our.internmc.facebook.com/intern/diff/D85685743)


torch/_inductor/config.py
```diff
@@ -2004,6 +2004,17 @@ class trace:
 external_matmul: list[Callable[[torch.Tensor, torch.Tensor, torch.Tensor], None]] = []
 
 
+class lookup_table:
+    # Lookup table for template config overrides
+    table: Optional[dict[str, list[dict[str, Any]]]] = None
+
+    # Enable template src_hash checking in lookup table to prevent using stale configs.
+    # If True, configs with 'template_hash' field will be compared against the template's
+    # src_hash at runtime and filtered out if they don't match. If False or None, no
```
eellison: nit: no need for `or None`

PaulZhang12 (approved):
LGTM! cc @jansel @eellison

torch/_inductor/lookup_table/README.md
```diff
@@ -0,0 +1,153 @@
...
+### Device Key Handling
+
+Configure device key behavior:
+```python
+# Control whether entries include device-specific keys for lookups
+# Device-agnostic entries work across different GPU models
+```
+
+**Lookup Behavior**: During lookup, the system automatically tries both key formats:
+1. **Device-specific key** (e.g., `"NVIDIA H100+input_data+mm"`) - tried first
```
eellison, line 45: nit: is there any specification of the key format?

torch/_inductor/lookup_table/README.md
```diff
@@ -0,0 +1,153 @@
...
+- `input_key`: Generated from `KernelInputs.key` property, represents tensor shapes/dtypes/strides
+- `op_name`: Operation name (`"mm"`, `"addmm"`, etc.)
+
+Each value is a list of configuration dictionaries containing:
+- `template_id`: Template identifier (`"triton:mm"`, `"triton::mm_persistent_tma"`, `"decompose_k"`, etc.)
+- Template-specific parameters (`BLOCK_M`, `BLOCK_N`, `BLOCK_K`, `num_warps`, etc.)
+
+### Example Table
+
+This is an example table for a single input showing two configurations
```
eellison, line 174: similarly - any specification for how input is represented ? i know it has the potential to change in future.

torch/_inductor/lookup_table/README.md
eellison: would it make sense to include cuda version here? e.g. cuda 12.9 gemms will be faster than 12.4, etc.

torch/_inductor/lookup_table/README.md
```diff
@@ -0,0 +1,153 @@
...
+    {
+      "template_id": "aten::bias_addmm"
+    },
+  ]
+}
+```
+
+## Source Hashing Safety
+
+The lookup table system includes source hashing to prevent using stale configurations when template code changes.
```
eellison, line 202: awesome !

torch/_inductor/lookup_table/README.md
```diff
@@ -0,0 +1,153 @@
...
+## Performance Impact
+
+- **Lookup Hit**: Eliminates heuristic choice generation and autotuning overhead (if a single choice)
+- **Lookup Miss**: Default behavior, including heuristic choice generation and autotuning
+- **Memory**: Table stored in memory, minimal overhead for key generation and lookup
+
+## Implementation Details
+
+### Key Generation
+- Device key: Uses `torch.cuda.get_device_properties().gcnArchName` (e.g., "NVIDIA H100")
```
eellison, line 240: I guess this is the key specification part. I think this is fine if it's a bit underspecified, and subject to change.

torch/_inductor/lookup_table/choices.py
```diff
@@ -0,0 +1,391 @@
...
+    All lookup functionality is contained within this class and can be customized by overriding methods.
+    """
+
+    def _get_lookup_table(self) -> dict[str, list[dict[str, Any]]]:
+        """
+        Get the template lookup table from config.
+        Override this method to use custom lookup table sources (database, API, etc.).
+        """
+        if not torch.cuda.is_available() or config.lookup_table.table is None:
+            return {}
```
eellison, line 38: nit: in the future, no specific need for `torch.cuda.is_available`

eellison: okay, i guess we need this for now bc of limitations in device key..

torch/_inductor/lookup_table/choices.py
```diff
@@ -0,0 +1,391 @@
...
+    def _generate_kernel_inputs_key(kernel_inputs: KernelInputs) -> str:
+        """
+        Generate a key based on input node properties and scalars.
+        The key includes dtype, size, and stride information for each input node,
+        plus scalar values as key=value pairs separated by & signs.
+        """
+        # Get node information using existing methods
+        dtypes = kernel_inputs.dtypes()
+        shapes = kernel_inputs.shapes_hinted()
+        strides = kernel_inputs.strides_hinted()
```
eellison, line 68: Nit: would it ever make sense to distinguish between static, and dynamic hints ? E.g. for decompose k, the shapes actually do need to be static to be divisible by k

eellison: similarly, alignment ? e.g. for some of the tma based kernels i think we need to know that

torch/_inductor/lookup_table/choices.py
```diff
@@ -0,0 +1,391 @@
...
+        # Create tuple of (dtype, shape_list, stride_list) for each node
+        node_info = tuple(
+            (dtype, list(shape), list(stride))
+            for dtype, shape, stride in zip(dtypes, shapes, strides)
+        )
+
+        # Create base key from node information
+        fmt_key = str(node_info)
+        # Add scalar information if present
+        if kernel_inputs._scalars:
```
eellison, line 79: it's a little silly that we'll need a different choices entry for different alpha, beta, but i think this is not particularly high prioroity to fix..

torch/_inductor/lookup_table/choices.py
```diff
@@ -0,0 +1,391 @@
...
+                    configs_by_template[template_id] = []
+                configs_by_template[template_id].append(cfg)
+
+        # Check template hashes and clean up template_id field
+        result = {}
+        for template_id, matching_configs in configs_by_template.items():
+            filtered_configs = []
+            for cfg in matching_configs:
+                # Check template hash if enabled
+                if config.lookup_table.check_src_hash and template_hash_map:
```
eellison: nit: factor out to a helper function ?

eellison (approved):

torch/_inductor/lookup_table/choices.py
PaulZhang12: cc @aorenste these are cases that we probably want to test for in distributed autotuning as well? Maybe this can be a handoff

torch/_inductor/lookup_table/README.md
coconutruben: I added a key schema section and a comment explaining the device specific key and encouraging the user to file a request if they need a CUDA specific key. device key is really only there to help deployments, rather than needing BUCK rules or other fancy things to swap out the table from the codebase depending on the target architecture - there are probably multitude of those use cases, I think I'd be more helpful for users to ask for them rather than building them proactively (e.g. can you have a cuda version without a specific device, etc)

torch/_inductor/lookup_table/choices.py
coconutruben: the way this works now is that the template would not be included (https://github.com/pytorch/pytorch/blob/main/torch/_inductor/utils.py#L1916) and so the entry would not be used, because it doesn't match any template we're using for that input. I'll take a note to evolve the templates a bit to state clearly _I can be used_ (input dependent) and _I should be used_ config dependent

mlazos (approved):

coconutruben:
D85812420 should have mitigated these issues internally

@pytorchbot merge

PR #164979, opened 2025-10-08
[inductor][lookup table] add recorder 2/3

coconutruben:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #164980
* __->__ #164979
* #164978

\# why

- enable users to record full tables to plug into the system

\# what

- recorder system
- default recorders that
  - write individual entries into debug logs
  - dump table progressively into output directory by setting
    `TORCH_INDUCTOR_LOOKUP_TABLE_RECORD_DIR`

\# testing

- added new unit tests

```
python3 -bb -m pytest test/inductor/test_lookup_table.py -v
```

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @jataylo @chenyang78

Differential Revision: [D85685744](https://our.internmc.facebook.com/intern/diff/D85685744)

PR #164980, opened 2025-10-08
[inductor][choices] use LookupTableChoices by default 3/3

coconutruben:
* __->__ #164980
* #164979
* #164978

\# why

- an empty table is a noop
- reduce friction for users to use the table

\# what

- virtualized uses LookupTableChoices as the default InductorChoices
- lookup table behavior is gated by an explicit bool that defaults to
  True, in addition to having any entries in the table at all

\# testing

- existing unit tests

Differential Revision: [D85685741](https://our.internmc.facebook.com/intern/diff/D85685741)


nmacchioni (approved):

torch/_inductor/config.py
```diff
@@ -2074,6 +2074,9 @@ class trace:
 
 
 class lookup_table:
+    # Enable/disable lookup table choices system (defaults to True)
+    active: bool = True
```
PaulZhang12, line 2078: Set through env var?

PaulZhang12 (approved):

PR #164986, opened 2025-10-08
Move test/functorch to test/functorch_test

ezyang:
* __->__ #164986

Apparently, there is some bad interaction with pip where because
the test/functorch folder has the name 'functorch', it ends up being
picked up when you do an editable install (instead of functorch).
This didn't used to happen with the old python setup.py develop
path, but it happens with pip install -e.  We should probably file
an upstream bug, but this makes the tests start working again with
pytest.

You can repro the old problem with `pytest test/functorch/test_aotdispatch.py` (if you just run python or directly import from CLI it works fine).

Fixes https://github.com/pytorch/pytorch/issues/164957


albanD (approved):
Ahhhhh that makes sense!

zou3519 (approved):

malfet (approved):

PR #164987, opened 2025-10-08
[BE] Use torch check the way its intended

malfet:
Replace
`if (!foo) TORCH_CHECK(false, "bar");` with `TORCH_CHECK(foo, "bar");`


albanD (approved):
Nice

Skylion007 (approved):

PR #164988, opened 2025-10-08
[fr] Enable reset the FR recording for fault tolerance

fduwjj:
* __->__ #164988
* #164752

We also want to have a python side API for users to reset FR recording for FR entries. We don't need to reset the PGNCCL's member counter since we are creating new PGNCCL anyway. FR is a global ring buffer, so we need to reset it.

### DIFF
--- a/torch/csrc/distributed/c10d/init.cpp
+++ b/torch/csrc/distributed/c10d/init.cpp
@@ -4091,6 +4091,10 @@ such as `dist.all_reduce(tensor, async_op=True)`.
             Stringified pickle work traces.
             Default settings return everything - i.e. contains NCCL comm dumps and collective traces.
       )");
+  module.def(
+      "_reset_fr_recording_nccl",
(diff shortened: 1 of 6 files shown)


torch/csrc/distributed/c10d/init.cpp
```diff
@@ -4091,6 +4091,10 @@ such as `dist.all_reduce(tensor, async_op=True)`.
             Stringified pickle work traces.
             Default settings return everything - i.e. contains NCCL comm dumps and collective traces.
       )");
+  module.def(
+      "_reset_fr_recording_nccl",
```
tushar00jain, line 4095: do we want to prefix the public api with _?

tushar00jain (approved):

torch/csrc/distributed/c10d/init.cpp
fduwjj: Let's make it a private API for now.

PR #164989, opened 2025-10-08
[CI] Use smaller amx + avx2 runners for inductor test?

clee2000:
Results from CI:
No failures but generally takes longer, maybe ~20% increase in time?
But the smaller runner is ~25% of the cost of the current runner, so in terms of cost this is a decrease

If the 20% is too much, we can try the 4x larger runners, which are about half the cost of the current runner, so it would probably still result in cost savings with hopefully less impact to time


BoyuanFeng (approved):

huydhn (approved):
LGTM! These are not CPU perf benchmark jobs, and only do accuracy check.  So, running them slower is ok I guess as long as they finish within 2 to 2h30 hour mark

PR #126135, opened 2024-05-14
[codemod] `c10:optional` -> `std::optional`

r-barnes:
Generated by running the following from PyTorch root:
```
find . -regex ".*\.\(cpp\|h\|cu\|hpp\|cc\|cxx\)$" | grep -v "build/" | xargs -n 50 -P 4 perl -pi -e 's/c10::optional/std::optional/'
```

`c10::optional` is just an alias for `std::optional`. This removes usages of that alias in preparation for eliminating it entirely.

cc @mrshenli @pritamdamania87 @zhaojuanmao @satgera @gqchen @aazzolini @osalpekar @jiayisuse @H-Huang @kwen2501 @awgu @penguinwu @fegin @XilunWu @wanchaol @fduwjj @wz337 @tianyu-l @wconstab @yf225 @chauhang @d4l3k @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @gujinghui @PenghuiCheng @jianyuh @min-jean-cho @yanbing-j @Guobing-Chen @Xia-Weiwen @snadampal @mcarilli @ptrblck @leslie-fang-intel @EikanWang @albanD @voznesenskym @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @peterbell10 @ipiszy @chenyang78 @kadeng @muchulee8 @ColinPeppler @amjames @desertfire

PR #101995, opened 2023-05-22
[PyTorch] Redirect c10::optional to std::optional

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #101995

We have C++17 now!

I am intentionally dropping the `c10::optional<c10::ArrayRef>` size optimization. It was intended to improve dispatch, but thanks to D34602980 / #70864 we don't use `optional<ArrayRef>` in function arguments anymore anyway.

Differential Revision: [D46079028](https://our.internmc.facebook.com/intern/diff/D46079028/)

cc @mrshenli @pritamdamania87 @zhaojuanmao @satgera @rohan-varma @gqchen @aazzolini @osalpekar @jiayisuse @H-Huang @kwen2501 @awgu @penguinwu @fegin @XilunWu @wanchaol @fduwjj @wz337 @tianyu-l @wconstab @yf225 @ezyang @gchanan @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @EikanWang @kiukchung @d4l3k @lucasllc

PR #59333, opened 2021-06-02
[PyTorch] RFC: optimize c10::optional<ArrayRef<T>>

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack):
* **#59333 [PyTorch] RFC: optimize c10::optional<ArrayRef<T>>**

Code comment should explain this in sufficient detail. In brief, making it 16 bytes should get it to be passed in registers.

Differential Revision: [D28843027](https://our.internmc.facebook.com/intern/diff/D28843027/)

PR #164991, opened 2025-10-08
Refactor out headeronly ArrayRef

janeyx99:
* #165953
* #165153
* #165152
* __->__ #164991

Differential Revision: [D85091961](https://our.internmc.facebook.com/intern/diff/D85091961)


torch/headeronly/util/HOArrayRef.h
```diff
@@ -0,0 +1,262 @@
...
+  using reverse_iterator = std::reverse_iterator<iterator>;
+
+ protected:
+  /// The start of the array, in an external buffer.
+  const T* Data;
+
+  /// The number of elements.
+  size_type Length;
+
+  // CHANGED void debugCheckNullptrInvariant() {
```
janeyx99: I will remove the CHANGED comments before landing--moreso wanted to show reviewers the diff between HOArrayRef and ArrayRef

c10/util/ArrayRef.h
```diff
@@ -40,58 +41,52 @@ namespace c10 {
 ///
 /// This is intended to be trivially copyable, so it should be passed by
 /// value.
+///
+/// NOTE: We have refactored out the headeronly parts of the ArrayRef struct
+/// into HOArrayRef. As adding `virtual` will change the performance of the
```
swolchok: nit: s/will/would

c10/util/ArrayRef.h
```diff
@@ -40,58 +41,52 @@ namespace c10 {
...
-
-  /// The number of elements.
-  size_type Length;
-
   void debugCheckNullptrInvariant() {
     TORCH_INTERNAL_ASSERT_DEBUG_ONLY(
-        Data != nullptr || Length == 0,
+        this->Data != nullptr || this->Length == 0,
         "created ArrayRef with nullptr and non-zero length! std::optional relies on this being illegal");
   }
```
swolchok: you can delete this function. "std::optional relies on this being illegal" is false. Relevant history here:

1) https://github.com/pytorch/pytorch/pull/59333 (by yours truly) added this debug check, together with a specialization of c10::optional<c10::ArrayRef> to reduce its size.
2) https://github.com/pytorch/pytorch/pull/101995 (by yours truly) removed the c10::optional and replaced it with a simple redirect to std::optional, "intentionally dropping the c10::optional<c10::ArrayRef> size optimization" (with a note that it wasn't as important anymore). I forgot that this debug check was here, oops! sorry.
3) https://github.com/pytorch/pytorch/pull/126135 find/replaced this message from c10::optional to std::optional.

I haven't specifically checked whether someone has tried to reintroduce this optimization, but if they had it would be undefined behavior -- you can't specialize std-namespaced things unless they explicitly allow you to, and I don't see any mention of specialization on https://en.cppreference.com/w/cpp/utility/optional.html .

if this check *was* important, I would instead ask you to guarantee it in HOArrayRef as well.

After you delete this function, you can get rid of all the constructor implementations that just call into HOArrayRef's constructors and simply write `using HOArrayRef::HOArrayRef`.

c10/util/ArrayRef.h
```diff
@@ -111,129 +106,86 @@ class ArrayRef final {
...
-    return Data;
-  }
-
-  /// size - Get the array size.
-  constexpr size_t size() const {
-    return Length;
-  }
-
   /// front - Get the first element.
+  /// We deviate from HOArrayRef by using TORCH_CHECK instead of STD_TORCH_CHECK
```
swolchok: is STD_TORCH_CHECK really so bad? if so, why is it always correct to call HOArrayRef functions instead? Is there some set of unwritten rules people have to be aware of about when it's OK for them to use HOArrayRef?

torch/headeronly/util/HOArrayRef.h
```diff
@@ -0,0 +1,250 @@
...
+#include <cstddef>
+#include <functional>
+#include <initializer_list>
+#include <iterator>
+#include <type_traits>
+#include <vector>
+
+namespace c10 {
+
+/// HOArrayRef (HO = HeaderOnly) - A subset of ArrayRef that is implemented only
```
swolchok: personally I think we should just call it HeaderOnlyArrayRef. Code is written once and read many times, and typing has generally not been the bottleneck, even before people were trying to have robots do the typing for them.

torch/headeronly/util/HOArrayRef.h
```diff
@@ -0,0 +1,250 @@
...
+/// in headers. This will be a base class from which ArrayRef inherits, so that
+/// we can keep much of the implementation shared.
+///
+/// [HOArrayRef vs ArrayRef note]
+/// As HOArrayRef is a subset of ArrayRef, it has slightly less functionality than
+/// ArrayRef. We document the differences below:
+/// 1. ArrayRef has a debug-only internal assert that prevents construction with a
+//     nullptr Data and nonzero length in the following constructors: from a pointer
+//     and length, from a range, from a templated Container with .data() and .size().
+//     HOArrayRef does not make that check for any constructor.
```
swolchok: delete

torch/headeronly/util/HOArrayRef.h
```diff
@@ -0,0 +1,250 @@
...
+/// 2. ArrayRef can be constructed from a SmallVector. HOArrayRef cannot.
```
swolchok: nitpick: you can still make a HOArrayRef given a SmallVector, you just don't get the convenience constructor

swolchok (changes requested):

c10/util/ArrayRef.h
janeyx99: lol ok, got it!!

c10/util/ArrayRef.h
janeyx99: From what I understand, c10::Error does some traceback handling to present the error in a nice way to the user. STD_TORCH_CHECK just calls std::runtime_error so it's less full. STD_TORCH_CHECK still checks properly though, but it scores fewer points in "user understandability", so it's not "incorrect", but not as ideal.

torch/headeronly/util/HOArrayRef.h
janeyx99: Yea, I wasn't fully committed to HOArrayRef, will update

c10/util/ArrayRef.h
swolchok: How do you propose avoiding the creation of a perceived mess where people don't understand when they should use HeaderOnlyArrayRef vs ArrayRef?

swolchok: I'm now wondering if we can get STD_TORCH_CHECK to do the right thing -- throw c10::Error when linking with libtorch, but throw std::runtime_error when not linking with libtorch. I suspect we might be able to get the job done with weak symbols. Let me see if I can get a PR up...

swolchok: weak symbols will do the job on macOS (and presumably Linux), but Googling seems to suggest that there isn't an equivalent for Windows.

swolchok: > I'm now wondering if we can get STD_TORCH_CHECK to do the right thing

per offline discussion, 1) @janeyx99 is concerned about code in headeronly changing its behavior depending on whether libtorch is linked 2) it is definitely unclear how to do this for windows, and we care about windows. therefore it is not clear that the behavior I proposed is the "right thing", shelving it.

c10/util/ArrayRef.h
```diff
@@ -40,200 +41,105 @@ namespace c10 {
 ///
 /// This is intended to be trivially copyable, so it should be passed by
 /// value.
+///
+/// NOTE: We have refactored out the headeronly parts of the ArrayRef struct
+/// into HeaderOnlyArrayRef. As adding `virtual` would change the performance of
+/// the underlying constexpr calls, we rely on apparent-type dispatch for
+/// inheritance. This should be fine because their memory format is the same,
+/// and it is never incorrect for ArrayRef to call HeaderOnlyArrayRef methods.
```
swolchok, line 49: I think we should probably add something like "However, you should prefer to use ArrayRef when possible, because its use of TORCH_CHECK will lead to better user-facing error messages."

torch/headeronly/util/HeaderOnlyArrayRef.h
```diff
@@ -0,0 +1,246 @@
...
+/// in headers. This will be a base class from which ArrayRef inherits, so that
+/// we can keep much of the implementation shared.
+///
+/// [HeaderOnlyArrayRef vs ArrayRef note]
+/// As HeaderOnlyArrayRef is a subset of ArrayRef, it has slightly less
+/// functionality than ArrayRef. We document the minor differences below:
+/// 1. ArrayRef has an extra convenience constructor for SmallVector.
+/// 2. ArrayRef uses TORCH_CHECK. HeaderOnlyArrayRef uses header-only
+/// STD_TORCH_CHECK,
+///    which will output a std::runtime_error vs a c10::Error.
```
swolchok: and here, something like "Consequently, you should use ArrayRef when possible and HeaderOnlyArrayRef only when necessary to support headeronly code."

c10/util/ArrayRef.h
```diff
@@ -40,200 +41,105 @@ namespace c10 {
...
-  /// Construct an ArrayRef from a single element.
-  // TODO Make this explicit
-  constexpr ArrayRef(const T& OneElt) : Data(&OneElt), Length(1) {}
-
-  /// Construct an ArrayRef from a pointer and length.
-  constexpr ArrayRef(const T* data, size_t length)
-      : Data(data), Length(length) {
-    debugCheckNullptrInvariant();
-  }
+  // Inherits all constructors from HeaderOnlyArrayRef<T>
```
swolchok: I don't love comments that explain C++ language features

torch/headeronly/util/HeaderOnlyArrayRef.h
```diff
@@ -0,0 +1,246 @@
...
+
+  /// equals - Check for element-wise equality.
+  constexpr bool equals(HeaderOnlyArrayRef RHS) const {
+    return this->Length == RHS.Length &&
+        std::equal(begin(), end(), RHS.begin());
+  }
+
+  /// slice(n, m) - Take M elements of the array starting at element N
+  constexpr HeaderOnlyArrayRef<T> slice(size_t N, size_t M) const {
+    STD_TORCH_CHECK(
```
swolchok, line 175: it occurs to me that having to repeat these error messages isn't great. if you wanted to cut that out, you could

```
#define ARRAY_REF_SLICE(check, class_name, N, M) \
  check( \
    N + M <= this->size(), \
// ...
#class_name ": invalid slice, N = ", \
// ...
```

up to you whether the macro ugliness is worth the deduplication, though.

swolchok (approved):

torch/headeronly/util/HeaderOnlyArrayRef.h
janeyx99: eh

PR #164992, opened 2025-10-08
[dynamo][DebugMode] mask python keys in dispatch_key_set guard checks

pianpwk:
I found that running any compiled function under DebugMode more than once will trigger recompilations, e.g. with the really simple modified test case in `test_compile`:
```
[0/1] [__recompiles] Recompiling function f in /data/users/pianpwk/ptclone/pytorch/test/distributed/tensor/debug/test_debug_mode.py:268
[0/1] [__recompiles]     triggered by the following guard failure(s):
[0/1] [__recompiles]     - 0/0: 
[0/2] [__recompiles] Recompiling function f in /data/users/pianpwk/ptclone/pytorch/test/distributed/tensor/debug/test_debug_mode.py:268
[0/2] [__recompiles]     triggered by the following guard failure(s):
[0/2] [__recompiles]     - 0/1: 
[0/2] [__recompiles]     - 0/0:
```

Digging deeper, the guard failures were due to TENSOR_MATCH guards failing on dispatch key set checks (seemingly on the Python dispatch key):
https://github.com/pytorch/pytorch/blob/5a1fbf45ad727353e367740ecd8825ca7ee857e9/torch/csrc/dynamo/guards.cpp#L199-L203

This seems to due to the `ignore_compile_internals=True` flag on custom dispatch modes being on, which causes these modes to "hide" themselves during compilation, making dynamo guard on the Python dispatch key being off.

The (maybe imperfect) solution is to mask out the Python keys for guard comparisons, when `_is_in_any_mode_without_ignore_compile_internals` is False.


torch/csrc/dynamo/guards.cpp
```diff
@@ -196,9 +195,20 @@ bool TensorCheck::check(
     const c10::SymIntArrayRef& sym_sizes,
     const c10::SymIntArrayRef& sym_strides,
     const bool& requires_grad) {
-  if (dispatch_key_ != state.apply(dispatch_key_set).raw_repr() ||
-      dtype_ != dtype || device_index_ != device.index() ||
-      requires_grad_ != requires_grad) {
+
+  // Mask to exclude Python keys, that can be artifically off during
+  // compilation, due to ignore_compile_internals being set on custom
+  // TorchDispatchModes.
```
albanD: I'm not very familiar with TensorCheck but the name seems to imply that this is the check associated to a given Tensor?
If so, I don't buy this as the dispatch mode would add the Python key to the global include set and not modify any Tensor's keyset.

pianpwk: Ah, I think the state.apply() call is adding the Python key in at runtime? https://github.com/pytorch/pytorch/blob/a2f29bcd6388acdc3202d8a90974c50ffb605104/torch/csrc/dynamo/guards.h#L22-L26

I double checked and the include set here is missing the Python key at compilation, and only has it afterwards.

Would it maybe make more sense to always have the Python keys in the exclude set here? Sorry, I'm not very familiar with dispatch key nuances.

williamwen42 (approved):

PR #164993, opened 2025-10-08
[DeviceMesh] Add a warning for slicing flattened dim from root mesh and types for _get_slice_mesh_layout

fduwjj:
* #164510
* __->__ #164993
* #164954
* #164750

As title, we want to add a deprecate warning for slicing flattened dim from root mesh. Also cosmetic changes for adding types for `_get_slice_mesh_layout`.

### DIFF
--- a/torch/distributed/device_mesh.py
+++ b/torch/distributed/device_mesh.py
@@ -72,145 +72,23 @@ else:
     class _MeshEnv(threading.local):
         def __init__(self) -> None:
             self.mesh_stack: list[DeviceMesh] = []
-            # TODO: Move the bookkeeping maps from _MeshEnv to DeviceMesh.
-            self.child_to_root_mapping: dict[DeviceMesh, DeviceMesh] = {}
-            # Record flatten mesh name to its flattened mesh in root mesh.
-            self.root_to_flatten_mapping: dict[DeviceMesh, dict[str, DeviceMesh]] = {}
 
         def get_current_mesh(self) -> "DeviceMesh":
             if len(self.mesh_stack) == 0:
                 raise RuntimeError("No device mesh is currently active!")
             return self.mesh_stack[-1]
 
-        def create_sub_mesh(
-            self,
-            device_mesh: "DeviceMesh",
-            layout: _MeshLayout,
-            submesh_dim_names: tuple[str, ...],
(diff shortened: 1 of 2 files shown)


torch/distributed/device_mesh.py
```diff
@@ -281,11 +283,17 @@ def _get_slice_mesh_layout(self, device_mesh, mesh_dim_names) -> _MeshLayout:
...
                     layout_sliced.append(
-                        device_mesh._layout[device_mesh.mesh_dim_names.index(name)]
+                        device_mesh._layout[
+                            not_none(device_mesh.mesh_dim_names).index(name)
+                        ]
                     )
                 elif name in flatten_name_to_root_layout:
+                    warnings.warn(
+                        "Slicing a flattened dim from root mesh will be deprecated soon. "
+                        "Users need to use and bookkeep the flattened mesh directly. "
```
fegin: The `to use and bookkeep` seems to be incorrect. Are you trying to say `Users need to bookkeedp the flattened mesh directly."?

torch/distributed/device_mesh.py
```diff
@@ -281,11 +283,17 @@ def _get_slice_mesh_layout(self, device_mesh, mesh_dim_names) -> _MeshLayout:
...
+                if name in not_none(device_mesh.mesh_dim_names):
                     layout_sliced.append(
-                        device_mesh._layout[device_mesh.mesh_dim_names.index(name)]
+                        device_mesh._layout[
+                            not_none(device_mesh.mesh_dim_names).index(name)
+                        ]
                     )
                 elif name in flatten_name_to_root_layout:
+                    warnings.warn(
+                        "Slicing a flattened dim from root mesh will be deprecated soon. "
```
fegin: Should we just say will be deprecated in 2.11? This PR will be included in 2.10, so 2.11 seems to be a reasonable time.

fegin (approved):

torch/distributed/device_mesh.py
fduwjj: sure

torch/distributed/device_mesh.py
fduwjj: right. I mean they should also use it directly instead of slicing it from root mesh.

PR #164995, opened 2025-10-08
[DTensor] Register einsum sharding strategy for inference_mode support

stmcgovern:
Fixes #157631
Adds sharding strategy registration for `aten.einsum.default` to fix AssertionError when using einsum with DTensor in `inference_mode()`.
Adds tests

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @aditvenk @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @aditew01 @voznesenskym @penguinwu @EikanWang @Guobing-Chen @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @Lucaskabela @chenyang78


stmcgovern:
resolved merge conflict.  Then switched target branch from viable/strict to main. Sorry for all the automatic review requests.

torch/distributed/tensor/_ops/_matrix_ops.py
```diff
@@ -267,6 +267,40 @@ def scaled_mm_strategy(op_schema: OpSchema) -> OpStrategy:
...
+    Multi-operand einsum (3+ tensors) is not yet supported and will raise an assertion.
+
+    See gen_einsum_strategies() in _einsum_strategy.py for details on supported
+    sharding patterns.
+    """
+    mesh = op_schema.get_mesh_from_args()
+    # einsum signature: einsum(equation, *operands)
+    # args_schema: [equation_str, OpStrategy1, OpStrategy2, ...]
+    equation = op_schema.args_schema[0]
+    assert isinstance(equation, str), (
```
zpcore: Not sure if this is necessary. I guess if it is not a str, it should already error out somewhere.

torch/distributed/tensor/_ops/_matrix_ops.py
```diff
@@ -267,6 +267,40 @@ def scaled_mm_strategy(op_schema: OpSchema) -> OpStrategy:
...
+
+    This handles einsum operations with DTensor inputs, supporting
+    various sharding patterns for batch dimensions, contracting dimensions,
+    and free dimensions based on the einsum equation.
+
+    The needs_pytree=True flag is crucial because einsum accepts
+    variable-length operands as torch.einsum(equation, *operands).
+
+    Note: Currently supports einsum with up to 2 tensor operands and single output.
+    Multi-operand einsum (3+ tensors) is not yet supported and will raise an assertion.
```
zpcore: Can you assert check if the op_schema[1] length is 2?

test/distributed/tensor/test_dtensor_ops.py
```diff
@@ -817,6 +817,78 @@ def run_opinfo_test(
     def assertEqualOnRank(self, x, y, msg=None, *, rank=0):
         self.assertEqual(x, y, msg)
 
+    def test_einsum_inference_mode(self):
```
zpcore: Suggest move the test to test/distributed/tensor/test_tensor_ops.py.

test/distributed/tensor/test_dtensor_ops.py
```diff
@@ -817,6 +817,78 @@ def run_opinfo_test(
...
+        with torch.inference_mode():
+            r_dtensor_replicate = torch.einsum(
+                "a b c, a d e c -> b d e", X_replicate, Y_replicate
+            )
+
+        # Verification: result should match local computation
+        r_local = torch.einsum("a b c, a d e c -> b d e", X_local, Y_local)
+
+        self.assertEqual(r_dtensor_replicate.to_local(), r_local)
+        self.assertEqual(r_dtensor_replicate.placements, (Replicate(),))
```
zpcore: See if we can reuse test util function [_test_op_on_dtensor](https://github.com/pytorch/pytorch/blob/150682ba7f5081f1b652316d2b822dbc3a0b2da5/torch/testing/_internal/distributed/_tensor/common_dtensor.py#L461) to clean up the code after relocating the test to test_tensor_ops.py.

zpcore (commented):
Generally LGTM, left a few comments.

ezyang:
Why do we not just decompose in this case? Saves you from having to write an entire new sharding strategy.

ezyang (approved):
ok

torch/distributed/tensor/_ops/_matrix_ops.py
stmcgovern: assert added

test/distributed/tensor/test_dtensor_ops.py
stmcgovern: done

test/distributed/tensor/test_dtensor_ops.py
stmcgovern: used in one of the added tests for correctness. the other two have placement checks.

torch/distributed/tensor/_ops/_matrix_ops.py
stmcgovern: I updated to mm_strategy style of checking.

torch/distributed/tensor/_ops/_matrix_ops.py
```diff
@@ -267,6 +268,60 @@ def scaled_mm_strategy(op_schema: OpSchema) -> OpStrategy:
...
+    # einsum signature: einsum(equation, *operands)
+    # args_schema: [equation_str, TupleStrategy(OpStrategy1, OpStrategy2, ...)]
+    equation = cast(str, op_schema.args_schema[0])
+    tensor_operands = cast(TupleStrategy, op_schema.args_schema[1])
+
+    # Validate number of operands (tensor inputs)
+    num_operands = len(tensor_operands.children)
+    assert num_operands <= 2, (
+        f"Currently only support einsum with up to 2 tensor operands, got {num_operands}. "
+        f"Multi-operand einsum (3+ tensors) is not yet supported."
```
ezyang: Nah, now this is worse than the decomp. Decomp would have worked with arbitrarily many tensors

stmcgovern: Thanks @ezyang. Decomposition is great! I made a change so that it now falls back to decomposing when 3 or more operands are used. I don't want to change the logic of https://github.com/pytorch/pytorch/blob/2df2c316e2929519a5bad029936e9e37ed637b12/torch/distributed/tensor/_ops/_einsum_strategy.py#L31

I wonder if the decision to use decomposition or fused strategies is something that can be evaluated in the placement selection logic (available at trace time)? This might consider possible decompositions with computed_costs and compare to strategies and not just use decomposition as a fallback if no strategy is available.

ezyang (changes requested):
unfortunately needs a rebase now...

stmcgovern:
> unfortunately needs a rebase now...

No worries, I've rebased it. Were there other changes requested @ezyang ?

wconstab (commented):
Note that @weifengpy is working on cutting all the matmul ops over to new single dim einsum strat, it would be good to do the same for this.

wconstab (approved):
Seems ok to land though, and migrate later, if that's what you prefer.

PR #164999, opened 2025-10-08
Simplifying computation of the final result for equals op on DTensor

dzmitry-huba:
Instead of collecting local results using all_gather_object followed by local reduction, with this change we switch to using a single all_reduce with MIN reduction operation to compute the final equals result.

This change is needed to enable LocalTensor work (all_gather_object introduces challenges in for DTensor and LocalTensor integration).

topic: not user facing


ezyang (approved):
what's the xfail?

PR #165004, opened 2025-10-08
[opaque_obj_v2] PyObject custom op schema type

angelayi:
* #165005
* __->__ #165004

This is a cleaner implementation of opaque objects (https://github.com/pytorch/pytorch/pull/162660). Instead now we just need to do:

Call `register_opaque_type` to register the type as being "opaque" and allowed by custom ops. You also need to pass a unique name that maps to the type.
```python
class OpaqueQueue:
    def __init__(self, queue: list[torch.Tensor], init_tensor_: torch.Tensor) -> None:
        super().__init__()
        self.queue = queue
        self.init_tensor_ = init_tensor_

    def push(self, tensor: torch.Tensor) -> None:
        self.queue.append(tensor)

    def pop(self) -> torch.Tensor:
        if len(self.queue) > 0:
            return self.queue.pop(0)
        return self.init_tensor_

    def size(self) -> int:
        return len(self.queue)

register_opaque_type(OpaqueQueue, "_TestOpaqueObject_OpaqueQueue")
```

When creating the custom op, the schema will then use the unique name: 
```python
self.lib = torch.library.Library("_TestOpaqueObject", "FRAGMENT")

torch.library.define(
    "_TestOpaqueObject::queue_push",
    "(_TestOpaqueObject_OpaqueQueue a, Tensor b) -> ()",
    tags=torch.Tag.pt2_compliant_tag,
    lib=self.lib,
)

@torch.library.impl(
    "_TestOpaqueObject::queue_push", "CompositeExplicitAutograd", lib=self.lib
)
def push_impl(queue: OpaqueQueue, b: torch.Tensor) -> None:
    assert isinstance(queue, OpaqueQueue)
    queue.push(b)
```

Using the custom op:
```python
queue = OpaqueQueue([], torch.zeros(3))
torch.ops._TestOpaqueObject.queue_push(queue, torch.ones(3))
self.assertTrue(queue.size(), 1)
```


torch/_library/opaque_object.py
```diff
@@ -150,3 +150,33 @@ def set_payload(opaque_object: torch._C.ScriptObject, payload: Any) -> None:
...
+_OPAQUE_TYPES: dict[Any, str] = {}
+
+
+def register_opaque_type(type_: Any, fqn: str) -> None:
+    """
+    Registers the given type as an opaque type which allows this to be consumed
+    by a custom operator.
+
+    Args:
+        type_ (type): The type to register as an opaque type.
```
albanD: "cls" is a classic name when taking a class as input

torch/_library/opaque_object.py
```diff
@@ -150,3 +150,33 @@ def set_payload(opaque_object: torch._C.ScriptObject, payload: Any) -> None:
...
+        fqn (str): A unique qualified name of the type.
```
albanD: name is more generic than fqn.
I would also make it optional and use `cls.__name__` by default.

torch/_library/opaque_object.py
```diff
@@ -150,3 +150,33 @@ def set_payload(opaque_object: torch._C.ScriptObject, payload: Any) -> None:
...
         )
     torch._C._set_opaque_object_payload(opaque_object, payload)
+
+
+_OPAQUE_TYPES: dict[Any, str] = {}
+
+
+def register_opaque_type(type_: Any, fqn: str) -> None:
+    """
+    Registers the given type as an opaque type which allows this to be consumed
```
albanD, line 160: We should define "opaque" better here. I expect: contains no Tensor or SymInt right?

albanD (approved):
SGTM !

PR #165005, opened 2025-10-08
[opaque_obj_v2] make_fx support

angelayi:
* #163284
* __->__ #165005

By wrapping the python objects with FakeScriptObject(FakeOpaqueQueue) we restrict users to do anything to this object. torch.compile support can be easily enabled by the rest of [this stack](https://github.com/pytorch/pytorch/pull/163936) and existing support for ScriptObjects.

One thing to note is that by default in functionalization we mark all ops that take in FakeScriptObjects as being effectful. Should this be the case for these custom ops that take in python objs?  


test/test_opaque_obj_v2.py
```diff
@@ -53,6 +80,15 @@ def pop_impl(queue: OpaqueQueue) -> torch.Tensor:
 
         self.lib.impl("queue_pop", pop_impl, "CompositeExplicitAutograd")
 
+        def pop_impl_fake(q: OpaqueQueue) -> torch.Tensor:
+            # This is not accurate since the queue could have tensors that are
+            # not rank 1
+            ctx = torch._custom_op.impl.get_ctx()
```
zou3519: isn't there a torch.library api for this

test/test_opaque_obj_v2.py
```diff
@@ -11,24 +19,39 @@ def __init__(self, queue: list[torch.Tensor], init_tensor_: torch.Tensor) -> Non
...
+        self._size_counter += 1
         return len(self.queue)
 
 
+class RNGState:
+    def __init__(self, seed):
+        self.rng = random.Random(seed)
+
+
+register_opaque_type(OpaqueQueue, "_TestOpaqueObject_OpaqueQueue")
```
zou3519, line 47: What's the point of passing a name? If the answer is that this is what shows up in the schema, then we're going to need to figure out how to avoid collisions... (not in this PR).

Ideally the string in the schema is just the fully qualified name for the class.

zou3519 (approved):

test/test_opaque_obj_v2.py
angelayi: Ah yeah that's the string that is used for the schema. Right now if `name` is not specified then I used `cls.__name__` ([code](https://github.com/pytorch/pytorch/blob/3f5401020b3fd66238a5eda31576b898d086e14e/torch/_library/opaque_object.py#L167-L168)) which I thought is the fully qualified name for the class?  [The registration function](https://github.com/pytorch/pytorch/blob/3f5401020b3fd66238a5eda31576b898d086e14e/torch/csrc/jit/frontend/schema_type_parser.cpp#L51-L58) errors if there are duplicate name registrations

PR #165020, opened 2025-10-09
[inductor] custom_graph_pass.get_hash_for_files: don't hash paths

masnesral:
* __->__ #165020

Summary: We have an internal user where caching broke because the paths that are unzipped are probably different per host. We can't think of a use case where a path change matters when the file content has not changed, so removing this part

### DIFF
--- a/torch/_inductor/custom_graph_pass.py
+++ b/torch/_inductor/custom_graph_pass.py
@@ -92,7 +92,9 @@ CustomGraphPassType: TypeAlias = Optional[
 
 
 @lru_cache(1)
-def get_hash_for_files(paths: tuple[str], extra: str = "") -> bytes:
+def get_hash_for_files(
+    paths: tuple[str], extra: str = "", hash_paths: bool = True
+) -> bytes:
     """
     Helper to compute a unique string by hashing the contents of a list of files.
     """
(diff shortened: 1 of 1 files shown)


torch/_inductor/custom_graph_pass.py
```diff
@@ -92,15 +92,18 @@ def uuid(self) -> Optional[Any]:
...
     """
     Helper to compute a unique string by hashing the contents of a list of files.
     """
     hasher = hashlib.sha256()
     hasher.update(extra.encode("utf-8"))
     for path in paths:
         with open(path, "rb") as f:
-            hasher.update(path.encode("utf-8"))
+            if hash_paths:
+                hasher.update(path.encode("utf-8"))
```
oulgen: i dont think this is good, the path can differ based on container/filesystem etc. Perhaps what you want is relative path to some root?

masnesral: @oulgen Yeah, the path can differ so I'm offering the option to _not_ consider the path. Do you mean that you think it's still important to hash some fragment of the path instead of the full path (rather than ignoring the path)? Is there a use case where the path changes, but the content does not, and we want the cache key to change?

oulgen: ah i'm confused, why did we care about the filepath in the first place?

oulgen (approved):

torch/_inductor/custom_graph_pass.py
masnesral: Yeah, in hindsight ... :/

masnesral: @oulgen maybe we should just remove the path instead of giving the option?? I kinda doubt anyone from OSS is relying on it and for internal it can only spell trouble. WDYT?

oulgen: yea lets kill it

PR #165023, opened 2025-10-09
Fix channels-last dimension mapping in CUDA parallel_cat

ngimel:
Fixes #164849
`dimension` was updated in-place, so for more than one batch of channels-last tensors the concat `dimension` for the second kernel launch was wrong

## Testing
- python -m compileall test/test_tensor_creation_ops.py

------
https://chatgpt.com/codex/tasks/task_e_68e708879b30832f89b10ae55faa68e8


ezyang (approved):
I verified all usages of dimension were updated

PR #165035, opened 2025-10-09
Remove unnecessary "static" for definitions in anonymous namespace

cyyever:
This PR removes unnecessary "static" for C++ functions and variables in anonymous namespace as detected by clang-tidy. This enhances code readability. The related rules are planed to be enabled in follow-up PRs.

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @mcarilli @ptrblck @leslie-fang-intel

### DIFF
--- a/c10/util/flags_use_no_gflags.cpp
+++ b/c10/util/flags_use_no_gflags.cpp
@@ -15,7 +15,7 @@ using std::string;
 C10_DEFINE_REGISTRY(C10FlagsRegistry, C10FlagParser, const string&)
 
 namespace {
-static bool gCommandLineFlagsParsed = false;
+bool gCommandLineFlagsParsed = false;
 // Since flags is going to be loaded before logging, we would
 // need to have a stringstream to hold the messages instead of directly
 // using caffe logging.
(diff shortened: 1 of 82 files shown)


c10/util/flags_use_no_gflags.cpp
```diff
@@ -15,15 +15,15 @@ using std::string;
...
+bool gCommandLineFlagsParsed = false;
 // Since flags is going to be loaded before logging, we would
 // need to have a stringstream to hold the messages instead of directly
 // using caffe logging.
 std::stringstream& GlobalInitStream() {
   static std::stringstream ss;
   return ss;
 }
-static const char* gUsageMessage = "(Usage message not set.)";
+const char* gUsageMessage = "(Usage message not set.)";
```
Skylion007, line 26: This should be constexpr, but probably should be another PR.

Skylion007 (approved):
Can you also enable the clang-tidy check for this PR?

c10/util/flags_use_no_gflags.cpp
cyyever: Let's put it in another PR

PR #165036, opened 2025-10-09
[Inductor][Grouped Gemm] Add Blackwell CuTeDSL Kernel

NikhilAPatel:
Make sure you're on cutlass 4.2.0+

Test Plan:
Tritonbench(oss):
`clear; CUDA_VISIBLE_DEVICES=2 TRITON_PRINT_AUTOTUNING=1 TRITON_ALWAYS_COMPILE=1 TORCH_LOGS=+inductor TORCHINDUCTOR_FORCE_DISABLE_CACHES=1 TORCHINDUCTOR_MAX_AUTOTUNE_GEMM=1 python run.py --op grouped_gemm --only aten_grouped_mm,preprocessed_pt2_triton_grouped_mm --precision bf16  --num-inputs 1 --metrics tflops,accuracy`

Unit Tests(oss):
`clear; python test/inductor/test_cutedsl_grouped_mm.py`

Differential Revision: D82010227


torch/_inductor/kernel/flex/common.py
```diff
@@ -337,9 +337,10 @@ def next_power_of_two(n):
 _TEMPLATE_DIR = Path(__file__).parent / "templates"
 
 
-def load_template(name: str) -> str:
+def load_template(name: str, template_dir: Path | None = None) -> str:
```
drisspg: we can move this up a level an maybe put somehwere in kernel common?

drisspg: also can you add a note about having to add templates to setup.py

alexsamardzic (approved):
This is awesome!  Looking forward to see the performance numbers.

EDIT:  Grabbed a B100 over a short period of time, some further comments:

1. Would be good to update `test_grouped_gemm_compiled` from `test/test_matmul_cuda.py` with this kernel as an option, as this test has some nasty offsets, and on a quick try it seems the kernel won't pass (2d/3d "op", both A and B "row-major" and max-autotune turned on).  Or, even better: to merge mentioned test and tests from `test/inductor/test_cutedsl_grouped_mm.py` into single place, to test with union of shapes/offsets, and to test both back-ends.
2. Tried to do a quick benchmark with the script pasted below, but latencies reported by this script for CuTe DSL are very large - apparently, proper compilation/tuning are yet to be implemented?   (Triton kernel speedups are low on Blackwell too, 0.1 and below for larger M, obviously for Blackwell and later, kernels will have to be written in a lower-level language.)
3. @ngimel: Not related to this PR, but I think "sm100a" should be added in the list [here](https://github.com/pytorch/pytorch/blob/a2601630cd9850ec68da8456d9f9584da9c0b0e8/cmake/Codegen.cmake#L132)?

<details>
<summary>Benchmarking script</summary>

```Python
import pandas as pd

from tqdm import tqdm

import torch
from triton.testing import do_bench


device = "cuda"
dtype_AB = torch.bfloat16
dtype_offset = torch.int32
dtype_C = torch.bfloat16
group_size = 4
strided = True

f_ref = torch._grouped_mm
f = torch.compile(
    f_ref,
    options={
        "max_autotune": True,
        "max_autotune_gemm_backends": "CUTE",
        "autotune_fallback_to_aten": False,
    },
)


def benchmark_microseconds(f, *args, **kwargs):
    return do_bench(lambda: f(*args, **kwargs), return_mode="median") * 1e3


def get_problem(M, N, K):
    A = torch.randn(M * group_size, K, device=device).to(dtype_AB)
    B = torch.randn(group_size, N, K, device=device).to(dtype_AB)
    offs = torch.arange(M, group_size * M + 1, M, device=device, dtype=dtype_offset)

    return A, B, offs


def benchmark(m: int, k: int, n: int):
    for i in range(5):
        A, B, offs = get_problem(m, n, k)
        C = f(A, B.transpose(-2, -1), offs, out_dtype=dtype_C)
        C_ref = f(A, B.transpose(-2, -1), offs, out_dtype=dtype_C)
        torch.testing.assert_close(C_ref, C)

    A, B, offs = get_problem(m, n, k)

    eager_time = benchmark_microseconds(
        f_ref,
        A,
        B.transpose(-2, -1),
        offs,
        out_dtype=dtype_C,
    )
    compiled_time = benchmark_microseconds(
        f,
        A,
        B.transpose(-2, -1),
        offs,
        out_dtype=dtype_C,
    )

    return {
        "m": m,
        "n": n,
        "k": k,
        "Eager (CUTLASS) latency (us)": eager_time,
        "Compiled (CuTe) latency (us)": compiled_time,
        "Compiled speedup (d/s)": eager_time / compiled_time,
    }


if __name__ == "__main__":
    k_vals = (4096, 8192)
    n_vals = (4096, 8192)
    results = []
    i_range = range(8, 14)
    for m in tqdm([1 << i for i in i_range]):
        for n, k in zip(n_vals, k_vals):
            results.append(benchmark(m, k, n))

        df = pd.DataFrame(results)
        df.to_csv(f"grouped_mm_2d_3d_time_results_cute.csv", index=False)
        print(df.to_markdown(index=False))
```
</details>

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
```diff
@@ -0,0 +1,287 @@
...
+            from_dlpack(input_b[0].unsqueeze(-1)),
+            from_dlpack(out_ptr0.unsqueeze(-1)),
+            probs,
+            strides,
+            ptrs,
+            tensormap_cute,
+            stream,
+        )
+
+        cache["{{kernel_name}}"] = (build_group_ptrs_jit_executor, group_gemm_jit_executor, total_num_clusters, tensormap_shape)
```
NikhilAPatel: cc @drisspg 
Right now we do a lot of manual caching since we don’t want to recompute total_num_clusters. Within an instantiated file, this number shouldn’t change assuming the shapes/strides of the inputs don’t change. 

Right now, the cache key is just "{{kernel_name}}" (which kind of means that this specific cache instance only ever has one key, since I think this is fixed when Inductor actually instantiates this file). But if its possible for Inductor to reuse this file for multiple different shapes, then we have to add that info to the cache key as well.

drisspg: for posterity, talked today, currently since we splat out a new file and thus a new cache = {} for all the constexprs we dont need to actually add any of these to the cache key. We also essentially have 1 cache per kernel name so don't need to include.

Empirically the kernels does specialize on strides and will get bad results we dont recompile, for the preproc kernel. 

We are going to only enable for static shapes and then cache all the things so expect lots of recomptue at first and then will follow up w/ more robust dynamic shape caching

test/inductor/test_cutedsl_grouped_mm.py
```diff
@@ -0,0 +1,149 @@
...
+from torch._inductor import config
+from torch._inductor.codegen.cuda.cuda_env import is_datacenter_blackwell_arch
+from torch._inductor.test_case import run_tests, TestCase
+from torch.testing._internal.common_utils import (
+    instantiate_parametrized_tests,
+    parametrize,
+)
+
+
+try:
```
drisspg: we can probaly make one of these https://github.com/pytorch/pytorch/blob/7778a58e7c3a9dfca8c4fa00d936581e7549d918/torch/_inductor/kernel/flex/flex_flash_attention.py#L23

for cutedsl

drisspg: and put in inductor

torch/_inductor/kernel/flex/flex_attention.py
```diff
@@ -50,6 +51,8 @@
 aten = torch.ops.aten
 Expr = sympy.Expr
 
+_TEMPLATE_DIR = Path(__file__).parent / "templates"
```
drisspg: rebasing on the other PR will make this easier to read lol

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
ngimel: The dimensions should be reasonably static though? Except possibly for M, though M with dropped tokens should be static too? Also can we have more detailed comments about CuTE caching, currently it's behavior is pretty opaque

drisspg: Agreed

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
```diff
@@ -0,0 +1,332 @@
...
+    base_C_u64 = int(out_ptr0.data_ptr())
+
+    ptrs_t = torch.empty((G, 3), device=dev, dtype=torch.int64)
+    probs_t = torch.empty((G, 4), device=dev, dtype=torch.int32)
+    strides_t = torch.empty((G, 3, 2), device=dev, dtype=torch.int32)
+    ptrs = from_dlpack(ptrs_t)
+    probs = from_dlpack(probs_t)
+    strides = from_dlpack(strides_t)
+
+    cache_key = get_cache_key("{{kernel_name}}", input_a, input_b, out_ptr0)
```
ngimel: btw do we need cache_key as a string, or can we directly use `tuple(input_a.shape)` etc? In my benchmarks the latter brings get_cache_key down from ~2.5us to sum-microsecond

NikhilAPatel: Nice find! I'll make the change

test/inductor/test_cutedsl_grouped_mm.py
```diff
@@ -0,0 +1,146 @@
...
+import torch
+from torch import Tensor
+from torch._inductor import config
+from torch._inductor.codegen.cuda.cuda_env import is_datacenter_blackwell_arch
+from torch._inductor.test_case import run_tests, TestCase as InductorTestCase
+from torch._inductor.utils import ensure_cute_available
+from torch.testing._internal.common_utils import (
+    instantiate_parametrized_tests,
+    parametrize,
+)
```
eellison, line 15: Do we need to add this test to b200 test files ?

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
```diff
@@ -0,0 +1,329 @@
...
+@functools.lru_cache
+def get_hardware_info():
+    hw = cutlass.utils.HardwareInfo()
+    sm_count = hw.get_max_active_clusters(1)
+    max_active_clusters = hw.get_max_active_clusters(CLUSTER_M * CLUSTER_N)
+
+    return (sm_count, max_active_clusters)
+
+
+def get_cache_key(input_a: torch.Tensor, input_b: torch.Tensor, out_ptr0: torch.Tensor):
```
eellison: does this also need to include alignment ?

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
```diff
@@ -0,0 +1,329 @@
...
+# Finally, note the @lru_cache decorator on get_hardware_info(). Empirically,
+# hw.get_max_active_clusters() triggers significant MLIR recompilation overhead,
+# despite depending only on the GPU type. We cache this function to mitigate
+# redundant recompiles even when shape/stride/dtype cache misses force kernel
+# regeneration. A follow-up study will investigate the root cause.
+
+cache = {}
+
+@functools.lru_cache
+def get_hardware_info():
```
eellison, line 47: nit: in the future, we'll want to factor out a lot of this file. is this already applicable with flex attention cc @drisspg ?

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
```diff
@@ -0,0 +1,329 @@
...
+    ptrs_t = torch.empty((G, 3), device=dev, dtype=torch.int64)
+    probs_t = torch.empty((G, 4), device=dev, dtype=torch.int32)
+    strides_t = torch.empty((G, 3, 2), device=dev, dtype=torch.int32)
+    ptrs = from_dlpack(ptrs_t)
+    probs = from_dlpack(probs_t)
+    strides = from_dlpack(strides_t)
+
+    cache_key = get_cache_key(input_a, input_b, out_ptr0)
+
+    if cache_key not in cache:
```
eellison: how long does it take to compile ? would we want to interleave the compilation in the future?

torch/_inductor/utils.py
```diff
@@ -1870,6 +1870,77 @@ def use_triton_blackwell_tma_template(
...
+
+    layout_dtypes = [torch.bfloat16]
+    if not _use_template_for_gpu(layout, layout_dtypes):
+        return False
+
+    if not (config.max_autotune or config.max_autotune_gemm):
+        return False
+
+    # Checks for 16B ptr and stride alignment
+    if not can_use_tma(mat_a, mat_b, output_layout=layout):
```
eellison, line 1971: i guess aligned is already constrained here

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
```diff
@@ -0,0 +1,329 @@
...
+    dev = input_a.device
+
+    base_A_u64 = int(input_a.data_ptr())
+    base_B_u64 = int(input_b.data_ptr())
+    base_C_u64 = int(out_ptr0.data_ptr())
+
+    ptrs_t = torch.empty((G, 3), device=dev, dtype=torch.int64)
+    probs_t = torch.empty((G, 4), device=dev, dtype=torch.int32)
+    strides_t = torch.empty((G, 3, 2), device=dev, dtype=torch.int32)
+    ptrs = from_dlpack(ptrs_t)
```
eellison, line 204: why do we need `from_dlpack ` , just wondering?

eellison (commented):
cool ! deferring to @drisspg and others for final review

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
```diff
@@ -0,0 +1,329 @@
...
+# Since this template currently targets static shapes, expect high recompute
+# frequency initially. Dynamic-shape-aware caching will be introduced later.
+#
+# Finally, note the @lru_cache decorator on get_hardware_info(). Empirically,
+# hw.get_max_active_clusters() triggers significant MLIR recompilation overhead,
+# despite depending only on the GPU type. We cache this function to mitigate
+# redundant recompiles even when shape/stride/dtype cache misses force kernel
+# regeneration. A follow-up study will investigate the root cause.
+
+cache = {}
```
eellison: in a follow up, an we try to factor this to work better with local, persistent disc caching ? similarly, do we cache across multiple invocations of same shape in the graph?

test/inductor/test_cutedsl_grouped_mm.py
NikhilAPatel: Yeah we probably do I'll add it.

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
NikhilAPatel: This kernel assumes 16B alignment. We verify this at compilation time. Do you know if this guarantees that tensors will be aligned at runtime as well?

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
NikhilAPatel: Yes I agree. This is one of the major followups planned.

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
NikhilAPatel: For a kernel like this, its on the order of 2ms. Yeah a more efficient method of compilation and caching is something we're planning to immediately follow up on.

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
NikhilAPatel: This is to convert from PyTorch tensors to CuTe tensors. You can see more information here: https://docs.nvidia.com/cutlass/media/docs/pythonDSL/cute_dsl_general/framework_integration.html

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
NikhilAPatel: Yeah this is one of the major followup areas I'm planning on working on next. We want to enable dynamic shapes, and a better, more robust caching system will be essential for that. Maybe we can chat a bit after PTC to discuss what the caching system for the cutedsl backend should look like.

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
eellison: wow that's pretty quick! if it's on the order of 2ms, not super high pri.

NikhilAPatel: Wait actually I was mistaken its around 800ms, so a lot longer

ngimel: Is it for all the autotuning instantiations or just one?

NikhilAPatel: Just a single instantiation

.ci/pytorch/test.sh
```diff
@@ -337,7 +337,7 @@ test_python() {
 
 test_python_smoke() {
   # Smoke tests for H100/B200
-  time python test/run_test.py --include test_matmul_cuda test_scaled_matmul_cuda inductor/test_fp8 inductor/test_max_autotune $PYTHON_TEST_EXTRA_OPTION --upload-artifacts-while-running
+  time python test/run_test.py --include test_matmul_cuda test_scaled_matmul_cuda inductor/test_fp8 inductor/test_max_autotune inductor/test_cutedsl_grouped_mm $PYTHON_TEST_EXTRA_OPTION --upload-artifacts-while-running
```
drisspg, line 340: One note is that this actually wont testing anything because we won't have cutedsl install I will fix when I figure out packaging

test/inductor/test_cutedsl_grouped_mm.py
```diff
@@ -0,0 +1,146 @@
...
+        self,
+        group_size: int,
+        M_hint: int,
+        K: int,
+        N: int,
+        device: str,
+        dtype: torch.dtype,
+    ) -> tuple[Tensor, Tensor, Tensor]:
+        # --- Random, tile-aligned M sizes ---
+        M_sizes = (
```
drisspg, line 35: can you make alignment an arg to the func w/ default val

drisspg: Also is 64 a requirement or is 16 bytes?

test/inductor/test_cutedsl_grouped_mm.py
```diff
@@ -0,0 +1,146 @@
...
+            }
+        ):
+            grouped_gemm_compiled = torch.compile(
+                grouped_gemm_fn, backend="inductor", dynamic=False
+            )
+            c_compiled = grouped_gemm_compiled(A, B, offsets)
+
+        self.assertEqual(c_eager.dtype, dtype)
+        self.assertEqual(c_compiled.dtype, dtype)
+        torch.testing.assert_close(c_eager, c_compiled, rtol=1e-2, atol=0.05)
```
drisspg: how does this comapare to other compiled v eager numerics for grouped_mm?

test/inductor/test_cutedsl_grouped_mm.py
```diff
@@ -0,0 +1,146 @@
...
+        offsets = torch.tensor(
+            [sum(M_sizes[: i + 1]) for i in range(G)], dtype=torch.int32, device=device
+        )
+
+        A = torch.randn(sum_M, K, device=device, dtype=dtype)
+
+        if layout_A == "offset":
+            # allocate bigger buffer than needed, use nonzero storage offset
+            storage = torch.randn(sum_M * K + 512, device=device, dtype=dtype)
+            offset = 128  # skip first 128 elements
```
drisspg, line 108: if you choose a more unfriendly number what happends?

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
drisspg: yup to both :)

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
drisspg: My quick pass makes me think that the copy_ alignment path will also just `just work` for inputs to a graph w/ cutedsl templates but feel free to correct me here @eellison

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
```diff
@@ -0,0 +1,329 @@
...
+
+        # --- Total clusters ---
+        def compute_total_num_clusters(
+            problem_sizes_mnkl,
+            cluster_tile_shape_mn,
+        ):
+            total_num_clusters = 0
+            for m, n, _, _ in problem_sizes_mnkl:
+                num_clusters_mn = tuple(
+                    (x + y - 1) // y for x, y in zip((m, n), cluster_tile_shape_mn)
```
drisspg: nit: ceildiv

torch/_inductor/template_heuristics/cutedsl.py
```diff
@@ -0,0 +1,306 @@
...
+
+    configs: list[CuTeGemmConfig] = []
+
+    # Non-2cta configs
+    for tensormap_update_mode in [
+        utils.TensorMapUpdateMode.SMEM,
+        utils.TensorMapUpdateMode.GMEM,
+    ]:
+        for tile_m in [64, 128]:
+            for tile_n in tile_n_vals:
```
drisspg: nit: might be prettier for these triple for's if you use itertools product

torch/_inductor/template_heuristics/cutedsl.py
```diff
@@ -0,0 +1,306 @@
...
+
+    return configs
+
+
+def get_default_groupgemm_configs() -> list[CuTeGemmConfig]:
+    """
+    Returns the default configuration set for the Blackwell CuTeDSL Grouped GEMM kernel.
+    """
+
+    return [
```
drisspg: nit: this is kinda hard to grok, I can think of a few ways to rewrite but curious how you got these specific configs

torch/_inductor/utils.py
```diff
@@ -1870,6 +1870,77 @@ def use_triton_blackwell_tma_template(
...
+    offs: Optional[Any],
+    bias: Optional[Any],
+    scale_result: Optional[Any],
+) -> bool:
+    """
+    Returns True if we can use the blackwell kernel for grouped mm.
+    Required conditions:
+        1. CuTeDSL is available
+        2. We are on a blackwell arch
+        3. The dtype is bf16
```
drisspg, line 1946: does the template only support bf16?

torch/_inductor/config.py
```diff
@@ -534,6 +534,10 @@ def prologue_fusion_enabled() -> bool:
     "TORCHINDUCTOR_MAX_AUTOTUNE_FLEX_SEARCH_SPACE", "DEFAULT"
 ).upper()  # type: ignore[assignment]
 
+cutedsl_enable_autotuning: bool = (
```
drisspg, line 549: whats this for?

drisspg (approved):
Small nits and 1 comment, overall looks good

We should track somewhere, maybe open an issue? on the follow up tasks

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
eellison: Yes, it should just work, we've already checked alignment in precondition for emitting this.

test/test_public_bindings.py
```diff
@@ -363,6 +363,7 @@ def onerror(modname):
             "torch._inductor.codegen.cpp_micro_gemm",
             "torch._inductor.codegen.cpp_template_kernel",
             "torch._inductor.runtime.triton_helpers",
+            "torch._inductor.template_heuristics.cutedsl",
```
albanD: Any way we could make this module actually importable? It's weird to ship in our package something that looks like a submodule but should not be imported

NikhilAPatel: Yeah I think I can make that possible

test/inductor/test_cutedsl_grouped_mm.py
NikhilAPatel: 16B is the requirement. Though that refers to the base pointer of the tensor. This alignment can be arbitrary.

test/inductor/test_cutedsl_grouped_mm.py
NikhilAPatel: Removed the manual overrides of rtol and atol and switched to the default for bf16 used by `torch.testing.assert_close` (rtol=1.6e-2, atol=1e-5). These are the same defaults used by `test_grouped_gemm_compiled` in test_matmul_cuda. We still pass, so numerics seem good.

test/inductor/test_cutedsl_grouped_mm.py
NikhilAPatel: This number must be 16B aligned for the kernel to be enabled as an option. It appears even eager mode requires this.

torch/_inductor/template_heuristics/cutedsl.py
NikhilAPatel: Yeah planning to rewrite this file to match the structure of the rest of the template heuristics. 

These configs were found by exhaustively autotuning over some Qwen3 shapes. Since then, I found some more shapes from llama and some other models, so after this PR is landed I plan to exhaustively tune over those too and converge to a small set of broadly performant configs. This is a very similar approach to how the other sets of default configs in Inductor are determined. But I agree with Natalia and others that maybe a heuristics based approach would be better to develop in the future.

torch/_inductor/utils.py
NikhilAPatel: It also supports fp16. But right now I believe that torch compile only accepts bf16 for groupgemm. So I was planning on opening up support for this whole path in the future.

torch/_inductor/config.py
NikhilAPatel: Right now autotuning is gated behind a config variable – CUTEDSL_ENABLE_AUTOTUNING – since I find that turning it on results in a flurry of call cuLaunchKernel/cuLaunchKernelEx failed, got error code: 1 errors. It still seems, however, that autotuning finishes successfully, so I’m still digging into why this occurs and what exactly the implications are. Soon I hope this can be enabled by default.

test/inductor/test_cutedsl_grouped_mm.py
```diff
@@ -0,0 +1,148 @@
...
+
+        # --- eager ---
+        c_eager = grouped_gemm_fn(A, B, offsets)
+
+        # --- compiled (CUTE backend) ---
+        with config.patch(
+            {
+                "max_autotune": True,
+                "max_autotune_gemm_backends": "CUTEDSL",
+                "test_configs.autotune_choice_name_regex": "cutedsl",
```
mlazos, line 139: hmm are there other choices that get added when you have only the single backend?

test/inductor/test_cutedsl_grouped_mm.py
```diff
@@ -0,0 +1,148 @@
...
+            )
+            c_compiled = grouped_gemm_compiled(A, B, offsets)
+
+        self.assertEqual(c_eager.dtype, dtype)
+        self.assertEqual(c_compiled.dtype, dtype)
+        torch.testing.assert_close(c_eager, c_compiled)
+
+    @parametrize("layout_A", ("contiguous", "offset", "padded"))
+    @parametrize("layout_B", ("contiguous", "broadcasted"))
+    def test_grouped_gemm_assorted_layouts(
```
mlazos, line 87: Can you also test passing a view?

torch/_inductor/config.py
mlazos: Wouldn't we only add the choices if cutedsl is in the backends?

torch/_inductor/template_heuristics/cutedsl.py
mlazos: CUTLASS has an instantiation level which let's the space exhaustive space be controlled by the user, I also did something similar in my own GEMM work. Bucketing the search space could work for now, but also using the matmul heuristics package I mentioned in our 1:1 might be the best way to constrain the search space

torch/_inductor/template_heuristics/cutedsl.py
```diff
@@ -0,0 +1,293 @@
...
+        ),
+        CuTeGemmConfig(
+            TILE_M=128,
+            TILE_N=64,
+            CLUSTER_M=1,
+            CLUSTER_N=1,
+            USE_2_CTA=False,
+            TENSORMAP_UPDATE_MODE=TensorMapUpdateMode.GMEM,
+        ),
+        CuTeGemmConfig(
```
mlazos: nit: for cleaning up the code a bit can you make these a list of tuples and then apply the constructor to the unpacked tuples? e.g. ([CuTeGemmConfig(*args) for args in configs]) and if there's a pattern can you autogenerate?

test/inductor/test_cutedsl_grouped_mm.py
NikhilAPatel: No just the cutedsl kernel

torch/_inductor/kernel/templates/cutedsl_mm_grouped.py.jinja
```diff
@@ -0,0 +1,331 @@
...
+        )
+        tensormap_workspace = from_dlpack(tensormap_workspace_t)
+
+        # --- Total clusters ---
+        def compute_total_num_clusters(
+            problem_sizes_mnkl,
+            cluster_tile_shape_mn,
+        ):
+            total_num_clusters = 0
+            for m, n, _, _ in problem_sizes_mnkl:
```
ngimel: how does this work for further cached invocations that have different m values? `m` are known on device only, and they can change even if the cache key is found, based on the `offs` value

NikhilAPatel: Yes it appears this was an oversight. I've updated the caching behavior such that now we maintain 2 separate caches for the preprocessing kernel and the gemm kernel. And we also recompute `total_num_clusters` on the host to verify that we can use the cached gemm executor. There's a more detailed explanation of the caching behavior in a comment in the jinja template file. 

Thanks for asking the Cutlass people if they can create a version of this kernel that allows all this information to remain completely on the gpu! Hopefully they finish that soon and we can update this behavior. In followups I'll also look into ways of mitigating the cost of this sync.

alexsamardzic:
In the meantime I came up with [this](https://github.com/pytorch/pytorch/pull/166063), while experimenting with Gluon; I've included this CuTe DSL kernel into benchmarking.  The CuTe DSL kernel is close to CUTLASS `torch._grouped_mm` with performance, but it seems there is a whole load of elementwise operators called before the MM kernel, and after `build_group_ptrs_from_bases` kernel, that seem to take almost the same time as the MM kernel itself.   Attached is a screenshot of profiler trace in Perfetto, maybe worth checking.

perfetto

mlazos (approved):

ngimel:
@NikhilAPatel did you have a chance to look into @alexsamardzic comment? What is causing the elementwise kernels before grouped_mm?

mlazos:
Also did you figure out the slowdown @NikhilAPatel ? The one we talked about in our 1:1, not the elementwise kernels.

PR #165046, opened 2025-10-09
Enable C407 of flake8

cyyever:
This PR enables C407 on flake8. The description is `C407` is `Unnecessary list comprehension - ‘<builtin>’ can take a generator`.


cyyever:
It is indeed covered so this PR simply removes the suppression from flake8 config.

albanD (approved):
Nice

PR #165048, opened 2025-10-09
[NVIDIA] Support CUDSS for CUDA 13 and Blackwell

johnnynunez:
Bump CUDSS to stable release

cc @Skylion007 @tinglvv @nWEIdia

changelog: [cuDSS v0.7.0](https://docs.nvidia.com/cuda/cudss/release_notes.html)

### DIFF
--- a/.ci/docker/common/install_cudss.sh
+++ b/.ci/docker/common/install_cudss.sh
@@ -5,20 +5,19 @@ set -ex
 # cudss license: https://docs.nvidia.com/cuda/cudss/license.html
 mkdir tmp_cudss && cd tmp_cudss
 
-if [[ ${CUDA_VERSION:0:4} =~ ^12\.[1-4]$ ]]; then
-    arch_path='sbsa'
-    export TARGETARCH=${TARGETARCH:-$(uname -m)}
-    if [ ${TARGETARCH} = 'amd64' ] || [ "${TARGETARCH}" = 'x86_64' ]; then
-        arch_path='x86_64'
-    fi
-    CUDSS_NAME="libcudss-linux-${arch_path}-0.3.0.9_cuda12-archive"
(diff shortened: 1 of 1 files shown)


.ci/docker/common/install_cudss.sh
```diff
@@ -5,20 +5,18 @@ set -ex
...
-    tar xf ${CUDSS_NAME}.tar.xz
-    cp -a ${CUDSS_NAME}/include/* /usr/local/cuda/include/
-    cp -a ${CUDSS_NAME}/lib/* /usr/local/cuda/lib64/
+arch_path='sbsa'
+CUDA_MAJOR="${CUDA_VERSION%%.*}"
+export TARGETARCH=${TARGETARCH:-$(uname -m)}
+if [ ${TARGETARCH} = 'amd64' ] || [ "${TARGETARCH}" = 'x86_64' ]; then
+    arch_path='x86_64'
 fi
+CUDSS_NAME="libcudss-linux-${arch_path}-0.7.0.20_cuda${CUDA_MAJOR}-archive"
```
Skylion007, line 14: Hmm, we only statically link currently? @atalman We should probably switch to this dynamic linking and use the pypi wheel at some point. We would need to add the pypi wheel dependency, change the CMake flags, and then finally add the relative path of the pypi wheel to our RPATH modification list in our build scripts. https://pypi.org/project/nvidia-cudss-cu12/ https://pypi.org/project/nvidia-cudss-cu13/ . Looks like potentially up to 70mb of space savings in our wheel.

johnnynunez: i don't know, i tried to use cudss and failed, and i've seen that it was not updated for cuda 13. So i updated to support it. Feel free to take any action here

Aidyn-A: Right, it does not work for quite some time: https://github.com/pytorch/pytorch/issues/164122

johnnynunez: @Aidyn-A i didn't see that, thanks! I was trying to use to accelerate SLAM for Robotics

PR #165056, opened 2025-10-09
Add float16 and bfloat16 support to sparse CSR sampled_addmm

stmcgovern:
Fixes #164666

- Add bfloat16 support via upcasting to float32 (cuSPARSE limitation)
- Add native float16 support via opmath_t 
- Enable tests

Float16 uses native cuSPARSE SDDMM with float32 accumulation. BFloat16 is not supported by cuSPARSE SDDMM, so we upcast to float32.

### DIFF
--- a/aten/src/ATen/native/sparse/cuda/SparseBlas.cpp
+++ b/aten/src/ATen/native/sparse/cuda/SparseBlas.cpp
@@ -45,6 +45,37 @@ Tensor& sparse_sampled_addmm_out_sparse_csr_cuda(
   at::native::sparse::sparse_sampled_addmm_check_inputs(
       self, mat1, mat2, beta, alpha, result);
 
+  // cuSPARSE SDDMM doesn't support BFloat16, so we need to upcast to Float32
+  //
+  // Performance Note: This workaround incurs significant memory overhead due to multiple copies:
+  // 1. Copy sparse tensor values from bfloat16 to float32 (self.to(at::kFloat))
+  // 2. Copy dense matrix mat1 from bfloat16 to float32 (mat1.to(at::kFloat))
+  // 3. Copy dense matrix mat2 from bfloat16 to float32 (mat2.to(at::kFloat))
+  // 4. Copy result values from float32 back to bfloat16 (result.values().copy_())
+  //
+  // This overhead is unavoidable given cuSPARSE limitations. Users requiring optimal
+  // performance should consider using float32 instead of bfloat16 for this operation.
+  if (self.scalar_type() == at::ScalarType::BFloat16) {
+    auto result_fp32 = at::empty({0, 0}, self.options().dtype(at::kFloat));
+    sparse_sampled_addmm_out_sparse_csr_cuda(
+        self.to(at::kFloat),
+        mat1.to(at::kFloat),
+        mat2.to(at::kFloat),
+        beta,
+        alpha,
+        result_fp32);
+    // Copy result back to bfloat16
+    if (&result != &self) {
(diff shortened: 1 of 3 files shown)


aten/src/ATen/native/sparse/cuda/SparseBlas.cpp
```diff
@@ -45,6 +45,37 @@ Tensor& sparse_sampled_addmm_out_sparse_csr_cuda(
...
+  // 4. Copy result values from float32 back to bfloat16 (result.values().copy_())
+  //
+  // This overhead is unavoidable given cuSPARSE limitations. Users requiring optimal
+  // performance should consider using float32 instead of bfloat16 for this operation.
+  if (self.scalar_type() == at::ScalarType::BFloat16) {
+    auto result_fp32 = at::empty({0, 0}, self.options().dtype(at::kFloat));
+    sparse_sampled_addmm_out_sparse_csr_cuda(
+        self.to(at::kFloat),
+        mat1.to(at::kFloat),
+        mat2.to(at::kFloat),
```
Aidyn-A, line 63: I do  not think this is a good idea. I believe it it would be more transparent is users have do it themselves, rather than making this unbeknownst to end users.

aten/src/ATen/native/sparse/cuda/SparseBlas.cpp
```diff
@@ -45,6 +45,37 @@ Tensor& sparse_sampled_addmm_out_sparse_csr_cuda(
   at::native::sparse::sparse_sampled_addmm_check_inputs(
       self, mat1, mat2, beta, alpha, result);
 
+  // cuSPARSE SDDMM doesn't support BFloat16, so we need to upcast to Float32
```
Aidyn-A, line 48: From the documentation I see that they do support `float16` and `bfloat16`: https://docs.nvidia.com/cuda/cusparse/#cusparse-types-reference
Only complex `float16` and complex `bfloat16` are deprecated.

test/test_sparse_csr.py
```diff
@@ -2410,8 +2410,12 @@ def remove_diagonal(t):
             run_test(n, k, upper, unitriangular, transpose, zero)
 
     @dtypes(torch.float32, torch.float64, torch.complex64, torch.complex128)
+    @dtypesIfCUDA(*floating_and_complex_types_and(
+                  *[torch.half] if SM53OrLater and TEST_CUSPARSE_GENERIC else [],
```
Aidyn-A, line 2414: PyTorch is no longer build for `sm_53`. Even latest CUDA versions (12+) do not support it.

test/test_sparse_csr.py
```diff
@@ -2460,6 +2464,12 @@ def run_test(c, a, b, op_a, op_b, *, alpha=None, beta=None):
                     run_test(c, a, b, op_a, op_b)
 
     @dtypes(torch.float32, torch.float64, torch.complex64, torch.complex128)
+    @dtypesIfCUDA(*floating_and_complex_types_and(
+                  *[torch.half] if SM53OrLater and TEST_CUSPARSE_GENERIC else [],
```
Aidyn-A, line 2468: PyTorch is no longer build for `sm_53`

aten/src/ATen/native/sparse/cuda/SparseBlas.cpp
stmcgovern: thanks for the feedback. What do you suggest for enabling bfloat16 dtype?

aten/src/ATen/native/sparse/cuda/SparseBlas.cpp
stmcgovern: thanks for the clarification!

PR #165060, opened 2025-10-09
Refactor _scaled_grouped_mm_cuda dispatch

slayton58:
* __->__ #165060

Summary:

* Clean & simplify different scaling recipe dispatch
* Split out recipes into separate dispatch functions

Test Plan:

```
pytest -svv -k grouped  test/test_scaled_matmul_cuda.py
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/aten/src/ATen/native/cuda/Blas.cpp
+++ b/aten/src/ATen/native/cuda/Blas.cpp
@@ -2546,79 +2546,27 @@ _scaled_mm_cuda_v2(
                       out);
 }
 
-Tensor
-_scaled_grouped_mm_cuda(const Tensor& mat_a, const Tensor& mat_b,
-const Tensor& scale_a, const Tensor& scale_b,
-const std::optional<at::Tensor>& offs,
-const std::optional<at::Tensor>& bias,
-const std::optional<at::Tensor>& scale_result,
-std::optional<c10::ScalarType> out_dtype,
-bool use_fast_accum) {
-  bool allowed_device = _scaled_mm_allowed_device(/*sm90_only*/true, /*sm100_only*/true);
-  TORCH_CHECK(allowed_device, "torch._scaled_grouped_mm is only supported on CUDA devices with compute capability = [9.0, 10.0], or ROCm MI300+");
-
-  TORCH_CHECK(!check_valid_strides_and_return_transposed(mat_a), "Expected mat1 to not be transposed");
-  TORCH_CHECK(check_valid_strides_and_return_transposed(mat_b), "Expected mat2 to be transposed");
-  TORCH_CHECK(mat_a.dim() == 2 || mat_a.dim() == 3, "mat_a has to be 2 or 3d");
-  TORCH_CHECK(mat_b.dim() == 2 || mat_b.dim() == 3, "mat_b has to be 2 or 3d");
-  const bool a_is_2d = mat_a.dim() == 2;
-  const bool b_is_2d = mat_b.dim() == 2;
-  if (!a_is_2d || !b_is_2d) {
-    TORCH_CHECK(mat_a.size(-1) == mat_b.size(-2), "contraction dimension of mat_a and mat_b must match");
-  }
-  TORCH_CHECK(
-    mat_a.size(-1) % 16 == 0,
-    "Expected trailing dimension of mat_a to be divisible by 16 ",
-    "but got mat1 shape: (",
-    mat_a.sizes(),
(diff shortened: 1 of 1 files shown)


aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -2546,51 +2546,184 @@ _scaled_mm_cuda_v2(
...
+        offs.value(),
+        out);
+#else
+    TORCH_CHECK_NOT_IMPLEMENTED(false, "mxfp8_mxfp8 grouped gemm requires compile with USE_FBGEMM_GENAI");
+#endif
+    return out;
+}
+
+// 2d-2d and 2d-3d cases
+// scaling=rowwise
```
danielvegamyhre, line 2584: should the name of the func make it clear it's using rowwise scaling, rather than relying on the comment? Or is the plan for this to also support tensorwise, blockwise eventually?

slayton58: I think I hadn't worked out what the support was when I wrote the name :) I'll add `_rowwise` to the name, it can always be changed if tensor-wise support is added later.

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -2610,31 +2743,22 @@ bool use_fast_accum) {
...
   }
-#endif
-
-#ifndef USE_ROCM
-  TORCH_CHECK(mat_a.dtype() == at::kFloat8_e4m3fn, "Expected mat_a to be Float8_e4m3 matrix got ", mat_a.scalar_type());
-  TORCH_CHECK(mat_b.dtype() == at::kFloat8_e4m3fn, "Expected mat_a to be Float8_e4m3 matrix got ", mat_b.scalar_type());
 
-  at::cuda::detail::f8f8bf16_grouped_mm(
+  // If we're not MXFP8, then we're row-wise scaling.
+  return _f8_f8_bf16_rowwise_grouped_mm(
```
danielvegamyhre, line 2761: optional nit but if we are using fbgemm style naming of "f8" instead of "fp8" for this func, maybe we should be consistent with mxfp8 and use the fbgemm style naming of `mx8_mx8_bf16`?  Would be more consistent and concise, up to you though

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -2546,51 +2546,184 @@ _scaled_mm_cuda_v2(
...
+  TORCH_CHECK_VALUE(check_valid_strides_and_return_transposed(mat_b), "Expected mat2 to be transposed");
+  TORCH_CHECK_VALUE(mat_a.dim() == 2 || mat_a.dim() == 3, "mat_a has to be 2 or 3d");
+  TORCH_CHECK_VALUE(mat_b.dim() == 2 || mat_b.dim() == 3, "mat_b has to be 2 or 3d");
   const bool a_is_2d = mat_a.dim() == 2;
   const bool b_is_2d = mat_b.dim() == 2;
+
+  // NOTE(slayton): For sub-1B formats want contraction_dim argument?
   if (!a_is_2d || !b_is_2d) {
-    TORCH_CHECK(mat_a.size(-1) == mat_b.size(-2), "contraction dimension of mat_a and mat_b must match");
+    TORCH_CHECK_VALUE(mat_a.size(-1) == mat_b.size(-2), "contraction dimension of mat_a and mat_b must match");
```
danielvegamyhre, line 2698: just curious why we're migrating TORCH_CHECK to TORCH_CHECK_VALUE, what is the difference?

danielvegamyhre (approved):
LGTM, @ngimel may want to review this as well

aten/src/ATen/native/cuda/Blas.cpp
slayton58: They raise different exceptions - `TORCH_CHECK_VALUE` raises `ValueError`, as-in "A runtime value is incompatible with what we want.

It was one of the feedbacks I got under review from the scaled_mm_v2 work - that we should be raising exceptions that matched well with the kind of errors we were checking for.

aten/src/ATen/native/cuda/Blas.cpp
slayton58: I went ahead and changed this to `_mx8_mx8_...` - might as well while I'm in there

ngimel (approved):

PR #165061, opened 2025-10-09
[pytorch] Composite backend potential fix for is_backend_available

wesbland:
Summary: `is_backend_available` takes in a string and expects it to only be backend, if its given a composite (device:backend) string, it fails.

Reviewed By: prashrock

Differential Revision: D81886736


kwen2501 (commented):
Are you proposing a UI like this?
```
is_backend_available("cuda:nccl")
```
May I know why a user would pass in `"cuda:nccl"` instead of just `is_backend_available("nccl")` ?

H-Huang (approved):
Seems reasonable to me

PR #165067, opened 2025-10-09
add device generalization support for distributed tests

harikodali:
## MOTIVATION
To generalize Distributed test cases for non-CUDA devices

## CHANGES
- Replaced hard coded device/backends with torch.accelerator.current_accelerator() and dist.get_default_backend_for_device
- Use DistributedTestBase instead of MultiProcessTestCase to use common utilities
- Remove instantiate_device_tests and make use of torch.accelerator.current_accelerator for test/distributed/test_c10d_object_collectives.py
- fix deterministic context issue for non-cuda devices in test/distributed/optim/test_zero_redundancy_optimizer.py
- use torch.accelerator.device_count() for multi-gpu check in torch/testing/_internal/distributed/_tensor/common_dtensor.py

### DIFF
--- a/test/distributed/optim/test_zero_redundancy_optimizer.py
+++ b/test/distributed/optim/test_zero_redundancy_optimizer.py
@@ -1236,12 +1236,15 @@ class TestZeroRedundancyOptimizerDistributed(TestZeroRedundancyOptimizer):
             )
         for model, inputs in models_to_test:
             # Select deterministic context based on device
+            if "cuda" not in device:
+                torch.use_deterministic_algorithms(True)
+
             det_ctx = (
                 torch.backends.cudnn.flags(
                     enabled=True, deterministic=True, benchmark=False
                 )
                 if "cuda" in device
-                else torch.use_deterministic_algorithms(True)
+                else nullcontext()
             )
             with det_ctx:
                 device_ids = [rank] if requires_ddp_rank(device) else None
--- a/test/distributed/optim/test_zero_redundancy_optimizer.py
+++ b/test/distributed/optim/test_zero_redundancy_optimizer.py
@@ -7,7 +7,7 @@
 
 import copy
 import sys
-from contextlib import nullcontext
+from contextlib import contextmanager, nullcontext
 from typing import Any, cast
 
 import numpy as np
@@ -40,7 +40,6 @@ from torch.testing._internal.common_distributed import (
     skip_if_rocm_multiprocess,
     skip_if_win32,
 )
-from torch.testing._internal.common_fsdp import get_devtype
 from torch.testing._internal.common_utils import (
     instantiate_parametrized_tests,
     parametrize,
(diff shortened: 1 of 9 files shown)


test/distributed/optim/test_zero_redundancy_optimizer.py
```diff
@@ -1236,12 +1236,15 @@ def _test_ddp_zero_overlap(
             )
         for model, inputs in models_to_test:
             # Select deterministic context based on device
+            if "cuda" not in device:
+                torch.use_deterministic_algorithms(True)
```
albanD: We shouldn't set it globally here and use it as a context manager as is done below.

albanD (commented):
One nit, but sgtm otherwise

test/distributed/optim/test_zero_redundancy_optimizer.py
harikodali: updated to use context manager in https://github.com/pytorch/pytorch/pull/165067/commits/740f43825d3bd350e5fd37091270b1c81f80e613

guangyey (dismissed):

albanD (dismissed):
SGTM

harikodali:
@cyyever @albanD looks like, some workflows need to be started by maintainers for merge, can you check

harikodali:
@albanD any issues with the PR? please provide next steps to get this PR merged

test/distributed/optim/test_zero_redundancy_optimizer.py
```diff
@@ -57,7 +56,17 @@
     HAS_TORCHVISION = False
 
 
-device_type = str(get_devtype())
+device_type = acc.type if (acc := torch.accelerator.current_accelerator()) else "cpu"
```
albanD: This one is missing the `check_available=True` flag in the current_accelerator() call.
Without it, running a cuda build on a cpu-only machine will try to run the tests and fail.

d4l3k:
@harikodali the root cause is due to running a CUDA build of the distributed tests on CPU machines.

Example behavior:

```
tristanr@devvm5553 ~/lintrunner (main)> CUDA_VISIBLE_DEVICES= python -c "import torch; print(torch.accelerator.current_accelerator())" 
cuda
tristanr@devvm5553 ~/lintrunner (main)> CUDA_VISIBLE_DEVICES= python -c "import torch; print(torch.accelerator.current_accelerator(check_available=True))"
None
```

Can you add `check_available=True` to everywhere you do

```
device_type = acc.type if (acc := torch.accelerator.current_accelerator()) else "cpu"
```

I.e.

```
device_type = acc.type if (acc := torch.accelerator.current_accelerator(check_available=True)) else "cpu"
```

The specific failure is test/distributed/optim/test_zero_redundancy_optimizer.py test_add_param_group but we should probably do this everywhere

test/distributed/optim/test_zero_redundancy_optimizer.py
harikodali: added check_available=True

albanD (approved):
Thanks for the update!

PR #165068, opened 2025-10-09
158232  Fix autocast cache incorrectly retaining no_grad state

### LINKED ISSUE #158232: Incompatibility between torch.no_grad, torch.autocast, and FSDP2
### 🐛 Describe the bug

Nesting `torch.no_grad` context and `torch.autocast` context when using FSDP2 with fp32 `param_dtype`, a critical incompatibility arises.

### Minimal Reproducible Code​​:
```python
import torch
from torch import nn
from torch.distributed.fsdp import fully_shard, MixedPrecisionPolicy

dist.init_process_group(backend="nccl")
torch.cuda.set_device(dist.get_rank())

m = nn.Linear(2, 2)

# setting both the following flags to True will cause the error
fail_condition1 = True
fail_condition2 = True

if fail_condition1:
    param_dtype = torch.float32
else:
    param_dtype = torch.bfloat16

m = fully_shard(m, mp_policy=MixedPrecisionPolicy(param_dtype=param_dtype))
m.requires_grad_(True)

inp = torch.randn(8, 2)

with torch.autocast("cuda", dtype=torch.bfloat16, enabled=True):
    if fail_condition2:
        with torch.no_grad():
            m(inp)
    m(inp).mean().backward()
```

### Error log
```
[rank0]: Traceback (most recent call last):
[rank0]:   File "/tmp/bug.py", line 30, in <module>
[rank0]:     m(inp).mean().backward()
[rank0]:   File "/usr/local/lib64/python3.12/site-packages/torch/_tensor.py", line 626, in backward
[rank0]:     torch.autograd.backward(
[rank0]:   File "/usr/local/lib64/python3.12/site-packages/torch/autograd/__init__.py", line 347, in backward
[rank0]:     _engine_run_backward(
[rank0]:   File "/usr/local/lib64/python3.12/site-packages/torch/autograd/graph.py", line 823, in _engine_run_backward
[rank0]:     return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
[rank0]:            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[rank0]: RuntimeError: element 0 of tensors does not require grad and does not have a grad_fn
```

Please note that the error arises ​​only when both conditions are met​​:

* Performing a forward pass first, which should be in the nested torch.no_grad-torch.autocast context
* FSDP's MixedPrecisionPolicy specifies param_dtype=torch.float32

### Versions

PyTorch version: 2.6.0+cu124
Is debug build: False
CUDA used to build PyTorch: 12.4

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @zhaojuanmao @mrshenli @rohan-varma @chauhang @mori360

stmcgovern:
Fixes #158232 
The autocast caching heuristic in `aten/src/ATen/autocast_mode.cpp:139` did not account for gradient mode state when deciding whether to cache. FSDP2 is not directly related. 

~~This PR adds `GradMode::is_enabled()` check to caching condition. Caching is now disabled in `no_grad()` contexts to prevent storing tensors with incorrect gradient state. Ensures correctness at the cost of using cache.~~
~~This PR proposes separate caches for gradient-enabled and gradient-disabled modes.~~
This PR uses a single cache and always creates cached tensors with grad enabled. Skips caching in inference_mode (not no_grad). Cached tensors work in both no_grad and grad contexts.
Adds tests.

cc @mcarilli @ptrblck @leslie-fang-intel @jgong5

### DIFF
--- a/aten/src/ATen/autocast_mode.cpp
+++ b/aten/src/ATen/autocast_mode.cpp
@@ -2,6 +2,8 @@
 
 #include <mutex>
 #include <ATen/CachedTensorUtils.h>
+#include <c10/core/GradMode.h>
+#include <c10/core/InferenceMode.h>
 #include <c10/util/flat_hash_map.h>
 
 namespace at::autocast {
@@ -121,10 +123,14 @@ Tensor cached_cast(at::ScalarType to_type, const Tensor& arg, DeviceType device_
   if (is_eligible(arg, device_type) && (arg.scalar_type() != to_type)) {
     // Heuristic:  Do what Apex does, and cache lower_precision_fp casts of fp32 model weights (leaves).
     // See cached_casts declaration above for detailed strategy.
+    //
+    // Fix #158232: Don't cache in inference_mode - those tensors can't be reused
+    // for training since enable_grad() cannot override inference_mode.
     bool can_try_cache = (to_type == get_lower_precision_fp_from_device_type(device_type) &&
                          arg.scalar_type() == at::kFloat && arg.requires_grad() &&
                          arg.is_leaf() && !arg.is_view() && cache_enabled &&
-                         !at::caching::is_cached_tensor(arg));
+                         !at::caching::is_cached_tensor(arg) &&
+                         !c10::InferenceMode::is_enabled());
 
     if (can_try_cache) {
       const std::lock_guard<std::mutex> lock(cached_casts_mutex);
@@ -132,6 +138,12 @@ Tensor cached_cast(at::ScalarType to_type, const Tensor& arg, DeviceType device_
       if (it != get_cached_casts().end()) {
         return std::get<1>(it->second);
       } else {
+        // Fix #158232: When caching in no_grad() context, we still need grad_fn
+        // on the cached tensor so it can be reused in grad-enabled contexts.
+        // The AutoGradMode RAII guard temporarily enables grad for .to() only.
+        // Note: arg.requires_grad() is guaranteed true here (checked in can_try_cache)
+        // and inference_mode is excluded above since enable_grad can't override it.
+        c10::AutoGradMode enable_grad(true);
         auto casted_arg = arg.to(to_type);
         get_cached_casts().emplace(arg.unsafeGetTensorImpl(), val_type{weakref_type(arg.getIntrusivePtr()), casted_arg});
         return casted_arg;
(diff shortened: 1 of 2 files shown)


aten/src/ATen/autocast_mode.cpp
```diff
@@ -121,10 +122,22 @@ Tensor cached_cast(at::ScalarType to_type, const Tensor& arg, DeviceType device_
...
+    // Performance note: This disables caching during inference (torch.no_grad() contexts),
+    // which likely will slightly reduce performance. This trade-off ensures
+    // correctness. Future optimization could implement gradient-mode-aware cache keys
+    // to allow caching in both modes separately.
     bool can_try_cache = (to_type == get_lower_precision_fp_from_device_type(device_type) &&
                          arg.scalar_type() == at::kFloat && arg.requires_grad() &&
                          arg.is_leaf() && !arg.is_view() && cache_enabled &&
-                         !at::caching::is_cached_tensor(arg));
+                         !at::caching::is_cached_tensor(arg) &&
+                         at::GradMode::is_enabled());  // Only cache when gradients are enabled
```
weifengpy: @janeyx99 : this unblocks fsdp2 training but regress inference perf. any opinions?

@stmcgovern great finding! any chance to not affecting inference? this can be a big perf change

stmcgovern: @weifengpy Thanks! Yes, I was just trying to highlight the root cause. One idea is to maintain separate caches for gradient-enabled and gradient-disabled modes.

weifengpy: got you - just being able to identify the root cause is a big step

ngimel (dismissed):

janeyx99 (commented):
The change looks good to me, but I don't feel great that we're just skipping dynamo tests. Do you know why those 3 tests run into KeyErrors?

At the very least, let's track the fact that dynamo doesn't work with the new autocast cache system in an issue and link to that issue in the test skips.

ngimel:
Dynamo/inductor already turn off autocast caching

janeyx99 (dismissed):
Oh. Then sounds good to me lol

ngimel (approved):

PR #165074, opened 2025-10-09
 Support python slicing with tensor inputs.

laithsakka:
* __->__ #165074

when the slice is tensor, we decompose it to .item() call and pass the unbacked symbol to the slice to avoid DDE. 
the diff also fix an existing bug in codegen_dynamic_slice_size in the cpp wrapper.  a +1 should be -1 making it match 
python codegen.


tools/autograd/templates/python_variable_methods.cpp
```diff
@@ -322,7 +322,8 @@ template<typename T>
 static T dispatch_to(const Tensor & self) {
   pybind11::gil_scoped_release no_gil;
   OptionalDeviceGuard device_guard(device_of(self));
-  TORCH_CHECK_VALUE(self.sym_numel() == 1, "only one element tensors can be converted to Python scalars");
+  auto cond = sym_eq(self.sym_numel(), 1);
```
laithsakka: maybe this not needed actually

laithsakka:
great i just need to add unit tests!

laithsakka:
AI told me ....
idk if this is ok 

It is not always the case that calling .item() on a size-1 tensor preserves the semantics of slicing, because there is a crucial difference in how the resulting value is handled: its impact on the computational graph. 

The key difference: gradient tracking

When you use a size-1 tensor directly for slicing, the slicing operation itself is non-differentiable. Any upstream gradient information from the tensor used for the index is discarded, as you cannot differentiate with respect to a discrete index.

When you call.item()on the tensor first, you explicitly break the computational graph. The single value is extracted as a standard Python number, and any gradient information associated with that tensor is completely lost. 

For the purpose of the slice operation itself, the two methods are semantically equivalent—the same slice is performed. The difference lies in what happens if you later compute gradients on a loss function that depends on the result of the slicing operation.

torch/_dynamo/variables/lists.py
```diff
@@ -1399,18 +1399,27 @@ def __init__(self, items, **kwargs) -> None:
...
-                hints=[
-                    *graph_break_hints.SUPPORTABLE,
-                ],
+        # Convert TensorVariable to SymIntVariable by calling .item()
+        # This decomposes a[:t] to u=t.item(); a[:u] at the dynamo level
+        if isinstance(start, variables.TensorVariable):
+            assert tx is not None, (
+                "tx is required when slice indices are TensorVariables"
             )
+            assert start.size is None or all(s == 1 for s in start.size)
```
Lucaskabela: What is expected currently if `dynamic=True` for the unit tests?  Can we add a few to check?

torch/_dynamo/variables/lists.py
```diff
@@ -1399,18 +1399,27 @@ def __init__(self, items, **kwargs) -> None:
...
+            start = start.call_method(tx, "item", [], {})
+        if isinstance(stop, variables.TensorVariable):
+            assert tx is not None, (
+                "tx is required when slice indices are TensorVariables"
+            )
+            assert stop.size is None or all(s == 1 for s in stop.size)
```
Lucaskabela: I think we are moving away from asserts in the variables code, let's avoid them here - see https://github.com/pytorch/pytorch/pull/165430 for reference

Lucaskabela (commented):
Generally this looks good, to me but needs some cleanup of asserts and I would like to see a test with dynamic=True to make sure the checks before `.item()` calls have reasonable behavior

Lucaskabela (changes requested):
See comments - few minor changes suggested but this looks good otherwise

torch/_dynamo/variables/lists.py
laithsakka: i can remove that. I mean calling .item will anyway do the assertion for us

torch/_dynamo/variables/lists.py
laithsakka: do you mean what happens before this PR? we used to throw data dependent error because we end up calling some c++ functions that are not data dependent friendly that handles python slicing. 
however we recently made slicing data dependent free when inputs are unbacked dynamic shapes
hence if we have an input as tensor x we can do x.item() and pass that instead.

laithsakka:
> Generally this looks good, to me but needs some cleanup of asserts and I would like to see a test with dynamic=True to make sure the checks before `.item()` calls have reasonable behavior

ah i see i will just remove those checks as you suggested probably size will be none with the input tensor is dynamic.

torch/_dynamo/variables/lists.py
laithsakka: when the input tensor is dynamic size is none i will update i believe i will update the unit tests.

Lucaskabela (approved):
Thanks for addressing feedback and fixing the test failures - the dtensor failure seems unrelated to me, so approving

PR #165083, opened 2025-10-09
[ONNX] Remove overload matching logic from dispatcher

justinchuby:
Complex overload type matching logic has been removed to keep this simple. There should no longer be overloads (for the same opset version) in torchlib anymore.

## Breaking change

The custom_translation_table now takes only one function for each torch op.

cc @titaiwangms

### DIFF
--- a/torch/onnx/_internal/exporter/_dispatching.py
+++ b/torch/onnx/_internal/exporter/_dispatching.py
@@ -2,321 +2,18 @@
 from __future__ import annotations
 
 import logging
-from collections.abc import Callable, Sequence
-from typing import Any
+from typing import TYPE_CHECKING
 
 import torch
 import torch.fx
-from torch.onnx._internal._lazy_import import onnx_ir as ir
-from torch.onnx._internal.exporter import _registration, _schemas
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+
+    from torch.onnx._internal.exporter import _registration
 
 
 logger = logging.getLogger(__name__)
-
-# Define utilities to convert PyTorch data types so users do not need to specify manually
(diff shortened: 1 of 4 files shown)


titaiwangms (commented):
Suggest we put this to draft until onnxscript side is ready?

xadupre (approved):

torch/onnx/_internal/exporter/_dispatching.py
```diff
@@ -360,11 +53,6 @@ def dispatch(
...
-        return (
-            decomp_metas[0].onnx_function,
-            "Fast path: Only one decomposition is defined",
-        )
-
-    overload, message = get_matching_overload(node, decomp_metas)
-    return overload, message
+    # NOTE: Complex overload type matching logic has been removed to keep this simple
+    # There should no longer be overloads (for the same opset version) in torchlib anymore
+    return (decomp_metas[0].onnx_function, "The first implementation is used")
```
titaiwangms, line 58: I see complex is a different and independent decom_meta: https://github.com/microsoft/onnxscript/blob/68e0458e1416ae11f5cd44eb27d07752be5e4960/onnxscript/function_libs/torch_lib/ops/core.py#L116

I am confused by the first line of NOTE means, and the message "The first implementation is used". Given the fact that we now have no overloads, should not this always be the firtst implementation?

titaiwangms (approved):
unblocking question

torch/onnx/_internal/exporter/_dispatching.py
justinchuby: Yes. Although technically there can still be multiple overloads registered, we just don't care about them. For now I think this is ok. For further clean ups we can make this even more simple.

Complex = complicated here (not complex number). I can edit after this merge.

PR #164888, opened 2025-10-07
remove more

laithsakka:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #164889
* __->__ #164888
* #164887
* #164886
* #164885
* #164884

PR #164887, opened 2025-10-07
use check_size instead of check_is_size in ops.py

laithsakka:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #164889
* #164888
* __->__ #164887
* #164886
* #164885
* #164884

cc @ezyang @EikanWang @jgong5 @wenzhe-nrv @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela

PR #164885, opened 2025-10-07
remove size-like based size-oblivious special simplifications

laithsakka:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #164889
* #164888
* #164887
* #164886
* __->__ #164885
* #164884

cc @ezyang @EikanWang @jgong5 @wenzhe-nrv @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela

Issue #137515, opened 2024-10-08
When comparing tensor sizes with dynamic shapes, avoid adding guards if len(size) is not the same. 


laithsakka:
```
@torch.compile(fullgraph=True, dynamic=True, backend="inductor")
def f(x):
   b = x[0]
   if(x.size()==b.size()):
      return x
   return x*2
```

len(x.size()) is 2 
and len(b.size()) is 1
sounds like we do not need to add the guards bellow in that case, because it always false

```
 [0/1] [__guards] +- LAMBDA_GUARD: Ne(L['x'].size()[0], L['x'].size()[1])  # if a != b:  # _dynamo/polyfills/__init__.py:61 in list_cmp (_dynamo/variables/tensor.py:1124 in evaluate_expr)
 [0/0] [__guards] +- LAMBDA_GUARD: L['x'].size()[1] == L['x'].size()[0]  # if a != b:  # _dynamo/polyfills/__init__.py:61 in list_cmp (_dynamo/variables/tensor.py:1124 in evaluate_expr)
```

laithsakka:
ok we basically calling sym_eq does what i want here for size()
```
def sym_eq(x: _T, y: _T) -> Union[bool, SymBool]:
    """
    Like ==, but when run on list/tuple, it will recursively test equality
    and use sym_and to join the results together, without guarding.
    """
    if (isinstance(x, tuple) and isinstance(y, tuple)) or (
        isinstance(x, list) and isinstance(y, list)
    ):
        if len(x) != len(y):
            return False
        return functools.reduce(operator.and_, map(sym_eq, x, y), True)
    elif isinstance(x, (int, torch.SymInt)) and isinstance(y, (int, torch.SymInt)):
        return x == y
    else:
        raise AssertionError(f"unexpected sym_eq between {type(x)} {type(y)}")
```

ezyang:
It's a little hard to believe that this is happening, because you have to work pretty hard to guard on the inner element equalities when the lists are not equal. Like if I wrote `xs == ys` in traced code that would test length first...

exclamaforte:
Doing some issue scrubbing: the guards look different now
```python
import torch

@torch.compile(fullgraph=True, dynamic=True, backend="inductor")
def f(x):
   b = x[0]
   if(x.size()==b.size()):
      return x
   return x*2
# Example usage of the function f

# Create a tensor
x = torch.tensor([[1, 2, 3], [1, 2, 3]])

# Call the function f with the tensor x
[...]

PR #165091, opened 2025-10-09
do not overguard when comparing lists

laithsakka:
* #165108
* __->__ #165091
* #164889
* #164888
* #164887
* #164886
* #164885
* #164884

if we are comparing two lists l1, l2 of different lengths for equality. 
we should early exist if len(l1) != len(l2)
and avoid guarding/comparing inner elements.

This avoids recompilations as in the unit test.  
address https://github.com/pytorch/pytorch/issues/137515

### DIFF
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -3194,7 +3194,7 @@ class TestGuardsExpressions(TestCase):
         self.assertIn("float(", guards)
         self.assertTrue(shape_env.evaluate_guards_expression(guards, [hint_int(s0)]))
         self.assertFalse(shape_env.evaluate_guards_expression(guards, [hint_int(s1)]))
-        
+
     @skipIfTorchDynamo("Attempt to trace generator")
     @torch.fx.experimental._config.patch("use_duck_shape", False)
     def test_size_comparison_no_recompile(self):
--- a/test/test_dynamic_shapes.py
+++ b/test/test_dynamic_shapes.py
@@ -3194,7 +3194,8 @@ class TestGuardsExpressions(TestCase):
         self.assertIn("float(", guards)
         self.assertTrue(shape_env.evaluate_guards_expression(guards, [hint_int(s0)]))
         self.assertFalse(shape_env.evaluate_guards_expression(guards, [hint_int(s1)]))
-
+        
+    @skipIfTorchDynamo("Attempt to trace generator")
     @torch.fx.experimental._config.patch("use_duck_shape", False)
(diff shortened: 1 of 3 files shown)


test/test_dynamic_shapes.py
```diff
@@ -3195,6 +3195,36 @@ def test_guards_float_div(self):
...
+            return x * 2
+
+        # First call: shapes differ (1, 2) vs (2, 4, 9), so if branch is False
+        f(torch.rand(10, 2), torch.rand(20, 4, 9))
+
+        # Second call: shapes differ again (1, 2) vs (1, 4, 9), so if branch is False
+        f(torch.rand(10, 2), torch.rand(10, 4, 9))
+
+        # Should only compile once despite different input shapes
+        self.assertEqual(
```
aorenste, line 3222: Should there be another one after this that checks that the same size *does* recompile?

aorenste (approved):

torch/_dynamo/polyfills/__init__.py
```diff
@@ -106,13 +106,25 @@ def accumulate_grad(x, new_grad):
...
+
+    # Optimization: For equality, short-circuit if lengths differ
+    # This avoids iterating through elements and triggering guards on SymInts
+    left_len = len(left)
+    right_len = len(right)
+
+    if op is operator.eq and left_len != right_len:
+        return False
+    if op is operator.ne and left_len != right_len:
+        return True
```
aorenste: nit: This might be a little more readable as:

```suggestion
    if left_len != right_len:
        return op(left_len, right_len)
```

laithsakka: ,,,, i dont get this change?

calling return op(left_len, right_len)?

laithsakka: ahhh now i get it

laithsakka: wer probably should not do it for < and > no?

laithsakka: ..etc

mlazos (approved):

torch/_dynamo/polyfills/__init__.py
aorenste: I thought of that but the problem is that < and > are not necessarily short-circuited by list size - you have to look at the elements:
```
>>> [1, 2, 3] < [3]
True
>>> [1, 2, 3] < [0]
False
```

laithsakka: right so i will keep it as is then

PR #150815, opened 2025-04-08
[C10D] Document object collectives limitations

wconstab:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #150880
* __->__ #150815

Adds louder warning labels in the doc page and docstring for object
collectives in hopes of raising awareness of several footgun issues
including accidental creation of cuda contexts by serializing and
sending 'device-local' gpu tensors over the object-* apis.

Preview:
image

addresses #150798

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @d4l3k

PR #165092, opened 2025-10-09
[6/N][DTensor device order] User API for device ordered distribution

zpcore:
Using `**placements**` and `**shard_order**` in DTensor Distribution
-----------------

This PR updates the documentation and implementation for **placements** and **shard_order** parameters in `distribute_tensor()` and `redistribute()` methods.

### Overview

DTensor supports two complementary ways to specify tensor distribution:

*   **placements**: Describes _what_ placement each mesh dimension has (PyTorch-style, mesh-centric)
*   **shard_order**: Describes _the order_ in which shardings are applied (JAX-style, tensor-centric)

#### User-Facing API

*   **placements**: `Sequence[Placement]` - e.g., `[Shard(0), Shard(1), Replicate()]`
*   **shard_order**: `dict[int, Sequence[int | str]]` - e.g., `{0: [1], 1: [0, 2]}` or using mesh dim string name `{0: [1], 1: ['dp', 'tp']}`
    *   Keys: tensor dimensions
    *   Values: mesh dimensions (or names) in execution order

#### Internal Representation

Internally, **shard_order** is converted to `ShardOrder` (tuple of `ShardOrderEntry` objects):
```python
shard_order:  ShardOrder = (
    ShardOrderEntry(tensor_dim=0, mesh_dims=(1,)), 
    ShardOrderEntry(tensor_dim=1, mesh_dims=(0, 2))
)
```

#### 1. Using Only **placements**

When only **placements** is specified, default left-to-right ordering is used:

```python
# Shard tensor dim 0 on mesh dim 0, tensor dim 1 on mesh dim 1 and 2 (mesh dim order is fixed)
dt = distribute_tensor( 
    tensor,
    device_mesh,
    placements=[Shard(0), Shard(1), Shard(1)]
) # Internally: shard_order = (ShardOrderEntry(0, (0,)), ShardOrderEntry(1, (1, 2)))
```

#### 2. Using Only **shard_order** (preferred way)

When only **shard_order** is specified, **placements** are inferred:

```python
# Shard tensor dim 0 on mesh dim 1, tensor dim 1 on mesh dims 2 then 0 (mesh dim order can be specified)
dt = distribute_tensor(
     tensor,
     device_mesh,
     shard_order={0: [1], 1: [2, 0]}   # Inferred placements: [Shard(1), Shard(0), Shard(1), Replicate()]
)
```

#### 3. Using Both (Maximum Control)

Specify both when you need:

*   Explicit control over multi-mesh-dimension sharding order

```python
# Tensor dim 1 sharded across mesh dims 2 and 0 (in that order) 
dt = distribute_tensor(
    tensor,
    device_mesh,
    placements=[Shard(1), Shard(0), Shard(1), Replicate()],
    shard_order={0: [1], 1: [2, 0]} )
```

(in theory **shard_order** should be expressive enough without **placements**. We leave the option there considering special placement types (e.g., uneven sharding) may be introduced in the future)

**Important**: When both are specified, they must be consistent. The function validates that placements and shard_order describe the same distribution. Note that ``shard_order`` can be a **sparse specification** - it does not need to mention every tensor dimension that is sharded. For tensor dimensions not mentioned in ``shard_order``, the default left-to-right mesh dimension order is used. When a tensor dimension IS mentioned in ``shard_order``, it must include ALL mesh dimensions that shard that tensor dimension in ``placements``. You cannot specify only a subset of the mesh dimensions for a given tensor dimension.

#### 4. Neither Specified (Default Replication)

```python
dt = distribute_tensor(tensor, device_mesh) # Results in: placements=[Replicate()] * mesh.ndim
```

### Use in `redistribute()`

Same API applies to `redistribute()`:

```python
# Redistribute with new sharding order
dt_redistributed = dt.redistribute(
    device_mesh,
    placements=[Shard(1), Shard(0), Shard(1), Replicate()],
    shard_order={1: [0, 2]}
) 
# Or just change shard order
dt_redistributed = dt.redistribute(shard_order={0: [1], 1: [2, 0]} )
```

### Constraints

1.  **Consistency**: When both are specified, placements and shard_order must describe the same distribution
2.  **No _StridedShard with shard_order**: `_StridedShard` placement cannot be used with explicit **shard_order**
3.  **No Partial inference**: User doesn't allow to specify `Partial` placement, and `Partial` placements cannot be inferred from **shard_order**
4.  **Dimension validation**: All tensor and mesh dimensions must be valid for the given tensor rank and mesh shape

--------
* #165197
* __->__ #165092


torch/distributed/tensor/_api.py
```diff
@@ -502,6 +737,20 @@ def redistribute(
                 describes how to place the DTensor into the DeviceMesh, must
                 have the same number of elements as ``device_mesh.ndim``.
                 default: replicate on all mesh dimensions
+            shard_order (dict[int, Sequence[int | str]], optional):
+                Specifies the mapping of tensor dimensions to the order of device mesh
+                dimensions they are sharded over. It is a dictionary where keys are tensor
+                dimensions and values are sequences of mesh dimensions (or mesh dimension names)
+                that the tensor dimension is sharded across, in execution order.
+
+                Internally, this is converted to a ShardOrder (tuple of ShardOrderEntry objects).
```
wanchaol: Hmm we probably don't want to mention the internal impl details in the public facing API

torch/distributed/tensor/_api.py
```diff
@@ -762,52 +1080,70 @@ def distribute_tensor(
...
-            )
-        else:
-            raise RuntimeError(
-                f"Trying to distribute tensor with unsupported placements {placement} on device mesh dimension {idx}!"
+    use_strided_shard = placements is not None and any(
+        isinstance(p, _StridedShard) for p in placements
+    )
+
+    if use_strided_shard:
+        # keep original code for backward compatibility considering
```
wanchaol, line 1168: is it being used anywhere? I would prefer to just delete `_StridedShard` instead of leaving the dead code here

torch/distributed/tensor/_api.py
```diff
@@ -612,6 +874,36 @@ def _raise_if_contains_partial_placements(self) -> None:
                 "DTensor with partial placements!"
             )
 
+    @property
+    def shard_order(self) -> ShardOrder:
```
wanchaol: I think this property should return the exact Dict[int, Sequence[str, int]] format instead of exposing ShardOrder: 1. It's an internal representation so we might want to keep it private to maximize the control without BC concern 2. the returned `shard_order` property should be able to directly feeding to the public facing API again, just like the `device_mesh` and `placements` property

torch/distributed/tensor/_api.py
zpcore: It is used in torchtitan. Once we roll out device ordered sharding, we can remove `_StridedShard`.

torch/distributed/tensor/_api.py
zpcore: We need to rename the dataclass attribute `shard_order` to a different name like `_shard_order` if we want the property to return the dict format. Let me try fix that.

torch/distributed/tensor/_api.py
zpcore: SGTM

torch/distributed/tensor/_redistribute.py
```diff
@@ -926,6 +938,10 @@ def backward(ctx, grad_output: "dtensor.DTensor"):  # type: ignore[override]
         if output.dtype != ctx.original_dtype:
             output = output.to(ctx.original_dtype)
 
+        # TODO(zpcore): During backward, some Partial related transform ops got
+        # silently skipped. This will be an issue for the graph-based
+        # redistribute planner. Need fix.
```
ezyang, line 943: Are you able to add expectedFailure tests for these?

torch/distributed/tensor/_redistribute.py
```diff
@@ -871,9 +872,20 @@ def forward(  # type: ignore[override]
         ctx.current_spec = current_spec
 
         if current_spec.placements != placements:
-            target_spec = DTensorSpec(
-                device_mesh, placements, tensor_meta=current_spec.tensor_meta
-            )
+            if shard_order is not None:
```
ezyang: This seems wrong? The placements can be the same but the shard order different, and you would need to do some handling for that no?

torch/distributed/tensor/_api.py
```diff
@@ -338,6 +562,7 @@ def __torch_dispatch__(cls, func, types, args=(), kwargs=None):  # type: ignore[
             kwargs or {},
         )
 
+    # TODO(zpcore): support `shard_order` argument
```
ezyang, line 622: This one's pretty important!

torch/distributed/tensor/_api.py
```diff
@@ -693,11 +984,28 @@ def distribute_tensor(
...
+
+            Example: For a 3D tensor and a 4D device mesh, to shard tensor dimension
+            0 over mesh dim 1, and tensor dimension 1 over mesh dim 2 and then mesh dim 0,
+            use ``shard_order = {0: [1], 1: [2, 0]}``. In this case, you can also specify the
+            ``placements`` as [Shard(1), Shard(0), Shard(1), Replicate()] along with ``shard_order``.
+
+            Note: As long as there are no Partial placements (though this should
+            not happen as of today), you may specify only shard_order (and not
+            placements if device order is not the default left-to-right), and the
+            correct placements will be inferred.
```
ezyang: I probably wouldn't have this note here at all.

torch/distributed/tensor/_api.py
```diff
@@ -721,6 +1029,19 @@ def distribute_tensor(
...
     device_mesh = device_mesh or _mesh_resources.get_current_mesh()
     device_type = device_mesh.device_type
+
+    if placements is not None:
+        for placement in placements:
+            if placement.is_partial():
+                raise RuntimeError(
+                    f"Can not distribute to {placements}, "
+                    "redistributing to Partial is for internal use only!"
+                )
```
ezyang, line 1107: Do this in a separate PR 🤔

torch/distributed/tensor/_api.py
```diff
@@ -729,7 +1050,8 @@ def distribute_tensor(
                 xla_distribute_tensor,
             )
 
-            return xla_distribute_tensor(tensor, device_mesh, placements)  # type:ignore[return-value]
+            # TODO: use `shard_order_tuple` instead of `placements_tuple` for XLA
+            return xla_distribute_tensor(tensor, device_mesh, placements_tuple)  # type:ignore[return-value]
```
ezyang, line 1128: Need to assert that shard order is standard, otherwise the XLA path will just be silently wrong here

torch/distributed/tensor/_api.py
ezyang: Another way to do it that isn't BC breaking is to transparently translate StridedShard into its shard order equivalent. This could be done separately but it would let you seamlessly port old use of StridedSharded to the new code.

torch/distributed/tensor/_api.py
```diff
@@ -762,52 +1075,70 @@ def distribute_tensor(
...
+            mesh_dims = entry.mesh_dims
+            for mesh_dim in mesh_dims:
+                assert isinstance(mesh_dim, int)
+                replicate_on_mesh_dims.remove(mesh_dim)
+                local_tensor = Shard._make_shard_tensor(
+                    tensor_dim, local_tensor, device_mesh, mesh_dim, src_data_rank
+                )
+        for mesh_dim in replicate_on_mesh_dims:
+            local_tensor = Replicate._make_replicate_tensor(
+                local_tensor, device_mesh, mesh_dim, src_data_rank
```
ezyang: The new logic here doesn't error when there is an unsupported placement which makes this highly suspicious from an error checking perspective.

torch/distributed/tensor/_api.py
```diff
@@ -762,52 +1075,70 @@ def distribute_tensor(
...
+                )
+    else:
+        replicate_on_mesh_dims = set(range(device_mesh.ndim))
+        for entry in shard_order_tuple:
+            tensor_dim = entry.tensor_dim
+            mesh_dims = entry.mesh_dims
+            for mesh_dim in mesh_dims:
+                assert isinstance(mesh_dim, int)
+                replicate_on_mesh_dims.remove(mesh_dim)
+                local_tensor = Shard._make_shard_tensor(
```
ezyang, line 1207: It's interesting that this is different from `_shard_tensor`, I guess this is the static methodification

torch/distributed/tensor/_api.py
```diff
@@ -211,6 +220,221 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+                        f"Conflict sharding annotation for Shard {original_mesh_dim.dim} "
+                        "detected between `placement` and `shard_order`."
+                    )
+                if isinstance(converted_mesh_dim, Shard):
+                    assert isinstance(original_mesh_dim, Shard), (
+                        f"Conflict sharding annotation detected for Shard {converted_mesh_dim.dim} "
+                        "between `placement` and `shard_order`."
+                    )
+            shard_order_tuple = _from_dict_to_ShardOrder(normalized_shard_order)
+    return placement_tuple, shard_order_tuple
```
ezyang, line 492: Need to review this. One thing to check is if not everything can be mentioned in shard order but it still works (not an all or nothing deal).

torch/distributed/tensor/_api.py
```diff
@@ -211,6 +220,221 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+                        f"mesh dim {mesh_dim} specified in `shard_order` is out of range "
+                        f"for placements of length {device_mesh.ndim}"
+                    )
+                normalized_shard_order[tensor_dim].append(mesh_dim)
+
+    # set default placements to replicated if not specified
+    placement_tuple: tuple[Placement, ...]
+    if placements is None:
+        if shard_order is None:
+            placement_tuple = tuple(Replicate() for _ in range(device_mesh.ndim))
```
ezyang: This is duplicated with L337

torch/distributed/tensor/_redistribute.py
zpcore: Good catch!!!

torch/distributed/tensor/_redistribute.py
zpcore: I managed to create a test with mocked redistribution path in order to trigger the case in the backward path. Will update the PR.

torch/distributed/tensor/_api.py
zpcore: Yes, Wanchao suggested that we should change `_shard_tensor` etc. to the static method call eventually.

torch/distributed/tensor/_api.py
zpcore: Considering the necessary tests for `from_local` and `to_local` when we add `shard_order` arg, I shall address them in a different PR.

zpcore: OK, I also need to add the `shard_order` support to those DTensor factory ops like `ones` in the file :<

torch/distributed/tensor/_api.py
zpcore: Cleaned up!

torch/distributed/tensor/_api.py
```diff
@@ -211,6 +220,221 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+        else:
+            # convert from shard_order to placements
+            placement_tuple = _convert_shard_order_to_placements(
+                cast(dict[int, Sequence[int]], normalized_shard_order), device_mesh
```
ezyang: you can avoid the variance disaster here by having the function accept Mapping (which is read only, and then you are covariant in the value position)

torch/distributed/tensor/_api.py
```diff
@@ -211,6 +220,221 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+            if placement.is_shard():
+                tensor_dim = placement.dim  # type: ignore[attr-defined]
+                if tensor_dim < -tensor_rank or tensor_dim >= tensor_rank:
+                    raise ValueError(
+                        f"tensor dim {tensor_dim} specified in `placements` is out of range for tensor_rank {tensor_rank}."
+                    )
+                tensor_dim = tensor_dim + tensor_rank if tensor_dim < 0 else tensor_dim
+
+                # reconstruct `placement` object in case it is `_StridedShard` for backward compatibility
+                if isinstance(placement, _StridedShard):
```
ezyang, line 442: This is not written very defensively. Shouldn't you assert that the type of placement is exactly Shard in the fallthrough case?

torch/distributed/tensor/_api.py
```diff
@@ -211,6 +220,221 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+    else:
+        normalized_placements = list(placements)  # type: ignore[assignment]
+        for i, placement in enumerate(placements):
+            if placement.is_shard():
+                tensor_dim = placement.dim  # type: ignore[attr-defined]
+                if tensor_dim < -tensor_rank or tensor_dim >= tensor_rank:
+                    raise ValueError(
+                        f"tensor dim {tensor_dim} specified in `placements` is out of range for tensor_rank {tensor_rank}."
+                    )
+                tensor_dim = tensor_dim + tensor_rank if tensor_dim < 0 else tensor_dim
```
ezyang: It is pretty annoying that you have this code over and over again

torch/distributed/tensor/_api.py
```diff
@@ -211,6 +220,221 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+                if isinstance(original_mesh_dim, Shard):
+                    assert (
+                        isinstance(converted_mesh_dim, Shard)
+                        and original_mesh_dim.dim == converted_mesh_dim.dim
+                    ), (
+                        f"Conflict sharding annotation for Shard {original_mesh_dim.dim} "
+                        "detected between `placement` and `shard_order`."
+                    )
+                if isinstance(converted_mesh_dim, Shard):
+                    assert isinstance(original_mesh_dim, Shard), (
```
ezyang: This is a very strange way to write these conditionals... why are you not just testing equality here

torch/distributed/tensor/_api.py
```diff
@@ -211,6 +220,221 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
         return grad_output.to_local(), None, None, None, None, None
 
 
+def _prepare_placements_and_shard_order(
```
ezyang, line 260: I would be happier if the canonicalization code was written more "tightly", i.e., didn't assume that the universe was only Shard and StridedShard, but that there could be potentially more sharding types. Overall the logic is very verbose which also makes it difficult to review.

torch/distributed/tensor/_api.py
zpcore: Updated to allow this specification. Will refresh the PR.

torch/distributed/tensor/_api.py
ezyang: There is actually a decently large can of worms for how exactly to best expose shard order here. I also consulted JAX's API but they have some weird setup where the entire placement is itself a subclass'able class. To unblock, one strategy is to expose a method for a *small* slice of information: e.g., an API that lets you specifically query the order of mesh dims for a single tensor dim. You can then have separate APIs for string mesh names as well as mesh dim indices. Another strategy is to return an object and then we can slowly fill in methods on that object for getting information (internally, it can be continue to be canonical form). The object is also good because it means we can easily create another tensor with exactly the same placement as the original, including shard order.

torch/distributed/tensor/_api.py
zpcore: OK, I added the raise condition similar to `use_strided_shard` case above to be safe.

torch/distributed/tensor/_api.py
zpcore: One is for `placements` and another for `shard_order`.  Not sure if there is an elegant way to handle this...

torch/distributed/tensor/_api.py
zpcore: This part is to normalize for `placements`.  So the `_StridedShard` still need to be adjusted as long as it doesn't come with `shard_order`.

ezyang: I'm talking about the else branch

zpcore: I see, updated

torch/distributed/tensor/_api.py
zpcore: This part is updated.

torch/distributed/tensor/_api.py
zpcore: I checked how jax show the tensor sharding. Here is how they print out tensor.sharding: https://github.com/jax-ml/jax/blob/48c876227a67840d7bcd44d80d16edbc0e910335/jax/_src/named_sharding.py#L139-L144.
```
>> print(tensor.sharding)
SingleDeviceSharding(device=CpuDevice(id=0), memory_kind=device)  # single device
```
or for sharded tensor:
```
NamedSharding(mesh=Mesh('a': 2, 'b': 2, axis_types=(Auto, Auto)), spec=PartitionSpec(('b', 'a'), None), memory_kind=unpinned_host)  # multiple device
```
The `axis_types` is to control sharding type (e.g., Explicit, Manual), which we don't support.  I think for now we can just print out the `mesh` and `spec` information, and make printed `spec` in a format that user can copy to use directly.

torch/distributed/tensor/_api.py
zpcore: Your hunch is correct, fixed in https://github.com/pytorch/pytorch/pull/165533.

torch/distributed/tensor/_api.py
wanchaol: > We need to rename the dataclass attribute shard_order to a different name like _shard_order if we want the property to return the dict format. Let me try fix that.

Hmmm curious why we need to do the renaming? iiuc `_spec` is a private attribute already, in `_spec` it can still be the `shard_order` attribute that is the ShardOrder object for internal representation. Here we are adding a public attribute to the DTensor object directly, in the DTensor object level we can just translate the `ShardOrder` to the original dict input format, i.e. maybe just add a method to the `ShardOrder` object like `to_dict` to translate it? We can also make this not a property and allow user to pass in `tensor_dim` as an argument to query the mesh dim order as @ezyang suggested.

zpcore: Sorry I mess up with the shard_order property from DTensor and DTensorSpec previously. I have updated the property in DTensor to return the dict, which should be able to reused directly to create another DTensor (as in the test).

torch/distributed/tensor/_api.py
```diff
@@ -212,6 +221,244 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+    tuple[Placement, ...],
+    ShardOrder,
+]:
+    """
+    This function places `placements` and `shard_order` in a redundant but
+    canonical form, whereas in the input users can elide redundant information
+    when specifying `placements`. For example, when there is never a tensor
+    dimension that is sharded by multiple mesh dims, `shard_order` can be elided
+    (this is traditional "PyTorch" style). Similarly, if a user specifies a
+    `shard_order` that has entry for every device mesh dim, the `placements` can
```
wconstab: "has entry" -> "has an entry"

torch/distributed/tensor/_api.py
```diff
@@ -212,6 +221,244 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+    be inferred (this is traditional "JAX" style).
+
+    You can also specify both arguments, which may be necessary in some
+    situations as `placements` and `shard_order` have different expressivity.
+    For example, to express that a device mesh is `Partial` or a non-standard
```
wconstab: to express that a Placement is something other than `Shard` (e.g.`Partial`)...

torch/distributed/tensor/_api.py
```diff
@@ -212,6 +221,244 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+    sharding differently from Shard, you must use `placements` (as only it takes
```
wconstab: ~you must use `placements` (as only it takes Placement)~
this can only be specified via the 'placements' kwarg.

torch/distributed/tensor/_api.py
```diff
@@ -212,6 +221,244 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+    Placement). To express the order multiple shardings apply to a single tensor
```
wconstab: To express the order of multiple shardings applied to a single tensor dimension, you must use `shard_order`.

torch/distributed/tensor/_api.py
```diff
@@ -212,6 +221,244 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+    dimension, you must use `shard_order`. If you want to express both of these
+    things, you will need to use both arguments.
+
+    `placements` and `shard_order` must be consistent with each other. When both
+    are set:
+
+    1. For each entry in `shard_order` (tensor_dim: [mesh_dims...]), each mesh_dim
+       must correspond to a `Shard(tensor_dim)` placement at that position in
```
wconstab: is this strictly true?  (i.e. if someone makes a new placement type besides Shard, the order could apply to that too right?

What happened with @sanketpurandare's work on DynamicShard for MoE- that's a good example, i'm not sure if it is a subclass of Shard or a separate class.

torch/distributed/tensor/_api.py
```diff
@@ -212,6 +221,244 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+      - Invalid: `shard_order={0: [1]}` - ERROR! Tensor dim 0 is sharded on mesh
+        dims [0, 1] but only [1] is specified
+      - Valid: `shard_order={0: [1, 0], 1: [3, 2]}` - fully customizes both dims
+
+    3. A mesh dim mentioned in `shard_order` must be a Shard placement; no other
+       placement types are supported.
+
+    In the returned canonical representation, no information is omitted. In
+    particular, `shard_order` is no longer a dict, it is a sparse tuple with
+    each inner tuple corresponds to a tensor dimension as the first element and
```
wconstab: corresponding to a sharded tensor dimension

torch/distributed/tensor/_api.py
```diff
@@ -212,6 +221,244 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+        RuntimeError: If attempting to redistribute from a non-Partial to a
+            Partial placement, from one Partial type to a different Partial type,
+            or use _StridedShard with `shard_order`.
+        AssertionError: If a placement's shard dim normalization would result in
+            a negative value, or if there is a conflict between placements and
+            shard_order for sharding annotation.
+
+    """
+
+    def _from_dict_to_ShardOrder(
```
wconstab, line 372: just curious, how come we prefer a dict as the user interface and a tuple as the final type?  i suppose having a tuple is a good choice for immutability, and perhaps a dict is more flexible for a user (?)

torch/distributed/tensor/_api.py
wconstab: well at least make it a local helper, it can capture 'tensor_rank' and take unnormalized_tensor_dim as arg, throw the ValueError, and return the normalized dim

torch/distributed/tensor/_api.py
```diff
@@ -212,6 +221,244 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+                    )
+                tensor_dim = tensor_dim + tensor_rank if tensor_dim < 0 else tensor_dim
+
+                # reconstruct `placement` object in case it is `_StridedShard` for backward compatibility
+                if isinstance(placement, _StridedShard):
+                    has_stridedshard_in_placements = True
+                    normalized_placements[i] = _StridedShard(
+                        tensor_dim, split_factor=placement.split_factor
+                    )
+                else:
```
wconstab, line 447: i have a couple thoughts here

1- as mentioned above, we might have other types of shard placements (e.g. DynamicShard), and it'd be good to not have to change this code.
2- it's kind of annoying to normalize the tensor dim so late in the game.  I wonder if we can do it earlier.  The Shard constructor can't actually do it becuase it doesn't know anything about the tensor, right?  We could add a 'normalize(tensor_rank)' helper to the base 'Placement' class, and call that on the user-passed placements, which would avoid the normalization code being here and also ensure it works to normalize any kind of placement.

torch/distributed/tensor/_api.py
```diff
@@ -212,6 +221,244 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+
+        placement_tuple = tuple(normalized_placements)
+
+        if shard_order is None:
+            shard_order_tuple = DTensorSpec.compute_default_shard_order(placement_tuple)
+        else:
+            # both shard_order and placements are specified; need to validate their correctness
+            if has_stridedshard_in_placements:
+                # _StridedShard doesn't work with shard_order
+                raise RuntimeError(
```
wconstab: ValueError?

torch/distributed/tensor/_api.py
```diff
@@ -212,6 +221,244 @@ def backward(ctx, grad_output: "DTensor"):  # type: ignore[override]
...
+            # Build map from tensor_dim -> list of mesh_dims (in left-to-right order)
+            tensor_dim_to_mesh_dims: dict[int, list[int]] = defaultdict(list)
+            for mesh_dim, placement in enumerate(placement_tuple):
+                if placement.is_shard():
+                    tensor_dim = placement.dim  # type: ignore[attr-defined]
+                    tensor_dim_to_mesh_dims[tensor_dim].append(mesh_dim)
+
+            # Validate: if a tensor_dim is in shard_order, all its sharding mesh dims must be listed
+            for tensor_dim, mesh_dims_in_order in normalized_shard_order.items():
+                expected = set(tensor_dim_to_mesh_dims.get(tensor_dim, []))
```
wconstab, line 475: should we use set here? wouldn't it be better to check that the specified order matches between the two?

wconstab (commented):
only had time to review `_prepare_placements_and_shard_order` function so far, will circle back to the rest

torch/distributed/tensor/_api.py
```diff
@@ -613,6 +908,51 @@ def _raise_if_contains_partial_placements(self) -> None:
...
+        .. note:: ``shard_order`` is a read-only property derived from the DTensor's
+            internal :class:`ShardOrder` specification (a tuple of :class:`ShardOrderEntry`
+            objects). The returned dictionary provides a more user-friendly interface
+            to the underlying shard order information.
+
+        .. note:: If you need to redistribute a DTensor with a different shard order,
+            use the :meth:`redistribute` method with the ``shard_order`` parameter.
+        """
+        tensor_mesh_dim_dict = defaultdict(list)
+        for entry in self._spec.shard_order:
```
wconstab, line 962: i wish we didn't have to construct a new dict every time someone called an accessor that they probably expect to be cheap.  We might find another way to do this, or, at least use 'cache'

torch/distributed/tensor/_api.py
```diff
@@ -694,11 +1035,35 @@ def distribute_tensor(
...
             by default replicate the tensor across the ``device_mesh`` from the
-            first rank of each dimension of the `device_mesh`.
+            first rank of each dimension of the ``device_mesh``.
+        shard_order (dict[int, Sequence[int | str]], optional):
+            Specifies the mapping of tensor dimensions to the order of device mesh
+            dimensions they are sharded over. It is a dictionary where keys are tensor
+            dimensions and values are sequences of mesh dimensions (or mesh dimension names)
+            that the tensor dimension is sharded across, in execution order.
+
+            ``shard_order`` can be a **sparse specification** - it does not need to
```
wconstab, line 1059: I think you have repeated the docstring details for shard_order in many places.  I would prefer if you put one doc section (using an anchor) into the DTensor doc page, and then link to that header from the different code sections. let me see if i can find an example of doing this as it's tricky to do.

wconstab: take a look at https://github.com/pytorch/pytorch/pull/150815/files

see how i defined it in the main doc 

```
.. _object_collectives:

Object collectives
------------------
...
```

and then referenced it form the code docstrings

```
...
See :ref:`object_collectives`
...
```

torch/distributed/tensor/_api.py
zpcore: I think as long as the new type of placement do partition on tensor, device order should matter. I prefer to make those (like _StridedShard and DynamicShard) all be subclass of Shard. Can't make a decision now...

torch/distributed/tensor/_api.py
zpcore: One benefit is that, with a dict, we don't need to related it with the tensor rank. I am not sure if this is useful, but it makes the writing easier.

torch/distributed/tensor/_api.py
zpcore: Right, fixed

torch/distributed/tensor/_api.py
zpcore: `tensor_dim_to_mesh_dims` is derived from placements to map tensor dim to mesh dims (no order information). So we can't compare the order between the two.

torch/distributed/tensor/_api.py
zpcore: Yes, let me revisit this once we switch to `StridedShard` internally.

torch/distributed/tensor/_api.py
zpcore: Fixed, with the help of AI.

PR #165094, opened 2025-10-09
Fix error, remove file from pyrefly checking

maggiemoss:
Reported issue with formatting and parsing. 

Removing suppressions and avoiding this file in future type checking until we can get a more complete fix in .


albanD (approved):
Thanks!

PR #164705, opened 2025-10-05
[cuda] fix triu/tril int32 overflow for large matrices

Aminsed:
Fixes #136611

Cast blockIdx.x to int64_t before multiplication to prevent overflow when computing linear_idx for matrices larger than 2^31 elements.

cc @ngimel @ptrblck @eqy

PR #165095, opened 2025-10-09
Fix int32 overflow in embedding_dense_backward

yinghai:
If `max_partial_segment` is large we can overflow `gid` and cause a bunch of IMA.

### DIFF
--- a/aten/src/ATen/native/cuda/EmbeddingBackwardKernel.cu
+++ b/aten/src/ATen/native/cuda/EmbeddingBackwardKernel.cu
@@ -88,9 +88,9 @@ __global__ void compute_grad_weight_bags(
     const int64_t stride_warped) {
 
   int64_t num_of_segments = *num_of_segments_ptr;
-  const int gid = blockIdx.x * blockDim.x + threadIdx.x;
-  const int id = gid / stride_warped;
-  const int startFeature = gid % stride_warped;
+  const int64_t gid = blockIdx.x * blockDim.x + threadIdx.x;
+  const int64_t id = gid / stride_warped;
+  const int64_t startFeature = gid % stride_warped;
   if (startFeature >= stride) {
     return;
   }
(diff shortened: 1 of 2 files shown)


test/nn/test_embedding.py
```diff
@@ -632,6 +632,66 @@ def gen_2D_indices_from_1D(
                     weights.grad, weights_check.grad, msg=msg, atol=atol, rtol=rtol
                 )
 
+    @onlyCUDA
+    @dtypes(torch.bfloat16,)
```
ngimel: decorate is with @largeTensorTest to avoid ooms

ngimel (approved):

aten/src/ATen/native/cuda/EmbeddingBackwardKernel.cu
```diff
@@ -88,9 +88,9 @@ __global__ void compute_grad_weight_bags(
     const int64_t stride_warped) {
 
   int64_t num_of_segments = *num_of_segments_ptr;
-  const int gid = blockIdx.x * blockDim.x + threadIdx.x;
-  const int id = gid / stride_warped;
-  const int startFeature = gid % stride_warped;
+  const int64_t gid = blockIdx.x * blockDim.x + threadIdx.x;
```
eqy: One of these values needs to be cast to 64-bit, otherwise the math is still done in 32bit
see e.g., https://github.com/pytorch/pytorch/pull/142010 https://github.com/pytorch/pytorch/pull/164705

eqy (changes requested):

aten/src/ATen/native/cuda/EmbeddingBackwardKernel.cu
yinghai: thanks updated.

ngimel: ugh interesting that the test isn't catching it

eqy (approved):

PR #164181, opened 2025-09-29
[a2av] Add all_to_all_v API accepting exchange plan

kwen2501:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #168047
* #167792
* #167676
* #167144
* #166896
* #166899
* #166898
* #166774
* #163815
* #163814
* #166772
* #165482
* #165101
* __->__ #164181
* #164164

The `all_to_all_v` API accepts an `ExchangePlan`, so that the user can exchange multiple sets of data without paying the meta data exchange cost.
For example:
```
all_to_all_v(topk_idx_in, topk_idx_out, plan)
all_to_all_v(topk_weights_in, topk_weights_out, plan)
all_to_all_v(tokens_in, tokens_out, plan)
```

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci

PR #164164, opened 2025-09-29
[a2av] Add make_exchange_plan for reuse of meta data

kwen2501:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #168047
* #167792
* #167676
* #167144
* #166896
* #166899
* #166898
* #166774
* #163815
* #163814
* #166772
* #165482
* #165101
* #164181
* __->__ #164164

Adding a new API to exchange all-to-all-v splits and offsets.

```
in_splits = ...
plan = dist.symm_mem.make_a2a_exchange_plan(
    in_splits, src_offsets, out_splits, dst_offsets, group_name
)
```

The end goal is so that users can reuse the `plan` to exchange multiple data following the same splits:
```
all_to_all_v(topk_idx_in, topk_idx_out, plan)
all_to_all_v(topk_weights_in, topk_weights_out, plan)
all_to_all_v(tokens_in, tokens_out, plan)
```

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci

PR #165101, opened 2025-10-09
[a2av] 2D all-to-all-v exchange plan

kwen2501:
* #168047
* #167792
* #167676
* #167144
* #166896
* #166899
* #166898
* #166774
* #163815
* #163814
* #166772
* #165482
* __->__ #165101
* #164181
* #164164

If we want an intra-node A2A kernel to run concurrently with the inter-node A2A kernel, the intra-node kernel would need  some meta information ahead of time, for example, the offset of a token in a destination rank. 

The `make2dExchangePlan` kernel in this PR prepares such information, and would be run ahead of the intra-node A2A kernel.

Re "the offset of a token in a destination rank":
I am assuming that the inter-node A2A is in a pull style, while the intra-node A2A is in a push style, so that they two kernels can connect better (e.g. via a signal). For push style, we would need the dest offset for each token.


torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
```diff
@@ -1007,6 +1007,150 @@ void _all_to_all_get(
...
+// This kernel is used to exchange output splits and dest offsets between peers.
+__global__ void make2dExchangePlan(int64_t* in_splits, int64_t* src_offsets, int64_t* out_splits, int64_t* dst_offsets, nvshmem_team_t team, int ne) {
+#ifndef _NVSHMEM_DEVICELIB_SUPPORTED
+  CUDA_KERNEL_ASSERT_MSG(false, "SM arch unsupported for NVSHMEM");
+#else
+  int peer = threadIdx.x / WARP_SIZE;
+  int my_pe = nvshmem_team_my_pe(team);
+  int npes = nvshmem_team_n_pes(team);
+  int nsplits = ne * npes;
+  CUDA_KERNEL_ASSERT(npes <= MAX_N_PEERS && " Number of peers must be smaller than MAX_N_PEERS\n");
```
Skylion007, line 1235: Ugh typo, shouldn't it be?
```suggestion
  CUDA_KERNEL_ASSERT(npes <= MAX_N_PEERS , " Number of peers must be smaller than MAX_N_PEERS\n");

File: torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu [hunk 2]
```diff
@@ -1007,6 +1007,150 @@ void _all_to_all_get(
...
+  // Make an exchange plan for a AllToAllv_2D shuffle operation.
+  auto in_splits_hdl = c10d::symmetric_memory::rendezvous(in_splits, group_name);
+  auto src_offsets_hdl = c10d::symmetric_memory::rendezvous(src_offsets, group_name);
+  auto out_splits_hdl = c10d::symmetric_memory::rendezvous(out_splits, group_name);
+  auto dst_offsets_hdl = c10d::symmetric_memory::rendezvous(dst_offsets, group_name);
+
+  // Verify inputs
+  auto npes = in_splits_hdl->get_world_size();
+  int nsplits = in_splits.size(0);
+  TORCH_CHECK(nsplits % npes == 0, "Number of splits must be a multiple of number of peers");
```
Skylion007, line 1324:
```suggestion
  TORCH_CHECK_VALUE(nsplits % npes == 0, "Number of splits must be a multiple of number of peers");
```
etc for all other checks of invalid input args.

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
```diff
@@ -1007,6 +1007,150 @@ void _all_to_all_get(
...
+
+  auto device = in_splits.device();
+  c10::cuda::CUDAGuard guard(device);
+  auto& team_manager = TeamManager::get(device);
+  auto team = team_manager.get_team(group_name, in_splits_hdl->get_rank_to_global_rank());
+  auto stream = at::cuda::getCurrentCUDAStream(device.index());
+
+  // Exchange output splits and source offsets
+  // Use collective launch because kernel involves nvshmem barrier
+  void* args0[] = {
```
Skylion007, line 1346: Nit, can't be an std::array that passes a data pointer?

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
kwen2501: I thought the assert accepts 1 field only. Can double check.

ngimel: assert accepts 1 field only, correct

torch/csrc/distributed/c10d/symm_mem/nvshmem_extension.cu
```diff
@@ -1007,6 +1007,150 @@ void _all_to_all_get(
...
+  }
+  __syncthreads();
+
+  // Write to output splits
+  int local_expert = threadIdx.x % WARP_SIZE;
+  if (local_expert < ne) {
+    // experts I own
+    int expert_id = ne * my_pe + local_expert;
+    // out_splits is (ne, npes) flattened
+    out_splits[local_expert * npes + peer] = in_splits_all[peer][expert_id];
```
ngimel, line 1253: seem out_splits doesn't need to be symmetric?

PR #165106, opened 2025-10-09
[PP] move FSDP reduce scatters to end of step

H-Huang:
* __->__ #165106
* #164976

Move FSDP reduce scatters to the end of the PP step. The reduce scatter compute stream sync blocks the other stages from executing their backwards leading to bubbles. There should be a way to execute these RS earlier, but doing this for now as a quick fix.

image


torch/distributed/pipelining/schedules.py
```diff
@@ -1554,6 +1554,12 @@ def step(self, *args, target=None, losses: Optional[list] = None, **kwargs):
         # Run microbatches
         self._step_microbatches(args_split, kwargs_split, targets_split, losses)
 
+        # This is only applicable in the FSDP case, any outstanding events need to be waited on
+        for stage in self._stages:
+            for event in stage.reduce_scatter_events:
+                if event is not None:
+                    torch.cuda.current_stream().wait_event(event)
```
weifengpy: when are we clearing reduce_scatter_events after wait ?

weifengpy (approved):

torch/distributed/pipelining/schedules.py
H-Huang: Good point, thanks, i will update this

PR #165112, opened 2025-10-10
Use sym_eq in _check_rms_norm_inputs_symint

ColinPeppler:
Summary:
### Problem
ArrayRef's `equals()`does elementwise quality using `==` operator. This can cause a DDE for unbacked symints since `==`  operator calls `guard_bool`.
```
// SymInt.h  
bool operator==(const SymInt& o) const {
  return sym_eq(o).guard_bool(__FILE__, __LINE__);
}
```

### Solution
Adds `sym_equals()` to do elementwise equality for `SymIntArrayRef`. Use this instead of `equals()` for `SymIntArrayRef`.

Reviewed By: guangy10, pianpwk, muchulee8

Differential Revision: D84168401

### DIFF
--- a/aten/src/ATen/native/layer_norm.h
+++ b/aten/src/ATen/native/layer_norm.h
@@ -3,6 +3,9 @@
 #include <ATen/core/Tensor.h>
 #include <ATen/native/DispatchStub.h>
 #include <c10/util/accumulate.h>
+#include <c10/core/SymBool.h>
+#include <c10/util/StringUtil.h>
+
 
 namespace at::native {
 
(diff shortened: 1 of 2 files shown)


aten/src/ATen/native/layer_norm.h
```diff
@@ -19,28 +20,33 @@ C10_ALWAYS_INLINE void _check_rms_norm_inputs_symint(
...
+      "Input tensor must have at least ", normalized_ndim, " dimensions, but got ", input_ndim);
+
+  std::stringstream ss;
+  ss << "Given normalized_shape=" << normalized_shape
+     << ", expected input with shape [*";
+  for (auto size : normalized_shape) {
+    ss << ", " << size;
   }
+  ss << "], but got input of size" << input_shape;
+
```
Skylion007: Probbbably cleaner to rewrite this with fmtlib and use fmt::join here.

aten/src/ATen/native/layer_norm.h
```diff
@@ -19,28 +20,33 @@ C10_ALWAYS_INLINE void _check_rms_norm_inputs_symint(
...
-           .equals(normalized_shape)) {
-    std::stringstream ss;
-    ss << "Given normalized_shape=" << normalized_shape
-       << ", expected input with shape [*";
-    for (auto size : normalized_shape) {
-      ss << ", " << size;
-    }
-    ss << "], but got input of size" << input_shape;
-    TORCH_CHECK(false, ss.str());
+  TORCH_CHECK(
```
Skylion007:
```suggestion
  TORCH_CHECK_VALUE(

File: aten/src/ATen/native/layer_norm.h (hunk 1 above)
Comment (ColinPeppler) [reply aten/src/ATen/native/layer_norm.h]: didn't know about this, will give it a ttry

File: c10/core/SymIntArrayRef.h [hunk 3]
```diff
@@ -86,4 +86,12 @@ inline SymIntArrayRef fromIntArrayRefSlow(IntArrayRef array_ref) {
       reinterpret_cast<const SymInt*>(array_ref.data()), array_ref.size());
 }
 
+inline c10::SymBool sym_equals(SymIntArrayRef LHS, SymIntArrayRef RHS) {
+  c10::SymBool result = sym_eq(LHS.size(), RHS.size());
+  for (size_t i = 0; i < RHS.size(); ++i) {
+    result = result.sym_and(sym_eq(LHS[i], RHS[i]));
```
Skylion007: Slight nit: we could do an early return if results evaluated to false, right?

aten/src/ATen/native/layer_norm.h
ColinPeppler: okay could not get `fmtlib` to work with meta's internal linter (maybe I could've just ignored it) but I did find `c10::str` /`c10::Join` which offers a similar API, wdyt?

Skylion007: Yeah, just use that. What's I would have suggested as a backup

Skylion007 (approved):
Looks good assuming c10::join

PR #164953, opened 2025-10-08
[Inductor][ATen] Fix stride rounding on Blockwise128x128 to accommodate for small shapes

jananisriram:
Summary: Fix rounding issue on `Blockwise128x128` to accommodate for small shapes. The original implementation rounded all strides to 4, which caused failures for `test_fp8.py` tests as well as `test_scaled_matmul_cuda.py::test_scaled_mm_vs_emulated_block_wise` tests ([GitHub PR](https://github.com/pytorch/pytorch/pull/164259)).

Test Plan:
`test_fp8.py`
`test_scaled_matmul_cuda.py`

Differential Revision: D84103213

PR #165132, opened 2025-10-10
[Inductor][Triton][FP8] Support tile-wise (1x128) scaling in Inductor

jananisriram:
Summary:
Support tile-wise `1x128` scaling in Inductor Triton for FP8 GEMMs, i.e. scaling values along tensors `a` and `b` represent a `1x128` slice of input.

NOTE: Block-wise `128x128` and `1x128` scaling is only supported in CUDA 12.9+; therefore, tile-wise scaling is currently unsupported in `fbcode` (CUDA 12.4). Use OSS PyTorch to run tile-wise scaling (as with deepseek-style scaling).

Test Plan:
Works out-of-the-box with TritonBench:
```
TORCHINDUCTOR_CACHE_DIR=~/personal/cache_dir_inductor CUDA_LAUNCH_BLOCKING=1 TORCH_USE_CUDA_DSA=1 TRITON_PRINT_AUTOTUNING=1 TRITON_ALWAYS_COMPILE=1 TORCH_LOGS=+inductor TORCHINDUCTOR_FORCE_DISABLE_CACHES=1 ENABLE_PERSISTENT_TMA_MATMUL=1 TORCHINDUCTOR_MAX_AUTOTUNE_GEMM=1 buck2 run mode/{opt,inplace} pytorch/tritonbench:run -- --op fp8_gemm --only torch_fp8_gemm,pt2_fp8_gemm --metrics tflops,accuracy --m 256 --n 768 --k 512 --output="/home/jananisriram/personal/random_bench.csv" --scaling-pair=BlockWise1x128,BlockWise1x128 --atol=1e-2 --rtol=0.5
```

Differential Revision: D84025878

### DIFF
--- a/test/inductor/test_fp8.py
+++ b/test/inductor/test_fp8.py
@@ -794,14 +794,16 @@ class TestFP8Lowering(TestCase):
         _get_torch_cuda_version() < (12, 9),
         "cuBLAS blockwise scaling added in CUDA 12.9",
     )
-    @parametrize(
-        "shape", ((16, 256, 256), (1024, 512, 1024))
-    )  # TODO (jananisriram): add scaling recipe overrides for shapes like (16, 256, 64) and (256, 16, 64)
+    @parametrize("shape", ((16, 256, 256), (1024, 512, 1024)))
     @parametrize("use_fast_accum", (False, True))
-    def test_blockwise1x128_blockwise128x128_scaling(
+    @parametrize(
+        "scaling_block_sizes", ((1, 128, 128, 128), (1, 128, 1, 128))
+    )  # (BlockWise1x128, BlockWise128x128), (BlockWise1x128, BlockWise1x128)
+    def test_main_loop_scaling(
         self,
         shape: tuple[int, int, int],
         use_fast_accum: bool,
+        scaling_block_sizes: tuple[int, int, int, int],
     ):
         # Only bf16 output type is supported for non-tensorwise scaling, not fp32
         dtype: torch.dtype = torch.bfloat16
@@ -814,20 +816,28 @@ class TestFP8Lowering(TestCase):
         w = torch.randn(N, K, dtype=dtype, device=device)
         bias = None
 
+        am, ak, bn, bk = scaling_block_sizes
+
         # quantize weight (prior to inference)
         w_fp8, w_inverse_scale = _quantize_blockwise(
-            w, dtype_float8, block_outer=128, block_inner=128
+            w, dtype_float8, block_outer=bn, block_inner=bk
         )
         w_t_fp8 = w_fp8.t()
-        w_inverse_scale = w_inverse_scale.t()  # scale_b should be (1, N)
+        if (bn, bk) == (1, 128):
+            w_inverse_scale = (
+                w_inverse_scale.t().contiguous().t().t()
+            )  # 1x128 blocks need scales to be outer-dim-major
+        else:
+            w_inverse_scale = w_inverse_scale.t()  # scale_b should be (1, N)
 
         # quantize input x
         x_fp8, x_inverse_scale = _quantize_blockwise(
-            x, dtype_float8, block_outer=1, block_inner=128
+            x, dtype_float8, block_outer=am, block_inner=ak
         )
-        x_inverse_scale = (
-            x_inverse_scale.t().contiguous().t()
-        )  # 1x128 blocks need scales to be outer-dim-major
+        if (am, ak) == (1, 128):
+            x_inverse_scale = (
+                x_inverse_scale.t().contiguous().t()
+            )  # 1x128 blocks need scales to be outer-dim-major
 
         def linear(x_fp8, x_inverse_scale, w_t_fp8, w_inverse_scale, bias):
             y = torch._scaled_mm(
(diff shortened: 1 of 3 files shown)


aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1177,7 +1177,7 @@ bool is_blockwise_128x128_scaling(const at::Tensor& t, const at::Tensor& scale)
           scale,
           0,
           ceil_div<int64_t>(t.size(0), 128),
-          round_up<int64_t>(ceil_div<int64_t>(t.size(1), 128), 4)) &&
+          ceil_div<int64_t>(t.size(1), 128)) &&
```
PaulZhang12: Is this safe? Will change aten behavior

slayton58: It's actually from #164953 and tested there - it'd go away on a rebase, along with the associated tests in `test/test_scaled_matmul_cuda.py`

eqy (approved):

test/inductor/test_fp8.py
```diff
@@ -768,7 +770,94 @@ def linear(x_fp8, x_inverse_scale, w_t_fp8, w_inverse_scale, bias):
...
+        use_fast_accum: bool,
+        scaling_block_sizes: str,
+    ):
+        # Only bf16 output type is supported for non-tensorwise scaling, not fp32
+        dtype: torch.dtype = torch.bfloat16
+        device = "cuda"
+        dtype_float8 = torch.float8_e4m3fn
+        dtype_float8 = _fix_fp8_dtype_for_rocm(dtype_float8, device)
+
+        shape = [int(dim) for dim in shape.split(",")]
```
drisspg: nit: just use tuples of ints?

torch/_inductor/kernel/mm.py
```diff
@@ -558,10 +562,222 @@ def apply_scaling(
...
+        offs_k = ki * BLOCK_K
+
+        a = tl.load_tensor_descriptor(a_desc, [offs_am, offs_k])
+        b = tl.load_tensor_descriptor(b_desc, [offs_bn, offs_k])
+
+        m_blocks = tl.cdiv(M, 128)
+        n_blocks = tl.cdiv(N, 128)
+        k_blocks = tl.cdiv(K, 128)
+
+        if SCALE_RECIPE_A == 2:  # ScalingType.Blockwise1x128
```
drisspg: Nit: all these if else could be jinja if elses so that the final output code is easier to read / debug if needed

drisspg (approved):
Much cleaner, a couple small comments. Great job!

torch/_meta_registrations.py
```diff
@@ -6444,6 +6444,20 @@ def ceil_div(a, b):
                     scale_a.is_contiguous() and scale_b.is_contiguous(),
                     lambda: "Both scale_a and scale_b must be contiguous for rowwise scaling.",
                 )
+            elif (
+                scale_a.size(0) == m
+                and scale_a.size(1) == scale_b.size(0) == (_k + 128 - 1) // 128
+                and scale_b.size(1) == (n + 128 - 1) // 128
```
njriasan: Nit: This is just ceil div 128? Seems true for k as well.

torch/_inductor/kernel/mm.py
```diff
@@ -378,15 +376,16 @@ def apply_scaling(
...
+    SCALE_RECIPE_B: tl.constexpr,
     offs_cm,
     offs_cn,
     M,
     N,
     stride_a_scale_m,
     stride_b_scale_n,
 ):
-    if SCALING_ROWWISE:
+    if SCALE_RECIPE_A == 1 and SCALE_RECIPE_B == 1:  # (ScalingType.RowWise, ScalingType.RowWise)
```
njriasan, line 400: What happens if `SCALE_RECIPE_A == 1` and `SCALE_RECIPE_A != SCALE_RECIPE_B`? The comments seem to imply you can't mix scaling types, so should we add an assert to the else case?

torch/_inductor/kernel/mm.py
```diff
@@ -1253,6 +1475,93 @@ def tuned_sparse_semi_structured_mm(
     )
 
 
+scaling_pairs = [
```
njriasan, line 1544: Is this the limited pairs of allowed values or is any combination legal. For example, could I ever mix tensorwise and rowwise across A & B? If not, then maybe it would be better to encode these in a single variable which would be easier to validate.

torch/_inductor/kernel/mm.py
```diff
@@ -1253,6 +1475,93 @@ def tuned_sparse_semi_structured_mm(
...
+def get_scaling_options(
+    mat_a: Any,
+    mat_b: Any,
+    scale_a_size: torch.Tensor,
+    scale_b_size: torch.Tensor,
+) -> tuple[ScalingType, ScalingType]:
+    for scale_option_a, scale_option_b in scaling_pairs:
+        if is_desired_scaling(
+            mat_a, scale_a_size, scale_option_a
+        ) and is_desired_scaling(mat_b, scale_b_size, scale_option_b, transpose=True):
```
njriasan, line 1627: Why is transpose hardcoded here? Is this not a function of the layouts?

njriasan (approved):
@jananisriram See my feedback. Overall this seems good, but I have some suggestions to have a single point of failure for the allowed scaling options. I'll defer to you though regarding what combinations should actually be supported and how to enforce those.

torch/_inductor/kernel/mm.py
jananisriram: this is the limited set of pairs that are legal for now (since these are the only combinations supported in aten) and seems to be as similar as we can get to the aten approach. by "encoding in a single variable," do you mean that we assign some variable to each pair, like `tensorwise_scaling = (TensorWise, TensorWise)`? i feel that the latter assignment will get complicated, considering that it's potentially possible that we support other combinations in the future

torch/_inductor/kernel/mm.py
jananisriram: we need to verify the stride values based on the transposed version of the second tensor, like aten's current implementation; it's possible, however, that this is a function of max-autotune enablement

torch/_meta_registrations.py
jananisriram: good point, thanks! moved the `ceil_div` function to a location where i can use it here as well

torch/_inductor/kernel/mm.py
jananisriram: we add asserts for scaling recipes earlier (when we check that the pairing is valid), so in this function we can assume that the pairing is either (1, 1) (RowWise, RowWise) or (0, 0) (TensorWise, TensorWise)

PR #165119, opened 2025-10-10
[CodeClean] Replace std::runtime_error with TORCH_CHECK

fffrog:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.12.0) (oldest at bottom):
* __->__ #165119
* #165139

As the title stated.

**Changes**:
- torch/csrc/inductor(Part 2)

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben

PR #165139, opened 2025-10-10
[CodeClean] Remove the Unused MACRO for AOT Inductor Runtime

fffrog:
* #165119
* __->__ #165139

As the title stated.

- AOTI_TORCH_CHECK depend on TORCH_CHECK_MSG which located in c10/util/Exception.h, which maybe break BC
- AOTI_TORCH_CHECK is not used everywhere
- STD_TORCH_CHECK have ABI check tests.


janeyx99 (commented):
Hi! shim.h is promised to be ABI stable, which means that we cannot remove functions in it like normal with a BC breaking policy.

I do agree that aoti_torch_check and the macros are less recommended than STD_TORCH_CHECK, so I'd recommend documenting that instead of removing. One day much later down the road, we can free ourselves to remove the shims that are long outdated/unused, but not now.

Skylion007 (approved):

janeyx99 (approved):
cc @desertfire

jeanschmidt:
So, after internal review, we identified that we do still rely on this API. This means that this should not be removed for now. In this case, I have no intent to re-merge this PR.

cc @janeyx99 @fffrog

fffrog:
> So, after internal review, we identified that we do still rely on this API. This means that this should not be removed for now. In this case, I have no intent to re-merge this PR.

Got it, thank you for detailed explanation.

PR #165152, opened 2025-10-10
Widen ops support to take in IntHOArrayRef vs only std::vec

janeyx99:
* #165953
* #165153
* __->__ #165152
* #164991


test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
```diff
@@ -311,10 +311,9 @@ void boxed_fill_infinity(
 }
 
 Tensor my_pad(Tensor t) {
-  std::vector<int64_t> padding = {1, 2, 2, 1};
   std::string mode = "constant";
   double value = 0.0;
-  return pad(t, padding, mode, value);
+  return pad(t, {1, 2, 2, 1}, mode, value);
```
mikaylagawarecki, line 316: n00b q for my understanding: since ArrayRef does not own the underlying data what keeps the initializer list alive

mikaylagawarecki (approved):

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
janeyx99: this function local scope, i believe

PR #165154, opened 2025-10-10
Add scaled_grouped_mm_v2 and python API

slayton58:
* __->__ #165154

Summary:

* Add `torch._scaled_grouped_mm_v2` with more functionality and
  extensibility for future formats
* Add `torch.nn.functional.scaled_grouped_mm` as public entrypoint
* Test both original and v2 functionality

Test Plan:

```
pytest -svv -k grouped test/test_scaled_matmul_cuda.py
```

Reviewers:

Subscribers:

Tasks:

Tags:


aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -2789,6 +2789,130 @@ _scaled_grouped_mm_cuda(
...
+          IntArrayRef scale_recipe_a,
+          IntArrayRef swizzle_a,
+          ArrayRef<Tensor> scale_b,
+          IntArrayRef scale_recipe_b,
+          IntArrayRef swizzle_b,
+          const std::optional<Tensor>& offs,
+          const std::optional<Tensor>& bias,
+          const std::optional<c10::ScalarType> out_dtype,
+          bool use_fast_accum) {
+  bool allowed_device = _scaled_mm_allowed_device(/*sm90_only*/true, /*sm100_only*/true);
```
drisspg, line 2833: does the Mi300 path work w/ this?

slayton58: It does, `_scaled_mm_allowed_device` has a `USE_ROCM` guarded path ([here](https://github.com/pytorch/pytorch/blob/9107d14463cdc7babc084c8c40e9fc49d29bdbbc/aten/src/ATen/native/cuda/Blas.cpp#L1063)) which ignores the `sm{90,100}_only` arguments and checks for applicable AMD GPUs

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -2789,6 +2789,130 @@ _scaled_grouped_mm_cuda(
...
+  TORCH_CHECK_VALUE(!check_valid_strides_and_return_transposed(mat_a), "Expected mat1 to not be transposed");
+  TORCH_CHECK_VALUE(check_valid_strides_and_return_transposed(mat_b), "Expected mat2 to be transposed");
+  TORCH_CHECK_VALUE(mat_a.dim() == 2 || mat_a.dim() == 3, "mat_a has to be 2 or 3d");
+  TORCH_CHECK_VALUE(mat_b.dim() == 2 || mat_b.dim() == 3, "mat_b has to be 2 or 3d");
+  const bool a_is_2d = mat_a.dim() == 2;
+  const bool b_is_2d = mat_b.dim() == 2;
+
+  // NOTE(slayton): For sub-1B formats want contraction_dim argument?
+  if (!a_is_2d || !b_is_2d) {
+    TORCH_CHECK_VALUE(mat_a.size(-1) == mat_b.size(-2), "contraction dimension of mat_a and mat_b must match");
```
slayton58: @danielvegamyhre In real-world use-cases, do you end up needing to pass any transposed inputs? If so, I think we want a `contraction_dim` argument, as `.t()` becomes non-"free" for sub-1B formats (like `e2m1x2`)

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -2789,6 +2789,130 @@ _scaled_grouped_mm_cuda(
...
+          scale_b[0],
+          offs,
+          bias,
+          use_fast_accum,
+          out);
+    case ScaledGemmImplementation::MXFP8_MXFP8:
+      // MXFP8 expects float8_e8m0fnu scales.
+      TORCH_CHECK_VALUE(scale_a[0].scalar_type() == at::kFloat8_e8m0fnu && scale_b[0].scalar_type() == at::kFloat8_e8m0fnu,
+          "For MXFP8 grouped gemm, both scales must be float8_e8m0fnu tensors.");
+      TORCH_CHECK_VALUE(swizzle_a_enum[0] == SwizzleType::SWIZZLE_32_4_4 && swizzle_b_enum[0] == SwizzleType::SWIZZLE_32_4_4,
```
slayton58: Note to self: ROCM doesn't need swizzle afaik

drisspg: cc @jeffdaily @petrex  can you confirm? I read through https://rocm.blogs.amd.com/software-tools-optimization/matrix-cores-cdna/README.html

but still wasn't 100% sure

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -2585,6 +2587,11 @@ _mx8_mx8_bf16_grouped_mm_fbgemm(
     TORCH_CHECK_VALUE(is_2d_2d || is_2d_3d, "MXFP8 grouped GEMM currently only supports 2d-2d and 2d-3d cases");
     TORCH_CHECK_VALUE(offs.has_value(), "MXFP8 2d-2d and 2d-3d grouped GEMMs requires offsets");
     TORCH_CHECK_VALUE(out.scalar_type() == at::kBFloat16, "Only bf16 out_dtype is supported for MXFP8 grouped gemm");
+    // MXFP8 expects float8_e8m0fnu scales.
+    TORCH_CHECK_VALUE(scale_a.scalar_type() == at::kFloat8_e8m0fnu && scale_b.scalar_type() == at::kFloat8_e8m0fnu,
+        "For MXFP8 grouped gemm, both scales must be float8_e8m0fnu tensors.");
+    TORCH_CHECK_VALUE(swizzle_a == SwizzleType::SWIZZLE_32_4_4 && swizzle_b == SwizzleType::SWIZZLE_32_4_4,
```
drisspg, line 2597: This is not run on rocm? I can't find the macros for IS ROCM lol

aten/src/ATen/native/cuda/Blas.cpp
danielvegamyhre: (sorry missed this comment somehow) - discussed this offline but for posterity's sake:
- fbgemm mx8mx8bf16 grouped mm API requires the B tensor be non-transposed (e.g., E,N,K)
- torch._scaled_grouped_mm requires the B tensor be pre-transposed (e.g., E,K,N) so before dispatching to fbgemm we do a `B.transpose(-2, -1)`

>If so, I think we want a contraction_dim argument, as .t() becomes non-"free" for sub-1B formats (like e2m1x2)

Could we just update fbgemm to have consistent API with torch._scaled_grouped_mm, accepting B as pre-transposed? Would need to take a look at how this affects perf but it should theoretically be fine I believe

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -2789,6 +2803,126 @@ _scaled_grouped_mm_cuda(
       out);
 }
 
+std::array<std::tuple<std::string, acceptance_fn, ScaledGemmImplementation>, 2> scale_grouped_kernel_dispatch = {{
```
drisspg, line 2813: nit, can this go in an anonymous?

drisspg (approved):

aten/src/ATen/native/cuda/Blas.cpp
slayton58: In this case that'd do it, but in general we do want to be able to support the full matrix of potential contractions (just like regular gemm does with it's `T/NT` modes), so both cases should ideally work.

aten/src/ATen/native/cuda/Blas.cpp
slayton58: Yes it can, and it now has :)

aten/src/ATen/native/cuda/Blas.cpp
slayton58: My reading is the same as yours @drisspg, and I'll make the appropriate change in the code, but a confirmation would be very welcome!

aten/src/ATen/native/cuda/Blas.cpp
slayton58: Had the change locally just not pushed :D Explicitly guarding with `USE_ROCM` now.

danielvegamyhre (approved):

PR #163792, opened 2025-09-24
[aoti] Save compute information

angelayi:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #163792
* #163779

Metadata looks like:
```
{
  'AOTI_DEVICE_KEY': 'cpu',
  'AOTI_PLATFORM': 'linux',
  'AOTI_MACHINE': 'x86_64',
  'AOTI_CPU_ISA': 'AVX512',
  'AOTI_COMPUTE_CAPABILITY': '90'
}
```

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben

PR #165157, opened 2025-10-10
fix(load_aoti): Explicitly import codecache to get device info

cbensimon:
Currently leads to an `AttributeError` if `torch._inductor.codecache` import has not been triggered by another code path.

Follow-up of:
- https://github.com/pytorch/pytorch/pull/163792

cc @angelayi


angelayi (approved):
thanks!

PR #165171, opened 2025-10-10
[lint] workflow consistency linter to look at all files instead of just changed files

clee2000:
As in title

If you change only one workflow file, lintrunner (default arg, also the one in CI since it only inputs changed files) won't look at other files in the repo, but the sync-tag might come from those other files

This makes it so that it looks at all workflow files so it will catch those failures

Also change output line so it prints which file + which job it is different from

Pros:
catches errors

Cons:
unusual behavior (getting around what lintrunner says the linter should run on)

### DIFF
--- a/tools/linter/adapters/workflow_consistency_linter.py
+++ b/tools/linter/adapters/workflow_consistency_linter.py
@@ -122,13 +122,15 @@ if __name__ == "__main__":
         if not path.is_file() or path.suffix not in {".yml", ".yaml"}:
             continue
         workflow = load_yaml(path)
-        if is_workflow(workflow):
-            jobs = workflow.get("jobs", {})
-            for job_id, job in jobs.items():
-                res = get_jobs_with_sync_tag(job)
-                if res is not None:
-                    sync_tag, path, job_dict = res
-                    tag_to_jobs[sync_tag].append((path, {job_id: job}))
+        if not is_workflow(workflow):
+            continue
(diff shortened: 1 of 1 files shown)


tools/linter/adapters/workflow_consistency_linter.py
```diff
@@ -93,41 +115,36 @@ def print_lint_message(path: Path, job: dict[str, Any], sync_tag: str) -> None:
...
+        if not path.is_file() or path.suffix not in {".yml", ".yaml"}:
+            continue
+        workflow = load_yaml(path)
+        if is_workflow(workflow):
+            jobs = workflow.get("jobs", {})
+            for job_id, job in jobs.items():
+                res = get_jobs_with_sync_tag(job)
+                if res is not None:
+                    sync_tag, path, job_dict = res
+                    tag_to_jobs[sync_tag].append((path, {job_id: job}))
```
malfet: Disassemble towers of doom
```suggestion
        if not is_workflow(workflow):
            continue
        jobs = workflow.get("jobs", {})
        for job_id, job in jobs.items():
            res = get_jobs_with_sync_tag(job)
            if res is None:
                continue
            sync_tag, path, job_dict = res
            tag_to_jobs[sync_tag].append((path, {job_id: job}))
```

malfet (approved):

izaitsevfb (approved):

atalman (approved):

PR #165179, opened 2025-10-10
[RFC] Add pyrefly to lintrunner

maggiemoss:
This will add pyrefly to lint runner as a warning only - and allow us to collect feedback about the tool before switching to pyrefly as the main type checker. 

References the steps outlined here: : https://github.com/pytorch/pytorch/issues/163283: 

test plan:
`lintrunner init`
`lintrunner` 
confirm when pyrefly errors are present results look like: https://gist.github.com/maggiemoss/e6cb2d015dd1ded560ae1329098cf33f

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @voznesenskym @penguinwu @EikanWang @Guobing-Chen @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela


.lintrunner.toml
```diff
@@ -209,6 +209,42 @@ command = [
     '@{{PATHSFILE}}'
 ]
 
+
+[[linter]]
+code = 'PYREFLY'
+include_patterns = [
+    'torch/**/*.py',
```
ezyang, line 216: Hmm what about pyi files

.lintrunner.toml
```diff
@@ -209,6 +209,42 @@ command = [
...
+    'types-setuptools==79.0.0.20250422',
+    'types-jinja2==2.11.9',
+    'types-colorama==0.4.6',
+    'filelock==3.18.0',
+    'junitparser==2.1.1',
+    'rich==14.1.0',
+    'pyyaml==6.0.2',
+    'optree==0.13.0',
+    'dataclasses-json==0.6.7',
+    'pandas==2.2.3',
```
ezyang: Do you really need this many deps here?

.lintrunner.toml
maggiemoss: oh, nice catch! thank you :)

.lintrunner.toml
maggiemoss: I copied from mypy - a lot of these are making sure we have the stubs, etc. installed

maggiemoss: Update: I went through and tried to determine if removing these would cause type errors - we need to install them to resolve imports in some files.

.lintrunner.toml
```diff
@@ -209,6 +209,42 @@ command = [
     '@{{PATHSFILE}}'
 ]
 
+
+[[linter]]
+code = 'PYREFLY'
+include_patterns = [
+    'torch/**/*.py',
+    'torch/**/*.pyi',
```
ezyang, line 217: I feel torchgen and functorch should probably also be added here 🤔

ezyang (approved):

ezyang (approved):

.lintrunner.toml
maggiemoss: I'll add that in!

PR #165180, opened 2025-10-10
[CI] Default keep-going true for tags of form ciflow/something/commitsha

clee2000:
Tags of the form `ciflow/something/commitsha` are usually created by running the workflow from HUD


huydhn (approved):
LGTM!

PR #161013, opened 2025-08-19
[CI] Add basic CUDA 13.0 periodic test

tinglvv:
https://github.com/pytorch/pytorch/issues/159779

cc @atalman

PR #165183, opened 2025-10-10
[CI] Put the no gpu tests on machines that don't have gpus

clee2000:
I think this is just a copy paste error?

NS: Introduced by https://github.com/pytorch/pytorch/pull/161013

Not sure where it got copied from though, the other set of no gpu tests for the other cuda version already have cpu runners


malfet (approved):
Thank you!

PR #165185, opened 2025-10-10
Fix h100 daily inductor running dispatch

yangw-dev:
casued by merged pr: https://github.com/pytorch/pytorch/commit/e7ed1a00eb5510d1c7dccd17b5c0ebb54231284f

the if condition should also updated


malfet (approved):
LGTM, can we write a lint for it though?

huydhn (approved):
Good catch!  We wouldn't be to spot this without the notification

PR #165191, opened 2025-10-10
[Bugfix][vLLM] Explicitly do not support instead of crashing for named tuples in infer schema

Lucaskabela:
Fixes https://github.com/vllm-project/vllm/issues/25270 by being explicit in erroring; previously we had a cryptic `__origin__ undefined` error, but now should give proper error message that we don't support NamedTuples in schema

Test with
```
python test/test_custom_ops.py TestCustomOp.test_unsupported_param_types
```

### DIFF
--- a/test/test_custom_ops.py
+++ b/test/test_custom_ops.py
@@ -1070,8 +1070,9 @@ class TestCustomOp(CustomOpTestCaseBase):
             del foo
 
         # Define a named tuple for a Point with x and y coordinates
-        Point = collections.namedtuple('Point', ['x', 'y'])
+        Point = collections.namedtuple("Point", ["x", "y"])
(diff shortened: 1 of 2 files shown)


test/test_custom_ops.py
```diff
@@ -1069,6 +1069,16 @@ def foo(x: Tensor, y: Callable) -> Tensor:
 
             del foo
 
+        # Define a named tuple for a Point with x and y coordinates
+        Point = collections.namedtuple("Point", ["x", "y"])
+        with self.assertRaisesRegex(ValueError, "unsupported type"):
```
zou3519, line 1074: can you make sure "Point" is in the error message? (by putting it into the string.

zou3519 (approved):

PR #165197, opened 2025-10-10
[7/N][DTensor device order] Support DCP with shard_order

zpcore:
DCP for tensor with `shard_order` works as below:
```
+------------------------------------+
|     DTensor (dt) + shard_order     |
+------------------------------------+
         |
         |  calls
         v
+---------------------------------------------+
| torch.distributed.checkpoint.save           |
+---------------------------------------------+
         |
         |  triggers
         v
+-------------------------------------------------------------+
| compute_local_shape_and_global_offset (adjusts shard order) |
+-------------------------------------------------------------+
         |
         |  after save, becomes
         v
+-------------------------+
|      checkpoint file    |
+-------------------------+
         |
         |  (later) call
         v
+---------------------------------------------+
| torch.distributed.checkpoint.load           |
+---------------------------------------------+
         |
         |  on top of
         v
+-------------------+
|  Tensor (empty)   |
+-------------------+
         |
         |  after load, becomes
         v
+----------------------------------------------------------------------+
| Tensor, which equals to full DTensor (dt)  with default device order |
+----------------------------------------------------------------------+
```

* __->__ #165197
* #165092


torch/distributed/tensor/_utils.py
```diff
@@ -414,12 +473,12 @@ def try_find_mesh_from_args(
...
     for arg in args:
-        if isinstance(arg, (dtensor.DTensor, DTensorSpec)):
+        if isinstance(arg, dtensor.DTensor | DTensorSpec):
             return arg.device_mesh
         elif (
-            isinstance(arg, (list, tuple))
+            isinstance(arg, list | tuple)
             and len(arg) > 0
-            and isinstance(arg[0], (dtensor.DTensor, DTensorSpec))
+            and isinstance(arg[0], dtensor.DTensor | DTensorSpec)
```
ezyang, line 481: What's generating these edits 🤔

test/distributed/checkpoint/test_state_dict.py
```diff
@@ -1094,5 +1101,59 @@ def test_no_dist(self) -> None:
...
+        """
+        device_mesh = init_device_mesh(self.device_type, (2, 2))
+        checkpoint_path = os.path.join(self.temp_dir, "checkpoint")  # pyright: ignore[reportAttributeAccessIssue]
+
+        # Create a distributed tensor with custom shard_order
+        x = torch.randn(8, 4, device=self.device_type)
+        x_dt = distribute_tensor(x, device_mesh, shard_order=shard_order)
+
+        # Save the distributed tensor to checkpoint. By default full_tensor()
+        # will be triggered to gather DTensor.
```
ezyang: No, I'm confused. Isn't the point of DCP that we don't have to gather when we checkpoint?

torch/distributed/tensor/_utils.py
zpcore: `lintrunner -a` forced me to update to the union syntax `|` once I have modification to that file.

test/distributed/checkpoint/test_state_dict.py
zpcore: I think there is an option to control whether to gather or not, but we must make sure the device order is correctly adjusted when saving.

torch/distributed/tensor/_utils.py
```diff
@@ -19,60 +19,110 @@
...
+            mesh_shape = (2, 2, 2)  # 3D device mesh
+            placements = [Shard(0), _StridedShard(0, split_factor=2), Shard(0)]
+            shard_order = None  # must be None when `_StridedShard` exists
+        Output:
+            [(0, Shard(0)), (2, Shard(0)), (1, Shard(0))]
+
+        Explanation: The strided shard on mesh dim 1 is deferred and appears after
+        the regular shard on mesh dim 2.
+
+    Example 2 (Explicit mutated order, must without _StridedShard):
```
ezyang: Not sure why you say "mutated" here, * must be without

ezyang (commented):
I'm endorsing the change, it looks plausible (since StridedShard seems to imply DCP already supports reordered sharding), but you need a DCP review

test/distributed/checkpoint/test_state_dict.py
fegin: The "gather" is to gather the control plan data not the data plan. So `full_tensor()` is never called during checkpoint. If your PR is going to change this behavior, then it is wrong.

test/distributed/checkpoint/test_state_dict.py
```diff
@@ -1094,5 +1101,59 @@ def test_no_dist(self) -> None:
...
+            {},  # default to all replicate
+            {1: [1, 0]},  # reversed order on tensor dim 1
+        ],
+    )
+    def test_dtensor_checkpoint_with_shard_order(self, shard_order):
+        """
+        Test that DTensor with arbitrary shard_order can be saved and loaded via checkpoints.
+        """
+        device_mesh = init_device_mesh(self.device_type, (2, 2))
+        checkpoint_path = os.path.join(self.temp_dir, "checkpoint")  # pyright: ignore[reportAttributeAccessIssue]
```
fegin: You should just use `self.temp_dir` as the checkpoint_path. `with_temp_dir` should guarantee that the folder exists for all ranks.

test/distributed/checkpoint/test_state_dict.py
```diff
@@ -1094,5 +1101,59 @@ def test_no_dist(self) -> None:
...
+    @parametrize(
+        "shard_order",
+        [
+            {0: [0, 1]},  # default order
+            {0: [1, 0]},  # reversed order on tensor dim 0
+            {},  # default to all replicate
+            {1: [1, 0]},  # reversed order on tensor dim 1
+        ],
+    )
+    def test_dtensor_checkpoint_with_shard_order(self, shard_order):
```
fegin: You should add another test to verify that the existing checkpoint saved with Strided DTensor can be loaded back.

test/distributed/checkpoint/test_state_dict.py
```diff
@@ -1094,5 +1101,59 @@ def test_no_dist(self) -> None:
...
+        x = torch.randn(8, 4, device=self.device_type)
+        x_dt = distribute_tensor(x, device_mesh, shard_order=shard_order)
+
+        # Save the distributed tensor to checkpoint. By default full_tensor()
+        # will be triggered to gather DTensor.
+        state_dict_to_save = {"x_dt": x_dt}
+        save(state_dict=state_dict_to_save, checkpoint_id=checkpoint_path)
+
+        # Create a new tensor with the same shape for loading
+        x_new = torch.zeros_like(x)
```
fegin: This is doing DTensor->full tensor resharding in DCP. 

We should also verify DTensor(order_shard)->DTensor(StridedSharding) and DTensor(order_shard)->(StridedSharding). These two cases shouldn't invoke any communication.

test/distributed/checkpoint/test_state_dict.py
zpcore: Oh, good to know! I didn't really check if full_tensor() has been called or not. Let me double check.

torch/distributed/tensor/_utils.py
zpcore: I was going to say the order is changed from default. Removed to make it clear.

test/distributed/checkpoint/test_state_dict.py
zpcore: TIL

test/distributed/checkpoint/test_state_dict.py
zpcore: Added those two tests.

test/distributed/checkpoint/test_state_dict.py
zpcore: Added those two tests, PTAL, thanks!

zpcore:
I further worked on DCP for AutoParallel in https://github.com/meta-pytorch/autoparallel/pull/206 and noticed that this PR is not sufficient to support DCP. This is because when we work with DCP for torch.nn.module, we will call sth like:
```
model_state_dict, optim_state_dict = get_state_dict(
                    parallel_mod,
                    optimizer,
                )
```
Internally, getting a state dict will trigger aten.detach op to create a dict with reference to DTensor parameters in the model. If our DTensor propagation strategy (especially for `detach`)  [link](https://github.com/pytorch/pytorch/blob/51319ca090bc7458168a8451c04ca7e021a72693/torch/distributed/tensor/_ops/_tensor_ops.py#L59-L69) doesn't propagate the `shard_order` then we will lose the device order information, thus results in incorrect state when resume.

zpcore:
Optimizer also doesn't work because optimizer state dict will call op like `toch.zeros_like` as, for example, [here](https://github.com/pytorch/pytorch/blob/eaeaa08e3a8071be46f833f7b46aa642ec14e0f7/torch/optim/adam.py#L178).

So we need to support `shard_order` propagation at least for `detach` (fixes model state dict) and `zeros_like` (fixes optimizer state). I should prioritize on enabling propagating for `shard_order` op strategy first.

ezyang:
This PR is obsolete under the new StridedShard plan right?

PR #165202, opened 2025-10-10
[annotate] Annotation should be mapped across submod

yushangdi:
The match for backward nodes might be in a different submod, so we should check all submod for potential matches. 

In flex attention, this could happen if `mask_mod` has operations (such as index) that increase the seq_nr of the forward graph nodes. Then the backward flex_attention nodes cannot find a match in its own subgraph.

```
python test/functorch/test_aot_joint_with_descriptors.py -k preserve_annotate
```

Also tested on torchtitan joint_graph_runner branch. The flex_attention backward nodes are annotated now. 

```
NGPU=8   CONFIG_FILE="./torchtitan/models/llama3/train_configs/debug_model.toml"   LOG_RANK=0   TRAIN_FILE="torchtitan.train"   TORCHFT_LIGHTHOUSE="http://localhost:29510"   PYTORCH_ALLOC_CONF="expandable_segments:True"   torchrun     --nproc_per_node=8     --rdzv_backend c10d     --rdzv_endpoint="localhost:0"     --local-ranks-filter 0     --role rank     --tee 3     -m torchtitan.train     --job.config_file ./torchtitan/models/llama3/train_configs/debug_model.toml     --model.name joint_graph_runner.llama3     --compile.enable     --parallelism.data_parallel_shard_degree=2     --parallelism.tensor_parallel_degree=4     --model.flavor=debugmodel_flex_attn
```

### DIFF
--- a/torch/_functorch/_aot_autograd/utils.py
+++ b/torch/_functorch/_aot_autograd/utils.py
@@ -404,28 +404,28 @@ def root_module_when_exporting_non_strict(flat_fn):
         return None
 
 
-def _copy_fwd_metadata_to_bw_nodes(fx_g):
-    def _is_forward_node_with_seq_nr(node):
-        # For now, assume that if nn_module_stack_metadata is populated, this
-        # node is from the forward. Ignore nodes without `seq_nr`.
-        # TODO(future): there is likely a less brittle way to do this by walking
-        # the descendants of graph inputs corresponding to fwd inputs, didn't
-        # seem obvious at first glance on how to partition graph inputs into
-        # fwd vs bwd without relying on string names.
-        return (
-            node.meta.get("partitioner_tag") != "is_backward" and "seq_nr" in node.meta
-        )
+def _is_forward_node_with_seq_nr(node: torch.fx.Node) -> bool:
+    # For now, assume that if nn_module_stack_metadata is populated, this
+    # node is from the forward. Ignore nodes without `seq_nr`.
+    # TODO(future): there is likely a less brittle way to do this by walking
+    # the descendants of graph inputs corresponding to fwd inputs, didn't
+    # seem obvious at first glance on how to partition graph inputs into
+    # fwd vs bwd without relying on string names.
+    return node.meta.get("partitioner_tag") != "is_backward" and "seq_nr" in node.meta
 
-    def _is_backward_node_with_seq_nr(node):
-        # For now, assume that if nn_module_stack_metadata is not populated,
-        # this node is from the backward. Ignore nodes without `seq_nr`.
-        # TODO(future): there is likely a less brittle way to do this, same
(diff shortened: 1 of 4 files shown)


torch/fx/traceback.py
```diff
@@ -388,3 +389,18 @@ def get_graph_provenance_json(graph: Graph) -> dict[str, Any]:
             },
         )
         return {}
+
+
+def get_custom_metadata(gm):
```
SherlockNoMad: do you expect this to be used by end user direction, or is it for testing only?

make it private if later?

torch/fx/traceback.py
```diff
@@ -388,3 +389,18 @@ def get_graph_provenance_json(graph: Graph) -> dict[str, Any]:
...
+        custom_metadata = []
+        for node in gm.graph.nodes:
+            if hasattr(node, "meta") and node.meta.get("custom", None):
+                custom_metadata.append((node.op, node.name, node.meta["custom"]))
+            if node.op == "get_attr" and isinstance(
+                getattr(gm, node.target), GraphModule
+            ):
+                custom_metadata.append(helper(getattr(gm, node.target)))
+        return custom_metadata
+
```
SherlockNoMad, line 407: assert input is GraphModule, not fx.Node

torch/fx/traceback.py
yushangdi: it's for testing only. Since it's used across different testing folders, I moved it here. Yeah let me make it private

torch/_functorch/_aot_autograd/utils.py
```diff
@@ -458,15 +464,35 @@ def copy_fwd_metadata_to_bw_nodes(fx_g):
...
-    for node in fx_g.graph.nodes:
-        if node.op == "get_attr":
-            submod = getattr(fx_g, node.target)
-            if isinstance(submod, torch.fx.GraphModule):
-                copy_fwd_metadata_to_bw_nodes(submod)
-    _copy_fwd_metadata_to_bw_nodes(fx_g)
+    # Build a global mapping of seq_nr to forward nodes across all subgraphs
+    fwd_seq_nr_to_node: dict[str, torch.fx.Node] = {}
+
+    def collect_all_submodules(module):
```
SherlockNoMad: for nn.module, you can use modules() or named_modules(). 

Not sure if these two functions are reliable for GraphModule.

SherlockNoMad (approved):
lgtm with minor comments. 
get_custom_metadata doesn't seems useful to user, let's make it private.

torch/_functorch/_aot_autograd/utils.py
yushangdi: good point. I used .modules() now.  I see it used in export, so it should be ok.

PR #165204, opened 2025-10-11
Quick fix of torch.save memory leak

ppwwyyxx:
Fix the memory leak shown in https://github.com/pytorch/pytorch/issues/149846#issuecomment-3392634572

### DIFF
--- a/torch/serialization.py
+++ b/torch/serialization.py
@@ -1140,6 +1140,9 @@ def _legacy_save(obj, f, pickle_module, pickle_protocol) -> None:
     pickler = PyTorchLegacyPickler(f, protocol=pickle_protocol)
     pickler.dump(obj)
 
+    # The class def keeps the persistent_id closure alive, leaking memory.
+    persistent_id = None
+
     serialized_storage_keys = sorted(serialized_storages.keys())
     pickle_module.dump(serialized_storage_keys, f, protocol=pickle_protocol)
     f.flush()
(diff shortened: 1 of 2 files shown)


torch/serialization.py
```diff
@@ -1140,6 +1140,9 @@ def persistent_id(self, obj):
     pickler = PyTorchLegacyPickler(f, protocol=pickle_protocol)
     pickler.dump(obj)
 
+    # The class def keeps the persistent_id closure alive, leaking memory.
+    persistent_id = None
```
cyyever: To avoid the liner, can `del persistent_id` releases memory here?

ppwwyyxx: done

torch/serialization.py
```diff
@@ -1221,6 +1224,10 @@ def persistent_id(self, obj):
 
     pickler = PyTorchPickler(data_buf, protocol=pickle_protocol)
     pickler.dump(obj)
+
+    # The class def keeps the persistent_id closure alive, leaking memory.
+    del persistent_id
```
cyyever, line 1229: It is deleted twice..

ppwwyyxx: It's two separate functions each having a delete.

Also, delete is insufficient to silence the linter. So I had to add a noqa

cyyever: Sometimes delete cannot trigger gc immediately. You may need to check the memory usage and use empty_cache..

ppwwyyxx: delete is not for triggering the gc. delete makes sure the closure will not be referenced by the class definition.

This issue is not related to cuda memory management so 'empty_cache' is irrelevant.

ezyang (approved):

Issue #164878, opened 2025-10-07
Ban and remove plain asserts with no message in our python code


albanD:
In a similar spirit to https://github.com/pytorch/pytorch/issues/148114

We should remove asserts without any message explaining what is happening.
On top of that, we should move them to proper errors to avoid any issue with python -O.

There are two parts here:
- [ ] Enable Ruff lint for this https://docs.astral.sh/ruff/rules/assert/ (with appropriate skips)
- [ ] Remove all the existing ones

cc @malfet

zhudada0120:
@RohitRathore1  I'll handle torch/jit/ .

Skylion007:
There is a ruff rule that seperates out commpand assets joined with and into seperate assert. I would advocate doing that as well

PR #165226, opened 2025-10-11
[distributed] Replace 54 assert statements in tensor/_ops/_tensor_ops.py

RohitRathore1:
Replace assert statements with explicit if/raise patterns to prevent assertions from being disabled with Python -O flag.

Fixes partially #164878

cc: @albanD 


albanD (approved):
Thanks!

PR #165229, opened 2025-10-11
[distributed] Replace 94 assert statements in tensor ops files

RohitRathore1:
Replace assert statements with explicit if/raise patterns in:
- _math_ops.py (43 asserts)
- _matrix_ops.py (27 asserts)
- _view_ops.py (24 asserts)

This prevents assertions from being disabled with Python -O flag.

Fixes partially #164878. 

cc: @albanD 


albanD (approved):
Thanks!

PR #165242, opened 2025-10-11
[dynamo, guards] Better error messages when generated guard fails on the same frame

jmaczan:
Not sure what exactly we want to have in the message, but that's easy to adjust. I tried to find a reliable test to reproduce this message (happens only when a guard fails right after it's created), but I ended up mocking a `guard_manager.check` function to return `False` to trigger this behavior. I think that's fine, because any other case that we pick (like datetime.now()), we want to patch one day anyway, so every time we make the next patch, will need to chase for another repro test

@williamwen42 

Fixes #164990


test/dynamo/test_repros.py
```diff
@@ -7369,6 +7369,40 @@ def fn():
...
+                    return x + 1
+
+            model = Model()
+            x = torch.randn(5)
+
+            with self.assertRaises(AssertionError) as e:
+                torch.compile(model)(x)
+
+        msg = str(e.exception)
+        assert (
```
williamwen42: use `self.assertIn`?

williamwen42 (approved):

test/dynamo/test_repros.py
jmaczan: Changed

jmaczan (commented):
Updated and resolved lint issue with too long line

PR #165254, opened 2025-10-12
[MPS] Fix internal assertion in torch.linalg.solve for singular matrices

inventshah:
Fixes #163962 by special casing MPS in the negative status code branch in `_linalg_check_errors`.

Checks if info is [`MPSMatrixDecompositionStatus.singular`](https://developer.apple.com/documentation/metalperformanceshaders/mpsmatrixdecompositionstatus/singular) (which has a raw value of -2). I didn't find an official Apple source with this raw value (besides printing the enum value), so I'm not sure if we can (or should) depend on it? Is there a way to directly get the Objective-C enum value in C++?

### DIFF
--- a/aten/src/ATen/native/mps/operations/LinearAlgebra.mm
+++ b/aten/src/ATen/native/mps/operations/LinearAlgebra.mm
@@ -196,6 +196,28 @@ bool use_metal_mm(const Tensor& self, const Tensor& other, const Tensor& output)
        other.size(0) > max_stride_size || other.size(1) > max_stride_size);
 }
 
+void map_mps_decomposition_error_code_to_blas(const Tensor& status) {
+  Tensor status_flat = status.view(-1);
+
+  for (const auto i : c10::irange(status_flat.size(0))) {
+    int code = status_flat[i].item<int>();
+    switch (code) {
+      case MPSMatrixDecompositionStatusSuccess:
+        status_flat[i] = 0;
+        break;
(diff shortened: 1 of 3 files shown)


malfet (approved):

malfet (changes requested):
@inventshah do you mind moving the error handling mapping logic back to LinearAlgebra.mm, where it checks if error code is such and such and maps it back to BLAS error code?

inventshah:
@malfet moved. Note there is a minor correctness issue where the BLAS error code signals the first zero pivot but the MPS code doesn't return that detail, so its just positive.

I also realized `torch.linalg.lu_factor` also needs this treatment, but unless you'd like to do that in this PR, I'll do that separately once we settle on this new mapping function.

malfet:
@inventshah it's up to you, you can do in this PR or another one

aten/src/ATen/native/mps/operations/LinearAlgebra.mm
```diff
@@ -196,6 +196,28 @@ bool use_metal_mm(const Tensor& self, const Tensor& other, const Tensor& output)
        other.size(0) > max_stride_size || other.size(1) > max_stride_size);
 }
 
+void map_mps_decomposition_error_code_to_blas(const Tensor& status) {
+  Tensor status_flat = status.view(-1);
```
malfet:
```suggestion
  auto& status_flat = status.view(-1);
```

malfet (approved):

PR #165255, opened 2025-10-12
[distributed] Replace 89 assert statements in pipelining directory

RohitRathore1:
Replace assert statements with explicit if/raise patterns across 6 files:
- schedules.py (33 asserts)
- stage.py (22 asserts)
- _IR.py (11 asserts)
- microbatch.py (10 asserts)
- _schedule_visualizer.py (8 asserts)
- _backward.py (5 asserts)

This prevents assertions from being disabled with Python -O flag.

Fixes partially #164878 

cc: @albanD 

### DIFF
--- a/torch/distributed/pipelining/schedules.py
+++ b/torch/distributed/pipelining/schedules.py
@@ -1246,7 +1246,8 @@ def _add_send_recv(
         return False
 
     def _get_comms(action: _Action) -> tuple[_Action, _Action]:
-        assert _has_comms(action), f"{action} is not a valid comm action"
+        if not _has_comms(action):
+            raise AssertionError(f"{action} is not a valid comm action")
         stage_idx = action.stage_index
         ctype = action.computation_type
         mb_idx = action.microbatch_index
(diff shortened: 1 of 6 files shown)


albanD (approved):
Thanks!

albanD (approved):
Looks great, thank you!

H-Huang (approved):

torch/distributed/pipelining/schedules.py
```diff
@@ -1692,9 +1703,10 @@ def _step_microbatches(
                     computation_type = action.computation_type
                     mb_index = action.microbatch_index
                     stage_index = action.stage_index
-                    assert mb_index is not None, (
-                        "All currently supported action types require valid microbatch_index"
-                    )
+                    if not (mb_index is not None):
```
H-Huang, line 1706: nit: theres a few of these double negatives which can be simplified. I can fix it after this PR is landed though

wconstab (commented):
I didn't realize this was a thing. Makes sense, I guess, but it's a bummer to make every assertion 2 lines instead of one!

PR #165256, opened 2025-10-12
[distributed] Replace assert statements in distributed checkpoint with explicit checks

RohitRathore1:
Fixes partially #164878

cc: @albanD 

### DIFF
--- a/torch/distributed/checkpoint/format_utils.py
+++ b/torch/distributed/checkpoint/format_utils.py
@@ -84,7 +84,8 @@ class BroadcastingTorchSaveReader(StorageReader):
         # the entire checkpoint on each rank, hopefully preventing OOM issues
         # TODO: read on each host, instead of only the coordinator
         if self.is_coordinator:
-            assert self.checkpoint_id is not None
+            if self.checkpoint_id is None:
+                raise AssertionError()
             torch_state_dict = torch.load(
                 self.checkpoint_id, map_location="cpu", weights_only=False
             )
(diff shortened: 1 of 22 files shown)


torch/distributed/checkpoint/format_utils.py
```diff
@@ -84,7 +84,8 @@ def read_data(self, plan: LoadPlan, planner: LoadPlanner) -> Future[None]:
         # the entire checkpoint on each rank, hopefully preventing OOM issues
         # TODO: read on each host, instead of only the coordinator
         if self.is_coordinator:
-            assert self.checkpoint_id is not None
+            if self.checkpoint_id is None:
+                raise AssertionError
```
albanD: Missing message here and below in this file?

torch/distributed/checkpoint/optimizer.py
```diff
@@ -184,7 +182,8 @@ def create_local_plan(self) -> LoadPlan:
 
             offset = self.fqn_to_offset[fqn]
 
-            assert len(obj.local_shards()) == 1
+            if not len(obj.local_shards()) == 1:
+                raise AssertionError
```
albanD: Message

albanD (commented):
Thanks!
A couple instances are missing messages but overall looks great!

torch/distributed/checkpoint/format_utils.py
RohitRathore1: Fixed it, thanks for pointing!

albanD (approved):
Thanks!

PR #165261, opened 2025-10-12
feat: add out_dtype parameter support to torch.einsum

rudymentale:
## 🚀 Feature
Adds `out_dtype` parameter to `torch.einsum` for consistent precision control, aligning with other linear algebra functions like `torch.mm` and `torch.matmul`.

## Changes
- Modified C++ implementation in `Linear.cpp`
- Updated Python bindings in `functional.py` 
- Added parameter declaration in `native_functions.yaml`

## Usage
```python
result = torch.einsum("ij,jk->ik", A, B, out_dtype=torch.float32)


ngimel (changes requested):
1) LLM assistance on PRs has to be disclosed
2) This PR converts final result to requested dtype. By this time the accuracy loss has already happened

PR #165265, opened 2025-10-12
Fix loop pipelining for 2d/2d case of Triton grouped MM

alexsamardzic:
* __->__ #165265


ngimel (approved):
Awesome! So the gist of the fix is you are doing unmasked loads/mmas for most of the loop, and only the last iteratio possibly goes through masking?

Issue #99221, opened 2023-04-15
the error message of torch.nn.functional.embedding_bag is not clear


cheyennee:
### 🐛 Describe the bug

```
import torch
arg_1 = torch.as_tensor([2])
arg_2 = torch.rand([2, 3], dtype=torch.float32)
arg_3 = torch.as_tensor([0])
res = torch.nn.functional.embedding_bag(input=arg_1,weight=arg_2,offsets=arg_3)
print(res)
```
Above code will throw an exception: `RuntimeError: Index 0 is out of bounds: 2, range 0 to 2`. This error message is not clear, because the range of `Index 0` is greater than or equal to 0 and less than 2. The message `range 0 to 2` is a little confused and people will think that `Index 0` can take the value of 2.

### Versions

pytorch version: 2.0.0+cu118

cc @albanD @mruberry @jbschlosser @walterddr @mikaylagawarecki @malfet

MylesJP:
Hi Cheyennee, I can take a look at this if it's still open. I'm looking for a documentation-related issue for a school project so this should work well.

jbschlosser:
@MylesJP Feel free to open a PR addressing this and we'll review it

PR #165268, opened 2025-10-12
Fix MPS embedding_bag out-of-bounds error handling (#163630)

### LINKED ISSUE #163630: [MPS] torch.nn.functional.embedding_bag does not raise an error for invalid input
### 🐛 Describe the bug

```python
import torch

inputs = torch.tensor([0, 1, 6])  # Note: 6 is out of bounds for weight with size 4
weight = torch.randn(4, 2)
offsets = torch.tensor([0, 3])

try:
    out_cpu = torch.nn.functional.embedding_bag(inputs, weight, offsets)
    print("CPU output:", out_cpu)
except Exception as e:
    print("CPU error:", e) # catch the error for invalid indices

weight_mps = weight.to('mps')
indices_mps = inputs.to('mps')
offsets_mps = offsets.to('mps')
try:
    out_mps = torch.nn.functional.embedding_bag(indices_mps, weight_mps, offsets_mps)
    print("MPS output:", out_mps.cpu())
except Exception as e:
    print("MPS error:", e) # does not catch the error for invalid indices
```

### Versions

PyTorch version: 2.10.0a0+gitedafc90

cc @malfet @kulinseth @DenisVieriu97 @jhavukainen

roei-shlezinger:
## Summary
- align MPS embedding_bag with CPU by checking indices up front and raising the detailed “Index N … not in range” error
- keep a device-specific regression test in test/test_mps.py to ensure MPS throws the same style of message as CPU for OOB indices

Fixes #163630
Refs #99221

### DIFF
--- a/aten/src/ATen/native/mps/operations/EmbeddingBag.mm
+++ b/aten/src/ATen/native/mps/operations/EmbeddingBag.mm
@@ -7,6 +7,7 @@
 #include <ATen/native/Pool.h>
 #include <ATen/native/mps/OperationUtils.h>
 #include <ATen/native/mps/kernels/EmbeddingBag.h>
+#include <ATen/ops/aminmax.h>
 
 #include <fmt/format.h>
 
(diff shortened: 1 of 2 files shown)


aten/src/ATen/native/mps/operations/EmbeddingBag.mm
```diff
@@ -106,6 +108,24 @@
   params.mode = static_cast<EmbeddingBagMode>(mode);
   params.padding_idx = padding_idx;
 
+  if (num_indices > 0) {
+    auto indices_cpu = indices.to(kCPU);
+    AT_DISPATCH_INDEX_TYPES(indices_cpu.scalar_type(), "embedding_bag_mps_check_indices", [&]() {
+      const auto* indices_data = indices_cpu.const_data_ptr<index_t>();
+      for (const auto i : c10::irange(indices_cpu.numel())) {
```
can-gaa-hou: Can it be faster just finding the min/max value and compare it with `num_embeddings`?

roei-shlezinger: Thank you for the suggestion! 
I considered the min/max approach, but in my opinion the current loop is more straightforward and already matches the CPU path’s error message with minimal code. Since we still have to touch each entry, the performance difference should be negligible. If you feel strongly about switching, I can certainly revisit it.

malfet (changes requested):
This is indeed correct behavior, but extremely slow because it requires copy of the tensor to CPU and running checks on CPU as well

roei-shlezinger:
@malfet Just a friendly ping - the requested updates are in. Thanks!

malfet:
Closing this one, as I'm a about to propose one that reports errors asynchronously

Issue #25845, opened 2019-09-09
[cppdocs] the docs of class Tensor is missing


jimages:
## 📚 Documentation

After the 1.2 version of pytorch release, the torch::Tensor docs(original url:https://pytorch.org/cppdocs/api/classat_1_1_tensor.html) is missing. it does provide much helpful information about class torch::Tensor. 

It was there. (https://github.com/pytorch/cppdocs/blob/59620f68901b1e5f6bc2fa30c9ecbe8d0d86a1c4/api/classat_1_1_tensor.html) but in the latest commit, the file is absent.

It is really helpful for c++ development. Could you recovery it?

cc @ezyang @gchanan @zou3519 @yf225

soumith:
cc: @yf225

PR #165270, opened 2025-10-12
Update vLLM pinned commit to #25845

huydhn:
Testing, no need to review


huydhn:
Bisect to https://github.com/vllm-project/vllm/pull/24274#discussion_r2433424860, the new logic is executed when testing against PyTorch main

PR #165279, opened 2025-10-13
[torch/utils][Code Clean] Clean asserts in `torch/utils/_sympy`

KarhouTam:
Including: `torch/utils/_sympy/`

Fixes part of #164878

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @malfet @albanD

### DIFF
--- a/torch/utils/_sympy/functions.py
+++ b/torch/utils/_sympy/functions.py
@@ -510,8 +510,10 @@ class Mod(sympy.Function):
 
         # Evaluate if they are both literals.
         if q.is_Number and p.is_Number:
-            assert p >= 0, p
-            assert q >= 1, q
+            if not (p >= 0):
+                raise AssertionError(p)
+            if not (q >= 1):
+                raise AssertionError(q)
             return p % q
 
         # If q == 2, it's a matter of whether p is odd or even.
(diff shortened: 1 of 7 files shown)


torch/utils/_sympy/functions.py
```diff
@@ -510,8 +510,10 @@ def eval(cls, p, q):
 
         # Evaluate if they are both literals.
         if q.is_Number and p.is_Number:
-            assert p >= 0, p
-            assert q >= 1, q
+            if not (p >= 0):
```
albanD, line 513: Here and below, we could flip the condition and we should improve the error message

albanD (approved):
Small nit on one of the condition flip. But SGTM !

torch/utils/_sympy/functions.py
albanD: Scratch that, we can handle that in a follow up in the name of simplicity and easiness to review these PRs!

KarhouTam: Certainly, there's some oversight here. It will be addressed in the upcoming PRs. Thank you!

PR #164356, opened 2025-10-01
Add TORCH_TARGET_VERSION for stable ABI

mikaylagawarecki:
And update it so comparisons can be done by the preprocessor

**Note: We also need to gate in shim.h and figure out how to enforce this**

Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #165285
* #165284
* #165283
* #164332
* #163683
* #166373
* __->__ #164356

Differential Revision: [D85683549](https://our.internmc.facebook.com/intern/diff/D85683549)

PR #165284, opened 2025-10-13
End to end ABI versioning PoC (version 2)

mikaylagawarecki:
This is part 2 of the PoC started in https://github.com/pytorch/pytorch/pull/165283 

This PR (version 2) 
- Hardcodes TORCH_ABI_VERSION to 2.9.0 (simulating a fake state of the world in 2.9.0)
- Changes the memory layout of `Dummy` (prepending an `int8_t` foo)
       - Keeping the old version in a separate namespace
       - Updates StableIValue conversions such that based on `extension_abi_version` or `aoti_torch_abi_version`, `from` and `to` will appropriately translate the old struct into the new struct
- Adds a kwarg `scale` to ` _test_versioning(Tensor self, Dummy dummy, *, int scale=1) -> Tensor` (fills Tensor with Dummy.id * scale). (no default behavior change)
       - Registers an adapter for the schema for `aoti_torch_call_dispatcher`

## Testing
We extend the cpp_extension in this PR to add an another op that calls the C++ wrapper in `torch/csrc/stable/ops.h` and passes `scale` as 3, in addition to the existing op that does not pass scale we described in https://github.com/pytorch/pytorch/pull/165283

We build this with
- TARGETING 2.8
And run this with versions 1, 2 and 3. This verifies stableivalue conversions for `Dummy` are working.
`test_op_with_dummy`: In all 3 versions the tensor is filled with 42 as expected
`test_op_with_dummy_scale3`: errors when running with v1 due to `scale` being passed. (We could choose to make this a build time error but we don't).
- No target version (i.e. TORCH_ABI_VERSION as the target version)
And run with torch builds from versions 2 and 3. In all cases, the tests in this PR pass.

* #165285
* __->__ #165284
* #165283
* #164332
* #163683
* #166373
* #164356


aten/src/ATen/core/ATen_fwd.h
```diff
@@ -47,5 +47,10 @@ using c10::TensorOptions;
 
 // Forward declarations for dummy types used in operator signatures
 namespace dummy_types {
+namespace v2_8 {
 struct Dummy;
+}
+inline namespace v2_9 {
+struct Dummy;
+}
```
janeyx99, line 55: How come this distinction is necessary when the `struct Dummy` line is the same?

aten/src/ATen/core/ivalue.h
```diff
@@ -1027,18 +1027,19 @@ struct TORCH_API IValue final {
   at::Generator toGenerator() &&;
   at::Generator toGenerator() const&;
 
-  IValue(const dummy_types::Dummy& d) : tag(Tag::Dummy) {
+  IValue(const dummy_types::v2_9::Dummy& d) : tag(Tag::Dummy) {
```
janeyx99, line 1037: Technically could this stay as 
```suggestion
  IValue(const dummy_types::Dummy& d) : tag(Tag::Dummy) {
```
?

torch/csrc/stable/stableivalue_conversions.h
```diff
@@ -230,12 +235,21 @@ struct FromImpl<dummy_types::Dummy> {
...
       uint64_t extension_build_version,
       bool is_internal) {
-    (void)extension_build_version; // We don't actually use this for the legacy
-                                   // type
-    (void)is_internal; // Unused parameter
-    // Pack only the id into the higher 32 bits
-    uint64_t result = (static_cast<uint64_t>(val.id) & 0xFFFFFFFF) << 32;
-    return static_cast<StableIValue>(result);
+    uint64_t version_to_check =
+        is_internal ? extension_build_version : aoti_torch_abi_version();
```
janeyx99, line 232: nit: maybe a clearer name for `is_internal` is `called_from_libtorch`?

torch/headeronly/dummy.h
```diff
@@ -14,5 +15,26 @@ struct Dummy {
     return id;
   }
 };
+} // namespace v2_8
+
+inline namespace v2_9 {
```
janeyx99, line 20: Does it matter what we name these? e.g., if Dummy is the same for versions 2.9 through 2.13, does the namespace name stay 2_9 here? (i would guess it does) (is there anything that depends on the name?)

torch/headeronly/dummy.h
```diff
@@ -4,6 +4,7 @@
 
 namespace dummy_types {
 
+namespace v2_8 {
```
janeyx99, line 7: should we wrap every API we have today in `inline namespace v2_9` ?

torch/csrc/Module.cpp
```diff
@@ -2861,12 +2861,16 @@ Call this whenever a new thread is created in order to propagate values from
 #ifdef USE_KINETO
   torch::global_kineto_init();
 #endif
-  py::class_<dummy_types::Dummy>(py_module, "_Dummy")
+  py::class_<dummy_types::v2_9::Dummy>(py_module, "_Dummy")
```
janeyx99, line 2869: why do we need to specify 2_9 on the libtorch side when that's the default?

torch/csrc/stable/ops.h
```diff
@@ -236,16 +236,36 @@ inline torch::stable::Tensor clone(const torch::stable::Tensor& self) {
...
 
 inline torch::stable::Tensor op(
     const torch::stable::Tensor& self,
-    const dummy_types::Dummy& dummy) {
-  const auto num_args = 2;
-  std::array<StableIValue, num_args> stack{from(self), from(dummy)};
-  TORCH_ERROR_CODE_CHECK(aoti_torch_call_dispatcher_v2(
-      "aten::_test_versioning", "", stack.data(), TORCH_ABI_VERSION));
-  return to<torch::stable::Tensor>(stack[0]);
+    const dummy_types::Dummy& dummy,
```
janeyx99, line 284: If we can offer a build time error for someone building with 2.9 and running on an older version, we should do that instead of letting them hit the runtime error.

We can gate the kwarg signature with TORCH_FEATURE_VERSION right?

janeyx99 (commented):
I think a lot of my questions fall under the same umbrella of "how much of libtorch outside of headeronly/stable would we need to change to support versioning for a struct?"

aten/src/ATen/core/ivalue.h
mikaylagawarecki: I think I got errors that `dummy_types::Dummy` is ambiguous here

aten/src/ATen/core/ATen_fwd.h
mikaylagawarecki: You're right, it might not be necessary here specifically, but I recall getting errors elsewhere when I made the change that `dummy_types::Dummy` was ambiguous

torch/headeronly/dummy.h
mikaylagawarecki: The name can be anything indeed

torch/headeronly/dummy.h
mikaylagawarecki: Do you mean in torch/headeronly?

I don't think this is necessary, we only need the versioned namespaces when there are two or more versions of a class that we want to keep around. imo since these are headeronly it should always be ok to add the "old namespace" when a change is actually needed

torch/csrc/Module.cpp
mikaylagawarecki: I think the reason was similar to https://github.com/pytorch/pytorch/pull/165284#discussion_r2429825164

torch/csrc/stable/ops.h
mikaylagawarecki: What runtime error are you referring to? There shouldn't be one here

Even if they run in an older version, they will have the latest `dummy_types::Dummy` because an extension always gets the current state of torch/headeronly in the version it was built with **regardless of `TORCH_TARGET_VERSION`** (rmb we discussed why last week!) :)

torch/csrc/stable/stableivalue_conversions.h
mikaylagawarecki: Good point, will update in the scaffolding PR

aten/src/ATen/core/ivalue.h
janeyx99: Hmm then doesn't that show that the inline namespace isn't working as expected? I'd expect dummy_types::Dummy to automatically only refer to the one in the inlined namespace.

torch/csrc/stable/ops.h
janeyx99: Oops, commented on the wrong line--I meant to comment on the line below--the "scale" kwarg!

mikaylagawarecki: Hmm I debated about whether that should be a runtime or compile time error as well (in my doc on versioning I wrote compile time indeed)

Initially I wanted to go with compile time, but I landed on runtime since technically if a user targets an old it's nice to allow them to still be able to run with a version that has a new kwarg 🤔

PR #165298, opened 2025-10-13
Refine CUDA BackendStaticInitializer for allocator select

guangyey:
* #165304
* __->__ #165298
* #165291
* #165289
* #165288


c10/cuda/CUDACachingAllocator.cpp
```diff
@@ -4448,37 +4449,34 @@ struct BackendStaticInitializer {
...
-              return CudaMallocAsync::allocator();
-            if (kv[1] == "native")
-              return &Native::allocator;
+          ) {
+            return CudaMallocAsync::allocator();
           }
+          break;
+        } else {
+          // Skip the key and its value
+          i = tokenizer.skipKey(i);
```
albanD, line 4466: Isn't something like this missing here?
        // handle comma separators between config entries
        if (i + 1 < tokenizer.size()) {
          tokenizer.checkToken(++i, ",");
        }

If so, we should have a specific test to make sure we parse these properly

guangyey: You're right, I missed that. Added a test case to cover this scenario.

albanD (approved):

PR #165309, opened 2025-10-13
[2/N] Don't return values in void functions 

cyyever:
This PR fixes returning from void functions, a follow-up of #164809 .

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @EikanWang @jgong5 @wenzhe-nrv @sanchitintel @mingfeima @XiaobingSuper @ashokei @jingxu10 @jerryzh168 @aditew01


ezyang:
I don't think this is worth it

PR #165311, opened 2025-10-13
[torch/utils][Code Clean] Clean asserts in `hipify/`, `jit/`, `model_dump` and `tensorboard` of `torch/utils`

KarhouTam:
Including:
- `torch/utils/hipify/`
- `torch/utils/jit/`
- `torch/utils/model_dump/`
- `torch/utils/tensorboard/`

Fixes part of #164878

cc @EikanWang @jgong5 @wenzhe-nrv @sanchitintel @mingfeima @XiaobingSuper @ashokei @jingxu10 @jerryzh168 @malfet @albanD

### DIFF
--- a/torch/utils/tensorboard/_embedding.py
+++ b/torch/utils/tensorboard/_embedding.py
@@ -25,9 +25,10 @@ def make_tsv(metadata, save_path, metadata_header=None):
     if not metadata_header:
         metadata = [str(x) for x in metadata]
     else:
-        assert len(metadata_header) == len(
+        if len(metadata_header) != len(
             metadata[0]
-        ), "len of header must be equal to the number of columns in metadata"
+        ):
+            raise AssertionError("len of header must be equal to the number of columns in metadata")
         metadata = ["\t".join(str(e) for e in l) for l in [metadata_header] + metadata]
 
     metadata_bytes = tf.compat.as_bytes("\n".join(metadata) + "\n")
--- a/torch/utils/tensorboard/_utils.py
+++ b/torch/utils/tensorboard/_utils.py
@@ -76,10 +76,12 @@ def _prepare_video(V):
 
 def make_grid(I, ncols=8):
     # I: N1HW or N3HW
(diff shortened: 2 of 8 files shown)


torch/utils/tensorboard/_embedding.py
```diff
@@ -25,9 +25,10 @@ def make_tsv(metadata, save_path, metadata_header=None):
     if not metadata_header:
         metadata = [str(x) for x in metadata]
     else:
-        assert len(metadata_header) == len(
+        if len(metadata_header) != len(
```
can-gaa-hou, line 28: This assertion does give a proper message here; is it still necessary to change it?

torch/utils/tensorboard/_utils.py
```diff
@@ -76,10 +76,12 @@ def is_power2(num):
 
 def make_grid(I, ncols=8):
     # I: N1HW or N3HW
-    assert isinstance(I, np.ndarray), "plugin error, should pass numpy array here"
+    if not isinstance(I, np.ndarray):
```
can-gaa-hou, line 79: Same above. BTW, is it better to raise a TypeError rather than an AssertionError?

torch/utils/tensorboard/_embedding.py
KarhouTam: Yes. According to https://github.com/pytorch/pytorch/issues/164878#issuecomment-3381584729, all asserts will finally be converted and `AssertionError` is the one PyTorch core maintainer recommended for consistency.

torch/utils/tensorboard/_utils.py
KarhouTam: Ditto.

albanD (approved):
Looks good!

torch/utils/tensorboard/_utils.py
albanD: We can do this kind of change in a follow up.
The problem is that any code that catches this error (in CI or otherwise) will also need updating which will make these PR more challenging to review.
Splitting things makes it more tractable.

PR #165315, opened 2025-10-13
[Autograd] Add Default Autograd Fallback for PrivateUse1 in PyTorch

fffrog:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.12.0) (oldest at bottom):
* #165316
* __->__ #165315

Please refer to this [link](https://github.com/pytorch/pytorch/issues/163979) for more background.

- Allow register fallback for AutogradPrivateUse1 multiple.
- Add Autograd fallback implemetation for AutogradPrivateUse1

PyTorch can privide a common implementation for AutogradPrivateUse1, and the user can override it based on the need of specififc accelerator.

PR #165316, opened 2025-10-13
[OpenReg] Remove the Unnecessary Fallback Implementation for AutogradPrivate1

fffrog:
* __->__ #165316
* #165315

As the title stated.

The fallback for AutogradPrivateUse1 is builtin in PyTorch, so it is no need to register general implementation for out of tree backend.


ezyang (approved):
conditioned on the first PR being approved this would be fine

albanD (approved):

PR #165317, opened 2025-10-13
[Code Clean] Clean asserts in torch/ao/quantization/experimental/* and torch/ao/quantization/pt2e/*

zhudada0120:
Replace assert statements with explicit if/raise patterns in:
- torch/ao/quantization/experimental/* (11 errors)
- torch/ao/quantization/pt2e/* (68 errors)

fix partialy #164878

### DIFF
--- a/torch/ao/quantization/experimental/adaround_fake_quantize.py
+++ b/torch/ao/quantization/experimental/adaround_fake_quantize.py
@@ -42,7 +42,8 @@ class AdaroundFakeQuantizer(FakeQuantize):
         if quant_min is not None and quant_max is not None:
             if quant_min > quant_max:
                 raise AssertionError(
-                    "quant_min must be less than or equal to quant_max"
+                    "quant_min must be less than or equal to quant_max, "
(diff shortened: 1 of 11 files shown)


torch/ao/quantization/experimental/adaround_fake_quantize.py
```diff
@@ -40,13 +40,15 @@ def __init__(
         )
         # Populate quant_min/quant_max to observer_kwargs if valid
         if quant_min is not None and quant_max is not None:
-            assert quant_min <= quant_max, (
-                "quant_min must be less than or equal to quant_max"
-            )
+            if quant_min > quant_max:
+                raise AssertionError(
+                    "quant_min must be less than or equal to quant_max"
```
albanD: For these, it would be great to print the actual values.

albanD (approved):

torch/ao/quantization/experimental/adaround_fake_quantize.py
zhudada0120: I've updated the error message to include the actual values. PTAL, thanks.

PR #165098, opened 2025-10-09
bf16 support for fake_quantize_learnable_per_channel_affine

liangel-02:
Adding bf16 support for `torch._fake_quantize_learnable_per_channel_affine()` op by relaxing the type check on scale

TODO: need to add bf16 support to `per_tensor_affine_` as `torch._fake_quantize_learnable_per_tensor_affine_backward` gets called in the backward pass 

**Test**
Modified unit test in `test_workflow_ops.py`

PR #165325, opened 2025-10-13
bf16 support for per_channel bwd

liangel-02:
Follow up to #165098 - adding bf16 support for the backward pass. To avoid BC breaking changes/losing precision, we upcast the parameters to fp32 after the op gets called, and downcast the gradients to bf16 before returning.

For testing, we upcast to fp32 before calling the reference function. We increase the tolerance to 1e-2 for bf16 inputs because of a difference in casting calculations between python's `x.to(torch.bfloat16)` and cpp's `x.to(at::kBFloat16)` (after comparing intermediate tensors, we found that the numerics diverge after the final casting). We don't explicitly cast in the CPP op but rather let autograd/optimizer handle it.

* #165362
* __->__ #165325

### DIFF
--- a/test/quantization/core/test_workflow_ops.py
+++ b/test/quantization/core/test_workflow_ops.py
@@ -919,10 +919,6 @@ class TestFakeQuantizeOps(TestCase):
         for n_bits in (4, 8):
             quant_min, quant_max = 0, 2 ** n_bits - 1
 
-            X_base = X_base.to(dtype=torch.float32)
-            scale_base = scale_base.to(device, dtype=torch.float32)
-            zero_point_base = zero_point_base.to(device=device, dtype=torch.float32)
-
             X_curr = X_base.clone()
             X_curr.requires_grad_()
             scale_curr = scale_base.clone()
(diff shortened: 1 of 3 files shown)


test/quantization/core/test_workflow_ops.py
```diff
@@ -931,7 +943,9 @@ def _test_learnable_backward_per_channel(self, X_base, device, scale_base, zero_
                 dScale_actual = scale_curr.to(device).grad.detach()
                 dZeroPoint_expected = dZeroPoint.to(device).detach()
                 dZeroPoint_actual = zero_point_curr.to(device).grad.detach()
-                tolerance = 1e-4
+
+                # increasing tolerance for bf16 due to differences in python's x.to(torch.bfloat16) and cpp's x.to(at::kBFloat16)
```
andrewor14, line 948: can you put an actual numerical example where the rounding differs? Like X is rounded to 0.168 in python but 0.167 in cpp or something?

test/quantization/core/test_workflow_ops.py
```diff
@@ -152,6 +158,12 @@ def _fake_quantize_learnable_per_channel_affine_grad_reference(
 
         grad_scale[i] = grad_scale_i
         grad_zero_point[i] = grad_zp_i
+
+    if dtype is torch.bfloat16:
```
andrewor14, line 163: I feel this needs a short comment, maybe something like mimic autograd downcasting to bf16, even though this part does not exist in the actual cpp op

andrewor14 (approved):
Looks great, thanks!

PR #165336, opened 2025-10-13
[hop] run local_map with interpreter to preserve fx_traceback annotations

xmfan:
* __->__ #165336

We have an issue when using fx_traceback.annotate and HOPs that trace joint graphs. HOPs have bodies that have already been traced by Dynamo, and after Animesh's PR, does have the annotations. But when we lower that Dynamo HOP body to aten in either pre-dispatch or post-dispatch, we need to propagate the annotations to the aten nodes.

AOTAutograd does this indirectly by piggybacking off the `PropagateUnbackedSymInts` fx.Interpreter. I'm not sure if all HOPs should be using it to trace their joints or not. This PR adds an interpreter to local_map's implementation.

### DIFF
--- a/torch/_higher_order_ops/local_map.py
+++ b/torch/_higher_order_ops/local_map.py
@@ -257,9 +257,14 @@ def create_hop_fw_bw(
             primals = primals_and_tangents[:num_fw_inputs]
             tangents = primals_and_tangents[num_fw_inputs:]
 
-            def prepare_fw_with_masks(fn: Callable[..., Any]) -> Callable[..., Any]:
+            def prepare_fw_with_masks(
+                fw_gm: torch.fx.GraphModule,
+            ) -> Callable[..., Any]:
                 def fw_with_masks(*args: Any) -> tuple[tuple[Any], list[bool]]:
-                    fw_out = fn(*args)
+                    # The Interpreter here is required to propagate metadata
+                    # from the dynamo graph body to the local_map graph body.
+                    # This is required for fx_traceback.annotate for work.
+                    fw_out = torch.fx.Interpreter(fw_gm).run(*args)
                     assert isinstance(fw_out, tuple), (
                         "Dynamo traced submodule should return tuple"
                     )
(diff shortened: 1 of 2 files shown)


torch/_higher_order_ops/local_map.py
```diff
@@ -257,9 +257,14 @@ def joint_f(
...
-            def prepare_fw_with_masks(fn: Callable[..., Any]) -> Callable[..., Any]:
+            def prepare_fw_with_masks(
+                fw_gm: torch.fx.GraphModule,
+            ) -> Callable[..., Any]:
                 def fw_with_masks(*args: Any) -> tuple[tuple[Any], list[bool]]:
-                    fw_out = fn(*args)
+                    # The Interpreter here is required to propagate metadata
+                    # from the dynamo graph body to the local_map graph body.
+                    # This is required for fx_traceback.annotate for work.
+                    fw_out = torch.fx.Interpreter(fw_gm).run(*args)
```
xmfan, line 267: @anijain2305 this was the only place i think we need for most HOPs

torch/_higher_order_ops/local_map.py
```diff
@@ -437,7 +447,8 @@ def autograd_key(
             fw_gm, bw_gm, num_fw_ins, num_fw_outs, filtered_grads_idx, *args, **kwargs
         )
 
-    return fw_gm(*args, **kwargs)
+    # TODO: get rid of this when we can install as a subgraph
+    return torch.fx.Interpreter(fw_gm).run(*args, **kwargs)
```
xmfan, line 451: so because fw_gm is not mounted as a submodule of the larger graph (issues w/ pre-dispatch and some autograd apis), when we run an interpreter on the larger graph, all the nodes in `fw_gm` would be ran and associated with the metadata of the `call_local_hop(fw_gm, ...)` node. For proper metadata propagation, we need a nested interpreter that will actually set the correct node meta before running each node of `fw_gm`.

torch/_higher_order_ops/local_map.py
yushangdi, line 267: would be good to add a test example that test the custom node meta on this HOP.

xmfan: there's more in the AP repo

yushangdi (approved):

PR #165337, opened 2025-10-13
[ROCm] moved atomics to LDS for grid sampler bilinear interpolation

glen-amd:
I'm closing this PR for now (maybe re-open if I could get new ideas) and about to file a separate one (https://github.com/pytorch/pytorch/pull/166825). Reason:

- After @jerrymannil wrote a script (below is the simplified one) and found the accuracy issue, I got aware of a critical logic error in the core premise of the optimization.
  ```python
  from typing import Tuple
  import torch


  def create_gpu_tensors() -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
      # Input tensor [112, 128, 60, 80] - needs gradients
      input_tensor = torch.randn(112, 128, 60, 80, device=torch.device("cuda"), requires_grad=True)

      # Grid tensor [112, 1, 25000, 2] - needs gradients
      # Grid values should be in [-1, 1] range for valid sampling
      grid = torch.rand(112, 1, 25000, 2, device=torch.device("cuda"), requires_grad=False) * 2 - 1

      # Target tensor for loss computation [112, 128, 1, 25000]
      target = torch.randn(112, 128, 1, 25000, device=torch.device("cuda"))

      # Make tensors contiguous to match expected strides
      input_tensor = input_tensor.contiguous()
      grid = grid.contiguous()
      target = target.contiguous()

      return input_tensor, grid, target


  def verify_grid_sampler_accuracy():
      input_tensor, grid, target = create_gpu_tensors()

      # Forward pass - grid sampling
      # This uses bilinear interpolation with zero padding
      sampled = F.grid_sample(
          input_tensor,
          grid,
          mode="bilinear",  # interpolation_mode = 0
          padding_mode="zeros",  # padding_mode = 0
          align_corners=False,  # align_corners = False
      )
      # Loss computation
      loss = F.mse_loss(sampled, target)
      # Backward pass - this will call aten::grid_sampler_2d_backward
      loss.backward()

      input_tensor_cpu = input_tensor.clone().detach().cpu().requires_grad_(True)
      grid_cpu = grid.clone().detach().cpu()
      target_cpu = target.clone().detach().cpu()

      sampled_cpu = F.grid_sample(
          input_tensor_cpu,
          grid_cpu,
          mode="bilinear",  # interpolation_mode = 0
          padding_mode="zeros",  # padding_mode = 0
          align_corners=False,  # align_corners = False
      )
      loss_cpu = F.mse_loss(sampled_cpu, target_cpu)
      loss_cpu.backward()

      assert torch.allclose(input_tensor.grad.cpu(), input_tensor_cpu.grad)


  if __name__ == "__main__":
    verify_grid_sampler_accuracy()
  ```
- Root cause leading to the accuracy issue: **mixing coordinate systems**
  - The numerical error lies in the mapping between global input coordinates and the LDS tile.
    - The kernel is launched by tiling the output space, but it incorrectly tries to map to a LDS tile that represents a local input space.
    - The direct faulty code in the optimized kernel function `grid_sampler_2d_backward_kernel_optimized`:
      ```cpp
      // Inside the `accumulate` lambda:
      // `iy_g` and `ix_g` are global input coords.
      // `tile_h_out` and `tile_w_out` are global output coords.
      // We are subtracting an output coord from an input one.
      // These two coordinate systems are unrelated in Grid Sampler.
      // The resulting `s_iy` and `s_ix` indices are meaningless, eventually causing gradients to be written to the wrong LDS locations or discarded entirely.
      const index_t s_iy = iy_g - tile_h_out + HALO;
      const index_t s_ix = ix_g - tile_w_out + HALO;
      ```
  - Accordingly, the write-back loop has the same flaw in reverse, reconstructing a garbage input coord from an output base.
    ```cpp
    // Inside the write-back loop:
    index_t h_in = tile_h_out - HALO + i;
    index_t w_in = tile_w_out - HALO + j;
    ```
- Fundamental flaw: Grid Sampler has **no locality** at all.
  - The optimization in the PR is made based on a **false assumption**.
    - It is assumed that a local 16x16 tile of output pixels would only scatter gradients to a local region of the input tensor.
    - This local input region is represented by the LDS tile (`smem_raw`).
  - The reallity is that the assumption is only true for up sample, but false to grid sample.
    - As shown in the test script, the `grid` tensor can be completely random (`grid = torch.rand(112, 1, 25000, 2, device=torch.device("cuda"), requires_grad=False) * 2 - 1`).
  - Because the output tile scatters gradients to arbitrary, non-local input locations, a local LDS tile makes no sense.

---

N.B.: The optimized kernel in this PR is only for `GridSamplerInterpolation::Bilinear` and `GridSamplerPadding::Zeros` for now.<br>
# Analysis and dissection of possible root causes
- Primarily similar to the [`upsample_bilinear2d_backward` issue]()
# Corresponding implementation
- Files
  - /aten/src/ATen/native/native_functions.yaml
  - /aten/src/ATen/native/GridSampler.h
  - /aten/src/ATen/native/cuda/GridSampler.h
  - /aten/src/ATen/native/cuda/GridSampler.cuh
  - **/aten/src/ATen/native/cuda/GridSampler.cu**
- Code snippets
  - https://github.com/pytorch/pytorch/blob/12d2ef557f6e127100267c31a31572d8ab5cc788/aten/src/ATen/native/cuda/GridSampler.cu#L370-L400
  - https://github.com/pytorch/pytorch/blob/12d2ef557f6e127100267c31a31572d8ab5cc788/aten/src/ATen/native/cuda/GridSampler.cuh#L247-L261
- `grid_sampler_2d_backward_cuda_kernel` defined in the file "/aten/src/ATen/native/cuda/GridSampler.cu"
  - Two distinct backward computations
    - `grad_input` (primary bottleneck)
      - Similar to `upsample_bilinear2d_backward`, employing a scatter-add pattern that can potentially cause high contention on global memory atomics.
        - The sampling locations for "grid_sampler" are *arbitrary* and data-dependent (from the `grid` tensor), leading to even more irregular and unpredictable memory access patterns than the regular grid of "upsample".
      - `safe_add_2d` defined in the file "/aten/src/ATen/native/cuda/GridSampler.cuh"
        - `fastAtomicAdd`
      - Proposed fix: Using shared memory (LDS) privatization to mitigate global atomic contention for `grad_input`
      - Partial feasibility of alternative fix: Employing a gather-sum approach like `upsample_bilinear2d_backward`
        - It's ONLY partially feasible because the grid region is arbitray and irregular, compared to the deterministic, small grid region in `upsample_bilinear2d_backward`, as shown below:
          ```
            if (align_corners) {
              lo = static_cast<accscalar_t>(input_pos - 1) / scale;
              hi = static_cast<accscalar_t>(input_pos + 1) / scale;
            } else {
              lo = (input_pos - static_cast<accscalar_t>(0.5)) / scale - static_cast<accscalar_t>(0.5);
              hi = (input_pos + static_cast<accscalar_t>(1.5)) / scale - static_cast<accscalar_t>(0.5);
            }
            min_output = max(0, static_cast<int>(ceil(lo)));
            max_output = min(output_size - 1, static_cast<int>(floor(hi)));
          ```
    - `grad_grid` (another bottleneck)
      - Employing a gather-read pattern with 2D spatial locality on reads from the `input` tensor
      - Proposed fix: Leveraging texture objects to accelerate the read-intensive path for `grad_grid`.
  - Other potention optimizations
    - Runtime function parameters to compile time template parameters
# Solution to mitigation the global atomic contention: shared memory privatization
- Thread-block private tile in the LDS for accumulating `grad_input`, by performing `atomicAdd` (`fastAtomicAdd`) on the fast, on-chip shared memory.
- Global writes (done by `fastAtomicAdd`) are reduced by a factor of `4` to reduce global memory atomic contention.
# Grid Sampler vs. Up Sampling
- Up Sampling
  - The relationship between an output pixel and an input pixel is a fixed mathematical formula (e.g., `input_coord = output_coord * scale_factor`) that is a structured, affine transformation.
  - Since we can write an equation for it, we can also invert it to figure out which output pixels depend on a given input pixel.
    - This is right what allows the `upsample_2d_backward` optimization to work by gathering contributions.
- Grid Sampler
  - The relationship is defined by the values inside the `grid` (output) tensor.
    - The `grid` tensor is essentially a lookup table that has no invertible formula, compared to the Up Sampling case.
    - We cannot efficiently determine which output pixels are influenced by a specific input pixel.
  - For each output pixel location `(h, w)`, the value `grid[n, h, w]` contains an `(x, y)` coordinate telling the kernel where to read from the input image.
  - The `(x, y)` coordinates in the grid can be anything, which has no guaranteed pattern or locality.


jeffdaily (approved):
Approved, pending CI.

aten/src/ATen/native/cuda/GridSampler.cu
```diff
@@ -513,6 +515,317 @@ namespace {
...
+            // If opmath_t could be `double`, casting to `float` may lose precision.
+            auto atomic_add_smem = [] __device__ (opmath_t* addr, opmath_t v) {
+              if constexpr (std::is_same<opmath_t, float>::value) {
+                atomicAdd(reinterpret_cast<float*>(addr), static_cast<float>(v));
+              } else if constexpr (std::is_same<opmath_t, double>::value) {
+                atomicAdd(reinterpret_cast<double*>(addr), static_cast<double>(v));
+              } else {
+                static_assert(!std::is_same<opmath_t, opmath_t>::value, "Unsupported opmath_t for LDS atomicAdd");
+              }
+            }
```
jerrymannil: I don't think this builds. We need a `;` here to finish lambda definition

glen-amd: Good catch.
We do need more good catches.

aten/src/ATen/native/cuda/GridSampler.cu
```diff
@@ -513,6 +515,317 @@ namespace {
...
+              }
+            };
+            accumulate(iy_nw, ix_nw, nw);
+            accumulate(iy_nw, ix_nw + 1, ne);
+            accumulate(iy_nw + 1, ix_nw, sw);
+            accumulate(iy_nw + 1, ix_nw + 1, se);
+          }
+
+          // Accumulate grad_grid contributions in registers.
+          scalar_t v_nw = get_value_bounded(inp_ptr_NC, static_cast<opmath_t>(ix_nw), static_cast<opmath_t>(iy_nw), inp_W, inp_H, input.strides[3], input.strides[2], PADDING_MODE, ALIGN_CORNERS);
```
jerrymannil: * i dont think we need to cast to `opmath_t`
* get_value_bounded<scalar_t>

These fixes are needed for CI to build successfully and run test.
Can you add CI tags and re-run the tests after fixing these ?

glen-amd: You mean automatic upcast is good?

glen-amd: How about half-precision cases?
I know we have type conflicts during type deduction and am working on a fix.

jerrymannil: > yes. Doing oppmath_t cast was giving me compilation issues
Same with not providing <scalar_t>

You can also check the CI logs. There are some build failures

glen-amd: > * i dont think we need to cast to `opmath_t`
> * get_value_bounded<scalar_t>
> 
> These fixes are needed for CI to build successfully and run test. Can you add CI tags and re-run the tests after fixing these ?

Only `get_value_bounded<scalar_t>` doesn't suffice to cover all cases - needing to explicitly cast the 2nd and 3rd args, and then casting the the result.
Or maybe we can just fall back to how the original kernel does?

pruthvistony (approved):

PR #165339, opened 2025-10-13
Change the variable source impl into the repr in Lazy

fxdawnn:
Addressed comment on moving items into the ``__repr__``. The test plan is inherited from the previous PR for variable tracker within the symbolic convert. 

### DIFF
--- a/torch/_dynamo/variables/lazy.py
+++ b/torch/_dynamo/variables/lazy.py
@@ -104,10 +104,6 @@ class LazyVariableTracker(VariableTracker):
             self._cache.name_hint = name
 
     def __str__(self) -> str:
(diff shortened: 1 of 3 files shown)


torch/_dynamo/variables/lazy.py
```diff
@@ -104,9 +104,19 @@ def set_name_hint(self, name: str) -> None:
             self._cache.name_hint = name
 
     def __str__(self) -> str:
+        # if self.is_realized():
```
Lucaskabela: No longer used so we can remove this

fxdawnn: Sounds good! Removed!

PR #165342, opened 2025-10-13
[trymerge] Do not check for rules when reverting

clee2000:
Why do we need to check for merge rules when reverting?


malfet (approved):
SGTM

PR #165344, opened 2025-10-13
[CI] [AOTI] Windows cross compilation CI test

yushangdi:
Add a to `trunk` workflow for AOTI windows cross-compilation test. 

Overall, the workflow looks like:

- "win-vs2022-cuda12_8-py3-build": 
        - build torch on windows
        - new: upload some `.lib` artifacts: `cuda.lib, cudart.lib`
        
- "cross-compile-linux-test": 
         - download the  cuda .lib artifacts, also unzip to get the torch .lib artifacts from the windows .whl
         - use `linux.g6.4xlarge.experimental.nvidia.gpu` which has L4, it should match the CUDA arch with the windows GPU runner that we use to load the compiled artifact later.
         - test the cross-compilation
 
We change the `pytorch-linux-jammy-cuda12.8-cudnn9-py3-gcc11` docker image to install `mingw`

We also need a test that load the compiled artifacts on windows and run it. This will be added in the next PR. 

On the AOTI side, this PR adds support for the posix flavor of mingw. If posix flavor is used, it will add the `-Wl,-Bstatic -lwinpthread -Wl,-Bdynamic` compilation flags. 

It also add the `aoti_shim_path` to build path when link_torch=True.

A successful run example (compilation part only, not the loading test) on commit [0e70ffb](https://github.com/pytorch/pytorch/commit/0e70ffb4ad423023ce6565d91be2e17257c615fc):
https://github.com/pytorch/pytorch/actions/runs/18507489766/job/52743592865?pr=165344

Example running this test on devserver:
```
# devvm
sudo dnf install mingw64-gcc-c++
x86_64-w64-mingw32-gcc --version


# CUDA toolkit
 wget https://developer.download.nvidia.com/compute/cuda/12.6.0/local_installers/cuda_12.6.0_560.76_windows.exe
sudo dnf install p7zip p7zip-plugins
 7z x cuda_12.6.0_560.76_windows.exe -o./cuda-windows

export WINDOWS_CUDA_HOME=/data/users/shangdiy/cuda-windows/cuda_cudart/cudart/

mkdir torch_win_libs
# obtained from windows pytorch/build/lib folder
manifold get pytorch/tree/shangdiy/lib/torch.lib
manifold get pytorch/tree/shangdiy/lib/torch_cuda.lib
manifold get pytorch/tree/shangdiy/lib/torch_cpu.lib
manifold get pytorch/tree/shangdiy/lib/c10.lib
manifold get pytorch/tree/shangdiy/lib/c10_cuda.lib

python test/inductor/test_aoti_cross_compile_windows.py --win-torch-lib-dir /data/users/shangdiy/torch_win_libs

```


malfet (changes requested):
Can we just have periodic/once in a bluemoon job that uploads those `.lib` files to some S3 buckets, and just installs them (together with mingw as part of docker container)?

yushangdi:
> Can we just have periodic/once in a bluemoon job that uploads those `.lib` files to some S3 buckets, and just installs them (together with mingw as part of docker container)?

but then the test would be broken if some one adds a new shim, but didn't trigger a .lib upload job in the CI?

.github/workflows/aoti-cross-compile-windows.yml
```diff
@@ -0,0 +1,77 @@
+name: aoti-cross-compile-windows
+
+on:
+  push:
+    tags:
+      - ciflow/inductor/*
```
desertfire: Let's use inductor-periodic.

yushangdi:
update: discussed with @malfet offline. I will change to use libs from torch-2.10.0a0%2Bgitdbe337a-cp310-cp310-win_amd64.whl  and change the docker config to install mingw inside the docker.

desertfire (commented):
The AOTI changes look good to me. I will leave it to @malfet to approve the CI part of changes.

desertfire:
BTW, I don't think c10 libs are needed. Testing in https://github.com/pytorch/pytorch/pull/165489.

.github/workflows/_win-build.yml
```diff
@@ -50,6 +50,13 @@ on:
         default: "windows.4xlarge.nonephemeral"
         description: |
           Label of the runner this job should run on.
+      upload-win-cuda-libs-for-cross-compile:
```
malfet, line 53: Why wheel is not sufficient for you? It contains the same binaries already

.github/workflows/_win-build.yml
```diff
@@ -168,6 +175,42 @@ jobs:
         run: |
           .ci/pytorch/win-build.sh
 
+      # Collect Windows torch libs and CUDA libs for cross-compilation
```
malfet, line 178: This should not be needed, they are already part of the wheel, aren't they?

test/run_test.py
```diff
@@ -394,6 +394,8 @@ def __contains__(self, item):
 ]
 
 INDUCTOR_TESTS = [test for test in TESTS if test.startswith(INDUCTOR_TEST_PREFIX)]
+# Cross-compilation tests require special setup and should only run in their dedicated workflow
+WIN_CROSS_COMPILE_TESTS = ["inductor/test_aoti_cross_compile_windows"]
```
malfet, line 398: I would have kept this logic as part of `test_aoti_XYZ` file

torch/_inductor/cpp_builder.py
```diff
@@ -333,9 +335,9 @@ def get_msvc_cl_path() -> tuple[bool, str]:
 
 
 @functools.cache
-def check_mingw_win32_flavor(compiler: str) -> None:
+def check_mingw_win32_flavor(compiler: str) -> str:
```
malfet, line 338: I like the change, but feels like it could have been moved to separate PR (as this PR title is just `add test`)

malfet (changes requested):
Looks very good, thank you for all the changes. If you don't mind, let's land changes to `cpp_builder` (one that extracts mingw flavor) as part of separate PR, and if you want one can land changes to docker image separately as well (as it painfully slow to rebuild docker image all the time)

IMO two areas needs slight improvements:
 - We should not modify _wind-build.yml at all, as artifcat it produces (wheel) should be sufficient, we can extract necessary files during test.sh preparation
 - test_aoti_cross_compile_win feels a bit too long, some of the steps outlined there probably should be part of cpp_builder.py (though I could be wrong, only quickly looked into it)

.github/workflows/_win-build.yml
yushangdi: I don't think it contains cuda.lib and cudart.lib? those are from the windows cuda toolkit.

.github/workflows/_win-build.yml
yushangdi: It does not collect any torch libs anymore, but we still need it to collect the CUDA libs (cuda.lib and cudart.lib). Those are not in the wheel

test/run_test.py
yushangdi: sorry can you elaborate a little more? I don't quite understand. The point of adding this line  here is to prevent all other linux jobs from running this test (as the test will fail without proper mingw setup etc.)

.github/workflows/_win-build.yml
desertfire: I think cuda.lib and cudart.lib can be extracted on the Windows side from the standard CUDA Windows installation package.

torch/_inductor/cpp_builder.py
yushangdi: sure, created a separate PR: https://github.com/pytorch/pytorch/pull/165557

yushangdi:
also, the mingw docker part is separated to https://github.com/pytorch/pytorch/pull/165560

PR #165347, opened 2025-10-13
[Inductor][CuTeDSL] Move load_template up two directories

NikhilAPatel:
Summary: Moves the function used to load CuTeDSL Jinja templates up one level out of the flex attention folder. This way it can be used for more generate Inductor templates in the future.

Test Plan: `INDUCTOR_TEST_DISABLE_FRESH_CACHE=1 TORCHINDUCTOR_CACHE_DIR=~/cutetest buck2 run mode/opt //caffe2/test/inductor:flex_flash -c fbcode.nvcc_arch=b200a -c fbcode.enable_gpu_sections=true -c fbcode.platform010_cuda_version=12.8`

Differential Revision: D84527470

### DIFF
--- a/torch/_inductor/kernel/flex/flex_attention.py
+++ b/torch/_inductor/kernel/flex/flex_attention.py
@@ -29,7 +29,7 @@ from .common import (
     freeze_irnodes,
     get_fwd_subgraph_outputs,
     infer_dense_strides,
-    load_template,
+    load_flex_template,
     maybe_realize,
     set_head_dim_values,
     SubgraphResults,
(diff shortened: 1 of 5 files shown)


torch/_inductor/kernel/mm_common.py
```diff
@@ -192,3 +193,13 @@ def is_batch_stride_largest_or_zero(mat1, mat2, layout) -> bool:
             return False
 
     return True
+
+
+_TEMPLATE_DIR = Path(__file__).parent / "templates"
```
drisspg: nit, can you put this in a `common.py` or a utils.py that is not related to mm

torch/_inductor/kernel/flex/flex_attention.py
```diff
@@ -50,6 +51,8 @@
 aten = torch.ops.aten
 Expr = sympy.Expr
 
+_TEMPLATE_DIR = Path(__file__).parent / "templates"
```
drisspg: super nit, use  partial to set the template dir and call it load_flex_template

drisspg (approved):

PR #165348, opened 2025-10-13
Replace insert with std::rotate_copy for RingBuffer

lakshayg:


eqy (approved):

c10/cuda/CUDACachingAllocator.cpp
```diff
@@ -1080,19 +1080,12 @@ class RingBuffer {
...
-    result.reserve(alloc_trace->size());
-    result.insert(
-        result.end(),
-        alloc_trace->begin() +
-            static_cast<typename std::vector<T>::difference_type>(
-                alloc_trace_next),
-        alloc_trace->end());
-    result.insert(
-        result.end(),
+    result.reserve(result.size() + alloc_trace->size());
```
eqy, line 1083: Was the previous reserve call insufficient?

lakshayg: I don't think it was insufficient based on how it was being used, however, the way this function is designed, it is possible to use the function in a way where it would be insufficient.

The function appends elements to a vector. We need to reserve space _in addition_ to what the vector already has. This call was just setting the total capacity which would be insufficient if the vector was already populated.

Skylion007 (approved):

PR #165355, opened 2025-10-13
Nan to num complex args

### LINKED ISSUE #131283: `input` argument of `nan_to_num()` works with `complex` type but `nan`, `posinf` and `neginf` argument don't work with `complex` type
### 🐛 Describe the bug

`input` argument of [nan_to_num()](https://pytorch.org/docs/stable/generated/torch.nan_to_num.html) works with `complex` type but `nan`, `posinf` and `neginf` argument don't work with `complex` type as shown below:

```python
import torch

my_tensor = torch.tensor([complex('nan'), complex('inf-infj'), complex('-inf+infj'), 4.+0.j])
                                    # ↓       # ↓       # ↓
torch.nan_to_num(input=my_tensor, nan=1, posinf=2, neginf=3)
# tensor([1.+0.j, 2.+3.j, 3.+2.j, 4.+0.j])

                                    # ↓↓       # ↓↓       # ↓↓
torch.nan_to_num(input=my_tensor, nan=1., posinf=2., neginf=3.)
# tensor([1.+0.j, 2.+3.j, 3.+2.j, 4.+0.j])

                                    # ↓↓↓↓       # ↓↓↓↓↓       # ↓↓↓↓
torch.nan_to_num(input=my_tensor, nan=True, posinf=False, neginf=True)
# tensor([1.+0.j, 0.+1.j, 1.+0.j, 4.+0.j])

                                    # ↓↓↓↓↓↓       # ↓↓↓↓↓↓       # ↓↓↓↓↓↓
torch.nan_to_num(input=my_tensor, nan=1.+0.j, posinf=2.+0.j, neginf=3.+0.j) # Error
```

> TypeError: nan_to_num(): argument 'nan' must be float, not complex

> TypeError: nan_to_num(): argument 'posinf' must be float, not complex

> TypeError: nan_to_num(): argument 'neginf' must be float, not complex

In addition, `input` argument of `nan_to_num()` works with `int`, `float` and `bool` type as shown below:

```python
import torch

my_tensor = torch.tensor([1, 2, 3, 4])

torch.nan_to_num(input=my_tensor)
# tensor([1, 2, 3, 4])

my_tensor = torch.tensor([1., 2., 3., 4.])

torch.nan_to_num(input=my_tensor)
# tensor([1., 2., 3., 4.])

my_tensor = torch.tensor([True, False, True, False])

torch.nan_to_num(input=my_tensor)
# tensor([True, False, True, False])
```

### Versions

```python
import torch

torch.__version__ # 2.3.1+cu121
```

cleonard530:
Fixes #131283

Updated nan_to_num to allow for arguments nan, posinf, and neginf to be complex when the tensor is complex. Currently works for torch.complex128 and torch.complex64 dtypes. To allow for torch.complex32 dtype, casting from complex double to complex Half would need to be implemented first.  

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @aditew01 @malfet

### DIFF
--- a/aten/src/ATen/native/UnaryOps.cpp
+++ b/aten/src/ATen/native/UnaryOps.cpp
@@ -819,45 +819,8 @@ Tensor& nan_to_num_out(const Tensor& self,
     return result;
   }
 
-  // Check if any of the scalar parameters are complex
-  bool has_complex_scalar = false;
-  if (nan.has_value() && nan.value().isComplex()) {
-    has_complex_scalar = true;
-  } else if (pos_inf.has_value() && pos_inf.value().isComplex()) {
-    has_complex_scalar = true;
-  } else if (neg_inf.has_value() && neg_inf.value().isComplex()) {
-    has_complex_scalar = true;
-  }
-
-  if (has_complex_scalar) {
-    // Convert Scalar to complex<double>
-    std::optional<c10::complex<double>> complex_nan = nan.has_value()
-        ? std::make_optional(nan.value().toComplexDouble())
-        : std::nullopt;
-    std::optional<c10::complex<double>> complex_pos_inf = pos_inf.has_value()
-        ? std::make_optional(pos_inf.value().toComplexDouble())
-        : std::nullopt;
-    std::optional<c10::complex<double>> complex_neg_inf = neg_inf.has_value()
-        ? std::make_optional(neg_inf.value().toComplexDouble())
-        : std::nullopt;
-
-    auto iter = TensorIterator::unary_op(result, self);
-    nan_to_num_complex_stub(iter.device_type(), iter, complex_nan, complex_pos_inf, complex_neg_inf);
-  } else {
-    // Convert Scalar to double
-    std::optional<double> double_nan = nan.has_value()
-        ? std::make_optional(nan.value().toDouble())
-        : std::nullopt;
-    std::optional<double> double_pos_inf = pos_inf.has_value()
-        ? std::make_optional(pos_inf.value().toDouble())
-        : std::nullopt;
-    std::optional<double> double_neg_inf = neg_inf.has_value()
-        ? std::make_optional(neg_inf.value().toDouble())
-        : std::nullopt;
-
-    auto iter = TensorIterator::unary_op(result, self);
-    nan_to_num_stub(iter.device_type(), iter, double_nan, double_pos_inf, double_neg_inf);
-  }
+  auto iter = TensorIterator::unary_op(result, self);
+  nan_to_num_stub(iter.device_type(), iter, nan, pos_inf, neg_inf);
   return result;
 }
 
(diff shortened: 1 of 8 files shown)


aten/src/ATen/native/cpu/UnaryOpsKernel.cpp
```diff
@@ -515,6 +553,30 @@ static void nan_to_num_kernel(
...
+          : scalar_t(std::numeric_limits<value_t>::max(), 0.0);
+      scalar_t neg_inf_replacement = neg_inf.has_value()
+          ? static_cast<scalar_t>(neg_inf.value())
+          : scalar_t(std::numeric_limits<value_t>::lowest(), 0.0);
+      cpu_kernel(iter, [=](scalar_t a) -> scalar_t {
+        return _nan_to_num_replace(a, nan_replacement, pos_inf_replacement, neg_inf_replacement);
+      });
+    });
+  } else {
+    TORCH_CHECK(false, "nan_to_num does not work with complex nan, pos_inf, or neg_inf and non-complex tensors. Expected complex tensor, but got ", iter.dtype());
```
amjames: This if/else is redundant AT_DISPATCH_COMPLEX_TYPES will already produce an error if the scalar type is not handled by any of the generated cases.

aten/src/ATen/native/UnaryOps.cpp
```diff
@@ -1016,6 +1053,7 @@ DEFINE_DISPATCH(special_ndtri_stub); // NOLINT(cppcoreguidelines-avoid-non-const
 DEFINE_DISPATCH(special_log_ndtr_stub); // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
 DEFINE_DISPATCH(neg_stub); // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
 DEFINE_DISPATCH(nan_to_num_stub); // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
+DEFINE_DISPATCH(nan_to_num_complex_stub); // NOLINT(cppcoreguidelines-avoid-non-const-global-variables)
```
amjames: Is this used anywhere?

test/test_unary_ufuncs.py
```diff
@@ -507,6 +507,32 @@ def test_nan_to_num(self, device, dtype):
             torch.nan_to_num(x, out=out, nan=nan, posinf=posinf, neginf=neginf)
             self.assertEqual(result, out)
 
+    @dtypes(torch.complex128, torch.complex64)
+    def test_nan_to_num_complex_args(self, device, dtype):
```
amjames: It should be possible to fold these into the test case above? It already covers  complex dtypes we just need to add complex replacement values to the mix.

aten/src/ATen/native/UnaryOps.cpp
```diff
@@ -819,25 +819,62 @@ Tensor& nan_to_num_out(const Tensor& self,
...
-  nan_to_num_stub(iter.device_type(), iter, nan, pos_inf, neg_inf);
+  // Check if any of the scalar parameters are complex
+  bool has_complex_scalar = false;
+  if (nan.has_value() && nan.value().isComplex()) {
+    has_complex_scalar = true;
+  } else if (pos_inf.has_value() && pos_inf.value().isComplex()) {
+    has_complex_scalar = true;
+  } else if (neg_inf.has_value() && neg_inf.value().isComplex()) {
+    has_complex_scalar = true;
+  }
```
amjames:
```suggestion
  const bool has_complex_scalar = (nan.has_value() && nan.value().isComplex()) ||  (pos_inf.has_value() && pos_inf.value().isComplex()) || (neg_inf.has_value() && neg_inf.value().isComplex());

  
```

cleonard530:
@isuruf, I updated how **nan_to_num** is dispatched (closer to how it was before my original updates) and I think it should pass the xpu build that was failing before. However, I do not have any way to test that locally can will only be able to tell from a new CI run.

amjames (commented):
This kind of opens up some questions about how we define NAN/Inf for complex types. The existing restrictions of having only real replacement vals allowed us to skirt that by effectively composing with view_as_real for complex dtyped tensors, at least on CPU. 

I think two big questions to answer 1) can we refactor this so that the complex vs real replacement value handling does not require a split in logic at such a high level. and 2) does this implementation naturally decay to the same behavior if a real replacement value is passed as a complex type (i.e imaginary part is zero) or similar for the tensor arg?

PR #165359, opened 2025-10-13
[Bugfix][Precompile][vLLM] Support for pickling einops for aot_autograd serialization in vLLM

Lucaskabela:
Fixes issue with compiling `Qwen2_5_vl` in https://github.com/vllm-project/vllm/pull/23207 (issue happens with `aot_autograd_cache`)

### DIFF
--- a/torch/fx/_graph_pickler.py
+++ b/torch/fx/_graph_pickler.py
@@ -4,6 +4,7 @@ import io
 import pickle
 from abc import abstractmethod
 from typing import Any, Callable, NewType, Optional, TypeVar, Union
+from typing_extensions import override, Self
 
 import torch
 import torch.utils._pytree as pytree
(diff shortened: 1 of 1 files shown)


jamesjwu (approved):
qq: what's the difference in behavior between this and _OpBuiltinPickleData? I assume OpBuiltinPickleData also has to import a module and access its attribute, can we not just add it to OpBuiltin?

torch/fx/_graph_pickler.py
```diff
@@ -432,6 +432,9 @@ def pickle(cls, op: object, options: Options) -> "_OpPickleData":
         elif name.startswith("operator."):
             _, detail = name.split(".", 1)
             return _OpOperatorPickleData(detail)
+        elif name.startswith("einops."):
+            _, detail = name.split(".", 1)
+            return _OpEinopsPickleData(detail)
```
zou3519: @jamesjwu wasn't there an API that allowed registering things to cacheable out of tree? Would that work here, or is this something else?

jamesjwu: Here I think they are pickling the fx graph itself (so to get a proper cache key), so the AOTAutograd API won't handle it properly.

jamesjwu (approved):

PR #165362, opened 2025-10-13
bf16 support for per tensor backward

liangel-02:
Adding bf16 for the backward pass of `torch._fake_quantize_learnable_per_tensor_affine()`. 

Note that for testing, we modified the seed to avoid increasing tolerance due to cases where difference in Python vs CPP downcasting causes tensor mismatches. (e.g. 27.87704 vs  27.8408 before downcasting, 27.7500 vs 27.8750 after downcasting for Python vs CPP op)

* __->__ #165362


test/quantization/core/test_workflow_ops.py
```diff
@@ -948,7 +964,7 @@ def _test_learnable_backward_per_channel(self, X_base, device, scale_base, zero_
                 # increasing tolerance for bf16 due to differences in python's x.to(torch.bfloat16) and cpp's x.to(at::kBFloat16)
```
andrewor14, line 964: this comment is outdated now? Or maybe we should rephrase it to be like, note: if we pick a different seed we may need to increase tolerance due to differences in...

andrewor14 (approved):

aten/src/ATen/native/quantized/FakeQuantPerTensorAffine.cpp
```diff
@@ -184,44 +184,52 @@ std::tuple<Tensor, Tensor, Tensor> _fake_quantize_learnable_per_tensor_affine_ba
           0 & \text{ else }
         \end{cases}
   */
-  float scale_val = scale[0].item<float>();
+
+  bool is_bfloat16 = (X.scalar_type() == at::kBFloat16);
```
Skylion007, line 188: So we shouldn't cast fp16

aten/src/ATen/native/quantized/FakeQuantPerTensorAffine.cpp
```diff
@@ -184,44 +184,52 @@ std::tuple<Tensor, Tensor, Tensor> _fake_quantize_learnable_per_tensor_affine_ba
...
-    X.device().type(), iter, scale_val, inv_scale_val, zero_point_val, quant_min, quant_max, grad_factor);
+    X_.device().type(), iter, scale_val, inv_scale_val, zero_point_val, quant_min, quant_max, grad_factor);
 
   // The total sums over the scale and zero point gradient vectors are what will be returned in the end.
-  auto dScale = dScale_vec.sum().unsqueeze(0).to(scale.device());
-  auto dZeroPoint = dZeroPoint_vec.sum().unsqueeze(0).to(zero_point.device());
+  auto dScale = dScale_vec.sum().unsqueeze(0).to(scale_.device());
+  auto dZeroPoint = dZeroPoint_vec.sum().unsqueeze(0).to(zero_point_.device());
 
   return std::make_tuple(dX, dScale, dZeroPoint);
```
Skylion007, line 234: All these should be std::move into make tuple btw

test/quantization/core/test_workflow_ops.py
liangel-02: not outdated, reverted tolerance change

aten/src/ATen/native/quantized/FakeQuantPerTensorAffine.cpp
liangel-02: we enabled bf16 support for per_tensor alongside per_channel in this pr #165325, so if we want to enable fp16 we can do it in a separate pr for both these ops?

aten/src/ATen/native/quantized/FakeQuantPerTensorAffine.cpp
liangel-02: should we make this change in a follow up PR since this PR doesn't actually touch this code?

PR #165370, opened 2025-10-13
[inductor] generate fused rms/layer norm bwd

shunting314:
* #166053
* __->__ #165370
* #166204

RMS/Layer norm backward would generated 2 kind of reductions:
- the reduction computing dx which reduce across the hidden dimension (in the context of transformer)
- the reduction computing dw/db which reduce across the BxT (batch size , sequence length) dimension.

These 2 set of reductions have common input buffers but inductor can not fuse them because of different loop orders.

There are multiple sources of custom kernels that implement fused version of such kernel (Liger-Kernel, quack, Paul Zhang's internal post). This PR enable Inductor to generate such kernels automatically.

The generated kernel is very similar to https://github.com/linkedin/Liger-Kernel/blob/33924d20b6fa5dfc9391e4630ae82cc810114a30/src/liger_kernel/ops/rms_norm.py#L114 .

To make the implementation simple and performing, we enable such fusion only if the inner reduction (computing dx) is a persistent reduction. This should be true for representative inputs. Persistent reduction is critical for the perf here to make sure a loaded tensor does not need to be reload.

To make sure the inner reduction (computing dx) and outer reductions (computing dw/db) being fusible, the PR does the following:
1. convert the outer reductions to pointwise by replacing 'reduction' & 'store_reduction' node with a new type of node 'parital_accumulate'. The new node will collect the reduction type, buffer name, input of reduction etc, which is essential for proper codegening.
2. do loop reordering (rely on the earlier loop ordering after fusion work) to reorder the loops of the converted pointwise so it can be fused with the inner reduction
3. there can be epilogues that need to be added in the end. E.g. the outer reduction may be followed by a division for mean , or followed by a down cast if dw/db is in low precision (fp16/bf16).

Some early benchmarking on H100 shows about 2X speedup for both RMSNorm and LayerNorm backward for shape (1152 * 500, 384 ) used in some internal model. Note that, I manually disable split reduction in this benchmarking since otherwise the fusion will be skipped right now. The next PR will make the mix-order-reduction compose better with split reduction


torch/_inductor/codegen/simd.py
```diff
@@ -1489,27 +1514,148 @@ def requires_closing_previous_reduction(node, node_schedule):
...
+        )
+
+        # split epilogue out of node2
+        node2_reductions, node2_epilogue = self._split_mix_order_reduction_epilogue(
+            node2
+        )
+
+        # decide the split size
+        nrow, ncol = node1.group[1]
+        split_size = 64  # TODO need add heuristics
```
PaulZhang12, line 1557: Is it possible to autotune over this?

torch/_inductor/codegen/triton.py
```diff
@@ -121,6 +122,15 @@
 async_compile = AsyncCompile()
 
 
+def get_triton_reductoin_function(reduction_type):
```
PaulZhang12: nit: spelling

torch/_inductor/codegen/simd.py
shunting314: yes, this is critical to perf. The current status is, if we do split reduction, the split size is already decided by how we pick 'num-split'. 

A few steps I'm trying
1. make the split size independent to num-splits decided by split reduction . The next PR is a first step to make sure we can still fuse mix order reductions even when the outer reduction is split
2. I'll first check if I can come up with a good enough heuristics. If there is no heuristics that works well for all important shapes, then do autotuning before codegening the kernel. (Similar implementation as how we benchmark fusion)

torch/_inductor/codegen/simd.py
```diff
@@ -1489,27 +1514,148 @@ def requires_closing_previous_reduction(node, node_schedule):
...
+            end = f"({idx} + 1) * {stride_str}"
+            reduction_type2op = {
+                "min": "amin",
+                "max": "amax",
+            }
+            opname = reduction_type2op.get(
+                partial_accum.reduction_type, partial_accum.reduction_type
+            )
+            V.graph.wrapper_code.writeline(
+                f"{buffer_name} = workspace_0[{start} : {end}].view({nsplit}, {rnumel}).{opname}(dim=0)",
```
shunting314: This causes issue when the graph has multiple workspace allocations. Fixing is coming

torch/_inductor/codegen/common.py
```diff
@@ -1573,7 +1581,6 @@ def workspace(
         - A new argument "ws_ptr" will be present in the generated code.
 
         Args:
-            nbytes (sympy.Expr): The number of bytes to allocate.
```
jansel: ?

jansel (approved):

torch/_inductor/config.py
```diff
@@ -1530,6 +1530,10 @@ class triton:
     # If set to false, will never generate PDL code.
     enable_pdl = False
 
+    mix_order_reduction = (
+        os.environ.get("TORCHINDUCTOR_MIX_ORDER_REDUCTION", "0") == "1"
```
shunting314, line 1534: Off by default in this PR. Will reenable in following PRs

PR #165376, opened 2025-10-13
[DebugMode] refactor logs into _DebugCalls

pianpwk:
Refactors `DebugMode.operators` to be more structured `_DebugCall` objects, instead of (op, args, kwargs, call_depth) tuples. Useful going forward for attaching more information (e.g. output info, call metadata).

Is BC-breaking, but attaches an `__iter__` method for `_OpCall` and `_RedistributeCall` so previous tuple usage is accessible.

* #165499
* #165462
* #165497
* #165377
* #165461
* #165498
* __->__ #165376

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @ezyang @gchanan


SherlockNoMad (dismissed):

torch/utils/_debug_mode.py
```diff
@@ -138,7 +171,7 @@ def __torch_function__(self, func, types, args=(), kwargs=None):
         if kwargs is None:
             kwargs = {}
 
-        self.operators.append((func, args, kwargs, self.call_depth))
+        self.operators.append(_OpCall(func, args, kwargs, self.call_depth))
```
malfet, line 194: This is BC breaking, isn't it? Though I agree, that `torch.utils._debug_mode` seems to be internal only method, so 3rd party users rely on it, so it worth at least documenting, that type of operators is changing, and perhaps provide fallback to `_OpCall` to make it's attributes accessible as tuple

And last but not least: unittest?

yushangdi (approved):

PR #165379, opened 2025-10-13
[CUDA][cuBLAS] Only `xFail` `addmm` with reduced precision reductions on non-RTX skus

eqy:
RTX Blackwells don't behave quite like their datacenter counterparts here

cc @ptrblck @msaroufim @jerryzh168 @mruberry @csarofeen @xwang233


Skylion007 (approved):
Huh, can we raise the tolerance just for datacenter GPUs?

PR #165390, opened 2025-10-14
[user-events] Add inductor event types

mlazos:
This PR adds support for Symbolic events in the TorchInductor scheduler. In short, we create an EventFactory which provides event indices which monotonically increase as compile proceeds and enables reuse of events that are no longer used. It also adds codegen support for events. 

* #174224
* #174223
* #165505
* #165391
* #165504
* __->__ #165390


torch/_inductor/event.py
```diff
@@ -0,0 +1,226 @@
...
+
+    def __eq__(self, rhs: object) -> bool:
+        """Whether the current event is identical to the rhs event."""
+        if not isinstance(rhs, CudaEventSym):
+            return NotImplemented
+        return (
+            self.idx == rhs.idx
+            and self.originate_stream_idx == rhs.originate_stream_idx
+            and self.factory is rhs.factory
+        )
```
eellison: any reason to not make them singletons per idx ?

torch/_inductor/event.py
```diff
@@ -0,0 +1,226 @@
...
+        Returns:
+            An internal data structure that depicts stream <-> event dependency.
+
+        Note:
+            This method doesn't necessarily generate a event waiting in the final program. Instead
+            it records the dependence between the stream and the current event and also increase
+            the reference count of this event. If an event object has called this method, it is
+            guaranteed to be generated in the final program.
+        """
+        assert stream_idx != self.originate_stream_idx
```
eellison: if we're going to have this assertion, do we have any machinery to avoid no-op waits ? there's nothing illegal about waiting on record on same stream it just might not do anything

torch/_inductor/config.py
```diff
@@ -2268,6 +2268,9 @@ class rocm:
 # Backend to use for CPU codegen either "cpp" or "triton" (experimental) or "halide" (experimental) or "pallas" (experimental)
 cpu_backend: Literal["cpp", "triton", "halide", "pallas"] = "cpp"
 
+# Whether to reuse CUDA events across different stream synchronizations
+reuse_cuda_event: bool = True
```
eellison: Do you have any sort of benchmarks of the overhead of creating a cuda event ? Is this a premature optimization?

torch/_inductor/event.py
```diff
@@ -0,0 +1,226 @@
...
+
+    Note:
+        In most cases this class should not be used standalone. Use
+        `CUDAEventFactory.get_sym_event()` to instantiate one.
+    """
+
+    factory: CudaEventFactory
+    idx: int
+    originate_stream_idx: int
+    ref_count: int = 0
```
eellison: why do we need ref count ? is this always going to be statically knowable, e.g. with control flow?

torch/_inductor/event.py
```diff
@@ -0,0 +1,226 @@
...
+    This factory maintains internal states to ensure that created cuda events get monotonically
+    increasing indices as compilation goes along. It also maintains a pool of materialized cuda
+    events that symbolic events can reuse.
+    """
+
+    def __init__(self) -> None:
+        """Initialize a event factory."""
+        self.symbolic_event_idx: itertools.count = itertools.count(start=1)
+        self.materialized_event_idx: itertools.count = itertools.count(start=1)
+        self.available_materialized_events: OrderedSet[str] = OrderedSet()
```
eellison: any reason to not just make these the actual event objects, and not strings ?

torch/_inductor/event.py
```diff
@@ -0,0 +1,226 @@
...
+        if self._reuse_cuda_event and self.available_materialized_events:
+            return self.available_materialized_events.pop()
+        else:
+            event = EVENT_NAME_TEMPLATE.format(
+                event_idx=next(self.materialized_event_idx)
+            )
+            code.writeline(f"{event} = torch.cuda.Event()")
+            return event
+
+    def deposit_materialized_event(self, event: str) -> None:
```
eellison: Would it make sense to give `event` a different type than string, with a default repr?

torch/_inductor/event.py
```diff
@@ -0,0 +1,226 @@
...
+        self._reuse_cuda_event: bool = config.reuse_cuda_event
+
+    def get_entrance_event(self) -> CudaEventSym:
+        """Return the cuda event that corresponding to compute graph entering."""
+        if self._entrance_event is None:
+            self._entrance_event = CudaEventSym(
+                factory=self,
+                idx=0,
+                originate_stream_idx=DEFAULT_STREAM_IDX,
+            )
```
eellison, line 112: Is this always true, what if this is created on a side stream ?

eellison (approved):
a lil hard to review without tests but i understand we're breaking it up..

torch/_inductor/event.py
mlazos: I'll remove

torch/_inductor/event.py
mlazos: I'll remove, fwiw, I think we would have to have some constraints on both branches of the control flow, but since we are just doing what the user wants, I don't need this. If we don't reuse events, ref count just cleans up unused events, and I don't want to clean up events that the user put there even if they are dead.

torch/_inductor/event.py
mlazos: I might be able to remove the entrance event stuff - the purpose of it is to allow any work queued on the default stream before the compiled region to complete before beginning any side stream work, it's compile-inserted. We're leaving this to the user in the non-automatic scenario. I will leave this for now and remove later if not needed once I have the end to end.

torch/_inductor/config.py
mlazos: Yeah I removed these are really lightweight

PR #165401, opened 2025-10-14
Turn some const variables into constexpr in C++ code

cyyever:
This PR checks the C++ code and turns some const variables into constexpr.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @fritzo @neerajprad @alicanb @nikitaved @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168

### DIFF
--- a/aten/src/ATen/native/Distributions.h
+++ b/aten/src/ATen/native/Distributions.h
@@ -127,7 +127,7 @@ C10_DEVICE scalar_t sample_gamma(scalar_t alpha, BaseSampler<accscalar_t, unifor
 
 template<typename scalar_t>
 C10_DEVICE scalar_t stirling_approx_tail(scalar_t k) {
-  const static scalar_t kTailValues[] = {
+  constexpr static scalar_t kTailValues[] = {
     0.0810614667953272,
     0.0413406959554092,
     0.0276779256849983,
@@ -139,7 +139,7 @@ C10_DEVICE scalar_t stirling_approx_tail(scalar_t k) {
     0.00925546218271273,
     0.00833056343336287
   };
-  if (k <= 9) {
+  if (k <= sizeof(kTailValues)/sizeof(scalar_t)) {
     return kTailValues[static_cast<size_t>(k)];
   }
   scalar_t kp1sq = (k + 1) * (k + 1);
--- a/aten/src/ATen/native/Distributions.h
+++ b/aten/src/ATen/native/Distributions.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <array>
 #include <ATen/native/Math.h>
 #include <c10/macros/Macros.h>
 #include <c10/util/MathConstants.h>
@@ -139,7 +140,7 @@ C10_DEVICE scalar_t stirling_approx_tail(scalar_t k) {
     0.00925546218271273,
     0.00833056343336287
   };
-  if (k < sizeof(kTailValues)/sizeof(scalar_t)) {
+  if (k < std::size(kTailValues)) {
     return kTailValues[static_cast<size_t>(k)];
   }
   scalar_t kp1sq = (k + 1) * (k + 1);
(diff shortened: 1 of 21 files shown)


aten/src/ATen/native/cuda/IGammaKernel.cu
```diff
@@ -197,7 +197,7 @@ __host__ __device__ scalar_t _igamc_helper_series(scalar_t a, scalar_t x) {
   accscalar_t sum = 0;
   accscalar_t term, logx;
   static const int MAXITER = 2000;
```
Skylion007: Why isn't this constexpr?

torch/csrc/distributed/c10d/ProcessGroupUCC.cpp
```diff
@@ -118,7 +118,7 @@ std::unordered_map<std::string, std::string> torch_ucc_envs_map = {
 };
 
 std::vector<OpType> parse_blocking_wait(std::string op_list_string) {
-  const static std::unordered_map<std::string, OpType> str2op = {
+  constexpr static std::unordered_map<std::string, OpType> str2op = {
```
Skylion007: Wait? Std::string supports constexpr before C++20?

Skylion007: Pretty sure this constexpr is a noop.

aten/src/ATen/native/quantized/cpu/qlinear.cpp
```diff
@@ -1388,7 +1388,7 @@ namespace at::native {
     TORCH_CHECK(act_scale.numel() == 1 && act_zero_point.numel() <= 1,
         "onednn int8 linear: act scale/zp size should be 1/<=1");
     static std::optional<at::Tensor> other = std::nullopt;
-    static const std::string_view binary_post_op = "none";
+    constexpr std::string_view binary_post_op = "none";
```
Skylion007, line 1391: Same.
Not sure if std::string_view supports constexpr construction before C++20

cyyever: I have checked cppreference, std::string_view supports constexpr since C++17.

torch/csrc/distributed/c10d/ProcessGroupUCC.cpp
cyyever: Yes, let's revert it

aten/src/ATen/native/cuda/IGammaKernel.cu
cyyever: Fixed

Skylion007 (approved):

yushangdi:
looks like this PR broke `test/distributions/test_distributions.py::TestDistributions::test_binomial_sample` test? Looking at https://hud.pytorch.org/hud/pytorch/pytorch/main/1?per_page=50, before this PR, the test is only failing with doctest, but after this PR, the distributions test start failing, maybe consider revert or forward fix?

cyyever:
@yushangdi @seemethere There is an indexing error in modifications of `aten/src/ATen/native/Distributions.h`, I have fixed it.

aten/src/ATen/native/Distributions.h
```diff
@@ -139,7 +139,7 @@ C10_DEVICE scalar_t stirling_approx_tail(scalar_t k) {
     0.00925546218271273,
     0.00833056343336287
   };
-  if (k <= 9) {
+  if (k < sizeof(kTailValues)/sizeof(scalar_t)) {
```
Skylion007: Since we are on C++17, consider std::size which should be constexpr: https://en.cppreference.com/w/cpp/iterator/size.html

cyyever: Good, let's try it.

PR #165410, opened 2025-10-14
[torch/utils][Code Clean] Clean asserts in `torch/utils/*.py`

KarhouTam:
Including:
- `torch/utils/*.py`

Fixes part of #164878

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @malfet @albanD

### DIFF
--- a/torch/utils/_appending_byte_serializer.py
+++ b/torch/utils/_appending_byte_serializer.py
@@ -125,7 +125,8 @@ class AppendingByteSerializer(Generic[T]):
         reader = BytesReader(data)
         if reader.read_uint64() != _ENCODING_VERSION:
             raise AssertionError(
-                "Encoding version mismatch in AppendingByteSerializer.to_list"
+                f"Encoding version mismatch in AppendingByteSerializer.to_list, \
+                    got {reader.read_uint64()}"
             )
 
         result: list[T] = []
--- a/torch/utils/_appending_byte_serializer.py
+++ b/torch/utils/_appending_byte_serializer.py
@@ -38,7 +38,8 @@ class BytesWriter:
         digest = zlib.crc32(self._data[CHECKSUM_DIGEST_SIZE:]).to_bytes(
             4, byteorder="big", signed=False
         )
-        assert len(digest) == CHECKSUM_DIGEST_SIZE
+        if len(digest) != CHECKSUM_DIGEST_SIZE:
+            raise AssertionError("Computed checksum digest has unexpected size")
         self._data[0:CHECKSUM_DIGEST_SIZE] = digest
         return bytes(self._data)
 
@@ -46,11 +47,13 @@ class BytesWriter:
 class BytesReader:
     def __init__(self, data: bytes) -> None:
         # Check for data corruption
-        assert len(data) >= CHECKSUM_DIGEST_SIZE
+        if len(data) < CHECKSUM_DIGEST_SIZE:
+            raise AssertionError("Input data is too short to contain checksum")
         digest = zlib.crc32(data[CHECKSUM_DIGEST_SIZE:]).to_bytes(
             4, byteorder="big", signed=False
         )
-        assert len(digest) == CHECKSUM_DIGEST_SIZE
+        if len(digest) != CHECKSUM_DIGEST_SIZE:
+            raise AssertionError("Computed checksum digest has unexpected size")
         if data[0:CHECKSUM_DIGEST_SIZE] != digest:
             raise RuntimeError(
                 "Bytes object is corrupted, checksum does not match. "
@@ -120,7 +123,10 @@ class AppendingByteSerializer(Generic[T]):
     @staticmethod
     def to_list(data: bytes, *, deserialize_fn: Callable[[BytesReader], T]) -> list[T]:
         reader = BytesReader(data)
-        assert reader.read_uint64() == _ENCODING_VERSION
+        if reader.read_uint64() != _ENCODING_VERSION:
+            raise AssertionError(
+                "Encoding version mismatch in AppendingByteSerializer.to_list"
+            )
 
         result: list[T] = []
         while not reader.is_finished():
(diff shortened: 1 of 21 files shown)


torch/utils/_appending_byte_serializer.py
```diff
@@ -120,7 +123,10 @@ def to_bytes(self) -> bytes:
     @staticmethod
     def to_list(data: bytes, *, deserialize_fn: Callable[[BytesReader], T]) -> list[T]:
         reader = BytesReader(data)
-        assert reader.read_uint64() == _ENCODING_VERSION
+        if reader.read_uint64() != _ENCODING_VERSION:
+            raise AssertionError(
+                "Encoding version mismatch in AppendingByteSerializer.to_list"
```
albanD: Can we print the versions?

torch/utils/_config_module.py
```diff
@@ -85,12 +85,15 @@ def __post_init__(self) -> None:
...
-                and self.env_name_force is None
-            ), "if alias is set, none of {default, justknob and env var} can be set"
+            if (
+                self.default is not _UNSET_SENTINEL
+                or self.justknob is not None
+                or self.env_name_default is not None
+                or self.env_name_force is not None
+            ):
+                raise AssertionError(
+                    "if alias is set, none of {default, justknob and env var} can be set"
```
albanD: missing f ?

torch/utils/_config_module.py
```diff
@@ -100,7 +103,8 @@ def string_or_list_of_string_to_list(
             return None
         if isinstance(val, str):
             return [val]
-        assert isinstance(val, list)
+        if not isinstance(val, list):
+            raise TypeError("val is not a list")
```
albanD: print val type?

torch/utils/_python_dispatch.py
```diff
@@ -88,7 +88,8 @@ class TorchDispatchMode:
 
     def __init__(self, _dispatch_key=None):
         if _dispatch_key is not None:
-            assert isinstance(_dispatch_key, torch._C.DispatchKey)
+            if not isinstance(_dispatch_key, torch._C.DispatchKey):
+                raise AssertionError("_dispatch_key must be a torch._C.DispatchKey")
```
albanD, line 87: print type

torch/utils/_python_dispatch.py
```diff
@@ -217,16 +218,22 @@ def _get_current_dispatch_mode() -> Optional[TorchDispatchMode]:
...
 
 def _detect_infra_mode(key):
-    assert key in [
+    if key not in (
         torch._C._TorchDispatchModeKey.FUNCTIONAL,
         torch._C._TorchDispatchModeKey.PROXY,
-    ]
+    ):
+        raise AssertionError(
+            "key must be either FUNCTIONAL or PROXY _TorchDispatchModeKey"
```
albanD: print key

torch/utils/_python_dispatch.py
```diff
@@ -538,9 +553,12 @@ def _correct_storage_aliasing(func, schema_info, args, outs):
     It does this by unsafely overwriting the storage field of the output tensor
     to be the same storage as the input.
     """
-    assert isinstance(func, torch._ops.OpOverload)
-    assert isinstance(args, tuple)
-    assert isinstance(outs, (list, tuple))
+    if not isinstance(func, torch._ops.OpOverload):
```
albanD, line 553: we can print the types here as well

torch/utils/flop_counter.py
```diff
@@ -62,7 +62,8 @@ def mm_flop(a_shape, b_shape, *args, out_shape=None, **kwargs) -> int:
     # Inputs contains the shapes of two matrices.
     m, k = a_shape
     k2, n = b_shape
-    assert k == k2
+    if k != k2:
```
albanD, line 65: We can print values for the errors in this file as well

albanD (commented):
A few small things, but looks pretty good overall!

torch/utils/_config_module.py
KarhouTam: No. The message is as is. If an error is raised here, both the if-branch and the message are shown and user can know, oh one of 
```python
self.default is not _UNSET_SENTINEL
                or self.justknob is not None
                or self.env_name_default is not None
                or self.env_name_force is not None
```
is not fulfilled. BTW, most of them are None when error raised.

KarhouTam:
@albanD Thanks for your reviewing! All mentioned nits are fixed. PTAL again.

torch/utils/_config_module.py
albanD: Ho I read too fast, my bad!

albanD (dismissed):
Looks great!

cyyever (approved):

torch/utils/throughput_benchmark.py
```diff
@@ -5,7 +5,8 @@
 
 def format_time(time_us=None, time_ms=None, time_s=None):
     """Define time formatting."""
-    assert sum([time_us is not None, time_ms is not None, time_s is not None]) == 1
+    if time_us is not None or time_ms is not None or time_s is not None:
```
Yuxingwang-intel: Hi, I would like to make sure that the condition is not inverted, as the current logic raises an error when any argument is not None instead of only when all arguments are None.

KarhouTam: Oh, my bad. Thanks for the review. Will fix it right now.

KarhouTam: Changed to
```
if sum([time_us is not None, time_ms is not None, time_s is not None]) != 1:
        raise AssertionError("Expected only one of time_us, time_ms, time_s is given.")
```
Again, thanks for your correction.

KarhouTam:
> @pytorchbot revert -m "sorry I'm going to revert this since I want to try to back out some other things that are conflicting with this, there is nothing wrong with this PR, rebasing and resolving the merge conflicts should be enough, sorry for the churn" -c weird

Hi, @albanD . Sorry for the bother. 
This PR is reverted due to PR conflict and now is solved and ready to be merged again.

albanD (approved):
Sounds good!

PR #165430, opened 2025-10-14
[dynamo] Clean up assert in dynamo [1/N]

can-gaa-hou:
Fixes some part of #162852 and #164878. These two issues have some relationship though.

* __->__ #165430

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela @williamwen42 @albanD

### DIFF
--- a/torch/_dynamo/variables/builtin.py
+++ b/torch/_dynamo/variables/builtin.py
@@ -1930,11 +1930,23 @@ class BuiltinVariable(VariableTracker):
     def call_custom_dict_fromkeys(
         tx: "InstructionTranslator", user_cls, *args, **kwargs
     ):
-        assert user_cls in {dict, OrderedDict, defaultdict}
+        if user_cls not in {dict, OrderedDict, defaultdict}:
+            msg = ConstantVariable.create(
+                f"{user_cls.__name__}.fromkeys() is not supported"
+            )
+            raise_observed_exception(TypeError, tx, args=[msg])
         if kwargs:
             # Only `OrderedDict.fromkeys` accepts `value` passed by keyword
-            assert user_cls is OrderedDict
-            assert len(args) == 1 and len(kwargs) == 1 and "value" in kwargs
+            if user_cls is not OrderedDict:
+                msg = ConstantVariable.create(
+                    f"{user_cls.__name__}.fromkeys() takes no keyword arguments"
+                )
+                raise_observed_exception(TypeError, tx, args=[msg])
+            if len(args) != 1 or len(kwargs) != 1 or "value" not in kwargs:
+                msg = ConstantVariable.create(
+                    f"{user_cls.__name__}.fromkeys() takes no keyword arguments"
+                )
+                raise_observed_exception(TypeError, tx, args=[msg])
             args = (*args, kwargs.pop("value"))
         if len(args) == 0:
             msg = ConstantVariable.create(
@@ -1943,7 +1955,11 @@ class BuiltinVariable(VariableTracker):
             raise_observed_exception(TypeError, tx, args=[msg])
         if len(args) == 1:
             args = (*args, ConstantVariable.create(None))
-        assert len(args) == 2
+        if len(args) != 2:
+            msg = ConstantVariable.create(
+                f"fromkeys expected at most 2 arguments, got {len(args)}"
+            )
+            raise_observed_exception(TypeError, tx, args=[msg])
         arg, value = args
         DictVariableType = (
             ConstDictVariable if user_cls is not defaultdict else DefaultDictVariable
@@ -2039,7 +2055,11 @@ class BuiltinVariable(VariableTracker):
 
     def call_zip(self, tx: "InstructionTranslator", *args, **kwargs):
         if kwargs:
-            assert len(kwargs) == 1 and "strict" in kwargs
+            if len(kwargs) != 1 or "strict" not in kwargs:
+                msg = ConstantVariable.create(
+                    f"zip() should only have 'strict' keyword argument, but ({len(kwargs)} given)"
+                )
+                raise_observed_exception(TypeError, tx, args=[msg])
         strict = kwargs.pop("strict", False)
         args = [BuiltinVariable(iter).call_function(tx, [arg], {}) for arg in args]
         return variables.ZipVariable(
--- a/torch/_dynamo/variables/builtin.py
+++ b/torch/_dynamo/variables/builtin.py
@@ -1930,19 +1930,15 @@ class BuiltinVariable(VariableTracker):
     def call_custom_dict_fromkeys(
         tx: "InstructionTranslator", user_cls, *args, **kwargs
     ):
-        if user_cls not in {dict, OrderedDict, defaultdict}:
-            msg = ConstantVariable.create(
-                f"{user_cls.__name__}.fromkeys() is not supported"
-            )
-            raise_observed_exception(TypeError, tx, args=[msg])
+        assert user_cls in {dict, OrderedDict, defaultdict}
         if kwargs:
             # Only `OrderedDict.fromkeys` accepts `value` passed by keyword
-            if user_cls is not OrderedDict:
-                msg = ConstantVariable.create(
-                    f"{user_cls.__name__}.fromkeys() takes no keyword arguments"
-                )
-                raise_observed_exception(TypeError, tx, args=[msg])
-            if len(args) != 1 or len(kwargs) != 1 or "value" not in kwargs:
+            if (
+                user_cls is not OrderedDict
+                or len(args) != 1
+                or len(kwargs) != 1
+                or "value" not in kwargs
+            ):
                 msg = ConstantVariable.create(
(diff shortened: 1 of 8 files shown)


torch/_dynamo/variables/builtin.py
```diff
@@ -1930,11 +1930,23 @@ def call_custom_dict(tx: "InstructionTranslator", user_cls, *args, **kwargs):
...
+        if user_cls not in {dict, OrderedDict, defaultdict}:
+            msg = ConstantVariable.create(
+                f"{user_cls.__name__}.fromkeys() is not supported"
+            )
+            raise_observed_exception(TypeError, tx, args=[msg])
         if kwargs:
             # Only `OrderedDict.fromkeys` accepts `value` passed by keyword
-            assert user_cls is OrderedDict
-            assert len(args) == 1 and len(kwargs) == 1 and "value" in kwargs
+            if user_cls is not OrderedDict:
```
Lucaskabela: Can we combine these two bool messages since they are the same?

torch/_dynamo/variables/builtin.py
```diff
@@ -2039,7 +2055,11 @@ def call_frozenset(self, tx: "InstructionTranslator", *args, **kwargs):
 
     def call_zip(self, tx: "InstructionTranslator", *args, **kwargs):
         if kwargs:
-            assert len(kwargs) == 1 and "strict" in kwargs
+            if len(kwargs) != 1 or "strict" not in kwargs:
```
Lucaskabela: For readability sake for this particular boolean argument, I personally prefer the non-demorgan version:

```suggestion
            if not (len(kwargs) == 1 and "strict" in kwargs):
```

I just personally find it easier to read

torch/_dynamo/variables/constant.py
```diff
@@ -149,7 +149,9 @@ def call_method(
                 tx, [self, *args], kwargs
             )
         elif name == "join" and istype(self.value, str):
-            assert len(args) == 1 and len(kwargs) == 0
+            if len(args) != 1 or len(kwargs) != 0:
```
Lucaskabela:
```suggestion
            if not (len(args) == 1 and len(kwargs) == 0):
```
Similar thought here

torch/_dynamo/variables/dicts.py
```diff
@@ -506,7 +506,11 @@ def call_method(
                 raise_unhashable(args[0])
 
             self.install_dict_keys_match_guard()
-            assert not kwargs and len(args) == 2
+            if kwargs or len(args) != 2:
```
Lucaskabela, line 511: This one seems good like this (since one argument is not and other is truthy)

torch/_dynamo/variables/lists.py
```diff
@@ -1115,11 +1117,20 @@ def call_method(
         kwargs: dict[str, "VariableTracker"],
     ) -> "VariableTracker":
         if name == "__getitem__":
-            assert not kwargs and len(args) == 1
+            if len(args) != 1:
```
Lucaskabela: Just curious why did we split this one into two checks?

albanD (commented):
@anijain2305 can you confirm these are the right way to update these?

torch/_dynamo/variables/misc.py
```diff
@@ -1220,7 +1220,11 @@ def call_function(
         if is_tensor_base_attr_getter(self.method_wrapper) and isinstance(
             args[0], variables.TensorVariable
         ):
-            assert len(args) == 1 and len(kwargs) == 0
+            if len(args) != 1 or len(kwargs) != 0:
```
Lucaskabela: nit: Please change this one for readability (imo)

torch/_dynamo/variables/lists.py
Lucaskabela: I would prefer keeping this one check for consistency (also what is done on L1200 in this PR in `variables/torch.py`)

Lucaskabela (changes requested):
Some minor nits but generally this looks close to ready

For ease of review, can you also add an example of running a program that triggers one of these exceptions and the error before/after in the summary? That can serve as the testplan here - thanks!

torch/_dynamo/variables/builtin.py
```diff
@@ -1930,11 +1930,23 @@ def call_custom_dict(tx: "InstructionTranslator", user_cls, *args, **kwargs):
     def call_custom_dict_fromkeys(
         tx: "InstructionTranslator", user_cls, *args, **kwargs
     ):
-        assert user_cls in {dict, OrderedDict, defaultdict}
+        if user_cls not in {dict, OrderedDict, defaultdict}:
```
williamwen42, line 1938: This should be a graph break (`unimplemented_v2`) or remain an assert since we don't expect this condition to happen - if it does, it's a bug.

torch/_dynamo/variables/dicts.py
```diff
@@ -506,7 +506,11 @@ def call_method(
...
+                msg = ConstantVariable.create(
+                    f"dict.__setitem__ takes exactly two arguments ({len(args)} given)"
+                )
+                raise_observed_exception(AssertionError, tx, args=[msg])
```
williamwen42: `TypeError` here?

williamwen42 (commented):
meta point (cc @anijain2305)

The pattern
```python
msg = ConstantVariable.create(...)
raise_observed_exception(TypeError, tx, args=[msg])
```

shows up a lot - could we refactor this?

torch/_dynamo/variables/builtin.py
can-gaa-hou: Thanks for pointing it out! Fixed.

torch/_dynamo/variables/lists.py
can-gaa-hou: Yes, you are right! There is no need to split it.

torch/_dynamo/variables/misc.py
can-gaa-hou: Ditto

torch/_dynamo/variables/dicts.py
can-gaa-hou: Yes, should be TypeError here. Fixed.

torch/_dynamo/variables/builtin.py
can-gaa-hou: Thanks for pointing it out! Should be a graph break here.

torch/_dynamo/variables/lists.py
```diff
@@ -153,7 +153,9 @@ def call_method(
                 )
                 raise_observed_exception(TypeError, tx, args=[msg])
 
-            assert not kwargs and len(args) == 1
+            if kwargs:
```
Lucaskabela: I think we dropped the check for len(args) here

torch/_dynamo/variables/lists.py
```diff
@@ -1115,11 +1117,17 @@ def call_method(
         kwargs: dict[str, "VariableTracker"],
     ) -> "VariableTracker":
         if name == "__getitem__":
-            assert not kwargs and len(args) == 1
+            if len(args) != 1 or kwargs:
```
Lucaskabela: nit: let's keep the order (kwargs or len(args) != 1) for consistency

Lucaskabela (changes requested):
Thanks for the changes - I will second @williamwen42's comment that we should put this in a helper fn since it appears so much.  We can probably add it to `variables/base.py`.  Something like

```
def raise_type_error_exc(tx: InstructionTranslator, msg_str: str) -> None:
    msg = ConstantVariable.create(msg_str)
    raise_observed_exception(TypeError, tx, args=[msg])
```

torch/_dynamo/variables/lists.py
can-gaa-hou: https://github.com/pytorch/pytorch/blob/21697feff257ad04dd916ef63b8b841c38f7e9ee/torch/_dynamo/variables/lists.py#L150-L156

Yeah.. But we can combine these two into one check because `len(args) != 1` promises `len(args) == 1`, doesn't it?

torch/_dynamo/variables/lists.py
can-gaa-hou: Fixed

Lucaskabela (approved):
LGTM - will probably need to rebase since another change to graph_break_registry landed

Might be easiest to:
1) Rebase on main
2)`git checkout main torch/_dynamo/graph_break_registry.json`
3) Run lintrunner to repopulate the graph_break_registry change automagically

Thanks for working hard on this!

williamwen42 (approved):

Lucaskabela (approved):
Thanks for working hard to clean up asserts here! LGTM

williamwen42:
@can-gaa-hou the registry and the `unimplemented_v2` fields are out of sync - just revert your graph_break_registry.json changes and apply `lintrunner -a` again

PR #165434, opened 2025-10-14
[ISSUES] Update ci:sev template to include ci: disable-autorevert label

jeanschmidt:
We noticed that disabling autorevert in any and all ci:sevs is too impactful, as ci: sevs are sometimes created just to communicate an action or a impactful change. But sometimes durring a SEV we might not want to disable autorevert anyways, a example is a ci: sev impacting jobs we don't use as basis for autorevert.

So, by default we add the cl: disable-autorevert tag, that can optionally be removed, allowing a ci:sev be nonblocking for autorevert.

### DIFF
--- a/.github/ISSUE_TEMPLATE/ci-sev.md
+++ b/.github/ISSUE_TEMPLATE/ci-sev.md
@@ -2,12 +2,13 @@
 name: "⚠️ CI SEV"
 about: Tracking incidents for PyTorch's CI infra.
 title: ''
-labels: ''
+labels: 'ci: sev, ci: disable-autorevert'
 assignees: ''
 
 ---
 
-> NOTE: Remember to label this issue with "`ci: sev`"
+> NOTE: - Remember to keep this issue with "`ci: sev`" label
+>.      - If you want autorevert to be disabled, keep the ci: disable-autorevert label
 
  <!-- Add the `merge blocking` label to this PR to prevent PRs from being merged while this issue is open -->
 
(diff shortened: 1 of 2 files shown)


atalman (changes requested):
Are third_party cutlass changes intentional ?

.github/ISSUE_TEMPLATE/ci-sev.md
```diff
@@ -2,12 +2,13 @@
 name: "⚠️ CI SEV"
 about: Tracking incidents for PyTorch's CI infra.
 title: ''
-labels: ''
+labels: 'ci: sev, ci: disable-autorevert'
```
malfet: I'm not sure what pattern it follows, but shouldn't it be `ci: sev`, `ci: disable-autorevert`?

jeanschmidt: I am not sure as well, the quotes were there already, not sure if I should keep them or remove them...

jeanschmidt: ahh, because of the `:` it must be there, otherwise it is a wrong syntax, according to my editor

malfet: Please remove the labels, otherwise anyone will be able to create CI:sevs

jeanschmidt: good point, makes sense

PR #165436, opened 2025-10-14
Refactor and unify v1/v2 _scaled_mm codes

slayton58:
* __->__ #165436

Summary:

* Refactor out some core routines (scaled_gemm, auto-tuned scaled_gemm)
* Unify v1/v2 dispatch calls where possible
* Simplify call pattern w.r.t. CUDA/ROCM for easier readability.

Test Plan:

```
pytest -svv test/test_scaled_matmul_cuda.py
```

Reviewers:

Subscribers:

Tasks:

Tags:

### DIFF
--- a/aten/src/ATen/native/cuda/Blas.cpp
+++ b/aten/src/ATen/native/cuda/Blas.cpp
@@ -1230,7 +1230,202 @@ std::pair<ScalingType, ScalingType> get_joint_scaling(
   );
 }
 
+Tensor&
+_tunable_scaled_gemm_rocm(
+          cublasCommonArgs& args,
+          const Tensor& mat1, const Tensor& mat2,
+          const Tensor& scale_a, const Tensor& scale_b,
+          const ScalingType scaling_choice_a, const ScalingType scaling_choice_b,
+          const std::optional<Tensor>& bias,
+          const bool use_fast_accum,
+          Tensor& out) {
+#ifdef USE_ROCM
+#define TUNABLE_DISPATCH(BLASOP_A, BLASOP_B)                            \
+      if (mat1.scalar_type() == ScalarType::Float8_e4m3fnuz) {        \
+        if (mat2.scalar_type() == ScalarType::Float8_e4m3fnuz) {      \
+          static at::cuda::tunable::ScaledGemmTunableOp<              \
+              at::Float8_e4m3fnuz, at::Float8_e4m3fnuz, scalar_t,     \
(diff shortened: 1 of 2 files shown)


aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1230,8 +1230,205 @@ std::pair<ScalingType, ScalingType> get_joint_scaling(
...
+// but can fallback to at::cuda::blas::scaled_gemm
+#ifdef USE_ROCM
+  auto tuning_ctx = at::cuda::tunable::getTuningContext();
+  bool tunable_op_enabled = tuning_ctx->IsTunableOpEnabled();
+#else
+  bool tunable_op_enabled = false;
+#endif
+  if (tunable_op_enabled) {
+      // Only available on ROCM
+      return _tunable_scaled_gemm_rocm(
```
drisspg, line 1377: nit: namespace?

test/test_scaled_matmul_cuda.py
```diff
@@ -1050,15 +1050,24 @@ def test_scaled_mm_vs_emulated_block_wise_verify_small_shapes(
...
+
         if rhs_block == 1:
             y_scales = y_scales.t().contiguous().t()
+            rhs_recipe = ScalingType.BlockWise1x128
+        else:
+            rhs_recipe = ScalingType.BlockWise128x128
 
         # Verify that actual F8 mm doesn't error
-        mm_float8(
+        scaled_mm_wrap(
```
drisspg, line 1056: can we remove `mm_float8`? Honestly not sure if this is still used

drisspg (approved):
🚀

test/test_scaled_matmul_cuda.py
slayton58: I've just replaced the last few (missed) uses of this in the test with `scaled_mm_wrap` - if everything still passes ok I'll go ahead and delete.

aten/src/ATen/native/cuda/Blas.cpp
slayton58: i'm a little confused about this - `_scaled_gemm` and `_tunable_scaled_gemm_rocm` are both already in an anonymous namespace.

slayton58: Talked offline - this was checking that locally-defined, non-namespaced `_tunable_scaled_gemm_rocm` on one path, and fully-namespaced `at::cuda::blas::scaled_gemm` on the other was intentional.

PR #161916, opened 2025-09-01
[CD] Add cuda 13.0 libtorch builds, remove CUDA 12.9 builds

atalman:
Related to https://github.com/pytorch/pytorch/issues/159779

Adding CUDA 13.0 libtorch builds, followup after https://github.com/pytorch/pytorch/pull/160956
Removing CUDA 12.9 builds, See https://github.com/pytorch/pytorch/issues/159980

cc @albanD @ptrblck @nWEIdia @malfet

PR #165465, opened 2025-10-14
Don't link with libnvToolsExt when building for 12.9

huydhn:
This is to bring back this logic from https://github.com/pytorch/pytorch/pull/161916/files#diff-bf46b4a09ca67e50622bf84fefc0d11b584ffcc24ee6cc5019cf0fc7565d81a8L170.  Building libtorch on 12.9 is failing otherwise https://github.com/pytorch/pytorch/actions/runs/18458531395/job/52610761895:

```
cp: cannot stat '/usr/local/cuda/lib64/libnvToolsExt.so.1': No such file or directory
```

### DIFF
--- a/.ci/manywheel/build_cuda.sh
+++ b/.ci/manywheel/build_cuda.sh
@@ -199,7 +199,6 @@ if [[ $CUDA_VERSION == 12* || $CUDA_VERSION == 13* ]]; then
                 "libnvrtc.so.12"
                 "libcupti.so.12")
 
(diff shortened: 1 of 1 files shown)


.ci/manywheel/build_cuda.sh
atalman, line 160: I believe this comment can be removed

atalman (approved):

malfet (approved):

PR #165467, opened 2025-10-14
[PP] Fix edge case with FSDP when stages_per_rank > 3

H-Huang:
There is an edge case with FSDP + PP when we add UNSHARD + RESHARD, we at max have 3 stages unsharded, https://github.com/pytorch/pytorch/blob/3f83e8915e86a93da2fe01fda45602dcd0e3ebfd/torch/distributed/pipelining/schedules.py#L1029-L1031

This change is need to be able to unshard and reshard a stage multiple times.


wwwjn (approved):
LGTM!

PR #165469, opened 2025-10-14
use_cpp_bmm_template supports more use cases

helloguo:
Summary: In certain scenarios, such as when the first stride is 0, the entire tensor may not be contiguous, but the 2D matrix within each batch can still be contiguous, allowing us to apply max autotune. This diff specifically checks for contiguity within the 2D matrix of each batch, and enables more uses for cpp bmm template.

Differential Revision: D84561331


desertfire (commented):
LGTM. Can you add a test for this?

desertfire (approved):

PR #165473, opened 2025-10-14
 [dynamo] Fix tensor factory functions ignoring default device in torch.compile

hinriksnaer:
Fixes #159041 #151202 
## Summary

This PR fixes a bug where `torch.compile` does not respect `torch.set_default_device()` for tensor creation functions like `torch.randn`, `torch.zeros`, `torch.ones`, etc.

## Problem

When a default device is set using `torch.set_default_device()` before compilation, tensor factory functions inside the compiled region ignore this setting and create tensors on CPU instead:

```python
torch.set_default_device("cuda")

@torch.compile
def foo(x):
    return x * torch.randn(1)  # Creates tensor on CPU instead of CUDA

x = torch.randn(1)  # Correctly on CUDA
result = foo(x)  # Fails: device mismatch between cuda:0 and cpu
```

## Root Cause

Tensor factory functions (e.g., `torch.randn`, `torch.zeros`, `torch.ones`) were not included in the `get_overridable_functions()` list. This prevented `__torch_function__` dispatch from being triggered during Dynamo tracing, so the existing `DeviceContext` torch function mode could not intercept these calls to inject the device parameter.

## Solution

**Single-line fix** in `torch/_dynamo/variables/torch.py`:

Replace the manual list of device constructors with the authoritative `_device_constructors()` function:

```python
@functools.cache
def get_overridable_functions():
    from itertools import chain
    from torch.overrides import get_overridable_functions as get_overridable_functions_
    from torch.utils._device import _device_constructors

    funcs = set(chain.from_iterable(get_overridable_functions_().values()))

    # Add all device constructor functions to support torch.set_default_device()
    funcs.update(_device_constructors())

    return funcs
```

### Before (Fails)
```python
import torch

torch.set_default_device("cuda")

@torch.compile
def fn(input_tensor):
    input_tensor = input_tensor / torch.tensor([1000], dtype=torch.int64)
    return input_tensor

fn(torch.randn([10, 10], dtype=torch.float32))
# RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```

### After (Works)
```python
import torch

torch.set_default_device("cuda")

@torch.compile
def fn(input_tensor):
    input_tensor = input_tensor / torch.tensor([1000], dtype=torch.int64)
    return input_tensor

result = fn(torch.randn([10, 10], dtype=torch.float32))
print(result.device)  # cuda:0
# Success! All tensors are on CUDA
```

### DIFF
--- a/test/dynamo/test_modes.py
+++ b/test/dynamo/test_modes.py
@@ -820,7 +820,9 @@ class TorchFunctionModeTests(torch._dynamo.test_case.TestCase):
         self.assertEqual(input_tensor.device.type, GPU_TYPE)
 
         # Verify the computation worked correctly
-        expected = input_tensor / torch.tensor([1000], dtype=torch.int64, device=GPU_TYPE)
+        expected = input_tensor / torch.tensor(
+            [1000], dtype=torch.int64, device=GPU_TYPE
+        )
         self.assertEqual(result, expected)
 
         torch.set_default_device(None)
--- a/test/dynamo/test_modes.py
+++ b/test/dynamo/test_modes.py
@@ -790,6 +790,41 @@ class TorchFunctionModeTests(torch._dynamo.test_case.TestCase):
                         torch.ones(2, 2, 2, 2),
                     )
 
+    @requires_gpu
+    def test_set_default_device_with_factory_functions(self):
+        """Test that torch.set_default_device works with tensor factory functions in torch.compile"""
+        torch.set_default_device(GPU_TYPE)
+
+        @torch.compile(fullgraph=True)
+        def fn(input_tensor):
+            # Test various factory functions that should respect default device.
+            # This can be expanded to fully cover all factory functions.
+            t1 = torch.randn(10, 10)
+            t2 = torch.zeros(5, 5)
+            t3 = torch.ones(3, 3)
+            t4 = torch.empty(2, 2)
+            t5 = torch.tensor([1000], dtype=torch.int64)
+
+            result = input_tensor / t5
+            return result, t1, t2, t3, t4
+
+        input_tensor = torch.randn([10, 10], dtype=torch.float32)
+        result, t1, t2, t3, t4 = fn(input_tensor)
+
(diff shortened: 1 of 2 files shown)


hinriksnaer:
Upon further inspection this seems to be caused by our functions being included in the following

```python
@functools.cache
@_disable_user_warnings
def get_ignored_functions() -> set[Callable]:
    """
    Return public functions that cannot be overridden by ``__torch_function__``.

    Returns
    -------
    set[Callable]
        A tuple of functions that are publicly available in the torch API but cannot
        be overridden with ``__torch_function__``. Mostly this is because none of the
        arguments of these functions are tensors or tensor-likes.

    Examples
    --------
    >>> torch.Tensor.as_subclass in torch.overrides.get_ignored_functions()
    True
    >>> torch.add in torch.overrides.get_ignored_functions()
    False
    """
    Tensor = torch.Tensor
    return {
        torch.typename,
        torch.is_tensor,
        torch.is_storage,
        ...
        torch.tensor, # <- Removing this resolves the issue on an otherwise unedited branch
        ...
    }
```

Given all of this, the following rationale might not necessarily fully hold up since device constructors can take tensor arguments and should respect `torch.set_default_device()` via `DeviceContext`, which relies on `__torch_function__` dispatch.

```python
  """
  A tuple of functions that are publicly available in the torch API but cannot
  be overridden with __torch_function__. Mostly this is because none of the
  arguments of these functions are tensors or tensor-likes.
  """
```
It might however be more risky to make changes here so best course of action might need maintainer input.

mlazos:
Hmm if nothing else breaks I'm okay with updating `get_overrideable_functions`, if something does end up breaking, we can just augment the list in TorchDynamo. 

I think `get_overrideable_functions` existed before torch function modes. If you only have a torch function subclass then the comment you quoted is correct. Torch function modes were introduced specifically to hit the case of the constructors w/o tensor arguments.

hinriksnaer:
@mlazos fixed the formatting so now all tests should pass

PR #165487, opened 2025-10-14
[simplefsdp] add manual bucketing pass

ruisizhang123:
As titled, this PR adds manual bucketing pass to SimpleFSDP. Users will need to parse FQNs they wanted to bucket together using `module_bucket_plans`. Then, `_manual_bucket_collectives` will get the node of the subgraphs correspond to each `bucket_module`, and bucket bucketable (FSDP-style) AG/RS together. `_manual_reorder_graph` reorders them for overlapping. 

For detailed performance, see this torchtitan PR: https://github.com/pytorch/torchtitan/pull/1881.

There are a few todo items isted in torchtitan PR. Let's start with this PR that implements FSDP+TP+llama3 manual bucketing. I will fix/add the rest in follow up PRs. 

### DIFF
--- a/torch/_inductor/fx_passes/overlap_manual_scheduling.py
+++ b/torch/_inductor/fx_passes/overlap_manual_scheduling.py
@@ -1,0 +1,357 @@
+from __future__ import annotations
+
+import heapq
+from collections import Counter, defaultdict
+from typing import Any, Optional
+
+import torch
+import torch.fx as fx
+from torch._dynamo.graph_deduplication import _stable_topological_sort
+from torch._inductor.fx_passes.bucketing import (
+    is_all_gather_into_tensor as is_all_gather,
+    is_reduce_scatter_tensor as is_reduce_scatter,
+    is_wait_tensor,
+    merge_all_gather_bucket,
+    merge_reduce_scatter_bucket,
+)
+from torch._inductor.fx_passes.overlap_preserving_bucketer import (
+    bucket_key,
+    OverlapPreservingBucketer,
+)
+from torch._inductor.fx_passes.overlap_scheduling import (
+    CollectiveInfo,
+    is_compute_node,
+    OverlapScheduler,
+)
+from torch.utils._ordered_set import OrderedSet
+
+from .graph_view import get_subgraph_by_path, GraphView, make_graph_view
+
+
+class ManualOverlapPreservingBucketer(OverlapPreservingBucketer):
+    """
+    Buckets collective operations based on user specifications.
+    The actual bucket happens in bucket_collectives, where all-gathers/reduce-scatters in
+        `nodes` will be buckted one single all-gather/reduce-scatter.
+    """
+
+    def __init__(
+        self, node_users: dict[fx.Node, OrderedSet[fx.Node]], *args: Any, **kwargs: Any
+    ):
+        super().__init__(*args, **kwargs)
+        self.node_users = node_users
+        self.wait_to_node_map: dict[fx.Node, fx.Node] = defaultdict()
+
+    def _check_recursive_dep(
+        self,
+        node: fx.Node,
+        target_op: str,
+        dep_dict: dict[torch.fx.Node, OrderedSet[torch.fx.Node]],
+    ) -> bool:
+        """
+        Check if the node is directly used for fetch parameters/gradients
+
+        TODO (ruisizhang123): currently, we assume the node only pre-fetch/update one parameter/gradient
+            We should handle multiple parameters/gradients update case by checking if there are non closure
+            computes along the path from primal/output to coll_node
+        """
+        deps: OrderedSet[fx.Node] = dep_dict[node]
+        seen_target_op = 0
+        for d in deps:
+            if d.op == target_op:
+                seen_target_op += 1
+
+        return seen_target_op == 1
+
+    def _bucket_group(self, coll_nodes: list[fx.Node]) -> None:
+        assert len(coll_nodes) > 0, "bucketed coll_nodes should have nonzero node"
+
+        waits = [self.collective_info[n].wait_node for n in coll_nodes]
+        # Use earliest wait insertion point
+        first_wait = min(waits, key=lambda w: self.node_idx[w])
+        # Find insertion location
+        first = coll_nodes[0]
+        next_node = first
+        while next_node in coll_nodes:
+            next_node = next_node.next
+
+        if is_all_gather(first):
+            new_nodes, replacements = merge_all_gather_bucket(
+                self.graph,
+                coll_nodes,
+                wait_insertion_point=first_wait,
+                insert_before=next_node,
+                mode="custom_ops",
+            )
+        elif is_reduce_scatter(first):
+            new_nodes, replacements = merge_reduce_scatter_bucket(
+                self.graph,
+                coll_nodes,
+                wait_insertion_point=first_wait,
+                insert_before=next_node,
+                mode="custom_ops",
+            )
+        else:
+            raise ValueError(
+                "bucket non all_gather/reduce_scatter node is not supported"
+            )
+
+        # Identify the new wait and start
+        new_waits = [n for n in new_nodes if is_wait_tensor(n)]
+        assert len(new_waits) == 1, f"Expected exactly one new wait, got {new_waits}"
+        new_wait = new_waits[0]
+        new_start = new_wait.args[0]
+        assert isinstance(new_start, fx.Node)
+
+        node_type = (
+            "bucketed_all_gather" if is_all_gather(first) else "bucketed_reduce_scatter"
+        )
+        for n in new_nodes:
+            n.meta["nn_module_stack"] = coll_nodes[0].meta.get("nn_module_stack", "")
+            n.meta["fwd_nn_module_stack"] = coll_nodes[0].meta.get(
+                "fwd_nn_module_stack", ""
+            )
+            if n == new_wait:
+                node_type = node_type + "_wait"
+            n.meta["manual_bucket_node_type"] = node_type
+            if "wait" in node_type:
+                self.wait_to_node_map[n] = new_wait
+
+    def manual_bucket_collectives(self, nodes: list[fx.Node]) -> None:
+        """
+        Bucket all all-gather/reduce-scatter nodes from nodes into one all-gather/reduce-scatter.
+        """
+        # Filter out valid collectives
+        collectives = [n for n in nodes if n in self.collective_info]
+        if collectives == []:
+            return
+        grouped_collectives: dict[object, OrderedSet[fx.Node]] = defaultdict(OrderedSet)
+        for node in collectives:
+            key = bucket_key(node)
+            if not (is_all_gather(node) or is_reduce_scatter(node)):
+                continue
+            # We only want to bucket all-gather/reduce-scatter that
+            # 1. all_gather that have ancestors dependent only on input placeholder(parameters)
+            # 2. reduce scatter that the wait user node is returned as output(gradients)
+            if is_all_gather(node) and not self._check_recursive_dep(
+                node, "placeholder", self.node_ancestors
+            ):
+                continue
+            if is_reduce_scatter(node) and not self._check_recursive_dep(
+                self.collective_info[node].wait_node, "output", self.node_users
+            ):
+                continue
+            if key is not None:
+                grouped_collectives[key].add(node)
+
+        for key, nodes in grouped_collectives.items():  # type: ignore[arg-type]
+            self._bucket_group(list(nodes))
+
+
+class ManualOverlapScheduler(OverlapScheduler):
+    """
+    Scheduler that manual buckets and reorders collective nodes based on module_bucket_plans
+    """
+
+    def __init__(
+        self,
+        gm: fx.GraphModule,
+        module_bucket_plans: list[list[str] | str],
+        insert_overlap_deps: bool,
+    ):
+        super().__init__(
+            gm,
+            max_in_flight_gb=0.0,
+            max_compute_pre_fetch=0,
+            collective_bucketing=True,
+            insert_overlap_deps=insert_overlap_deps,
+            compute_overlap_multipler=0.0,
+            max_coll_distance=0,
+            custom_runtime_estimation=None,
+            collective_estimator="analytical",
+        )
+        self.module_bucket_plans = module_bucket_plans
+        self.nodes_in_subgraph: list[list[fx.Node]] = []
+
+        self.node_users: dict[fx.Node, OrderedSet[fx.Node]] = self._collect_node_users()
+        self.bucketer = ManualOverlapPreservingBucketer(
+            graph=self.graph,
+            collective_info=self.collective_info,
+            node_ancestors=self.node_ancestors,
+            node_users=self.node_users,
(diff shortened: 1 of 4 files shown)


torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,583 @@
...
+    if len(s) == 1:
+        return s_
+    first: str = s.pop(0)
+    s: list[str] = [(num_spaces * " ") + line for line in s]
+    joint_s: str = "\n".join(s)
+    joint_s = first + "\n" + joint_s
+    return joint_s
+
+
+class Container:
```
ezyang: docstring would be appreciated here

torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,583 @@
...
+    return new_graph
+
+
+def _is_root(stack: str) -> bool:
+    return stack == ""
+
+
+def make_graph_view(graph: fx.Graph) -> Container:
+    """
+    Code from: https://github.com/meta-pytorch/autoparallel/pull/158
```
ezyang: If we're putting this in core, let's do it for real. Dedicated module, docs, tests. cc @fmassa @eellison. I haven't used this API yet so I don't have a real world opinion on it.

torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,583 @@
...
+
+def manual_overlap_bucketing(
+    gm: torch.fx.GraphModule,
+    module_bucket_plans: list[str],
+) -> torch.fx.GraphModule:
+    """Schedule nodes based on user specifications in module_bucket_plans
+    The manual overlapping consists of two steps:
+    Step 1: bucket all-gather/reduce-scatter in each module in module_bucket_plans
+    Step 2: reorder all-gather to overlap with last module_bucket &
+        reorder reduce-scatter to overlap with next module_bucket
```
ezyang, line 344: The reordering here is just the SimpleFSDP strategy that you used to have in Inductor IR, right?

ruisizhang123:
> Can't land this without some tests. Would it be better for this to live in torchtitan? I can't tell... but since simple fsdp is in torchtitan and not in core, maybe it should live there!

This is the starting of manual optimization pass that is independent of simplefsdp, which I think should live in core. Currently, it uses similar reordering strategy, but I have a list of todo items to make it general here: https://github.com/pytorch/torchtitan/pull/1881. I will add them in followup PRs. 

I will add tests to it to make sure it can live in core :)

ezyang:
Bucketing, absolutely. Overlapping, I think you potentially may need something even more manual than this.

torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,583 @@
...
+from torch._inductor.fx_passes.bucketing import (
+    is_all_gather_into_tensor as is_all_gather,
+    is_reduce_scatter_tensor as is_reduce_scatter,
+    is_wait_tensor,
+    merge_all_gather_bucket,
+    merge_reduce_scatter_bucket,
+)
+from torch._inductor.fx_passes.overlap_preserving_bucketer import (
+    bucket_key,
+    OverlapPreservingBucketer,
```
ezyang, line 19: @eellison to comment on inheritance here; note that we have some internal code also inheriting from this class in a similar way too.

torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,599 @@
...
+        .replace("['", ".")
+        .replace("']", "")
+    )
+    return name
+
+
+def _find_key_nodes(nodes: list[fx.Node]) -> tuple[list[fx.Node], list[fx.Node]]:
+    root = []
+    outputs = []
+    nodes_set = OrderedSet(nodes)
```
wconstab: nodes_set appears to never have anything added to it.  what is the purpose of it?

ruisizhang123: good catch, I copied from the prototype loll

torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,571 @@
...
+    name: str = (
+        val.replace("L['self'].", "")
+        .replace("_modules['", "")
+        .replace("['", ".")
+        .replace("']", "")
+    )
+    return name
+
+
+def _find_key_nodes(nodes: list[fx.Node]) -> tuple[list[fx.Node], list[fx.Node]]:
```
wconstab: i'd appreciate a docstring at least, or a renamed function, that makes it more intuitive what this does.  (what is a "key" node?)

torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,571 @@
...
+    for node in nodes:
+        for x in node.all_input_nodes:
+            if x not in nodes:
+                root.append(x)
+        if all(x not in nodes for x in node.users):
+            outputs.append(node)
+    return root, outputs
+
+
+def _make_subgraph(nodes: list[fx.Node]) -> fx.Graph:
```
wconstab: similar comment as above: intuitively a function named "make_subgraph" i'd expect some input arg that helps identify what part of the graph to include in the subgraph.   In this case it sounds like it makes the 'subgraph containing key nodes', which, is still opaque to me.

torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,571 @@
...
+
+
+def get_subgraph_by_path(
+    graph_view: Container, paths: Union[str, list[str]]
+) -> list[fx.Node]:
+    """
+    Get subgraph by path(s).
+    Args:
+        graph_view (object): Root graph view object.
+        paths (str or list of str): Path(s) to subgraph.
```
wconstab: any requirements here? overlapping paths ok? disjoint paths ok?

torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,571 @@
...
+    out_node = [env[x] for x in outputs]
+    new_graph.output(out_node)
+    return new_graph
+
+
+def _is_root(stack: str) -> bool:
+    return stack == ""
+
+
+def make_graph_view(graph: fx.Graph) -> Container:
```
wconstab: its a little odd to me that this function returns a 'Container' type, and then on the Container you can further call `.graph_view()` to get an fx graphmodule type.  I might not be understanding the design yet, but i wonder if it would be cleaner to have 
- Container class renamed to GraphView class
- functions like `get_subgraph_by_path` moved to class methods of GraphView
- make_graph_view moves to GraphView.__init__
- GraphView.graph_view() still doesn't make sense to me

torch/_inductor/fx_passes/overlap_manual_scheduling.py
ruisizhang123: not, it has to be disjoint paths. will add an assertion here.

ruisizhang123: there is no requirement. this function will collect all of the nodes as long as they belong to the path(s). This is for the case that we may want to put nodes from multiple paths in one bucket. The disjoint graph assertion bucket should happen [here](https://github.com/pytorch/pytorch/pull/165487/files/c472083fc16e61b90b17ac05fa4a3c895c5e2234#diff-90e7edbccb860142a1bf9f7dca61f9aa449129d3a435eca1e3cae89c3aaf1cb1R542-R549) to ensure two buckets have no overlap. Will add an assertion in _obtain_nodes_in_subgraph func.

torch/_inductor/fx_passes/overlap_manual_scheduling.py
ruisizhang123: We are using Container class to have a hierarchical mapping between nodes & their module names. I agree it would be more intuitive to rename Container to Graph_view, will do this. We can also rename `GraphView.graph_view()` to sth like `GraphView.obtain_subgraph()`?

ruisizhang123: I realized `GraphView.graph_view()` is not actually used in our bucketing & reordering pass to get subgraphs. Instead, I decided to use get_subgraph_by_path func, which is more friendly to input FQNs as strings from torchtitan. Thus, I removed the legacy functions including`_make_subgraph`, `_find_key_nodes` and `graph_view` in the new code.

wconstab: ok removing those functions will address a lot of my questions.

torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,529 @@
...
+        .replace("']", "")
+    )
+    return name
+
+
+def _is_root(stack: str) -> bool:
+    return stack == ""
+
+
+def make_graph_view(graph: fx.Graph) -> GraphView:
```
eellison: Would it make sense to put this in a differnet file?

torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,529 @@
...
+        """
+        deps: OrderedSet[fx.Node] = dep_dict[node]
+        seen_target_op = 0
+        for d in deps:
+            if d.op == target_op:
+                seen_target_op += 1
+
+        return seen_target_op == 1
+
+    def _bucket_group(self, coll_nodes: list[fx.Node]) -> None:
```
eellison, line 66: reason you had to define this, just curious ?

torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,529 @@
...
+                get_node_by_path(graph_view, p).data for p in paths
+            )
+        )
+        return nodes
+    else:
+        node = get_node_by_path(graph_view, paths)
+        return node.data
+
+
+class ManualOverlapPreservingBucketer(OverlapPreservingBucketer):
```
eellison, line 31: What parts of the existing class are you using, just for my understanding?

torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,529 @@
...
+        for key, nodes in grouped_collectives.items():
+            self._bucket_group(list(nodes))
+
+
+class ManualOverlapScheduler(OverlapScheduler):
+    """
+    Scheduler that manual buckets and reorders collective nodes based on module_bucket_plans
+    """
+
+    def __init__(self, gm: fx.GraphModule, module_bucket_plans: list[list[str] | str]):
```
eellison: Same question here ? I guess this is heapq, mostly? anything else?

torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,529 @@
...
+from torch._inductor.fx_passes.overlap_scheduling import (
+    CollectiveInfo,
+    is_compute_node,
+    OverlapScheduler,
+)
+from torch.utils._ordered_set import OrderedSet
+
+
+def _get_module_stack(node: fx.Node) -> list[tuple[str, type[Any]]]:
+    if node.meta.get("nn_module_stack", "") == "":
```
ezyang: err why not just test if it's None?

torch/_inductor/fx_passes/overlap_manual_scheduling.py
ezyang: Also have Claude generate some simple unit tests for it

torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,529 @@
...
+    Example:
+        Input: "L['self']._modules['layers']['0']._modules['attention']"
+        Output: "layers.0.attention"
+    """
+    name: str = (
+        val.replace("L['self'].", "")
+        .replace("_modules['", "")
+        .replace("['", ".")
+        .replace("']", "")
+    )
```
ezyang: This is kind of terrible; also not sure if there's a preexisting utility for this

torch/_inductor/fx_passes/overlap_manual_scheduling.py
```diff
@@ -0,0 +1,529 @@
...
+def get_subgraph_by_path(
+    graph_view: GraphView, paths: Union[str, list[str]]
+) -> list[fx.Node]:
+    """
+    Get subgraph by path(s).
+    Args:
+        graph_view (object): Root graph view object.
+        paths (str or list of str): Path(s) to subgraph.
+    Returns:
+        object: Subgraph object or node.
```
ezyang: inconsistent with the type signature

torch/_inductor/fx_passes/overlap_manual_scheduling.py
ezyang: It is a bit different than `_apply_bucket` FWIW

ezyang (approved):
this has been stuck in review hell for a while, stamping to unblock, it's really low risk to put in

torch/_inductor/fx_passes/overlap_manual_scheduling.py
ruisizhang123: updated

torch/_inductor/fx_passes/overlap_manual_scheduling.py
ruisizhang123: I can rewrite it with regex function that extracts MODULE inside ['MODULE']. yeah, this one is a bit fragile.

torch/_inductor/fx_passes/overlap_manual_scheduling.py
ruisizhang123: for OverlapPreservingBucketer, I'm only using `__init__` func to get self.collective_info and self.node_ancestors. The manual bucketing logic is much simpler -- you basically put things in a bucket and call bucket helper functions.

torch/_inductor/fx_passes/overlap_manual_scheduling.py
ruisizhang123: I wanted to add more info to bucketed nodes to help reordering: (1) tag newly bucketed AG/RS's metadata, which helps to identify FSDP related comms in reordering; (2) keep track of newly bucketed wait and its mapping to bucketed AG/RS as `self.wait_to_node_map`. It would help adding dependencies in reordering.

torch/_inductor/fx_passes/overlap_manual_scheduling.py
ruisizhang123: yes!

torch/_inductor/fx_passes/overlap_manual_scheduling.py
ruisizhang123: yes, mostly helper variables (e.g., collective_info) and heapq to help reordering & bucketing.

PR #165490, opened 2025-10-14
[Mem Snapshot] Add Metadata Field

sraikund16:
Summary:
The implementation adds the ability to:

Set custom metadata strings that will be attached to all subsequent allocations
Clear or change the metadata at any point
View the metadata in memory snapshots via _dump_snapshot()

Test Plan: Added test in test_cuda.py and check manually in snapshot to see that metadata was added.

Differential Revision: D84654933


yushangdi:
do you know if it's hard to add the metadata to the visualization in https://docs.pytorch.org/memory_viz?

yushangdi (approved):

PR #165492, opened 2025-10-15
[DeviceMesh][2D] Use concatenate for 2D (FSDP+TP) instead of getting from root mesh

fduwjj:
* __->__ #165492
* #163358
* #166264

With concatenate API, we can directly combine two meshes together rather than getting the spmd mesh from root.

Differential Revision: [D85409698](https://our.internmc.facebook.com/intern/diff/D85409698)


fegin (approved):
Very nice, let's wait for all the tests and can you also import this PR so that we can also trigger all the internal tests?

PR #165055, opened 2025-10-09
[ATen] Vectorize 8 elements on 16 bit data types for sum/mean

PaulZhang12:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #165055
* #164790
* #165494

Benchmarks for a full reduction + reduction on the contiguous dimension. Vectorized loads do not occur on the non contiguous dimension. Benchmarking done for FP16/BF16, ~6% improvement on average across shapes, up to ~24% for single reduction on contiguous dimension and 46% for full reduce:
**BF16**
```
Tensor Shape         Operation    Full reduce (ms)     Contiguous dim (ms)  Full reduce (ms)     Contiguous dim (ms)  Full reduce diff %   Contiguous diff %   
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
(256, 256)           mean         0.022686             0.008263             0.015498             0.008117                          +46.38%               +1.80%
(256, 256)           sum          0.022769             0.008269             0.015628             0.008185                          +45.69%               +1.03%
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
(512, 512)           mean         0.014116             0.009545             0.012892             0.008839                           +9.49%               +7.99%
(512, 512)           sum          0.014110             0.009892             0.012891             0.008878                           +9.46%              +11.42%
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
(1024, 1024)         mean         0.014727             0.012642             0.014061             0.010519                           +4.74%              +20.18%
(1024, 1024)         sum          0.014376             0.012636             0.014069             0.010595                           +2.18%              +19.26%
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
(2048, 2048)         mean         0.018663             0.018294             0.018171             0.014678                           +2.71%              +24.64%
(2048, 2048)         sum          0.018638             0.017931             0.018142             0.014713                           +2.73%              +21.87%
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
(4096, 4096)         mean         0.034216             0.036953             0.033520             0.030585                           +2.08%              +20.82%
(4096, 4096)         sum          0.034196             0.036942             0.033518             0.030676                           +2.02%              +20.43%
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
(8192, 8192)         mean         0.087763             0.095201             0.085439             0.084960                           +2.72%              +12.05%
[...]

PR #164790, opened 2025-10-06
[ATen] Fix CUDA reduction warp shuffle order

PaulZhang12:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #164790

Typical warp shuffle reduction has the following pattern:
image

which is exhibited in Triton generated by torch.compile:
image

Switch the warp shuffle order to make bitwise equivalence between the 2 easier.
PTX difference between old and new, we see a few extra instructions: https://www.diffchecker.com/h6ly3INC/

Comparing the performance on different reduction operations, we see minimal differences. New represents the changes in this PR, old represents the past warp shuffle order:
```
Tensor Shape              Operation            New all dims (ms)       New dim=0 (ms)      New dim=1 (ms)     Old all dims (ms)    Old dim=0 (ms)      Old dim=1 (ms)
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
(1024, 1024)              mean                 0.015817             0.016259             0.013642             0.015990             0.016258             0.013631             
(1024, 1024)              sum                  0.015917             0.015906             0.013359             0.015707             0.016266             0.013226             
(1024, 1024)              min                  0.016021             0.024625             0.015631             0.015761             0.024485             0.015317             
(1024, 1024)              max                  0.016349             0.024971             0.015972             0.015771             0.025001             0.015314             
(1024, 1024)              argmin               0.018070             0.024448             0.015578             0.018135             0.025370             0.015322             
(1024, 1024)              argmax               0.018427             0.024859             0.015932             0.018164             0.024452             0.015639             
(1024, 1024)              var                  0.020078             0.026413             0.020295             0.020199             0.026381             0.020214             
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
(2048, 2048)              mean                 0.023826             0.023726             0.022273             0.023236             0.023776             0.022248             
(2048, 2048)              sum                  0.023840             0.023355             0.021974             0.023294             0.023354             0.021884             
(2048, 2048)              min                  0.024519             0.041263             0.024620             0.023292             0.041491             0.024358             
(2048, 2048)              max                  0.024509             0.041670             0.024277             0.023334             0.041231             0.024395             
(2048, 2048)              argmin               0.026125             0.041282             0.024567             0.026772             0.041773             0.024296             
(2048, 2048)              argmax               0.026117             0.041487             0.024572             0.026412             0.041477             0.024273             
[...]

PR #165494, opened 2025-10-15
[ATen] Remove explicit casting of complex nansum during accumulation

PaulZhang12:
* #165187
* #165178
* #165055
* #164790
* __->__ #165494

https://github.com/pytorch/pytorch/pull/164790 modifies aten to perform a different reduction order intra warp. However, this change exposed a large difference in a sum for complex32. Namely the case:

```
import torch

a = torch.tensor([[ 4.82031250+7.34765625j,
           -3.37109375-1.9501953125j],

         [ 3.7832031250-2.43359375j,
           -6.07812500+5.32812500j]], dtype=torch.complex32, device='cuda:0')

sum_out = torch.sum(a)
nansum_out = torch.nansum(a)
torch.testing.assert_close(
    sum_out,
    nansum_out,
    rtol=0,
    atol=0,
)
```

Here, the result of `sum` and `nansum` differed significantly by 1e-2. Further investigation showed that the explicit casting of b back to `arg_t` from `scalar_t` was the root cause. `arg_t` is the dtype of the accumulator, ComplexFloat, and `scalar_t` of the input dtype, ComplexHalf. When we cast in the reduction to the accumulator order, that means the input is still of ComplexHalf, which loses precision as it can store intermediate values.

### DIFF
--- a/aten/src/ATen/native/SharedReduceOps.h
+++ b/aten/src/ATen/native/SharedReduceOps.h
@@ -392,8 +392,8 @@ struct NormTwoOps {
 
 template <typename acc_t, typename data_t>
 struct NanSumOps {
-  inline C10_DEVICE acc_t reduce(acc_t a, acc_t b, int64_t /*idx*/) const {
-    return a + (at::_isnan(b) ? acc_t{0.} : b);
+  inline C10_DEVICE acc_t reduce(acc_t a, data_t b, int64_t /*idx*/) const {
+    return a + (at::_isnan(b) ? acc_t{0.} : acc_t{b});
   }
 
   inline C10_DEVICE acc_t combine(acc_t a, acc_t b) const {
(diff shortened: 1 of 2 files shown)


aten/src/ATen/native/SharedReduceOps.h
```diff
@@ -392,8 +392,8 @@ struct NormTwoOps {
 
 template <typename acc_t, typename data_t>
 struct NanSumOps {
-  inline C10_DEVICE acc_t reduce(acc_t a, data_t b, int64_t /*idx*/) const {
```
ngimel, line 395: this is actually fine, `reduce` is called only to add a single number `b` which would be representable in `data_t` to accumulator

aten/src/ATen/native/cuda/ReduceSumProdKernel.cu
```diff
@@ -48,6 +48,7 @@ struct sum_functor<c10::complex<at::Half>> {
       return a + b;
     }
     );
+    std::cout << "Jiterator" << std::endl;
```
ngimel: stray print

ngimel (approved):

PR #165391, opened 2025-10-14
[user-streams] Add stream support to inductor wrapper codegen

mlazos:
This PR implements stream codegen to the SubgraphWrapper Codegen. It supports codegen for enter/exiting stream contexts and calling record_stream on returned tensors. 

Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #174224
* #174223
* #165505
* __->__ #165391
* #165504
* #165390

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @chenyang78

PR #165504, opened 2025-10-15
[user-streams] Add support functions for stream codegen

mlazos:
This PR adds utility functions for managing streams and a stream pool. 

* #174224
* #174223
* #165505
* #165391
* __->__ #165504
* #165390


torch/_inductor/stream_utils.py
```diff
@@ -0,0 +1,138 @@
...
+    "CUDAStreamPool",
+    "get_cuda_stream_pool",
+    "get_stream_name",
+]
+
+DEFAULT_STREAM: str = "default_stream"
+DEFAULT_STREAM_IDX: int = 0
+ENTRANCE_EVENT: str = "event0"
+EVENT_NAME_TEMPLATE: str = "event{event_idx:d}"
+STREAM_NAME_TEMPLATE: str = "stream{stream_idx:d}"
```
eellison: isnt this duped in the other pr?

torch/_inductor/stream_utils.py
```diff
@@ -0,0 +1,138 @@
...
+    """Generate CUDA Stream name from stream index number.
+
+    Args:
+        stream_idx: Non-negative index number. 0 refers to the default stream, others refer to side
+            streams.
+    """
+    if stream_idx == 0:
+        return DEFAULT_STREAM
+    else:
+        return STREAM_NAME_TEMPLATE.format(stream_idx=stream_idx)
```
eellison, line 37: will this always be knowable? i think wed have to query what the stream is. especially bc streams are passed in as first class objects.

torch/_inductor/stream_utils.py
```diff
@@ -0,0 +1,138 @@
...
+        """Initializesthe CUDAStreamPool instance.
+
+        Args:
+            device (Optional[int], optional): The CUDA device ID.
+                Defaults to None (current device).
+            pool_size (int, optional): The maximum number of CUDA streams in the pool.
+                Defaults to 8.
+        """
+        self.pool_size: int = pool_size
+        self.stream_queue: queue.Queue[torch.cuda.Stream] = queue.Queue(maxsize=pool_size)
```
eellison: is this necessary ? Pytorch has its own cache of streams in eager

torch/_inductor/stream_utils.py
```diff
@@ -0,0 +1,138 @@
...
+        return self.stream_queue.get()
+
+    def release(self, stream: torch.cuda.Stream | None) -> None:
+        """Return a CUDA stream back to the pool.
+
+        Args:
+            stream (Optional[torch.cuda.Stream]): The CUDA stream to return to the pool.
+        """
+        if stream is not None:
+            self.stream_queue.put(stream)
```
eellison: when would this ever be none

eellison (commented):
i don't think we need a stream pool, since pytorch eager manages its own. is it necessary ?

torch/_inductor/stream_utils.py
mlazos: Yeah you're right, I'm curious why the NVIDIA folks did this, but I'll remove, creating streams is pretty lightweight as is.

torch/_inductor/stream_utils.py
mlazos: I'm going to remove this

torch/_inductor/stream_utils.py
mlazos: Yeah we assign these in dynamo, it's an identifier that we use and the actual stream object may be different from run to run if it's created within the compiled region.

eellison (approved):

PR #165505, opened 2025-10-15
[user-streams] Add stream support to scheduler

mlazos:
Summary

  Add CUDA stream context management to the inductor scheduler's codegen phase. This is the
  foundational infrastructure for multi-stream code generation: it tracks which stream is currently
  active and generates the appropriate with torch.cuda.stream(...) enter/exit wrappers around
  kernels.

  Key pieces:
  - _current_stream_ctx field tracks the active EnterCudaStreamContextLine during codegen
  - current_stream_idx / current_stream_name properties expose the current stream state (also used by
   the wrapper's nesting guard)
  - generate_stream_ctx_switching() is the main entry point called per-node during codegen — it
  compares the node's assigned stream to the current stream and emits enter/exit code only when
  switching is needed. NopKernelSchedulerNodes inherit the previous stream context since they
  generate no kernel code.

  This commit adds the codegen-time stream management; the stream assignment logic (node_to_stream,
  _populate_stream_assignments) and the codegen callsite are in the next commit in the stack.

  Test plan

  - Tests are in the third commit of the stack (test/inductor/test_user_streams.py)

* #174224
* #174223
* __->__ #165505


torch/_inductor/scheduler.py
```diff
@@ -6989,6 +7010,310 @@ def update_zero_dim_cpu_tensor(self) -> None:
                     ):
                         V.graph.zero_dim_cpu_tensor_list.add(read.name)
```
eellison, line 7612:
```
get_cross_stream_dependencies(node): Analyzes a node's reads to determine:
- Which CUDA events from other streams must be waited on
- Which buffers from other streams need record_stream calls
```

why do we need this ? aren't we following user explicit side stream and synchronization and such ?

eellison (commented):
one comment on design prior to full review

torch/_inductor/scheduler.py
mlazos: Yeah let me think about this - this might be an artifact from the NVIDIA automatic streams stuff. I think you're right.

mlazos: Removed

torch/_inductor/scheduler.py
```diff
@@ -7379,6 +7385,66 @@ def update_zero_dim_cpu_tensor(self) -> None:
...
+        if (stream_idx := self.current_stream_idx) is not None:
+            return get_stream_name(stream_idx)
+        else:
+            return None
+
+    def generate_stream_ctx_enter(self, node: BaseSchedulerNode) -> None:
+        """Code-gen to enter the Stream context assigned to node."""
+        assert not isinstance(node, NopKernelSchedulerNode)
+        # pyrefly: ignore[missing-attribute]
+        node_stream = self.node_to_stream[node]
```
shunting314: You have a PR ordering issue? Since node_to_stream is defined in the next PR. (Thanks Claude pointing this out)

torch/_inductor/scheduler.py
```diff
@@ -7379,6 +7385,66 @@ def update_zero_dim_cpu_tensor(self) -> None:
...
+        self._current_stream_ctx = V.graph.wrapper_code.codegen_cuda_stream_enter(
```
desertfire: Not sure if it exists in other PRs of your stack, but does CppWrapperCodegen support this?

mlazos: I don't think it will yet, however we can codgen stream indices directly into kernel calls if needed which would probably be enough to support it in cpp.

torch/_inductor/scheduler.py
mlazos: Yeah let me fix this

shunting314 (approved):

PR #165515, opened 2025-10-15
[9/N] Apply ruff UP035 rule

cyyever:
This is follow-up of #165214 to continue applying ruff UP035 rule to the code base. 

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @jeffdaily @sunway513 @jithunnair-amd @pruthvistony @ROCmSupport @dllehr-amd @jataylo @hongxiayang @naromero77amd @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela @xmfan

### DIFF
--- a/tools/stats/upload_stats_lib.py
+++ b/tools/stats/upload_stats_lib.py
@@ -9,10 +9,14 @@ import time
 import zipfile
 from functools import lru_cache
 from pathlib import Path
-from typing import Any, Callable, cast, Optional
+from typing import Any, cast, Optional, TYPE_CHECKING
 
 import boto3  # type: ignore[import]
 import requests
+
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
 
 
 PYTORCH_REPO = "https://api.github.com/repos/pytorch/pytorch"
--- a/benchmarks/dynamo/cachebench.py
+++ b/benchmarks/dynamo/cachebench.py
@@ -6,7 +6,7 @@ import os
 import subprocess
 import sys
 import tempfile
-from typing import Callable
+from collections.abc import Callable
 
 from torch._inductor.utils import fresh_cache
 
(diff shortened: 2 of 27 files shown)


tools/stats/upload_stats_lib.py
```diff
@@ -9,12 +9,16 @@
...
 from functools import lru_cache
 from pathlib import Path
-from typing import Any, Callable, cast, Optional
+from typing import Any, cast, Optional, TYPE_CHECKING
 
 import boto3  # type: ignore[import]
 import requests
 
 
+if TYPE_CHECKING:
```
Lucaskabela, line 18: I am a tad concerned the red CI signal related to nightly means there is some function in here that we have broken - do we have a unit test or can you include an example of running a function from this file using the Callable annotation to derisk? Thanks!

Lucaskabela (changes requested):
Trying to get to the source of this CI failure that I personally have never seen before - let's add a snippet of running the mentioned file to check/derisk

tools/stats/upload_stats_lib.py
cyyever: `Callable` is used for static type checking in `upload_stats_lib.py` because the change imports it under `TYPE_CHECKING` so it is not responsible for the red CI.
The red CI is likely a flaky test being triggered. Let's wait for some time for verification.

cyyever: The CI failures are now marked as flaky.

Lucaskabela (approved):

PR #165531, opened 2025-10-15
[1/N] Add strict parameter to Python zip calls 

cyyever:
Add `strict=True/False` to zip calls in test utils. `strict=True` is passed when possible.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168

### DIFF
--- a/torch/testing/_internal/common_cuda.py
+++ b/torch/testing/_internal/common_cuda.py
@@ -252,7 +252,7 @@ def tf32_on_and_off(tf32_precision=1e-5, *, only_if=True):
 
         @functools.wraps(f)
         def wrapped(*args, **kwargs):
-            kwargs.update(zip(arg_names, args))
+            kwargs.update(zip(arg_names, args, strict=False))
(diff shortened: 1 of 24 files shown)


torch/testing/_internal/common_cuda.py
```diff
@@ -252,7 +252,7 @@ def wrapper(f):
 
         @functools.wraps(f)
         def wrapped(*args, **kwargs):
-            kwargs.update(zip(arg_names, args))
+            kwargs.update(zip(arg_names, args, strict=False))
```
Skylion007, line 255: Why does arg_names not match args length?

Skylion007 (approved):

torch/testing/_internal/common_cuda.py
cyyever: This is a util function and some tests call it with unequal sequences.

PR #136116, opened 2024-09-15
[BE][Ez]: Add missing cpu float16 coverage for avg_pool_3d

Skylion007:
Fixes a edge case where avg pool 3d didn't support all half dtypes on CPU. Adds the missing kernel to give near parity between GPU and CPU.

cc @ptrblck @msaroufim @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10

PR #165534, opened 2025-10-15
Fix avg_pool3d precision for BFloat16/Half 

### LINKED ISSUE #140696: inconsistency in ```torch.nn.functional.adaptive_avg_pool3d``` on CPU and GPU
### 🐛 Describe the bug

consistency check of function ```torch.nn.functional.adaptive_avg_pool3d``` between CPU and GPU using a bfloat16 tensor.
```python 
import torch

input_tensor = torch.tensor([
    [
        [
            [-1.4062, 1.4609, 0.6797],
            [-0.6875, -0.9492, 0.4434],
            [-1.0312, -0.3730, 0.9453]
        ],
        [
            [0.9766, 0.2070, 0.8242],
            [-1.6484, 1.4531, 1.7891],
            [0.3945, 0.5352, -0.8711]
        ]
    ],
    [
        [
            [-2.5000, 0.2617, -0.3613],
            [-1.6094, -1.4219, -0.3281],
            [-1.3594, -2.3594, -0.5312]
        ],
        [
            [-1.9375, 1.0938, 1.5547],
            [-0.5820, -0.1167, 1.3438],
            [1.1953, -1.3750, -1.3438]
        ]
    ]
], dtype=torch.bfloat16)

output_size = (None, 1, None)

result_cpu = torch.nn.functional.adaptive_avg_pool3d(input_tensor, output_size)

input_cuda = input_tensor.cuda()
result_gpu = torch.nn.functional.adaptive_avg_pool3d(input_cuda, output_size)

print("CPU result:\n", result_cpu)
print("GPU result:\n", result_gpu.cpu())

inconsistent = not torch.allclose(result_cpu, result_gpu.cpu(), atol=1e-02, rtol=1e-03)
print("Inconsistency with atol=1e-02 and rtol=1e-03:", inconsistent)
```
outputs:
```
CPU result:
 tensor([[[[-1.0391,  0.0461,  0.6875]],

         [[-0.0923,  0.7305,  0.5781]]],


        [[[-1.8359, -1.1719, -0.4062]],

         [[-0.4395, -0.1328,  0.5195]]]], dtype=torch.bfloat16)

GPU result:
 tensor([[[[-1.0391,  0.0461,  0.6875]],

         [[-0.0923,  0.7305,  0.5820]]],


        [[[-1.8203, -1.1719, -0.4062]],

         [[-0.4414, -0.1328,  0.5195]]]], dtype=torch.bfloat16)

Inconsistency with atol=1e-02 and rtol=1e-03: True
```

### Versions

(executed on google colab)
PyTorch version: 2.5.0+cu121
Is debug build: False
CUDA used to build PyTorch: 12.1

OS: Ubuntu 22.04.3 LTS (x86_64)
GCC version: (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0
Clang version: 14.0.0-1ubuntu1.1
CMake version: version 3.30.5
Libc version: glibc-2.35

Python version: 3.10.12 (main, Sep 11 2024, 15:47:36) [GCC 11.4.0] (64-bit runtime)
Python platform: Linux-6.1.85+-x86_64-with-glibc2.35
Is CUDA available: True
CUDA runtime version: 12.2.140
CUDA_MODULE_LOADING set to: LAZY
GPU models and configuration: GPU 0: Tesla T4
Nvidia driver version: 535.104.05
cuDNN version: Probably one of the following:
/usr/lib/x86_64-linux-gnu/libcudnn.so.8.9.6
/usr/lib/x86_64-linux-gnu/libcudnn_adv_infer.so.8.9.6
/usr/lib/x86_64-linux-gnu/libcudnn_adv_train.so.8.9.6
/usr/lib/x86_64-linux-gnu/libcudnn_cnn_infer.so.8.9.6
/usr/lib/x86_64-linux-gnu/libcudnn_cnn_train.so.8.9.6
/usr/lib/x86_64-linux-gnu/libcudnn_ops_infer.so.8.9.6
/usr/lib/x86_64-linux-gnu/libcudnn_ops_train.so.8.9.6
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Address sizes:                        46 bits physical, 48 bits virtual
CPU(s):                               2
On-line CPU(s) list:                  0,1
Model name:                           Intel(R) Xeon(R) CPU @ 2.00GHz
Model:                                85
Core(s) per socket:                   1
Stepping:                             3
BogoMIPS:                             4000.31
Flags:                                fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush mmx fxsr sse sse2 ss ht syscall nx pdpe1gb rdtscp lm constant_tsc rep_good nopl xtopology nonstop_tsc cpuid tsc_known_freq pni pclmulqdq ssse3 fma cx16 pcid sse4_1 sse4_2 x2apic movbe popcnt aes xsave avx f16c rdrand hypervisor lahf_lm abm 3dnowprefetch invpcid_single ssbd ibrs ibpb stibp fsgsbase tsc_adjust bmi1 hle avx2 smep bmi2 erms invpcid rtm mpx avx512f avx512dq rdseed adx smap clflushopt clwb avx512cd avx512bw avx512vl xsaveopt xsavec xgetbv1 xsaves arat md_clear arch_capabilities
Hypervisor vendor:                    KVM
Virtualization type:                  full
L1d cache:                            32 KiB (1 instance)
L1i cache:                            32 KiB (1 instance)
L2 cache:                             1 MiB (1 instance)
L3 cache:                             38.5 MiB (1 instance)
NUMA node0 CPU(s):                    0,1
Vulnerability Gather data sampling:   Not affected
Vulnerability Itlb multihit:          Not affected
Vulnerability L1tf:                   Mitigation; PTE Inversion
Vulnerability Mds:                    Vulnerable; SMT Host state unknown
Vulnerability Meltdown:               Vulnerable
Vulnerability Mmio stale data:        Vulnerable
Vulnerability Reg file data sampling: Not affected
Vulnerability Retbleed:               Vulnerable
Vulnerability Spec store bypass:      Vulnerable
Vulnerability Spectre v1:             Vulnerable: __user pointer sanitization and usercopy barriers only; no swapgs barriers
Vulnerability Spectre v2:             Vulnerable; IBPB: disabled; STIBP: disabled; PBRSB-eIBRS: Not affected; BHI: Vulnerable (Syscall hardening enabled)
Vulnerability Srbds:                  Not affected
Vulnerability Tsx async abort:        Vulnerable

[pip3] numpy==1.26.4
[pip3] nvidia-cublas-cu12==12.6.3.3
[pip3] nvidia-cuda-cupti-cu12==12.6.80
[pip3] nvidia-cuda-runtime-cu12==12.6.77
[pip3] nvidia-cudnn-cu12==9.5.1.17
[pip3] nvidia-cufft-cu12==11.3.0.4
[pip3] nvidia-curand-cu12==10.3.7.77
[pip3] nvidia-cusolver-cu12==11.7.1.2
[pip3] nvidia-cusparse-cu12==12.5.4.2
[pip3] nvidia-nccl-cu12==2.23.4
[pip3] nvidia-nvjitlink-cu12==12.6.77
[pip3] nvtx==0.2.10
[pip3] optree==0.13.0
[pip3] pynvjitlink-cu12==0.4.0
[pip3] torch==2.5.0+cu121
[pip3] torchaudio==2.5.0+cu121
[pip3] torchsummary==1.5.1
[pip3] torchvision==0.20.0+cu121
[conda] Could not collect

vishalgoyal316:
Use higher precision accumulator (opmath_type) for sum computation

in adaptive_avg_pool3d CPU implementation to fix numerical errors and CPU-GPU inconsistencies with BFloat16/Half dtypes.

- Accumulate in float precision instead of BFloat16/Half
- Compute divide_factor as single multiplication vs sequential divisions
- Add test for CPU-GPU parity with reduced precision

Fixes #140696


test/nn/test_pooling.py
```diff
@@ -1011,6 +1011,49 @@ def test_adaptive_avg_pool3d_output_size_one(self, device):
         c = out.size(1)
         self.assertEqual(out.stride(), [c, 1, 1, 1, 1])
 
+    @onlyCUDA
```
Skylion007: Hmm, why only cuda, why not just skip CPU? Good to test RoCM, MPS, and XPU too, right?

vishalgoyal316: Thanks for your comments. You are right, I have added test support for all GPUs.
Please review and let me know if you have any other suggestions.

test/nn/test_pooling.py
```diff
@@ -1011,6 +1011,57 @@ def test_adaptive_avg_pool3d_output_size_one(self, device):
...
+        not (
+            TEST_CUDA
+            or torch.backends.mps.is_available()
+            or torch.xpu.is_available()
+            or (hasattr(torch, "hpu") and torch.hpu.is_available())
+        ),
+        "Requires at least one accelerator device",
+    )
+    @dtypes(torch.bfloat16, torch.half)
+    def test_adaptive_avg_pool3d_cpu_gpu_parity(self, device, dtype):
```
Skylion007: How come the nn.functional.adaptive_avg_pool3d test on `torch/testing/_internal/common_methods_invocations.py` doesn't catch this? Is the tolerance too high?

test/nn/test_pooling.py
```diff
@@ -1011,6 +1011,57 @@ def test_adaptive_avg_pool3d_output_size_one(self, device):
...
 
+    @unittest.skipIf(
+        not (
+            TEST_CUDA
+            or torch.backends.mps.is_available()
+            or torch.xpu.is_available()
+            or (hasattr(torch, "hpu") and torch.hpu.is_available())
+        ),
+        "Requires at least one accelerator device",
+    )
```
Skylion007: Ugh, this is ugly. Can we just run it on CPU? It's a useless test, but it shouldn't fail atleast...
```suggestion
```
Otherwise, there has to be some testing macro we can import.

test/nn/test_pooling.py
```diff
@@ -1011,6 +1011,57 @@ def test_adaptive_avg_pool3d_output_size_one(self, device):
...
+            ],
+            dtype=dtype,
+        )
+
+        output_size = (None, 1, None)
+
+        result_cpu = F.adaptive_avg_pool3d(input_tensor.cpu(), output_size)
+        result_accelerator = F.adaptive_avg_pool3d(input_tensor.to(device), output_size)
+
+        self.assertEqual(result_cpu, result_accelerator.cpu(), atol=1e-2, rtol=1e-3)
```
Skylion007: should this call assert close?

test/nn/test_pooling.py
vishalgoyal316: Right, now, I have modified the test case for CPU only, please check and let me know your feedback.

test/nn/test_pooling.py
vishalgoyal316: The tolerance contributes to the issue, but the primary reason is that randomly generated inputs don't reliably expose this particular precision accumulation edge case.

test/nn/test_pooling.py
```diff
@@ -1011,6 +1011,53 @@ def test_adaptive_avg_pool3d_output_size_one(self, device):
         c = out.size(1)
         self.assertEqual(out.stride(), [c, 1, 1, 1, 1])
 
+    @onlyCPU
+    @dtypes(torch.bfloat16, torch.half)
+    def test_adaptive_avg_pool3d_reduced_precision(self, device, dtype):
+        # test for reduced precision types (BFloat16/Half) on CPU
+        # this test ensures that the fix for using higher precision accumulation
+        # produces correct results by comparing against float32 reference
```
Skylion007, line 1031: Also add a link to the issue number it is resolving.

Skylion007 (dismissed):

Skylion007:
Does this test fail on CUDA? Why not just enable it everywhere though to test other kernels too

test/nn/test_pooling.py
```diff
@@ -1011,6 +1011,53 @@ def test_adaptive_avg_pool3d_output_size_one(self, device):
         c = out.size(1)
         self.assertEqual(out.stride(), [c, 1, 1, 1, 1])
 
+    @onlyCPU
+    @dtypes(torch.bfloat16, torch.half)
+    def test_adaptive_avg_pool3d_reduced_precision(self, device, dtype):
```
Skylion007, line 1027: Why not test this on all devices which support these dtypes?

Skylion007:
Ah found it: https://github.com/pytorch/pytorch/pull/136116

test/nn/test_pooling.py
vishalgoyal316: Yes, Done!

test/nn/test_pooling.py
vishalgoyal316: Done!

vishalgoyal316:
> This is also partially my fault lol: #136091

I have taken care this failure in this PR. Thanks!

Skylion007 (dismissed):

Skylion007:
Does the MPS kernel have the same bug?

vishalgoyal316:
> Does the MPS kernel have the same bug?

@Skylion007, you are right, I have fixed it now.

vishalgoyal316:
## Updated with MPS precision fix

The MPS test failures were due to a real precision bug in the backward pass. 
I've fixed it by using `opmath_t<T>` for division (matching CPU and forward pass).

**Changes:**
- `aten/src/ATen/native/mps/kernels/Pooling.metal`: Use high precision for backward division
- `test/nn/test_pooling.py`: Removed `@expectedFailureMPS` - test should now pass

**Note:** The `test_pynode_destruction_deadlock` failure is unrelated to this PR 
(autograd test, not pooling). This is a known flaky test.

Skylion007 (dismissed):

vishalgoyal316:
@Skylion007, after passing all checks, is this PR ready to merge?

aten/src/ATen/native/AdaptiveAveragePooling3d.cpp
```diff
@@ -63,19 +65,20 @@ void adaptive_avg_pool3d_out_frame(
...
             /* compute local average: */
-            scalar_t sum = 0;
+            accscalar_t sum = accscalar_t(0);
             for (const auto it : c10::irange(kT)) {
               for (const auto ih : c10::irange(kH)) {
                 for (const auto iw : c10::irange(kW)) {
                   scalar_t val =
                       *(ip + it * istrideT + ih * istrideH + iw * istrideW);
-                  sum += val;
+                  sum += accscalar_t(val);
```
Skylion007, line 74: Actually not sure this cast is necessary if it supports an overloaded. Probably fine, but don't want to override this in CUDA as some float type may have specific overloads for reduced precision dtypes using hardware available ops.

If this is just to silence warnings/errors probably fine and a super minor nit.

With as it is now, isn't this exactly equivalent doing the entire op in high precision and then downcasting?

aten/src/ATen/native/AveragePool3d.cpp
```diff
@@ -396,15 +399,16 @@ void avg_pool3d_backward_out_frame(
             }
 
             /* scatter gradients out to footprint: */
-            scalar_t val  = *op++;
+            accscalar_t val  = static_cast<accscalar_t>(*op++);
+            accscalar_t val_scaled = val / divide_factor;
```
Skylion007, line 403: This value is const and applied to all values in the for loop below, should be statically cast once to scalar_t and applied. Casting in the loop is inefficient and risks the compiler not being able to hoist the cast out of the loop.

aten/src/ATen/native/AveragePool3d.cpp
```diff
@@ -222,7 +224,7 @@ void avg_pool3d_out_frame(
             }
 
             /* compute local sum: */
-            scalar_t sum = 0.0;
+            accscalar_t sum = 0.0;
```
Skylion007, line 227: By contrast, here a static cast isn't necessary for the arg. IE: this op looks fine as is.

Skylion007 (commented):
Finally remaining performance nits

aten/src/ATen/native/AdaptiveAveragePooling3d.cpp
```diff
@@ -63,19 +65,20 @@ void adaptive_avg_pool3d_out_frame(
...
                 }
               }
             }
 
             /* set output to local average */
-            *op = sum / kT / kH / kW;
+            const accscalar_t divide_factor = accscalar_t(kT * kH * kW);
```
Skylion007, line 80: Likewise is this cast necessary? Shouldn't these remain nits? Or is linting complaining?

Skylion007 (commented):
Okay, now I think we have some unnecessary casts and things that can be improved for performance here.

vishalgoyal316:
> Okay, now I think we have some unnecessary casts and things that can be improved for performance here.

@Skylion007, you are right, and thanks for pointing out such improvements, I have pushed the code accordingly, please let me know if anything else need to correct/update.

aten/src/ATen/native/AveragePool3d.cpp
```diff
@@ -396,15 +399,16 @@ void avg_pool3d_backward_out_frame(
             }
 
             /* scatter gradients out to footprint: */
-            scalar_t val  = *op++;
+            accscalar_t val  = static_cast<accscalar_t>(*op++);
+            scalar_t val_scaled = static_cast<scalar_t>(val / divide_factor);
```
Skylion007:
```suggestion
            const accscalar_t val  = static_cast<accscalar_t>(*op++);
            const scalar_t val_scaled = static_cast<scalar_t>(val / divide_factor);
```
This upcasting of the division here is really needed to enable low precision support here? Fails without it?

vishalgoyal316: No fail without this. Done now!

Skylion007 (dismissed):

test/nn/test_pooling.py
```diff
@@ -1011,6 +1011,118 @@ def test_adaptive_avg_pool3d_output_size_one(self, device):
...
+        result_reduced = F.adaptive_avg_pool3d(
+            input_tensor.to(dtype).to(device), output_size
+        )
+
+        # compare reduced precision result against float32 reference
+        torch.testing.assert_close(
+            result_reduced.float(), result_fp32, atol=1e-2, rtol=1e-3
+        )
+
+    @onlyNativeDeviceTypes
```
Skylion007, line 1079: Is this test also failing on MPS?

test/nn/test_pooling.py
```diff
@@ -1011,6 +1011,118 @@ def test_adaptive_avg_pool3d_output_size_one(self, device):
...
+        # regression test for https://github.com/pytorch/pytorch/issues/140696
+        # test for reduced precision types (BFloat16/Half)
+        # test ensures that the fix for using higher precision accumulation
+        # produces correct results by comparing against float32 reference
+
+        input_tensor = torch.tensor(
+            [
+                [
+                    [
+                        [-1.4062, 1.4609, 0.6797],
```
malfet, line 1037: Why do you hardcode values there rather than use `torch.rand`/`torch.randn`/`torch.uniform`?

test/nn/test_pooling.py
```diff
@@ -1011,6 +1011,118 @@ def test_adaptive_avg_pool3d_output_size_one(self, device):
...
+        # related to https://github.com/pytorch/pytorch/issues/140696
+        # test for reduced precision types (BFloat16/Half)
+        # test ensures that the avg_pool3d path uses higher precision
+        # accumulation and produces correct results
+
+        input_tensor = torch.tensor(
+            [
+                [
+                    [
+                        [-1.4062, 1.4609, 0.6797],
```
malfet, line 1091: Once again, what so specific about those values?

test/nn/test_pooling.py
malfet: +1 on the above, why this decorator is needed here?

aten/src/ATen/native/cuda/AdaptiveAveragePooling3d.cu
```diff
@@ -106,7 +106,7 @@ __global__ void adaptiveaveragepool(
                                    (istartT + it) * istrideT +
                                    (istartH + ih) * istrideH + (istartW + iw) * istrideW;
             scalar_t val = input[input_offset];
-            sum += static_cast<accscalar_t>(val);
```
malfet, line 109: This change is benign, but why is it needed now?

aten/src/ATen/native/AveragePool3d.cpp
```diff
@@ -396,15 +398,15 @@ void avg_pool3d_backward_out_frame(
...
+            const scalar_t val_scaled = *op++ / divide_factor;
 
             for (auto z = tstart; z < tend; z++)
             {
               for (auto y = hstart; y < hend; y++)
               {
                 for (auto x = wstart; x < wend; x++)
                 {
-                  *(ip + z * iheight * iwidth + y * iwidth + x) += val / divide_factor;
+                  *(ip + z * iheight * iwidth + y * iwidth + x) += val_scaled;
```
malfet: This is a nice change, but have nothing to do with precision improvements, is it?

malfet (changes requested):
This PR tries to do many things at once, do you mind breaking it down into smaller changes? (For example small BE changes that precomputes divide factor, another one that extends types, and finally the one that adds new test?

aten/src/ATen/native/AveragePool3d.cpp
vishalgoyal316: @malfet, absolutely right.

test/nn/test_pooling.py
vishalgoyal316: The decorator ensures the test only runs on devices that actually compute values, so it can verify the precision fix works correctly.

test/nn/test_pooling.py
vishalgoyal316: These specific values likely triggered the original bug reported in issue #140696. They represent a real-world case where precision mattered.

test/nn/test_pooling.py
vishalgoyal316: These specific values likely triggered the original bug reported in issue #140696. They represent a real-world case where precision mattered.

PR #165538, opened 2025-10-15
[ROCm][Windows] Enable AOTriton runtime compile on Windows

tvukovic-amd:
AOTriton uses prebuilt runtime binaries if the user's ROCm version matches the ones used to generate the prebuilt runtime. However, since there's no prebuilt runtime available for Windows, this check needs to be bypassed for Windows. This PR enables it by changing condition to always build AOTriton runtime from source on Windows.

cc @peterjc123 @mszhanyi @skyline75489 @nbcsm @iremyux @Blackhex @jeffdaily @sunway513 @jithunnair-amd @pruthvistony @ROCmSupport @dllehr-amd @jataylo @hongxiayang @naromero77amd


xinyazhang (approved):
Although the ideal solution is matching the `manylinux_2_28_x86_64` substring in the package name or using a catalog file signed with sigstore, but it is good enough for now.

jeffdaily (approved):

PR #165539, opened 2025-10-15
In pipeline parallelism: Use same dtype for receive and send tensor when initializing p2p communication.

BlueCrescent:
When initializing the p2p communication for pipeline parallelism, currently different default dtypes are used for the send and receive tensor here:
https://github.com/pytorch/pytorch/blob/5c583e2573f29243742e00b9fa36b266c5c78bb3/torch/distributed/pipelining/stage.py#L935-L936

This caused hard to trace issues when training on multiple nodes. Multiple stages on one node seem to work for some reason which probably caused the unit tests not to catch this.

Fixes #165143


H-Huang (approved):
Thank you for root causing and fixing. Send/recv with mismatched dtypes indeed causes undefined behavior so we didn't see this in our runs. Really appreciate your investigation!

PR #165540, opened 2025-10-15
[dynamo] Trace datetime.now()

jmaczan:
Introduces DatetimeVariable based on RandomVariable. Only `now()` is supported, timestap is generated in DatetimeVariable constructor and when `now` is accessed on `datetime.datetime()`, stored timestamp is returned as `UserDefinedObjectVariable` (couldn't use ConstantVariable). No graph breaks, etc. The PR is not finished, but I'd like to know your comments @williamwen42 @Lucaskabela 

One important thing is that I didn't commit changes to `SideEffects.codegen_update_mutated`, because I'm not sure if we need it at all and if we do, then what for?
https://github.com/pytorch/pytorch/blob/f58f301313d4fc89499fb35cdfb2ffb91d14d896/torch/_dynamo/side_effects.py#L1157

Fixes #164922

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @kadeng @chauhang @amjames @jataylo @azahed98 @mlazos @Lucaskabela @chenyang78


torch/_dynamo/variables/misc.py
```diff
@@ -1958,3 +1959,79 @@ def reconstruct(self, codegen: "PyCodegen"):
...
+    }
+
+    def __init__(
+        self,
+        dt: Optional[datetime.datetime] = None,
+        **kwargs,
+    ) -> None:
+        super().__init__(**kwargs)
+        # maybe don't instantiate and do it only when now() is called? not sure
+        self.now = datetime.datetime.now()
```
Lucaskabela: One thing I am struggling to understand here is that since we save the time here, it will save this time at trace time, and not when the function is called.

As a toy example, let's take this:

```
import torch

def fn():
      # Example where someone may want to record time it takes 
      start = datetime.datetime.now()

      # Do expensive stuff - call matmul on two tensors 100,000 times
      for _ in range(int(1e6)):
          a, b = torch.randn((20, 20)), torch.randn((20, 20))
          res = a * b
      end = datetime.datetime.now()
      return end - start
```

My expectation is `end - start` > 1s; however, given the current implementation, I am not sure this will be the case

I think we could turn this into a unit test to guide development here
cc @williamwen42

Lucaskabela: As stated, my intuition is this current impl (saving .now) will serve to measure our tracing time as opposed to anything related to when the fn is called so we should probably move this to call

Lucaskabela (changes requested):

torch/_dynamo/variables/misc.py
williamwen42: I'm doubtful that we can preserve accurate timing. I think at minimum for now, we should support:
- datetime.now() is called for every invocation of the compiled function
- The output of datetime.now() is treated like a symbolic integer

mlazos:
Can you share more information about the desired use case? What do we then use the output of this function for? Mainly, I'm curious if the value will be used in the graph or not.

torch/_dynamo/variables/misc.py
jmaczan: Makes sense, I'm reworking it

jmaczan:
@mlazos Right now guards on `now()` fail just right after they're created and I'm preventing that in the first place. As far as I understand, we want to use it in graph, without constant folding, so fresh timestamp is generated every time compiled function is called, but at the same time it has to either not be guarded or return the same timestamp within a frame. I hope that makes sense? I'm still learning so I might have blind spots here, pls point out if that's the case

torch/_dynamo/variables/misc.py
```diff
@@ -1958,3 +1959,48 @@ def reconstruct(self, codegen: "PyCodegen"):
...
+    def reconstruct(self, cg):
+        cg.create_load_python_module(datetime)
+        cg.create_load_attr("datetime")
+        cg.create_load_attr("now")
+        cg.call_function(0, False)
+
+    def var_getattr(self, tx, name: str) -> VariableTracker:
+        # TODO: for now supporting only attributes; operations on now() won't work yet
+        if name in _supported_datetime_attrs:
+            vt = tx.output.synthetic_graph_input(
```
jmaczan: Gives us new value on each run @Lucaskabela @williamwen42 @mlazos

torch/_dynamo/variables/misc.py
```diff
@@ -1958,3 +1959,42 @@ def reconstruct(self, codegen: "PyCodegen"):
...
+    "hour",
+    "minute",
+    "second",
+    "microsecond",
+}
+
+
+class DatetimeNowVariable(VariableTracker):
+    def var_getattr(self, tx, name: str) -> VariableTracker:
+        # TODO: for now supporting only attributes; operations on now() won't work yet
```
jmaczan: It's the question of how far we want to go with support for `datetime`. Do we want to cover scenarios like `tensor * datetime` or are we good with just letting users to use attributes of `datetime` (year, second, etc)?

torch/_dynamo/variables/builtin.py
```diff
@@ -1261,6 +1261,40 @@ def _handle_insert_op_in_graph(self, tx: "InstructionTranslator", args, kwargs):
...
+                        False,
+                        False,
+                        None,
+                    ),
+                    {},
+                )
+
+                proxy = tx.output.create_proxy(
+                    "call_function",
+                    torch.ops.aten.add.Tensor,  # TODO: add-only for now
```
jmaczan: It's in progress, currently add is only supported binary op and seem to work fine, but I'm going to improve it

torch/_dynamo/variables/misc.py
```diff
@@ -1958,3 +1959,42 @@ def reconstruct(self, codegen: "PyCodegen"):
...
+        if name in _supported_datetime_attrs:
+            vt = tx.output.synthetic_graph_input(
+                _datetime_attr_tensor, (name, torch.get_default_dtype())
```
jmaczan: How to ensure that our new tensor wrapper for int is on the same device and with the same dtype as a tensor that will it be used with (added to, etc)?

torch/_dynamo/variables/misc.py
```diff
@@ -1958,3 +1959,42 @@ def reconstruct(self, codegen: "PyCodegen"):
...
         codegen(self.callback_vt)
         codegen.extend_output(create_call_function(2, False))
+
+
+class DatetimeClassVariable(VariableTracker):
+    def python_type(self):
+        return type(datetime.datetime)
+
+    def call_method(self, tx, name, args, kwargs):
+        if name == "now" and not args and not kwargs:
```
jmaczan, line 2123: only `now()` supported for now, there are a few other methods like `utcnow()` that we might want to support, too

jmaczan (commented):
@Lucaskabela @mlazos @williamwen42 
I have reworked it, so it uses fresh now() on each run. There still few things to do:
- support more ops than just `add`
- support more datetime constructors than just `now()`
- figure out how to ensure that tensor wrapper for datetime attribute (second, year, etc - all are ints) has the same dtype and device as a tensor that we will use it with
- if it's possible, maybe we shouldn't wrap it with tensor, but maybe with `SymInt` or something?

I'd like to know your opinion about the approach here - is it ok to you and I can continue or would you like to fundamentally change something?

jmaczan:
@williamwen42 Yes, I'm rewriting it now to not use tensors and will add tests as well

test/dynamo/test_misc.py
```diff
@@ -13286,6 +13287,208 @@ class C:
...
+        @torch.compile(backend=counter, fullgraph=True)
+        def fn(x):
+            t = datetime.datetime.now()
+            return x + t.second
+
+        x = torch.randn(4)
+
+        # multiple runs shouldn't trigger recompile
+        out1 = fn(x)
+        out2 = fn(x)
```
williamwen42, line 13302: nit: you can use `torch.compiler.set_stance("fail_on_recompile")` on the second call to error on recompiles

test/dynamo/test_misc.py
```diff
@@ -13286,6 +13287,208 @@ class C:
...
+
+        dt_now = misc.datetime_now
+        calls_counter = 0
+
+        def count_now_calls():
+            nonlocal calls_counter
+            calls_counter += 1
+            return dt_now()
+
+        misc.datetime_now = count_now_calls
```
williamwen42, line 13466: nit: use `unittest.mock.patch`?

test/dynamo/test_misc.py
```diff
@@ -13286,6 +13287,208 @@ class C:
...
+        counter = CompileCounter()
+
+        @torch.compile(backend=counter)
+        def fn(x):
+            start = datetime.datetime.now()
+            y = x + 1
+            end = datetime.datetime.now()
+            elapsed = end - start
+            s = elapsed.total_seconds()
+            if s > 0.0:
```
williamwen42, line 13410: Is it possible to graph break here and resume instead of skipping the frame? That logic should be here: https://github.com/pytorch/pytorch/blob/1328a02d2eb26ee67346048ee242327ea90d6315/torch/_dynamo/symbolic_convert.py#L648

williamwen42 (commented):
My main worry here is that we're introducing nodes in the FX graph that downstream backends may not be able to handle.

The `random.random` approach uses sources to access `random.random()` calls at bytecode runtime and uses `wrap_unspecialized_primitive` to bake the value into the graph. The downside of following this approach is that all `datetime.now()` will be called before the graph is called, leading to incorrect time measurements.

jmaczan:
Chill out @github-actions, I will come back to this PR soon!

PR #165542, opened 2025-10-15
[DeviceMesh] Fix layout calculation when flattening non-contiguous dims

lw:
* #165556
* #165555
* #165554
* __->__ #165542

### DIFF
--- a/torch/distributed/device_mesh.py
+++ b/torch/distributed/device_mesh.py
@@ -217,7 +217,7 @@ else:
                 "Please use a non-overlapping layout when creating a DeviceMesh."
             )
             # Because we still need to support slicing of flattened dim from root mesh, so we don't check stride here.
-            assert self._layout.numel() == self.mesh.numel(), (
+            assert self._layout.top_level_sizes == self.mesh.size(), (
                 "Please use a valid layout when creating a DeviceMesh."
                 f"The layout {self._layout} is not consistent with the mesh size {self.mesh.size()}."
             )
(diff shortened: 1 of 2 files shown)


torch/distributed/device_mesh.py
```diff
@@ -217,7 +217,7 @@ def __init__(
                 "Please use a non-overlapping layout when creating a DeviceMesh."
             )
             # Because we still need to support slicing of flattened dim from root mesh, so we don't check stride here.
-            assert self._layout.numel() == self.mesh.numel(), (
+            assert self._layout.top_level_sizes == self.mesh.size(), (
```
lw, line 220: Once I made this check somewhat stricter, the TestDeviceMeshGetItem.test_flatten_mesh_3d testcase started failing, because of a mismatch in ndim.

torch/distributed/device_mesh.py
```diff
@@ -672,6 +672,8 @@ def _create_flatten_mesh(
                 )
 
             flattened_mesh_layout = self._layout.coalesce()
```
lw, line 674: The result of coalesce is an int if the layout is contiguous, but it's a tuple of 2+ elements otherwise. However, this corresponds to _one_ DeviceMesh dim, so we must nest it in order for the top-level tuple to be of size 1.

fduwjj: Well since this is a flatten, it makes sense that it is one DeviceMesh dim? Do you have an actual case where the current code fails?

lw: See TestDeviceMeshGetItem.test_flatten_mesh_3d

ezyang (approved):

fduwjj (approved):

PR #165545, opened 2025-10-15
Fix nn.Dropout accuracy discrepancies between triton and torch implementations

EricKing626:
We're proposing a PyTorch PR to address Torch Compile vs. Eager inconsistencies. The primary concern we’ve heard is accuracy divergence under Torch Compile (Triton). We targeted a concrete source of mismatch— nn.Dropout—and implemented an approach that:
1. Fuses Dropout with adjacent pointwise kernels (preserving performance), and
2. Preserves exact results with Eager mode (no accuracy regressions).

This PR introduces a compiler switch, adding torch._inductor.config.align_random_eager = True, that reproduces GPU-eager Philox consumption so dropout masks align with eager without disabling Triton fusion. 
In contrast, torch._inductor.config.fallback_random = True routes RNG ops back to eager and typically prevents fusing dropout with surrounding elementwise code, which can hurt performance. Our flag only affects RNG alignment; non-random kernels are unchanged. In practice we see parity with eager RNG and slightly better throughput than a vanilla torch.compile run where RNG may diverge from eager. Verified on MI325/MI350/MI355/H100/H200.

The dropout mask can be aligned with eager mode in FP32, FP16, BF16
image

Performance tested in MI325, each batch size is 8. align means our PR (align_random_eager), fallback means fallback_random
vit_b_16 baseline 6.936 ms | align 6.954 ms | fallback 7.013 ms | align/fallback 1.008x
swin_v2_t baseline 12.734 ms | align 12.773 ms | fallback 12.797 ms | align/fallback 1.002x
inception_v3 baseline 16.905 ms | align 16.903 ms | fallback 16.951 ms | align/fallback 1.003x

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @mlazos @chenyang78


torch/_inductor/config.py
```diff
@@ -644,6 +644,9 @@ def use_autoheuristic(name: str) -> bool:
 # fallback to eager for random/dropout, this is slow but useful for debugging
 fallback_random = False
 
+# align random/dropout as eager mode(aten) behavior, maintaining fused possibility and faster gpu kernel
+align_random_eager = False
```
eellison, line 776: can you turn this on, for the purposes of seeing what fails ? Also, any reason for this to not be on by default ?

torch/_inductor/lowering.py
```diff
@@ -2432,8 +2432,26 @@ def inner_fn(_):
...
 
+def get_seed_and_threads_per_round(x_device_index: int, nelem: int):
+    UNROLL = 4
+    prop = torch.cuda.get_device_properties(x_device_index)
+    threads_per_round = (
+        prop.multi_processor_count * prop.max_threads_per_multi_processor
+    )
+    gen = torch.cuda.default_generators[x_device_index]
+    seed = int(gen.initial_seed())
+    return seed, threads_per_round
```
eellison: Doesn't initial seed need to be a runtime arg ?

torch/_inductor/fx_passes/replace_random.py
```diff
@@ -20,6 +20,70 @@
 patterns = PatternMatcherPass(subsystem="joint_graph_passes")
 aten = torch.ops.aten
 
+from torch.library import custom_op
+
+
+# ---- host helpers ----
+def _compute_grid_x(nelem: int, block: int, device_index: int) -> int:
```
eellison: unused?

test/inductor/test_dropout_align_random_eager.py
```diff
@@ -0,0 +1,252 @@
...
+torch._inductor.config.freezing = True
+torch._inductor.config.align_random_eager = True
+# torch._inductor.config.fallback_random = True
+# torch._inductor.config.max_autotune_pointwise = True
+
+
+# ───────────────────────────────────────────────────────────────
+# Model under test
+# ───────────────────────────────────────────────────────────────
+class LinearBlock(torch.nn.Module):
```
eellison, line 35: Can we have a test that invokes multiple dropouts ? and does multiple iterations ?

test/inductor/test_dropout_align_random_eager.py
```diff
@@ -0,0 +1,252 @@
...
+    seed = struct.unpack("<Q", st[0:8].cpu().numpy().tobytes())[0]
+    off = struct.unpack("<Q", st[8:24].cpu().numpy().tobytes())[0]
+    return seed, off
+
+
+# ───────────────────────────────────────────────────────────────
+# Tests: correctness (forward/backward)
+# ───────────────────────────────────────────────────────────────
+@pytest.mark.parametrize("training", [False, True])
+def test_linear_block_compile_parity_forward(device, input_tensor, training):
```
eellison: Can you please copy the test inductor idioms for test classes - see:

https://github.com/pytorch/pytorch/blob/aa4a8c9b9228a8999cac1fba37cd04e086ecfb9b/test/inductor/test_control_deps.py#L2-L16

test/inductor/test_dropout_align_random_eager.py
```diff
@@ -0,0 +1,252 @@
...
+# torch._inductor.config.fallback_random = True
+# torch._inductor.config.max_autotune_pointwise = True
+
+
+# ───────────────────────────────────────────────────────────────
+# Model under test
+# ───────────────────────────────────────────────────────────────
+class LinearBlock(torch.nn.Module):
+    def __init__(self, hidden_dim: int, ffn_dim: int, dropout: float = DROPOUT_P):
+        super().__init__()
```
eellison, line 37: Can we also get tests for :

a) dynamic shapes (you can use mark_dynamic or dynamic=True)
b) cudagraphs - you can invoke with mode='reduce-overhead'

torch/_inductor/fx_passes/replace_random.py
```diff
@@ -20,6 +20,70 @@
...
+    dev_index = device.index if isinstance(device, torch.device) else int(device)
+    gen = torch.cuda.default_generators[dev_index]
+    old_off = int(gen.get_offset())
+    gen.set_offset(old_off + used_offset)
+    return old_off // 4
+
+
+@custom_op("custom_op::rand_eager_offset", mutates_args={})
+def rand_eager_offset(offset: int, device: torch.device) -> torch.Tensor:
+    base = _reserve_offset(device, int(offset))
```
eellison: The `int` here will not work with symints.

eellison (commented):
Some questions from me. also cc @v0i0 re: inductor/eager numerics.

torch/_inductor/fx_passes/replace_random.py
```diff
@@ -20,6 +20,70 @@
...
+    )
+    used_offset = rounds_per_thread * UNROLL
+    return used_offset
+
+
+def _reserve_offset(device: torch.device, used_offset: int) -> int:
+    dev_index = device.index if isinstance(device, torch.device) else int(device)
+    gen = torch.cuda.default_generators[dev_index]
+    old_off = int(gen.get_offset())
+    gen.set_offset(old_off + used_offset)
```
eellison: And I don't think set_offset will work with cudagraph

torch/_inductor/lowering.py
EricKing626: Good point — right now I'm baking initial_seed() at lowering time, so it won't see any runtime changes to the RNG (e.g. manual_seed or a user-provided generator). I'll refactor this so the seed is passed in as a runtime argument from the call site and only threads_per_round is derived from the device properties here.

torch/_inductor/fx_passes/replace_random.py
EricKing626: I got your point. But is there any better place to reserve and push the RNG offset for supporting cudagraph in your opinion?

torch/_inductor/lowering.py
EricKing626: > Doesn't initial seed need to be a runtime arg ?

@eellison Solved

test/inductor/test_dropout_align_random_eager.py
```diff
@@ -0,0 +1,355 @@
...
+        compiled.eval()
+
+        # warm up
+        _timed_run(eager, x, backward=False)
+        _timed_run(compiled, x, backward=False)
+
+        t_eager, _ = _timed_run(eager, x, backward=False)
+        t_comp, _ = _timed_run(compiled, x, backward=False)
+
+        assert t_comp > 0 and t_eager > 0
```
v0i0: could you also test against torch primitive random generation fns, that should work now, too, right? i.e. randint/rand/randn? if some don't work, xfail it so we know where we have future work

EricKing626: @v0i0 Sure, I haven't done randn. I could add this in test and set as xfail if not work

test/inductor/test_dropout_align_random_eager.py
EricKing626: Dynamic_shape CI tests can be pass but I found it doesn't update new dropout when inter-step training because set_offset is host API and require concrete integer. So, I turn off it when dynamic_shape.

Dynamic shape related:
python test/inductor/test_fused_attention.py SDPAPatternRewriterGpuDynamicTests.test_sdpa_rewriter_12_gpu
python test/inductor/test_aot_inductor.py AOTInductorTestABICompatibleGpu.test_constant_int_kernel_input_cuda
pytest test/inductor/test_dropout_align_random_eager.py -k test_dropout_parity_dynamic_shapes

torch/_inductor/fx_passes/replace_random.py
EricKing626: Changed

torch/_inductor/fx_passes/replace_random.py
EricKing626: @eellison adding cudagraph_unsafe solved problem. 
Can do cudagraph with TORCHINDUCTOR_CUDAGRAPHS=1 (Inductor auto-CUDAGraphs)

test/inductor/test_dropout_align_random_eager.py
EricKing626: @eellison Add in test_multi_dropout_multi_iterations_parity(), I also updated different seed in different interations

test/inductor/test_dropout_align_random_eager.py
EricKing626: Added

EricKing626:
> @EricKing626 great! we'll take a look.

@eellison Can we re-run failed tests? This all just flaky errors. Or is leaving these flaky errors ok?
And another thing, do we need to close the flag as default when merging?
image

EricKing626 (commented):
Functionality is already complete. Performance improves compared to the fallback_random implementation.

test/inductor/test_dropout_align_random_eager.py
```diff
@@ -0,0 +1,439 @@
...
+            _set_seed(BASE_SEED)
+            y_eager = eager(x)
+
+            _set_seed(BASE_SEED)
+            y_comp = compiled(x)
+
+            torch.testing.assert_close(y_eager, y_comp, rtol=1e-3, atol=1e-4)
+
+    # ───────────────────────────────────────────────────────────
+    # cudagraphs test via mode='reduce-overhead' (b)
```
eellison, line 306: Once: https://github.com/pytorch/pytorch/pull/170733 lands, it would be great to extend this to dropout. cc @v0i0

test/inductor/test_dropout_align_random_eager.py
```diff
@@ -0,0 +1,439 @@
...
+
+        t_eager, _ = _timed_run(eager, x, backward=False)
+        t_comp, _ = _timed_run(compiled, x, backward=False)
+
+        assert t_comp > 0 and t_eager > 0
+
+    # ───────────────────────────────────────────────────────────
+    # Helper for primitive random parity (rand / randn / randint)
+    # ───────────────────────────────────────────────────────────
+    def _run_primitive_random_parity(self, kind, device, shape):
```
eellison, line 378: Here were some additional failing tests (helped find with claude). https://gist.github.com/eellison/88aa2bb71f64cc01d4f5f4b2a4fc4e8a

- float64 dtype vec mismatch
- Large seed (>32-bit) packing truncation
- Non-contiguous sliced
- numel % 4 != 0 produces wrong masks

torch/csrc/inductor/inductor_ops.cpp
```diff
@@ -108,4 +108,10 @@ TORCH_LIBRARY_FRAGMENT(inductor, m) {
       {at::Tag::pt2_compliant_tag});
 }
 
+TORCH_LIBRARY_FRAGMENT(inductor_prims, m) {
+  m.def(
+      "inductor_reserve_rng_state(Generator? generator, SymInt increment) "
+      "-> (Tensor, Tensor, Tensor)");
```
eellison: not sure if we need the `at::Tag::pt2_compliant_tag` tag

test/inductor/test_dropout_align_random_eager.py
```diff
@@ -0,0 +1,439 @@
...
+            seed = BASE_SEED + i
+            x = torch.randn(BATCH, SEQ_LEN, HIDDEN_DIM, device=device)
+
+            _set_seed(seed)
+            y_eager = eager(x)
+
+            _set_seed(seed)
+            y_comp = compiled(x)
+
+            torch.testing.assert_close(y_eager, y_comp, rtol=1e-3, atol=1e-4)
```
eellison: I know this is RNG, so should sort of flush this out anyway, but should we just be using `atol=0, rtol=0` across tests (when the only operation is dropout)

test/inductor/test_dropout_align_random_eager.py
```diff
@@ -0,0 +1,439 @@
...
+    HAS_CUDA_AND_TRITON,
+    requires_gpu,
+)
+
+
+# ───────────────────────────────────────────────────────────────
+# Global config
+# ───────────────────────────────────────────────────────────────
+BASE_SEED = 1234
+DROPOUT_P = 0.3
```
eellison: nit: maybe make .5, just to maximally test divergence?

torch/_inductor/codegen/cpp.py
```diff
@@ -1094,6 +1094,16 @@ def bitwise_right_shift(a, b):
...
 
+    @staticmethod
+    def rand_eager(
+        seed: sympy.Expr,
+        base_offset: sympy.Expr,
+        threads_per_round: sympy.Expr,
+        tid: sympy.Expr,
+        vec: sympy.Expr,
+    ):
+        return f"normalized_rand_cpu({seed}, {base_offset})"
```
eellison, line 1121: just to limit the scope of what youre attempting, you might consider not trying to give parity with cpu (maybe this is already the case).

torch/_inductor/fx_passes/replace_random.py
```diff
@@ -29,10 +64,59 @@ def replace_random_passes(gm: torch.fx.GraphModule):
...
+    Horizontally fuse all the seed generation on each device
+        a = inductor_prims.rand_eager_offset(offset, dev)
+        b = inductor_prims.rand_eager_offset(offset, dev)
+    Becomes:
+        offsets = inductor_prims.rand_eager_offsets([offset1, offset2...], dev)
+        a = inductor_lookup_seed(offsets, 0)
+        b = inductor_lookup_seed(offsets, 1)
+    We do this because seed creation is entirely launch overhead bound.
+    """
+    device_offsets = collections.defaultdict(list)
```
eellison, line 86: nit: type annotation ?

torch/_inductor/fx_passes/replace_random.py
```diff
@@ -124,8 +208,34 @@ def replacement(size):
...
+    if mode == "rand" and config.align_random_eager and device.type == "cuda":
+        # Only enable when align_random_eager is on.
+        def replacement_align(size):
+            offset = _shape_to_offset(size, device)
+
+            align_dtype = dtype
+            if isinstance(align_dtype, (tuple, list)):
+                align_dtype = align_dtype[0] if len(align_dtype) else None
+            if align_dtype is None:
+                align_dtype = torch.float32
```
eellison: should we always known align_dtype ? safer to not match if we dont know?

eellison (commented):
Looks great, thanks for your work here ! I left a comment on some remaining cases that don't yet work. 

Having said that, this is already useful as is, and the pr is getting large. What do you think about explicitly covering the cases we do not yet handle, and falling back to eager ? We could also config this option so we lower and warn, or always lower. 

What do you think?

torch/_inductor/codegen/cpp.py
EricKing626: Agree — CPU parity is out of scope. rand_eager is kept as a required C++ codegen fallback/hook (otherwise codegen/CI fails). The parity target of this PR is CUDA eager vs compiled only; I'll added a note to make that explicit.

torch/_inductor/fx_passes/replace_random.py
EricKing626: Thanks for the nit! This pass is modeled after fuse_seed_creation_pass, which also uses collections.defaultdict(list) without a local annotation. I’ll add a return type annotation (-> int) here (it returns 0 or len(device_offsets)) while keeping the local defaultdict un-annotated to avoid overly specific/brittle typing.

test/inductor/test_dropout_align_random_eager.py
EricKing626: First 2 fixed with added test cases in test_dropout_align_random_eager.py
"Non-contiguous sliced" & "numel % 4 != 0 produces wrong masks" might be fallback to eager

jansel (commented):
This looks great!  Two things:
1) Do we have any perf data on this?  If it is faster maybe this should be the default rather than behind a flag.
2) Looks like there are test failures.

torch/csrc/inductor/inductor_ops.cpp
EricKing626: Added

test/inductor/test_dropout_align_random_eager.py
EricKing626: Added

test/inductor/test_dropout_align_random_eager.py
```diff
@@ -203,7 +203,7 @@ def test_dropout_mask_parity_and_rng_offset_cuda(self):
         device = torch.device(GPU_TYPE)
         H, W = BATCH * SEQ_LEN, FFN_DIM
 
-        dtypes = [torch.float32, torch.float16, torch.bfloat16, torch.float64]
```
eellison: can i ask why you removed fp64 here ?

test/inductor/test_dropout_align_random_eager.py
eellison: these no longer fallback to eager , correct?

torch/csrc/inductor/inductor_ops_gpu.cpp
```diff
@@ -0,0 +1,89 @@
...
+#if defined(USE_CUDA) || defined(USE_ROCM)
+
+// Reserves RNG state for Inductor with CUDA Graph support.
+//
+// This function allows Inductor to reserve a specific amount of RNG offset
+// (increment) for a kernel. It is designed to be safe for CUDA Graph capture
+// by explicitly handling the internal generator state via public APIs.
+//
+// Behavior:
+// - Graph Mode: Advances the generator state and returns pointers (wrapped as
```
eellison, line 30: cc maybe @eee4017 for comments

eellison (approved):
Awesome! cc @albanD , do we have any tests we do to verify random distributions, I know you recently landed some changes here. 

Before we land, lets do a dashboard run to double check that there arent any failures or regressions.

Also, I see that this is on by default, and does not warn if it's one of the cases we have not yet handled (Non-contiguous sliced, numel % 4 != 0 ).

How are you envisioning the use case ? would it make sense to add an option to either warn or fallback to eager if we can't prove we'll match numerics ?

test/inductor/test_dropout_align_random_eager.py
EricKing626: I pass torch.float64 on my local machine but failed to test in CI. I don't have bandwidth to deal with this so I am wondering if this can be added into fallback list.

eellison: fallback is fine by me!

jansel (approved):

albanD (commented):
@eellison if the new algorithm is bitwise equal with eager, I guess you can simply compare with eager given a set seed?

If you want to check the statistical properties of a new op, we use scipy.stats. See test_torch.py for how we use it.

PR #165548, opened 2025-10-15
[CUDA][cuBLASLt] addmm -- enable 2D bias in the Lt path when followed by an activation

nikitaved:
As per title.
This one is based off [#163955](https://github.com/pytorch/pytorch/pull/163955), but I will rebase once it is merged.

cc @ptrblck @msaroufim @eqy @jerryzh168 @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @dcci

### DIFF
--- a/aten/src/ATen/native/cuda/Blas.cpp
+++ b/aten/src/ATen/native/cuda/Blas.cpp
@@ -147,14 +147,24 @@ static bool isGloballyDisabledAddmmCudaLt(const at::Device& device) {
 /*
  * Check whether for the given input we want to enable the Lt interface
  */
-static bool isInputCompliesAddmmCudaLt(Tensor& result, const Tensor& self, const Tensor& mat1, const Tensor& mat2, const Scalar& beta, const Scalar& alpha) {
+static bool isInputCompliesAddmmCudaLt(
+    Tensor& result,
+    const Tensor& self,
+    const Tensor& mat1,
+    const Tensor& mat2,
+    const Scalar& beta,
+    const Scalar& alpha,
+    Activation activation
+) {
+  #ifdef USE_ROCM
   // Implies 2D bias which we currently not send through Lt.
   // TODO: this check is done pre col-major input preparation,
   // so, this condition can be ralexed in cases when a col-major
   // copy of result is needed.
-  if (result.is_same(self)) {
+  if (self.is_same(result) || self.dim() == 2) {
     return false;
   }
+  #endif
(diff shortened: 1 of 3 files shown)


aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -328,10 +328,28 @@ static bool isInputCompliesAddmmCudaLt(Tensor& result, const Tensor& self, const
...
+    // self needs to be copied into result and the bias ptr
+    // will be ignored.
     && result.dim() == 2 && result.is_contiguous()
+    && ( // bias restrictions
+      // 1D bias restrictions. Implies no extra data preparation for the Lt path.
+      (self.dim() == 1 && self.sizes()[0] == mat2_sizes[1] && self.is_contiguous()) ||
+      // 2D bias restrictions. self.is_contiguous() is implicit when result.is_same(self),
+      // and we need to copy self into result otherwise, so the self's layout becomes irrelevant.
+      // See also TODO from above.
+      (self.dim() == 2 && self.sizes()[0] == mat1_sizes[0] && self.sizes()[1] == mat2_sizes[1])
```
nikitaved: sufficient to check that `self.shape` broadcasts over `(m1.shape[0], m2.shape[1])`.

aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -425,7 +443,12 @@ bool launchGemmAndBiasCublasLt(
     const Scalar& alpha,
     Activation activation = Activation::None
 ) {
-  const auto* self_ptr = self.const_data_ptr<scalar_t>();
+  const auto* self_ptr = [&]() -> auto {
+    if (self.dim() == 1) {
+      return self.const_data_ptr<scalar_t>();
+    }
+    return static_cast<const scalar_t*>(nullptr);
+  }();
```
nikitaved, line 312: Need a comment about that special treatment of `nullptr` bias

eqy (approved):

eqy:
For my own edification, does this mean that there is path towards using `blasPreferredBackend` to control the dispatch choice for 2D bias cases if it is not already possible?

nikitaved:
@eqy , no, unfortunately, there is no such path, iiuc, as it is being dispatched to an independent low-level function. From what I see, the backend selection only controls the gemm, and does not include bias nor activations.

PR #165554, opened 2025-10-15
[DeviceMesh] Prefer using _layout over _mesh for all sorts of things

lw:
* #165556
* #165555
* __->__ #165554

The goal of this PR is to avoid storing the explicit `mesh` Tensor inside each DeviceMesh, and instead compute it on-the-fly when the end user needs it, and try to replace all of its internal usages with `_layout` and the newly-introduced `_global_rank_permutation` Tensor. The name of this attribute is up for debate. The advantage of the `_global_rank_permutation` Tensor is that it is _the same_ Tensor for the root mesh and all its children, so it doesn't need to be copied/reallocated.


torch/distributed/device_mesh.py
```diff
@@ -173,7 +173,7 @@ class DeviceMesh:
         """
 
         _device_type: str
-        _mesh: torch.Tensor
+        _global_rank_permutation: torch.Tensor
```
lw: As I mentioned, I welcome suggestions for better names.

torch/distributed/device_mesh.py
```diff
@@ -190,44 +190,49 @@ def __init__(
...
+            mesh_tensor = (
                 mesh.detach().to(dtype=torch.int).contiguous()
                 if isinstance(mesh, torch.Tensor)
                 else torch.tensor(mesh, device="cpu", dtype=torch.int)
             )
+            self._global_rank_permutation = (
+                _root_mesh._global_rank_permutation
+                if _root_mesh is not None
+                else mesh_tensor.flatten()
+            )
```
lw: This is a bit of a mess, but the next PR in the stack will clean this up.

torch/distributed/device_mesh.py
```diff
@@ -258,7 +266,13 @@ def device_type(self) -> str:
         @property
         def mesh(self) -> torch.Tensor:
```
lw, line 269: This function is admittedly quite complicated. I think it has been a good exercise to write it out as it helps us understand what are the contents of the `mesh` Tensor. If we agree on this definition, we can work next on cleaning this up :)

torch/distributed/device_mesh.py
```diff
@@ -258,7 +266,13 @@ def device_type(self) -> str:
...
             """Returns the tensor representing the layout of devices."""
-            return self._mesh
+            full_mesh = self._layout.remap_to_tensor(self._global_rank_permutation)
+            my_coords = (full_mesh == get_rank()).nonzero()
+            if my_coords.size(0) > 0:
+                return full_mesh[my_coords[0, 0]]
+            if full_mesh.size(0) == 1:
+                return full_mesh[0]
```
lw: This case here is for DeviceMeshes that do not contain the rank on which they're being instantiated.

In general, I have a hard time understanding in which cases they would be useful, and I'd like to explore the possibility of forbidding such a scenario.

Concretely, this codepath is currently being triggered by `get_all_submeshes`, thus I'd like to look into whether we can rework the internal usages of that private method and hopefully remove it or improve it.

torch/distributed/device_mesh.py
fduwjj: I personally don't like the idea of permutation. Shall we call it `_global_ranks_alloc`?

torch/distributed/device_mesh.py
```diff
@@ -250,6 +255,9 @@ def __init__(
                     rank_coords[0].tolist() if rank_coords.size(0) > 0 else None
                 )
 
+            # private field to pre-generate DeviceMesh's hash
+            self._flatten_mesh_list = tuple(self.mesh.flatten().tolist())
```
fduwjj, line 259: you can directly use mesh_tensor here.

torch/distributed/device_mesh.py
lw: Why `alloc` (which I guess stand for allocation)? In which sense is this an allocation?

An alternative name could also be `_rank_order`, which has the benefit of being shorter.

torch/distributed/device_mesh.py
fduwjj: Agree we can also make `get_all_submeshes` not rely on this case..

fduwjj: One thing is required, can you kindly run DTensor CPU overhead benchmark here: https://github.com/pytorch/pytorch/issues/159169. At least we have an understanding of how much overhead this change will bring into.

torch/distributed/device_mesh.py
```diff
@@ -463,7 +479,8 @@ def __hash__(self):
                 self._hash = hash(
                     (
                         self._flatten_mesh_list,
-                        self._mesh.shape,
+                        self._layout.sizes,
```
fduwjj: layout has a __hash__, so self._layout should be enough?

torch/distributed/device_mesh.py
```diff
@@ -479,7 +496,8 @@ def __eq__(self, other: object) -> bool:
                 return False
             return (
                 self._flatten_mesh_list == other._flatten_mesh_list
-                and self._mesh.shape == other._mesh.shape
+                and self._layout.sizes == other._layout.sizes
```
fduwjj: ditto, we can directly compare the layout.

fduwjj (approved):
Nice, this looks make sense to me.. Thanks for continuing helping make device mesh cleaner. Just one question for naming and DTensor CPU overhead.

torch/distributed/device_mesh.py
fduwjj: Because it means rank allocation? The actual mesh tensor [0, 3, 5, 10] (let's say a random tensor) it means a device allocation. It refers to a device representation. That's why I think allocation might be better? I mean I don't have a strong opinion on the word allocation, and I just want to make sure the name here clearly represents what it means

torch/distributed/device_mesh.py
```diff
@@ -346,11 +360,13 @@ def _init_process_groups(
...
                 # create sub pgs base on the mesh argument specified
-                for dim in range(self.mesh.ndim):
+                for dim in range(len(self._layout)):
                     # swap the current dim to the last dim
                     # then reshape to flatten out other dims
-                    pg_ranks_by_dim = self.mesh.swapdims(-1, dim).reshape(
-                        -1, self.mesh.size(dim)
+                    pg_ranks_by_dim = (
+                        self._layout[dim]
+                        .nest()
```
fduwjj: I don't think nest is needed here? I mean inside `remap_to_tensor` we are flatten its size/strides anyway.

torch/distributed/device_mesh.py
```diff
@@ -346,11 +360,13 @@ def _init_process_groups(
...
+                        .remap_to_tensor(self._global_rank_permutation)
```
fduwjj: nice, this really makes the code cleaner.

torch/distributed/device_mesh.py
lw: I thought about what could be the shortest possible name (to keep things concise), and I like `_rank_map`, which I find also quite self-descriptive, since this is exactly what we're using this for: to _map_ coordinates to actual ranks. It's also the term we're already using in `remap_to_tensor`.

Are people ok with this name?

torch/distributed/device_mesh.py
lw: I prefer not to do that, since in the next PR this line will be moved to a separate constructor where we will not have access to `mesh_tensor`.

torch/distributed/device_mesh.py
lw: Good call!

torch/distributed/device_mesh.py
lw: Unfortunately we do. `remap_to_tensor` returns a tensor which has N+1 dims, where N is the size of the IntTuple we give to it, and the extra 1 is for the "outer" complement. `self._layout[dim]` is the inner layout of one specific dim, and we don't know how many elements it will have (1 for contiguous dims, 2+ when flattening non-adjacent dims, ...). Thus we need to call `nest` to make sure the top-level always has exactly 1 element, since we want `pg_ranks_by_dim` to be a 2-d tensor.

lw:
I ran the CPU benchmark. I looked at the iteration_4/rank0_trace.json files, and measured the e2e time of the first forward on CPU. Before this PR it took 1.526s, after this PR it took 1.521s. Hence no regression.

torch/distributed/device_mesh.py
fduwjj: `_rank_map` sounds good to me.

torch/distributed/device_mesh.py
fduwjj: I agree that we want a 2D tensor here, and I understand what you mean above. But since we already do nest in flatten which is in your first PR in this stack, why do we do it again here? That's the part I find confusing. I mean it is not blocker for merging, I just feel it is not necessary.

lw: I see what you mean, but I believe the two issues are unrelated. Let me try to explain.
- Before the fix in #165542, the result of `mesh["dp", "tp"]._flatten()` would be a layout with sizes `(2, 2)`, which would be mis-interpreted as _2_ contiguous dimensions.
- After the fix, that same layout now becomes `((2, 2),)`, which indicates that it's _1_ non-contiguous dimension. However, when we do `layout[0]` on that value we recover `(2, 2)`, which `remap_to_tensor` interprets as a 2-d layout
- Hence we need to again call nest to bring it back to `((2, 2),)`.

Note that another way to fix it could have been this:
```diff
- self._layout[dim].nest()
+ self._layout[dim : dim + 1]
```

fduwjj: oh I see. Thanks for explanation.

fduwjj:
somehow this test failed: python test/distributed/test_serialization.py TestSerialization.test_dtensor

torch/distributed/device_mesh.py
```diff
@@ -260,7 +268,17 @@ def device_type(self) -> str:
         @property
         def mesh(self) -> torch.Tensor:
             """Returns the tensor representing the layout of devices."""
-            return self._mesh
+            full_mesh = self._layout.remap_to_tensor(self._rank_map)
```
fduwjj, line 271: Looks like this line fails when the input is tensor(1)

fduwjj: probably we need to call nest here as well?

PR #165555, opened 2025-10-15
[DeviceMesh] Introduce private constructor instead of _create_mesh_from_ranks

lw:
* #165556
* __->__ #165555
* #165554

The refactoring of DeviceMesh is heavily constrained by the signature of its constructor, which is a public API which contains some "legacy" concepts which we'd love to get rid of, such as an explicit/materialized `mesh` Tensor.

In other languages the solution to this would be to add a private overload of the constructor. Python doesn't natively allow this, but in this PR I managed to build something that approximates it.

This new private constructor basically only takes `_layout`, `_global_rank_permutation`, and `mesh_dim_names`.

With such a constructor we can effectively simplify a lot of callsites and get rid of the `_create_mesh_from_ranks` helper method. That's a good thing because it was instantiating many DeviceMeshes in a for loop, which always felt unnecessary.


torch/distributed/device_mesh.py
```diff
@@ -258,6 +285,23 @@ def __init__(
...
+                    sig.bind(self, *args, **kwargs)
+                except TypeError as e:
+                    mismatches.append((f"{sig!s}", f"{e!s}"))
+                    continue
+                func(self, *args, **kwargs)
+                return
+            raise TypeError(
+                f"{args=} and {kwargs=} didn't match any overload:\n"
+                + "\n".join(f"- overload {s}: {e}" for s, e in mismatches)
+            )
```
lw: This is not very standard, but it works. How do people feel about it?

fduwjj: Interesting. Maybe @fegin can comment more on this one.

fduwjj (commented):
Aside the init override part, the rest looks good to me

fduwjj (approved):
LGTM

torch/distributed/device_mesh.py
```diff
@@ -183,44 +182,51 @@ class DeviceMesh:
         def __init__(
             self,
             device_type: str,
-            mesh: Union[torch.Tensor, "ArrayLike"],
+            mesh: Optional[Union[torch.Tensor, "ArrayLike"]] = None,
```
fegin, line 185: Can we update the docstring to explain when `mesh` can be None? Even if we just preserve it for the internal usage, we should mention because this is a public API.

fegin (approved):

PR #165556, opened 2025-10-15
[DeviceMesh] Simplify unflatten method

lw:
* __->__ #165556
* #165555
* #165554

By adding a few small helpers (e.g., a `splice` method to `_MeshLayout`, and making `_init_process_groups` static and thus stateless) we can substantially shorten the definition of the unflatten method, and help readability.


torch/distributed/_mesh_layout.py
```diff
@@ -156,50 +161,11 @@ def complement(self, world_size: int) -> "_MeshLayout":
         layout = complement(self, world_size)
         return _MeshLayout(layout.shape, layout.stride)
 
-    def unflatten(self, dim: int, unflatten_sizes: tuple[int, ...]) -> "_MeshLayout":
```
lw, line 155: I know that it was me who insisted for making this a method of `_MeshLayout`, but I hadn't realized that we then needed to re-extract the sub-layout in order to create the ProcessGroups. This made for some bulky code.

Thus my new suggestion is that we break this monolithic method into two: a composition one (which already exists) and a splice one. This allows the DeviceMesh to achieve the same result as this method in two lines of code, while easily getting access to the intermediate value it needs.

torch/distributed/device_mesh.py
```diff
@@ -373,8 +378,11 @@ def _setup_world_group_and_device(self):
 
             return _get_default_group()
 
+        @staticmethod
         def _init_process_groups(
```
lw, line 350: This could also be extracted to become a global private helper function. I didn't do it as I wanted to keep the diff small.

torch/distributed/_mesh_layout.py
```diff
@@ -156,50 +161,11 @@ def complement(self, world_size: int) -> "_MeshLayout":
...
-                f"These must be equal for unflatten to work correctly."
-            )
-
-        sizes = list(self.sizes)  # type: ignore[arg-type]
-        strides = list(self.strides)  # type: ignore[arg-type]
-        unflatten_layout = self[dim].composition(
-            _MeshLayout(tuple(unflatten_sizes), suffix_product(unflatten_sizes))
-        )
-        sizes[dim : dim + 1] = list(unflatten_layout.sizes)  # type: ignore[arg-type]
-        strides[dim : dim + 1] = list(unflatten_layout.strides)  # type: ignore[arg-type]
```
lw, line 198: Note that the `# type: ignore[arg-types]` were hiding actual bugs, since `.sizes` and `.strides` could be integers, but we can't pass these to `list(...)`! The new code fixes this thanks to `as_tuple`.

torch/distributed/device_mesh.py
```diff
@@ -373,8 +378,11 @@ def _setup_world_group_and_device(self):
...
-            self,
+            layout: _MeshLayout,
+            global_rank_permutation: torch.Tensor,
+            mesh_dim_names: Optional[tuple[str, ...]],
             backend_override: tuple[BackendConfig, ...],
         ):
```
fduwjj: Looks like we are changing the return signature of this function, can you also update it here?

torch/distributed/device_mesh.py
```diff
@@ -1124,30 +1141,13 @@ def _create_unflatten_mesh(
...
-                unflatten_length = len(mesh_sizes)
-                unflatten_layout = _MeshLayout(
-                    tuple(unflattened_layout.sizes[dim : dim + unflatten_length]),  # type: ignore[index]
-                    tuple(unflattened_layout.strides[dim : dim + unflatten_length]),  # type: ignore[index]
+                dim_group_names = self._dim_group_names.copy()
+                dim_group_names[dim : dim + 1] = self._init_process_groups(
+                    partial_layout,
+                    root_mesh._global_rank_permutation,
+                    mesh_dim_names,
+                    backend_override,
```
fduwjj, line 1111: I like this one so that we can reuse it later if we only want to init backend for some dims not all. And also to make _flatten generic like what Tensor is doing also needs this.

torch/distributed/_mesh_layout.py
```diff
@@ -156,50 +161,11 @@ def complement(self, world_size: int) -> "_MeshLayout":
...
+    def splice(self, start: int, end: int, layout: "_MeshLayout") -> "_MeshLayout":
+        sizes = list(as_tuple(self.sizes))
```
fduwjj, line 161: with as_tuple and flatten, do we still want to do the proposal of limiting the sizes and strides to be at most 2d?

fduwjj (approved):
this change looks reasonable to me as well. And we can build the shared_state on top of that so that we can also cache the PG for unflatten.

torch/distributed/device_mesh.py
lw: Good catch, fixed!

torch/distributed/_mesh_layout.py
lw: The way I see it, `as_tuple` and `flatten` are the tools we need to make the code correct, but moving from IntTuple to `tuple[tuple[int, ...], ...]` is what will make the type checker help us detect those bugs in the first place.

PR #165563, opened 2025-10-15
User-passed alpha to scaled_gemm

slayton58:
* __->__ #165563

Summary:

Add optional user-passed `alpha` argument to
`at::cuda::blas::scaled_gemm`, necessary for two-level-scaled NVFP4 gemm
calls (where the global de-scales are folded into the `alpha` argument.

Global de-scales are naturally device tensors, but using cublas'
device-pointer mode for `alpha`/`beta` has an interesting lifetime
implication - the `alpha` tensor must be valid & correct until the end
of the matmul call, *not* just the launch (as for host values). To
enable this, I added device-constant memory for `one` and `zero`, along
with a statically-held single-fp32-value tensor, which is valid from the
first passed-`alpha` invocation of `scaled_gemm` to the end of the
program. User-passed values are copied into this perpetual buffer to
ensure lifetime requirements are met.

Test Plan:

Reviewers:

Subscribers:

Tasks:

Tags:


aten/src/ATen/cuda/CUDABlas.cpp
```diff
@@ -1954,13 +1956,15 @@ void scaled_gemm(
...
+    const std::optional<Tensor>& alpha) {
   // Note: see `cublasCommonArgs` for various non-intuitive manupulations
   // of input arguments to this function.
   const auto computeType = CUBLAS_COMPUTE_32F;
   const auto scaleType = CUDA_R_32F;
-  const float alpha_val = 1.0;
-  const float beta_val = 0.0;
+  // Note: alpha_val may change later depending on user-passed argument
+  float alpha_val = 1.0;
+  float beta_val = 0.0;
```
drisspg, line 1967: super nit: keep beta const?

aten/src/ATen/cuda/CUDABlas.cpp
```diff
@@ -2031,6 +2035,27 @@ void scaled_gemm(
...
+    // if device-tensor
+    if (a.is_cuda()) {
+      at::Tensor& user_alpha = at::cuda::detail::get_user_alpha_tensor();
+      user_alpha.copy_(a);
+      auto pointer_mode = CUBLASLT_POINTER_MODE_DEVICE;
+      computeDesc.setAttribute(CUBLASLT_MATMUL_DESC_POINTER_MODE, pointer_mode);
+      alpha_ptr = user_alpha.data_ptr<float>();
+      beta_ptr = at::cuda::detail::get_cublas_device_zero();
+    } else {
+      alpha_val = *a.data_ptr<float>();
```
drisspg: maybe .item(), but then again this is probably faster

aten/src/ATen/cuda/CUDABlas.cpp
slayton58: To keep host / device paths as equivalent as possible, I create `float *` variables for alpha/beta that are then populated. If I keep `beta_val` const, I'm not allowed to do 
```
float *beta_ptr;
...
beta_ptr = &beta_val;
```
as I can't assign a `const float *` to a `float *`

aten/src/ATen/cuda/detail/BLASConstants.cu
```diff
@@ -0,0 +1,54 @@
...
+  return ptr;
+}
+
+at::Tensor& get_user_alpha_tensor() {
+  static at::Tensor alpha;
+
+  static std::once_flag init_flag;
+
+  std::call_once(init_flag, []() {
+    alpha = at::empty({1}, TensorOptions().device(kCUDA).dtype(kFloat));
```
eqy: If this isn't clearable it has a chance of blowing up tests that assume a blank slate of allocations. I think there are three immediately obvious possibilities here:
(0) patch fudge factors of 1-byte to any failing tests
(1) add functionality to clear this allocation for testing
(2) bypass the caching allocator and raw `cudaMalloc` this allocation since it's of minimal size to bypass CUDA caching allocator accounting

aten/src/ATen/cuda/CUDABlas.cpp
drisspg: ohh you are right

aten/src/ATen/cuda/CUDABlas.cpp
slayton58: Yeah, `.item<float>()` does feel cleaner.

aten/src/ATen/cuda/detail/BLASConstants.cu
slayton58: The allocation should only happen when a user passes a device-side tensor as `alpha` to `at::cuda::blas::scaled_gemm`, i.e. in very, _very_ specific cases - otherwise it's just a blank `at::Tensor`, no contents. Is that still a problem?

If it is still an issue, I think I'd go with option (2) - it's minimal code changes on both sides of the interface.

aten/src/ATen/cuda/detail/BLASConstants.h
```diff
@@ -0,0 +1,15 @@
+#pragma once
+
+#include <ATen/core/TensorBase.h>
+
+namespace at {
+namespace cuda {
+namespace detail {
+
+float *get_cublas_device_one();
```
drisspg, line 7: Super nit: nested namespaces

drisspg (approved):

eqy (approved):

aten/src/ATen/cuda/detail/BLASConstants.h
slayton58: Discussed offline - I had no idea one could even do
```
namespace foo::bar::baz { 
..
} // namespace foo::bar::baz
```
:)

Fixed.

PR #165568, opened 2025-10-15
tmp fix for compile internal logger issue

zou3519:
Summary: Catch runtime exception when garse and scrub uninteresting configs from inductor config

Test Plan: tested locally

Differential Revision: D84727788


luccafong (approved):

oulgen (approved):
good to unblock for now, but ideally the config does not have an nn.module in it

PR #165571, opened 2025-10-15
added flop formula for element wise operations

### LINKED ISSUE #165570: Add support for element-wise operations to flop counter

FlopCounterMode currently does not support operators that perform element wise operations such as element wise addition. Let's add support for these operations by registering a flop formula for them.

### Alternatives

_No response_

### Additional context

_No response_

alisyeMeta:
Fixes #165570

### DIFF
--- a/torch/utils/flop_counter.py
+++ b/torch/utils/flop_counter.py
@@ -160,6 +160,14 @@ def conv_flop(x_shape, w_shape, _bias, _stride, _padding, _dilation, transposed,
     return conv_flop_count(x_shape, w_shape, out_shape, transposed=transposed)
 
 
+@register_flop_formula([aten.add, aten.sub, aten.mul, aten.div, aten.pow, aten.remainder, aten.add_, aten.sub_, aten.mul_, aten.div_, aten.pow_, aten.remainder_])
+def element_wise_flop(x_shape, y_shape, *args, out_shape=None, **kwargs) -> int:
+    """Count flops for element-wise operations."""
+    # pyrefly: ignore  # bad-argument-type
+    broadcasted_shape = torch.broadcast_shapes(x_shape, y_shape)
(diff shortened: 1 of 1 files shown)


eellison (commented):
What is the intended use case for this ?

alisyeMeta:
> What is the intended use case for this ?

We have a tool that is estimating the flop count required to train models, this tool is using FlopCounterMode under the hood to generate this estimation. However there are a lot of gaps in FlopCounterMode's coverage, these element wise operators are used quite extensively by the users of our tool I'm hoping by adding support for these operators and others in follow up PRs we can improve the accuracy of our tool.

torch/utils/flop_counter.py
```diff
@@ -153,6 +153,14 @@ def conv_flop(x_shape, w_shape, _bias, _stride, _padding, _dilation, transposed,
     return conv_flop_count(x_shape, w_shape, out_shape, transposed=transposed)
 
 
+@register_flop_formula([aten.add, aten.sub, aten.mul, aten.div, aten.pow, aten.remainder, aten.add_, aten.sub_, aten.mul_, aten.div_, aten.pow_, aten.remainder_])
```
eellison: is `pow` exactly one ? no, right ? why not just check torch.tag.Pointwise if this isnt exactly right anyway

eellison (commented):
why not just check if `torch.tag.Pointwise` is in the operator instead ?

torch/utils/flop_counter.py
alisyeMeta: yeah I think i need to put some more thought into `pow`, I'll remove it from the PR and make a follow up for it later

eellison (changes requested):
still dont think we should do this, but you can subclass your own flop counter and use the `torch.tag.Pointwise` to assume single floating point operation

PR #165580, opened 2025-10-15
[scan x vmap] support scan in vmap

ydwu4:
* #165616
* __->__ #165580
* #165675

This is required by the chunked_with_scan work where two nested vmap(vmap) with chunk sizes > 1 are invoked, which produces a scan-> vmap -> scan -> vmap chain and we need to handle the case of vmap(scan) and scan(vmap).

The way we handle vmap(scan) is to turn it into scan(vmap(combine_fn)). The idea being that the combine_fn no longer do the combine_fn for a single slice, it vmaps over the combine_fn and do multiple combine_fns in one step. We need to need know how combine_fn propagates the batched tensor and what are the batched dims of the output. For this purpose, we use restore_vmap to give us the out_dims information. 


torch/_higher_order_ops/scan.py
```diff
@@ -866,6 +866,76 @@ def scan_functionalize(ctx, combine_fn, init, xs, additional_inputs):
...
+@scan_op.py_impl(torch._C._functorch.TransformType.Vmap)
+def scan_batch_rule(interpreter, combine_fn, init, xs, additional_inputs):
+    from torch._functorch.vmap import (
+        is_batchedtensor,
+        restore_vmap,
+        unwrap_batched,
+        wrap_batched,
+    )
+
+    if any(is_batchedtensor(x) for x in [*init, *xs, *additional_inputs]):
```
Skylion007:
```suggestion
    if any(is_batchedtensor(x) for x in itertools.chain(init, xs, additional_inputs)):

zou3519 (approved):

PR #165581, opened 2025-10-15
torch.compile: populate compiler_config

c00w:
Summary: This starts writing the compiler_config metadata into logger

Test Plan:
Modified existing test case to make sure this is not null.
(Also eyeballed what we're logging tomake sure it's reasonable

Reviewed By: masnesral

Differential Revision: D84014636

### DIFF
--- a/torch/_dynamo/utils.py
+++ b/torch/_dynamo/utils.py
@@ -1315,6 +1315,7 @@ class CompilationMetrics:
     config_inline_inbuilt_nn_modules: Optional[bool] = None
     specialize_float: Optional[bool] = None
     dynamo_config: Optional[str] = None
+    compiler_config: Optional[str] = None
     is_forward: Optional[bool] = None
     num_triton_bundles: Optional[int] = None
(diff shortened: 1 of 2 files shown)


torch/_dynamo/utils.py
```diff
@@ -1553,6 +1554,27 @@ def clean_for_json(d: dict[str, Any]) -> dict[str, Any]:
...
+
+        return {
+            key: sorted(value) if isinstance(value, set) else value
+            for key, value in d.items()
+            if key not in blocklist
+        }
+
+    if torch.compiler.config:
+        try:
+            compiler_config_copy = torch.compiler.config.get_config_copy()
```
masnesral: lint complaint looks legit?

masnesral (changes requested):

torch/_dynamo/utils.py
c00w: It's wrong, I made the test case explicit, but get_config_copy is on that module - not sure why my-py is flagging it, but I'm adding a supress.

masnesral (approved):

PR #143733, opened 2024-12-23
[CI] enable operator benchmark on CPU

LifengWang:
This is to enable operator benchmark for CPU to track op level performance. This PR is motivated by PR: https://github.com/pytorch/pytorch/issues/120982 and investigate feasibility in https://github.com/pytorch/pytorch/pull/127216

cc @albanD

PR #165583, opened 2025-10-15
[BE] Remove confusing `opbenchmark-on-demand-build`

malfet:
* #165585
* #165584
* __->__ #165583

As it doesn't have a test shard, so what's the point or running the build? Was added in https://github.com/pytorch/pytorch/pull/143733 and looks like test shard never existed for it

Moreover, allow one to specify benchmark size as argument, so one
technically can do a workflow dispatch with different opbenchmark sizes


huydhn (approved):
Good catch!

.github/workflows/operator_benchmark.yml
```diff
@@ -32,19 +32,6 @@ jobs:
     if: github.repository_owner == 'pytorch'
     name: opbenchmark-build
     uses: ./.github/workflows/_linux-build.yml
-    with:
-      build-environment: linux-jammy-py3.10-gcc11-build
-      docker-image-name: ci-image:pytorch-linux-jammy-py3-gcc11-inductor-benchmarks
-      test-matrix: |
-        { include: [
-          { config: "cpu_operator_benchmark_short", shard: 1, num_shards: 1, runner: "linux.12xlarge" },
```
huydhn, line 40: I don't like it when GH doesn't allow me to comment on line 53, but I think you will need to provide a default there https://github.com/pytorch/pytorch/actions/runs/18543654982/job/52858790226?pr=165583:

```
{ config: "cpu_operator_benchmark_${{ inputs.test_mode || 'short' }}", shard: 1, num_shards: 1, runner: "linux.12xlarge" },
```

The default for `inputs.test_mode` is only there for `workflow_dispatch`

malfet: Hmm, `inputs.test_mode` already have a default value, isn't it?

malfet: Removed `required: false` let's see how it changes the equation

PR #165584, opened 2025-10-15
[CI] Add arch prefix in front of op benchmark results

malfet:
* #165585
* __->__ #165584
* #165583

To be able to run x86 and aarch64 benchmarks later on


.github/workflows/operator_benchmark.yml
```diff
@@ -28,9 +28,9 @@ permissions:
   contents: read
 
 jobs:
-  opbenchmark-build:
+  x86-opbenchmark-build:
```
huydhn, line 33: Nit: I wonder if a strategy would work here, but having separate build and test jobs for x86 and aarch64 works too, and probably simpler

huydhn (approved):

.github/workflows/operator_benchmark.yml
malfet: Sure, I can do the matrix in followup PRs

PR #165592, opened 2025-10-15
Fix torch.clamp half/bfloat16 overflow on CPU and MPS

bbopen:
## Summary
- Resolves #155671 where CPU/MPS clamp overflowed while CUDA saturated
- Convert scalar bounds through double/opmath before casting in CPU and MPS helpers and adjust tests to use math.isnan with broader backend coverage
- Manually confirmed the rebuilt MPS path matches CPU using a repro script after installing with USE_MPS=1

## Testing
- python test/test_torch.py -k clamp_half_scalar_bounds_cpu -v
- python test/test_torch.py -k clamp_bfloat16_scalar_bounds_cpu -v
- python test/test_torch.py -k clamp_scalar_saturation_consistency -v

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @aditew01 @malfet @kulinseth @DenisVieriu97 @jhavukainen

### DIFF
--- a/test/test_torch.py
+++ b/test/test_torch.py
@@ -50,14 +50,14 @@ from torch.testing._internal.common_device_type import (
     expectedFailureMeta,
     expectedFailureXLA,
     instantiate_device_type_tests,
-    onlyCUDA, onlyCPU,
+    onlyCUDA, onlyCPU, onlyMPS,
     dtypes, dtypesIfCUDA, dtypesIfCPU, deviceCountAtLeast,
     skipMeta, PYTORCH_CUDA_MEMCHECK, largeTensorTest, onlyNativeDeviceTypes, skipCUDAIfNotRocm,
-    get_all_device_types, skipXLA)
+    get_all_device_types, skipXLA, TEST_MPS)
 import torch.backends.quantized
 import torch.testing._internal.data
 from torch.testing._internal.common_cuda import (
-    tf32_on_and_off, TEST_CUDNN, TEST_MULTIGPU,
+    tf32_on_and_off, TEST_CUDNN, TEST_MULTIGPU, TEST_CUDA,
     _create_scaling_case, _create_scaling_models_optimizers)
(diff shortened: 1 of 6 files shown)


bbopen:
This change fixes a user-facing bug in torch.clamp, so the `topic: not user facing` label doesn’t really apply. If a maintainer could update the labels accordingly, that’d be great.

test/test_torch.py
```diff
@@ -158,6 +158,177 @@ def test_constants(self, device):
         self.assertIsInstance(torch.inf, float)
         self.assertEqual(torch.inf, math.inf)
 
+    @onlyCPU
```
malfet: Why not test it across all devices?

test/test_torch.py
```diff
@@ -158,6 +158,177 @@ def test_constants(self, device):
...
+        self.assertEqual(out_min[1].item(), torch.clamp_min(torch.tensor(1.0, dtype=torch.float16, device=device), expected_min).item())
+
+        # NaN bounds should propagate entirely
+        nan_min_out = torch.clamp(x, min=float("nan"), max=max_scalar)
+        self.assertTrue(torch.isnan(nan_min_out).all().item())
+        nan_max_out = torch.clamp(x, min=min_scalar, max=float("nan"))
+        self.assertTrue(torch.isnan(nan_max_out).all().item())
+
+    @onlyCPU
+    def test_clamp_bfloat16_scalar_bounds_cpu(self, device):
```
malfet: Rather than repeating the same test multiple times, use `dtype` as argument and instantiate it across multiple dtypes (or better yet, just add it to OpInfo testing for clamp)

malfet (changes requested):
Please do not introduce 5 instances of the same times, but rather instantiate it across multiple devices/dtypes and use such `torch.finfo()` property to dynamically define dtype limits

Also, would be curious to hear why one behavior is preferable to another

PR #165599, opened 2025-10-16
[Inductor-FX] Support Tensor.item

blaine-rister:
# Feature
This PR supports compiling `Tensor.item` with Inductor's FX backend. This maps to a custom WrapperCodeGen method called `codegen_dynamic_scalar`.

# Implementation
The implementation is fairly mechanical, following the usual flow for these types of PRs.
1. Introduce a new Wrapper IR line for this, called `DynamicScalarLine`.
2. Split `PythonWrapperCodegen.codegen_dynamic_scalar` into 2 parts: a public method which generates the Wrapper IR line, and a private one generating Python from Wrapper IR.
3. Implement an FX codegen method for the wrapper IR line. This one calls `aten.where.Scalar` to handle code like `1 if x.item() else 0`, which is a bit tricky. It also calls `aten.item.default` to convert tensors to scalars.

# Test plan
Added CI tests mirroring the AOTI ones. They test float, int and bool types, the latter taking a distinct codegen path.

### DIFF
--- a/torch/_inductor/codegen/wrapper_fxir.py
+++ b/torch/_inductor/codegen/wrapper_fxir.py
@@ -769,7 +769,7 @@ class FxConverter:
                 args=(generate_item(input_fx_node), keypath[0].divisor),
             )
         else:
-            raise AssertionError(f"unrecognized keypath {keypath}")
+            raise NotImplementedError(f"Unsupported keypath {keypath}")
 
         result_symbol = ir_node.sym
         result_buffer = SymbolBuffer(result_symbol)
--- a/torch/_inductor/codegen/wrapper_fxir.py
+++ b/torch/_inductor/codegen/wrapper_fxir.py
@@ -29,6 +29,7 @@ from torch._library.triton import wrap_triton
 from torch.fx import GraphModule
 from torch.fx.experimental.symbolic_shapes import (
     CallMethodKey,
+    ConvertIntKey,
     DivideByKey,
     free_unbacked_symbols,
 )
@@ -54,6 +55,7 @@ from .wrapper import (
     CommBufferFreeLine,
     CommentLine,
     ConditionalLine,
+    DynamicScalarLine,
     EnterDeviceContextManagerLine,
     EnterSubgraphLine,
     ExitDeviceContextManagerLine,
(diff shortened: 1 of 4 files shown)


torch/_inductor/codegen/wrapper_fxir.py
```diff
@@ -736,6 +738,44 @@ def _generate_comment(self, line: WrapperLine) -> None:
...
+                args=(input_fx_node, 1, 0),
+            )
+            result_fx_node = generate_item(where_fx_node)
+        elif len(keypath) == 1 and isinstance(keypath[0], DivideByKey):
+            result_fx_node = graph.call_function(
+                operator.floordiv,
+                args=(generate_item(input_fx_node), keypath[0].divisor),
+            )
+        else:
+            raise AssertionError(f"unrecognized keypath {keypath}")
```
blaine-rister: TODO: `NotImplementedError`?

torch/_inductor/codegen/wrapper_fxir.py
```diff
@@ -736,6 +738,44 @@ def _generate_comment(self, line: WrapperLine) -> None:
...
+            result_fx_node = generate_item(input_fx_node)
+        elif len(keypath) == 1 and isinstance(keypath[0], ConvertIntKey):
+            where_fx_node = graph.call_function(
+                aten.where.Scalar,
+                args=(input_fx_node, 1, 0),
+            )
+            result_fx_node = generate_item(where_fx_node)
+        elif len(keypath) == 1 and isinstance(keypath[0], DivideByKey):
+            result_fx_node = graph.call_function(
+                operator.floordiv,
```
angelayi: do you think you could add some tests for these paths? It's not really clear to me when we run into these case 😅

angelayi (approved):

torch/_inductor/codegen/wrapper_fxir.py
blaine-rister: Sure. I think this `DivideByKey` path is only triggered by a pretty obscure optimization. Will add a CI test for it. The other paths are covered by the existing CI.

jansel (approved):

torch/_inductor/codegen/wrapper_fxir.py
```diff
@@ -736,6 +738,44 @@ def _generate_comment(self, line: WrapperLine) -> None:
...
+                args=(generate_item(input_fx_node), keypath[0].divisor),
+            )
+        else:
+            raise NotImplementedError(f"Unsupported keypath {keypath}")
+
+        result_symbol = ir_node.sym
```
blaine-rister, line 769: TODO: test dynamic shapes

torch/_inductor/codegen/wrapper_fxir.py
blaine-rister: I had difficulty constructing a test which would trigger this path, so I created https://github.com/pytorch/pytorch/pull/165954 to break the corresponding logic in the Python backend and see which tests failed. It seems that no tests exercise this path, so I decided to delete it from this PR.

PR #165616, opened 2025-10-16
[vmap] support chunk_with_scan for vmap

ydwu4:
* __->__ #165616
* #165580
* #165675


torch/_functorch/vmap.py
```diff
@@ -258,26 +259,95 @@ def _get_name(func: Callable):
...
+
+            num_chunks = batch_size // chunk_size
+
+            def _normalize_dims(in_dims):
+                if isinstance(in_dims, int) or in_dims is None:
+                    assert len(args) == 1
+                    in_dims = (in_dims,)
+                return in_dims
+
+            in_dims: tuple[Optional[int], ...] = _normalize_dims(in_dims)
```
zou3519, line 287: there's a broadcast function somewhere for broadcast semantics. can do in follow-up (shouldn't be difficult)

torch/_functorch/vmap.py
```diff
@@ -258,26 +259,95 @@ def _get_name(func: Callable):
...
+
+            chunked_args = tree_map(
+                lambda t, in_dim: t.unsqueeze(0).expand(num_chunks, *t.shape)
+                if in_dim is None
+                else t.view(
+                    num_chunks,
+                    *[s if i != in_dim else -1 for i, s in enumerate(t.shape)],
```
zou3519: let's just toss in a chunk_size

zou3519: NB: I'm not sure this works, let's talk offline

PR #165625, opened 2025-10-16
[AMD][gfx1100] test_decompose_mem_bound_mm.py tolerance increase

k-artem:
test_decompose_mem_bound_mm.py tolerance increase for navi3x(gfx11x)

(cherry picked from commit 03c7da05f61890bbf5ae41e23c8df6d5f6805bac) from 

Fixes for CI HUD for gfx1100

Signed-off-by: Artem Kuzmitckii <artem.kuzmitckii@amd.com>

### DIFF
--- a/test/inductor/test_decompose_mem_bound_mm.py
+++ b/test/inductor/test_decompose_mem_bound_mm.py
@@ -12,9 +12,9 @@ from torch._inductor.utils import run_and_get_code
 from torch.testing import FileCheck
 from torch.testing._internal.common_utils import (
     instantiate_parametrized_tests,
-    patch_test_members,
     is_navi3_arch,
     parametrize,
+    patch_test_members,
     TEST_XPU,
 )
 from torch.testing._internal.inductor_utils import GPU_TYPE, HAS_CUDA_AND_TRITON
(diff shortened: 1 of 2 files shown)


test/inductor/test_decompose_mem_bound_mm.py
```diff
@@ -71,31 +73,46 @@ def forward(
...
+    def __init__(self, method_name='runTest', methodName='runTest'):
+        super().__init__(method_name, methodName)
+        self.atol = 1e-3
+        self.rtol = 1e-3
+
+    def setup_tolerance(self, rtol=None, atol=None):
+        if rtol is None:
+            rtol = self.rtol
+        if atol is None:
+            atol = self.rtol
```
jeffdaily: How does this work?  rtol and atol local to the function.  Modifying them here should not affect call sites like line 88 below.  Right?

k-artem: 1) it works because `setattr(self, member, value)` inside decorator set correct values.
2) One more isssue found `atol = self.rtol` instead of `atol = self.atol`

I keep `setup_tolerance` (which actually can be removed) in case we need to update tolerance value at calls of `compare_*` functions.
Please let me know if I need to cp in back to ROCm fork or we will wait and get it via sync w/ upstream

jeffdaily (approved):

jithunnair-amd:
\easycla

PR #165627, opened 2025-10-16
[Code Clean] Clean asserts in torch/autograd.

mansiag05:
Replaces 78 assert statements across 10 files in torch.autograd with explicit if-checks raising AssertionError to prevent assertions from being disabled with Python -O flag. This ensures error checking remains active in optimized builds.

fix partially #164878


albanD (approved):
Looks great!

PR #165629, opened 2025-10-16
[Code Clean] Clean asserts in torch/optim.

mansiag05:
Replaces 50 assert statements across 15 files in torch.optim with explicit  if-checks raising AssertionError to prevent assertions from being disabled with Python -O flag.

fix partially #164878

### DIFF
--- a/torch/optim/_adafactor.py
+++ b/torch/optim/_adafactor.py
@@ -350,15 +350,15 @@ def _single_tensor_adafactor(
     maximize: bool,
     has_complex: bool,
 ):
-    assert grad_scale is None and found_inf is None, (
-        "Grad scaling should occur outside of optimizer.step()"
-    )
+    if grad_scale is not None or found_inf is not None:
+        raise AssertionError("Grad scaling should occur outside of optimizer.step()")
 
     if torch.jit.is_scripting():
         # this assert is due to JIT being dumb and not realizing that the ops below
         # have overloads to handle both float and Tensor lrs, so we just assert it's
         # a float since most people using JIT are using floats
-        assert isinstance(lr, float)
+        if not isinstance(lr, float):
+            raise AssertionError("Expected lr to be a float")
     else:
         lr = _to_scalar(lr)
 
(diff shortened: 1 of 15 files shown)


torch/optim/_adafactor.py
```diff
@@ -350,15 +350,15 @@ def _single_tensor_adafactor(
...
+    if grad_scale is not None or found_inf is not None:
+        raise AssertionError("Grad scaling should occur outside of optimizer.step()")
 
     if torch.jit.is_scripting():
         # this assert is due to JIT being dumb and not realizing that the ops below
         # have overloads to handle both float and Tensor lrs, so we just assert it's
         # a float since most people using JIT are using floats
-        assert isinstance(lr, float)
+        if not isinstance(lr, float):
+            raise AssertionError("Expected lr to be a float")
```
albanD: Printing the type would be good

torch/optim/adam.py
```diff
@@ -366,15 +366,19 @@ def _single_tensor_adam(
...
         # a float since most people using JIT are using floats
-        assert isinstance(lr, float)
-        assert isinstance(beta1, float)
-        assert isinstance(beta2, float)
+        if not isinstance(lr, float):
+            raise AssertionError("Expected lr to be a float")
+        if not isinstance(beta1, float):
+            raise AssertionError("Expected beta1 to be a float")
+        if not isinstance(beta2, float):
+            raise AssertionError("Expected beta2 to be a float")
```
albanD: show types for these

torch/optim/lbfgs.py
```diff
@@ -331,7 +332,8 @@ def step(self, closure):  # type: ignore[override]
             closure (Callable): A closure that reevaluates the model
                 and returns the loss.
         """
-        assert len(self.param_groups) == 1
+        if len(self.param_groups) != 1:
+            raise AssertionError("Expected exactly one param_group")
```
albanD: print the current len

albanD (commented):
A couple small nits, but sgtm

albanD (approved):
Thanks!

PR #165633, opened 2025-10-16
[Fix] Use sys.executable instead of hardcoded python

lingebeng:
Replace hardcoded "python" string with sys.executable to ensure correct Python interpreter is used. This fixes failures on systems with multiple Python runtimes or where "python" is not in PATH.

Similar to pytorch/pytorch#155918


Skylion007 (dismissed):
runpy stdlib is probably the proper way to do it, but this is definitely better.

lingebeng:
@Skylion007 Could you review the reason why the CI is failed. This hard code should to be fixed anyway.

PR #165642, opened 2025-10-16
fix fr reset api

tushar00jain:
Summary:
- there are various places that access fr's `entries_` field
- if we empty the entries_ on reset, the accesses can result in an error
- so we only perform a soft delete instead of clearing out the entries copletely
  - only reset id_ on the reset
  - keep track of a reset_epoch which increments everytime reset is called
  - dump_entries only returns entries from the latest epoch
  - api's that access entries also check if the reset epoch matches
- make the `next_` always track the index in the circular buffer - this change was needed to make the soft delete's implementation easier

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/pytorch/pytorch/pull/165642).
* #166103
* #166102
* __->__ #165642

### DIFF
--- a/test/distributed/test_c10d_nccl.py
+++ b/test/distributed/test_c10d_nccl.py
@@ -4607,6 +4607,229 @@ class NCCLTraceTest(NCCLTraceTestBase):
 
     @requires_nccl()
     @skip_but_pass_in_sandcastle_if(not TEST_MULTIGPU, "NCCL test requires 2+ GPUs")
+    @parametrize("timing_enabled", [True, False])
+    def test_fr_record_reset_circular_buffer_full(self, timing_enabled):
+        """
+        Test that when the circular buffer in entries_ is full and we call reset,
+        then fill the buffer with new entries, dump_entries returns only the new
+        entries and not the old ones.
+        """
+        if self.rank == self.MAIN_PROCESS_RANK:
+            return
+
+        # Override buffer size to 10 for faster testing
+        os.environ["TORCH_NCCL_TRACE_BUFFER_SIZE"] = "10"
+
+        pg = self._create_process_group_nccl()
+        if timing_enabled:
+            pg._enable_collectives_timing()
+        device = self.local_device
+        self.set_thread_name("fr_test_thread")
+        a = torch.full((3, 4), float(self.rank), device=device)
+
+        # Fill the buffer completely with 10 entries
+        for _ in range(10):
+            f = pg.allreduce(a)
+        f.wait()
+        torch.cuda.synchronize(device=device)
+        time.sleep(1)
+
+        # Verify buffer is full with 10 entries
+        t = pickle.loads(torch._C._distributed_c10d._dump_nccl_trace())
+        self.assertEqual(len(t["entries"]), 10)
+
+        # Now reset the flight recorder
+        torch._C._distributed_c10d._reset_fr_recording_nccl()
+
+        # Add new entries after reset - fill the buffer completely again
+        for _ in range(10):
+            f = pg.allreduce(a)
+        f.wait()
+        torch.cuda.synchronize(device=device)
+        time.sleep(1)
+
+        # Verify we get exactly 10 new entries, not 20
+        t = pickle.loads(torch._C._distributed_c10d._dump_nccl_trace())
+        self.assertEqual(len(t["entries"]), 10)
+
+        # Verify all entries have the expected properties (from after reset)
+        # After reset, record IDs should start from 0 again
+        for i, entry in enumerate(t["entries"]):
+            self.assertIn("profiling_name", entry)
+            self.assertEqual(entry["profiling_name"], "nccl:all_reduce")
+            self.assertIn("record_id", entry)
+            # Record IDs should be sequential starting from 0 after reset
+            self.assertEqual(entry["record_id"], i)
+
+        dist.destroy_process_group()
+
+    @requires_nccl()
+    @skip_but_pass_in_sandcastle_if(not TEST_MULTIGPU, "NCCL test requires 2+ GPUs")
+    @parametrize("timing_enabled", [True, False])
+    def test_fr_record_reset_partial_overwrite(self, timing_enabled):
+        """
+        Test that when the circular buffer is full, we reset, and then add fewer
+        entries than the buffer size, we only get the new entries.
+        This tests that old entries at the end of the circular buffer are properly
+        filtered out based on reset_epoch.
+        """
+        if self.rank == self.MAIN_PROCESS_RANK:
+            return
+
+        # Override buffer size to 10 for faster testing
+        os.environ["TORCH_NCCL_TRACE_BUFFER_SIZE"] = "10"
+
+        pg = self._create_process_group_nccl()
+        if timing_enabled:
+            pg._enable_collectives_timing()
+        device = self.local_device
+        self.set_thread_name("fr_test_thread")
+        a = torch.full((3, 4), float(self.rank), device=device)
+
+        # Fill the buffer completely
+        for _ in range(10):
+            f = pg.allreduce(a)
+        f.wait()
+        torch.cuda.synchronize(device=device)
+        time.sleep(1)
+
+        # Reset the flight recorder
+        torch._C._distributed_c10d._reset_fr_recording_nccl()
+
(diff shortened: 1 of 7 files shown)


torch/csrc/distributed/c10d/FlightRecorderDetail.hpp
fduwjj: Can we do the filter here?

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
```diff
@@ -4182,6 +4185,8 @@ c10::intrusive_ptr<Work> ProcessGroupNCCL::pointToPoint(
         options_->timeout,
         pgStatus_,
         /*isP2P=*/true);
+    work->trace_id_ = traceId.id;
+    work->trace_reset_epoch_ = traceId.reset_epoch;
```
fduwjj, line 4189: not a deal break, I think we can just record trace_reset_epoch_ in the PGNCCL and then when calling retire, if it is not in the current reset epoch, we just make it a no-op. But I don't have a strong opinion, so up to you.

torch/csrc/distributed/c10d/FlightRecorderDetail.hpp
```diff
@@ -214,8 +248,8 @@ void FlightRecorder<EventType>::retire_id(
 
   std::unique_lock<std::mutex> guard(mutex_);
 
-  Entry* entry = &entries_.at(*id % max_entries_);
-  if (entry->id_ == *id) {
+  Entry* entry = &entries_.at(getIdxFromId(*id, *reset_epoch));
+  if (entry->id_ == *id && entry->reset_epoch_ == *reset_epoch) {
```
fduwjj, line 247: see here if reset_epoch not the same it is already a no-op.

torch/csrc/distributed/c10d/FlightRecorderDetail.hpp
```diff
@@ -178,32 +187,57 @@ std::vector<typename FlightRecorder<EventType>::Entry> FlightRecorder<
...
-// std::nullopt.
+// Returns the index in entries_ for the given id and reset_epoch.
+// Caller must hold mutex_lock before calling this method.
+size_t FlightRecorder<EventType>::getIdxFromId(size_t id, size_t reset_epoch)
+    const {
+  // Look up the starting idx for the given reset epoch
+  auto it = reset_epoch_start_idx_.find(reset_epoch);
+  TORCH_CHECK(it != reset_epoch_start_idx_.end());
+  // Calculate idx based on where the epoch started
+  return (it->second + id) % max_entries_;
```
fduwjj, line 209: I think this is wrong? because you reset id to 0, we should not consider reset_epoch anymore? I thought reset_epoch is only for dump and retire_id. When you try to get entry, you can just directly get that entry using the index?

fduwjj:
Thanks for making FR working for TorchFT. Can we mention the context and rationale on why we are doing this in the PR description? Like because we have restart and we want to reset FR without race condition so that we want to do soft-reset?

Also I have questions on the correctness of that getting entry. Maybe you can clarify a bit more? Also you can write a unit test (we already have one in test_c10d_nccl) to validate your change as well.

torch/csrc/distributed/c10d/FlightRecorderDetail.hpp
tushar00jain: > because you reset id to 0

that's exactly why we need this logic i think? we store the index in the hashmap when we reset the id to 0, so we need to add it back.

tushar00jain:
@fduwjj 

> like because we have restart and we want to reset FR without race condition so that we want to do soft-reset?

this is the first 2 bullet points?

> Also you can write a unit test (we already have one in test_c10d_nccl) to validate your change as well.

added

torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp
tushar00jain: it's already going to be a noop when we call retire with an old epoch, think it's better to only capture this logic inside flight recorder and not expose the details to pg specially because we have multiple pg's and all of them will need to implement the same logic

torch/csrc/distributed/c10d/FlightRecorderDetail.hpp
tushar00jain: yep i think better to keep it a noop inside flight recorder so we don't have to implement the same logic in all pg's

torch/csrc/distributed/c10d/FlightRecorderDetail.hpp
```diff
@@ -64,8 +65,13 @@ std::optional<size_t> FlightRecorder<EventType>::record(
       torch::CapturedTraceback::gather(true, true, capture_cpp_stack_);
   std::lock_guard<std::mutex> guard(mutex_);
 
+  TORCH_CHECK(
+      reset_epoch_start_idx_.find(reset_epoch_) !=
```
fduwjj, line 69: what would happen for non-reset case?

torch/csrc/distributed/c10d/FlightRecorderDetail.hpp
```diff
@@ -104,15 +110,20 @@ std::optional<size_t> FlightRecorder<EventType>::record(
...
+    entries_[next] = std::move(te);
   }
-  return id_++;
+
+  if (next_ == max_entries_) {
+    next_ = 0;
+  }
+
+  const auto id = id_++;
+  return TraceIdentifier{id, reset_epoch_};
```
fduwjj, line 126: why can't we directly pass in id_++ to TraceIdentifier?

torch/csrc/distributed/c10d/FlightRecorderDetail.hpp
```diff
@@ -104,15 +110,20 @@ std::optional<size_t> FlightRecorder<EventType>::record(
...
-    if (next_ == max_entries_) {
-      next_ = 0;
-    }
+    entries_[next] = std::move(te);
   }
-  return id_++;
+
+  if (next_ == max_entries_) {
+    next_ = 0;
+  }
```
fduwjj, line 123: I don't think you need to move this?

test/distributed/test_c10d_nccl.py
```diff
@@ -4626,6 +4626,229 @@ def test_fr_record_reset(self, timing_enabled):
...
+    def test_fr_record_multiple_resets(self, timing_enabled):
+        """
+        Test multiple consecutive resets to ensure each reset properly increments
+        the epoch and filters out entries from previous epochs.
+        """
+        if self.rank == self.MAIN_PROCESS_RANK:
+            return
+
+        # Override buffer size to 10 for faster testing
+        os.environ["TORCH_NCCL_TRACE_BUFFER_SIZE"] = "10"
```
fduwjj, line 4784: nit: unset this ENV in the end so that it does not interfere with other unit tests.

fduwjj (approved):
I think this will work, just some nit in torch/csrc/distributed/c10d/FlightRecorderDetail.hpp. Thanks!

torch/csrc/distributed/c10d/FlightRecorderDetail.hpp
fduwjj: NVM I think we handled that well.

torch/csrc/distributed/c10d/FlightRecorderDetail.hpp
tushar00jain: this is needed, the pr changes how `next_` is tracked. we use it to track the index in the buffer at all times. we do this so we can create the mapping from reset epoch to the index in the buffer

torch/csrc/distributed/c10d/FlightRecorderDetail.hpp
tushar00jain: the increment operators always confuse me, it's more clear this way i think

test/distributed/test_c10d_nccl.py
tushar00jain: hmm seems other unit tests also don't reset, will just stick to the existing convention

PR #165645, opened 2025-10-16
Document limitations of weights_only in SECURITY.md and torch.load doc

mikaylagawarecki:
* __->__ #165645

cc @albanD


torch/serialization.py
```diff
@@ -1296,6 +1296,19 @@ def load(
 
     Loads an object saved with :func:`torch.save` from a file.
 
+    .. warning::
+        :func:`torch.load()` unless ``weights_only`` parameter is set to ``True``,
+        uses ``pickle`` module implicitly, which is known to be insecure.
+        It is possible to construct malicious pickle data which will execute arbitrary code
+        during unpickling. Never load data that could have come from an untrusted
```
albanD:
```suggestion
        during unpickling. Never load data from an untrusted
```

torch/serialization.py
```diff
@@ -1296,6 +1296,19 @@ def load(
...
+        source in an unsafe mode, or that could have been tampered with. **Only load data you trust**.
```
albanD:
```suggestion
        source, or that could have been tampered with. **Only load data you trust**.
```

torch/serialization.py
```diff
@@ -1296,6 +1296,19 @@ def load(
...
+
+        ``weights_only=True`` narrows the surface of remote code execution attacks via a more
+        restricted unpickler but has the following limitations:
+
+        1. ``weights_only=True`` does not guard against denial of service attacks.
+        2. ``weights_only=True`` safeguards against memory corruption are best effort.
```
albanD: We did actually audit and remove memory corruption during the loading process?
I would suggest:
- There are no RCE or memory corruption during the loading process.
- There are STILL denial of service attack during the loading process.
- Loading CAN create un-expected objects for the downstream code that can still lead to memory corruption (a Tensor of indices, made in to a sparse Tensor in the user code, can then write/read out of bound).
Thus, even with weights_only=True, we do not recommend loading untrusted checkpoints without extensive validation of the objects being loaded.

SECURITY.md
```diff
@@ -31,9 +31,12 @@ Be careful when running untrusted models. This classification includes models cr
...
+**Be mindful of risky model formats**. Give preference to share and load weights with the appropriate format for your use case. [safetensors](https://huggingface.co/docs/safetensors/en/index) gives the most safety but is the most restricted in what it supports. [`torch.load`](https://pytorch.org/docs/stable/generated/torch.load.html#torch.load) with `weights_only=True` reduces the surface of remote code execution attacks but offers a significantly larger surface of attack than safetensors. In particular:
 
+1. `weights_only=True` does not guard against denial of service attacks.
+2. `weights_only=True` safeguards against memory corruption are best effort.
```
albanD: To update like the one above

SECURITY.md
```diff
@@ -31,12 +31,9 @@ Be careful when running untrusted models. This classification includes models cr
...
 
-**Be mindful of risky model formats**. Give preference to share and load weights with the appropriate format for your use case. [safetensors](https://huggingface.co/docs/safetensors/en/index) gives the most safety but is the most restricted in what it supports. [`torch.load`](https://pytorch.org/docs/stable/generated/torch.load.html#torch.load) with `weights_only=True` reduces the surface of remote code execution attacks but offers a significantly larger surface of attack than safetensors. In particular:
+**Be mindful of risky model formats**. Give preference to share and load weights with the appropriate format for your use case. [safetensors](https://huggingface.co/docs/safetensors/en/index) gives the most safety but is the most restricted in what it supports. [`torch.load`](https://pytorch.org/docs/stable/generated/torch.load.html#torch.load) has a significantly larger surface of attack than safetensors. See the documentation for more details.
```
albanD:
```suggestion
**Be mindful of risky model formats**. Give preference to share and load weights with the appropriate format for your use case. [safetensors](https://huggingface.co/docs/safetensors/en/index) gives the most safety but is the most restricted in what it supports. [`torch.load`](https://pytorch.org/docs/stable/generated/torch.load.html#torch.load) has a significantly larger surface of attack but is more flexible in what it can serialize. See the documentation for more details.
```

albanD (approved):
One small nit, but SGTM !

PR #165646, opened 2025-10-16
Update test jobs in pull workflow to c7i

zxiiro:
Relates to pytorch/test-infra#7175. While moving to c7i costs 5% more, CPU intensive jobs should run roughly 15-20% faster resulting in a cost reduection of 10-15% for those jobs.

This PR updates for the following test job suite that seem to benefit from the newer hardware:

* backwards_compat
* numpy_2_x
* ONNX default


jordanconway (approved):
Can confirm I've seen @zxiiros timing tests behind this. Seems like an easy cost saving win.

huydhn (approved):
LGTM!

PR #165651, opened 2025-10-16
Use union syntax in torch/compiler

oulgen:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* #174839
* __->__ #165651


torch/compiler/__init__.py
```diff
@@ -1,7 +1,8 @@
 # mypy: allow-untyped-defs
-import io
+from __future__ import annotations
```
aorenste: Was this necessary? annotations is deprecated so we're trying to avoid adding more...

aorenste (approved):

torch/compiler/__init__.py
oulgen: `"Foo" | None` is not supported, so this is the only way

malfet (approved):

torch/compiler/__init__.py
aorenste: FYI: You can do "Foo | None"

PR #165663, opened 2025-10-16
mark fbgemm as torch files for hipify purposes

jeffdaily:


jeffdaily:
Diff added elsewhere.

PR #165665, opened 2025-10-16
[CD] Skip 12.9 build on Windows

huydhn:
Per title

cc @seemethere @malfet @pytorch/pytorch-dev-infra @albanD


Camyll (approved):

malfet (approved):

.github/scripts/generate_binary_build_matrix.py
```diff
@@ -241,6 +241,9 @@ def generate_libtorch_matrix(
             arches += CUDA_ARCHES
             arches += ROCM_ARCHES
         elif os == "windows":
+            # TODO (huydhn): Only build CUDA 12.9 for Linux. This logic is to be cleaned up
+            # in 2.10
+            CUDA_ARCHES.remove("12.9")
```
malfet: This can have unintended consequences (as it modifies global variable)
Better to just add 12.9 to local var if Linux

huydhn: Yeah.  Let me do a copy and modify that copy instead, that should work.  The reason why I don't just add `12.9` to Linux is that there are a couple of other global variables that we initialize using CUDA_ARCHS like WHEEL_CONTAINER_IMAGES.  So, I just pick the solution with the least amount of code change here as the change is temporary

PR #165679, opened 2025-10-16
[BE][Ez]: Use sys.executable instead of hardcoded Python

Skylion007:
Handles edgecase to ensure proper interpreter is called. Inspired by #165633


FindHao (approved):
LGTM

PR #165680, opened 2025-10-16
[DeviceMesh] Enable mesh universe concept in mesh comparison

fduwjj:
* __->__ #165680

Since we now use the same _rank_map everywhere, we can just use it to differentiate between different mesh universe and get rid of root mesh comparison.


lw (approved):
Yes please!!! So happy we can finally do this! Go!!

PR #165683, opened 2025-10-16
[dynamo] Use Variable Builder to build the property fget object

anijain2305:
* #165708
* #165707
* #165706
* __->__ #165683

### DIFF
--- a/torch/_dynamo/variables/user_defined.py
+++ b/torch/_dynamo/variables/user_defined.py
@@ -1458,12 +1458,8 @@ class UserDefinedObjectVariable(UserDefinedVariable):
                 # Get the getter function
                 source = AttrSource(source, "fget")
 
-            # Avoid using UserMethodVariable here because there is no way to
-            # access the method object here. Direct inline by creating the
-            # UserFunctionVariable.
-            return variables.UserFunctionVariable(
-                subobj.fget, source=source
-            ).call_function(tx, [self], {})
+            fget_vt = VariableTracker.build(tx, subobj.fget, source=source)
+            return fget_vt.call_function(tx, [self], {})
         elif isinstance(subobj, _collections._tuplegetter):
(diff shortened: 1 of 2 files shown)


ezyang (approved):
Thanks can you also check functools.partial works?

torch/_dynamo/variables/user_defined.py
```diff
@@ -1458,12 +1458,8 @@ def var_getattr(self, tx: "InstructionTranslator", name):
...
                 source = AttrSource(source, "fget")
 
-            # Avoid using UserMethodVariable here because there is no way to
-            # access the method object here. Direct inline by creating the
-            # UserFunctionVariable.
-            return variables.UserFunctionVariable(
-                subobj.fget, source=source
-            ).call_function(tx, [self], {})
+            fget_vt = VariableTracker.build(tx, subobj.fget, source=source)
+            return fget_vt.call_function(tx, [self], {})
```
ruomingp, line 1462: Does this also work with functools.partial?

williamwen42 (approved):

torch/_dynamo/variables/user_defined.py
ruomingp: UserFunctionVariable is also used in a few other places: https://github.com/search?q=repo%3Apytorch%2Fpytorch+UserFunctionVariable%28&type=code. Do we have a similar issue in some of those?

ruomingp (commented):
Thanks for the quick fix!

torch/_dynamo/variables/user_defined.py
anijain2305: Yes, we need an audit of other places, where we should directly use the variable builder. I will do that.

I will add a test for the functools.partial in the followup PR.

Update - Added the functools.partial test

PR #165689, opened 2025-10-16
Generalize the testcase for a given world size

newtdms:
Currently, the test case fails when more than 4-ranks are used with the following error:
NotImplementedError: Cannot copy out of meta tensor; no data!

The patch keeps existing functionality of supporting four-ranks or devices, in addition, make the test case more generic for a given world_size, e.g. 8-ranks.

Fixes internal PyTorch issue no 7192 


guangyey:
It is expected, please refer to https://github.com/pytorch/pytorch/blob/74e53d0761c0b34a720c59c59a433a0bfc9e4bf4/test/distributed/_composable/fsdp/test_fully_shard_state_dict.py#L37-L39

PR #165694, opened 2025-10-16
Move toUnderlying to headeronly

pearu:
As in the title. Required in upper PRs of this ghstack.

* #165695
* #165856
* __->__ #165694


torch/headeronly/core/ScalarType.h
```diff
@@ -300,10 +300,26 @@ inline const char* toString(ScalarType t) {
 
 inline std::ostream& operator<<(
     std::ostream& stream,
-    at::ScalarType scalar_type) {
+    c10::ScalarType scalar_type) {
```
janeyx99, line 303: Why this change?

janeyx99 (approved):
LGTM thanks

torch/headeronly/core/ScalarType.h
pearu: It just addresses the oversight that ought to been included in https://github.com/pytorch/pytorch/pull/164458 . While `at::ScalarType` and `c10::ScalarType` are the same, there is no reason to have old `at` dependency in headeronly.

PR #165323, opened 2025-10-13
Move AT_DISPATCH_..._TYPES... macros to headeronly and add tests for the macros.

pearu:
As in the title.

Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #165323
* #164669

cc @jbschlosser

PR #165695, opened 2025-10-16
Refactor AT_DISPATCH_CASE_... macros to headeronly

pearu:
`AT_DISPATCH_CASE_...` macros are refactored by moving torch library independent parts to `THO_DISPATCH_CASE_...` macros so that one could implement headeronly or stable `DISPATCH_..._TYPES` macros that are semantically equivalent to `AT_DISPATCH_..._TYPES` macros with respect to dtypes mappings.

The new `AT_DISPATCH_CASE_...` macro definitions as semantically equivalent to their old definitions.

This PR implements an alternative approach to the stack https://github.com/pytorch/pytorch/pull/165323.

* __->__ #165695
* #165856
* #166062


malfet (commented):
I would probably migrate only v2 version of the macros there, to encourage use of new APIs

Also, not sure what's the benefit of renaming `AT_DISPATCH_XYZ` with `THO_DISPATCH_XYZ`...

pearu:
> > I would probably migrate only v2 version of the macros there, to encourage use of new APIs
> 
> What are v2 version of `AT_DISPATCH...` macros?

Found it in aten/src/ATen/Dispatch_v2.h, it's nice!

Notice that V2 macros relay on AT_DISPATCH_SWITCH/CASE macros from V1, so the problem of primary and secondary features remains.

Also, V2 does not appear to support dispatching that correspond to V1 `AT_DISPATCH_QINT_AND_SUB_BYTE_TYPES` and `AT_DISPATCH_QINT_TYPES` macros.

pearu:
> > Also, not sure what's the benefit of renaming `AT_DISPATCH_XYZ` with `THO_DISPATCH_XYZ`...
> 
> It is just not a rename. Notice that `THO_DISPATCH_` macros have one or two new leading arguments depending on a specific macro, for example:
> 
> ```c++
> #define AT_DISPATCH_CASE_FLOATING_TYPES(...) \
>   THO_DISPATCH_CASE_FLOATING_TYPES(AT_DISPATCH_CASE, __VA_ARGS__)
> ```

I suggest renaming `THO_DISPATCH_CASE_XYZ` macros to `AT_DISPATCH_CASE_XYZ_TMPL` so that it's meaning is better highlighted and this avoids introducing a new macro prefix `THO_`. What do you think?

PR #165701, opened 2025-10-16
[pipeline][be] refactored pipeline composability tests

anshul-si:
**Summary:** The first thing I did was increase the world size to 8 because test_3d_with_tp_dp_pp wouldn't actually do fully shard as tp = 2, pp = 2, leaving dp = 1. The second thing was refactoring the tests using both single and multi stage schedules so that their logic is largely combined. This was accomplished by using the logic in test_replicate_pp_grad multi-stage schedule to determine the start and end indices for a partial model, but setting virtual_stage to 1 if we are using single stage schedules. Even if this approach isn't approved, multistage schedule logic in test_3d_with_tp_dp_pp and test_replicate_pp should be changed as the logic used is incorrect. 

**Test Case**
1. pytest test/distributed/_composable/test_composability/test_pp_composability.py

* __->__ #165701

### DIFF
--- a/test/distributed/_composable/test_composability/test_pp_composability.py
+++ b/test/distributed/_composable/test_composability/test_pp_composability.py
@@ -377,7 +377,7 @@ class ComposabilityTest(MultiProcessTestCase):
             torch.float32,
         ],
     )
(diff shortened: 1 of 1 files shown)


H-Huang (approved):
LGTM, thanks!

PR #165703, opened 2025-10-16
[annotate] add annotate_fn function decorator

yushangdi:
Example usage:

```
        @fx_traceback.annotate_fn({"pp_stage": 1})
        def example_function(x):
            return x * x

        class SimpleLinear(nn.Module):
            def __init__(self):
                super().__init__()
                self.linear = nn.Linear(3, 2)

            def forward(self, x):
                    y = self.linear(x)
                y = example_function(y)
                return y - 1
```

### DIFF
--- a/torch/fx/traceback.py
+++ b/torch/fx/traceback.py
@@ -18,6 +18,7 @@ log = logging.getLogger(__name__)
 
 __all__ = [
     "annotate",
+    "annotate_fn",
     "preserve_node_meta",
     "has_preserved_node_meta",
     "set_stack_trace",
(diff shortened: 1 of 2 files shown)


SherlockNoMad (approved):
thanks.

torch/fx/traceback.py
```diff
@@ -291,6 +293,43 @@ def annotate(annotation_dict: dict):
             del current_meta["custom"]
 
 
+@compatibility(is_backward_compatible=False)
+def annotate_fn(annotation_dict: dict):
```
Skylion007, line 297: Are the dict keys always strings? If so, that should be specified. Also we need to specify it returns a decorator, or a Callable that returns a Callable preserving the args, kwargs, and return type.

PR #165710, opened 2025-10-17
Fix #114629: ensure CPU DeviceMesh initializes gloo process groups

bbopen:
## Summary
- make `init_device_mesh("cpu", …)` pass `backend="gloo"` when it has to bootstrap distributed
- create gloo subgroups whenever the default process group lacks CPU support so CPU meshes can reuse DeviceMesh APIs on CUDA builds without GPUs
- document the mocked tests that exercise this edge case

## Testing
- python test/distributed/test_device_mesh.py -k DeviceMeshCPUFallbackTest -v
- ```python
  import os
  import tempfile
  import torch
  import torch.multiprocessing as mp

  def worker(rank, world_size, init_file):
      dist.init_process_group("gloo", rank=rank, world_size=world_size,
                              init_method=f"file://{init_file}")
      mesh = dist.init_device_mesh(device_type="cpu", mesh_shape=(world_size,))
      group = mesh.get_group()
      inp = torch.tensor([rank], dtype=torch.float32)
      out = torch.empty(world_size, dtype=torch.float32)
      dist.all_gather_into_tensor(out, inp, group=group)
      if rank == 0:
          print("all_gather result", out.tolist())
      dist.destroy_process_group()

  if __name__ == "__main__":
      world_size = 2
      fd, path = tempfile.mkstemp()
      os.close(fd)
      try:
          mp.spawn(worker, args=(world_size, path), nprocs=world_size, join=True)
      finally:
          if os.path.exists(path):
              os.remove(path)
  ```

## Not tested
- GPU/NCCL DeviceMesh coverage (will rely on CI or a separate GPU run)

Fixes #114629

### DIFF
--- a/torch/distributed/device_mesh.py
+++ b/torch/distributed/device_mesh.py
@@ -272,10 +272,10 @@ else:
             # TODO: think about how to allow pg options to be passed to world group
             # or mesh dimension groups
             if not default_initialized:
-                init_pg_kwargs: dict[str, object] = {}
+                backend_override: Optional[str] = None
                 if self._device_type == "cpu" and not torch.cuda.is_available():
-                    init_pg_kwargs["backend"] = "gloo"
-                init_process_group(**init_pg_kwargs)
+                    backend_override = "gloo"
+                init_process_group(backend=backend_override)
 
             world_size = get_world_size()
             if self.mesh.numel() > world_size:
@@ -350,7 +350,10 @@ else:
                             ranks=ranks,
                             group_desc="mesh_default",
                         )
-                    elif torch.cuda.is_available() and get_backend(default_group) == "gloo":
+                    elif (
+                        torch.cuda.is_available()
+                        and get_backend(default_group) == "gloo"
+                    ):
                         dim_group = new_group(
                             backend="cpu:gloo,cuda:nccl",
                             ranks=ranks,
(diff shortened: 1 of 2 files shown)


torch/distributed/device_mesh.py
```diff
@@ -335,16 +339,36 @@ def _init_process_groups(
...
+                        and get_backend(default_group) == "gloo"
+                    ):
+                        dim_group = new_group(
+                            backend="cpu:gloo,cuda:nccl",
+                            ranks=ranks,
+                            group_desc="mesh_default",
+                        )
+                    else:
+                        dim_group = default_group
+                elif torch.cuda.is_available() and get_backend(default_group) == "gloo":
```
Skylion007, line 364: Can we make this accelerator agnostic somehow? What about ROCM/XPU?

torch/distributed/device_mesh.py
```diff
@@ -272,7 +272,10 @@ def _setup_world_group_and_device(self):
             # TODO: think about how to allow pg options to be passed to world group
             # or mesh dimension groups
             if not default_initialized:
-                init_process_group()
+                backend_override: Optional[str] = None
+                if self._device_type == "cpu" and not torch.cuda.is_available():
```
Skylion007, line 276: This should be generic for torch.accelerator right? Not just cuda

PR #165711, opened 2025-10-17
(1/3)[DTensor][AllPermute] Support  over multi dimension mesh

zpcore:
Update `permute_tensor` to support permute a multi dimension tensor on first dimension. In addition, we should be able to permute over a multi dimension device mesh.

* #165712
* __->__ #165711

### DIFF
--- a/torch/distributed/_functional_collectives.py
+++ b/torch/distributed/_functional_collectives.py
@@ -466,11 +466,11 @@ def all_to_all_single(
     """
     if output_split_sizes is not None:
         assert all(
-            isinstance(size, (int, torch.SymInt)) for size in output_split_sizes
+            isinstance(size, int | torch.SymInt) for size in output_split_sizes
         ), output_split_sizes
     if input_split_sizes is not None:
         assert all(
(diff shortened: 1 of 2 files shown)


torch/distributed/_functional_collectives.py
```diff
@@ -722,14 +721,19 @@ def cast_listint(x):
...
+            rankset = dist.get_process_group_ranks(pg)
+            group_size = len(rankset)
+            tag = tag or c10d._get_group_tag(pg)
+        else:
+            # For multi-dimensional mesh, use all ranks in the global mesh
+            all_ranks = group.mesh.flatten().tolist()
+            rankset = all_ranks
+            group_size = len(rankset)
+            # Create a tag that represents the entire mesh
+            tag = tag or f"global_mesh_{id(group)}"
```
fegin, line 736: This can be problematic. How do you know if `global mesh` is actually created? Like if the input mesh is a submesh of a world mesh, this flatten PG tag is likely to hit an uncreated one.

fegin (changes requested):

PR #165712, opened 2025-10-17
(2/3)[DTensor][AllPermute] Implement AllPermute collective op for ordered shard

zpcore:
### Summary
Introduce the `AllPermute` collective operation as mentioned in https://arxiv.org/pdf/2112.01075 "section 2.6 Collective operations".  

### What is AllPermute?
AllPermute can transform any 𝜏1 to 𝜏2 if their local and global shapes match. For example:
Given mesh and size {X:4, Y:4, Z:16}, we have
- example 1: [32{X,Y}}512, 128] -> [32{Y,X}512, 128]
- example 2: [128{Y}512, 32{X}128] -> [128{X}512, 32{Y}128]
- example 3: [32{X,Y}512, 128] -> [32{Z}512, 128]
Note: annotation borrowed from https://arxiv.org/pdf/2112.01075 "section 2.1 Distributed array types"

### Why we need AllPermute?
With AllPermute, we can eliminate some AllGather ops during redistribution. This plays an important role in reducing the memory overhead. In theory, at most one AllPermute is needed to redistribute from any 𝜏1 to 𝜏2. The `AllPermute` can be performed as the final step, or moved before the last `AllGather` to minimize the amount of data relocated between shards in the `AllPermute`.

### Algorithm
The function `all_permute_mesh_dim` determines, for a given local tensor shard (the data on the current rank), which target rank should receive this data when redistributing a tensor from a source sharding distribution to a target sharding distribution across a device mesh. It computes a permutation mapping between source and target ranks based on how tensor dimensions are sharded on mesh dimensions, and then uses this mapping to permute the data accordingly. This enables efficient redistribution of tensor shards without intermediate gather operations.

We use device coordinate as a bridge to compute where the data should be send to. Let's use the 3D mesh (size: [X, Y, U]) and 3D tensor (size: [I, J, K]) as the example:  
Assume data is located on device with coordinate:

 $$\tau_1 = (x, y, u)$$

(using tensor-centric format) When we do AllPermute from source distribution ``S(I)[X, Y]S(K)[U]`` to target distribution ``S(I)[U]S(K)[X,Y]`` , the data will be relocated from `(x, y, u)` to:

$$\tau_2 = (\frac{u\lceil \frac{K}{U} \rceil}{\lceil\frac{K}{X}\rceil},  \frac{u\lceil\frac{K}{U}\rceil(mod\ \lceil \frac{K}{X} \rceil)}{\lceil \frac{\lceil \frac{K}{X}\rceil}{Y} \rceil},   \frac{x\lceil\frac{I}{X}\rceil + y\lceil \frac{\lceil \frac{I}{X} \rceil}{Y}\rceil}{\lceil \frac{I}{U} \rceil})$$

Luckily the main logic can be implemented with ~10 lines of code. Hope the algorithm is self-explainable with the example :) 

* __->__ #165712
* #165711

### DIFF
--- a/torch/distributed/tensor/debug/_comm_mode.py
+++ b/torch/distributed/tensor/debug/_comm_mode.py
@@ -249,6 +249,7 @@ class CommDebugMode(TorchDispatchMode):
             self.comm_registry.add(py_op)
 
         self.comm_registry.add(torch.ops._dtensor.shard_dim_alltoall)
+        self.comm_registry.add(torch.ops._c10d_functional.all_to_all_single)
         self.advanced_module_tracker = _CommModeModuleTracker()
 
     def generate_json_dump(self, file_name="comm_mode_log.json", noise_level=3):
--- a/test/distributed/tensor/test_redistribute.py
+++ b/test/distributed/tensor/test_redistribute.py
@@ -16,7 +16,10 @@ from torch.distributed.tensor import (
     Replicate,
     Shard,
 )
-from torch.distributed.tensor._collective_utils import shard_dim_alltoall
+from torch.distributed.tensor._collective_utils import (
+    all_permute_mesh_dim,
+    shard_dim_alltoall,
+)
 from torch.distributed.tensor._dtensor_spec import ShardOrderEntry
 from torch.distributed.tensor._redistribute import redistribute_local_tensor
 from torch.distributed.tensor.debug import CommDebugMode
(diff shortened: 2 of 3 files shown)


zpcore:
I shall schedule a brief discussion meeting to ensure everyone is aligned with me and understand how the device coordinate mapping works.

torch/distributed/tensor/debug/_comm_mode.py
```diff
@@ -249,6 +249,7 @@ def __init__(self):
             self.comm_registry.add(py_op)
 
         self.comm_registry.add(torch.ops._dtensor.shard_dim_alltoall)
+        self.comm_registry.add(torch.ops._c10d_functional.all_to_all_single)
```
ezyang, line 252: What's going on here?

ezyang:
Looks like it's permute_tensor so shouldn't be bad

PR #165716, opened 2025-10-17
Enable more DTensor tests in local tensor mode and fix more integration issues

dzmitry-huba:
* __->__ #165716

- During op dispatch local tensor is supposed to collect rng state from CPU and CUDA
devices so that it can be reset before execution of the op for each such that ops
with randomness produces the same result for all ranks (note that we are planning a
separate change to add support of per rank rng state). Previously we relied on
op input arguments to deduce which devices to get rng state from. Which doesn't work
for factory functions such torch.randn. Hence this changes switches to uncondionally
collecting rng state from all devices.

- Fixing per rank specific computations in _MaskedPartial and Shard placements discovered
during test enablement.


ezyang (approved):

dzmitry-huba:
Fixed failure and attempting to merge again

@pytorchbot merge

PR #165729, opened 2025-10-17
Thread deterministic config vars to subproc compilation

drisspg:
* __->__ #165729

# Summary

TIL (AFTER WAYYYY TOO MUCH INSANITY), that we do not serialize the full set of configs for the subproc compilation. 

I found this while working on Flex-attention determinism: https://github.com/meta-pytorch/attention-gym/pull/168

might be good to audit if we need to thread through any more

### DIFF
--- a/torch/_inductor/async_compile.py
+++ b/torch/_inductor/async_compile.py
@@ -423,9 +423,7 @@ class AsyncCompile:
             env_vars = ["TORCHINDUCTOR_CACHE_DIR", "TRITON_CACHE_DIR"]
             extra_env = {v: os.environ[v] for v in env_vars if v in os.environ}
             extra_config = {
-                "use_static_cuda_launcher": torch._inductor.config.use_static_cuda_launcher,
-                "deterministic": torch._inductor.config.deterministic,
-                "test_configs.force_filter_reduction_configs": torch._inductor.config.test_configs.force_filter_reduction_configs,
+                "use_static_cuda_launcher": torch._inductor.config.use_static_cuda_launcher
             }
 
             if len(torch._inductor.config.autotune_lookup_table) > 0:
--- a/torch/_inductor/async_compile.py
+++ b/torch/_inductor/async_compile.py
@@ -423,7 +423,9 @@ class AsyncCompile:
             env_vars = ["TORCHINDUCTOR_CACHE_DIR", "TRITON_CACHE_DIR"]
             extra_env = {v: os.environ[v] for v in env_vars if v in os.environ}
             extra_config = {
-                "use_static_cuda_launcher": torch._inductor.config.use_static_cuda_launcher
+                "use_static_cuda_launcher": torch._inductor.config.use_static_cuda_launcher,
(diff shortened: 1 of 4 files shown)


torch/_inductor/runtime/benchmarking.py
```diff
@@ -59,6 +59,7 @@ def wrapper(
 
 
 def may_ban_benchmarking() -> None:
+    # raise RuntimeError("wtf")
```
drisspg: this is not needed

torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -2971,6 +2971,12 @@ def filter_reduction_configs_for_determinism(
     assert len(configs) > 0
 
     def _do_filter_due_to_inductor_config():
+        detmerinistic = inductor_meta.get("deterministic", False)
+        force_filter = torch._inductor.config.test_configs.force_filter_reduction_configs
+        detmerinistic_algos = inductor_meta.get("are_deterministic_algorithms_enabled")
+        print(f"deterministic: {detmerinistic}, force_filter: {force_filter}, deterministic_algos: {detmerinistic_algos}")
+        return True
```
drisspg: need to return True

torch/_inductor/async_compile.py
```diff
@@ -423,7 +423,9 @@ def reload_kernel_in_parent():
             env_vars = ["TORCHINDUCTOR_CACHE_DIR", "TRITON_CACHE_DIR"]
             extra_env = {v: os.environ[v] for v in env_vars if v in os.environ}
             extra_config = {
```
v0i0, line 425: might be a silly question, but why isn't this the full set?

drisspg: I had this exact question as well....

My guess is its a potentially expensive? But I doubt its really that expensive, not sure didn't measure

shunting314 (approved):

eellison (approved):

torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -2973,7 +2973,7 @@ def filter_reduction_configs_for_determinism(
     def _do_filter_due_to_inductor_config():
         return (
```
eellison, line 2974: Can we do anything defensively to avoid this happening again? E.g., factor out `torch._inductor.config` accesses in this file to a file that asserts that the attribute is in the fields that we patch to subprocess ?

drisspg: cc @shunting314 

I like this idea and agree we should do something

shunting314: I think it's not hard to go thru the file and make sure there is no directly access to `torch._inductor.config` (e.g. access them thru inductor_meta dict), but it would be hard to

1. avoid future code in this file from accessing inductor config directly
2. avoid the code this file rely on from accessing inductor config directly. There accesses are harder to find.

Alternatively, maybe we should just detect and patch every config overriding when submitting a compilation task to child process.

PR #165731, opened 2025-10-17
fix wrong accuracy_status when exception.

### LINKED ISSUE #165753: wrong accuracy_status when check_accuracy occur exception.
### 🐛 Describe the bug

When I debug `XPU` accruacy issue, I found the script output wrong accuracy_status.
When the `try` block raise an exception, we should process the exception, but not return the `fail_accuracy`.

Error code:
https://github.com/pytorch/pytorch/blob/ce29d0d796df40f484884e7b8db8b60567dcd95b/benchmarks/dynamo/common.py#L2406-L2408

### Versions

Main branch

cc @chauhang @penguinwu

xuhancn:
When I debug `XPU` accruacy issue, I found the script output wrong accuracy_status.
When the `try` block raise an exception, we should process the exception, but not return the `fail_accuracy`.

Before fixing, it returned as `fail_accuracy`:
image

After fixing, it returned the exception message:
image

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @voznesenskym @penguinwu @EikanWang @Guobing-Chen @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela

### DIFF
--- a/benchmarks/dynamo/common.py
+++ b/benchmarks/dynamo/common.py
@@ -2284,9 +2284,11 @@ class BenchmarkRunner:
                     )
                 ):
                     is_same = False
-            except Exception:
+            except Exception as e:
                 # Sometimes torch.allclose may throw RuntimeError
-                is_same = False
+                exception_string = str(e)
+                accuracy_status = f"fail_exception: {exception_string}"
+                return record_status(accuracy_status, dynamo_start_stats=start_stats)
 
             if not is_same:
                 accuracy_status = "eager_two_runs_differ"
@@ -2403,9 +2405,11 @@ class BenchmarkRunner:
                     force_max_multiplier=force_max_multiplier,
                 ):
                     is_same = False
-            except Exception:
+            except Exception as e:
                 # Sometimes torch.allclose may throw RuntimeError
-                is_same = False
+                exception_string = str(e)
+                accuracy_status = f"fail_exception: {exception_string}"
+                return record_status(accuracy_status, dynamo_start_stats=start_stats)
 
             if not is_same:
                 if self.args.skip_accuracy_check:


benchmarks/dynamo/common.py
```diff
@@ -2284,9 +2284,11 @@ def record_status(accuracy_status, dynamo_start_stats):
                     )
                 ):
                     is_same = False
-            except Exception:
+            except Exception as e:
                 # Sometimes torch.allclose may throw RuntimeError
-                is_same = False
+                exception_string = str(e)
+                accuracy_status = f"fail_exception: {exception_string}"
+                return record_status(accuracy_status, dynamo_start_stats=start_stats)
```
Stonepia, line 2291: I am not sure that this change will affects CUDA. Because they may formerly fail with message "eager two runs differ". But I think this change is reasonable.

Stonepia (approved):

benchmarks/dynamo/common.py
xuhancn: It won't affects CUDA, because of the `accuracy_status` is string type, and not CUDA specific type. We can return the detailed message to it.

chuanqi129 (approved):

Lucaskabela (approved):
Thanks for working to improve the accuracy of our tooling here

zou3519:
This broke inductor periodic tests. Furthermore, it is incompatible with how the periodic tests report issues (we have a csv file that checks the exception type ("fail_accuracy") against the expected exception type "fail_accuracy"). By adding an additional error message in there, this is complicating the situation. 

Things that fail to run should be marked as fail_to_run, not fail_exception. I'm going to revert to get CI back into a good state.

PR #165745, opened 2025-10-17
[dynamo] Clean up assert in dynamo [2/N]

can-gaa-hou:
Extend from #165430
* #165903(Clean up for graph break)
* ->#165745
* #165430

One main refractor from the previous PR:
* For assertions like checking `len(args)` or `len(kwargs)`, using `raise_args_mismatch` instead of `raise_type_error_exc`

I am also considering moving `raise_type_error_exc` into `utils.py` for consistency.

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela @williamwen42

### DIFF
--- a/torch/_dynamo/utils.py
+++ b/torch/_dynamo/utils.py
@@ -2769,22 +2769,24 @@ def slice_length(s: slice, seq_len: int) -> int:
 
 
 def raise_args_mismatch(
-    tx: InstructionTranslatorBase, name: str, expect: str = ""
+    tx: InstructionTranslatorBase,
+    name: str,
+    expect: str = "",
+    actual: str = "",
 ) -> None:
     from torch._dynamo.exc import raise_observed_exception
     from torch._dynamo.variables import ConstantVariable
 
-    msg = [
-        ConstantVariable(
-            f"wrong number of arguments or keyword arguments for {name}() call. "
-            + expect
-        )
-    ]
+    msg_str = (
+        f"wrong number of arguments or keyword arguments for {name}() call.\n"
+        f"  Expect: {expect}\n"
+        f"  Actual: {actual}"
+    )
 
     raise_observed_exception(
         TypeError,
         tx,
-        args=msg,
+        args=[ConstantVariable(msg_str)],
     )
 
 
--- a/torch/_dynamo/utils.py
+++ b/torch/_dynamo/utils.py
@@ -2768,14 +2768,23 @@ def slice_length(s: slice, seq_len: int) -> int:
     return max(0, (stop - start + (step - (1 if step > 0 else -1))) // step)
 
 
-def raise_args_mismatch(tx: InstructionTranslatorBase, name: str) -> None:
+def raise_args_mismatch(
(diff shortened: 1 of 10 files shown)


Lucaskabela (changes requested):
See comment; I think we should split the `Except` string for readability and consistency on future uses of this API.  This is a good step for sure though!

torch/_dynamo/variables/dicts.py
```diff
@@ -630,14 +663,22 @@ def call_method(
...
+                    f"Expect: 1 or 2 args and 0 kwargs, Actual: {len(args)} args and {len(kwargs)} kwargs",
+                )
 
             if not arg_hashable:
                 raise_unhashable(args[0])
 
             self.install_dict_keys_match_guard()
-            assert not kwargs
-            assert len(args) <= 2
+            if kwargs and len(args) > 2:
```
Lucaskabela: Should this be an or?  As if we rewrite 639 to combine asserts it would be 

`assert not kwargs and len(args) <= 2`

so

`not (not kwargs and len(args) <= 2)` == `kwargs or len(args) > 2`

torch/_dynamo/utils.py
```diff
@@ -2768,14 +2768,23 @@ def slice_length(s: slice, seq_len: int) -> int:
     return max(0, (stop - start + (step - (1 if step > 0 else -1))) // step)
 
 
-def raise_args_mismatch(tx: InstructionTranslatorBase, name: str) -> None:
+def raise_args_mismatch(
+    tx: InstructionTranslatorBase, name: str, expect: str = ""
```
Lucaskabela: I think we can go a step further here by splitting the `except` string into two strings; except, and actual.

This gives us two benefits:
  1) Encapulating the the `Expect: ... Actual:` format inside this API
  2) Separate the string into two lines for readability (since these get kind of long)

williamwen42 (commented):
One concern that I have is that the error messages don't exactly match what Python produces.

I don't expect most users to do this, but if a user tries to compile code that catches an exception and checks for its error message, dynamo would silently trace this incorrectly.

torch/_dynamo/variables/dicts.py
can-gaa-hou: Thanks for the correction! Fixed.

torch/_dynamo/utils.py
can-gaa-hou: Yes, I have changed to the format you mentioned. Thanks for the advice.

can-gaa-hou:
> One concern that I have is that the error messages don't exactly match what Python produces.

@williamwen42 We can focus on the assertions that check for the length and type of arguments first. They are easy to fix and we can generate similar error messages as what Python produces. Also, we get benefit from cleaning up assertions without error messages (See #164878).

Lucaskabela (approved):
Thanks for the changes LGTM :)

PR #165750, opened 2025-10-17
[1/N] Change C-style casts to static_cast or reinterpret_cast

cyyever:
This series of changes try to cover C style casts into C++ alternatives.


Skylion007 (commented):
GitHub kept erroring my leaving inline comments. Some of the static casts from std::string.data to char cast, are they really necessary? 

Some of the reinterpret casts should be const casts, which are bit safer.

Also slight nit: I noticed unique_devices set checks for membership before insertion. I think you can just call.Insert on it and it will be a noop if the key is present and return an std::pair where the second value tells you if it succeeded or not.

Skylion007 (approved):

clee2000:
I think this revert is wrong, the failures are probably due to the AWS outage instead

PR #165752, opened 2025-10-17
filter out alloc-free pairs from trace plot

waysg:
Summary:
When dealing with a large memory trace, the resulting plot can be challenging to interpret and analyze.
This commit introduces a feature that enables filtering of allocations that have already been freed, providing a more focused view.
The remaining events in the plot often warrant closer examination, as they may be indicative of potential out-of-memory (OOM) issues.

### DIFF
--- a/torch/cuda/_memory_viz.py
+++ b/torch/cuda/_memory_viz.py
@@ -446,7 +446,43 @@ def _format_viz(data, viz_kind, device):
     )
 
 
-def trace_plot(data, device=None, plot_segments=False):
+def filter_alloc_free_pairs(data):
+    for dev_id in range(len(data["device_traces"])):
+        # set of indexes of trace events for alloc-free pairs
+        filterSet = set()
+        # map from addr to index of alloc event
+        allocMap = {}
+        # set of addrs from free_requested events
+        freeRequested = set()
+        for idx, event in enumerate(data["device_traces"][dev_id]):
+            if event["action"] == "alloc":
+                allocMap[event["addr"]] = idx
+            elif event["action"] == "free_requested":
+                freeRequested.add(event["addr"])
+                if allocMap.get(event["addr"]) is not None:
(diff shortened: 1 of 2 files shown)


torch/cuda/_memory_viz.py
```diff
@@ -446,18 +446,56 @@ def _format_viz(data, viz_kind, device):
...
+                if allocMap.get(event["addr"]) is not None:
+                    filterSet.add(idx)
+                    filterSet.add(allocMap[event["addr"]])
+                    allocMap.pop(event["addr"])
+            elif event["action"] == "free_completed":
+                if event["addr"] in freeRequested:
+                    freeRequested.remove(event["addr"])
+                    filterSet.add(idx)
+                else:
+                    print(f"free_completed without free_requested: {event}")
```
Skylion007, line 471: Should be a logger debug?

waysg: @Skylion007 It looks like this file uses print instead of logger. To be consistent, I used 'print' here.

torch/cuda/_memory_viz.py
```diff
@@ -446,18 +446,56 @@ def _format_viz(data, viz_kind, device):
     )
 
 
-def trace_plot(data, device=None, plot_segments=False):
+def filter_alloc_free_pairs(data):
+    for dev_id in range(len(data["device_traces"])):
+        filterSet = set()
```
zdevito, line 452: can you add type annotations? I think filterSet holds indices into the device traces, and allocMap is address -> index, but I am not sure.

zdevito (approved):
Looks good! Thanks for adding it

torch/cuda/_memory_viz.py
waysg: Correct. 
This file does not use type annotations, so I added comments to explain the variables.

PR #165762, opened 2025-10-17
Fix race condition and make CUDA kthvalue deterministic

nick-kuhn:
The gatherKthValue kernel had a race condition where multiple threads could write to the same output location without synchronization when duplicate k-th values exist, resulting in non-deterministic output.

Changes:
- aten/src/ATen/native/cuda/Sorting.cu: Use atomicMin with shared memory to deterministically find minimum index. Add early termination and remove redundant inRange checks. (We have to cast the index to `int32_t`, but this is already assumed to fit earlier in the kernel.)
- aten/src/ATen/native/cuda/Sorting.cpp: Remove non-deterministic alert since kthvalue is now deterministic on CUDA.
- torch/__init__.py: Remove kthvalue from non-deterministic operations list and remove kthvalue example from use_deterministic_algorithms() docstring.
- test/test_torch.py: Remove test_nondeterministic_alert_kthvalue since kthvalue no longer raises alerts on CUDA.

Benefits:
- Deterministic: always returns minimum index when duplicates exist
- Potential performance improvement on large arrays with repetitions

Test Results:
- All existing PyTorch tests pass (test_kthvalue)
- Custom determinism tests confirm consistent results
- Custom CUDA vs CPU correctness validated across 50+ scenarios
- Custom performance benchmarks show improvements with no visible regressions

Addresses #165227


aten/src/ATen/native/cuda/Sorting.cu
```diff
@@ -65,25 +65,34 @@ __global__ void gatherKthValue(
       &kValue);
 
   // Find the index of the k-th highest element
-  index_t kValueIndex = 0;
-  bool foundKValue = false;
+  __shared__ int32_t minIndexFound;
```
ngimel, line 68: this (and atomic below) won't correctly work when the kernel is launched with `index_t = int64_t`, for large slices that require 64-bit indexing

ngimel (changes requested):

aten/src/ATen/native/cuda/Sorting.cu
nick-kuhn: This is already an implicit condition for the radix sort earlier in the kernel and the function currently fails silently if the length exceeds int32. (e.g. see the simple example [here](https://colab.research.google.com/drive/1otrSfchmt4De0OJt6VRvjNmolyf8YsSw?usp=sharing)). Hence, I didn't attempt to address this case, but maybe one should add an error/warning to the current CUDA implementation.

nick-kuhn: Happy for suggestions on how to proceed.

ngimel: Thanks for checking! Can you please error out in this case then? Does instantiating with index_t = int64_t ever make sense, or will it only lead to silent wrong results?

nick-kuhn: I added the error message. The index_t is determined by the total size of the tensor, and can be int64_t, as long as the size of each slice (i.e. the dimension kthvalue is taken along) fits into int32_t (we have to cast back and forth).

ngimel (approved):

eqy (approved):

PR #165766, opened 2025-10-17
Enable kineto XPU scope profiler extending existing XPU profiler plugin

moksiuc:
**Summary:**

As XPU became a PyTorch built-in device, the profiler support is indispensable part of functionality completeness. In this PR, the XPU scope profiler is introduced by extending existing XPU profiler plugin. It allows to gather XPU hardware metrics. The LIBKINETO_NOXPUPTI option is used to enable or disable the whole XPU profiler plugin during kineto build stage.

**Changes:**

Enanced prepareTraceWithExperimentalOptions path in kineto_shim.cpp file
Added test


torch/csrc/profiler/kineto_shim.cpp
```diff
@@ -179,34 +179,47 @@ class ExperimentalConfigWrapper {
   }
 
   void prepareTraceWithExperimentalOptions(
-      std::set<libkineto::ActivityType>&& enabled_activities) {
+      std::set<libkineto::ActivityType>&& enabled_activities,
+      bool hasActivityTypeXPU) {
```
EikanWang: It is safe to assume that CUDA and XPU never coexist. So, this flag is not necessary because we can use a build Macro to do it.

torch/csrc/profiler/kineto_shim.cpp
```diff
@@ -179,34 +179,47 @@ class ExperimentalConfigWrapper {
...
     std::set<libkineto::ActivityType> k_activities =
         std::move(enabled_activities);
 #ifdef USE_KINETO
     k_activities.insert(libkineto::ActivityType::CUDA_PROFILER_RANGE);
+    if (hasActivityTypeXPU) {
+      k_activities.insert(libkineto::ActivityType::XPU_SCOPE_PROFILER);
+    }
```
EikanWang: ditto

torch/csrc/profiler/kineto_shim.cpp
```diff
@@ -179,34 +179,47 @@ class ExperimentalConfigWrapper {
...
 
     // Add CPU activities if we are measuring per kernel ranges
     if (config_.profiler_measure_per_kernel) {
       k_activities.insert(kCpuTypes.begin(), kCpuTypes.end());
     }
 
     const size_t num_metrics = config_.profiler_metrics.size();
-    std::stringstream configss;
 
     LOG(INFO) << "CUPTI profiler metrics size = " << num_metrics;
```
EikanWang: `CUPTI` should be generalized as well, right?

torch/csrc/profiler/kineto_shim.cpp
```diff
@@ -179,34 +179,47 @@ class ExperimentalConfigWrapper {
...
+      if (i > 0) {
+        metrics += ',';
       }
+      metrics += config_.profiler_metrics[i];
+    }
+
+    std::stringstream configss;
+    configss << "ACTIVITIES_WARMUP_PERIOD_SECS=0\n"
+             << "CUPTI_PROFILER_METRICS=" << metrics << '\n';
+    if (hasActivityTypeXPU && !metrics.empty()) {
```
EikanWang: Please use build marco to isolate different devices. Meanwhile, why does XPU need to check `metrics.empty`, while CUDA does not?

test/profiler/test_xpu_scope_profiler.py
EikanWang, line 1: I suppose we should update the profiler-related test cases directly. @chuanqi129 , do you know why we have not enabled profiler cases yet?

torch/csrc/profiler/kineto_shim.cpp
moksiuc: It is optimization. If metrics is empty it would generate empty "XPUPTI_PROFILER_METRICS=".

torch/csrc/profiler/kineto_shim.cpp
moksiuc: Actually I don't change code for cupti.
I'm only reusing the same code for both cupti and xpupti definitions.

torch/csrc/profiler/kineto_shim.cpp
moksiuc: Done

torch/csrc/profiler/kineto_shim.cpp
moksiuc: Done

EikanWang (approved):

EikanWang:
@sraikund16 , any comments?

test/profiler/test_xpu_scope_profiler.py
chuanqi129: We have enabled some profiler-related test cases under https://github.com/pytorch/pytorch/tree/main/test/profiler. @moksiuc May I know is there any existed test cases can reuse?

moksiuc: I don't think there existed test for scope profiler so far as it is completely new feature.

sraikund16 (approved):
LGTM

aostrowski-hbn:
Failures are unrelated (mostly S3 infra issues after the tests passed), also two unrelated failures in shards 7 & 10. 

@scotts @ryanzhang22 can we merge? This PR is just integrating https://github.com/pytorch/kineto/pull/1174 from PT side.

torch/csrc/profiler/kineto_shim.cpp
```diff
@@ -219,7 +219,13 @@ class ExperimentalConfigWrapper {
     std::set<libkineto::ActivityType> k_activities =
         std::move(enabled_activities);
 #ifdef USE_KINETO
+#if not defined(USE_CUDA) and defined(USE_XPU)
```
gujinghui: @aostrowski-hbn 
Why can USE_CUDA not be defined here? Is the XPU scoped profiler excluded when CUDA or cuPTI is enabled?

aostrowski-hbn: @gujinghui There's no support for simultaneous CUDA + XPU tracing, similarly as there's no PrivateUse1 + CUDA/XPU.

We use the same activity types and this is problematic, for instance: https://github.com/pytorch/pytorch/blob/main/torch/csrc/profiler/kineto_shim.cpp#L492-L503

This change was introduced due to https://github.com/pytorch/pytorch/pull/165766#discussion_r2586967882.

gujinghui (approved):

aostrowski-hbn:
@scotts please merge

aostrowski-hbn:
@scotts @ryanzhang22 failures seem unrelated. Please merge or comment if anything else is required from our side.

gujinghui:
@divyanshk @scotts  could you help review and merge this PR? Thanks.

scotts:
This now has serious merge conflicts with https://github.com/pytorch/pytorch/pull/187204.

We removed `_ExperimentalConfig(profiler_metrics, profiler_measure_per_kernel)` as it was dead code; the code in Kineto that was doing that work had been disabled for about a year. We are in the process of removing the code on the Kineto side as well. The feature also did not get much traction in five years. Things that are "experimental" need to either graduate to supported, documented features or be removed.

Do you have an alternate way of enabling support for the XPU scope profiler?

ryanzhang22:
If this is a profiler that needs first-class support in XPU, one potential option is to leverage the `activities` field. https://github.com/pytorch/pytorch/pull/176351 added the ability to specify filtered activity selection in profiler, so you could do something like:

```
with torch.profiler.profile(
    activities=[
        torch.profiler.ProfilerActivity.CPU,
        {torch.profiler.ProfilerActivity.XPU: ["GpuTime", "GpuCoreClocks"]},
    ]
):
```

That way we would avoid introducing more IFDEFs based on device type into the code + avoid using experimental options. You could set a default set of activities as well for bare `torch.profiler.ProfilerActivity.XPU`.

aostrowski-hbn:
@scotts @ryanzhang22 Thanks for the comments. The code is now using the new API. Please import and review.

torch/profiler/profiler.py
```diff
@@ -164,6 +164,23 @@ class _KinetoProfile:
...
+            values are attached to each kernel as Perfetto counters. Hardware
+            metric collection requires ``ZET_ENABLE_METRICS=1`` and access to
+            the GPU performance counters (``perf_stream_paranoid``/
+            ``observation_paranoid`` set to ``0``). Metric names are
+            device-specific and defined by the driver, not by PyTorch.
+            Because a ``dict`` entry collects exactly what it lists,
+            requesting only metrics (e.g. ``{ProfilerActivity.XPU: ["GpuTime"]}``)
+            does not also collect the default XPU activities; list the desired
+            activity type names alongside the metrics to keep tracing them, e.g.
+            ``{ProfilerActivity.XPU: ["CONCURRENT_KERNEL", "XPU_RUNTIME", "GpuTime"]}``.
```
scotts: @moksiuc ,this is a lot of device specific text in one of the most prominent fields in the main API entry point to the scheduler. Rather than putting device specific text in general areas, let's put this on the `ProfilerActivity.XPU` item itself.

That's actually defined in C++ at `torch/csrc/profiler/python/init.cpp`. Let's do something like:
```
  py::enum_<ActivityType>(
      m,
      "ProfilerActivity",
      "Device kinds that the profiler supports.")
      .value("CPU", ActivityType::CPU, "CPU events (operators, runtime, ...).")
      .value("XPU", ActivityType::XPU, "Intel XPU device activity. <NEW STUFF FROM THIS PR HERE>")
      .value("MTIA", ActivityType::MTIA, "MTIA device activity.")
      .value("CUDA", ActivityType::CUDA, "CUDA kernels and runtime.")
      .value("HPU", ActivityType::HPU, "HPU device activity.")
      .value("PrivateUse1", ActivityType::PrivateUse1, "PrivateUse1 backend activity.");
```

torch/csrc/profiler/kineto_shim.cpp
```diff
@@ -258,6 +310,9 @@ void prepareTrace(
   }
 
   std::set<libkineto::ActivityType> k_activities;
+  // Config fragment for the XPUPTI scope profiler, populated when the XPU
+  // group requests hardware metrics. Empty means no scope profiling.
+  std::string xpuScopeConfig;
```
scotts: Let's name this `extraConfig` and remove the comment. That this is only use by the XPU scope profiler is not important at this scope.

torch/csrc/profiler/kineto_shim.cpp
```diff
@@ -100,6 +101,57 @@ std::unordered_set<libkineto::ActivityType> filterActivities(
...
+}
+
+// Build the Kineto config fragment that drives the XPUPTI scope profiler.
+// The keys mirror those parsed by XpuptiScopeProfilerConfig::handleOption.
+// Per-kernel (PTI_METRICS_SCOPE_AUTO_KERNEL) is the only mode PTI supports
+// today, so it is intentionally hardcoded here; custom_profiler_config is the
+// reserved channel for a future PTI_METRICS_SCOPE_USER mode.
+std::string buildXpuScopeConfig(const std::vector<std::string>& metrics) {
+  std::stringstream configss;
+  configss << "ACTIVITIES_WARMUP_PERIOD_SECS=0\n"
```
ryanzhang22: for my info, why is ACTIVITIES_WARMUP_PERIOD_SECS hard coded to 0?

ryanzhang22 (approved):

torch/profiler/profiler.py
```diff
@@ -164,6 +164,10 @@ class _KinetoProfile:
             An empty list (e.g. ``{ProfilerActivity.CUDA: []}``) means collect
             nothing for that group.
             The same activity group must not appear more than once.
+            The valid activity type names for each group are the per-device
+            default sets defined in ``torch/csrc/profiler/kineto_shim.cpp``.
+            See :class:`~torch.profiler.ProfilerActivity` for device-specific
+            behavior (e.g. Intel XPU hardware metrics).
```
scotts: Remove the code reference to the `kineto_shim.cpp` source file. The sphinx relative reference to `ProfilerActivity` is what matters. Also remove `(eg. Intel XPU hardware metrics)`. Again, this is very close to the top of the documentation page. We want to be sparing with our words; the link should suffice.

torch/profiler/profiler.py
```diff
@@ -848,6 +852,10 @@ class profile(_KinetoProfile):
...
```
scotts: Same as above.

torch/profiler/profiler.py
aostrowski-hbn: Great point, done.

torch/csrc/profiler/kineto_shim.cpp
aostrowski-hbn: Done.

torch/csrc/profiler/kineto_shim.cpp
aostrowski-hbn: If was a Range Profiler artifact -- now removed. Thanks for noticing this!

torch/profiler/profiler.py
aostrowski-hbn: Done

torch/profiler/profiler.py
aostrowski-hbn: Done

gujinghui:
@ryanzhang22 @scotts
Could you please check whether all of your comments have been addressed? Thanks.

ryanzhang22 (approved):

PR #165769, opened 2025-10-17
[ONNX] Support renaming in dynamic axes to shapes conversion

titaiwangms:
Discovered in ##165748

This PR also deprecates the conversion. ONNX exporter team does not intend to maintain the conversion in long term.


torch/onnx/_internal/exporter/_dynamic_shapes.py
```diff
@@ -39,6 +39,14 @@ def from_dynamic_axes_to_dynamic_shapes(
 
     Detail on Dim.DYNAMIC: `#133620 <https://github.com/pytorch/pytorch/pull/133620>`_
     """
+
+    warnings.warn(
+        "from_dynamic_axes_to_dynamic_shapes is deprecated and will be removed in a future release. "
```
justinchuby, line 44: Describe the behavior instead? Since the function is not public.

justinchuby (approved):

torch/onnx/_internal/exporter/_dynamic_shapes.py
titaiwangms: Should be more informative now

torch/onnx/_internal/exporter/_dynamic_shapes.py
```diff
@@ -42,7 +42,8 @@ def from_dynamic_axes_to_dynamic_shapes(
 
     warnings.warn(
         "from_dynamic_axes_to_dynamic_shapes is deprecated and will be removed in a future release. "
-        "Please provide 'dynamic_shapes' directly.",
+        "This function converts 'dynamic_axes' format (including custom axis names) to 'dynamic_shapes' format. "
+        "Instead of relying on this conversion, provide 'dynamic_shapes' directly using torch.export.Dim objects with custom names.",
```
justinchuby: We do support strings right? Maybe just say supply dynamic_shapes?

titaiwangms: Removed "torch.export.Dim objects"

PR #165770, opened 2025-10-17
Remove unused exception parameter from some files, to work with -Wunused-exception-parameter

psiddh:
Summary: address compiler complains that were coming up to unblock the build

Test Plan:
before the change
```
aten/src/ATen/native/LinearAlgebra.cpp:3623:36: error: unused exception parameter 'e' [-Werror,-Wunused-exception-parameter]
 3623 |     } catch (const std::exception& e) {
      |
```

after: targets build with `-Wunused-exception-parameter`

Differential Revision: D84876246


Skylion007 (approved):
Huh, because TORCH_WARN might be removed based on various compile switches, interesting...

cyyever (approved):
Hope that we can enable Wunused-exception-parameter finally.

PR #165774, opened 2025-10-17
Add NVFP4 two-level scaling to scaled_mm

slayton58:
* __->__ #165774

Summary:

* Add second-level scaling dispatch to scaled_mm, tying into optional `alpha` passing
* Add two-level tests

Test Plan:

```
pytest -svv -k "nvfp4_global_scale" test/test_scaled_matmul_cuda.py
```

Reviewers:

Subscribers:

Tasks:

Tags:


aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -2323,11 +2323,20 @@ _scaled_nvfp4_nvfp4(
           const std::optional<Tensor>& bias,
           const c10::ScalarType out_dtype,
           const bool single_scale,
```
Skylion007: metapoint, super weird API design. Why have an explicit single_scale flag instead of implicitly checking where Scale_A, Scale_B is null?

test/test_scaled_matmul_cuda.py
```diff
@@ -1254,6 +1290,59 @@ def test_pack_uint4(self):
...
+        # Very unbalanced
+        (1023, 64, 48),
+        (31, 1024, 64),
+        (45, 96, 1024),
+
+        # Mixed large and small
+        (2, 1024, 128),
+        (127, 96, 1024),
+        (1025, 128, 96)
+    ], name_fn=lambda mkn: f"{mkn[0]}_{mkn[1]}_{mkn[2]}")
```
Skylion007, line 1313:
```suggestion
    ], name_fn=lambda mkn: "_".join(mkn))

File: aten/src/ATen/native/cuda/Blas.cpp (hunk 1 above)
Comment (slayton58) [reply aten/src/ATen/native/cuda/Blas.cpp]: Yeah, this is a reasonable point - originally `single_scale` was added to differentiate the different options, before the 2-level scale parts were fleshed out -- I'll revisit this

File: test/test_scaled_matmul_cuda.py (hunk 2 above)
Comment (slayton58) [reply test/test_scaled_matmul_cuda.py:1313]: Nit: this is directly from the `test_blockwise_mxfp8_nvfp4_mxfp4_numerics` test parametrize annotations - I can change both.

File: aten/src/ATen/native/cuda/Blas.cpp (hunk 1 above)
Comment (Skylion007) [reply aten/src/ATen/native/cuda/Blas.cpp]: https://read.engineerscodex.com/p/the-boolean-trap

File: test/test_scaled_matmul_cuda.py (hunk 2 above)
Comment (slayton58) [reply test/test_scaled_matmul_cuda.py:1313]: Actually, because the tuple is ints, not strings, it looks more like:
```
"_".join(str(i) for i in mnk)
```
which I don't think is any better than what's there in terms of clearness or brevity.

File: aten/src/ATen/native/cuda/Blas.cpp [hunk 3]
```diff
@@ -2322,12 +2322,20 @@ _scaled_nvfp4_nvfp4(
...
-          Tensor& out) {
+          Tensor& out,
+          const std::optional<Tensor>& global_scale_a = std::nullopt,
+          const std::optional<Tensor>& global_scale_b = std::nullopt) {
 #ifdef USE_ROCM
   TORCH_CHECK_NOT_IMPLEMENTED(false, "NVFP4 scaling not supported on ROCM");
 #endif
-  TORCH_CHECK_VALUE(single_scale, "Only single-scaled NVFP4 currently supported");
+  std::optional<Tensor> alpha = std::nullopt;
+  if (global_scale_a.has_value() && global_scale_b.has_value()) {
```
drisspg: should this be scale_a or scale_b?

drisspg (approved):

aten/src/ATen/native/cuda/Blas.cpp
slayton58: What do you mean? Each tensor should have a global scale attached to it, unless we want to also handle the case where only 1 does.

drisspg: Ohh well I mean, I think we can remove the torch_checks if keep it && or we keep the torch checks (to provide nicer error message) and turn the scales to or?

slayton58: I see - if one global scale is passed, trigger the check for both scales present, as opposed to now, where if you pass a single global scale, it'll silently do nothing with it.

PR #165777, opened 2025-10-17
Addmv backward fails

### LINKED ISSUE #76778: `torch.addmv` backward fails
### 🐛 Describe the bug

`torch.addmv` backward fails when `input` is `float64`, `mat` and `vec` are `complex128`

```python
import torch
input = torch.rand([2], dtype=torch.float64, requires_grad=True)
mat = torch.rand([2, 3], dtype=torch.complex128, requires_grad=True)
vec = torch.rand([3], dtype=torch.complex128, requires_grad=True)

res = torch.addmv(input, mat, vec)
print("addmv SUCCEED!")

res_2 = res.sum()
print("sum SUCCEED!")

res_2.backward()
# addmv SUCCEED!
# sum SUCCEED!
# RuntimeError: Expected isFloatingType(grad.scalar_type()) || (input_is_complex == grad_is_complex) to be true, but got false.  (Could this error message be improved?  If so, please report an enhancement request to PyTorch.)
```

However, when `input` is `complex128`, `mat` and `vec` are `float64`, it will succeed to backward

### Versions

pytorch: 1.11.0

cleonard530:
Fixes #76778

Added a check to make sure that the dtypes for addmv are either all floats or all complex. 

@soulitzer, @anjali411, @lezcano 


test/inductor/test_torchinductor.py
```diff
@@ -3758,8 +3758,8 @@ def fn(a, b, c):
             return torch.addmv(a, b, c)
 
         cfn = torch.compile(backend="inductor")(fn)
-        input = torch.tensor([2], dtype=torch.int32)
-        mat = torch.tensor(np.random.randn(0, 0), dtype=torch.int32)
+        input = torch.tensor([2], dtype=torch.float32)
```
thenumberouscode: oh, Changing an exist unit test may be a bad operation. you can add float32 as a new dtype.

cleonard530: The update I made and the comments from the original issue was to only allow floating point types or complex types. So, all implementations of addmv with integral types need to be removed. If I am misunderstanding and we want to include integral types as well, then I can update the fix, but I don't think they should be included.

isuruf (approved):

PR #165779, opened 2025-10-17
162481: synchronize CUDA operations at each profiler step

ahkush:
Fixes #162481. Fix Kineto profiler kernel duplication issue by synchronizing CUDA operations at each profiler step.
This ensures kernels from previous iterations don't bleed into subsequent profiling steps.


ngimel (changes requested):
This adds unacceptable synchronization overhead to multi-step profiled workloads

PR #156635, opened 2025-06-23
[ONNX] Simplify attention op tests

justinchuby:
Follow up of https://github.com/pytorch/pytorch/pull/156431

fix https://github.com/pytorch/pytorch/issues/156524

cc @titaiwangms

PR #165780, opened 2025-10-17
[ONNX] Create fake implementations for onnx ops; fix boolean mask in attention

### LINKED ISSUE #164909: [ONNX] Create meta implementations for onnx ops
Currently in torch.onnx.ops.*, we generate meta functions for the ops. However since the implementation can be complicated (Attention etc.), the genreated meta functions may see constraint issues when being traced. Instead, we should just create meta functions for them.

cc @titaiwangms @xadupre

justinchuby:
Previously we rely on the concreate implementation to generate fake implementation. This makes the fake implementation overly complicated and breaks in some cases when there are dynamic shapes.

This PR updates onnx op registration to instead take a dedicated fake implementation.

**Also fixed: When boolean mask is supplied to torch sdpa, it was previously taken the negation, which is incorrect.**

Fix https://github.com/pytorch/pytorch/issues/164909 Also taken changes from https://github.com/pytorch/pytorch/pull/156635

cc @titaiwangms

### DIFF
--- a/test/onnx/ops/test_ops.py
+++ b/test/onnx/ops/test_ops.py
@@ -427,7 +427,6 @@ class NativeOnnxOpsTest(common_utils.TestCase):
             dynamo=True,
             fallback=False,
             verbose=False,
-            optimize=False,
             **options,
         )
         assert onnx_program is not None
--- a/test/onnx/ops/test_ops.py
+++ b/test/onnx/ops/test_ops.py
@@ -4,9 +4,7 @@
 from __future__ import annotations
 
 import onnx_ir.passes.common as common_passes
-import onnxruntime
 from onnxscript import ir
-from packaging import version
 
 import torch
 from torch.onnx._internal.exporter import _testing as onnx_testing
@@ -418,6 +416,7 @@ class SymbolicOpsTest(common_utils.TestCase):
             )
 
 
+@common_utils.instantiate_parametrized_tests
 class NativeOnnxOpsTest(common_utils.TestCase):
     def export(self, model, args=(), kwargs=None, **options) -> torch.onnx.ONNXProgram:
         onnx_program = torch.onnx.export(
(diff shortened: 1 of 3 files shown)


torch/onnx/ops/_impl.py
```diff
@@ -327,18 +423,11 @@ def attention_23(
 
     if can_use_sdpa:
         # Use PyTorch's optimized scaled_dot_product_attention
-
-        # Prepare attention mask for SDPA
-        sdpa_attn_mask = None
-        if attn_mask is not None:
-            # Convert boolean mask: True means participate, SDPA expects True to mask out
```
justinchuby, line 334: This is incorrect

test/onnx/ops/test_ops.py
```diff
@@ -1411,8 +1398,19 @@ def forward(self, Q, K, V):
...
         )
+        onnx_testing.assert_onnx_program(onnx_program)
 
-    def test_attention_export_with_softmax_precision(self):
+    @common_utils.parametrize(
+        "precision_enum, precision_name",
+        [
+            (1, "FLOAT"),
+            (10, "FLOAT16"),
+            (11, "DOUBLE"),
```
xadupre, line 1408: BFLOAT16?

justinchuby: I removed because I don't know how well ort cpu can run it, but I can add back and try

justinchuby: Added

torch/onnx/ops/_impl.py
```diff
@@ -37,15 +37,27 @@ def decorator(func: Callable[_P, _R]) -> Callable[_P, _R]:
...
+    cos_cache: torch.Tensor,
+    sin_cache: torch.Tensor,
+    position_ids: Optional[torch.Tensor] = None,
+    *,
+    interleaved: bool = False,
+    num_heads: int = 0,
+    rotary_embedding_dim: int = 0,
+) -> torch.Tensor:
+    """Fake implementation for RotaryEmbedding-23 for torch.compile purposes."""
+    return x.clone()
```
titaiwangms, line 64: Is this correct? Can we put some instructions of how to write a fake impl? I am not sure I understand this.

justinchuby: This is correct in that we only need a tensor that is not the same input tensor and that it has desired device/dtype/shape of the output. Clone should do it just fine.

titaiwangms (approved):

torch/onnx/ops/_impl.py
titaiwangms: Could you add a ref link or some instructions for future work on adding more fake impl. Thanks!

justinchuby: Added to the file: https://docs.pytorch.org/docs/stable/library.html#torch.library.register_fake

PR #165784, opened 2025-10-17
Bugfix to forward autodiff causing different datatype 2

### LINKED ISSUE #160513: Forward autodiff : Multiplying by python float changes the dual dtype in some situations
Hello !
This bug report is based on a discussion from the pytorch forum : [https://discuss.pytorch.org/t/forward-autodiff-multiplying-by-python-float-changes-the-dual-dtype-in-some-situations/222163]

## Bug Description
When using `torch.func.jacfwd` (forward-mode automatic differentiation) on a 0‑dimensional dual tensor multiplied by a Python float literal (e.g., `0.5 * x[0]`), the primal retains its declared dtype (e.g. `float32`), but the tangent component is unexpectedly promoted to `float64`. This mismatch causes runtime errors in subsequent tensor operations requiring consistent dtypes, like `torch.tensordot` even though there is no dtype mismatch without autodiff.

## Minimal Repro Example

```python
import torch
print(torch.version.__version__)

fargs = {'dtype': torch.float32, 'device': 'cuda'}
A = torch.rand((5, 3, 100), **fargs)

def test(x):
    a = 0.5 * x[0]
    p, t = torch.autograd.forward_ad.unpack_dual(a)
    print(f"a dtype : {p.dtype}, a.tangent dtype : {t.dtype}")
    v = torch.stack([a, *x[1:]])
    p, t = torch.autograd.forward_ad.unpack_dual(v)
    print(f"v dtype : {p.dtype}, v.tangent dtype : {t.dtype}")
    return torch.tensordot(v, A, dims=1)

torch.func.jacfwd(test)(torch.rand(5, **fargs))
```

### Output
```
2.7.1+cu128
a dtype : torch.float32, a.tangent dtype : torch.float64
v dtype : torch.float32, v.tangent dtype : torch.float64

[...]

RuntimeError: expected mat1 and mat2 to have the same dtype, but got: double != float
```

This results in a dtype mismatch (`float64` vs `float32`) in the tangent part, triggering a runtime error during `tensordot`.

## K. Frank’s Diagnostic Example (Clarifying Scope of Bug)

```python
import torch
print(torch.__version__)

primal1d = torch.ones(1)
tangent1d = torch.ones(1)
primal0d = torch.ones(1)[0]
tangent0d = torch.ones(1)[0]

with torch.autograd.forward_ad.dual_level():
    # 1‑D tensor, no scalar multiplication — ok
    dual1d = torch.autograd.forward_ad.make_dual(primal1d, tangent1d)
    p1d, t1d = torch.autograd.forward_ad.unpack_dual(dual1d)
    print(p1d.dtype, t1d.dtype, t1d.dtype == p1d.dtype)

    # 1‑D tensor × Python scalar — still ok
    dual1db = 1.0 * dual1d
    p1db, t1db = torch.autograd.forward_ad.unpack_dual(dual1db)
    print(p1db.dtype, t1db.dtype, t1db.dtype == p1db.dtype)

    # 0‑D tensor, no scalar multiplication — ok
    dual0d = torch.autograd.forward_ad.make_dual(primal0d, tangent0d)
    p0d, t0d = torch.autograd.forward_ad.unpack_dual(dual0d)
    print(p0d.dtype, t0d.dtype, t0d.dtype == p0d.dtype)

    # 0‑D tensor × Python scalar — **BUG**: tangent gets promoted
    dual0db = 1.0 * dual0d
    p0db, t0db = torch.autograd.forward_ad.unpack_dual(dual0db)
    print(p0db.dtype, t0db.dtype, t0db.dtype == p0db.dtype)
```

**Output summary:**

- **1‑D tensor (no scalar)**: primal/tangent both `float32` → OK  
- **1‑D tensor × Python float**: both still `float32` → OK  
- **0‑D tensor (no scalar)**: both `float32` → OK  
- **0‑D tensor × Python float**: primal = `float32`, tangent = **`float64`** → **BUG**

## Workarounds

- Use a tensor when scaling:  
  ```python
  a = torch.tensor(0.5, **fargs) * x[0]
  ```
- Or apply the scalar before slicing:  
  ```python
  a = (0.5 * x)[0]
  ```
Both preserve matching primal and tangent dtypes (`float32`) and avoid the bug.

**Maybe torch can raise a warning if primal and dual values have different dtype during forward autodiff to help debugging ?**

### Versions

PyTorch version: 2.7.1+cu128
Is debug build: False
CUDA used to build PyTorch: 12.8

OS: Ubuntu 24.04.2 LTS (x86_64)
GCC version: (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0
CMake version: Could not collect
Libc version: glibc-2.39

Python version: 3.13.5 (main, Jul 23 2025, 00:37:22) [Clang 20.1.4 ] (64-bit runtime)
Python platform: Linux-6.14.0-27-generic-x86_64-with-glibc2.39
Is CUDA available: True
CUDA runtime version: 12.0.140
CUDA_MODULE_LOADING set to: LAZY
GPU models and configuration: GPU 0: NVIDIA RTX 3500 Ada Generation Laptop GPU
Nvidia driver version: 575.64.03
cuDNN version: Could not collect
Is XPU available: False
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Address sizes:                        46 bits physical, 48 bits virtual
CPU(s):                               28
On-line CPU(s) list:                  0-27
Model name:                           13th Gen Intel(R) Core(TM) i7-13850HX
Model:                                183
Core(s) per socket:                   20
Stepping:                             1
CPU(s) scaling MHz:                   31%
CPU max MHz:                          5300.0000
CPU min MHz:                          800.0000
BogoMIPS:                             4608.00
Flags:                                fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc art arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf tsc_known_freq pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm 3dnowprefetch cpuid_fault epb ssbd ibrs ibpb stibp ibrs_enhanced tpr_shadow flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 smep bmi2 erms invpcid rdseed adx smap clflushopt clwb intel_pt sha_ni xsaveopt xsavec xgetbv1 xsaves split_lock_detect user_shstk avx_vnni dtherm ida arat pln pts hwp hwp_notify hwp_act_window hwp_epp hwp_pkg_req hfi vnmi umip pku ospke waitpkg gfni vaes vpclmulqdq rdpid movdiri movdir64b fsrm md_clear serialize pconfig arch_lbr ibt flush_l1d arch_capabilities
Virtualization:                       VT-x
L1d cache:                            768 KiB (20 instances)
L1i cache:                            1 MiB (20 instances)
L2 cache:                             28 MiB (11 instances)
L3 cache:                             30 MiB (1 instance)
NUMA node0 CPU(s):                    0-27
Vulnerability Gather data sampling:   Not affected
Vulnerability Ghostwrite:             Not affected
Vulnerability Itlb multihit:          Not affected
Vulnerability L1tf:                   Not affected
Vulnerability Mds:                    Not affected
Vulnerability Meltdown:               Not affected
Vulnerability Mmio stale data:        Not affected
Vulnerability Reg file data sampling: Mitigation; Clear Register File
Vulnerability Retbleed:               Not affected
Vulnerability Spec store bypass:      Mitigation; Speculative Store Bypass disabled via prctl
Vulnerability Spectre v1:             Mitigation; usercopy/swapgs barriers and __user pointer sanitization
Vulnerability Spectre v2:             Mitigation; Enhanced / Automatic IBRS; IBPB conditional; PBRSB-eIBRS SW sequence; BHI BHI_DIS_S
Vulnerability Srbds:                  Not affected
Vulnerability Tsx async abort:        Not affected

[pip3] Could not collect
[conda] numpy                     2.0.0           py310h515e003_0    conda-forge

cc @ezyang @albanD @gqchen @nikitaved @soulitzer @Varal7 @xmfan @zou3519 @Chillee @samdow @kshitij12345

skpark-rh:
Fixes #160513

## The Problem Summary
The issue boiled down to data type promotion logic. The code base has two different functions that deal with dtype promotion logic. If it is purely multi-dimensional tensor operations, the cpp code gets triggered and that follows the numpy dtype promotion logic.  That is why in #160513 NDim tensors are fine as NDim dtypes gets precedence.  The issue came with python scalars and 0Dim tensors. When it detects "scalars", a python implementation of dtype promotion logic gets triggered (torch/_prims_common/__init__.py:1544). Since this is in python, the implementation can't distinguish what is from a wrapped tensor and a 0Dim tensor and thus will just take the highest dtype which is the python double wrapped number.

## The Fix
The python implementation for dtype promotion had to know where the scalar came from. Once the scalar can be distinguished then the appropriate dtype can be set. The first approach was to try and expose the `is_wrapped_number` method but this came with a big issue.  During the `forward_ad` the derivative of those scalars turned out to be `ZeroTensor`s.  The `ZeroTensor` internally uses a hack to initialize a meta dtype tensor which skips expensive dispatch operations. But the copy would not grab everything especially the `is_number_wrapped_` property.  I thought about modifying the copy but that seemed to go away from the spirit of what the copy was intended for and plus the tests for `is_wrapped_number_` requires `dim > 0` and a scalar `ZeroTensor` is a meta dtype tensor which complicates things.

So I chose the route of creating a new property called `was_wrapped_number` and exposed this property to the python tensor API. I had to modify the autograd code generation to set `was_wrapped_number` in the mul, add, and div operations in  `VariableType.cpp`.  Once this property was set, the dtype promotion logic could be updated to consider wrapped numbers and 0Dim numbers. Once that hierarchy was taken care of, the buggy behavior was fixed.

I wrote a new ops testing module `TestForwardADWithScalars`.  I saw that this bug was unique and required new testing paradigm. This only tests the multiply, add, and divide and I chose this because all operations boil down to these three operations.

[edit]: Just used `efficientzerotensor` meta and converted that to a python number. Since wrapped number is converted back to a python number, dtype promotion is preserved.  The constraint to achieve this happened by setting the forward grad zero tensor of a wrapped number with a wrapped number flag since the tangent of the wrapped number should still be a wrapped number. After that this specific zerotensor was then sent through as a meta type in the `BinaryOps.cpp` to get appropriate dtype for resulting arithmetic.

@ezyang @OihanJoyot


skpark-rh:
@ezyang - continuing from our conversation in the closed PR...
The first way I tried has its limitations. The `_efficientzerotensor_symint` copies the size of the input tensor but that input tensor is an empty meta tensor. Even if I passed in an option of `is_wrapped_number_`, the device_type of `ZeroTensor` is a meta device with dim=0.  So when the `set_wrap_number` is called it has a strict constraint saying that the tensor cannot be a scalar tensor and have `is_wrapped_number_` set. Either we would have to remove that constraint from the setter or modify the `_efficientzerotensor_symint` to not be a meta device type.

ezyang:
My intuition is that we should relax the is_wrapped_number setter so we are able to have a zerotensor 0d tensor be a wrapped number. This lets us repesent a zero 0d regular tensor. Do you know where the check you're running against is? I checked set_wrapped_number but it seems like it should work with 0d tensor.

skpark-rh:
In `VariableType_0.cpp:13495` -> `result_new_fw_grad_opt = other_t * self_p + self_t * other_p;` `other_t` and `self_t` becomes the `ZeroTensor`. I can set the `is_wrapped_number_` property just fine here. The problem happens when the scalar operations start to execute. Once the addition/multiplication starts to happen the code eventually lands itself in the python code,`torch/_prims_common/__init__.py:1537-1726`, to compute the correct dtype. When the code lands in the python side of things, I need to access the `is_wrapped_number_` or the `wrapped_number` property. I tried to expose the property through the python bindings but the following code, `torch/csrc/jit/python/pybind_utils.cpp:590`, prevented me from exposing the property because the incoming `ZeroTensor` was a device type `meta`.  There is a hard internal assert, `TORCH_INTERNAL_ASSERT(tensor.device().is_cpu());`, preventing me from grabbing the `is_wrapped_number_` property.

This is why I needed a new property `was_wrapped_number` to get around this constraint.  I believe this problem becomes difficult because the code suddenly goes to the python side to compute the appropriate dtypes...

ezyang:
Ok, this makes sense. I propose we relax the assert. Specifically if you have a zero tensor which is 0d then it is also ok for the device type to be meta.

Another strategy would be to avoid use of zero tensor when 0d but I do not know how easy that is.

ezyang:
we discussed this at PTC, there is a next step

skpark-rh:
@ezyang So I removed the `was_wrapped_number` property and used the `is_wrapped_number` property instead. I am exposing `is_wrapped_number` on the python side to allow the dtype promotion logic to consider wrapped python numbers and 0d tensor scalars.

aten/src/ATen/ZeroTensorFallback.cpp
```diff
@@ -86,22 +86,32 @@ namespace at {
 
       if (ivalue.isTensor()) {
         auto tensor = std::move(ivalue).toTensor();
+        bool is_wrapped_number = (tensor.unsafeGetTensorImpl()->is_wrapped_number()) ? true : false;
```
ezyang: Not sure why you have to ternary here

aten/src/ATen/ZeroTensorFallback.cpp
```diff
@@ -86,22 +86,32 @@ namespace at {
...
         for(const auto j : c10::irange(tensors.size())) {
           const Tensor& tensor = tensors[j];
+          bool is_wrapped_number = (tensor.unsafeGetTensorImpl()->is_wrapped_number()) ? true : false;
           if (tensor._is_zerotensor()) {
             // TODO: assert requires_grad=False
             //_like should not propagate zerotensor dispatch key
             TORCH_CHECK(!mut_arg, "ZeroTensors are immutable. Please use the materialized zero tensor ",
                     "obtained using .clone() if you want a mutable tensor.");
             tensors[j] = at::zeros({}, tensor.options()).expand(tensor.sizes());
+            Tensor updated_tensor = tensors[j];
```
ezyang: skip writing into `tensors[j]` and then writing into it again.

aten/src/ATen/native/TensorConversions.cpp
```diff
@@ -361,13 +361,19 @@ Tensor _to_copy(
...
         return r;
       } else if (!self.is_quantized() && self.layout() == kStrided) {
         Tensor r;
         auto strides = infer_dense_strides(self.sizes(), self.strides());
         r = at::empty_strided(
             self.sizes(), strides, options.pinned_memory(pin_out));
         r.copy_(self, non_blocking);
+        if (self.unsafeGetTensorImpl()->is_wrapped_number()) {
+          r.unsafeGetTensorImpl()->set_wrapped_number(true);
+        }
```
ezyang: These two are kind of sus, are you sure these were necessary?

tools/pyi/gen_pyi.py
```diff
@@ -1724,6 +1724,7 @@ def replace_special_case(hint: str) -> str:
             "is_mkldnn": ["is_mkldnn: _bool"],
             "is_vulkan": ["is_vulkan: _bool"],
             "is_ipu": ["is_ipu: _bool"],
+            "is_wrapped_number": ["is_wrapped_number: _bool"],
```
ezyang: It shouldn't be necessary to expose this in Python; wrapped numbers always turn into plain numbers when you get to Python. Why did you need it? Is it because of ZeroTensor wrapped number? I think ZeroTensor should turn into a plain number too.

torch/_prims_common/__init__.py
```diff
@@ -1657,22 +1659,28 @@ def _find_highest_dtype_filtered(
                 if float_as_complex and is_float_dtype(_dtype):
                     _dtype = corresponding_complex_dtype(_dtype)
                 if x.ndim == 0:
-                    zero_dim_tensor_dtype = get_higher_dtype(
-                        zero_dim_tensor_dtype, _dtype
-                    )
+                    if x.is_wrapped_number:
```
ezyang: so this shouldn't be possible here, specifically

ezyang (changes requested):
This is promising but I hope it can be tighter

ezyang (changes requested):
This is promising but I hope it can be tighter

aten/src/ATen/ZeroTensorFallback.cpp
skpark-rh: Yeah, this was dumb of me...

aten/src/ATen/ZeroTensorFallback.cpp
skpark-rh: Yep, dumb of me... Will get this updated.

aten/src/ATen/native/TensorConversions.cpp
skpark-rh: So when the copy_ gets invoked, the `is_wrapped_number` does not get copied over which I thought was weird. Without this setter, the flag gets lost.

torch/_prims_common/__init__.py
skpark-rh: So when I go through the stack trace and debug here specifically, the arguments are tensors and they are both meta devices.  I think this is because of the forward autograd operation requiring them to be tensors all the way through. I can post the stack trace up until this point.  The code uses two meta tensors to get the respective dtypes.
Screenshot From 2025-10-29 09-45-30

skpark-rh: This is why at this point, the code needs to know if those meta devices are wrapped numbers  to properly determine the right promotion.

tools/pyi/gen_pyi.py
skpark-rh: Yes the reason this is needed is because ZeroTensor is a meta tensor with an empty storage. There is no number to convert to get to the python side.

skpark-rh: I think if we turn ZeroTensor into a plain number then this become easier. I don't have to expose anything on the python side.  I do wonder what would happen to precison/accuracy if a double precision ZeroTensor gets converted to just a plain number. The resulting scalar tensor will be a lower precision if the said scalar tensor is a float.

skpark-rh: Although I do think the `mul_zerotensor` (pytorch/aten/src/ATen/native/BinaryOps.cpp:1012-1019) will be affected. Line 1017 is the one in question. I think it'll be okay if we handle the conversion in the JIT code.

aten/src/ATen/native/TensorConversions.cpp
ezyang: But if I `copy_` a wrapped number into a tensor, that tensor shouldn't turn into a wrapped number. If it was a tensor, it remains a tensor for dtype promo purposes.

tools/pyi/gen_pyi.py
ezyang: Yes just turn the ZeroTensor into a plain number??

ezyang: ZeroTensor is always zero! No precision problems

torch/_prims_common/__init__.py
ezyang: Update this function to correctly turn wrapped number Tensor IValue into plain Python numbers similar to the pybind jit code

```
std::pair<py::object, py::dict> parseIValuesToPyArgsKwargs(
    const c10::OperatorHandle& op,
    const std::vector<c10::IValue>& arguments) {
  TORCH_CHECK(
      PyGILState_Check(),
      "GIL must be held before you call parseIValuesToPyArgsKwargs");
  const auto& schema = op.schema();
  py::dict kwargs;
  // About all the pointers:
  //
  // f(int x, int y = 0, *, int z = 0)
  //                                  ^- arguments.size()
  //                        ^- kwarg_only_start
  //          ^- positional_default_start
  //   ^- 0

  // Find the split point between kwarg-only and regular.  Since most functions
  // don't have kwarg-only arguments, it is more efficient to scan from the
  // right (but ideally, this would just be precomputed in FunctionSchema
  // itself).  (NB: minus one in the loop is because we're testing if the
  // *next* argument is kwarg-only before we advance the starting index)
  int64_t kwarg_only_start = static_cast<int64_t>(arguments.size());
  for (; kwarg_only_start > 0; kwarg_only_start--) {
    const auto& arg = schema.arguments()[kwarg_only_start - 1];
    if (!arg.kwarg_only()) {
      break;
    }
  }

  // Find the first positional argument that isn't defaulted
  auto is_default = [&](size_t idx) -> bool {
    const auto& arg = schema.arguments()[idx];
    if (!arg.default_value().has_value()) {
      return false;
    }
    const auto& default_ivalue = *arg.default_value();
    const auto& ivalue = arguments[idx];
    if (default_ivalue != ivalue) {
      return false;
    }
    return true;
  };

  int64_t positional_default_start = kwarg_only_start;
  for (; positional_default_start > 0; positional_default_start--) {
```

The main part is below what I quoted. This is torch/csrc/autograd/python_variable.cpp

in

aten/src/ATen/native/TensorConversions.cpp
skpark-rh: But if the tensor you are copying has the flag `is_wrapped_number`, in this case `self`, shouldn't the resulting tensor, `r`, also have the flag set?

torch/_prims_common/__init__.py
skpark-rh: Okay I'll give it a try. I do have some concerns about how the code can recognize a meta tensor coming in as a ZeroTensor, `pytorch/aten/src/ATen/native/BinaryOps.cpp:1017`.

aten/src/ATen/native/TensorConversions.cpp
ezyang: nope

tools/pyi/gen_pyi.py
skpark-rh: I was mistaken about the ZeroTensors. What I am trying to propose is changing the dtype of `meta_out` at `aten/src/ATen/native/BinaryOps.cpp:1017` to a float.  The `self.to(device_)` is making a copy and changing the ZeroTensor to a meta tensor and so in `torch/csrc/autograd/python_variable.cpp` I am working with meta tensors and can't tell if they are ZeroTensors.  Once the meta tensors are converted to `args` for the python promotion operator, I need `is_wrapped_number` to tell what meta tensor is the wrapped number and what is the scalar tensor.  This helps me follow the numpy dtype promotion logic that handles Ndim tensors, scalar tensors, and wrapped numbers tensors.

ezyang: Yes, updating *this specific spot* to generate a wrapped number looks good.

aten/src/ATen/native/TensorConversions.cpp
skpark-rh: If I don't set this here, then the `is_wrapped_number` never shows up in the `torch/csrc/autograd/python_variable.cpp`. The `to` methods in `aten/src/ATen/native/BinaryOps.cpp:1017` do away with the `is_wrapped_number` flag from `self` and `other` and what is sent to `torch/csrc/autograd/python_variable.cpp` are just meta tensors without the flag `is_wrapped_number`.  Thus when the code reaches `torch/csrc/autograd/python_variable.cpp` there's no way to tell if any of the meta tensors coming in are from wrapped numbers.

skpark-rh: I think I get the gist of what you want. I'll modify the `mul_Tensor` in `BinaryOps.cpp` to take in the `_efficientzerotensor_meta_symint` to allow the ZeroTensor to go to the `python_variable.cpp`.  From there I'll have visibility of the ZeroTensor and can just pass in the python number.  This should eliminate the need for this setter in the copy_ here.

torch/csrc/jit/python/pybind_utils.cpp
```diff
@@ -586,8 +586,13 @@ py::object toPyObject(IValue ivalue) {
     return py::none();
   } else if (ivalue.isTensor()) {
     auto tensor = std::move(ivalue).toTensor();
-    if (tensor.unsafeGetTensorImpl()->is_wrapped_number()) {
-      TORCH_INTERNAL_ASSERT(tensor.device().is_cpu());
+    if (tensor.unsafeGetTensorImpl()->is_wrapped_number() ||
+        (tensor._is_zerotensor() &&
+         tensor.unsafeGetTensorImpl()->is_wrapped_number() &&
```
ezyang: This branch of the OR conjunction is pointless since if it's true, the lhs is always true too

torch/csrc/autograd/FunctionsManual.cpp
```diff
@@ -79,6 +79,12 @@ Tensor toNonOptPrimal(const std::optional<Tensor>& t) {
   return Tensor();
 }
 
+void update_wrapped_number(Tensor& input, Tensor& output) {
```
ezyang, line 82: severe parameter ordering blindness here lol

ezyang (approved):
looks good if ci passes

torch/csrc/jit/python/pybind_utils.cpp
skpark-rh: Yeah, that's true...

torch/csrc/autograd/FunctionsManual.cpp
skpark-rh: Yeah... for sure. T_T

PR #165790, opened 2025-10-17
[PT2 Compiler] Add flag in dynamo disable wrapper to indicate reursive disable

malaybag:
Summary: After torch._dynamo.disable is applied, wrapped method does not have any flag to indicate whether it was disabled recursively or not. This flag is needed if to preserve dynamo disable methods in torch.export-ed model

Test Plan:
```
buck test mode/opt caffe2/test/dynamo:test_dynamo -- 'test_disable_recursive_flags'
````
https://www.internalfb.com/intern/testinfra/testrun/7599824674075603

Differential Revision: D84949143

### CODE REVIEW

**Review by angelayi — CHANGES_REQUESTED** [review]
Can we hold off on landing this PR until we prototype out the entire "preserving disable in export"

**Comment by malaybag** [comment]
> Can we hold off on landing this PR until we prototype out the entire "preserving disable in export"

Yes, that makes sense. Will ping when e2e implementation is ready.

File: torch/_dynamo/decorators.py [hunk 1]
```diff
@@ -995,3 +996,10 @@ def error_on_graph_break(
     The default value of torch.compile's `error_on_graph_break` setting is False.
     """
     return ErrorOnGraphBreakDecoratorContextManager(error_on_graph_break)
+
+
+def is_dynamo_disable_recursive(method: Callable[[Any], Any]) -> bool:
+    """
+    Check if a method is marked as `dynamo_disable` recursively.
+    """
+    return getattr(method, "_torchdynamo_disable_recursive", False)
```
williamwen42: Maybe return None if there is no `_torchdynamo_disable_recursive`?

torch/_dynamo/decorators.py
```diff
@@ -1002,3 +1003,10 @@ def error_on_graph_break(
     The default value of torch.compile's `error_on_graph_break` setting is False.
     """
     return ErrorOnGraphBreakDecoratorContextManager(error_on_graph_break)
+
+
+def is_dynamo_disable_recursive(method: Callable[[Any], Any]) -> Optional[bool]:
+    """
+    Check if a method is marked as `dynamo_disable` recursively.
```
williamwen42: Document what should be returned:
- True if disable(recursive=True)
- False if disable(recursive=False)
- None if method is not a disable decorator

williamwen42 (approved):
Just a small doc nit

angelayi (approved):

PR #165792, opened 2025-10-17
[CP] Fix load balancer incorrectly assuming batch dimension exists

fegin:
* __->__ #165792

https://github.com/pytorch/pytorch/pull/163617 removes the if/else statement to check if the input buffers have the batch dimension.

This PR fixes the issue and also adds a test.

In the future, we should explicitly ask users to unsqueeze the batch dimension. This is a BC of the existing contract but implicitly infers the batch dimension existence is not safe.


torch/distributed/tensor/experimental/_attention.py
```diff
@@ -1068,27 +1068,37 @@ def _context_parallel_buffers(
...
-                    index = (
-                        load_balance_indices[0]  # identical load-balance in batch
-                        if idx_batch_size == 1
-                        else load_balance_indices[i]
-                    )
-                    buffer_batch_i = torch.index_select(
-                        buffer[i], dim=seq_dim - 1, index=index
+                if seq_dim == 0:
+                    buffer = torch.index_select(
+                        buffer, dim=0, index=load_balance_indices[0]
```
XilunWu, line 1090: alternatively, we can just `unsqueeze` inside this function

XilunWu (approved):
Thanks for catching the missing case

torch/distributed/tensor/experimental/_attention.py
fegin: I tried that, but we need to squeeze back before returning. This if/else is more straightforward to me.

PR #165793, opened 2025-10-17
[TD] Job level TD sort of

clee2000:
I tried to do a correlation analysis using copilot script: P2001937663 results: P2001936725

Jobs with the same test configs and similar arches tend to have the same results, so this PR attempts to use similar jobs as a sort of extra shard by distributing tests across similar jobs.  However, it's hard to tell if a test job exists prior to it actually running, so we have to be a bit conservative about it.  We always run the top 10% of tests for every job.  Then for the rest, we round robin them among the job groups, and also concat the round robining so that every job runs every test, just in different orders.  If we assume the job groups usually have the same results, this increases coverage when applying the threshold

Also
* move threshold determination to TD job instead of inside the test job.  previously the filter was applied after any filtering for tests that dont run on the job (ex they specify jobs, they filter some out b/c not supported by platform).  Now it happens with the rest of target determination
  * Pros: stuff like distributed tests which only get run on distributed jobs can be filtered in or out more strongly (ex td says to put distributed tests first, but if the distributed test jobs filter out non distributed tests, then the threshold is applied on this shorter list -> runs fewer distributed tests.  Now since the threshold is applied earlier, then the list will have all the tests -> threshold applied on longer list -> more distributed tests get run)
  * Cons: the pro, but in the opposite direction.  Since the threshold is applied earlier, on the longer list, then it's possible that you end up running fewer tests on some configs 
* very small threshold increase from from .25 -> .3 to help with the above

### DIFF
--- a/tools/testing/target_determination/do_td_with_job_info.py
+++ b/tools/testing/target_determination/do_td_with_job_info.py
@@ -1,0 +1,123 @@
+import re
+from pathlib import Path
+from typing import Any
+
+
+HAS_PYYAML = True
+try:
+    import yaml
+except ImportError:
+    print("Please install pyyaml to use target determination features.")
+    HAS_PYYAML = False
+
+
+REPO_ROOT = Path(__file__).resolve().parents[3]
+
+
+def get_job_info_from_workflow_file(workflow_file: str) -> list[list[dict[str, Any]]]:
+    """
+    Returns groups of jobs that are similar based on the test configurations
+    they run.
+
+    This is pretty hardcoded, so it is fragile, but it returns a pretty accurate
+    mapping
+
+    TODO replace with better (automated?) system. Maybe a separate workflow that
+    generates an artifact that says which jobs are similar according what tests
+    they run, correlation etc, also looks at jobs on main branch or merge base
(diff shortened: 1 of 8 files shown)


tools/testing/target_determination/do_td_with_job_info.py
```diff
@@ -0,0 +1,125 @@
...
+    import yaml
+except ImportError:
+    print("Please install pyyaml to use target determination features.")
+    HAS_PYYAML = False
+
+
+REPO_ROOT = Path(__file__).resolve().parents[3]
+
+
+def get_job_info_from_workflow_file(workflow_file: str) -> list[list[dict[str, Any]]]:
```
huydhn, line 17: This parsing logic is wicked :) I think we should add a mock test case for it, maybe just use `pull.yml`

huydhn (approved):

tools/testing/target_determination/do_td_with_job_info.py
clee2000: Added a test that just does brief sanity checking on pull.yml

PR #165808, opened 2025-10-18
[dynamo, nested graph breaks] fix RETURN_VALUE tx skipping in nested graph breaks

williamwen42:
* #163503
* #165597
* #163335
* #166016
* #165809
* __->__ #165808
* #166015
* #166013

Previously, we would completely skip building and calling any resume function if the leaf frame's resume instruction was RETURN_VALUE/RETURN_CONST. Now, we only skip building/calling resume functions for frames that are resuming on RETURN_VALUE.


Lucaskabela (approved):
Thanks for the comments they are helpful in trying to parse out what all these stack operations and instruction pushes do.

As a minor nit, they seem a bit shorthanded still - could we transform them to long hand so new users can onboard to this logic easier?  i.e

```
do the same with cells
```
becomes
```
Filter out the cells of skipped tx'es from codegen
```

PR #165817, opened 2025-10-18
[FlexAttention] Add mechanism to get optimal autotune decision

drisspg:
* #166359
* __->__ #165817

Script: https://github.com/meta-pytorch/attention-gym/pull/169

Feels directionally okay but there is some bike shedding / this could be quite prone to collision of keys depending on mask mod and score mod changes and simple cache key.

Usecase: https://github.com/meta-pytorch/attention-gym/pull/169 

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Chillee @yanboliang @BoyuanFeng


torch/_inductor/select_algorithm.py
```diff
@@ -2395,6 +2397,35 @@ def get_mm_log_filename() -> Optional[str]:
     return mm_file_name
 
 
+_FLEX_ATTENTION_TUNABLE_KEYS = frozenset(
```
Skylion007: Is the better abstraction here just to create a typed_dict class and statically go over the keys maybe? With a default sentinel value? Or maybe that's just too complicated.

test/inductor/test_flex_attention.py
```diff
@@ -6860,6 +6863,120 @@ def test_flex_attention_with_dynamic_max_autotune(self, device):
     def test_flex_attention_with_dynamic_max_autotune_graph_partition(self, device):
         self._test_flex_attention_with_dynamic_max_autotune(device)
 
+    @skip_on_cpu
+    def test_flex_attention_logging(self, device):
```
Skylion007, line 6917: There is a tmpdir pytest fixture you can use, I think just adding arg the test, tmpdir or tmp_dir might trigger it

drisspg: I dont think pytorch supports this fixture

Chillee (approved):

PR #165820, opened 2025-10-18
Set prefer_deferred_runtime_asserts_over_guards to True

justinchuby:
Set prefer_deferred_runtime_asserts_over_guards to True and allow a flag to control the behavior, just in case.

This option has enable the gemma3 model export with transformers==4.57. I am not sure how best to test it though.

cc @titaiwangms

### DIFF
--- a/torch/onnx/_flags.py
+++ b/torch/onnx/_flags.py
@@ -48,3 +48,8 @@ ENABLE_DRAFT_EXPORT: bool = _load_boolean_flag(
     this_will="enable torch.export.draft_export as a strategy for capturing models",
     default=False,
 )
+PREFER_DEFERRED_RUNTIME_ASSERTS_OVER_GUARDS: bool = _load_boolean_flag(
+    "TORCH_ONNX_PREFER_DEFERRED_RUNTIME_ASSERTS_OVER_GUARDS",
+    this_will="set prefer_deferred_runtime_asserts_over_guards when calling torch.export",
+    default=True,
+)
--- a/torch/onnx/_internal/exporter/_capture_strategies.py
+++ b/torch/onnx/_internal/exporter/_capture_strategies.py
@@ -166,6 +166,7 @@ class TorchExportStrictStrategy(CaptureStrategy):
                     kwargs=kwargs,
                     dynamic_shapes=dynamic_shapes,
                     strict=True,
+                    prefer_deferred_runtime_asserts_over_guards=_flags.PREFER_DEFERRED_RUNTIME_ASSERTS_OVER_GUARDS,
                 )
             except torch._dynamo.exc.UserError as exc:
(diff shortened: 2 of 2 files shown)


torch/onnx/_internal/exporter/_capture_strategies.py
```diff
@@ -215,6 +221,7 @@ def _capture(
                     kwargs=kwargs,
                     dynamic_shapes=dynamic_shapes,
                     strict=False,
+                    prefer_deferred_runtime_asserts_over_guards=_flags.PREFER_DEFERRED_RUNTIME_ASSERTS_OVER_GUARDS,
```
titaiwangms, line 224: This is how we activate it?

justinchuby: By default it is True (see flag definition). We can turn it off by setting the envvar.

titaiwangms: Would yhere be any potential regression for other models? I am not sure what this does..

justinchuby: In case of regression yes. I am not sure what could cause it. But if you prefer just setting it to true I can do that too

titaiwangms: I thought default was False?

https://github.com/pytorch/pytorch/blob/1891239a1d876b15f6f97be7ada9aac1f84dab38/torch/export/__init__.py#L173

https://github.com/pytorch/pytorch/blob/1891239a1d876b15f6f97be7ada9aac1f84dab38/torch/_dynamo/config.py#L280

torch/onnx/_flags.py
```diff
@@ -48,3 +48,8 @@ def _load_boolean_flag(
     this_will="enable torch.export.draft_export as a strategy for capturing models",
     default=False,
 )
+PREFER_DEFERRED_RUNTIME_ASSERTS_OVER_GUARDS: bool = _load_boolean_flag(
+    "TORCH_ONNX_PREFER_DEFERRED_RUNTIME_ASSERTS_OVER_GUARDS",
+    this_will="set prefer_deferred_runtime_asserts_over_guards when calling torch.export",
+    default=True,
```
justinchuby, line 54: @titaiwangms default is set here

titaiwangms: I mean default by torch, it's False? I am wondering if we change the default, would it cause regression?

justinchuby: True, the default in torch is False. If the arg does what it claims to do though it is just going to not enforce the guards during export but instead make them runtime asserts (which we ignore). I think that’s a reasonable thing to do. I propose that we at least try it and run the benchmarks.

titaiwangms (approved):

PR #165839, opened 2025-10-18
Fix/faketensor deepcopy

### LINKED ISSUE #165547: error message for deepcopy a cuda tensor under FakeTensorMode is pretty cryptic
```
>>> y = torch.randn([], device="cuda:0")
>>> with FakeTensorMode():
...     copy.deepcopy(y)
...
ERROR:torch._subclasses.fake_tensor:failed while attempting to run meta for aten.set_.source_Storage
  File "/home/rzou/dev/2.8/env/lib/python3.12/site-packages/torch/_subclasses/fake_tensor.py", line 2717, in _dispatch_impl
    r = func(*args, **kwargs)
        ^^^^^^^^^^^^^^^^^^^^^
  File "/home/rzou/dev/2.8/env/lib/python3.12/site-packages/torch/_ops.py", line 829, in __call__
           ^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Attempted to set the storage of a tensor on device "meta" to a storage on different device "cuda:0".  This is no longer allowed; the devices must match.
```
to be clear I don't think this should be supported but at least it should throw something nicer

cc @chauhang @penguinwu @eellison @bdhirsh

adityasingh345:
Fixes #165547

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @aditvenk @jeffdaily @sunway513 @jithunnair-amd @pruthvistony @ROCmSupport @jataylo @hongxiayang @naromero77amd @jerrymannil @xinyazhang @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @aditew01 @gujinghui @PenghuiCheng @jianyuh @min-jean-cho @yanbing-j @Guobing-Chen @Xia-Weiwen @snadampal @voznesenskym @penguinwu @EikanWang @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela @dllehr-amd @chenyang78


Lucaskabela (changes requested):
Can we isolate these changes to just the fake tensor? Weird amount of updates to other submodules

adityasingh345:
> Can we isolate these changes to just the fake tensor? Weird amount of updates to other submodules

Thanks @Lucaskabela! I’ve cleaned up unrelated submodule changes and limited the fix to FakeTensorMode. Please re-review.”

adityasingh345:
Fix: Handle deepcopy safely for real tensors in FakeTensor.from_real_tensor 
when a non-meta tensor passed to FakeTensor.from_real_tensor, it can lead to a storage mismatch or invalid memory reference. to fix this, the real tensors are converted to meta tensors before fakification, while preserving dtype, shape, requires_grad. Deepcopy of real tensors can fail because Fake Tensors expects a meta tensor. 

# Explanation:
- Convert real CPU/CUDA tensors to meta to avoid storage mismatch and preserve shape, dtype, requires_grad, and contiguity.
- Restore symbolic context from active TracingContext if missing, so deepcopy maintains tracing info.
- Wrap tensor as FakeTensor and memoize, ensuring safe deepcopy and consistency inside FakeTensorMode.

torch/_subclasses/fake_tensor.py
```diff
@@ -381,13 +381,44 @@ def from_real_tensor(
...
+        if t.device.type != "meta":
+    # convert real tensor (CPU/CUDA) to meta to avoid storage mismatch
+            t_meta = torch.empty(
+            t.shape,
+            dtype=t.dtype,
+            device="meta",
+            requires_grad=t.requires_grad
+            )
+    # Preserve contiguity if needed
+        if t.is_contiguous():
```
thenumberouscode: I'am just curious. Does torch.empty not return a contiguous format by default?

adityasingh345: Thanks for the note! I added .contiguous() to be safe with meta tensors and potential non-standard layouts. Just to confirm — should I go ahead and remove it, or keep it for consistency with other FakeTensor code paths?
I added .contiguous() as a defensive step to ensure consistent strides in case of meta or symbolic tensors, but I can remove it if we prefer to keep the code cleaner.

torch/_subclasses/fake_tensor.py
```diff
@@ -381,13 +381,44 @@ def from_real_tensor(
         if maybe_memo is not None:
             return maybe_memo
         # not yet supported in metatensors
+        # --- HANDLE deepcopy / real tensors safely ---
```
Lucaskabela: I tend to avoid these `---` comments - lets simplify to just an inline one :)

torch/_subclasses/fake_tensor.py
```diff
@@ -381,13 +381,44 @@ def from_real_tensor(
...
+        if t.device.type != "meta":
+    # convert real tensor (CPU/CUDA) to meta to avoid storage mismatch
+            t_meta = torch.empty(
```
Lucaskabela: Formatting seems inconsitent (probably needs to run `lintrunner init` then `lintrunner -m origin/main -a`

torch/_subclasses/fake_tensor.py
```diff
@@ -381,13 +381,44 @@ def from_real_tensor(
...
+                    symbolic_context = tracing_context.tensor_to_context[t]
+                    from torch.fx.experimental.symbolic_shapes import StatefulSymbolicContext
+                    assert isinstance(symbolic_context, StatefulSymbolicContext)
+                    source = symbolic_context.tensor_source
+
+    # -----------------------------
+    # Wrap as FakeTensor and return
+    # -----------------------------
+        from torch._subclasses.fake_tensor import FakeTensor
+        fake = FakeTensor(t, example_value=constant)
```
Lucaskabela: When creating a new fake tensor, we should run this inside the `FakeTensorMode` context so that we are properly tracking it

test/test_fake_tensor.py
```diff
@@ -94,7 +94,9 @@ def checkType(self, t, device_str, size):
         self.assertTrue(isinstance(t, FakeTensor))
         self.assertEqual(t.device.type, device_str)
         self.assertEqual(list(t.size()), size)
+
```
Lucaskabela: Can we add the issue as a unit test and include in the testplan to show that we are fixing this isue?

Lucaskabela (changes requested):
Still some unrelated changes here (from package-lock.json); lets clean these up please!

Lucaskabela:
I would also wonder (similar to what is voiced by Richard) if trying to support this deepcopy is the right approach or might result in some memory issues; maybe instead when we detect this we should be graph breaking/raising an error?

torch/_subclasses/fake_tensor.py
adityasingh345: Okk I will apply the changes

torch/_subclasses/fake_tensor.py
adityasingh345: Ok I will improve and modify it

adityasingh345:
> I would also wonder (similar to what is voiced by Richard) if trying to support this deepcopy is the right approach or might result in some memory issues; maybe instead when we detect this we should be graph breaking/raising an error?

thank you for raising it.

I see the issue now. Supporting deepcopy could indeed lead to unexpected memory usage, which goes against the purpose of FakeTensor. It also makes sense that it's not a standard graph operation, and trying to support it silently could be problematic.

I agree that raising an error is a much safer.

test/test_fake_tensor.py
adityasingh345: done add a file test_fake_tensor_dummy.py

torch/_subclasses/fake_tensor.py
adityasingh345: is it good now ?

torch/_subclasses/fake_tensor.py
thenumberouscode: oh, I think remove it may be reasonable, you can use empty_like to insure the memory format

adityasingh345: Ah, torch.empty_like is a much cleaner solution!. It handles the memory format automatically and lets me remove the extra if t.is_contiguous(): block.

I'll make that change right now. Thanks for the great suggestion sir.

zou3519 (changes requested):
deepcopying a real tensor under a faketensormode should error unless there's a good usecase for it. there's room to provide a better error message, that's what my issue was about.

test/test_fake_tensor_dummy.py
```diff
@@ -0,0 +1,19 @@
+
+import torch
+import copy
+from torch._subclasses.fake_tensor import FakeTensorMode
+
+# ... inside the test class ...
+
+def test_deepcopy_real_tensor_in_fake_mode(self):
```
Lucaskabela: this can probably be moved to the `test_fake_tensor` class instead

torch/_subclasses/fake_tensor.py
```diff
@@ -2363,7 +2385,23 @@ def _dispatch_impl(
                 "FakeTensorMode unrecognized subclass(es): %s", unrecognized_types
             )
             return NotImplemented
-
+        # -
```
Lucaskabela: Is this change still needed here?

Lucaskabela (changes requested):
package-lock change still here let's be sure to remove it

Lucaskabela (changes requested):
See comment; needs formatting and some potential to further minimize changes

torch/_subclasses/fake_tensor.py
```diff
@@ -377,17 +377,37 @@ def from_real_tensor(
                     assert isinstance(symbolic_context, StatefulSymbolicContext)
                     source = symbolic_context.tensor_source
 
-        maybe_memo = self._get_memo(t)
+        maybe_memo = fake_mode._get_memo(t)
         if maybe_memo is not None:
             return maybe_memo
         # not yet supported in metatensors
+        # HANDLE deepcopy / real tensors safely
+        if t.device.type != "meta":
```
Lucaskabela: Trying to understand here why the other changes are also necessary - this one makes sense but what are the other additions for? 

Also, let's run `lintrunner init` and `lintrunner -m origin/main -a` before pushing to catch linter errors

Lucaskabela (changes requested):
Thanks for making progress on this - I think we have sufficiently minimized to relevant files, but I am still not sure if all these changes are necessary - can you help me understand why we need the changes (besides `NotImplementedError`)

torch/_subclasses/fake_tensor.py
```diff
@@ -377,17 +377,37 @@ def from_real_tensor(
                     assert isinstance(symbolic_context, StatefulSymbolicContext)
                     source = symbolic_context.tensor_source
 
-        maybe_memo = self._get_memo(t)
```
Lucaskabela, line 380: This should be able to stay as `self` right?

torch/_subclasses/fake_tensor.py
```diff
@@ -377,17 +377,37 @@ def from_real_tensor(
...
+            )
+
         if t.is_quantized:
             raise UnsupportedFakeTensorException("quantized nyi in meta tensors")
         if type(t) is torch.nn.Parameter:
             assert not make_constant
 
         constant = t if make_constant else None
 
+
```
Lucaskabela: What is this needed for here? Shouldn't this be handled above already?

torch/_subclasses/fake_tensor.py
```diff
@@ -2364,6 +2386,22 @@ def _dispatch_impl(
             )
             return NotImplemented
 
+        if func == torch.ops.aten.set_.source_Storage:
```
Lucaskabela: Similar comment here - what is this logic for? My understanding would be we should just generally error instead of trying to patch the device to meta?

Lucaskabela (changes requested):
Left some feedback where I would like to see clarification on why the code needs to change - for the lintrunner, perhaps your local branch is outdated? can you try `git pull` first then rerun the init/lintrunner application?

If that doesn't work, you can see the signals with `lintrunner` here and make the suggested changes manually.  Hope that helps

torch/_subclasses/fake_tensor.py
adityasingh345: Thanks for catching this! I double-checked the code path, and t.is_quantized or torch.nn.Parameter aren’t handled in any of the earlier branches of from_real_tensor().

torch/_subclasses/fake_tensor.py
adityasingh345: yes sir

torch/_subclasses/fake_tensor.py
```diff
@@ -381,13 +382,34 @@ def from_real_tensor(
...
+            )
+
         if t.is_quantized:
             raise UnsupportedFakeTensorException("quantized nyi in meta tensors")
         if type(t) is torch.nn.Parameter:
             assert not make_constant
 
         constant = t if make_constant else None
 
+        # Wrap as FakeTensor and return
```
Lucaskabela: I do not understand why this is needed - please remove

torch/_subclasses/fake_tensor.py
```diff
@@ -2364,6 +2387,20 @@ def _dispatch_impl(
             )
             return NotImplemented
 
+        if func == torch.ops.aten.set_.source_Storage:
```
Lucaskabela: I do not understand why this is needed - please remove

test/test_fake_tensor.py
```diff
@@ -115,6 +127,10 @@ def test_basic(self):
             self.assertEqual(z.device, torch.device("cpu"))
             self.assertTrue(isinstance(z, FakeTensor))
 
+
```
Lucaskabela: Please see the formatting changes suggested in the `lintrunner` tests and manually patch them

Lucaskabela (changes requested):
Added very concrete feedback here - please address these direct comments before pinging again

adityasingh345:
> Added very concrete feedback here - please address these direct comments before pinging again

Thanks @Lucaskabela — I’ve rebased with main, removed the unnecessary block, fixed all lint issues, and verified tests locally. Please take another look when you get a chance.

test/test_fake_tensor.py
```diff
@@ -105,8 +117,8 @@ def test_cuda_initialized(self):
             y.backward()
 
     def test_basic(self):
-        x = torch.empty(2, 2, device="cpu")
-        y = torch.empty(4, 2, 2, device="cpu")
+        x = torch.empty(2, 2, device="meta")
```
Lucaskabela: Why do these need to change?

torch/_subclasses/fake_tensor.py
```diff
@@ -381,13 +381,33 @@ def from_real_tensor(
         if maybe_memo is not None:
             return maybe_memo
         # not yet supported in metatensors
+        # HANDLE deepcopy / real tensors safely
```
Lucaskabela: I think we need to re-examine the original issue here - this change is preventing us from creating from_real_tensor for any non-meta tensors.  This is not really how we want to use `from_real_tensor` - we want to only error when we are calling `deepcopy()` on a tensor in FakeTensorMode that is not meta device

Lucaskabela: In otherwords, I think this logic needs to move to intercept where `deepcopy()` gets called - perhaps we can breakpoint in the torch_dispatch to find where that is?

Lucaskabela (changes requested):
Still many changes that seem unrelated (docs/jupyter_execute)

Now that we have the changes minimized though, I think we need to revisit where the change is applied

test/test_fake_tensor.py
adityasingh345: These were changed to use the "meta" device since FakeTensorMode tests are intended to run without real tensor allocations. Using "meta" ensures the tensors are lightweight and consistent with how FakeTensor operates internally, instead of allocating memory on "cpu".

Lucaskabela (changes requested):
What happened to our unit test checking behavior here? Can we re-add it?

Also there are other files still touched here - we can probably minimize to just `fake_tensor.py` and `test_fake_tensor.py`

torch/_subclasses/fake_tensor.py
```diff
@@ -366,7 +366,7 @@ def from_real_tensor(
         trace: bool = True,
     ) -> FakeTensor:
         # see note [Tensor Fakification and Symbol Caching]
-        if not symbolic_context and not source and shape_env:
+        if not symbolic_context and not source and not shape_env:
```
Lucaskabela, line 369: Why does this need to change to not?

docs/requirements.txt
Lucaskabela, line 1: More misc files changed please remove those

Lucaskabela (changes requested):

torch/_subclasses/fake_tensor.py
adityasingh345: sir i have changed this because the previous condition only ran when shape_env was already provided, but this block mean to handle the case where no shape_env exists  yet.
the check now correctly run the symbolic context inference when shape_env is none, which matches. 

sir if it will be shape_env it means it will run only if shape_env exists -- skips needed inference 

and if it not shape_env run when shape_env is missin -- triggers symbolic inference. 

if i am wrong please tell me about my mistakes.

adityasingh345:
made the change please review my pr @Lucaskabela

PR #165840, opened 2025-10-18
[BE][Ez]: Update Eigen to 5.0.0. C++14 support and more!

Skylion007:
Update Eigen pin to 5.0.0 . Tons of new features and perf improvements. Most importantly updates minimum from C++03 to C++14 giving a ton of performance optimizations like properly implemented move operators, simplified code, etc. Also improved vectorization particularily on ARM. We really only use this library as a fallback for sparse operators, but still useful to update it.


albanD (approved):
Thx!

PR #165841, opened 2025-10-18
[BE][Ez]: Update torch.is_tensor documentation

Skylion007:
TypeIs propogates the isinstance check with the typing system. They are now equivalent.

cc @svekars @sekyondaMeta @AlannaBurke


albanD (approved):
Sure

PR #165842, opened 2025-10-18
[BE]: Update kleidai submodule to v1.15.0

Skylion007:
This mostly just adds a few new kernels and fixes some IMA and performance improvement of prev kernels. Also improves compiler support.


albanD (approved):
Thx

PR #165843, opened 2025-10-18
Implement an AOT precompile mode for standalone_compile

jamesjwu:
* #165844
* __->__ #165843

This PR introduces an `aot` flag to standalone_compile that uses BundledAOTAutogradCacheEntry, and then allows regional_inductor to use this so that we can start aot compiling regional compiler graphs. The diff above this will attempt to allow GraphPickler to fully serialize graphs that have regionally compiled subgraphs. 


torch/_inductor/standalone_compile.py
```diff
@@ -187,13 +188,84 @@ def load(
             return CompiledArtifact(lambda *args: compiled_fn(list(args)), None)
 
 
+class AOTCompiledArtifact(CompiledArtifact):
```
oulgen, line 264: this literally overrides every method of the base class, maybe make an interface and have both of these be implementations?

alternatively, can this class just be inlined into the  CompiledArtifact class?

oulgen (changes requested):
tests?

jamesjwu:
There are existing tests in `test_regional_inductor.py` to make sure compilation still works. It's hard to test serialization end to end without implementing GraphPickler above this diff, but I can write some basic unit tests as well for the main data structure.

jamesjwu:
- Update so that torch._inductor.CompiledArtifact.load and torch._inductor.standalone_compile's interfaces are BC compatible, i.e. old usages do not break at all.

torch/fx/passes/regional_inductor.py
```diff
@@ -56,13 +58,12 @@ def _compile_submod(gm, prefix):
 
             submod = getattr(gm, node.target)
 
-            # _dummy_wrapper is to make call_function happy
-            compiled_submod = _dummy_wrapper(
-                torch._inductor.standalone_compile(
-                    submod, fake_inputs, dynamic_shapes="from_tracing_context"
-                )
+            compiled_fn = torch._inductor.standalone_compile(
+                submod, fake_inputs, dynamic_shapes="from_tracing_context", aot=True
```
oulgen, line 62: why do we want this True?

oulgen (approved):

torch/fx/passes/regional_inductor.py
jamesjwu: TLDR: in regional inductor, we want to be able to load and save entire GraphModules that have multiple nodes each of which call standalone compile. Using AOT mode means that it's much easier to save the bigger graph module with all of its compile artifacts, instead of relying on indirect caching

PR #165844, opened 2025-10-18
Allow GraphPickler to pickle graph modules containing AOTCompiled subgraphs

jamesjwu:
* #166002
* #165996
* __->__ #165844
* #165843

This PR allows GraphPickler to pickle aot_eager graph modules that have regional inductor bits in them, with a few exceptions: 
- FlexAttentionBackward isn't marked cacheable, so those tests don't work immediately since we're not sure how to serialize it. But it's safe to serialize/cache, so the next PR fixes those unit tests. 
- It seems that when reloading a GraphPickled object, we don't recompile subgraphs. Will investigate this in a future PR

All unit tests in test_regional_inductor are parameterized so that we try serializing and deserializing the returned graph module before returning. 


torch/_inductor/utils.py
```diff
@@ -2226,15 +2226,15 @@ def run_and_get_code(
 ) -> tuple[_T, list[str]]:
     from .graph import GraphLowering
 
-    source_codes: list[str] = []
+    source_codes: OrderedSet[str] = OrderedSet()
```
jamesjwu, line 2229: In the serialize path, we will call output_codes twice: the first time before serialization, and a second time on deserialization. There's no particular reason we need duplicate source_codes so I changed this to be a set. Will confirm no other test failures because of this

oulgen (approved):

PR #165851, opened 2025-10-19
Enable PLW0127 in ruff

cyyever:
This PR enables `PLW0127` in ruff, which checks self-assignment of variables with the form `var=var`.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela

### DIFF
--- a/torch/nn/attention/__init__.py
+++ b/torch/nn/attention/__init__.py
@@ -36,7 +36,6 @@ WARN_FOR_UNFUSED_KERNELS = False
 
 # Hacks for Sphinx documentation:
 # https://stackoverflow.com/questions/38765577/overriding-sphinx-autodoc-alias-of-for-import-of-private-class
-SDPBackend = SDPBackend
 r"""An enum-like class that contains the different backends for scaled dot product attention.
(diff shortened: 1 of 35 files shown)


Lucaskabela (approved):
LGTM

torch/nn/attention/__init__.py
```diff
@@ -36,7 +36,6 @@
 
 # Hacks for Sphinx documentation:
 # https://stackoverflow.com/questions/38765577/overriding-sphinx-autodoc-alias-of-for-import-of-private-class
-SDPBackend = SDPBackend
```
albanD, line 39: Should we remove the comment above if we're removing this?

cyyever: Yes we should, let me remove them.

PR #165856, opened 2025-10-19
Move AT_DISPATCH_V2 helper macros to headeronly and add THO_DISPATCH_V2_TMPL

pearu:
Problem: the migration of `AT_DISPATCH_V2` macros to headeronly cannot be a simple copy-paste of macro definitions from one header file to another because the macros `AT_DISPATCH_SWITCH` and `AT_DISPATCH_CASE` may use functions that cannot be migrated to headeronly, e.g. when a selective build feature is enabled, there will be functions that are generated. On the other hand, when not using selective build, the dtype-dispatch macros are perfectly suitable for migrating to headeronly.

In this PR, the migration problem above is tackled by refactoring `AT_DISPATCH` related macros into headeronly macros and non-headeronly macros while preserving the current API and semantics. For instance, consider the current V2 macro definitions:
```c++
#define AT_DISPATCH_V2(TYPE, NAME, BODY, ...) \
  AT_DISPATCH_SWITCH(TYPE, NAME, AT_AP_VAR(AT_WRAP(BODY), TYPE, __VA_ARGS__))
#define AT_AP_VAR(N, T, ...) \
  AT_EXPAND(AT_CONCAT(AT_AP, AT_NUM_ARGS(__VA_ARGS__))(AT_WRAP(N), __VA_ARGS__))
#define AT_AP1(N, _1) AT_DISPATCH_CASE(_1, N)
...
```
where the headeronly-migration-problematic parts are using AT_DISPATCH_SWITCH and AT_DISPATCH_CASE macros (defined in ATen/Dispatch.h). In this PR, we introduce parametric versions of `AT_DISPATCH_V2` and `AT_AP1` macros that have `_TMPL` suffices, have DISPATCH_SWITCH and DISPATCH_CASE arguments, and are define in `torch/headeronly/core/Dispatch_v2.h`:
```c++
#define THO_DISPATCH_V2_TMPL(                               \
    DISPATCH_SWITCH, DISPATCH_CASE, TYPE, NAME, BODY, ...) \
  DISPATCH_SWITCH(                                         \
      TYPE,                                                \
      NAME,                                                \
      THO_AP_VAR_TMPL(DISPATCH_CASE, AT_WRAP(BODY), TYPE, __VA_ARGS__))
#define THO_AP_VAR_TMPL(C, N, T, ...) \
  AT_EXPAND(                         \
      AT_CONCAT(THO_AP, AT_NUM_ARGS(__VA_ARGS__))(C, AT_WRAP(N), __VA_ARGS__))
#define THO_AP1(C, N, _1) C(_1, N)
...
```
so that original V2 macro definition, defined in ATen/Dispatch_v2.h,  becomes:
```c++
#define AT_DISPATCH_V2(TYPE, NAME, BODY, ...) \
  THO_DISPATCH_V2_TMPL(                        \
      AT_DISPATCH_SWITCH,                     \
      AT_DISPATCH_CASE,                       \
      TYPE,                                   \
      NAME,                                   \
      AT_WRAP(BODY),                          \
      __VA_ARGS__)
```
that has exactly the same API and semantics as the original definition.

Note 1: ~we have changed the definition of `AT_AP1(N, _1) ...` to `AT_AP1(C, N, _1) ...` without renaming `AT_AP1` because `AT_AP1` is a helper macro that is not a part of public API (for instance, nothing in pytorch explicitly uses `AT_AP1`).~ UPDATE: restored the original `AT_AP` macros and introduced new `THO_AP` macros.

Note 2: this PR introduces a new API macro THO_DISPATCH_V2_TMPL that will be available for stable ABI users who can use it by providing custom versions of `AT_DISPATCH_SWITCH` and `AT_DISPATCH_CASE macros, say, with selective build features removed.

* #166629
* __->__ #165856


torch/headeronly/core/Dispatch_v2.h
janeyx99, line 1: Can you split this file up into what APIs are new (not in the original header) and how they are different from the non header only APIs? Right now it is confusing what is a migration vs what APIs have inherently changed.

From what I understand of the current PR, the generated macros, like `AT_AP1` will have different behavior depending on whether the headeronly version or the Aten version of the file is included, which is not the desired semantic. I'd prefer having two sets of different APIs that remain consistent with themselves.

pearu: > Can you split this file up into what APIs are new (not in the original header) and how they are different from the non header only APIs? Right now it is confusing what is a migration vs what APIs have inherently changed.

I have addressed this question in the PR description by providing the migration problem and how this PR resolves it. The aim is not to change APIs during the migration, but see the AT_AP1 note below.

> From what I understand of the current PR, the generated macros, like `AT_AP1` will have different behavior depending on whether the headeronly version or the Aten version of the file is included, which is not the desired semantic. I'd prefer having two sets of different APIs that remain consistent with themselves.

Note that `AT_AP1` is a helper macro that is not part of the public API and in fact is never used explicitly in torch source code. Hence, changing `AT_AP1` by introducing a new argument should be safe. If we would want to migrate really safely, we can introduce `AT_AP1_TMPL` as in
```c++
#define AT_AP1_TMPL(C, N, _1) C(_1, N)
#define AT_AP1(N, _1) AT_AP1_TMPL(AT_DISPATCH_CASE, N, _1)
```
so that `AT_AP1` will stay equivalent to its current definition, both API and semantics wise.

janeyx99: However, having AT_DISPATCH_V2 do different things based on what headers are included is inherently confusing/not explicit. Even if it makes migration a little harder (API names would have to be find+replace'd), the explicitness is worth keeping.

pearu: > However, having AT_DISPATCH_V2 do different things based on what headers are included is inherently confusing/not explicit.

This PR merely moves the AT_DISPATCH V2 helper macros to headeronly but the definition of `AT_DISPATCH_V2` macro still stays in `ATen/Dispatch_v2.h` as it uses non-headeronly `AT_DISPATCH_SWITCH/CASE` macros (I agree that the title of this PR was misleading). So, 
"having AT_DISPATCH_V2 do different things based on what headers are included" is incorrect as only one of the header files provides `AT_DISPATCH_V2`.

> Even if it makes migration a little harder (API names would have to be find+replace'd), the explicitness is worth keeping.

This PR has two header files, torch/headeronly/core/Dispatch_v2.h and ATen/Dispatch_v2.h, and I guess from the comment that your assumption is that including torch/headeronly/core/Dispatch_v2.h gives the AT_DISPATCH_V2 macro. With this PR, this is not true and it cannot be so in general. The headeronly Dispatch_v2.h only provides `AT_DISPATCH_V2_TMPL` macro that users may use to define custom DISPATCH_V2 macros, similarly to how `AT_DISPATCH_V2` is defined in ATen/Dispatch_v2.h, just with custom DISPATCH_SWITCH/CASE macros.

That said, to make torch/headeronly/core/Dispatch_v2.h useful for stable ABI programs, it should provide, say, a `THO_DISPATCH_V2` macro without the selective build support, exactly as is the MY_DISPATCH_V2 macro defined in test/cpp/aoti_abi_check/test_dispatch.cpp.
I suggest implementing this feature (provided that the naming `THO_DISPATCH_V2` is ok) in a follow-up PR as it is a new public feature. What do you think?

pearu: https://github.com/pytorch/pytorch/pull/166629 adds THO_DISPATCH_V2.

torch/headeronly/core/Dispatch.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
+
+#include <torch/headeronly/core/ScalarType.h>
+
+#define AT_PRIVATE_CASE_TYPE_USING_HINT_TMPL(PRELUDE, enum_type, HINT, ...)   \
```
janeyx99:
```suggestion
#define THO_PRIVATE_CASE_TYPE_USING_HINT_TMPL(PRELUDE, enum_type, HINT, ...)   \
```
and below too for clarity

torch/headeronly/core/Dispatch.h
```diff
@@ -0,0 +1,40 @@
...
+}
+} // namespace detail
+
+#define AT_DISPATCH_SWITCH_TMPL(                                            \
+    PRELUDE, CHECK_NOT_IMPLEMENTED, TYPE, NAME, ...)                        \
+  [&] {                                                                     \
+    const auto& the_type = TYPE;                                            \
+    constexpr const char* at_dispatch_name = NAME;                          \
+    /* don't use TYPE again in case it is an expensive or side-effect op */ \
+    c10::ScalarType _st = ::detail::scalar_type(the_type);                  \
```
janeyx99:
```suggestion
    torch::headeronly::ScalarType _st = ::detail::scalar_type(the_type);                  \
```

torch/headeronly/core/Dispatch_v2.h
```diff
@@ -0,0 +1,148 @@
...
+// documentation): AT_DISPATCH_V2_TMPL extends AT_DISPATCH_V2 with
+// extra DISPATCH_SWITCH and DISPATCH_CASE arguments for specifying
+// custom implementations of the original AT_DISPATCH_SWITCH and
+// AT_DISPATCH_CASE macros. Use the provided macros
+// AT_DISPATCH_SWITCH_TMPL and AT_DISPATCH_CASE_TMPL to define the
+// custom implementations of the switch and case macros, respectively.
+
+// Public API macros
+
+#define AT_DISPATCH_V2_TMPL(                               \
```
janeyx99: same here let's make the TMPL's THO_DISPATCH_V2_TMPL and etc.

torch/headeronly/core/Dispatch_v2.h
```diff
@@ -0,0 +1,148 @@
...
+      NAME,                                                \
+      AT_AP_VAR_TMPL(DISPATCH_CASE, AT_WRAP(BODY), TYPE, __VA_ARGS__))
+
+// This macro lets you pass an arbitrary expression that may contain internal
+// commas to another macro without having the commas causing the expression
+// to be interpreted as being multiple arguments
+#define AT_WRAP(...) __VA_ARGS__
+
+#define AT_FLOAT8_TYPES                                                 \
+  c10::ScalarType::Float8_e5m2, c10::ScalarType::Float8_e5m2fnuz,       \
```
janeyx99: torch::headeronly instead of c10 please!

torch/headeronly/core/Dispatch_v2.h
```diff
@@ -0,0 +1,148 @@
...
+#define AT_QINT_TYPES \
+  c10::ScalarType::QInt8, c10::ScalarType::QUInt8, c10::ScalarType::QInt32
+// NB: not *actually* all types
+#define AT_ALL_TYPES AT_EXPAND(AT_INTEGRAL_TYPES), AT_EXPAND(AT_FLOATING_TYPES)
+#define AT_ALL_TYPES_AND_COMPLEX \
+  AT_EXPAND(AT_ALL_TYPES), AT_EXPAND(AT_COMPLEX_TYPES)
+
+// Helper macros
+
+#define AT_AP_VAR_TMPL(C, N, T, ...) \
```
janeyx99: rename to THO... and document how it's diff from AT_AP_VAR

torch/headeronly/core/Dispatch_v2.h
```diff
@@ -0,0 +1,148 @@
...
+    print(f'#define AT_AP{i}(C, N, {args}) {cases}')
+
+#endif
+
+// Begin generated code
+// clang-format off
+
+#define AT_NUM_ARGS(...) AT_EXPAND(AT_NUM_ARGS_AUX(__VA_ARGS__, 60, 59, 58, 57, 56, 55, 54, 53, 52, 51, 50, 49, 48, 47, 46, 45, 44, 43, 42, 41, 40, 39, 38, 37, 36, 35, 34, 33, 32, 31, 30, 29, 28, 27, 26, 25, 24, 23, 22, 21, 20, 19, 18, 17, 16, 15, 14, 13, 12, 11, 10, 9, 8, 7, 6, 5, 4, 3, 2, 1, 0))
+#define AT_NUM_ARGS_AUX(_1, _2, _3, _4, _5, _6, _7, _8, _9, _10, _11, _12, _13, _14, _15, _16, _17, _18, _19, _20, _21, _22, _23, _24, _25, _26, _27, _28, _29, _30, _31, _32, _33, _34, _35, _36, _37, _38, _39, _40, _41, _42, _43, _44, _45, _46, _47, _48, _49, _50, _51, _52, _53, _54, _55, _56, _57, _58, _59, _60, N, ...) N
+#define AT_AP1(C, N, _1) C(_1, N)
```
janeyx99: Curious: where are these used?

These are also subtlely different from AT_AP... so we shouldn;t delete those and we should rename these.

janeyx99 (commented):
Thanks. This is indeed closer.

torch/headeronly/core/Dispatch_v2.h
pearu: AT_AP1, etc are used in AT_AP_VAR and AT_AP_VAR_TMPL macros. Notice the AT_CONCAT result in
```
#define AT_AP_VAR_TMPL(C, N, T, ...) \
  AT_EXPAND(                         \
      AT_CONCAT(AT_AP, AT_NUM_ARGS(__VA_ARGS__))(C, AT_WRAP(N), __VA_ARGS__))
```

torch/headeronly/core/Dispatch.h
pearu: Replaced `AT_` prefix with `THO_`.

torch/headeronly/core/Dispatch.h
pearu: Applied.

torch/headeronly/core/Dispatch_v2.h
pearu: Applied.

torch/headeronly/core/Dispatch_v2.h
pearu: Fixed.

torch/headeronly/core/Dispatch_v2.h
pearu: Done.

torch/headeronly/core/Dispatch_v2.h
pearu: `AT_AP` macros are now restored and added `THO_AP` macros.

aten/src/ATen/Dispatch_v2.h
```diff
@@ -74,41 +79,26 @@
...
+      AT_DISPATCH_SWITCH,                     \
+      AT_DISPATCH_CASE,                       \
+      TYPE,                                   \
+      NAME,                                   \
+      AT_WRAP(BODY),                          \
+      __VA_ARGS__)
+
+// Helper macros, kept for BC:
 #define AT_WRAP(...) __VA_ARGS__
+#define AT_EXPAND(X) X
```
janeyx99: nit: These can be header only--no need to keep two versions of these.

torch/headeronly/core/Dispatch_v2.h
```diff
@@ -0,0 +1,161 @@
...
+    DISPATCH_SWITCH, DISPATCH_CASE, TYPE, NAME, BODY, ...) \
+  DISPATCH_SWITCH(                                         \
+      TYPE,                                                \
+      NAME,                                                \
+      THO_AP_VAR_TMPL(DISPATCH_CASE, THO_WRAP(BODY), TYPE, __VA_ARGS__))
+
+// This macro lets you pass an arbitrary expression that may contain internal
+// commas to another macro without having the commas causing the expression
+// to be interpreted as being multiple arguments
+#define THO_WRAP(...) __VA_ARGS__
```
janeyx99: Since this is just a normal helper, it doesn't really need the THO prefix. It could just be WRAP and EXPAND--what do you think?

torch/headeronly/core/Dispatch_v2.h
```diff
@@ -0,0 +1,161 @@
...
+// Helper macros
+
+// THO_AP_VAR_TMPL is same as AT_AP_VAR but with a custom
+// DISPATCH_CASE macro argument:
+#define THO_AP_VAR_TMPL(C, N, T, ...)                       \
+  THO_EXPAND(THO_CONCAT(THO_AP, THO_NUM_ARGS(__VA_ARGS__))( \
+      C, THO_WRAP(N), __VA_ARGS__))
+#define THO_CONCAT(a, b) THO_CONCAT_AUX(a, b)
+#define THO_CONCAT_AUX(a, b) a##b
+#define THO_EXPAND(X) X
```
janeyx99: these are also just helpers we don't need to duplicate.

torch/headeronly/core/Dispatch_v2.h
```diff
@@ -0,0 +1,161 @@
...
+// Python code to regenerate generate code below:
+#if 0
+
+num_args = 60
+
+nums = ', '.join(str(i) for i in reversed(range(num_args+1)))
+args = ', '.join(f'_{i}' for i in range(1, num_args+1))
+
+print(f'#define THO_NUM_ARGS(...) THO_EXPAND(THO_NUM_ARGS_AUX(__VA_ARGS__, {nums}))')
+print(f'#define THO_NUM_ARGS_AUX({args}, N, ...) N')
```
janeyx99: this bit is the same as the AT bits right? like AT_NUM_ARGS is the same as THO_NUM_ARGS? Would it be possible to define these APIs in this file and update the ATen version to use these?

janeyx99 (commented):
Approach is good, there are some things we could share more though i've commented on below

aten/src/ATen/Dispatch_v2.h
pearu: AT_WRAP/EXPAND are now defined in header only.

torch/headeronly/core/Dispatch_v2.h
pearu: I renamed these back to AT_WARP/EXPAND as these are public macros that users may need. 

In fact, for BE, AT_WRAP could have a more general usage as it is a known workaround to a CPP quirk where commas in code blocks that are passed as macro arguments need to be protected from misinterpretation. For instance, the following code

https://github.com/pytorch/pytorch/blob/a51208c656fb3e9a8b091a4d181f9a9cda783c04/aten/src/ATen/core/TensorAccessor.h#L261-L268

could be simplified to

```c++
// Old name for `GenericPackedTensorAccessor`
template <typename T, size_t N, template <typename U> class PtrTraits = DefaultPtrTraits, typename index_t = int64_t>
C10_DEFINE_DEPRECATED_USING(PackedTensorAccessor, AT_WRAP(GenericPackedTensorAccessor<T, N, PtrTraits, index_t>))
```

torch/headeronly/core/Dispatch_v2.h
pearu: Removed duplication.

Btw, `AT_EXPAND` is generally useful while `AT_CONCAT` has only a specific use in this file.

torch/headeronly/core/Dispatch_v2.h
pearu: Correct and yes. Your suggestion is applied.

aten/src/ATen/Dispatch_v2.h
janeyx99, line 1: lines 108 and 109 can be deleted, right?

janeyx99 (approved):

janeyx99:
thanks @pearu !!

aten/src/ATen/Dispatch_v2.h
pearu: Yes, good catch! Removed.

PR #165869, opened 2025-10-19
Add "where" to arg_parser

sumantro93:
Fixes #165799


torch/csrc/utils/python_arg_parser.cpp
```diff
@@ -119,7 +119,8 @@ bool should_allow_numbers_as_tensors(const std::string& name) {
       "floor_divide",
       "floor_divide_",
       "floor_divide_out",
-      "_conj"}; // _conj needed because mul.Tensor backward calls it
+      "_conj", // _conj needed because mul.Tensor backward calls it
+      "where"};
```
albanD, line 123: cc @zou3519 any concern with adding this one?

albanD (changes requested):
The removed file is definitely wrong.

Also we would need to add test to this in common_method_invocations.py by adding new sample inputs there for where.

torch/csrc/utils/python_arg_parser.cpp
sumantro93: I can add the test, yes the file deletion was wrong.. :( I will fix this PR

PR #144651, opened 2025-01-12
[MPS] lu factor ex implementation

Isalia20:
Implements `torch.linalg.lu_factor_ex`

PR #99269, opened 2023-04-16
[MPS] Add lu_factor

qqaatw:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #99269

### <samp>🤖 Generated by Copilot at d75cde1</samp>

Added MPS support and autograd formulas for LU factorization of tensors. Implemented the `linalg_lu_factor` and `linalg_lu_factor.out` functions for the MPS backend in `LinearAlgebra.mm` and added tests in `test_mps.py`. Added the corresponding dispatch entries in `native_functions.yaml` and the backward and forward formulas in `derivatives.yaml`.

PR #165871, opened 2025-10-19
[MPS] Fix parity between CPU and MPS on singular matrices in linalg.lu_factor

### LINKED ISSUE #165870: [MPS] torch.linalg.lu_factor on singular matrices different between CPU and MPS
### 🐛 Describe the bug

Follow up from https://github.com/pytorch/pytorch/pull/165254#issuecomment-3408776145.

`lu_factor` on a singular matrix should result in an error like it does on the CPU, but currently does not:
```python
import torch

t = torch.tensor([[1.0, 2.0], [2.0, 4.0]], device="mps")
torch.linalg.lu_factor(t)
```

On the CPU:
> RuntimeError: torch.linalg.lu_factor: U[2,2] is zero and using it on lu_solve would result in a division by zero. If you still want to perform the factorization, consider calling linalg.lu(A, pivot) or linalg.lu_factor_ex(A, pivot)

### Versions

```
PyTorch version: 2.9.0
Is debug build: False

OS: macOS 15.6.1 (arm64)
Clang version: 17.0.0 (clang-1700.3.16.4)
CMake version: version 3.31.5
Libc version: N/A

Python version: 3.12.8 (main, Mar 24 2025, 16:58:11) [Clang 16.0.0 (clang-1600.0.26.6)] (64-bit runtime)
Python platform: macOS-15.6.1-arm64-arm-64bit
Is CUDA available: False
CUDA runtime version: No CUDA
cuDNN version: No CUDA
Is XPU available: False
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M4 Max

[pip3] numpy==2.3.3
[pip3] torch==2.9.0
[conda] Could not collect
```

inventshah:
Fixes #165870. Follow up from #165254.

This PR [a] removes the MPS specific version of `lu_factor` in favor of the version in BatchedLinearAlgebra.cpp which uses `lu_factor_ex`, and [b] updates `lu_factor_ex` error codes to match expectations.

When `lu_factor` was first implemented for MPS (#99269), it bypassed the implementation in BatchedLinearAlgebra.cpp since we did not have `lu_factor_ex`. Since #144651 implements `lu_factor_ex`, we can now remove the MPS specific wrapper.

### DIFF
--- a/aten/src/ATen/native/mps/operations/LinearAlgebra.mm
+++ b/aten/src/ATen/native/mps/operations/LinearAlgebra.mm
@@ -338,6 +338,8 @@ static void linalg_lu_factor_ex_out_mps_impl(const Tensor& A,
           ". See https://developer.apple.com/documentation/metalperformanceshaders/mpsmatrixdecompositionstatus for details.");
     }
   }
+
+  map_mps_decomposition_error_code_to_blas(info);
 }
 
 static void linalg_solve_out_mps_impl(const Tensor& A,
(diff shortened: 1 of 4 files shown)


kulinseth (approved):

albanD (approved):
Nice simplification, thanks!

PR #165874, opened 2025-10-19
[rfc] add debug mode to print meta in fx graphs

bobrenjc93:
* __->__ #165874
* #165893

quite useful in debugging things like unbacked bindings (and presumably other mechanisms that dependent on meta including activation checkpointing and stack trace printing)

CleanShot 2025-10-21 at 09 41 54@2x


ezyang (approved):
suuuure why not

PR #165883, opened 2025-10-20
[CI] Update ONNX CI packages to latest

cyyever:
This PR updates ONNX related packages to their latest versions used in CI environments.


justinchuby (approved):

albanD (approved):
That sounds ok.
Not sure if we should do like other dependencies and run the ci on the oldest supported version though?
But will let @justinchuby decide on this one.

cyyever:
@albanD We want to update these environment packages to test the latest ONNX packages because third-party ONNX has been in 1.18 for a long time.

PR #165886, opened 2025-10-20
[xpu][test]port embedding indexing and native_mha test files for Intel GPU

wincent8:
we port test_indexing, test_native_mha and test_embedding for Intel GPU in this pr.
We could enable Intel GPU with following methods and try the best to keep the original code styles:

Use torch.accelerator for general gpu
Skip the case if running on xpu which has known issues
using torch.nn.attention.sdpa_kernel() to replace torch.backends.cuda.sdp_kernel() for Intel GPU as torch.backends.cuda.sdp_kernel() is depricated and Intel xpu did not support it.

cc @frank-wei @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10

### DIFF
--- a/test/nn/test_embedding.py
+++ b/test/nn/test_embedding.py
@@ -13,8 +13,8 @@ from torch.testing._internal.common_device_type import (
     dtypesIfXPU,
     instantiate_device_type_tests,
     largeTensorTest,
-    onlyCUDA,
     onlyNativeDeviceTypes,
+    onlyOn,
     skipCUDAIf,
     skipMeta,
     skipXPUIf,
@@ -316,6 +316,7 @@ class TestEmbeddingNNDeviceType(NNTestCase):
                 torch.nn.functional.embedding(indices, weight)
 
     @dtypesIfCUDA(torch.float16, torch.float64)
+    @dtypesIfXPU(torch.float16, torch.float64)
     @dtypes(torch.float64)
     def test_embedding_backward(self, device, dtype):
         embedding = nn.Embedding(10, 3, sparse=True)
@@ -356,6 +357,13 @@ class TestEmbeddingNNDeviceType(NNTestCase):
             else (torch.float, torch.double, torch.half)
         )
     )
+    @dtypesIfXPU(
+        *(
+            (torch.float, torch.double, torch.bfloat16, torch.half)
+            if TEST_WITH_ROCM
+            else (torch.float, torch.double, torch.half)
+        )
+    )
     @dtypes(torch.float32)
     def test_embedding_max_norm_backward(self, device, dtype):
         # can't use gradcheck since in place renorm makes analytical gradients different from produced ones
@@ -380,6 +388,13 @@ class TestEmbeddingNNDeviceType(NNTestCase):
             else (torch.float, torch.double, torch.half)
         )
     )
+    @dtypesIfXPU(
+        *(
+            (torch.float, torch.double, torch.bfloat16, torch.half)
+            if TEST_WITH_ROCM
+            else (torch.float, torch.double, torch.half)
+        )
+    )
     @dtypes(torch.float32)
     def test_embedding_max_norm_fwd_AD(self, device, dtype):
         if torch.device(device).type == "xla":
@@ -398,6 +413,13 @@ class TestEmbeddingNNDeviceType(NNTestCase):
         self.assertEqual(jvp, expected_grad)
 
     @dtypesIfCUDA(
+        *(
+            (torch.float, torch.double, torch.bfloat16, torch.half)
+            if TEST_WITH_ROCM
+            else (torch.float, torch.double, torch.half)
+        )
+    )
+    @dtypesIfXPU(
         *(
             (torch.float, torch.double, torch.bfloat16, torch.half)
             if TEST_WITH_ROCM
@@ -496,6 +518,7 @@ class TestEmbeddingNNDeviceType(NNTestCase):
     @onlyNativeDeviceTypes
     @dtypes(torch.float32, torch.float64)
     @dtypesIfCUDA(torch.half, torch.bfloat16)
+    @dtypesIfXPU(torch.half, torch.bfloat16)
     def test_embedding_bag_1D_padding_idx(self, device, dtype):
         num_features = 3
         max_indices_per_bag = 10
@@ -640,11 +663,12 @@ class TestEmbeddingNNDeviceType(NNTestCase):
                     weights.grad, weights_check.grad, msg=msg, atol=atol, rtol=rtol
                 )
 
-    @onlyCUDA
+    @onlyOn(["cuda", "xpu"])
     @dtypes(
         torch.bfloat16,
     )
     @largeTensorTest("80GB", device="cuda")
+    @largeTensorTest("80GB", device="xpu")
     def test_embedding_backward_large_batch_overflow(self, device, dtype):
         """
         Test that embedding_dense_backward handles large batches that exceed INT32_MAX thread IDs.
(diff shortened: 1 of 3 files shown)


test/test_indexing.py
```diff
@@ -598,8 +601,9 @@ def validate_setting(x):
 
         # test invalid index fails
         reference = torch.empty(10, dtype=dtype, device=device)
-        # can't test cuda because it is a device assert
-        if not reference.is_cuda:
+        print(reference.device.type, flush=True)
```
guangyey: remove debug info

test/test_indexing.py
```diff
@@ -598,8 +601,9 @@ def validate_setting(x):
...
+        # can't test cuda/xpu because it is a device assert
+        if not reference.is_cuda and not reference.device.type == "xpu":
```
guangyey:
```suggestion
        if reference.device.type == "cpu":
```

test/nn/test_embedding.py
```diff
@@ -348,6 +357,13 @@ def test_embedding_backward(self, device, dtype):
             else (torch.float, torch.double, torch.half)
         )
     )
+    @dtypesIfXPU(
+        *(
+            (torch.float, torch.double, torch.bfloat16, torch.half)
+            if TEST_WITH_ROCM
```
guangyey: TEST_WITH_ROCM is used to differ nvdia card or amd card. We should not use this.

test/nn/test_embedding.py
```diff
@@ -372,6 +388,13 @@ def test_embedding_max_norm_backward(self, device, dtype):
...
```
guangyey: ditto

test/nn/test_embedding.py
```diff
@@ -396,6 +419,13 @@ def test_embedding_max_norm_fwd_AD(self, device, dtype):
...
```
guangyey: ditto

test/nn/test_embedding.py
wincent8: i refer to the cuda code
    @dtypesIfCUDA(
        *(
            (torch.float, torch.double, torch.bfloat16, torch.half)
            if TEST_WITH_ROCM
            else (torch.float, torch.double, torch.half)
        )
    )

test/nn/test_embedding.py
```diff
@@ -1391,7 +1451,7 @@ def run_tests(mode, sparse, trainable_per_sample_weights):
             run_tests(mode, sparse, trainable_per_sample_weights)
 
         # Test CUDA Dense on half precision
-        if device == "cuda":
+        if device == "cuda" or device == "xpu":
```
guangyey:
```suggestion
        if device != "cpu":
```

test/nn/test_embedding.py
```diff
@@ -1582,7 +1649,7 @@ def test_embedding_bag_device(self, device, dtypes):
             )
 
             test_backward = False
-            if self.device_type == "cuda":
+            if self.device_type == "cuda" or self.device_type == "xpu":
```
guangyey:
```suggestion
            if self.device_type != "cpu":
```

test/test_indexing.py
wincent8: got it, thanks

test/test_indexing.py
```diff
@@ -2361,8 +2377,8 @@ def test_broaderrors_indexing(self, device):
     def test_trivial_fancy_out_of_bounds(self, device):
         a = torch.zeros(5, device=device)
         ind = torch.ones(20, dtype=torch.int64, device=device)
-        if a.is_cuda:
-            raise unittest.SkipTest("CUDA asserts instead of raising an exception")
+        if a.is_cuda or a.device.type == "xpu":
```
guangyey:
```suggestion
        if a.device.type in ["cuda", "xpu"]:
```

test/test_indexing.py
wincent8: if we using cpu here, we can not make sure if there will be a issue on other device type like HPU?

test/test_native_mha.py
```diff
@@ -285,23 +292,41 @@ def test_native_multihead_self_attention(self, device, dtype, use_nt,
...
-                        dtype,
-                        "self",
-                        use_nt=use_nt,
-                        use_padding=use_padding,
-                        pad_all=pad_all,
-                        need_weights=need_weights,
-                        average_attn_weights=average_attn_weights,
-                    )
+                if device == "cuda":
+                    with torch.backends.cuda.sdp_kernel(
```
guangyey: Could `torch.backends.cuda.sdp_kernel` be replaced by `torch.nn.attention.sdpa_kernel`?

test/nn/test_embedding.py
guangyey: Please remove it.

test/test_indexing.py
guangyey: I think the other accelerator should have the same behavior as CUDA.

test/test_native_mha.py
wincent8: my goal is only changing the things related to xpu, but yes, i can have a try to change cuda related also

test/test_indexing.py
wincent8: got it, thanks for suggestion.

test/nn/test_embedding.py
wincent8: ok

test/test_indexing.py
wincent8: ok, that will make things easy.

guangyey (approved):
Thanks for your update!
LGTM.

test/nn/test_embedding.py
```diff
@@ -1604,6 +1604,8 @@ def _test_EmbeddingBag(
     def test_embedding_bag_device(self, device, dtypes):
         if IS_JETSON and torch.bfloat16 in dtypes and device == "cpu":
             self.skipTest("bfloat16 not supported with Jetson cpu")
+        if dtypes == (torch.int32, torch.int32, torch.float64) and "xpu" in device:
```
guangyey: two int32?

wincent8: yes, it's the case original designed.

test/nn/test_embedding.py
```diff
@@ -29,14 +30,21 @@
     run_tests,
     set_default_dtype,
     skipIfTorchDynamo,
+    TEST_CUDA,
+    TEST_XPU,
+)
+
+
+device_type = (
```
albanD, line 38: Don't we have a generic utility for that already?

test/nn/test_embedding.py
```diff
@@ -348,6 +357,7 @@ def test_embedding_backward(self, device, dtype):
             else (torch.float, torch.double, torch.half)
         )
     )
+    @dtypesIfXPU(torch.float, torch.double, torch.half)
```
albanD: nit: for consistency, please use float32 instead of float.

albanD (approved):
Small nit. SGTM

test/nn/test_embedding.py
guangyey: Yes. We can do this using `torch._C._get_accelerator().type`.

guangyey: However, `torch._C._get_accelerator()` could poison fork, `torch.accelerator.current_accelerator(True)` does not. So, it would differ behavior in the distribution test.

guangyey: I will try to add a utility properly.

test/nn/test_embedding.py
wincent8: thanks for reminder, updated.

PR #165888, opened 2025-10-20
Add `tabulate` as a PyTorch runtime dependency

xuantengh:
Fixes #144564, cc @zou3519


malfet (changes requested):
First we need an agreement, that out of number of options we want to standardize on `tabulate`

PR #165891, opened 2025-10-20
[2/N] Change C-style casts to static_cast or reinterpret_cast

cyyever:
A follow-up of #165750 to clean up C casts.


aten/src/ATen/native/GridSampler.cpp
```diff
@@ -487,13 +487,13 @@ static Tensor _grid_sampler_2d_cpu_quantized(
   int64_t out_sC = output.stride(1);
   int64_t out_sH = output.stride(2);
   int64_t out_sW = output.stride(3);
-  uint8_t* inp_ptr = (uint8_t*)input.data_ptr<quint8>();
-  uint8_t* out_ptr = (uint8_t*)output.data_ptr<quint8>();
+  const uint8_t* inp_ptr = reinterpret_cast<const uint8_t*>(input.data_ptr<quint8>());
```
Skylion007: Why not just use?
```suggestion
  const uint8_t* inp_ptr = input.const_data_ptr<uint8_t>());
```
or fi that is not supported:
```suggestion
  const uint8_t* inp_ptr = reinterpret_cast<const uint8_t*>(input.const_data_ptr<quint8>());

File: aten/src/ATen/cudnn/Descriptors.cpp [hunk 2]
```diff
@@ -138,7 +138,7 @@ void FilterDescriptor::set(const at::Tensor &t, const at::MemoryFormat memory_fo
 
   int size[CUDNN_DIM_MAX];
```
Skylion007: Nit should probably be std::array.

cyyever: Fixed

cyyever:
@Skylion007 Applied your advice

aten/src/ATen/native/GridSampler.cpp
```diff
@@ -487,13 +487,13 @@ static Tensor _grid_sampler_2d_cpu_quantized(
   int64_t out_sC = output.stride(1);
   int64_t out_sH = output.stride(2);
   int64_t out_sW = output.stride(3);
-  uint8_t* inp_ptr = (uint8_t*)input.data_ptr<quint8>();
-  uint8_t* out_ptr = (uint8_t*)output.data_ptr<quint8>();
+  const uint8_t* inp_ptr = input.const_data_ptr<uint8_t>();
+  uint8_t* out_ptr = output.data_ptr<uint8_t>();
   float* grid_ptr = grid.data_ptr<float>();
```
Skylion007: I think these grid_ptr should also be const even though they aren't currently? Right?

aten/src/ATen/native/cudnn/GridSampler.cpp
```diff
@@ -61,7 +61,7 @@ void setSamplerDescriptor(
     const at::Tensor& tensor) {
   int inputSize[4] = {0};
```
Skylion007: Should probably also be an std::array.

Skylion007 (approved):

Skylion007 (commented):
Remaining points are nits. Feel free to merge those and fix other stuff in later PR.

aten/src/ATen/native/GridSampler.cpp
cyyever: Fixed

aten/src/ATen/native/cudnn/GridSampler.cpp
cyyever: Fixed

PR #125623, opened 2024-05-06
Memoize local_scalar_dense calls, refactor all memos

ezyang:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #125623

Signed-off-by: Edward Z. Yang <ezyang@meta.com>

PR #165893, opened 2025-10-20
[hops] fix unbacked runtime asserts for cond higher order op

bobrenjc93:
* __->__ #165893

At a high level after this fix we get the following nice tlparse https://manifold.edge.x2p.facebook.net/v0/read/tree/logs/bobren/54a57665-7dcc-41e0-8ca7-df01393cd4aa/custom/index.html?bucketName=tlparse_reports&apiKey=tlparse_reports-key&withPayload=1&timeoutMsec=10000

As seen in this doc, previously we were simply dropping assert post
dynamo: https://docs.google.com/document/d/1nRQwvw_gWL0_9T3VKb5Ly3_tNI1fgqG9WtryeD6qaZI/edit?tab=t.0

The fixes are a couple things:

1) Actually run the runtime assertion fx graph pass on subgraphs
2) Reset fake mode unbacked memo across speculate subgraph invocations
   since the memos actually break the runtime assertion insertions since
   calls like nonzero end up not allocating new unbacked symints and
   hence not populating pending_unbacked which then results in incorrect
   unbacked_bindings on fx_nodes in subgraphs.

This is a first step in hardening runtime asserts across all phases of
the compiler (eager, aot_eager, inductor, etc.). I will continue kicking
tires and fixing bugs until we get runtime assert generations in a good
place. One obvious next step is the added test case in this PR fails
when compiled with inductor with the following error (NB: it fails before this PR as well):

```
  File "/data/users/bobren/a/pytorch/torch/_inductor/ir.py", line 659, in get_dtype
    return self.dtype
torch._dynamo.exc.BackendCompilerFailed: backend='inductor' raised:
LoweringException: AttributeError: 'ShapeAsConstantBuffer' object has no attribute 'dtype'
  target: cond
  args[0]: Eq(Mod(s77, 4), 0)
  args[1]: Subgraph(name='true_graph_0', graph_module=<lambda>(), graph=<torch._inductor.graph.SubgraphLowering object at 0x7fbcbb11e110>)
  args[2]: Subgraph(name='false_graph_0', graph_module=<lambda>(), graph=<torch._inductor.graph.SubgraphLowering object at 0x7fbcbb21cf70>)
  args[3]: (s77, TensorBox(StorageBox(
    ComputedBuffer(name='buf0', layout=FlexibleLayout('cuda:0', torch.float32, size=[s77, s77], stride=[s77, 1]), data=Pointwise(device=device(type='cuda', index=0), dtype=torch.float32, inner_fn=<function make_pointwise.<locals>.inner.<locals>.inner_fn at 0x7fbcbb2f37f0>, ranges=[s77, s77]))
  )))
```


torch/_dynamo/variables/higher_order_ops.py
```diff
@@ -1371,6 +1371,11 @@ def speculate_branch(branch):
                 supports_aliasing=self.supports_aliasing,
             )
 
+            # need to ensure we increase epoch so we don't memoize unbacked bindings
+            # across different subgraphs which can interfere with runtime assertion
+            # generation.
+            tx.fake_mode.epoch += 1
```
zou3519, line 1377: Could you explain the epoch thing a bit more please? Why does incrementing it "Reset [the] fake mode unbacked memo " ?

bobrenjc93: This idea of memoizing data dependent op calls was originally introduced by @ezyang in https://github.com/pytorch/pytorch/pull/125623/files

It's helpful in the case where you are repeatedly calling the same data dependent operation in the same graph, but in the HOP case we are actually memoizing across different graphs. For example we'd call nonzero in the True graph and memoize the call such that in the False graph doesn't allocate a new unbacked symbol and thus doesn't properly set unbacked_bindings on the nonzero fx node.

```
@register_op_impl(torch.ops.aten.nonzero.default)
def nonzero(fake_mode, func, arg):
    if (
        fake_mode.shape_env is None
        or not fake_mode.shape_env.allow_dynamic_output_shape_ops
    ):
        # Without symints/symfloats, cannot handle this
        raise DynamicOutputShapeException(func)

    if (nnz := arg.nonzero_memo) is None:
```

This is problematic since our runtime assert fx pass depends on these unbacked bindings to figure out what asserts to insert into the graph

```
@compatibility(is_backward_compatible=True)
def insert_deferred_runtime_asserts(
    gm: GraphModule,
    shape_env: ShapeEnv,
    name: str,
    export: bool = False,
) -> None:

            ...

            if unbacked_bindings := resolve_unbacked_bindings(
                shape_env, node.meta.get("unbacked_bindings")
            ):
                for s, keypath in unbacked_bindings.items():
                    defs.append(s)
```

In practice this meant, even after my fix to run the fx graph passes, we were only inserting the asserts in one of the subgraphs. The memo epoch increase resets these memos such that we do allocate new unbacked symbols in both branches resulting in proper unbacked_bindings population resulting in correct runtime assert insertions.

Happy to add this comment to the code but not sure if it'd be too verbose lol

ydwu4: Will it cause problem for parent graph since now we would have different  tx.fake_mode.epoch when resume tracing parent graph?

In fact, this is a general problem for hop x dynamic shape. We need to be able to checkpoint and resume fake_mode/shape_env.  Another real use case from Executorch is to support different shape assumptions made in two branches. e.g. 
```python
def true_fn(x):
  return x @ self.small_param # (4, 8)
 
def false_fn(x):
  return x @ self.large_param # (128, 8)

torch.cond(x.shape[1] == 4, true_fn, false_fn)
```


After tracing true_fn or false_fn:

1. no new guards should be introduced to parent graph because cond shoudn't assume which branch is taken at tracing time and the decision of which branch to run is only available when the graph sees a real input. Parent graph's shape env should stay unchanged. 
2. We want to insert run time assertions in the traced graph of true_fn and false_fn even for backed symints.

A workaround for this use case is to export two/more programs with each of them specialize on x.shape in a different way but they are not happy about this and x.shape[1] could be unbacked.

bobrenjc93: 1) My understanding of the memoization is it's a mechanism to reduce number of unbacked allocations to make it easier on sympy (cc @ezyang for verification). Allocating more symbols across HOP boundaries shouldn't be a soundness problem. In contrast, today we actually do have a soundness problem where runtime asserts are lost post dynamo and rehydrated via the ShapeEnv during inductor. Crucially this means aot_eager is missing runtime asserts (see PR description). This PR fixes this with the slight tradeoff of a few more unbacked allocations, but I think that's a worthy tradeoff to make.
2) It is also true that subgraphs introducing runtime asserts can cause problems today. That's a separate problem that I hopeful to have a PR up for soon where we will prepend the torch.cond's predicate as a prelude to all runtime asserts in the subgraph.

zou3519 (approved):
I don't completely follow but I think this improves the current state (we're not silently dropping the asserts) so let's ship it

PR #165897, opened 2025-10-20
[OpenReg][Feat][Docs] Enrich OpenReg device management implementation and add focused documentation

KarhouTam:
## Summary
This PR enriches OpenReg device management codes and adds focused documentation.

## Key Changes
- Introduced device management documentation in `device.md`.
- Updated `OpenRegFunctions.h` and `OpenRegFunctions.cpp` to use `DeviceIndex` and added error handling.
- Implemented `check_device_index` function for validating device indices.
- Enhanced Python bindings in `Module.cpp` for device management.
- Added tests for invalid device index handling in `test_device.py`.

### DIFF
--- a/docs/source/accelerator/device.md
+++ b/docs/source/accelerator/device.md
@@ -2,37 +2,34 @@
 
 ## Background
 
-Device management is a foundational component of PyTorch's accelerator support system, enabling PyTorch to interact with different hardware accelerators in a unified manner. For custom accelerators like OpenReg, device management provides the essential infrastructure for:
+Device management handles basic operations like querying how many devices are available and switching between them. Accelerator backends need to wrap their device runtime's APIs and expose them to PyTorch.
 
-- **Device Discovery and Enumeration**: Identifying how many accelerator devices are available in the system
-- **Device Context Management**: Tracking and switching between different devices during computation
-- **Error Handling and Validation**: Ensuring device operations are safe and providing meaningful error messages
-- **Integration with PyTorch Ecosystem**: Enabling device-aware features like tensors, streams, events, and automatic mixed precision
-
-The OpenReg device management implementation (`OpenRegFunctions.h/cpp`) serves as an adapter layer that wraps the underlying third-party OpenReg runtime and exposes a PyTorch-compatible C++ API. This module provides simple, well-tested functions that higher layers (streams, events, generators, Python bindings) rely on to manage device state throughout the backend.
+The OpenReg implementation ([`OpenRegFunctions.h/cpp`][OpenReg Device Management]) shows how to wrap a third-party runtime. These functions are used throughout the backend - by streams, events, generators, and Python bindings.
 
 ## Design
 
-The device management system must support the following basic functionalities:
+Accelerator vendors need to implement these core functions:
 
-| Functionality            | Description                                               | Application Scenarios                                                                                          |
-| ------------------------ | --------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------- |
-| **Device Count Query**   | Query the total number of available devices in the system | - Application initialization<br>- Multi-device workload distribution<br>- Validating device indices before use |
-| **Current Device Query** | Get the currently active device for the calling thread    | - Debugging and logging<br>- Determining tensor placement<br>- Guard implementations                           |
-| **Set Device**           | Change the active device for subsequent operations        | - Switching context between devices<br>- Initializing specific device resources<br>- Multi-GPU training loops  |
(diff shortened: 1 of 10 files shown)


test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegFunctions.h
```diff
@@ -9,10 +9,20 @@
...
-OPENREG_EXPORT c10::DeviceIndex current_device();
-OPENREG_EXPORT void set_device(c10::DeviceIndex device);
+OPENREG_EXPORT DeviceIndex device_count() noexcept;
+OPENREG_EXPORT DeviceIndex current_device();
+OPENREG_EXPORT void set_device(DeviceIndex device);
 
 OPENREG_EXPORT DeviceIndex ExchangeDevice(DeviceIndex device);
 
+static inline void check_device_index(int64_t device) {
+  TORCH_CHECK((device >= 0 && device < c10::openreg::device_count()) || device == -1,
```
fffrog: Why do we need to explicitly allow device equal to -1?

KarhouTam: Sorry about that. I was thought that `-1` is the default and allowed. Fix already. Thanks!

test/cpp_extensions/open_registration_extension/torch_openreg/torch_openreg/openreg/__init__.py
```diff
@@ -42,7 +42,8 @@ def current_device():
 
 
 def set_device(device) -> None:
-    return torch_openreg._C._set_device(device)
+    if device >= 0:
```
fffrog, line 46: Please don't let this failure happen silently.

KarhouTam: Well. I think this can be set as silently, like `torch.cuda.set_device()` does. All negatives are going to do nothing.
https://github.com/pytorch/pytorch/blob/4a94591321a62b31658f9259680de87e4be9e50d/torch/cuda/__init__.py#L564-L576
Or maybe add a warn or something else to hint the user? Because in the end, negative device index is invalid. Wdyt?

docs/source/accelerator/device.md
```diff
@@ -0,0 +1,64 @@
+# Device Management
```
fffrog, line 1: Thank you for writing the documentation on "Device Management".

We should follow this outline:

## Background
Describe the background of Device Management.

## Design
Please use a table to list all the basic functionalities the new accelerator needs to support, along with their descriptions and application scenarios.

## Implementation

Use just one function as an example to illustrate how we should implement it.

### Python Side

How to implement it on the Python side.

### C++ Side

How to implement it in C++.

## Integration

### Python

These functionalities need to be exposed from C++ to Python for easier user access. Please use one function as an example to illustrate how to expose it to Python.

Also, a table should be provided showing the list of APIs that need to be exposed to Python, along with their descriptions and application scenarios.

### Guard

In addition to being used in Python, these functionalities also need to be integrated into Guard. Guard is crucial within PyTorch and in many other places.

Show an example of how to integrate it.

KarhouTam: Will modify doc asap! Thanks for your reviewing.

KarhouTam: Docs modified

KarhouTam:
Hi, @fffrog.  I think this PR should be merged after the hook one: #165980 . Cause there must have conflict in `index.md`, but this PR is enoughly fine to be merged (at least all CI passed).

docs/source/accelerator/device.md
```diff
@@ -0,0 +1,122 @@
...
+```
+
+**Key implementation points:**
+
+1. **Error Checking**: All OpenReg runtime calls are wrapped with `OPENREG_CHECK`, which converts error codes to exceptions
+2. **Optimization**: Avoids redundant device switching by checking if the target device is already current
+3. **Bounds Validation**: The public `set_device()` function calls `check_device_index()` to validate the device index is within valid range
+4. **Caching**: `device_count()` uses static initialization to cache the device count, avoiding repeated system calls
+
+### Python Side
```
fffrog: Thank you so much for the update, it looks great.

Could you move the part about exposing C++ to Python to after the C++ part and before the Python part? This is because these three parts are presented as a whole to represent the entire device integration process.

fffrog (approved):
LGTM, thank you a lot.

Just some tiny nits.

docs/source/accelerator/device.md
KarhouTam: Got it.

PR #165707, opened 2025-10-17
[dynamo][misc] Replace UserFunctionVariable with VariableTracker build

anijain2305:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #165707
* #166251

Audit: To prevent future issues with functools.partial or callable
objects.

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela

PR #165903, opened 2025-10-20
[dynamo] Clean up assert in dynamo [3/N]

can-gaa-hou:
Some previous PRs have been merged. This PR aims for some **assert** that the users can trigger, and it may be better to turn them into a graph break. Correct me if there are any problems.

* ->#165903(Clean up for graph break)
* #165745
* #165430

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela @williamwen42

### DIFF
--- a/torch/_dynamo/variables/ctx_manager.py
+++ b/torch/_dynamo/variables/ctx_manager.py
@@ -877,20 +877,11 @@ class DisabledSavedTensorsHooksVariable(ContextWrappingVariable):
 class AutocastModeVariable(ContextWrappingVariable):
     @staticmethod
     def create(func, args, kwargs):
-        if func not in [
+        assert func in [
             torch.amp.autocast_mode.autocast,
             torch.cuda.amp.autocast,
             torch.cpu.amp.autocast,
-        ]:
-            unimplemented_v2(
-                gb_type="Unsupported autocast function",
-                context=str(func),
-                explanation="Dynamo doesn't support compiling a region that returns a unsupported autocast function.",
-                hints=[
-                    "Use supported autocast functions: torch.amp.autocast_mode.autocast, torch.cuda.amp.autocast, torch.cpu.amp.autocast",
-                    "Or move the autocast usage outside the compiled region",
-                ],
-            )
+        ]
         # device_type : str,
         # dtype : Optional[_dtype] = None,
         # enabled : bool = True,
--- a/torch/_dynamo/variables/ctx_manager.py
+++ b/torch/_dynamo/variables/ctx_manager.py
@@ -877,11 +877,20 @@ class DisabledSavedTensorsHooksVariable(ContextWrappingVariable):
 class AutocastModeVariable(ContextWrappingVariable):
     @staticmethod
     def create(func, args, kwargs):
-        assert func in [
+        if func not in [
             torch.amp.autocast_mode.autocast,
             torch.cuda.amp.autocast,
             torch.cpu.amp.autocast,
-        ]
+        ]:
+            unimplemented_v2(
+                gb_type="Unsupported autocast function",
+                context=str(func),
+                explanation="Dynamo doesn't support compiling a region that returns a unsupported autocast function.",
+                hints=[
+                    "Use supported autocast functions: torch.amp.autocast_mode.autocast, torch.cuda.amp.autocast, torch.cpu.amp.autocast",
+                    "Or move the autocast usage outside the compiled region",
+                ],
+            )
         # device_type : str,
         # dtype : Optional[_dtype] = None,
         # enabled : bool = True,
--- a/torch/_dynamo/variables/distributed.py
+++ b/torch/_dynamo/variables/distributed.py
@@ -210,9 +210,16 @@ class PlacementVariable(DistributedVariable):
         if name in constant_fold_functions:
             try:
                 value_type = type(self.value)
-                assert (
-                    inspect.getattr_static(value_type, "__getattr__", None) is None
-                ), "no custom getattr allowed!"
+                if inspect.getattr_static(value_type, "__getattr__", None) is not None:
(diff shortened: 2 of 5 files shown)


torch/_dynamo/variables/ctx_manager.py
```diff
@@ -877,11 +877,20 @@ def fn_name(self):
 class AutocastModeVariable(ContextWrappingVariable):
     @staticmethod
     def create(func, args, kwargs):
-        assert func in [
+        if func not in [
```
williamwen42: This assert is fine to leave as-is since we should only be creating this variable if the condition is true from the caller. (code pointer https://github.com/pytorch/pytorch/blob/9901d4441862cc0be881f913e1a7f932f7cb0177/torch/_dynamo/variables/torch.py#L398)

torch/_dynamo/variables/distributed.py
```diff
@@ -213,9 +213,16 @@ def call_method(
         if name in constant_fold_functions:
             try:
                 value_type = type(self.value)
-                assert (
-                    inspect.getattr_static(value_type, "__getattr__", None) is None
-                ), "no custom getattr allowed!"
+                if inspect.getattr_static(value_type, "__getattr__", None) is not None:
+                    unimplemented_v2(
+                        gb_type="custom __getattr__ not supported",
```
williamwen42:
```suggestion
                        gb_type="Placement with custom __getattr__ not supported",
```

torch/_dynamo/variables/distributed.py
```diff
@@ -213,9 +213,16 @@ def call_method(
...
+                        context=f"{value_type.__name__} with custom __getattr__",
+                        explanation="Dynamo does not support Placement types with custom __getattr__ methods",
+                        hints=[
+                            "Use Placement types without custom __getattr__ methods",
+                            "Or move the Placement usage outside the compiled region",
```
williamwen42:
```suggestion
                            "Move the Placement usage outside the compiled region",
```

torch/_dynamo/variables/misc.py
```diff
@@ -103,7 +103,15 @@ def reconstruct(self, codegen: "PyCodegen"):
             codegen.extend_output(create_call_function(1, False))
 
     def _resolved_getattr_and_source(self, tx: "InstructionTranslator", name):
-        assert self.objvar, "1-arg super not implemented"
+        if not self.objvar:
+            unimplemented_v2(
+                gb_type="1-arg super not implemented",
+                context="",
+                explanation="Dynamo failed to trace attribute `{name}` accessed "
```
williamwen42:
```suggestion
                explanation=f"Dynamo failed to trace attribute `{name}` accessed "
```

torch/_dynamo/variables/misc.py
```diff
@@ -103,7 +103,15 @@ def reconstruct(self, codegen: "PyCodegen"):
...
+                f"via `super()` (for type `{self.typevar}` and object `{self.objvar}`) "
+                "because only one argument of super() is not yet supported.",
```
williamwen42:
```suggestion
                "because one-argument of super() is not supported.",
```

torch/_dynamo/variables/misc.py
```diff
@@ -103,7 +103,15 @@ def reconstruct(self, codegen: "PyCodegen"):
...
+                hints=[],
```
williamwen42:
```suggestion
                hints=[
                    "Use two-argument super(type, object_or_type).",
                ],
```

torch/_dynamo/variables/misc.py
```diff
@@ -1033,9 +1041,16 @@ def call_method(
     ) -> "VariableTracker":
         if name == "queue_callback":
             if torch._dynamo.compiled_autograd.in_compiled_autograd_region:
-                assert tx.one_graph or tx.error_on_graph_break, (
-                    "queue_callback() is only supported when Compiled Autograd is enabled with fullgraph=True"
-                )
+                if not (tx.one_graph or tx.error_on_graph_break):
+                    unimplemented_v2(
```
williamwen42: @xmfan is it ok to graph break here? or should we error out?

torch/_dynamo/variables/nn_module.py
```diff
@@ -822,9 +822,23 @@ def gen_source(source, name):
...
+                        gb_type="non-const argument in nn.Module __getitem__",
+                        context=f"call_method: {self} {name} {args} {kwargs}",
+                        explanation="Dynamo does not support calling "
+                        f"method `{name}` of ``nn.Module`` {module} with non-constant arguments.",
+                        hints=["Use constant arguments"],
+                    )
                 key = args[0].as_python_constant()
-                assert isinstance(key, (str, int))
+                if not isinstance(key, (str, int)):
+                    unimplemented_v2(
```
williamwen42, line 829: Can we combine this graph break with the one above?

torch/_dynamo/variables/ctx_manager.py
can-gaa-hou: Thanks for the suggestion :) I have updated it.

torch/_dynamo/variables/nn_module.py
can-gaa-hou: Sure. SGTM. Fix :)

torch/_dynamo/variables/misc.py
can-gaa-hou: Seems like there has been a change [here](https://github.com/pytorch/pytorch/pull/165707) recently. Let's keep that first.

williamwen42 (approved):

Issue #165654, opened 2025-10-16
PyTorch on aarch64 crashes with ConvTranspose1d


malfet:
```
$ lldb -- python -c "import torch;torch.nn.ConvTranspose1d(2016, 1026, 1024, stride=256)(torch.rand(1, 2016, 224))"
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'lldb'
(lldb) target create "python"
Current executable set to 'python' (aarch64).
(lldb) settings set -- target.run-args  "-c" "import torch;torch.nn.ConvTranspose1d(2016, 1026, 1024, stride=256)(torch.rand(1, 2016, 224))"
(lldb) r
Process 1623 launched: '/opt/conda/envs/py_3.10/bin/python' (aarch64)
warning: (aarch64) /opt/conda/envs/py_3.10/lib/python3.10/site-packages/numpy.libs/libgfortran-daac5196.so.5.0.0 No LZMA support found for reading .gnu_debugdata section
Process 1623 stopped
* thread #1, name = 'python', stop reason = signal SIGSEGV: invalid address (fault address: 0xfff88d79d0c0)
    frame #0: 0x0000ffffe9002400 libarm_compute.so`void arm_compute::run_reverse<unsigned int>(arm_compute::Window const&, arm_compute::ITensor const*, arm_compute::ITensor const*, arm_compute::ITensor*, bool)::'lambda'(arm_compute::Coordinates const&)::operator()(arm_compute::Coordinates const&) const + 512
libarm_compute.so`void arm_compute::run_reverse<unsigned int>(arm_compute::Window const&, arm_compute::ITensor const*, arm_compute::ITensor const*, arm_compute::ITensor*, bool)::'lambda'(arm_compute::Coordinates const&)::operator()(arm_compute::Coordinates const&) const:
->  0xffffe9002400 <+512>: str    q0, [x26, w0, sxtw]
    0xffffe9002404 <+516>: ldp    x1, x0, [x19, #0x8]
    0xffffe9002408 <+520>: ldr    w0, [x0]
    0xffffe900240c <+524>: ldr    w1, [x1]
  thread #17, name = 'python', stop reason = signal SIGSEGV: invalid address (fault address: 0xfff88d71f0c0)
[...]

PR #165904, opened 2025-10-20
 [cpu][fix] Update ACL version to fix crashes with tensor sizes > 2^31-1

fadara01:
* #166731
* __->__ #165904

----

- Updates Arm Compute Library (ACL) to v52.6.0
- v52.6.0 contains https://github.com/ARM-software/ComputeLibrary/pull/1201 which fixes crashes with tensors of sizes > 2^31-1

fixes: #165654

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @malfet @snadampal @milpuz01 @aditew01 @nikhil-arm

### DIFF
--- a/.ci/docker/common/install_acl.sh
+++ b/.ci/docker/common/install_acl.sh
@@ -3,7 +3,7 @@
 
 set -eux
 
-ACL_VERSION=${ACL_VERSION:-"v25.02"}
+ACL_VERSION=${ACL_VERSION:-"v52.6.0"}
 ACL_INSTALL_DIR="/acl"
 
 # Clone ACL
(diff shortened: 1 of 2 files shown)


malfet (changes requested):
Hmm, if problem is with tensor sizes, I'm confused why it should be excluded only for transposed tensors. Very likely the same will happens for regular tensor as well, wouldn't it? 

And IMO the right workaround would be to return `false` from `use_mkldnn` if weight or input tensors numel exceed `2**31`

malfet (approved):

PR #155576, opened 2025-06-10
[BE]: Update cudnn to 9.10.2.21

Skylion007:
Update to CUDNN 9.10.2.21

PR #165913, opened 2025-10-20
Update cuDNN 9.10.2 in Manylinux 2.28 Docker files

### LINKED ISSUE #165801: [CUDA][cuDNN][2.9] CUDA 12.8 wheels ship with cuDNN 9.10.2 but CI is installing 9.8.0
### 🐛 Describe the bug

@drisspg noticed that cuDNN wasn't being selected as the default backend for SDPA where expected.
This seem to be because the wheel build checks cuDNN version at compile time (9.8) which is too old for the minimum on by default setting (9.9).

CI install:
https://github.com/pytorch/pytorch/blob/e50dc40d28ba409930023c77a031ec0dd20fd73b/.ci/docker/common/install_cuda.sh#L132

Binary build matrix:
https://github.com/pytorch/pytorch/blob/9315f44cd64d6e25abc30e8db35fb0e4d6141eb6/.github/scripts/generate_binary_build_matrix.py#L67

Meanwhile the wheels are shipping with cuDNN 9.10.2 which should support cuDNN-on-by-default for SDPA.

### Versions

`release/2.9`

cc @ezyang @gchanan @zou3519 @kadeng @msaroufim @seemethere @malfet @atalman @csarofeen @ptrblck @xwang233 @jerryzh168

atalman:
Fixes https://github.com/pytorch/pytorch/issues/165801
Smoke test: https://github.com/pytorch/pytorch/pull/165922/files

### DIFF
--- a/.ci/docker/common/install_cuda.sh
+++ b/.ci/docker/common/install_cuda.sh
@@ -129,7 +129,7 @@ function install_129 {
 }
 
 function install_128 {
-  CUDNN_VERSION=9.8.0.87
+  CUDNN_VERSION=9.10.2.21
   echo "Installing CUDA 12.8.1 and cuDNN ${CUDNN_VERSION} and NVSHMEM and NCCL and cuSparseLt-0.7.1"
   # install CUDA 12.8.1 in the same container
   install_cuda 12.8.1 cuda_12.8.1_570.124.06_linux


Camyll (approved):

malfet (commented):
How about test plan?

malfet (changes requested):
Since there are no urgency to land this one (2.9 release is out), let's incorporate some sort of BE mechanism to avoid making those changes in 25 different places as https://github.com/pytorch/pytorch/pull/155576 did and miss 26th one

.ci/docker/common/install_cuda.sh
```diff
@@ -129,7 +129,7 @@ function install_129 {
 }
 
 function install_128 {
-  CUDNN_VERSION=9.8.0.87
+  CUDNN_VERSION=9.10.2.21
```
Skylion007, line 132: install_124 needs to be updated as well.

Skylion007 (approved):

.ci/docker/common/install_cuda.sh
atalman: HI @Skylion007 install_124 is only used for CI CUDA 12.4 jobs https://hud.pytorch.org/hud/pytorch/pytorch/main/1?per_page=50&name_filter=12.4
I believe at this point we probably can leave them as is

clee2000:
@pytorchbot revert -m "I think something here is causing CI tests to segfault at exit on cuda, ex [GH job link](https://github.com/pytorch/pytorch/actions/runs/18857880394/job/53811917713) [HUD commit link](https://hud.pytorch.org/pytorch/pytorch/commit/9a91486e453f300804c7031d0e904b0e2af7d802) says no tests failed but it segfaulted afterwards.  I can't tell if it's because of this change, or an unpinned dependency in docker that got triggered by this.  Note to self, would have been bad TD except trunk didn't run either" -c nosignal

PR #165914, opened 2025-10-20
[aoti][win] add support for a list of shim libraries

yushangdi:
As title, support passing in a list of shim libraries when cross compiling artifacts

### DIFF
--- a/torch/_inductor/config.py
+++ b/torch/_inductor/config.py
@@ -1670,7 +1670,7 @@ class aot_inductor:
 
     # If link_libtorch is False and cross_target_platform is windows,
     # a library needs to be provided to provide the shim implementations.
-    aoti_shim_library: Optional[str] = None
+    aoti_shim_library: Optional[str | list[str]] = None
     aoti_shim_library_path: Optional[str] = None
 
 
--- a/torch/_inductor/cpp_builder.py
+++ b/torch/_inductor/cpp_builder.py
@@ -1147,10 +1147,16 @@ def _get_torch_related_args(
     else:
         libraries_dirs = []
         if config.aot_inductor.cross_target_platform == "windows":
-            assert config.aot_inductor.aoti_shim_library, (
+            aoti_shim_library = config.aot_inductor.aoti_shim_library
+
+            assert aoti_shim_library, (
                 "'config.aot_inductor.aoti_shim_library' must be set when 'cross_target_platform' is 'windows'."
             )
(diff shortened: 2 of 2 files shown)


torch/_inductor/config.py
```diff
@@ -1670,7 +1670,7 @@ class aot_inductor:
 
     # If link_libtorch is False and cross_target_platform is windows,
     # a library needs to be provided to provide the shim implementations.
-    aoti_shim_library: Optional[str] = None
+    aoti_shim_library: None | str | list[str] = None
```
JacobSzwejbka: TIL this typing syntax

torch/_inductor/cpp_builder.py
```diff
@@ -1147,10 +1147,16 @@ def _get_torch_related_args(
...
+            aoti_shim_library = config.aot_inductor.aoti_shim_library
+
+            assert aoti_shim_library, (
                 "'config.aot_inductor.aoti_shim_library' must be set when 'cross_target_platform' is 'windows'."
             )
-            libraries.append(config.aot_inductor.aoti_shim_library)
+            if isinstance(aoti_shim_library, str):
+                libraries.append(aoti_shim_library)
+            else:
+                assert isinstance(aoti_shim_library, list)
```
JacobSzwejbka, line 1158: I still dont 100% understand why aoti_shim_library is different from aoti_shim_library_path.

yushangdi: when `link_libtorch=True`, no aoti_shim_library needs to be provided (it automatically uses the same libtorch libraries as non-cross compiling case), but aoti_shim_library_path needs to be provided.

desertfire, line 1158: Is it possible `aoti_shim_library` is None for this else branch?

desertfire (commented):
Is it possible to add a test for this one?

torch/_inductor/cpp_builder.py
yushangdi: not because we have `assert aoti_shim_library` above.

yushangdi:
> Is it possible to add a test for this one?

@desertfire I don't have a good way to test it right now. if we don't want to land this without a test, we can wait until ET needs to use it, and then land it with test in ET. ET doesn't need it right now as they produce a single .dll/.lib , but it might be needed in the future.

desertfire (approved):

PR #165922, opened 2025-10-20
[cuDNN] Smoke-test runtime cuDNN version matches compile time version in CI

eqy:
Fix and regression test for https://github.com/pytorch/pytorch/issues/165801

cc @seemethere @malfet @atalman @tinglvv @nWEIdia @csarofeen @ptrblck @xwang233

### DIFF
--- a/.ci/pytorch/smoke_test/smoke_test.py
+++ b/.ci/pytorch/smoke_test/smoke_test.py
@@ -284,13 +284,14 @@ def smoke_test_cuda(
         torch_cudnn_compile_version = torch._C._cudnn.getCompileVersion()
         print(f"Torch cuDNN compile-time version: {torch_cudnn_compile_version}")
         torch_cudnn_runtime_version = tuple(
-            [int(x) for x in torch_cudnn_version.split('.')]
+            [int(x) for x in torch_cudnn_version.split(".")]
         )
         if torch_cudnn_runtime_version != torch_cudnn_compile_version:
             raise RuntimeError(
                 "cuDNN runtime version doesn't match comple version. "
                 f"Loaded: {torch_cudnn_runtime_version} "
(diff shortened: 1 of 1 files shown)


malfet (approved):

atalman (approved):
lgtm, lets confirm with ciflow/binaries check

.ci/pytorch/smoke_test/smoke_test.py
```diff
@@ -272,6 +272,15 @@ def smoke_test_cuda(
         torch_cudnn_version = cudnn_to_version_str(torch.backends.cudnn.version())
         print(f"Torch cuDNN version: {torch_cudnn_version}")
 
+        torch_cudnn_compile_version = torch._C._cudnn.getCompileVersion()
+        print(f"Torch cuDNN compile-time version: {torch_cudnn_compile_version}")
+        torch_cudnn_runtime_version = tuple([int(x) for x in torch_cudnn_version.split('.')])
```
Skylion007:
```suggestion
        torch_cudnn_runtime_version = tuple(map(int, torch_cudnn_version.split('.')))

Skylion007 (approved):

.ci/pytorch/smoke_test/smoke_test.py
malfet: Not sure if it's faster or more readable though

drisspg (approved):
Noob q; so we build pytorch based off of cudnn frontend X with cudnn bakckend installed on build machine Y

We then ship PyTorch with cudnn backend wheel installed version Z

In theory as long as Y and Z are they same major version as X's frontend all is good.

Should we update the dispatching for cudnn SDPA to be a runtime check against the version of Z, and not a static build time check against Y? 

I know in general we should build against what we ship but trying to understand why this didnt cause other errors

PR #165936, opened 2025-10-20
Use `is` rather than `==` to work around slow enum comparion in _ops.py

swolchok:
* __->__ #165936

This shows up (under _are_we_tracing) in DTensor dispatch. I have some work in flight to speed up enum comparison in pybind11, but `is` is just much faster and easy to use.


Skylion007 (approved):

swolchok:
lintrunner-mypy appears to have a known failure on HUD, bypassing

torch/_ops.py
```diff
@@ -628,11 +628,13 @@ def unset_mode_pre_dispatch(mode_key, schema_check=False):
         assert mode_key is None
 
     def _unset_mode():
-        if mode_key == torch._C._TorchDispatchModeKey.PROXY:
+        # NOTE: Using `is` rather than `==` to work around slow enum comparison in
```
rec, line 631: Maybe this isn't a good place for this comment, particularly since you have to repeat it below!

Perhaps it would be better in the documentation of `torch._C._TorchDispatchModeKey` where others would see it?

zou3519 (approved):

torch/_ops.py
swolchok: > Maybe this isn't a good place for this comment, particularly since you have to repeat it below!
> 
> Perhaps it would be better in the documentation of `torch._C._TorchDispatchModeKey` where others would see it?

it applies to all pybinded enums, so the documentation of an individual enum is also not a good place. What I really think we should do about it is 1) take the approach in this PR (use `is`) to unblock for perf-sensitive paths, 2) patch pybind11 to reduce the cost of enum operations, and also 3) start requiring version 3 of pybind11 and use `native_enum` rather than `enum_` for new enums.

PR #165937, opened 2025-10-20
Support multiple commits on push events in trunk tagging workflow

izaitsevfb:
Context:
* this workflow is used to create tags like `trunk/{sha}` for all `main` commits
* those tags are used by [autorevert](https://github.com/pytorch/test-infra/blob/main/aws/lambda/pytorch-auto-revert/README.md) to rerun selected workflows

Problem: currently the workflow creates only a single tag per push event, while ghstack pushes multiple commits per single push.

This PR supports tag creation for all commits in the push event.

Complimentary autorevert PR: https://github.com/pytorch/test-infra/pull/7291

---

### Testing

I created an identical copy of this workflow in my personal repo: https://github.com/izaitsevfb/pr-head-test/actions/workflows/trunk-tagging.yml

See action runs there.

### DIFF
--- a/.github/workflows/trunk-tagging.yml
+++ b/.github/workflows/trunk-tagging.yml
@@ -114,6 +114,24 @@ jobs:
             return 1
           }
 
+          # Counters for summary reporting
+          created_count=0
+          skipped_count=0
+          failed_count=0
+
+          # Always write outputs once on exit
+          finish() {
+            set +e
+            if [ -n "${GITHUB_OUTPUT:-}" ]; then
+              {
+                echo "created_count=${created_count}"
+                echo "skipped_count=${skipped_count}"
+                echo "failed_count=${failed_count}"
+              } >> "${GITHUB_OUTPUT}"
+            fi
+          }
+          trap finish EXIT
+
           # Retry configuration
           MAX_RETRIES=5
           BASE_DELAY=2
@@ -190,11 +208,6 @@ jobs:
           # New behavior for push events: enumerate commits in the push and tag each one.
           # For workflow_dispatch, retain existing single-SHA behavior.
 
-          # Counters for summary reporting
-          created_count=0
-          skipped_count=0
-          failed_count=0
-
           # Always fetch tags once up front to improve idempotency in loops
           git fetch origin --tags --quiet || true
 
@@ -208,12 +221,6 @@ jobs:
 
             if [ ! -s "${commits_file}" ]; then
               echo "No new commits found between ${BEFORE_SHA}..${AFTER_SHA}; nothing to tag."
-              {
-                echo "exists=true"
-                echo "created_count=0"
-                echo "skipped_count=0"
-                echo "failed_count=0"
-              } >> "${GITHUB_OUTPUT}"
               rm -f "${commits_file}"
               exit 0
             fi
@@ -247,21 +254,6 @@ jobs:
 
             rm -f "${commits_file}"
 
-            # Determine aggregate existence flag: true iff nothing was created and nothing failed
-            if [ "${created_count}" -eq 0 ] && [ "${failed_count}" -eq 0 ]; then
-              exists_flag=true
-            else
-              exists_flag=false
-            fi
-
-            # Expose summary outputs for downstream step in a single redirection
-            {
-              echo "exists=${exists_flag}"
-              echo "created_count=${created_count}"
-              echo "skipped_count=${skipped_count}"
-              echo "failed_count=${failed_count}"
-            } >> "${GITHUB_OUTPUT}"
-
             if [ "${failed_count}" -gt 0 ]; then
               exit 1
             fi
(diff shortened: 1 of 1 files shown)


.github/workflows/trunk-tagging.yml
```diff
@@ -194,31 +187,143 @@ jobs:
...
+
+          # Always fetch tags once up front to improve idempotency in loops
+          git fetch origin --tags --quiet || true
+
+          if [ "${{ github.event_name }}" = "push" ]; then
+            BEFORE_SHA="${{ github.event.before }}"
+            AFTER_SHA="${{ github.sha }}"  # same as event.after
+
+            # List commits introduced by this push (old..new), oldest first for stable ordering
+            mapfile -t COMMITS < <(git rev-list --reverse "${BEFORE_SHA}..${AFTER_SHA}")
```
malfet: `mapfile` is a bash extension, isn't it? Is it guaranteed anywhere that default shell for github action is bash?

.github/workflows/trunk-tagging.yml
```diff
@@ -194,31 +187,143 @@ jobs:
...
+
+            if [ ${#COMMITS[@]} -eq 0 ]; then
+              echo "No new commits found between ${BEFORE_SHA}..${AFTER_SHA}; nothing to tag."
+              {
+                echo "exists=true"
```
malfet: Where this one is used?

.github/workflows/trunk-tagging.yml
```diff
@@ -194,31 +187,143 @@ jobs:
...
+                failed_count=$((failed_count + 1))
+              fi
+            done
+
+            # Determine aggregate existence flag: true iff nothing was created and nothing failed
+            if [ "${created_count}" -eq 0 ] && [ "${failed_count}" -eq 0 ]; then
+              exists_flag=true
+            else
+              exists_flag=false
+            fi
```
malfet: Was this written by LLM? I'm bad at shell scripting, but in C/C++/Python `if (cond) var = true; else var = false;` are usually replaced with `var = cond;`
```suggestion
            exists_flag = $([ "${created_count}" -eq 0 ] && [ "${failed_count}" -eq 0 ]) 
```

.github/workflows/trunk-tagging.yml
izaitsevfb: it works currently, but I'll see if I can replace it to minimize dependencies

.github/workflows/trunk-tagging.yml
```diff
@@ -194,31 +187,143 @@ jobs:
...
+
+            # Exit early if tag already exists
+            if check_tag_exists; then
+              echo "✅ Tag already exists - no action needed"
+              {
+                echo "exists=true"
+                echo "created_count=0"
+                echo "skipped_count=1"
+                echo "failed_count=0"
+              } >> "${GITHUB_OUTPUT}"
```
malfet: This repeated pattern feels like it could benefit from [trap](https://man7.org/linux/man-pages/man1/trap.1p.html)

.github/workflows/trunk-tagging.yml
izaitsevfb: this won't work unfortunately.

`$([ ... ])` captures stdout, but [ ... ] prints nothing

in bash it's already more or less optimal as currently written. 

this one is slightly shorter, but I'm not sure what we're trying to optimize here:
```
exists_flag=false; [ "$created_count" -eq 0 ] && [ "$failed_count" -eq 0 ] && exists_flag=true
```

izaitsevfb: and yes to your question, it was written by LLM

.github/workflows/trunk-tagging.yml
izaitsevfb: good idea, will do

.github/workflows/trunk-tagging.yml
izaitsevfb: here:
https://github.com/pytorch/pytorch/pull/165937/files#diff-abe52f8587c6584de77a02a9cfd50e786ac967740f57c0e695ae5f5a0e0e16d7R320-R321

```
 if [ "${{ steps.check_tag.outputs.exists }}" = "true" ]; then
              echo "✅ Tag ${{ steps.commit.outputs.tag_name }} already existed - no action needed"
```

but it could be simplified a bit

.github/workflows/trunk-tagging.yml
izaitsevfb: replaced with a temp file

.github/workflows/trunk-tagging.yml
izaitsevfb: removed this flag

huydhn (approved):
LGTM!

.github/workflows/trunk-tagging.yml
```diff
@@ -194,31 +205,111 @@ jobs:
...
-          if retry_with_backoff "tag_with_retry" "Creating tag ${TAG_NAME} for commit ${COMMIT_SHA}"; then
-            echo "exists=false" >> "${GITHUB_OUTPUT}"
+          # New behavior for push events: enumerate commits in the push and tag each one.
+          # For workflow_dispatch, retain existing single-SHA behavior.
+
+          # Always fetch tags once up front to improve idempotency in loops
+          git fetch origin --tags --quiet || true
+
+          if [ "${{ github.event_name }}" = "push" ]; then
+            BEFORE_SHA="${{ github.event.before }}"
```
huydhn, line 215: I double check and confirm that `github.event.before` is only available for `push` event.  However, I think you could default `BEFORE_SHA` to `${{ github.sha }}^1` in that case.  The benefit would be that you don't need to keep the `else` branch for `workflow_dispatch` anymore, making the code a bit shorter.  It will be a commits file of just one commit

.github/workflows/trunk-tagging.yml
```diff
@@ -194,31 +205,111 @@ jobs:
...
+            AFTER_SHA="${{ github.sha }}"  # same as event.after
+
+            # List commits introduced by this push (old..new), oldest first for stable ordering
+            commits_file="$(mktemp)"
+            git rev-list --reverse "${BEFORE_SHA}..${AFTER_SHA}" > "${commits_file}"
+
+            if [ ! -s "${commits_file}" ]; then
+              echo "No new commits found between ${BEFORE_SHA}..${AFTER_SHA}; nothing to tag."
+              rm -f "${commits_file}"
```
huydhn, line 224: Nit: It's also nicer to have this clean up step in the exit trap

.github/workflows/trunk-tagging.yml
izaitsevfb: thought about this (who am i kidding, discussed it with llm), and this might not be safe in case we ever encounter a merge commit (doesn't apply currently, but just in case let's play it safe)

Issue #119930, opened 2024-02-14
[dtensor] ops coverage tracker


wanchaol:
This issue is the overall tracker of the ops we are implementing currently and plan to implement soon, when we plan to add new ops we'll also try to update this issue to keep things in sync. 

### Current supported ops consolidation
replace the existing ops that implemented in rule with op strategy to increase op sharding coverage
- [ ] aten.slice_backward.default, 
- [ ] aten.bernoulli.default/aten.bernoulli_.float
- [ ] aten.split.Tensor, aten.split_with_sizes.default, aten.split_with_sizes_copy.default
- [ ] aten.cat.default
- [ ] aten.index.Tensor
- [ ] aten.index_select.default
- [ ] aten.squeeze.default, aten.squeeze.dim
- [ ] aten.view.default, aten._unsafe_view.default, aten.reshape.default
- [ ] aten.unsqueeze.default
- [ ] aten.expand.default
- [ ] aten.permute.default
- [ ] aten.transpose.int, aten.t.default
- [ ] aten.repeat.default
- [ ] aten.convolution.default
- [ ] aten.convolution_backward.default

### Ops to be implemented (planned)
- [ ] aten._scaled_dot_product_flash_attention.default
- [ ] aten._scaled_dot_product_flash_attention_backward.default
- [ ] complex ops

Vandermode:
Hi @wanchaol , thanks for the great work!

I was wondering if that is possible to add user-customized partition rules to certain operators e.g., `torch.fft.fft `
Specifically, can we extend FFT to dtensor op without diving into the cpp implementation? 
Note such implementation is provided in jax (https://jax.readthedocs.io/en/latest/jax.experimental.custom_partitioning.html) to make it work in shared arrays.
 
Thank you for your patience.

awgu:
https://github.com/pytorch/pytorch/issues/119930
https://github.com/pytorch/pytorch/issues/132016
https://github.com/pytorch/pytorch/issues/132017

tianyu-l:
Hi @Vandermode , we've added the support for defining op partition/sharding rules in DTensor (in https://github.com/pytorch/pytorch/pull/131108). Please feel free to try out the API and let us know any feedbacks you may have.

XilunWu:
https://github.com/pytorch/pytorch/pull/149106 is another tracker for DTensor op coverage for some important models.

stmcgovern:
[...]

PR #165940, opened 2025-10-20
[DTensor] dist() sharding strategy

stmcgovern:
Contributes to #119930

 ### Summary

  Adds DTensor sharding strategy for torch.dist() operation, enabling distributed p-norm distance computation.

   ### Motivation

  Previously, torch.dist() failed with DTensor inputs, forcing inefficient to_local() conversion. This was evident in test_linalg_eigh
  where eigenvalue verification required converting distributed tensors to local.

  ###  Implementation

  - Aligns input tensors (prefers more sharded for parallelism)
  - Forces Partial→Replicate when inputs have Partial placement (prevents correctness bugs)
  - Uses NormReduction(p) → returns _NormPartial for lazy communication
  - Follows softmax_backward_strategy pattern for two-tensor alignment

   ### Testing

  - Tests p-norms (1, 2, inf) with Shard/Replicate placements
  - Includes Partial input edge case (critical for correctness)
  - Updated test_linalg_eigh to use dist directly on DTensors
  - Removed from xfail list

  Resolves TODO at test/distributed/tensor/test_math_ops.py:740

### DIFF
--- a/torch/distributed/tensor/_ops/_math_ops.py
+++ b/torch/distributed/tensor/_ops/_math_ops.py
@@ -27,6 +27,7 @@ from torch.distributed.tensor._ops.utils import (
     normalize_dims,
     register_op_strategy,
 )
+from torch.distributed.tensor._ops._pointwise_ops import linear_pointwise_strategy
 from torch.distributed.tensor._utils import normalize_to_torch_size
 from torch.distributed.tensor.placement_types import (
     _StridedShard,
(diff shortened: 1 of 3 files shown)


stmcgovern:
cc @zpcore

torch/distributed/tensor/_ops/_math_ops.py
```diff
@@ -450,6 +451,70 @@ def vector_norm_strategy(op_schema: OpSchema) -> OpStrategy:
...
+    sub_schema = OpSchema(
+        op=aten.sub.Tensor,
+        args_schema=(input_strategy, other_strategy),
+        kwargs_schema={},
+    )
+    aligned_strategy = cast(OpStrategy, linear_pointwise_strategy(sub_schema))
+
+    # Step 2: Apply reduction using existing placement logic
+    # NOTE: We can't delegate fully to common_reduction_strategy because:
+    # - linear_pointwise_strategy doesn't set tensor_meta on output specs
```
wconstab: i'm thinking about whether we should fix this issue directly - setting tensor_meta in linear_pointwise strat.  Otoh, I am hopefully going to be able to replace pointwise with a single_dim rule in the near future, so I am not so concerned about nonessential fixes to linear_pointwise.

torch/distributed/tensor/_ops/_math_ops.py
```diff
@@ -450,6 +451,70 @@ def vector_norm_strategy(op_schema: OpSchema) -> OpStrategy:
...
+    # - common_reduction_strategy needs tensor_meta to access .ndim
+    # - So we compute dimensions manually and reuse map_placements_after_reduction
+    common_shape = torch.broadcast_shapes(input_strategy.shape, other_strategy.shape)
+    result_ndim = len(common_shape)
+
+    reduce_dims = list(range(result_ndim))
+    reduce_dims_map = _infer_reduce_dims_map(reduce_dims, result_ndim, keep_dim=False)
```
wconstab: i'm confused about this logic. reduce_dims is range (0,...n) and result_ndim is n -- right?  If this is the case, doesn't reduce_dims_maps always come out as range(n) and offer a trivial mapping?  or what am I missing

stmcgovern: with keep_dim=False it will always come out as [-1, -1,...]. keep_dim=True gets range(n). But your point that we already know the answer is definitely right. For the full reduction, any shard goes to partial. This can be simplified without using that more general function.

stmcgovern: shard goes to normPartial and when handling partial inputs (not done before), goes to replicate

torch/distributed/tensor/_ops/_math_ops.py
```diff
@@ -476,6 +476,87 @@ def vector_norm_strategy(op_schema: OpSchema) -> OpStrategy:
     )
 
 
+@register_op_strategy([aten.dist.default], schema_info=RuntimeSchemaInfo(2))
+def dist_strategy(op_schema: OpSchema) -> OpStrategy:
```
wconstab, line 480: i'm hoping to make a simpler implementation for this.  One thing I wonder is, could we use the decomp based approach from @pianpwk (https://github.com/pytorch/pytorch/pull/171652)?  If dist() already has a decomp, we should just use that.  If it doesn't, it sounds from your description like it might be easy to write a decomp combining sub and vector_norm.  In either case, we could then avoid writing a rule specifically for dist().

pianpwk: we do add support for dist in https://github.com/pytorch/pytorch/pull/171652, though it's unclear to me if there's any difference in sharding rules

PR #165943, opened 2025-10-20
A fix to incorrect size/stride hints for unbacked symbols

guangy10:
Summary:
Autotune issue during lowering w/ AOTI:
```
setStorage: sizes [1536, 32, 8192], strides [8192, 8192, 1], storage offset 0, and itemsize 2 requiring a storage size of 25673728 are out of bounds for storage of size 25362432
```
Need a hack to create new base tensor with sufficient storage

Test Plan: Finally be able to see the e2e test passes on CI. See the detailed Test Plan in D83520844

Differential Revision: D84872792

### DIFF
--- a/torch/_inductor/select_algorithm.py
+++ b/torch/_inductor/select_algorithm.py
@@ -3147,35 +3147,82 @@ class AlgorithmSelectorCache(PersistentCache):
             for i, x in enumerate(input_nodes)
         }
         example_inputs = list(unique_example_inputs.values())
-        example_inputs_extern = [
-            (
-                unique_example_inputs[input_node.get_name()]
-                if unique_example_inputs[input_node.get_name()].is_mkldnn
-                else torch.as_strided(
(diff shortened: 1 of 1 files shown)


torch/_inductor/select_algorithm.py
```diff
@@ -3147,35 +3147,82 @@ def get_inputs(
...
+                    torch.as_strided(base, sizes, strides, storage_offset)
                 )
-            )
-            for input_node in input_nodes
-        ]
         out = cls.benchmark_example_value(layout, hint_override=hint_override)
-        out_extern = torch.as_strided(
-            out, out.size(), out.stride(), V.graph.sizevars.size_hint(layout.offset)
+
+        # Also check the output tensor for storage size
```
laithsakka, line 3205: do we need this part? did it fail with out it ?

guangy10: All CI passed

laithsakka (approved):

PR #165945, opened 2025-10-20
[DO NOT REVIEW] fx pass for cudagraph partition

BoyuanFeng:
This PR implements cudagraph partition as a fx pass.


eellison (commented):
[Ignoring DO NOT REVIEW]

have you seen this existing pass https://github.com/pytorch/pytorch/blob/4c963a68d784b6c4bc8724ffca5cb9f3fbf75ca9/torch/fx/passes/backends/cudagraphs.py#L45-L61?

torch/fx/passes/backends/cudagraph_partition.py
```diff
@@ -0,0 +1,232 @@
...
+    split_gm, split_items = split_graph(gm, resolved_split_ops)
+    submod_names_to_cudagraph = [
+        item.submod_name for item in split_items if not item.is_cudagraph_unsafe
+    ]
+
+    # split_gm.input_buffers refers to the tensors in the list, instead of the list itself
+    split_gm.input_buffers = [t for t in example_inputs]
+
+    # 2. Wrap submodules with CUDAGraphWrapper
+    PiecewiseCompileInterpreter(split_gm, submod_names_to_cudagraph).run(
```
SherlockNoMad: when do we warm up cuda graph?

torch/fx/passes/backends/cudagraph_partition.py
```diff
@@ -0,0 +1,232 @@
...
+        *example_inputs
+    )
+
+    # 3. Copy inputs to static tensors
```
SherlockNoMad: what about output? 
do you need to copy out the output?

torch/fx/passes/backends/cudagraph_partition.py
SherlockNoMad: I guess the questions is, this is JIT-style cudagraph, I imagine it also could be a AOT cudagraph, where you warm up with example inputs.

torch/fx/passes/backends/cudagraph_partition.py
BoyuanFeng: no need to copy out the output

torch/fx/passes/cudagraph_partition.py
```diff
@@ -0,0 +1,232 @@
...
+    for node in gm.graph.nodes:
+        if node.op in ("output", "placeholder"):
+            continue
+        # Match node.target against resolved_ops
+        # node.target can be OpOverloadPacket, need to check .default
+        if node.op == "call_function" and (
+            node.target in resolved_ops
+            or (hasattr(node.target, "default") and node.target.default in resolved_ops)
+        ):
+            subgraph_id += 1
```
SherlockNoMad, line 74: if you have a few resolved_ops in a sequence, subgraph_id will keep increasing. 
but it's not indicating a new cuda graph region

BoyuanFeng: yes this is fine. we record this cudagraph_unsafe op in `node_to_subgraph_id[node] = subgraph_id` and skip cudagraph for it later by checking `(graph_id in split_op_graphs)`.

PR #165952, opened 2025-10-21
CustomOp Inline Fusion

tianrengao:
Add Inline Fusion Support for Custom Op Autotuning
--------------------------------------------------

This PR extends PyTorch Inductor's custom op autotuning with inline fusion capabilities, enabling the winning decomposition to be inlined directly into the computation graph for fusion with surrounding operations.

### Usage

```python

def decompose_k_implementation(
    a: torch.Tensor, b: torch.Tensor, k_splits: int = 4
) -> torch.Tensor:
    """Matrix multiply with k-way decomposition."""
    ...

@torch.library.custom_op("my_lib::matmul_relu", mutates_args={})
def custom_matmul_relu_dk(
    a: torch.Tensor, b: torch.Tensor, k_splits: int 
) -> torch.Tensor:
    return torch.relu(decompose_k_implementation(a, b, k_splits))

register_custom_op_autotuning(
    custom_op=custom_matmul_relu_dk,
    configs=[
        CustomOpConfig(k_splits=2),
        CustomOpConfig(k_splits=4),
        CustomOpConfig(k_splits=8),
        CustomOpConfig(k_splits=32),
        CustomOpConfig(k_splits=64),
    ],
    name="decompose_k_autotuned",
    input_gen_fns={
        "a": lambda fake: torch.randn_like(fake, device='cuda'),
        "b": lambda fake: torch.randn_like(fake, device='cuda'),
    }
)
```

### How It Works
Enable optimizations from Inductor by inlining the best decomposition, allowing fusion with surrounding elementwise operations and other graph-level optimizations. This provide potentially better performance and memory efficiency.
During customop autotuning phase, we still benchmarks all CustomOpConfigs to find the fastest implementation. Then during inline fusion, inductor inline the decompositions into the main graph, converting the winning choice to individual ComputedBuffer IR nodes (fusable). At the end, Inductor automatically fuses inlined operations with surrounding elementwise ops (e.g., bias add, ReLU, scaling). Note that the winning choice must be a SubgraphChoiceCaller (decomposition-based) rather than an ExternKernelChoice for inlining to work. If the ExternKernelChoice is returned, no inline happens.

Performance Results
Benchmarked on matmul+relu workload with decompose-k fusion (H100 GPU, 15 test shapes):
Screenshot 2025-11-04 at 12 43 11 AM

Metric | Result
-- | --
Average Speedup vs ATen | 1.28x
Max Speedup vs ATen | 1.41x

<br class="Apple-interchange-newline">

The performance comparison are detailed in the below plots. We spot that on most use cases, the inline fusion gains better performance compared to aten baseline and the current torch.compile. 
image

**Test**: `test_decompose_k_with_fusion` demonstrates decompose-k with inline fusion enabled.

--------------

### Integration to mm.py decomposeK with a flag enable_inline_subgraph_fusion=True in config (deprecated to avoid breaking async compilation. removed from the PR already)
FP32:
Screenshot 2025-11-04 at 12 05 08 AM
FP16:
Screenshot 2025-11-04 at 12 13 49 AM

The TCF column represents torch compile fusion, which is close to custom_op decomposek. The difference might due to different candidate k values.

#### Usage:
Note: this only happens when we don't benchmark_epilogue_fusion, i.e., not using multi_template_buffer.

```python
# Define the matmul+relu function
    def matmul_relu(x, y):
        return torch.nn.functional.relu(torch.matmul(x, y))

    # Compile with inline subgraph fusion enabled
    @torch.compile
    def compiled_matmul_relu(x, y):
        return matmul_relu(x, y)

    # Reset dynamo to ensure clean compilation
    torch._dynamo.reset()

    with config.patch(
        {
            "max_autotune": True,
            # CRITICAL: These two flags enable inline subgraph fusion
            "benchmark_epilogue_fusion": False,  # Must be False for inline fusion!
            "enable_inline_subgraph_fusion": True,  # Enable inline fusion
        }
    ):
        # Compile and run
        result = compiled_matmul_relu(a, b)
        torch.cuda.synchronize()
```

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @voznesenskym @penguinwu @EikanWang @Guobing-Chen @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela

### DIFF
--- a/torch/_inductor/codegen/subgraph.py
+++ b/torch/_inductor/codegen/subgraph.py
@@ -24,6 +24,22 @@ from torch._inductor.virtualized import V
 log = logging.getLogger(__name__)
 
 
+def inline_subgraph_to_ir_nodes(
+    gm: torch.fx.GraphModule, inputs: list[Any], name: str
+) -> Any:
+    """Inline a subgraph by converting its FX operations to individual IR nodes.
+
+    This converts a subgraph to multiple ComputedBuffer nodes (fusable),
+    enabling epilogue fusion with subsequent operations.
+
+    Returns:
+        TensorBox containing the final operation result as individual IR nodes
+    """
+    from torch._inductor.lowering import process_subgraph_nodes
+
+    return process_subgraph_nodes(gm, inputs)
+
+
 class SubgraphChoiceCaller(ir.ChoiceCaller):
     """
     Represents a Subgraph Autotuning choice, and the subgraph can be any arbitrary
@@ -261,7 +277,14 @@ class SubgraphTemplate(KernelTemplate):
                 # decomp_kwargs contains all merged parameters: CustomOpConfig params + runtime kwargs
                 from torch.fx.experimental.proxy_tensor import make_fx
 
-                return make_fx(functools.partial(decomp, **decomp_kwargs))(*args)
+                from ..decomposition import select_decomp_table
+
+                decomposition_table = select_decomp_table()
+
+                return make_fx(
+                    functools.partial(decomp, **decomp_kwargs),
+                    decomposition_table=decomposition_table,
+                )(*args)
 
             # Generate descriptive name for this variant
             variant_name = self._generate_variant_name(decomp, decomp_kwargs)
--- a/torch/_inductor/kernel/custom_op.py
+++ b/torch/_inductor/kernel/custom_op.py
@@ -5,6 +5,7 @@ import logging
 from typing import Any, Callable, Optional, Union
 
 import torch
+from torch._inductor import config
 from torch._inductor.codegen.subgraph import SubgraphTemplate
 from torch._inductor.ir import Buffer, FixedLayout, ir_node_to_tensor, TensorBox
 from torch._inductor.lowering import lowerings, validate_ir
@@ -157,7 +158,6 @@ def _adapt_user_input_gen_fns(
 
     Uses V.graph.sizevars.size_hints() to guess best for dynamic shapes.
     """
-    from torch._inductor import config
 
     name_to_index = {name: i for i, name in enumerate(arg_names)}
     index_based_fns = {}
@@ -237,6 +237,7 @@ def autotune_custom_op(
 
     This function generates multiple implementation choices for a custom operation and
     uses Inductor's autotuning system to select the best performing variant at runtime.
+    After selecting the best choice, applies inline fusion if the winning choice has a graph.
 
     Args:
         name: Unique identifier for the autotuning operation
@@ -319,13 +320,33 @@ def autotune_custom_op(
         )
         input_gen_fns = _adapt_user_input_gen_fns(inputs, arg_names, user_input_gen_fns)
 
-    return autotune_select_algorithm(
+    # Run autotuning and get both result and winning choice
+    selected_result, winning_choice = autotune_select_algorithm(
         name=name,
         choices=choices,
         input_nodes=list(inputs),
         layout=choices[0].layout,
         input_gen_fns=input_gen_fns,
+        return_choice=True,
     )
+
+    # Apply inlining for fusion if winning_choice has graph; otherwise return result as-is(default fallback impl)
+    if winning_choice.gm is not None:
+        log.debug(
+            "Inlining winning choice: %s (name=%s)",
+            getattr(winning_choice, "name", type(winning_choice).__name__),
+            name,
+        )
+        from torch._inductor.codegen.subgraph import inline_subgraph_to_ir_nodes
+
+        return inline_subgraph_to_ir_nodes(winning_choice.gm, inputs, name)
+
+    log.debug(
+        "Winning choice does not support inlining: %s (name=%s)",
+        getattr(winning_choice, "name", type(winning_choice).__name__),
+        name,
+    )
+    return selected_result
 
 
 def register_custom_op_autotuning(
@@ -359,7 +380,7 @@ def register_custom_op_autotuning(
                 "query": lambda fake: torch.randn_like(fake, device='cuda'),
                 "key": lambda fake: torch.randn_like(fake, device='cuda'),
                 "value": lambda fake: torch.randn_like(fake, device='cuda'),
-            }
+            },
         )
     """
     from torch._library.custom_ops import CustomOpDef
@@ -377,12 +398,12 @@ def register_custom_op_autotuning(
         raise TypeError(f"configs must be a list or tuple, got {type(configs)}")
 
     processed_configs = []
-    for config in configs:
-        if isinstance(config, CustomOpConfig):
-            processed_configs.append(config)
+    for cfg in configs:
+        if isinstance(cfg, CustomOpConfig):
+            processed_configs.append(cfg)
         else:
             raise TypeError(
-                f"Each config must be a CustomOpConfig object, got {type(config)}"
+                f"Each config must be a CustomOpConfig object, got {type(cfg)}"
             )
 
     if not processed_configs:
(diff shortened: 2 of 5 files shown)


PaulZhang12:
Nice! This is awesome, curious why is there 1 case in which the perf was worse?

torch/_inductor/scheduler.py
```diff
@@ -3463,6 +3463,7 @@ def speedup_by_fusion(
             and isinstance(n.get_template_node(), ir.MultiTemplateBuffer)
             for n in (node1, node2)
         )
+
```
PaulZhang12: nit: unrelated?

torch/_inductor/select_algorithm.py
```diff
@@ -2700,6 +2700,7 @@ def __call__(
         precompilation_timeout_seconds: int = 60 * 60,
         return_multi_template=False,
         best_config_future=None,
+        return_choice=False,
```
PaulZhang12: Fine for now, perhaps a refactor where we only return choice is better, as we can get node from choice

PaulZhang12 (approved):
LGTM! Probably wait until @eellison takes a look as well :). Also, does this automatically enable decomposeK fusions now if decomposeK wins with max_autotune_gemm? Or that is still yet to be implemented?

torch/_inductor/kernel/custom_op.py
```diff
@@ -0,0 +1,544 @@
...
+    ExternKernelChoice,
+)
+from torch._inductor.virtualized import V
+from torch.utils._ordered_set import OrderedSet
+
+
+log = logging.getLogger(__name__)
+
+
+class CustomOpConfig:
```
eellison, line 23: Can you rebase this onto main ? this code already exists in main. don't want to review code that's already landed. in the future i recommend using ghstack for this workflow. both for yourself and to make it easier for your reviewers.

eellison (commented):
please rebase onto main since much of this code has already landed

torch/_inductor/kernel/custom_op.py
tianrengao: Rebased it onto main and composed the inline fusion to mm.py when not using multiTemplateBuffer. Updated the benchmark result for Torch Compile with gemm max-autotune and inline fusion in the PR description.

Yea will make sure no landed code for review in the future.

torch/_inductor/config.py
```diff
@@ -693,6 +701,11 @@ def use_autoheuristic(name: str) -> bool:
 # Take how many of the top triton kernels to benchmark epilogue
 max_epilogue_benchmarked_choices = 1
 
+# Enable inline fusion for subgraphs (custom_op, mm.py decomposek, etc.)
+# Only effective when benchmark_epilogue_fusion=False (immediate autotune path)
+# When True, inline the winning subgraph choice to IR nodes for epilogue fusion
+enable_inline_subgraph_fusion = False
```
tianrengao: shall we make this defualt or a flag?

torch/_inductor/codegen/subgraph.py
```diff
@@ -23,6 +23,69 @@
...
+        TensorBox containing the final operation result as individual IR nodes
+    """
+    from torch._inductor.lowering import lowerings
+
+    # Create mapping from placeholder nodes to actual inputs
+    node_to_value = {}
+    placeholder_idx = 0
+
+    # Process each node in the graph
+    for node in gm.graph.nodes:
```
eellison: https://github.com/pytorch/pytorch/blob/a5f3035aafd5113dd7641a95a3e919d4a4c8781f/torch/_inductor/lowering.py#L7341-L7352 is more along the lines of what you want.

torch/_inductor/kernel/custom_op.py
```diff
@@ -318,20 +321,70 @@ def autotune_custom_op(
...
         name=name,
         choices=choices,
         input_nodes=list(inputs),
         layout=choices[0].layout,
         input_gen_fns=input_gen_fns,
+        return_choice=True,
     )
 
+    # Apply inlining for fusion if winning_choice has graph; otherwise return result as-is(default impl)
+    if hasattr(winning_choice, "gm"):
```
eellison: We shouldn't be conditionally assigning attrs. can this be None if it's not set ?

torch/_inductor/kernel/mm.py
```diff
@@ -1315,6 +1315,39 @@ def _to_dtype(x):
         # The future will be awaited at scheduling time in select_algorithm.py
         best_config_future = gen_best_config(mat1, mat2)
 
+    # If not using benchmark_epilogue_fusion, try to inline subgraph for fusion
+    if (
+        not inductor_config.benchmark_epilogue_fusion
+        and inductor_config.enable_inline_subgraph_fusion
+    ):
+        result, winning_choice = autotune_select_algorithm(
+            name,
```
eellison: This interferes with async compilation. if you were to turn this on a do a benchmark run with  max-autotune, youd see a huge regresssion.

torch/_inductor/kernel/mm.py
```diff
@@ -1315,6 +1315,39 @@ def _to_dtype(x):
...
+            choices,
+            kernel_inputs.nodes(),
+            layout,
+            best_config_future=best_config_future,
+            return_choice=True,
+        )
+
+        # If choice has graph to lower, inline it; otherwise use result as-is(cublas mm)
+        if hasattr(winning_choice, "gm"):
```
eellison: can we do a better check then hasatt?

torch/_inductor/select_algorithm.py
eellison: +1 - this is incorrect types now.

torch/_inductor/codegen/subgraph.py
eellison: is there a refactor we could do ?

torch/_inductor/kernel/custom_op.py
```diff
@@ -318,20 +321,70 @@ def autotune_custom_op(
...
+
+    return inline_subgraph_to_ir_nodes(winning_choice.gm, inputs, name)
+
 
 def register_custom_op_autotuning(
     custom_op: torch._library.custom_ops.CustomOpDef,
     configs: Union[list[CustomOpConfig], list[Callable[..., Any]]],
     name: Optional[str] = None,
     input_gen_fns: Optional[dict[str, Callable[[torch.Tensor], torch.Tensor]]] = None,
+    enable_fusion: bool = False,
```
eellison: Can we remove this arg if we don't have any expectation of when they would want this to be False ?

eellison (commented):
The main reason we have MultiTemplateBuffer and deferred choice is so that we can overlap compilations of max-autotune. We shouldn't be turning this for our own mm implementations, at least if there are any non decomposition /extern choices.

I'm willing to approve this pr for user custom kernels, since it's unlikely they're spawning 10+ triton mm compilations, and we don't have the overlapping working there yet. It's a little unfortunate that this pr doesn't fix the mm decomposition case (without eagerly awaiting all async compilations). 

have a few, comments, first.

eellison (commented):
The main reason we have MultiTemplateBuffer and deferred choice is so that we can overlap compilations of max-autotune. We shouldn't be turning this for our own mm implementations, at least if there are any non decomposition /extern choices.

I'm willing to approve this pr for user custom kernels, since it's unlikely they're spawning 10+ triton mm compilations, and we don't have the overlapping working there yet. It's a little unfortunate that this pr doesn't fix the mm decomposition case (without eagerly awaiting all async compilations). 

have a few, comments, first.

eellison (commented):
The main reason we have MultiTemplateBuffer and deferred choice is so that we can overlap compilations of max-autotune. We shouldn't be turning this for our own mm implementations, at least if there are any non decomposition /extern choices.

I'm willing to approve this pr for user custom kernels, since it's unlikely they're spawning 10+ triton mm compilations, and we don't have the overlapping working there yet. It's a little unfortunate that this pr doesn't fix the mm decomposition case (without eagerly awaiting all async compilations). 

have a few, comments, first.

torch/_inductor/select_algorithm.py
tianrengao: Yea, this has been used by many places in the codebase. I prefer to have a separate refactor PR to fix this: return choice and get nodes from choice.

eellison (approved):

PR #165953, opened 2025-10-21
Introducing the StableIValue representation of list :D

janeyx99:
Some important notes:
a) Just like IValues steal the ownership of ArrayRefs and any std::vectors in order to convert the inner elements into IValues, we do the same thing with StableIValue. This O(N) traverse is ineluctable.
b) As a result, since StableIValues are owning and our contract is that to<T>(StableIValue) transfers ownership, you cannot ever convert from StableIValue to a nonowning HeaderOnlyArrayRef<V>.

We handle memory similar to AtenTensorHandle, but we have a StableListHandle!

* #167126
* __->__ #165953
* #165153
* #165152
* #164991


test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
```diff
@@ -478,6 +478,64 @@ STABLE_TORCH_LIBRARY_IMPL(libtorch_agnostic, CompositeExplicitAutograd, m) {
...
+
+std::vector<Tensor> make_tensor_clones_and_call_foreach(Tensor t1, Tensor t2) {
+  // This function tests that my__foreach_mul can take in non-owning HeaderOnlyArrayRefs
+  // in addition to std::vectors.
+  Tensor t1_1 = my_clone(t1);
+  Tensor t1_2 = my_clone(t1);
+  Tensor t2_1 = my_clone(t2);
+  Tensor t2_2 = my_clone(t2);
+  auto l1 = torch::headeronly::HeaderOnlyArrayRef<Tensor>({t1_1, t2_1});
+  auto l2 = torch::headeronly::HeaderOnlyArrayRef<Tensor>({t1_2, t2_2});
```
swolchok: this sets up dangling arrayrefs. The arrays they refer to are temporary and their lifetime therefore ends at the end of the statement. You have to declare the arrays as named variables instead. I would expect ASAN complaints about this test

torch/csrc/stable/c/shim.h
```diff
@@ -37,6 +37,34 @@ AOTI_TORCH_EXPORT AOTITorchError torch_library_impl(
...
+struct StableListOpaque;
+using StableListHandle = StableListOpaque*;
+
+// returns an owning reference of a StableList. callee is responsible for
+// freeing memory.
+AOTI_TORCH_EXPORT AOTITorchError
+torch_new_list_reserve_size(size_t size, StableListHandle* ret);
+
+// new list handle steals ownership of underlying elements from orig_handle
+AOTI_TORCH_EXPORT AOTITorchError torch_new_list_handle(
```
swolchok: maybe call it torch_move_list_handle? also this seems like an odd function to need; why can't the caller just continue to use `orig_handle`?

torch/csrc/shim_common.cpp
```diff
@@ -4,12 +4,64 @@
...
+  AOTI_TORCH_CONVERT_EXCEPTION_TO_ERROR_CODE(
+      { *ret = list_pointer_to_list_handle(list_ptr.release()); });
+}
+
+AOTITorchError torch_new_list_handle(
+    StableListHandle orig_handle,
+    StableListHandle* new_handle) {
+  AOTI_TORCH_CONVERT_EXCEPTION_TO_ERROR_CODE({
+    std::vector<StableIValue>* list = list_handle_to_list_pointer(orig_handle);
+    *new_handle = new_list_handle(std::vector<StableIValue>(*list));
```
swolchok: because this is stealing ownership, is orig_handle even usable afterwards? if not, why not std::move() from *list? (and see previous "why do I need this" comment)

torch/csrc/stable/c/shim.h
```diff
@@ -37,6 +37,34 @@ AOTI_TORCH_EXPORT AOTITorchError torch_library_impl(
...
+    StableListHandle list_handle,
+    size_t index,
+    StableIValue* element);
+
+AOTI_TORCH_EXPORT AOTITorchError
+torch_list_emplace_back(StableListHandle list_handle, StableIValue element);
+
+// deletes the underlying list referenced by list_handle
+AOTI_TORCH_EXPORT AOTITorchError
+torch_delete_list_object(StableListHandle list_handle);
```
swolchok: why does this function have "object" in the name when none of the other list functions do?

torch/csrc/stable/c/shim.h
```diff
@@ -37,6 +37,34 @@ AOTI_TORCH_EXPORT AOTITorchError torch_library_impl(
...
+AOTI_TORCH_EXPORT AOTITorchError
+torch_list_size(StableListHandle list_handle, size_t* size);
+
+AOTI_TORCH_EXPORT AOTITorchError torch_list_get(
+    StableListHandle list_handle,
+    size_t index,
+    StableIValue* element);
+
+AOTI_TORCH_EXPORT AOTITorchError
+torch_list_emplace_back(StableListHandle list_handle, StableIValue element);
```
swolchok: why is it emplace and not push

torch/csrc/stable/c/shim.h
```diff
@@ -37,6 +37,34 @@ AOTI_TORCH_EXPORT AOTITorchError torch_library_impl(
...
+AOTI_TORCH_EXPORT AOTITorchError
+torch_new_list_reserve_size(size_t size, StableListHandle* ret);
+
+// new list handle steals ownership of underlying elements from orig_handle
+AOTI_TORCH_EXPORT AOTITorchError torch_new_list_handle(
+    StableListHandle orig_handle,
+    StableListHandle* new_handle);
+
+AOTI_TORCH_EXPORT AOTITorchError
+torch_list_size(StableListHandle list_handle, size_t* size);
```
swolchok, line 49: torch_list_get_size?

torch/csrc/stable/c/shim.h
```diff
@@ -37,6 +37,34 @@ AOTI_TORCH_EXPORT AOTITorchError torch_library_impl(
...
+
+AOTI_TORCH_EXPORT AOTITorchError torch_list_get(
+    StableListHandle list_handle,
+    size_t index,
+    StableIValue* element);
```
swolchok: torch_list_get_item?

torch/csrc/stable/stableivalue_conversions.h
```diff
@@ -192,6 +192,38 @@ struct FromImpl<torch::stable::Tensor> {
   }
 };
 
+// Specialization for torch::headeronly::HeaderOnlyArrayRef<T> => StableIValue
+// Returns a new OWNING reference of the underlying list!
```
swolchok: why are we yelling about ownership here (caps and exclamation point) but not above?

swolchok (commented):
I had a lot of confusion around the ownership model before I went back and reviewed how ownership of tensors works and confirmed that (I think) StableIValue owns lists to the same extent that it owns tensors -- it takes care of its own cleanup. I think it's probably OK but might be something to think about.

torch/csrc/stable/c/shim.h
janeyx99: I was misguided in thinking that emplace_back is faster, but a google search tells me that emplace_back is only better if I haven't already constructed the thing to push back. Will modify.

torch/csrc/stable/c/shim.h
janeyx99: it was to say to delete the underlying list but i agree, will remove. i had just copied this from tensor_object but it doesn't really illuminate anything.

torch/csrc/stable/c/shim.h
janeyx99: I'm following the semantic where if it's blah->api(), I'm naming it torch_blah_api, so this one stays.

torch/csrc/stable/c/shim.h
janeyx99: But this one should be `torch_list_at(...)`

torch/csrc/stable/stableivalue_conversions.h
janeyx99: ok cuz i'm dramatic. i will tone it down and be professional here too.

torch/csrc/stable/c/shim.h
janeyx99: yea this was leftover when i had an overcomplicated design for memory, but we do not need it anymore.

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
```diff
@@ -478,6 +478,62 @@ STABLE_TORCH_LIBRARY_IMPL(libtorch_agnostic, CompositeExplicitAutograd, m) {
   m.impl("my_amax_vec", &boxed_my_amax_vec);
 }
 
+std::vector<Tensor> my__foreach_mul(torch::headeronly::HeaderOnlyArrayRef<Tensor> self, torch::headeronly::HeaderOnlyArrayRef<Tensor> other) {
+  const auto num_args = 2;
+  StableIValue stack[num_args];
+  stack[0] = from(self);
+  stack[1] = from(other);
```
swolchok: nit: this is acceptable, but since num_args isn't actually used anywhere else and I hate being nagged by linters, I would write

std::array<StableIValue, 2> stack = {{from(self), from(other)}};

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
```diff
@@ -478,6 +478,62 @@ STABLE_TORCH_LIBRARY_IMPL(libtorch_agnostic, CompositeExplicitAutograd, m) {
...
+void boxed_my__foreach_mul(StableIValue* stack, uint64_t num_args, uint64_t num_outputs) {
+  // Why is the following NOT to<HeaderOnlyArrayRef<Tensor>>(stack[0])? Because calling `to`
+  // on a StableIValue means that the result is owning its underlying data now! HeaderOnlyArrayRef
+  // is not owning, so it cannot safely steward the result of the to<>.
+  auto res = my__foreach_mul(to<std::vector<Tensor>>(stack[0]), to<std::vector<Tensor>>(stack[1]));
+  stack[0] = from(res);
+}
+
+void my__foreach_mul_(torch::headeronly::HeaderOnlyArrayRef<Tensor> self, torch::headeronly::HeaderOnlyArrayRef<Tensor> other) {
+  const auto num_args = 2;
```
swolchok: ditto

test/cpp_extensions/libtorch_agnostic_extension/libtorch_agnostic/csrc/kernel.cpp
```diff
@@ -478,6 +478,62 @@ STABLE_TORCH_LIBRARY_IMPL(libtorch_agnostic, CompositeExplicitAutograd, m) {
...
+  aoti_torch_call_dispatcher("aten::_foreach_mul_", "List", stack);
+}
+
+void boxed_my__foreach_mul_(StableIValue* stack, uint64_t num_args, uint64_t num_outputs) {
+  my__foreach_mul_(to<std::vector<Tensor>>(stack[0]), to<std::vector<Tensor>>(stack[1]));
+}
+
+std::vector<Tensor> make_tensor_clones_and_call_foreach(Tensor t1, Tensor t2) {
+  // This function tests that my__foreach_mul can take in non-owning HeaderOnlyArrayRefs
+  // in addition to std::vectors.
```
swolchok: this no longer seems particularly true; they're `std::initializer_list`s

torch/csrc/stable/c/shim.h
malfet: Hmm, on one hand I understand the desire to keep `error_t foo(inp, *out)` semantic, but on the other hand, the only it it could fail if `list_handle` is nullptr, isn't it?  Should we indeed for the sake of speed change semantic a bit to `size_t torch_list_get_size(StableListHandle)`?

torch/csrc/stable/c/shim.h
```diff
@@ -37,6 +37,29 @@ AOTI_TORCH_EXPORT AOTITorchError torch_library_impl(
...
+AOTI_TORCH_EXPORT AOTITorchError
+torch_new_list_reserve_size(size_t size, StableListHandle* ret);
+
+AOTI_TORCH_EXPORT AOTITorchError
+torch_list_size(StableListHandle list_handle, size_t* size);
+
+AOTI_TORCH_EXPORT AOTITorchError torch_list_at(
+    StableListHandle list_handle,
+    size_t index,
+    StableIValue* element);
```
malfet: Same as above: if  `at::List::operator[]` doesn't do boundary checks, this API shouldn't do them either

torch/csrc/stable/stableivalue_conversions.h
```diff
@@ -192,6 +193,37 @@ struct FromImpl<torch::stable::Tensor> {
...
 };
 
+// Specialization for torch::headeronly::HeaderOnlyArrayRef<T> => StableIValue
+// Returns a new owning reference of the underlying list.
+template <typename T>
+struct FromImpl<torch::headeronly::HeaderOnlyArrayRef<T>> {
+  static StableIValue call(
+      const torch::headeronly::HeaderOnlyArrayRef<T>& val,
+      [[maybe_unused]] uint64_t extension_build_version,
+      [[maybe_unused]] bool is_internal) {
```
malfet, line 203: Just curious, what are the purpose of those?

torch/csrc/stable/stableivalue_conversions.h
```diff
@@ -342,6 +374,33 @@ struct ToImpl<torch::stable::Tensor> {
...
+  static std::vector<T> call(
+      StableIValue val,
+      [[maybe_unused]] uint64_t extension_build_version,
+      [[maybe_unused]] bool is_internal) {
+    auto list_handle = to<StableListHandle>(val);
+    size_t size;
+    TORCH_ERROR_CODE_CHECK(torch_list_size(list_handle, &size));
+    std::vector<T> result;
+    result.reserve(size);
+    for (size_t i = 0; i < size; i++) {
```
malfet: Why can't we use `irange` for stable ABI?
```suggestion
    for (const auto i  : c10::irange(size)) {
```

torch/csrc/shim_common.cpp
```diff
@@ -4,12 +4,55 @@
...
 
+AOTITorchError torch_new_list_reserve_size(size_t size, StableListHandle* ret) {
+  auto list_ptr = std::make_unique<std::vector<StableIValue>>();
+  list_ptr->reserve(size);
+  AOTI_TORCH_CONVERT_EXCEPTION_TO_ERROR_CODE(
+      { *ret = list_pointer_to_list_handle(list_ptr.release()); });
+}
+
+AOTI_TORCH_EXPORT AOTITorchError
+torch_list_size(StableListHandle list_handle, size_t* size) {
```
malfet, line 21: Do you want to check first that `list_handle` and `size` are not nullptrs?

torch/csrc/stable/stableivalue_conversions.h
mikaylagawarecki: These are scaffolding so that in the unlikely case that the memory layout of a struct passed as a stableivalue changes between two releases, we can appropriately modify `from` and `to` (and `from_ivalue` + `to_ivalue` that call into these) to handle this. See https://github.com/pytorch/pytorch/pull/165284/files#diff-e260bc7cf1dcf33d3ce5fa484f918553929e57bc8ae76d398d4b1d884fe8654eR420-R437 for an example use

(I'll have better developer documentation to come, there are also tests that ensure the memory layout of the relevant structs don't change coming)

torch/csrc/stable/stableivalue_conversions.h
janeyx99: We'd have to move irange to headeronly first, but otherwise we can

torch/csrc/stable/c/shim.h
swolchok: 1) it seems odd that there is no corresponding way to set an item at an index
2) I wonder if we should consider exposing a raw data() pointer for efficiency. I suppose we can always add it later if we skip it for now, but we can't take it away later if we do it now and then decide it was a bad idea.

torch/csrc/stable/c/shim.h
```diff
@@ -37,6 +37,29 @@ AOTI_TORCH_EXPORT AOTITorchError torch_library_impl(
...
+
+AOTI_TORCH_EXPORT AOTITorchError
+torch_list_push_back(StableListHandle list_handle, StableIValue element);
```
swolchok, line 62: is it permissible to be wrong about the reserved size?

torch/csrc/stable/c/shim.h
```diff
@@ -37,6 +37,29 @@ AOTI_TORCH_EXPORT AOTITorchError torch_library_impl(
     void (*fn)(StableIValue*, uint64_t, uint64_t),
     uint64_t extension_build_version);
 
+struct StableListOpaque;
+using StableListHandle = StableListOpaque*;
+
+// returns an owning reference of a StableList. callee is responsible for
+// freeing memory.
+AOTI_TORCH_EXPORT AOTITorchError
+torch_new_list_reserve_size(size_t size, StableListHandle* ret);
```
swolchok, line 46: interestingly, this does not correspond to any constructor of std::vector, though I like avoiding a second crossing of the ABI boundary just to do a reserve.

torch/csrc/stable/stableivalue_conversions.h
```diff
@@ -342,6 +374,33 @@ struct ToImpl<torch::stable::Tensor> {
   }
 };
 
+// Specialization for StableIValue => std::vector<T>
+// std::vector<T> should be represented as a StableListHandle
+// filled with StableIValues
+// The new std::vector steals ownership of the underlying elements
+// and StableListHandle is freed.
```
swolchok: this isn't quite right -- the list the handle refers to is freed, not the handle itself, right?

torch/csrc/stable/c/shim.h
janeyx99: I'm leaning towards no for now, because it would mean we should guarantee in the long term that the function should never fail, and that's not worth the small speed bump to me.

torch/csrc/shim_common.cpp
janeyx99: Ehhhh these are meant to be used mostly internal so I'm not convinced we need to add checks

torch/csrc/stable/c/shim.h
janeyx99: Oh I see....this function indeed doesn't have the same semantic as the C++ API as there's no way to set with my C API. I'll rename and add a way to set.

I agree we should not rush to add things

torch/csrc/stable/c/shim.h
janeyx99: I think reserving size is just a performance optimization, it doesn't actually populate the tensor. So after reserving the size is still 0.

torch/csrc/stable/stableivalue_conversions.h
```diff
@@ -192,6 +193,37 @@ struct FromImpl<torch::stable::Tensor> {
...
+  static StableIValue call(
+      const torch::headeronly::HeaderOnlyArrayRef<T>& val,
+      [[maybe_unused]] uint64_t extension_build_version,
+      [[maybe_unused]] bool is_internal) {
+    StableListHandle new_list_handle;
+    TORCH_ERROR_CODE_CHECK(
+        torch_new_list_reserve_size(val.size(), &new_list_handle));
+    for (const auto& elem : val) {
+      TORCH_ERROR_CODE_CHECK(torch_list_push_back(new_list_handle, from(elem)));
+    }
```
swolchok: you're going to leak memory for the list and for previously-pushed elements if this TORCH_ERROR_CODE_CHECK ever fails, courtesy of StableIValue not having a destructor

torch/csrc/stable/stableivalue_conversions.h
```diff
@@ -342,6 +374,33 @@ struct ToImpl<torch::stable::Tensor> {
...
+      [[maybe_unused]] uint64_t extension_build_version,
+      [[maybe_unused]] bool is_internal) {
+    auto list_handle = to<StableListHandle>(val);
+    size_t size;
+    TORCH_ERROR_CODE_CHECK(torch_list_size(list_handle, &size));
+    std::vector<T> result;
+    result.reserve(size);
+    for (size_t i = 0; i < size; i++) {
+      StableIValue element;
+      TORCH_ERROR_CODE_CHECK(torch_list_get_item(list_handle, i, &element));
```
swolchok: ditto leaky

swolchok (commented):
looking pretty good

torch/csrc/stable/stableivalue_conversions.h
```diff
@@ -202,12 +202,21 @@ struct FromImpl<torch::headeronly::HeaderOnlyArrayRef<T>> {
...
-      TORCH_ERROR_CODE_CHECK(torch_list_push_back(new_list_handle, from(elem)));
+    try {
+      TORCH_ERROR_CODE_CHECK(
+          torch_new_list_reserve_size(val.size(), &new_list_handle));
+      for (const auto& elem : val) {
+        TORCH_ERROR_CODE_CHECK(
+            torch_list_push_back(new_list_handle, from(elem)));
+      }
+      return from(new_list_handle);
+    } catch (const std::runtime_error& e) {
```
swolchok, line 213: don't you need to catch all flavors of exception? why is it ok to leak on things that aren't runtime_error?

torch/csrc/stable/stableivalue_conversions.h
```diff
@@ -202,12 +202,21 @@ struct FromImpl<torch::headeronly::HeaderOnlyArrayRef<T>> {
...
+      if (new_list_handle != nullptr) {
+        // clean up memory if an error was thrown
+        TORCH_ERROR_CODE_CHECK(torch_delete_list(new_list_handle));
+      }
+      throw e;
```
swolchok: drop the `e`, just `throw;`. it's cleaner.

torch/csrc/stable/stableivalue_conversions.h
```diff
@@ -387,17 +396,22 @@ struct ToImpl<std::vector<T>> {
...
+        StableIValue element;
+        TORCH_ERROR_CODE_CHECK(torch_list_get_item(list_handle, i, &element));
+        result.push_back(to<T>(element));
+      }
+      TORCH_ERROR_CODE_CHECK(torch_delete_list(list_handle));
+      return result;
+    } catch (const std::runtime_error& e) {
+      // clean up memory if an exception is thrown, and rethrow
+      TORCH_ERROR_CODE_CHECK(torch_delete_list(list_handle));
+      throw e;
```
swolchok: same as above

torch/csrc/stable/stableivalue_conversions.h
janeyx99: The only thing that TORCH_ERROR_CODE_CHECK can throw is std::runtime_error

torch/csrc/stable/stableivalue_conversions.h
janeyx99: Wait does that do the same thing? Does that rethrow the error?

janeyx99: Ohhh it issss

torch/csrc/stable/stableivalue_conversions.h
swolchok: fair enough, nothing should get thrown through ABI boundary

torch/csrc/stable/stableivalue_conversions.h
swolchok: generally rethrowing can also do stuff like cause backtraces to be preserved.

malfet: `throw e;` can cause object slicing, while `throw` preserve type of original exception?

torch/csrc/shim_common.cpp
malfet: OK, then why do you want to wrap this code with `AOTI_TORCH_CONVERT_EXCEPTION_TO_ERROR_CODE`? Semantically this code will never throw, would it?

malfet (approved):

torch/csrc/shim_common.cpp
swolchok: FWIW, it's harmless if size() gets inlined and so I am weakly in favor of it because it makes everything visually uniform (lack of `AOTI_TORCH_CONVERT_EXCEPTION_TO_ERROR_CODE` is known-bad). https://godbolt.org/z/zbqbhez87

janeyx99: I see your point; I have just learned that dereferencing a bad pointer will lead to crash/UB but not throw an Exception. Regardless, the way I'm thinking of adding these APIs is to be defensive against future code modifications. I would feel more comfortable removing the macro once we have a test to ensure that all shim APIs never throw Exceptions. 

I indeed will land the PR as is but this is a good consideration. Whatever policy we adopt I would want to apply across all APIs --> if we check for nullptr here we should check in the Tensor shims too.

PR #165957, opened 2025-10-21
Make Backend::setGroupUid virtual

Zha0q1:
As titled, so that we may customize this function in custom backends

### DIFF
--- a/torch/csrc/distributed/c10d/Backend.hpp
+++ b/torch/csrc/distributed/c10d/Backend.hpp
@@ -419,7 +419,7 @@ class TORCH_API Backend : public torch::CustomClassHolder {
   }
 
   // Do not call this directly, use ProcessGroup::setGroupName instead.
-  void setGroupUid(const std::string& pg_uid) {
+  virtual void setGroupUid(const std::string& pg_uid) {
     pg_uid_ = pg_uid;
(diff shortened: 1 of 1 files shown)


d4l3k:
This change seems fine but need to double check rollout strategy with some folks since we had backwards compatibility issues the last time we modified the c10d::Backend vtable

torch/csrc/distributed/c10d/Backend.hpp
```diff
@@ -419,7 +419,7 @@ class TORCH_API Backend : public torch::CustomClassHolder {
   }
 
   // Do not call this directly, use ProcessGroup::setGroupName instead.
-  void setGroupUid(const std::string& pg_uid) {
+  virtual void setGroupUid(const std::string& pg_uid) {
```
Skylion007, line 422: Shouldn't getGroupUid also be virtual then?

Zha0q1: My use case is to establish mapping from the pg_id to some internal resources owned by a custom backend. For me I don't need to override getGroupUid. For symmetry we could add it too

d4l3k (approved):
LGTM, no backwards compatibility concerns

PR #165962, opened 2025-10-21
[DTensor][Op] fix for DTensor ops with Partial placements

anshul-si:
**Summary:** When operations are done on partial placements, we use sharding logic to incorrectly determine whether we should redistribute the tensor to replicate. By delaying the redistribution, we do the operation first, and then the partial reduction. This leads to incorrect results for max, min, gradient norm clipping, and more. We solve this by setting reduction_linear to False when there is a Partial placement to force the redistribution before completing the op.

**Test Cases**
1. pytest test/distributed/tensor/test_math_ops.py -k test_partial_reduction_ops
2. pytest test/distributed/tensor/test_math_ops.py -k test_matching_partial_reduction_ops

* __->__ #165962

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @tianyu-l @XilunWu @SherlockNoMad

### DIFF
--- a/torch/distributed/tensor/_ops/_math_ops.py
+++ b/torch/distributed/tensor/_ops/_math_ops.py
@@ -291,11 +291,10 @@ def common_reduction_strategy(
                     reduction_linear = False
                     break
 
-        # TODO: there are some combinations of local and global reductions that allow for
-        # delaying redistribute such as max, max
-
         for p in op_spec.output_spec.placements:
-            if isinstance(p, Partial):
+            # when the partial reduction op matches the global reduction op,
+            # we can delay redistribution (i.e max, max)
+            if isinstance(p, Partial) and p.reduce_op != reduction_op:
                 reduction_linear = False
                 break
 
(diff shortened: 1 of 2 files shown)


torch/distributed/tensor/_ops/_math_ops.py
```diff
@@ -291,6 +291,13 @@ def common_reduction_strategy(
                     reduction_linear = False
                     break
 
+        for p in op_spec.output_spec.placements:
```
wconstab, line 294: I suspect this change is too aggressive, shouldn't it consider 'reduction_dims' and only set reduction_linear to False if the reduce op is happening on a dim that is partial and differs from reduce_op?

anshul-si: I may be misunderstanding reduce_dims, but I thought reduce_dims refers to tensor dims rather than mesh dims. I get how Shard(n) relates to tensor dims, but it doesn't seem like Partial can be connected in the same way? 

image

wconstab: You're right! Partial applies to a mesh dim and a reduce op applies to a tensor dim, and I was confused in my original comment.  I think this code is correct.

wconstab (approved):
lgtm. thanks for the fix!

PR #165963, opened 2025-10-21
[MTIAGraph][Pytorch][2/n] Add binding for Python to C++, and hook for Pytorch to Fbcode

andyanwang:
Summary:
This diff is the binding and hook layer for MTIA Graph, including
1. binding between Python and C++
2. hook between Pytorch and mtia fbcode
image

[Doc](https://docs.google.com/document/d/1Q3xdZAIqhBvuy2HxGDfJyXVmxYXUEeYSZSwsp7bcJF8/edit?tab=t.osb46a42t6wb#heading=h.ayp9tkk08x00)

Test Plan: Will be tested in the python implementation which will use the binding and hook

Differential Revision: D84457757

cc @egienvalue

### DIFF
--- a/aten/src/ATen/detail/MTIAHooksInterface.h
+++ b/aten/src/ATen/detail/MTIAHooksInterface.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <c10/core/CachingDeviceAllocator.h>
 #include <c10/core/Device.h>
 #include <c10/util/Exception.h>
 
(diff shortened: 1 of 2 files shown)


torch/csrc/mtia/Module.cpp
```diff
@@ -10,6 +10,42 @@
...
+struct _MTIAGraph {
+  // MTIA use accelorator hooks to connect pytorch and outside.
+  // We need to provide the MTIAGraph class at Python layer, but the hooks only
+  // support hooking functions, not classes. Thus we store all MTIAGraph C++
+  // instances in a map, and use a handle to choose the right instance.
+  int64_t handle_;
+
+  _MTIAGraph(bool keep_graph = false)
+      : handle_(at::detail::getMTIAHooks().mtiagraphCreate(keep_graph)) {}
+  ~_MTIAGraph() {}
```
Skylion007: Shouldn't this equal default?

malfet (approved):
LGTM

aten/src/ATen/detail/MTIAHooksInterface.h
```diff
@@ -8,8 +9,8 @@
 
 #include <c10/core/Allocator.h>
 
-#include <c10/util/python_stub.h>
 #include <ATen/detail/AcceleratorHooksInterface.h>
+#include <c10/util/python_stub.h>
```
malfet, line 12: Why change the header order here (and may be move all `c10` headers to one block)

aten/src/ATen/detail/MTIAHooksInterface.h
```diff
@@ -91,7 +92,7 @@ struct TORCH_API MTIAHooksInterface : AcceleratorHooksInterface {
     return c10::Stream::unpack3(-1, 0, c10::DeviceType::MTIA);
   }
 
-  virtual void setCurrentStream(const c10::Stream& /*stream*/ ) const {
+  virtual void setCurrentStream(const c10::Stream& /*stream*/) const {
```
malfet: Feels like could be done by separate commit

aten/src/ATen/detail/MTIAHooksInterface.h
andyanwang: I think it's done by lintrunner. I'll double check

albanD (approved):
SGTM

PR #156872, opened 2025-06-25
chunk_size should always be int64_t for Foreach functors

janeyx99:
See https://github.com/pytorch/pytorch/issues/156261#issuecomment-3002394773

Testing is a valid q--it is pretty expensive to test such large tensors for all these ops.

Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #156872
* #156871
* #156876

PR #165971, opened 2025-10-21
[BugFix] chunk_size should always be int64_t

lingebeng:
aspired by https://github.com/pytorch/pytorch/pull/156872


cyyever:
While it is too expensive to create large dense tensors, is it possible to create large sparse tensors to test the ops?

lingebeng:
Thanks,I see. Maybe it's not a bug!

cyyever:
@lingebeng We can have further chats via e-mail.

albanD (approved):
@cyyever my layman understanding is that `int` is `int64_t` on linux/mac but `int32_t` on windows. So I would only expect to see this fail on windows.

Generally, we do want to make these type explicit to avoid windows-only issues.

PR #165983, opened 2025-10-21
[inductor] Fix argmin/argmax returning incorrect indices for non-contiguous tensor

### LINKED ISSUE #163929: [inductor] `argmax` outputs wrong when working on transposed and mutated matrix
### 🐛 Describe the bug

Here is the code:

```python
import torch

def foo(x):
    x.tan_()
    x = x.t()
    return x.argmin()


torch.manual_seed(0)
x1 = torch.randn(4, 6)
x2 = x1.clone()
out1 = foo(x1)
cf = torch.compile(foo)
out2 = cf(x2)
torch.testing.assert_close(out1, out2)
```

Error:

```
AssertionError: Scalars are not equal!

Expected 1 but got 4.
Absolute difference: 3
Relative difference: 3.0
```
Please kindly note that `x.tan_()` and `x=x.t()` is necessary to reproduce the error.
The affected backend is `inductor`. The `aot_eager` and `eager` output correct results.

### Versions

[pip3] numpy==1.26.4
[pip3] optree==0.17.0
[pip3] pytorch-triton==3.5.0+git5ae38bdb
[pip3] torch==2.10.0a0+gitf34744d
[conda] numpy                      1.26.4               pypi_0              pypi
[conda] optree                     0.17.0               pypi_0              pypi
[conda] pytorch-triton             3.5.0+git5ae38bdb    pypi_0              pypi
[conda] torch                      2.10.0a0+gitf34744d  pypi_0              pypi

cc @ezyang @gchanan @zou3519 @kadeng @msaroufim @chauhang @penguinwu @voznesenskym @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @muchulee8 @amjames @aakhundov @coconutruben

karthickai:
* __->__ #165983

Fixes #163929

Fixes argmin/argmax operations to return correct logical indices instead of physical memory offsets when applied to transposed/permuted tensors.  When `argmin()` or `argmax()` is called on a transposed tensor, Inductor was returning physical memory indices instead of logical row-major indices. This caused incorrect results that don't match eager mode behavior.  


torch/_inductor/lowering.py
```diff
@@ -6088,6 +6090,17 @@ def _make_reduction_inner(x, *, axis, keepdims, dtype, override_return_dtype):
             kept_idx.append(i)
             kept_sizes.append(size[i])
 
+    # For argmax/argmin compute logical indices when the tensor has non-contiguous layout.
```
thenumberouscode, line 6094: oh, I have a question: why we only need compute logical indices for argmax/argmin?

karthickai: because argmin/argmax are the only reductions that return indices rather than values

torch/_inductor/lowering.py
```diff
@@ -6089,6 +6091,21 @@ def _make_reduction_inner(x, *, axis, keepdims, dtype, override_return_dtype):
...
+        and len(reduced_sizes) > 1
+        and is_triton(x)
+    ):
+        if isinstance(x.data, PermuteView):
+            should_compute_logical_index = True
+        elif isinstance(x.data, ir.ReinterpretView):
+            layout = x.get_layout()
+            should_compute_logical_index = (
+                layout.is_transposed() or not layout.is_contiguous()
+            )
```
shunting314, line 6109: This does not cover the case when the function input is in column major layout already.

test/inductor/test_torchinductor.py
```diff
@@ -14305,6 +14305,31 @@ def fn(repeat, output_size, data):
...
+        self.common(fn, (torch.randn(6, 4, device=GPU_TYPE),))
+        self.common(fn, (torch.randn(128, 64, device=GPU_TYPE),))
+        self.common(fn, (torch.randn(8, 6, device=GPU_TYPE, dtype=torch.float16),))
+
+        def fn(x):
+            # Permute: (A, B, C) -> (C, A, B)
+            permuted = x.permute(2, 0, 1)
+            return (permuted.argmin(), permuted.argmax())
+
+        self.common(fn, (torch.randn(4, 6, 8, device=GPU_TYPE),))
```
shunting314, line 14331: To be complete, can we also test for the case where the layout of the tensor contains 'gap', e.g.:
```
torch.randn(10, 20)[:, :10]
```

torch/_inductor/lowering.py
karthickai: thanks for the catch! I've added `isinstance(x.data, ir.StorageBox)` to handle column major tensors passed as function input.

test/inductor/test_torchinductor.py
karthickai: I've added a test case for sliced tensors with gaps in memory

torch/_inductor/lowering.py
karthickai: also added a test case to verify column-major inputs

torch/_inductor/lowering.py
```diff
@@ -6089,6 +6091,24 @@ def _make_reduction_inner(x, *, axis, keepdims, dtype, override_return_dtype):
...
+    if (
+        reduction_type in ("argmax", "argmin")
+        and len(reduced_sizes) > 1
+        and is_triton(x)
+    ):
+        if isinstance(x.data, PermuteView):
+            should_compute_logical_index = True
+        elif isinstance(x.data, ir.ReinterpretView) or (
+            isinstance(x.data, ir.StorageBox)
+            and isinstance(x.data.data, ir.InputBuffer)
```
shunting314: hmm, it works for graph input now but still does not work for internally generated buffers  (e.g. ComputedBuffer).

Here are a few choices:
1. either change `isinstance(x.data.data, ir.InputBuffer)` to `isinstance(x.data.data, ir.Buffer)`
2. or always use logical index disregarding the layout.  Logical index should still work even if the input for argmin/max is contiguous right?

karthickai: you're correct. I've changed it to `ir.Buffer`. Yes, using logical index for contiguous will work, but I feel it's less efficient.

shunting314 (approved):

PR #165760, opened 2025-10-17
[FlexAttention][CUDA] Add flex configs for Blackwell

Aidyn-A:
This PR fixes ULFs on `max_autotune` mode for high head-dim sizes on B200. Closes https://github.com/pytorch/torchtitan/issues/1791

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Chillee @drisspg @yanboliang @BoyuanFeng

PR #165990, opened 2025-10-21
[CUDA] Add torch._C._cudart.cudaGetLastError() method

Aidyn-A:
Just a useful enhancement, since there are bindings for `cudaError_t` and `cudaGetErrorString`.

cc @ptrblck @msaroufim @eqy @jerryzh168 @tinglvv @nWEIdia

### DIFF
--- a/torch/csrc/cuda/shared/cudart.cpp
+++ b/torch/csrc/cuda/shared/cudart.cpp
@@ -120,6 +120,14 @@ void initCudartBindings(PyObject* module) {
         C10_CUDA_CHECK(cudaMemGetInfo(&device_free, &device_total));
         return {device_free, device_total};
       });
+  cudart.def(
+      "cuda"
+      "GetLastError",
+      []() -> cudaError_t {
+        py::gil_scoped_release no_gil;
+        // NOLINTNEXTLINE(performance-no-int-to-ptr)
+        return C10_CUDA_ERROR_HANDLED(cudaGetLastError());
+      });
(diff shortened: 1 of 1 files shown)


syed-ahmed (commented):
Thanks for the adding the binding @Aidyn-A ! Could you please also add an integration which shows the usage of this? For instance, in this PR: https://github.com/pytorch/pytorch/pull/165760, I understand ULF can't be cleared by cudaGetLastError, but can we clear the OOM using cudaGetLastError before it becomes the ULF?

syed-ahmed:
> > > Thanks for the adding the binding @Aidyn-A ! Could you please also add an integration which shows the usage of this? For instance, in this PR: #165760, I understand ULF can't be cleared by cudaGetLastError, but can we clear the OOM using cudaGetLastError before it becomes the ULF?
> > 
> > 
> > Yes, I believe we can clear OOM. Lemme double check whether the behavior is indeed "N-OOMs --> ULF".
> 
> Correction: OOM check is done prior to the kernel call, so the `cudaGetLastError` will not help.
> 
> https://github.com/pytorch/pytorch/blob/d97f6550a25492c747d341feefa181bf2ac7e416/torch/csrc/inductor/static_cuda_launcher.cpp#L109-L124

That's interesting! Is it possible to find what from those lines is the ULF? Instead of cudaGetLastError are we then missing a cuModuleUnload when abandoning the launch of a kernel after loading?

PR #159157, opened 2025-07-25
Add host protoc script back

albanD:
Following comment from https://github.com/pytorch/pytorch/pull/158475#issuecomment-3116518904

Also this is a fake issue as protoc is dead anyways: https://github.com/pytorch/pytorch/issues/159156

Also also, macos cross compilation is not something that is tested :/ But I guess we're ok with that given how niche it it...

PR #165991, opened 2025-10-21
Bring back the executable bit on build_host_protoc.sh

blawrence-ont:
When this was brought back in #159157 it didn't have the executable bit set like it was previously. This caused macOS builds to fail if `CMAKE_OSX_ARCHITECTURES` was set, even if it's set to the same arch as the host.


blawrence-ont:
@albanD for visibility

albanD (approved):
Thanks for the follow up!

Skylion007 (approved):

PR #165993, opened 2025-10-21
Fix triu_/tril_ overlap handling

malfet:
## Summary
- add a shared helper to check triu/tril memory overlap
- guard the CPU and CUDA implementations against overlapping in-place writes
- add a regression test covering in-place triu_/tril_ on expanded tensors

## Testing
- not run (PyTorch build unavailable in sandbox)

------
https://chatgpt.com/codex/tasks/task_e_68f797fd9a38832799f271526881d0f1

### DIFF
--- a/aten/src/ATen/native/TriangularOpsUtils.h
+++ b/aten/src/ATen/native/TriangularOpsUtils.h
@@ -1,3 +1,4 @@
+#include <ATen/MemoryOverlap.h>
 #include <ATen/core/Tensor.h>
 #include <ATen/native/LinearAlgebraUtils.h>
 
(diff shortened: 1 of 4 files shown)


aten/src/ATen/native/TriangularOpsUtils.h
```diff
@@ -54,4 +55,13 @@ static inline std::tuple<bool, Tensor> checkTrilTriuBatchContiguous(const Tensor
   return std::make_tuple(true, tensor);
 }
 
+static inline void checkTrilTriuMemoryOverlap(const Tensor& result, const Tensor& self) {
+  if (result.is_same(self)) {
+    at::assert_no_internal_overlap(result);
+  } else {
```
nikitaved, line 61: Looks rather conservative -- having overlaps in the batch dims should be fine. Maybe something along the lines of `(stride[-1] >= 1 && stride[-2] >= stride[-1] * size[-1]) || (stride[-2] >= 1 && stride[-1] >= stride[-2] * size[-2])` is enough?

PR #165995, opened 2025-10-21
[dirsync] Switch to top-level `xplat/third-party/pthreadpool`

mzlee:
Summary: `fbcode//xplat/third-party/pthreadpool:` just redirects to the xplat version. Switch to the real location

Test Plan: This should be a no-op, so CI?

Differential Revision: D83999534


bigfootjon (approved):

Skylion007 (approved):

PR #165996, opened 2025-10-21
Mark FlexAttentionBackward as cacheable

jamesjwu:
* #166002
* __->__ #165996

This probably should have been marked cacheable a long time ago, no reason that it isn't. 

Test Plan: 
New regional inductor tests for test_flex_attention now are serializable. 


oulgen (approved):

zou3519 (approved):

drisspg (approved):

test/inductor/test_flex_attention.py
```diff
@@ -6618,7 +6618,7 @@ def bias_func(score, b, h, q_idx, kv_idx):
         )
         # Error in backwards
         with self.assertRaisesRegex(
-            torch._inductor.exc.LoweringException,
+            torch._inductor.exc.InductorError,
```
jamesjwu, line 6621: NB: Before, we would throw a cache bypass here instead of a cache miss. For some reason, in the cache bypass branch of our FXGraphCache setup, we wouldn't wrap a resulting exception in fx_codegen_and_compile in a `InductorError`, whereas we would for the cache miss and cache hit paths. 

I fixed `compile_fx` to be consistent about this below, and also updated this test so that we expect an InductorError (which wraps around the existing LoweringException).

PR #166000, opened 2025-10-21
[WIP] remove unreachable `TORCH_USE_CUDA_DSA` code

eqy:
currently a pytorch source build with `TORCH_USE_CUDA_DSA` or even `PYTORCH_USE_CUDA_DSA` doesn't seem to fire any of the macros?

cc @ptrblck @msaroufim @jerryzh168 @tinglvv @nWEIdia


c10/cuda/CUDADeviceAssertionHost.cpp
```diff
@@ -207,7 +74,7 @@ bool CUDAKernelLaunchRegistry::check_env_for_enable_launch_stacktracing()
 }
 
 bool CUDAKernelLaunchRegistry::check_env_for_dsa_enabled() const {
```
Aidyn-A, line 76: Would removing this function be BC-breaking? (It doesn't seem to be used anywhere)

PR #166002, opened 2025-10-21
Add GraphModule.recompile_submodules, use for regional inductor

jamesjwu:
* __->__ #166002
* #165996

This makes it so that `GraphModule.recompile()` will also recompile any submodules that are also graph modules, which allows us to pass all existing regional inductor tests without skipping. 


torch/fx/graph_module.py
```diff
@@ -848,6 +848,17 @@ def call_wrapped(self, *args, **kwargs):
 
         return python_code
 
+    def recompile_submodules(self) -> list[tuple[str, PythonCode]]:
```
anijain2305: I think this should be done automatically for `recompile` method

anijain2305: Something similar here - https://github.com/pytorch/pytorch/pull/152772/files

jamesjwu: @anijain2305 Is it dangerous to add this to the recompile method? I see that the recompile method has a BC flag on it. I feel like it shouldn't matter, but just making sure

oulgen (approved):

PR #166008, opened 2025-10-21
[inductor] add in-kernel nan-check

shunting314:
* __->__ #166008


torch/_inductor/config.py
```diff
@@ -2055,6 +2055,7 @@ class test_configs:
 
     runtime_triton_dtype_assert = False
     runtime_triton_shape_assert = False
+    runtime_triton_nan_assert = False
```
eellison: let's move this next to the other nan checker, since these checks are related to internal, inductor invariants, whereas the nan checker we might see users trigger more

eellison (approved):

PR #166010, opened 2025-10-21
[For discussion][DeviceMesh] Use a shared_state to cache pg per layout, root_mesh and rank_map

fduwjj:
* __->__ #166010
* #167753
* #167590
* #167581

To avoid creating extra PGs when not needed, (for example when calling unflatten many times and some dims share the same layout), we want to create a pg cache mechanism. We cache a PG by a pair of layout (_MeshLayout) and its pg_option, so if users flatten or unflatten into the same layout+pg option (by default pg_option will be None if not set, that way the cache key falls back to layout itself) we will not create process group if already created.

Also to further consolidate the bookkeeping of DM, we created a device_type, shared_state to store root_mesh, rank_map and pg caches. This way all shared info for a given device mesh universe now becomes a singleton.

cc @H-Huang @awgu @wanchaol @fegin @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @aditvenk


torch/distributed/device_mesh.py
```diff
@@ -123,6 +123,44 @@ def _get_device_handle(device_type: str = "cuda"):
...
+        def __init__(
+            self, rank_map: torch.Tensor, root_mesh: Optional["DeviceMesh"] = None
+        ) -> None:
+            self._rank_map = rank_map
+            self._root_mesh = root_mesh
+            self._backend_cache: dict[_MeshLayout, str] = {}
+            self._lock = threading.Lock()
+
+        def get_rank_map(self) -> torch.Tensor:
+            with self._lock:
```
wconstab: I have not reviewed the earlier PRs in this stack, so my questions might be dumb.

I'm wondering what our goal is with the lock.

If we're supporting 'multiple python threads', then I would expect python's GIL would perform the same thing as the lock.  (Perhaps you were thinking forward to GIL-free python?)

On the other hand, if we lock and unlock just around returning the _rank_map or _root_mesh object, we only ensure that threads do not overwrite each other by assigning a new object to the shared state.  We do not protect the case where 2 threads sequentially retrieve the 'root_mesh' and then go on to mutate it differently.  Shouldn't we also protect this case if we're trying to be thread safe?  (What is the goal actually?)

torch/distributed/device_mesh.py
```diff
@@ -123,6 +123,44 @@ def _get_device_handle(device_type: str = "cuda"):
         """
         return getattr(torch, device_type, None)
 
+    class _SharedState:
```
wconstab, line 128: I feel like we need a more intuitive name for this class, to tie it into the design docs a bit better.  Are we using the name "Mesh Universe" still in our docs?  Is a universe defined as all the meshes that share a shared state?  Should we then call this _MeshUniverse?

torch/distributed/device_mesh.py
fduwjj: I think in this case, `_rank_map` is always read-only. We don't expect users to change it. Somehow for `_root_mesh` we only need to set it once if it is still None.

The lock mostly is to mimic the original thread-local situation that we store and update PG using a lock. And it happens to be compatible with GIL-free python as well.

torch/distributed/device_mesh.py
fduwjj: This is the designed proposed by Luca in https://github.com/pytorch/pytorch/pull/163213#discussion_r2387806468 and I think it makes sense. Now this PR is trying to do both: pg caching and shared state. And each universe will share the same state.

torch/distributed/device_mesh.py
```diff
@@ -238,6 +277,11 @@ def __init__(
                     f"but got {len(backend_override)} and {len(self._layout)}."
                 )
 
+            if _shared_state is None:
+                self._shared_state = _SharedState(rank_map, self)
+            else:
+                self._shared_state = _shared_state
```
lw: This branch looks more natural for me if we merge it with the `if mesh is not None` at the very beginning:
- if mesh is not None, we convert it to a `rank_map` and instantiate a SharedState with it
- if mesh is None, `_shared_state` must be given and we use it as-is

If we go down this route, then we can move all the `assert rank_map....` to the constructor of SharedState.

torch/distributed/device_mesh.py
```diff
@@ -174,12 +212,13 @@ class DeviceMesh:
         """
 
         _device_type: str
-        _rank_map: torch.Tensor
+        _rank_map: torch.Tensor  # TODO: remove
         _mesh_dim_names: Optional[tuple[str, ...]]
         _layout: _MeshLayout
-        _root_mesh: Optional["DeviceMesh"] = None
+        _root_mesh: Optional["DeviceMesh"] = None  # TODO: remove
```
lw: Any reason why these aren't removed yet?

torch/distributed/device_mesh.py
```diff
@@ -367,26 +411,39 @@ def _init_process_groups(
...
             ):
-                # Append the default pg to the first dim groups only if the default pg is compatible with `self._device_type`.
-                # Otherwise, create new pg.
-                ranks = list(range(get_world_size()))
-                dim_group = (
-                    new_group(
-                        backend="cpu:gloo,cuda:nccl",
-                        ranks=ranks,
-                        group_desc="mesh_default",
+                backend_cache = shared_state.get_backend_from_cache(layout)
```
lw: I think it would be muuuch cleaner if `_init_process_groups` became a method of SharedState. In order for that to be easy, I believe we could do a preliminary change to how this method works.

Currently, this method is invoked _once_ and, internally, it does a for loop over the sub-layouts of the given layout, and each individual element of `mesh_dim_names` and `backend_override`. I think we should extract the body of the for loop and make it its own separate helper method, with a signature of this kind:
```py
def _init_one_process_group(sub_layout: _MeshLayout, rank_map: torch.Tensor, dim_name: str, backend_override: BackendConfig) -> str:
```

The `if len(layout) == 1 and ...` conditional will have to be rewritten accordingly.

Once this change is done, we can move `_init_one_process_group` to be a method of the SharedState, and the caching will simply become a preamble that looks like this:
```py
if sub_layout in self.cached_pg_names:
    return self.cached_pg_names[sub_layout]
```

Does this make sense? If so, we could do a preliminary PR to extract the for-loop body to a separate method, in order to keep this PR clean.

torch/distributed/device_mesh.py
```diff
@@ -463,11 +520,15 @@ def _init_process_groups(
...
+                            shared_state.update_backend_cache(
+                                layout[dim],
+                                dim_group.group_name,  # type: ignore[union-attr]
+                            )
                             dim_group_names.append(dim_group.group_name)  # type: ignore[union-attr]
             return dim_group_names
 
         def _get_root_mesh(self) -> "DeviceMesh":
-            return self._root_mesh if self._root_mesh else self
+            return not_none(self._shared_state.get_root_mesh())
```
lw: This is the only instance where we're accessing the root mesh of the shared state. In principle, once we introduce the concept of shared state, we can get rid of the concept of root meshes.

What is preventing us from simply removing this `_get_root_mesh` method?

If there's any internal usage of `_get_root_mesh`, could we find ways of codemod it away so that we can fully remove this method?

torch/distributed/device_mesh.py
lw: I'm open to rediscuss naming.

I don't have strong preferences one way or another. Here are some random thoughts to fuel the discussion.

In my mental model I do still oftentimes go back to the idea that a DeviceMesh is morally equivalent to a "tensor" of ranks: it has a number of dimensions, a shape, strides, etc. and it can be sliced, flattened, unflattened, ... just like tensors. When we do this kind of operations on real torch.Tensors, we obtain a new Tensor instance, which however shared the same torch.Storage with the original Tensor. In a sense, this shared state is equivalent to the torch.Storage! It's the underlying "global" store of information, of which each individual DeviceMesh is a (global or partial) view.

With that in mind we could call this `_DeviceMeshStorage`?

torch/distributed/device_mesh.py
```diff
@@ -125,6 +125,44 @@ def _get_device_handle(device_type: str = "cuda"):
...
+        _root_mesh: Optional["DeviceMesh"]
+        _backend_cache: dict[_MeshLayout, str]
+
+        def __init__(
+            self, rank_map: torch.Tensor, root_mesh: Optional["DeviceMesh"] = None
+        ) -> None:
+            self._rank_map = rank_map
+            self._root_mesh = root_mesh
+            self._backend_cache: dict[_MeshLayout, str] = {}
+            self._lock = threading.Lock()
```
lw: Just like @wconstab, I'm also very puzzled by the fact that we need a lock. We've never needed that before, even though we already knew that DeviceMeshes were being accessed by multiple threads (notably: the main thread during forward and the autograd thread pool during backward). And yet, we have never encountered race conditions that motivated introducing locks.

What has changed?

lw (commented):
Thanks!

I think this PR is a great starting point for the discussion around PG caching, and defining a clear mental model for how different DeviceMesh should share state.

torch/distributed/device_mesh.py
fduwjj: I mean we can remove it and see if it works. I don't have a strong opinion on this.

torch/distributed/device_mesh.py
fduwjj: Updated and they are removed.

torch/distributed/device_mesh.py
fduwjj: yes it was merged sorry I just updated it yesterday.

torch/distributed/device_mesh.py
fduwjj: sure, let me try to make a preliminary PR.

torch/distributed/device_mesh.py
lw: Some other ideas for naming.

Another way of seeing this is that the shared state determines when/how PGs are being cached and reused. That is: all the DeviceMeshes that belong to the same shared state will share PGs between themselves, but will not share them at all with DeviceMeshes that belong to different shared states.

In this sense, we could think of shared states as separate "domains" or "realms".

torch/distributed/device_mesh.py
wconstab: well, in this case i'm pretty sure we don't need them.  but in general, its not a good approach to add / remove locks and see if it works- you have to have a design and justify whether you need locks, it should not be guesswork!

torch/distributed/device_mesh.py
fduwjj: We might need more PRs to remove `_get_root_mesh` not in this PR.

torch/distributed/device_mesh.py
fduwjj: I personally prefer the name shared state better because we not only store PG cache, but also rank_map and root_mesh. And it is shared so I kind of like its current name.

test/distributed/test_device_mesh.py
```diff
@@ -1051,6 +1051,26 @@ def test_unflatten_mesh_3d(self):
...
+        # This also turns on pg cache
+        DeviceMesh.decouple_backend_at_save = True
+        # Test unflatten from a dummy world mesh, which is the case we need for Expert Parallelism(EP).
+        global_mesh = init_device_mesh(
+            self.device_type,
+            (8,),
+            mesh_dim_names=("world",),
+        )
+        non_ep_mesh = global_mesh._unflatten(0, (2, 2, 2), ("dp", "cp", "tp"))
+        ep_mesh = global_mesh._unflatten(0, (2, 2, 2), ("dp", "ep", "ep_tp"))
```
kwen2501, line 1066: What's the difference between these two lines vs a user giving multiple names to a dimension? Or a complex name such as "cp_or_ep"?

fduwjj: I think this is to generate a use case where we want to test PG cache. The cache is per layout so when multiple names are assigned to one dimension, the PG will be shared.

PR #165809, opened 2025-10-18
[dynamo, nested graph breaks] disable nested graph breaks in generators; enable nested_graph_breaks in test_ctx_manager.py and test_generator.py

williamwen42:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #163503
* #165597
* #163335
* #166016
* __->__ #165809
* #165808
* #166015
* #166013

Generators should not support nested graph breaks.

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela

PR #166015, opened 2025-10-21
[dynamo, nested graph breaks] remove _dynamo.utils.counter patch on inlined tx'es

williamwen42:
* #163503
* #165597
* #163335
* #166016
* #165809
* #165808
* __->__ #166015
* #166013

This `patch.dict(counters, ...` appears to be ancient code that doesn't really seem to be doing anything? It causes issues in nested graph breaks because the patch cleanup clears out the record of the nested graph break. Removing the patch to see if it's even needed in the first place.


torch/_dynamo/variables/functions.py
```diff
@@ -712,8 +710,7 @@ def next_variable(self, tx):
             # Hierarchically, tx can be seen as the parent of the inline tracer
             # created on call_function. Any exception needs to be propagated to tx
             # for Dynamo to behave correctly
-            with patch.dict(counters, {"unimplemented": counters["inline_call"]}):
```
Lucaskabela, line 716: Can we run this with context manager tests? Seems that was the last time this patch was moved so may be worth double checking those tests.

Presuming those are green (and since nothing showing in CI), we can approve

williamwen42: CI is green now.

Lucaskabela (approved):

PR #166026, opened 2025-10-21
Replace NUMA inheritance approach

pdesupinski:
# Context
Previously, we would modify the parent process's NUMA bindings in order to force child process to inherit them.

However, this would not work correctly if `start_method="forkserver"`, because the subprocesses would actually inherit their bindings from the forkserver middleman process. In this case, the inherited affinity would actually be incorrect for all but the first subprocess (because the forkserver process would get created lazily, and hence inherit and then stick with the bindings intended for the first subprocess).

# This PR
* `str` entrypoints: Use `numactl` CLI
* `Callable` entrypoints: Wrap the `Callable` entrypoint and call `os.sched_setaffinity` inside it.

Hopefully this will be the last necessary iteration.

# Test Plan
## Automated
`$ pytest test/test_numa_binding.py`

## Manual
Verified flops/sec and memory locality wins on several different types of jobs
* `Callable` with forkserver
* `str` entrypoint with spawn
* `Callable` entrypoint with spawn

More details in [this doc (Meta-only).](https://docs.google.com/document/d/1vxD-OKYBTT27jbBwtW9iz9g0tNM0u-i0tiTJg_ieQA8/edit?tab=t.scjv58yswi64)

# Later PR
Update all the documentation when we're confident this has stabilized.

### DIFF
--- a/torch/numa/binding.py
+++ b/torch/numa/binding.py
@@ -1,12 +1,13 @@
 import os
+import shutil
 import traceback
 from collections import defaultdict
-from collections.abc import Callable, Iterable, Iterator
-from contextlib import contextmanager
+from collections.abc import Callable, Iterable
 from dataclasses import asdict, dataclass
 from enum import Enum
+from functools import wraps
 from logging import getLogger
-from typing import Optional, TypeVar
+from typing import Optional, ParamSpec, TypeVar
 
 import torch
 from torch._utils_internal import signpost_event
@@ -14,7 +15,8 @@ from torch._utils_internal import signpost_event
 
 __all__ = [
     "AffinityMode",
-    "maybe_temporarily_apply_numa_binding_to_current_thread",
+    "maybe_wrap_command_args_with_numa_binding",
+    "maybe_wrap_with_numa_binding",
     "NumaOptions",
 ]
 
@@ -39,7 +41,7 @@ class NumaOptions:
 
     """
     If true, we will fall back to using the original command/entrypoint if we fail to compute
-    or apply NUMA bindings.
+    NUMA bindings.
 
     You should avoid using this option! It is only intended as a safety mechanism for facilitating
     mass rollouts of numa binding.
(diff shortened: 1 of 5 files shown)


torch/numa/binding.py
```diff
@@ -39,65 +41,128 @@ class NumaOptions:
...
+        _maybe_apply_numa_binding_to_current_thread(
+            gpu_index=gpu_index,
+            numa_options=numa_options,
+        )
+        return func(*args, **kwargs)
+
+    return wrapped
+
+
+def _maybe_apply_numa_binding_to_current_thread(
```
d4l3k: can we rename this since below it does "_bind_all_threads_in_current_process"? rather than just current_thread?

torch/numa/binding.py
```diff
@@ -39,65 +41,128 @@ class NumaOptions:
...
+    except Exception:
+        _handle_exception(numa_options=numa_options, logger_kwargs=kwargs)
+        return command_args
+
+
+_TParams = ParamSpec("_TParams")
+_TReturn = TypeVar("_TReturn")
+
+
+def maybe_with_numa_binding(
```
d4l3k: `maybe_wrap`? to distinguish between apply vs wrap

torch/numa/binding.py
```diff
@@ -39,65 +41,128 @@ class NumaOptions:
...
     mass rollouts of numa binding.
     """
     should_fall_back_if_binding_fails: bool = False
 
 
-@contextmanager
-def maybe_temporarily_apply_numa_binding_to_current_thread(
-    *, gpu_index: int, numa_options: Optional[NumaOptions]
-) -> Iterator[None]:
+def maybe_apply_numa_binding_to_command_args(
```
d4l3k: this should also probably be wrap rather than apply?

d4l3k (approved):
LGTM

Issue #166033, opened 2025-10-21
[dynamo] Key error in bytecode_transformation


williamwen42:
```python
        flag = True
        dummy = lambda: None

        def fn(x):
            x = x + 1
            torch._dynamo.graph_break()
            x = x + 2
            if flag:
                dummy.attr0 = x
            else:
                with torch.no_grad():
                    dummy.attr1 = x
            return x + 4

        inp = torch.ones(3)
        opt_fn = torch.compile(fn, backend="eager")
        assert torch.allclose(fn(inp), opt_fn(inp))
        flag = False
        assert torch.allclose(fn(inp), opt_fn(inp))
```

results in this KeyError:

```
  File "/data/users/williamwen/pytorch4/test/dynamo/test_repros.py", line 7306, in test_311_resume_block_keyerror2
    self.assertEqual(fn(inp), opt_fn(inp))
                              ^^^^^^^^^^^
  File "/data/users/williamwen/pytorch4/torch/_dynamo/eval_frame.py", line 912, in compile_wrapper
    return fn(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^
  File "/data/users/williamwen/pytorch4/test/dynamo/test_repros.py", line 7293, in fn
    torch._dynamo.graph_break()
  File "/data/users/williamwen/pytorch4/torch/_dynamo/convert_frame.py", line 2034, in __call__
    result = self._torchdynamo_orig_backend(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/users/williamwen/pytorch4/torch/_dynamo/convert_frame.py", line 1784, in __call__
    result = self._inner_convert(
             ^^^^^^^^^^^^^^^^^^^^
  File "/data/users/williamwen/pytorch4/torch/_dynamo/convert_frame.py", line 701, in __call__
    result = _compile(
             ^^^^^^^^^
  File "/data/users/williamwen/pytorch4/torch/_dynamo/convert_frame.py", line 1654, in _compile
    raise InternalTorchDynamoError(
  File "/data/users/williamwen/pytorch4/torch/_dynamo/convert_frame.py", line 1591, in _compile
    guarded_code, tracer_output = compile_inner(code, one_graph, hooks)
                                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/users/williamwen/pytorch4/torch/_utils_internal.py", line 98, in wrapper_function
[...]

PR #166036, opened 2025-10-21
[dynamo] fix store attr graph break in with block

williamwen42:
* #166040
* __->__ #166036

Fixes https://github.com/pytorch/pytorch/issues/166033

Differential Revision: [D85198055](https://our.internmc.facebook.com/intern/diff/D85198055)

### DIFF
--- a/torch/_dynamo/symbolic_convert.py
+++ b/torch/_dynamo/symbolic_convert.py
@@ -2638,7 +2638,9 @@ class InstructionTranslatorBase(
             reason=GraphCompileReason("store_attr", [self.frame_summary()]),
             stack_pops=2,
         )
-        self.output.add_output_instructions([copy.copy(inst)])
+        inst_copy = copy.copy(inst)
+        inst_copy.exn_tab_entry = None
+        self.output.add_output_instructions([inst_copy])
         self.popn(2)
         self.output.add_output_instructions(
             self.codegen_fix_leaf_stack(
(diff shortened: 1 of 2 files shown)


williamwen42:
@williamwen42 has imported this pull request.  If you are a Meta employee, you can view this diff [on Phabricator](https://www.internalfb.com/diff/D85198055).

torch/_dynamo/symbolic_convert.py
```diff
@@ -2638,7 +2638,9 @@ def store_attr_graph_break(self, inst: Instruction) -> None:
             reason=GraphCompileReason("store_attr", [self.frame_summary()]),
             stack_pops=2,
         )
-        self.output.add_output_instructions([copy.copy(inst)])
+        inst_copy = copy.copy(inst)
+        inst_copy.exn_tab_entry = None
```
Lucaskabela, line 2642: So this is the root fix (need to set exn_tab_entry to None) - can you provide a bit more context for me as to why this fixes the error? I am assuming it is because when this is set bytecode_transformation assumes certain things about the instr which isn't true

williamwen42: Before, we would add an instruction to the output code with an invalid `exn_tab_entry` (since it refers to the old set of instructions). This correction is done elsewhere in symbolic_convert.py - i.e. https://github.com/pytorch/pytorch/blob/a988510c3396a668e92801c61b93f7b06219ad0e/torch/_dynamo/symbolic_convert.py#L1019

Lucaskabela (approved):
Thanks for elaborating, lgtm

Lucaskabela:
I will try to get this cherry-pick request tomorrow

PR #166049, opened 2025-10-22
[Pytorch] Enable autovec on aarch64 for type conversion

Nicoshev:
Summary:
Implementing autovec template for type conversions on aarch64-NEON

Generated code can be seen here: https://godbolt.org/z/1K6T1d9TE

We've seen significant performance improvements for converting to and from bytes, compiling using clang with -march=armv9-a+sve2:

Before
float->uint8->float ===> 683.212us
float->int8->float ===> 687.846us
int32->uint8->int32 ===> 497.121us
int32->int8->int32 ===> 481.889us

After:
float->uint8->float ===> 198.204us  ----> 245% higher throughput
float->int8->float ===> 200.241us ----> 244% higher throughput
int32->uint8->int32 ===> 197.970us ----> 151% higher throughput
int32->int8->int32 ===> 198.206us ----> 143% higher throughput

Test Plan:

buck2 test mode/opt //caffe2/test:test_ops
buck2 test mode/opt //caffe2/test:torch

Differential Revision: D85213420

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @aditew01 @snadampal @milpuz01 @nikhil-arm @fadara01

### DIFF
--- a/aten/src/ATen/cpu/vec/vec128/vec128_convert.h
+++ b/aten/src/ATen/cpu/vec/vec128/vec128_convert.h
@@ -5,6 +5,114 @@
 namespace at::vec {
 inline namespace CPU_CAPABILITY {
 #if (defined(__aarch64__) && !defined(CPU_CAPABILITY_SVE256))
+
+// Enable auto-vectorization for GCC-13+ and clang-17+
+// GCC-12 has a bug: gcc.gnu.org/bugzilla/show_bug.cgi?id=117001
+#if __GNUC__ > 12 || (defined(__clang__) && (__clang_major__ >= 17))
+
+template <typename from_type, typename to_type>
+inline void convertImpl(
+    const from_type* __restrict src,
+    to_type* __restrict dst,
+    int64_t n) {
+  uint64_t len = static_cast<uint64_t>(n);
+  for (uint64_t i = 0; i < len; i++) {
+    dst[i] = static_cast<to_type>(src[i]);
+  }
+}
+
+#define CONVERT_TEMPLATE(from_type, to_type)                           \
+  template <>                                                          \
+  inline void convert(const from_type* src, to_type* dst, int64_t n) { \
+    return convertImpl<from_type, to_type>(src, dst, n);               \
+  }
+
+CONVERT_TEMPLATE(uint8_t, uint8_t)
+CONVERT_TEMPLATE(uint8_t, int8_t)
+CONVERT_TEMPLATE(uint8_t, int16_t)
+CONVERT_TEMPLATE(uint8_t, int32_t)
+CONVERT_TEMPLATE(uint8_t, int64_t)
+CONVERT_TEMPLATE(uint8_t, float)
+CONVERT_TEMPLATE(uint8_t, double)
+CONVERT_TEMPLATE(int8_t, uint8_t)
+CONVERT_TEMPLATE(int8_t, int8_t)
+CONVERT_TEMPLATE(int8_t, int16_t)
+CONVERT_TEMPLATE(int8_t, int32_t)
+CONVERT_TEMPLATE(int8_t, int64_t)
+CONVERT_TEMPLATE(int8_t, float)
+CONVERT_TEMPLATE(int8_t, double)
+CONVERT_TEMPLATE(int16_t, uint8_t)
+CONVERT_TEMPLATE(int16_t, int8_t)
+CONVERT_TEMPLATE(int16_t, int16_t)
+CONVERT_TEMPLATE(int16_t, int32_t)
+CONVERT_TEMPLATE(int16_t, int64_t)
+CONVERT_TEMPLATE(int16_t, float)
+CONVERT_TEMPLATE(int16_t, double)
+CONVERT_TEMPLATE(int32_t, uint8_t)
+CONVERT_TEMPLATE(int32_t, int8_t)
+CONVERT_TEMPLATE(int32_t, int16_t)
+CONVERT_TEMPLATE(int32_t, int32_t)
+CONVERT_TEMPLATE(int32_t, int64_t)
+CONVERT_TEMPLATE(int32_t, float)
+CONVERT_TEMPLATE(int32_t, double)
+CONVERT_TEMPLATE(int64_t, uint8_t)
+CONVERT_TEMPLATE(int64_t, int8_t)
+CONVERT_TEMPLATE(int64_t, int16_t)
+CONVERT_TEMPLATE(int64_t, int32_t)
+CONVERT_TEMPLATE(int64_t, int64_t)
+CONVERT_TEMPLATE(int64_t, float)
+CONVERT_TEMPLATE(int64_t, double)
+CONVERT_TEMPLATE(float, uint8_t)
+CONVERT_TEMPLATE(float, int8_t)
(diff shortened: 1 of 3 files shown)


aten/src/ATen/cpu/vec/vec128/vec128_convert.h
```diff
@@ -5,6 +5,117 @@
...
+  }
+}
+
+#define CONVERT_TEMPLATE(from_type, to_type)                           \
+  template <>                                                          \
+  inline void convert(const from_type* src, to_type* dst, int64_t n) { \
+    return convertImpl<from_type, to_type>(src, dst, n);               \
+  }
+
+CONVERT_TEMPLATE(uint8_t, uint8_t)
```
malfet, line 30: Why do you need to implicitly instantiate those conversions? I.e. it looks like you are saying any conversion will be faster using explicit vectorization. If that is the case, why not just define a template a be done with it?

aten/src/ATen/cpu/vec/vec128/vec128_convert.h
```diff
@@ -5,6 +5,117 @@
 namespace at::vec {
 inline namespace CPU_CAPABILITY {
 #if (defined(__aarch64__) && !defined(CPU_CAPABILITY_SVE256))
+
+// Enable auto-vectorization for GCC-13+ and clang
+// GCC-12 has a bug: gcc.gnu.org/bugzilla/show_bug.cgi?id=117001
+#if __GNUC__ > 12 || defined(__clang__)
```
malfet: Shouldn't this code be enabled if and only if auto-vectorization is enabled?
Also, are you sure this will work with any version of clang? Some [godbolt](https://godbolt.org/) example of the generated code in PR description would be good

aten/src/ATen/cpu/vec/vec128/vec128_convert.h
Nicoshev: @malfet There is already a template, function "convert" is a template which we are overriding.
For some reason autovec was not occurring on the generic convert template, so we are switching to explicit instantiations.
The case where from/to are the same type is actually covered by other platforms for some data types, so I thought convenient to leave it here too.

aten/src/ATen/cpu/vec/vec128/vec128_convert.h
Nicoshev: @malfet If auto-vectorization is not enabled, this code will still be as good as it can be. This is because we have no alternative conversion routines using explicit SIMD intrinsics.
You can see the clean code generated here: https://godbolt.org/z/1K6T1d9TE
It should work well with any version of clang. For older versions some functions may just not vectorize.

aten/src/ATen/cpu/vec/vec128/vec128_convert.h
MatzeB: By my reading`convertImpl` mainly does two things:
- Adding `restrict` qualifiers to the arguments. Let's hope this is always true. Though the old manually vectorized code made the same assumption, so I guess that's not really new here...

- The `memcpy` seem superfluous to me here. Seems to be a pattern to avoid strict aliasing problems, but I don't see any indication here that we have to worry about that...

MatzeB: Generally this code is simpler/easier and leaves the compiler more freedom to vectorize than the previous manually vectorized code. So it's mostly a task of proving with benchmarks that the vectorization is happening? Code looks nicer in my eyes (disregarding the `memcpy` stuff)

Nicoshev:
@malfet 
> @Nicoshev for your godbolt example, you are compiling it for SVE, but replacing the NEON code, so I think it would be fair to see the output generated for vectorized for those and compared to the code generated before

New code: https://godbolt.org/z/oY6s9K5sM
Old code: https://godbolt.org/z/h9n6czqdz

Code size is greatly reduced. This won't have an impact on benchmarks, but it will affect prod positively

aten/src/ATen/cpu/vec/vec128/vec128_convert.h
```diff
@@ -5,6 +5,114 @@
...
+template <typename from_type, typename to_type>
+inline void convertImpl(
+    const from_type* __restrict src,
+    to_type* __restrict dst,
+    int64_t n) {
+  uint64_t len = static_cast<uint64_t>(n);
+  for (uint64_t i = 0; i < len; i++) {
+    dst[i] = static_cast<to_type>(src[i]);
+  }
+}
```
malfet, line 22: This is exactly identical to the default implementation of convert template, isn't it?

aten/src/ATen/cpu/vec/vec128/vec128_convert.h
```diff
@@ -5,6 +5,114 @@
...
+CONVERT_TEMPLATE(int8_t, int64_t)
+CONVERT_TEMPLATE(int8_t, float)
+CONVERT_TEMPLATE(int8_t, double)
+CONVERT_TEMPLATE(int16_t, uint8_t)
+CONVERT_TEMPLATE(int16_t, int8_t)
+CONVERT_TEMPLATE(int16_t, int16_t)
+CONVERT_TEMPLATE(int16_t, int32_t)
+CONVERT_TEMPLATE(int16_t, int64_t)
+CONVERT_TEMPLATE(int16_t, float)
+CONVERT_TEMPLATE(int16_t, double)
```
malfet, line 50: It's very easy to make mistake in this endless repetition, any way it could be refactored further?  (for example, it looks like you've missed float16<->bfloat16 conversions in this list don't you?)

malfet (changes requested):
PR title claims throughput improvements, but does not update any benchmark files. 

Moreover, `convertImpl` feels just like a code duplicate of default template implementation already, isn't it?

aten/src/ATen/cpu/vec/vec128/vec128_convert.h
malfet: How this new `convertImpl` are different from vanialla implementation of `convert` which should do the same scalar operation, isn't it?

aten/src/ATen/cpu/vec/vec128/vec128_convert.h
Nicoshev: @malfet float16<->bfloat16 cannot be done, there is no static_cast between these types

aten/src/ATen/cpu/vec/vec128/vec128_convert.h
Nicoshev: @malfet no, it adds the restrict qualifiers and casts the count as uint64

aten/src/ATen/cpu/vec/vec128/vec128_convert.h
Nicoshev: @malfet  it adds the restrict qualifiers and casts the count as uint64

ezyang (approved):

mcfi (approved):

aditew01 (approved):

Nicoshev:
@ezyang Issue got fixed in https://github.com/pytorch/pytorch/pull/166739

PR #166058, opened 2025-10-22
Skip test_intra_node_comm_all_reduce for CC < 9.0

Flamefire:
The test uses a multicast operation in SymmetricMemory that requires CUDA CC 9.0 and CUDA RT version 12.1

Failure:

```
[rank1]:E1022 09:55:08.823000 3580472 torch/testing/_internal/common_distributed.py:721] RuntimeError: CUDA error: device-side assert triggered...
[rank1]:E1022 09:55:08.823000 3580472 torch/testing/_internal/common_distributed.py:721]  exiting process 1 with exit code: 10
...
:318: st_vec: block: [0,0,0], thread: [87,0,0] Assertion `false` failed.
/pytorch-v2.7.1/torch/csrc/distributed/c10d/CUDASymmetricMemory-inl.h:318: st_vec: block: [0,0,0], thread: [88,0,0] Assertion `false` failed.

```

### DIFF
--- a/test/distributed/test_c10d_nccl.py
+++ b/test/distributed/test_c10d_nccl.py
@@ -4630,7 +4630,7 @@ class CommTest(test_c10d_common.AbstractCommTest, MultiProcessTestCase):
     )
     def test_intra_node_comm_all_reduce(self, custom_group_name):
         from torch._C._distributed_c10d import _get_intra_node_comm_usage_counter
-        from torch.testing._internal.common_cuda import SM80OrLater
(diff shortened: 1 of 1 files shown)


kwen2501 (commented):
TBH I am not sure if the SymmMem implementation of all-reduce uses multimem operations.
From the code I read, the intra_node_comm allreduce dispatches to either one-shot or two-shot in SymmMem. Neither of them uses multimem instructions; they just perform per-peer load and store.
I think Ampere with NVSwitch should be able to support such.

PR #166065, opened 2025-10-22
[Dynamo] Support for xor

### LINKED ISSUE #146688: [Dynamo] Support for more binary ops
### 🐛 Describe the bug

Below example gives graph break `torch._dynamo.exc.Unsupported: builtin: and_ [<class 'torch._dynamo.variables.user_defined.UserDefinedObjectVariable'>, <class 'torch._dynamo.variables.constant.ConstantVariable'>] False`

Looks like intended for now and it can be supported with little effort by adding entry for `fns` in `BuiltinVariable::_binops` function at `torch/_dynamo/variables/builtin.py`
`operator.and_: (["__and__", "__rand__", "__iand__"], operator.and_),` and it's alternatives for other binary operators.

There's such comment added 2 years ago - is it still valid blocking dynamo from supporting other binary operators? 
Maybe it can support other cases and keep it disabled for dynamic shapes only?
```
# NB: The follow binary operators are not supported for now, since the
# corresponding magic methods aren't defined on SymInt / SymFloat:
# operator.matmul
# divmod
# operator.and_
# operator.or_
# operator.xor
```

```python
import torch


class DummyClass:
    def __init__(self, value):
        self.value = value

    def __and__(self, other):
        return self.value & int(other)

def test_fn():
    v = DummyClass(1)
    return (v & 1)


compiled_test_fn = torch.compile(test_fn)
compiled_test_fn()
```

### Versions

PT 2.6

jmaczan:
Add missing support for xor (and maybe some other binary ops later on)

Fixes #146688

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @aditew01 @ezyang @EikanWang @wenzhe-nrv @voznesenskym @penguinwu @Guobing-Chen @zhuhaozhe @blzheng @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela


Lucaskabela (dismissed):
This generally looks good  but we should add some unit test coverage here and include in summary as testplan :)

ezyang (changes requested):
The ValueRange logic seems to have exactly been copied from bitwise_or which seems unlikely.

torch/_dynamo/variables/builtin.py
```diff
@@ -2779,7 +2779,31 @@ def _comparison_with_symnode(self, tx: "InstructionTranslator", left, right):
...
+
+        if isinstance(
+            a,
+            (
+                ConstDictVariable,
+                DictKeysVariable,
+                MutableMappingVariable,
+                SetVariable,
+                UserDefinedDictVariable,
+                UserDefinedObjectVariable,
```
ezyang: This seems like a behavior difference

torch/utils/_sympy/value_ranges.py
```diff
@@ -513,6 +513,10 @@ def or_(a, b):
     def and_(a, b):
         return ValueRanges.coordinatewise_increasing_map(a, b, sympy.And)
 
+    @staticmethod
+    def xor_(a, b):
+        return ValueRanges.coordinatewise_increasing_map(a, b, sympy.Xor)
```
ezyang: Pretty sure this is dead?  Also, this is not coordinatewise increasing: 0 ^ 1 = 1, but increasing first coord to 1 gives 1 ^ 1 = 0

torch/utils/_sympy/value_ranges.py
```diff
@@ -566,6 +570,30 @@ def bitwise_or(cls, a, b):
...
+            b = cls._bool_to_int(b)
+
+        try:
+            lower_lower = a.lower ^ b.lower
+            lower_upper = a.lower ^ b.upper
+            upper_lower = a.upper ^ b.lower
+            upper_upper = a.upper ^ b.upper
+
+            lower = min(lower_lower, lower_upper, upper_lower, upper_upper)
+            upper = max(lower_lower, lower_upper, upper_lower, upper_upper)
```
ezyang: This seems unlikely to be right. Please reason through this ChatGPT review:

----

Counterexample:
	•	Let a ∈ [8, 15], b ∈ [8, 15].
	•	Your code checks: 8^8=0, 8^15=7, 15^8=7, 15^15=0 → returns range [0, 7].
	•	But pick interior values: 10 ^ 5 = 15. The true max is at least 15, so [0, 7] is wrong (too tight).

Other issues:
	•	With negatives, Python’s unbounded two’s-complement semantics make “bit ranges” tricky; the try/except won’t throw, so you’ll quietly produce misleading bounds.
	•	Even for non-negative ranges, the minimum isn’t determined by corners either (unless some special structure holds).

ezyang (changes requested):
.

torch/_dynamo/variables/builtin.py
jmaczan: Correct, I have reverted that. I have used `call_or` as a reference, but seems like I shouldn't

torch/utils/_sympy/value_ranges.py
jmaczan: Removed. Instead, in `bitwise_xor` I directly compute lower and upper when `a` and `b` are booleans. I use similar code to what I previously assumed I can use for general case for `ints`, but here I think it's safe to use

Initially, I didn't dig deep enough when coded it and used `or_` as a source of truth and tbh this was just incorrect. `xor`, unlike `or` and `and` is less (not at all?) monotonic, and from what I checked `coordinatewise_increasing_map` assumes that next values always increase, which is not always a case, as you pointed out

torch/utils/_sympy/value_ranges.py
jmaczan: @ezyang I have underestimated a difficulty of xor. Now I always use the -inf,inf range, unless there are two booleans or both ranges have same lower and upper, respectively (and aren't infinites). As you mentioned some time ago, it's fine to have too wide bound here, so I think that's ok now. At least it's sound

To optimize the ranges further is a bit rabbit hole, and it seems to be possible but chasing it led me to things like Known Bits Analysis in LLVM (https://llvm.org/docs/GlobalISel/KnownBits.html), which sounds exciting, but it's too big and too difficult for me to implement in this PR

Where I can add tests for `bitwise_xor`? I didn't find a related test file, can I create a new one?

ezyang: You didn't update the bool handling?

Look at test/test_sympy_utils.py

jmaczan: I've added `"bitwise_xor"` to be tested in `test/test_sympy_utils.py` if that's what you're asking about. If that's all we need, then great and pls rereview

@ezyang

jmaczan:
Ready for a rereview

torch/utils/_sympy/value_ranges.py
```diff
@@ -111,6 +111,10 @@ def vr_is_expr(vr: ValueRanges[_T]) -> TypeGuard[ValueRanges[sympy.Expr]]:
     return not vr.is_bool
 
 
+def is_sympy_integer(value) -> TypeGuard[sympy.Integer]:
```
Skylion007: Actually should be TypeIs instead of TypeGuard here, we have a stronger invariant.

jmaczan: @Skylion007  Updated, thanks

ezyang (approved):

PR #166073, opened 2025-10-22
[Inductor][Autotune] Gracefully restart the autotune process after ULF failure

### LINKED ISSUE #1791: Deepseekv3 CUDA illegal memory access on B200 from increasing seq_len to 8192
### Bug description

Using B200 devgpu

## Repro 

- fsdp=4, ep=4, AC=none, compile=False

- Using debug model with DSV3 671b config in single node testing via setting `n_layers=2` and `n_dense_layers=1` to avoid OOM.
    - Is this not a valid config? Could this explain the error? 

```python
    "debugmodel": DeepSeekV3ModelArgs(
        vocab_size=129280,
        dim=7168,
        inter_dim=18432,
        moe_inter_dim=2048,
        n_layers=2,
        n_dense_layers=1,
        n_heads=128,
        moe_args=MoEArgs(
            num_experts=256,
            num_shared_experts=1,
            top_k=8,
            score_func="sigmoid",
            route_norm=True,
            route_scale=2.5,
            score_before_experts=False,
        ),
        n_expert_groups=8,
        n_limited_groups=4,
        q_lora_rank=1536,
        kv_lora_rank=512,
        qk_nope_head_dim=128,
        qk_rope_head_dim=64,
        v_head_dim=128,
        use_flex_attn=True,
        attn_mask_type="block_causal",
    ),
```

Default seq len (2048) works:

```
CUDA_VISIBLE_DEVICES="4,5,6,7" NGPU=4 CONFIG_FILE="./torchtitan/models/deepseek_v3/train_configs/debug_model.toml" ./run_train.sh --training.steps=200 --parallelism.data_parallel_shard_degree=4 --parallelism.expert_parallel_degree=4 --parallelism.tensor_parallel_degree=1 --activation_checkpoint.mode=none 
```

Increasing seq_len to 8192 -> CUDA illegal memory access:

```
CUDA_VISIBLE_DEVICES="4,5,6,7" NGPU=4 CONFIG_FILE="./torchtitan/models/deepseek_v3/train_configs/debug_model.toml" ./run_train.sh --training.steps=200 --parallelism.data_parallel_shard_degree=4 --parallelism.expert_parallel_degree=4 --parallelism.tensor_parallel_degree=1 --activation_checkpoint.mode=none --training.seq_len=8192
```

### Versions

- torchtitan latest main branch
- torch 10/2 nightly for cuda 12.8: `torch                     2.10.0.dev20251002+cu128  `

Aidyn-A:
This PR partially fixes https://github.com/pytorch/torchtitan/issues/1791, as it will work with `TORCHINDUCTOR_AUTOTUNE_IN_SUBPROC=1` setting only.

The core of the problem: In `max-autotune` mode Inductor runs multiple benchmarks to determine the best config. If one of these benchmarks fails with `cudaErrorLaunchFailure`, all other CUDA calls within the same process will fail including the rest of the benchmarks. 

The solution: Restart the child process gracefully and continue benchmarking.

Unfortunately, if autotuning is done in the main process, the whole program falls into unrecoverable state. In this case, the only way of successful execution would be just preventing the ULF.

Here is some info from [CUDA documentation](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html): 
>cudaErrorLaunchFailure = 719
An exception occurred on the device while executing a kernel. ... . This leaves the process in an inconsistent state and any further CUDA work will return the same error. To continue using CUDA, the process must be terminated and relaunched.

cc @ptrblck @msaroufim @eqy @jerryzh168 @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben

### DIFF
--- a/torch/_inductor/autotune_process.py
+++ b/torch/_inductor/autotune_process.py
@@ -231,6 +231,13 @@ class TuningProcess:
             self.process.kill()
         self.close()
 
+    def restart(self) -> None:
+        """
+        Gracefully restarts the child process.
+        """
+        self.shutdown(wait=True)
+        self.start()
+
 
 class TuningProcessPool:
     """
@@ -311,11 +318,16 @@ class TuningProcessPool:
             )
             # Set to INF so this choice will be ignored
             return float("inf")
-        except Exception:
+        except Exception as process_exception:
             warnings.warn(
                 f"Failed to benchmark choice '{choice}'. It will be ignored. "
                 "Please debug the root cause in case the choice can bring perf gains."
             )
+            # An unspecified launch failure (cudaErrorLaunchFailure) corrupts the
+            # CUDA context, making it unrecoverable. All subsequent CUDA calls will
+            # fail as well. The process must be restarted to restore CUDA functionality.
+            if "cudaErrorLaunchFailure" in str(process_exception):
+                process.restart()
             # Set to INF so this choice will be ignored
             return float("inf")
         finally:
--- a/torch/_inductor/select_algorithm.py
+++ b/torch/_inductor/select_algorithm.py
@@ -3290,7 +3290,7 @@ class AlgorithmSelectorCache(PersistentCache):
                 elif "illegal memory access" in msg:
                     msg += "\n\nEither error in template or triton bug.\n"
                 elif "unspecified launch failure" in msg:
-                    msg += "\n\nAn unspecified launch failure was caught. It is an unrecoverable failure."
+                    msg += "\n\nAn unrecoverable unspecified launch failure was caught during autotuning."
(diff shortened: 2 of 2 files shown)


syed-ahmed (approved):
I am ok with this. But we should figure out how not to get to a ULF scenario - perhaps continue the work in this PR: https://github.com/pytorch/pytorch/pull/165990.

drisspg (approved):
Looks good, cc @eellison  for any other comments if you have them

PR #166076, opened 2025-10-22
Add HIDDEN_NAMESPACE_BEGIN and END macros for hiding header APIs

janeyx99:
Spurred by the conversation started in https://github.com/pytorch/pytorch/issues/163343. 

Context:
* Header implementations may be inlined _but_ are not necessarily inlined, even when using the `inline` keyword.
* When someone wants to use multiple extensions in the same runtime, e,g., with FA3 and AO, then 2 `.so`s are loaded that may have been built with different libtorch versions. Thus, if an API is not inlined and are differently implemented, one implementation will be arbitrarily picked up and used across the runtime, depending on link order. This is bad!
* Consequently, we need to be very good at guaranteeing that we don't modify header implementations within a namespace. This is easy to mess up by accident, which would be a dire mistake.

Solution:
In essence, we want APIs in torch::headeronly and torch::stable to be visible in each individual extension only, and nowhere else. We want to hide these symbols! Thankfully, pybind already solved this problem (thanks @malfet for bringing that to my attention). This PR is heavily inspired by the code in pybind here: https://github.com/pybind/pybind11/blob/e6984c805ec09c0e5f826e3081a32f322a6bfe63/include/pybind11/detail/pybind11_namespace_macros.h#L73-L82.

In this PR, we introduce the macros for defining hidden namespaces in PyTorch.

* #166203
* #166079
* #166078
* #166077
* __->__ #166076


malfet (approved):

swolchok:
> if an API is not inlined and are differently implemented, one implementation will be arbitrarily picked up and used across the runtime, depending on link order. This is bad!

I'm not certain this is true; RTLD_LOCAL should prevent it everywhere, and on Mac-like OSes, I think the two-level namespace might prevent it even if RTLD_GLOBAL is used.

However, I see https://github.com/pybind/pybind11/commit/a859dd67a29372e478ad862d8d2979930471532e is the pybind11 commit that did something similar. It references https://github.com/pybind/pybind11/pull/949, which does cite specific failures of RTLD_LOCAL to prevent symbol conflicts. So, I think this is fine to merge, especially since it should be harmless regardless.

PR #166081, opened 2025-10-22
Enable local tensor model for  DTensor redistribute tests

dzmitry-huba:
* #166105
* __->__ #166081

Redistribute test exercise extensively various sharding schemes and
redistribution between them. These tests uncovered more edge cases
that were not supported by the local tensor primarily different flavors
of uneven sharding. In order to handle these cases this change implements
missing functional collectives and adds support for uneven sharding
case where sharding group (ranks) is larger than the size of the dimension
being sharded. In the latter case the "missing" shards are represented
by zero sized tensors so that the rest of the local tensor machinery
can stay oblivious to this special case.


test/distributed/tensor/test_redistribute.py
```diff
@@ -407,7 +412,7 @@ def test_replicate_to_partial(self):
     def test_partial_to_shard(self, dtype):
         device_mesh = self.build_device_mesh()
         partial_spec = [Partial()]
-        my_rank = device_mesh.get_rank()
+        my_rank = self.rank
```
ezyang, line 418: Arguably this should work without this modification

test/distributed/tensor/test_redistribute.py
```diff
@@ -534,10 +544,11 @@ def test_redistribute_shard_dim_change(self, dtype):
                         1,
                     )
                 else:
-                    self.assertEqual(
-                        comm_mode.get_comm_counts()[funcol.all_gather_into_tensor],
-                        1,
-                    )
+                    if not self.is_local_tensor_enabled:
+                        self.assertEqual(
+                            comm_mode.get_comm_counts()[funcol.all_gather_into_tensor],
```
ezyang, line 553: arguably LocalTensor should update the comm counts correctly; another good task to farm out

torch/distributed/_local_tensor/__init__.py
```diff
@@ -111,7 +112,7 @@ def _map_to_rank_local_val(val: Any, rank: int) -> Any:
     return val
 
 
-def collect_cuda_rng_states() -> list[torch.Tensor]:
+def _collect_cuda_rng_states() -> list[torch.Tensor]:
```
ezyang, line 115: oh oh what's going on here? Potentially BC breaking?

torch/distributed/_local_tensor/_c10d.py
```diff
@@ -134,31 +249,31 @@ def _local_broadcast_(
 
 
 def _local_reduce(
-    reduce_op: ReduceOp,
+    reduce_op: ReduceOp | str,
     tensors: list[torch.Tensor],
 ) -> torch.Tensor:
-    if reduce_op == ReduceOp.SUM:
+    if reduce_op == ReduceOp.SUM or reduce_op == "sum":
```
ezyang, line 259: weird

torch/distributed/_local_tensor/_c10d.py
```diff
@@ -291,8 +406,8 @@ def _local_reduce_scatter_tensor_coalesced_(
             )
 
             # Update all tensors in the group with the reduced result
-            for rank in group_ranks:
-                output_tensor._local_tensors[rank].copy_(reduced_inpit_splits[rank])
+            for i, rank in enumerate(group_ranks):
+                output_tensor._local_tensors[rank].copy_(reduced_inpit_splits[i])
```
ezyang: s/inpit/input/

torch/distributed/_local_tensor/_c10d.py
```diff
@@ -92,6 +97,116 @@ def _prepare_collective_groups(
     return ranks, group_offsets, offset
 
 
+def _local_functional_all_gather_into_tensor(
```
ezyang, line 104: It's probably worth explaining here why we need both this and https://github.com/pytorch/pytorch/blob/gh/dzmitry-huba/9/head/torch/distributed/_local_tensor/_c10d.py#L464

torch/distributed/_local_tensor/__init__.py
```diff
@@ -164,7 +165,51 @@ def _set_rng_state(cpu_state: torch.Tensor, cuda_states: list[torch.Tensor]) ->
     will get their rng state set.
     """
     torch.set_rng_state(cpu_state)
-    set_cuda_rng_states(cuda_states)
+    _set_cuda_rng_states(cuda_states)
+
+
+def _combine_singular_rank_results(rank_results: dict[int, Any]) -> Any:
```
ezyang: two overload signatures probably better for downstream typing (a generic sig and an int to SymInt overload)

torch/distributed/_local_tensor/__init__.py
```diff
@@ -187,40 +232,24 @@ def _for_each_rank_run_func(
...
         rank_ret = func(*rank_args, **rank_kwargs)
         flat_rank_rets[r] = rank_ret
 
-    rr_key = next(iter(flat_rank_rets.keys()))
-    rr_val = flat_rank_rets[rr_key]
+        if default_value is None and func is torch.ops.aten.split.Tensor:
+            # If split happens over the dimension smaller than the number of chunks
+            # it is possible that some ranks will produce shorter lists of chunks.
+            # In order to make the result across all ranks of the same length we
+            # append empty tensors (zero size on the split dimension).
```
ezyang, line 252: lol

torch/distributed/tensor/placement_types.py
```diff
@@ -103,7 +103,7 @@ def _split_tensor(
         pad_sizes: list[int] = []
         for shard in tensor_list:
             if with_padding:
-                pad_size = full_chunk_size - shard.size(self.dim)
+                pad_size = Shard._get_shard_pad_size(full_chunk_size, shard, self.dim)
```
ezyang, line 106: oh huh you already defined this

torch/distributed/tensor/placement_types.py
```diff
@@ -412,23 +402,82 @@ def _to_new_shard_dim(
...
+        old_dim_logical_size = current_logical_shape[old_shard_dim]
+        new_dim_logical_size = current_logical_shape[new_shard_dim]
+
+        old_dim_padding = old_dim_logical_size % num_chunks != 0
+        new_dim_padding = new_dim_logical_size % num_chunks != 0
 
         if old_dim_padding:
+            old_dim_full_chunk_size = (
+                old_dim_logical_size + num_chunks - 1
+            ) // num_chunks
```
ezyang: The code duplication here is unfortunate

ezyang (approved):

test/distributed/tensor/test_redistribute.py
dzmitry-huba: It may work if we will patch DeviceMesh.get_rank() to return all ranks of the mesh. I thought that keeping patching to the minimum is worth it. If we are going to see more use cases like this, I plan to patch DeviceMesh it to support it.

test/distributed/tensor/test_redistribute.py
dzmitry-huba: Good point, added todo to address this in a separate PR

torch/distributed/_local_tensor/__init__.py
dzmitry-huba: No, this is module internal function and is not meant to be used outside of it. This change fixes the naming to reflect that.

torch/distributed/_local_tensor/__init__.py
dzmitry-huba: Lol indeed.

torch/distributed/_local_tensor/_c10d.py
dzmitry-huba: Functional collectives pass reduction operation as a string whereas regular collectives as a ReduceOp. Both functional and regular collectives call into this function to perform reduce over collected tensors.

torch/distributed/_local_tensor/_c10d.py
dzmitry-huba: Renamed

torch/distributed/tensor/placement_types.py
dzmitry-huba: Yes, every test helps to discover missing implementation - the unfortunate nature of runtime checks.

torch/distributed/_local_tensor/__init__.py
dzmitry-huba: Good point, updated

torch/distributed/tensor/placement_types.py
dzmitry-huba: Did rewrite to avoid code duplication

PR #166090, opened 2025-10-22
[MPS] Add `linalg.householder_product` for MPS

### LINKED ISSUE #166089: Support `torch.linalg.householder_product` for MPS
cc @jianyuh @nikitaved @mruberry @walterddr @xwang233 @Lezcano @kulinseth @malfet @DenisVieriu97 @jhavukainen

kurtamohler:
* __->__ #166090

Fixes #166089


aten/src/ATen/native/mps/kernels/LinearAlgebra.metal
```diff
@@ -640,6 +641,179 @@ kernel void applyPivots(
...
+      *H_prod_elem_ptr = H_irc;
+    } else {
+      *H_elem_ptr = H_irc;
+
+      // Need this sync because the below matmul requires all threads to finish
+      // writing their entries to `H_prod` and `H`.
+      threadgroup_barrier(mem_flags::mem_threadgroup);
+
+      T H_prod_0_to_i_rc =
+          calc_matmul_rc(H_prod, H, H_stride_r, H_stride_c, m, r, c);
```
kurtamohler, line 784: At the moment, performance is much worse than that of the CPU impl, except in some cases where the number of batches is greater than the number of `A` matrix elements times `tau` vector elements.

The vast majority of runtime is spent in this matrix multiplication. I'm using a naive implementation of matmul, so we should be able to get much better performance if I change it to a tiled matmul. I suppose it should be possible to just reuse the tiled matmul defined earlier in this file, so I will look into that

aten/src/ATen/native/mps/kernels/LinearAlgebra.metal
```diff
@@ -640,6 +641,179 @@ kernel void applyPivots(
...
+
+template <>
+half2 conj(half2 a) {
+  return half2(a.x, -a.y);
+}
+
+template <>
+float2 conj(float2 a) {
+  return float2(a.x, -a.y);
+}
```
malfet: I wonder if something like that should be moved ot `c10/metal/complex.h` or `c10/metal/utils.h`

malfet (approved):

aten/src/ATen/native/mps/kernels/LinearAlgebra.metal
kurtamohler: I attempted to improve performance ([in this branch](https://github.com/kurtamohler/pytorch/commit/0439c3f5806de48c40ce3bf55f4c95a59b08e8d8)) by changing the kernel to just generate the householder matrices and use the existing `do_metal_bmm` for the matrix multiply. It improved performance slightly in some cases, and decreased in others, but overall it didn't make too much of a difference. Maybe the CPU impl isn't actually doing a series of full matrix multiplies and instead uses some simplified formula. I'll have to take a look at the lapack impl. But I guess this is probably somewhat low priority

kurtamohler:
I will follow up with a performance improvement PR once I understand how to do it

PR #166095, opened 2025-10-22
[PoC and not for landing][Flight Recorder] Added flight recorder NCCL collective visualization

VieEeEw:
Summary:
Usage:
Suppose we have Flight Recorder traces from all ranks stored in a directory "./fr_traces" in the format of "nccl_trace_rank_<rank>".
First, run the process trace python script
```bash
cd path/to/collective_visualizer
python process_fr_trace.py --dump-dir="./fr_traces"
```
We also have some other customized support including fr trace file name pattern, check the help.
```bash
python process_fr_trace.py --help
```
By default the intermediate traces will be dumped under visualize/public/trace.json.
Then we can start the react application to visualize the NCCL collectives.
```
cd visualize
npm install
npm start
```

### DIFF
--- a/tools/flight_recorder/collective_visualizer/process_fr_trace.py
+++ b/tools/flight_recorder/collective_visualizer/process_fr_trace.py
@@ -1,0 +1,203 @@
+#!/usr/bin/env python3
+
+# pyre-strict
+import asyncio
+import json
+import os
+import pickle
+import re
+from argparse import ArgumentParser, Namespace
+from glob import glob
+from tempfile import NamedTemporaryFile
+from typing import Any, Dict, List, Optional, Tuple
+
+STACK_FRAME_FORMAT_STR: str = "{filename}:{line}, in `{name}`"
+
+
+def get_args() -> Namespace:
+    parser = ArgumentParser(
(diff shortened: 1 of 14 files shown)


wang-mask:
hello @VieEeEw , i have a quetion about this feature. When using this tool to analyze and visualize the Flight Recorder data, I noticed that on many ranks, the NCCL op records show a state of "completed" but the `retired` field is 0. This even applies to NCCL ops that have timed out. Could you help explain why this happens? Thank you.

VieEeEw:
@wang-mask I won't be able to help you on specific cases without looking at the code, but assuming you are using the default PyTorch distributed implementation:
1. It is impossible for a completed collective to time out, 
2. It is possible for a completed collective to still be not retired, as the status check happens every 100ms [by default](https://github.com/pytorch/pytorch/blob/a83f2a90820f3e78455aa1060c8285c377ed0bfc/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp#L2284) asynchronously with the collective execution , but I would say it is abnormal if it happened a lot.

Given you stated that you saw NCCL watchdog timeout happened on a completed collective, I had questions:
1. How did you determine which collective, on which rank, at which collective sequence id from which process group was timing out?
2. How did you pick the collective from FR dump to analyze? On which rank, at which collective sequence id, from which process group.

My suspicion is that you were looking at two different collectives, maybe from different ranks or even different collective on the same rank.

PR #166098, opened 2025-10-22
[triton][sigmoid] Fix kernel cache and serialization issue for triton sigmoid + CUDA kernel bug

shawnxu26:
Differential Revision: D84930778


torch/_export/serde/serialize.py
```diff
@@ -552,17 +539,103 @@ def get_triton_kernel_and_cache_entry(node: torch.fx.Node):
...
+            constexpr_vals[actual_kernel.arg_names[idx]]
+            for idx in actual_kernel.constexprs
+            if actual_kernel.arg_names[idx] in constexpr_vals
+        ]
+
+        matching_entries = []
+        for sig_key, cache_entry in cache.items():
+            constexpr_matches = re.findall(
+                r"\('constexpr',\s*(-?\d+(?:\.\d+)?)\)", sig_key
+            )
```
minjang: what if a number is in scientific notation? is it even allowed or possible? you may try a big floating number to see.

if so, I would extract as a general string and attempt to convert a string to a number (and decide int or float after).

torch/_export/serde/serialize.py
```diff
@@ -552,17 +539,103 @@ def get_triton_kernel_and_cache_entry(node: torch.fx.Node):
...
     else:
-        for cache_entry in cache.values():
-            if cache_entry.metadata.num_warps == num_warps:
-                return kernel, cache_entry
+        matching_entries = list(cache.items())
+
+    if len(matching_entries) == 1:
+        return kernel, matching_entries[0][1]
+
+    if len(matching_entries) > 1:
```
minjang: nit. just coding style suggestion. you may early raise if the entries is empty. then you can reduce indentations for a large chunk of the code.

minjang (approved):
LGTM. A minor coding style suggestion and number format in constexpr.

torch/_export/serde/serialize.py
```diff
@@ -552,17 +539,108 @@ def get_triton_kernel_and_cache_entry(node: torch.fx.Node):
...
+        expected_values = [
+            constexpr_vals[actual_kernel.arg_names[idx]]
+            for idx in actual_kernel.constexprs
+            if actual_kernel.arg_names[idx] in constexpr_vals
+        ]
+
+        matching_entries = []
+        for sig_key, cache_entry in cache.items():
+            constexpr_matches = re.findall(
+                r"\('constexpr',\s*(-?\d+(?:\.\d+)?(?:[eE][+-]?\d+)?)\)", sig_key
```
minjang: +3.0, or +3e10 is also valid. the beginning should be `[-+]?`

torch/_export/serde/serialize.py
```diff
@@ -552,17 +539,108 @@ def get_triton_kernel_and_cache_entry(node: torch.fx.Node):
...
+                    entry_metadata, "num_stages"
+                ):
+                    if config.num_stages != entry_metadata.num_stages:
+                        metadata_matches = False
+                        continue
+
+                if hasattr(config, "num_ctas") and hasattr(entry_metadata, "num_ctas"):
+                    if config.num_ctas != entry_metadata.num_ctas:
+                        metadata_matches = False
+                        continue
```
minjang: just another style/refactoring suggestion. we could have a for-loop with `["num_cpu_threads", "num_warps", ...]`, and have a general code, using `getattr`. the code will be much simplified. or, you can refactor into a separate method, then you don't need to have the `metadata_matches` boolean: something like `if is_metadata_unmatched(config, entry_metadata): continue`.

torch/_export/serde/serialize.py
shawnxu26: make sense, changed to `[+-]?`

torch/_export/serde/serialize.py
shawnxu26: Thanks for the comments! I've refactored the code as suggested.

torch/_export/serde/serialize.py
minjang: yes: it should be `[+-]?` I think your regex seems complete that captures int, float and scientific numbers in python.

but, the current code has some weakness: although it seems complete, if it fails, it currently silently ignores it, which could complicate debugging. if we must extract a number after `constexpr`, then we should have an exception handling.

alternatively, you can simply take `r"\('constexpr',\s*(.+)\)"`, extracting whatever sting after `constexpr`. and just let Python do type conversion to a number. it will then also throw an exception if it is an invalid number.

shawnxu26: That's a good point. For the pattern, we may need to use `r"\('constexpr',\s*([^)]+)\)"` to match everything until a `")"`.

minjang (approved):

PR #166107, opened 2025-10-23
[xpu][fix] Use default warps for inner reductions on XPU 

weishi-deng:
Fixes [[#triton-xpu-issue](https://github.com/intel/intel-xpu-backend-for-triton/issues/5367)], the config of less warps for reduction on XPU devices causes register spill and performance drop of 70%.

### DIFF
--- a/torch/_inductor/runtime/triton_heuristics.py
+++ b/torch/_inductor/runtime/triton_heuristics.py
@@ -2436,7 +2436,6 @@ def triton_config_reduction(
         else:
             num_warps = total_numel() // 128
 
-
     max_num_warps = 16 if r <= 8192 else 32
     num_warps = _num_warps(
         num_warps, max_num_warps=max_num_warps, register_intensive=register_intensive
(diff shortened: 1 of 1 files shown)


torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -2431,6 +2431,12 @@ def total_numel() -> int:
         else:
             num_warps = total_numel() // 128
 
+    # Workaround here since less warps cause register spills on XPU.
```
EikanWang: It does not make sense to ignore the `num_warps`.

EikanWang (changes requested):

torch/_inductor/runtime/triton_heuristics.py
weishi-deng: Updated the workaround heuristic for XPU Triton num_warps, which reverts the PR https://github.com/pytorch/pytorch/pull/162447 on xpu until we fix the regression on XPU Triton.

PR #166115, opened 2025-10-23
Strengthen the implementation of OpenReg Stream

zhudada0120:
This PR redesigns the implementation of OpenRegStream, enabling it to support default streams and normal stream pools with multiple priorities. It also adds robustness checks such as device ID validation and supplements relevant test cases for stream context management.

**Changes and Reasons**

1. OpenReg runtime (OpenRegStream.cpp)

1）Redesigned the ID of OpenReg Stream. In the new implementation:
- StreamIdType=6 indicates the default stream, and StreamIdType=7 represents the external stream. Values 0-5 are reserved as priority codes for normal stream pools. OpenReg currently supports stream pools with priorities 0-1. This design makes the priority of stream pools more intuitive and easier to understand.
- Modified StreamIdType from an enum to a regular class to support multi-priority scenarios.
- Updated the implementation of OpenRegStream::stream() for retrieving orStream_t.
- Adjusted the code in the original implementation that restricted stream pools to single-priority, enabling multi-priority support.

2）Enhanced robustness: added device validation, DeviceGuard initialization. 

2. Tests (test_streams.py)

Added assertions for "original stream restoration" in context exception paths and stream switching checks, covering key code paths.

### DIFF
--- a/test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
+++ b/test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
@@ -14,12 +14,11 @@ namespace c10::openreg {
 namespace {
 
 // Global stream state and constants
-static c10::once_flag init_flag;
-
-static DeviceIndex num_devices = -1;
-static constexpr int kStreamsPerPoolBits = 5;
-static constexpr int kStreamsPerPool = 1 << kStreamsPerPoolBits;
-static constexpr int kStreamTypeBits = 2;
+c10::once_flag init_flag;
+DeviceIndex num_devices = -1;
+constexpr int kStreamsPerPoolBits = 5;
+constexpr int kStreamsPerPool = 1 << kStreamsPerPoolBits;
+constexpr int kStreamTypeBits = 2;
 
 /*
  * The stream pools are lazily initialized when the first queue is requested
@@ -27,16 +26,16 @@ static constexpr int kStreamTypeBits = 2;
  * a queue is requested, the next queue in the pool to be returned in a
  * round-robin fashion, see Note [Stream Management].
  */
-static std::deque<c10::once_flag> device_flags;
-static std::vector<std::array<
+std::deque<c10::once_flag> device_flags;
+std::vector<std::array<
     std::array<orStream_t, kStreamsPerPool>,
     c10::openreg::max_compile_time_stream_priorities>>
     streams;
-static std::deque<
+std::deque<
     std::array<std::atomic<uint32_t>, max_compile_time_stream_priorities>>
     priority_counters;
 
-static thread_local std::unique_ptr<StreamId[]> current_streams = nullptr;
+thread_local std::unique_ptr<StreamId[]> current_streams = nullptr;
 
 /*
  * Note [StreamId assignment]
@@ -81,29 +80,28 @@ inline std::ostream& operator<<(std::ostream& stream, StreamIdType s) {
   return stream << static_cast<int16_t>(s);
 }
 
-static inline StreamIdType streamIdType(StreamId s) {
+inline StreamIdType streamIdType(StreamId s) {
   // Externally allocated streams have their id being the orStream_ptr
   // so the last bit will be 0
-  if (!(s & 1)) {
+  if (!(s & 1) && s) {
     return StreamIdType(StreamIdType::EXT);
   }
-
   int mask_for_type = (1 << kStreamTypeBits) - 1;
   auto st = static_cast<StreamIdType>((s >> 1) & mask_for_type);
   TORCH_CHECK(
       st == StreamIdType::DEFAULT || st == StreamIdType::NORMAL,
-      "invalid StreamId: ",
+      "invalid StreamIdType: ",
       s);
   return st;
 }
 
-static inline size_t streamIdIndex(StreamId s) {
+inline size_t streamIdIndex(StreamId s) {
   return static_cast<size_t>(
       (s >> (kStreamTypeBits + 1)) & ((1 << kStreamsPerPoolBits) - 1));
 }
 
 StreamId makeStreamId(StreamIdType st, size_t si) {
-  if (st == StreamIdType::EXT) {
+  if (st == StreamIdType::DEFAULT) {
     return static_cast<StreamId>(0);
   }
 
@@ -111,25 +109,22 @@ StreamId makeStreamId(StreamIdType st, size_t si) {
       (static_cast<StreamId>(st) << 1) | 1;
 }
 
-static void initGlobalStreamState() {
+void initGlobalStreamState() {
   num_devices = device_count();
   device_flags.resize(num_devices);
   streams.resize(num_devices);
   priority_counters.resize(num_devices);
 }
 
-static void initSingleDeviceStream(
-    int priority,
-    DeviceIndex device_index,
-    int i) {
+void initSingleDeviceStream(int priority, DeviceIndex device_index, int i) {
   auto& stream = streams[device_index][priority][i];
-
   OPENREG_CHECK(orStreamCreateWithPriority(&stream, 0, priority));
   priority_counters[device_index][priority] = 0;
 }
 
 // Creates stream pools for the specified device. It should be call only once.
-static void initDeviceStreamState(DeviceIndex device_index) {
+void initDeviceStreamState(DeviceIndex device_index) {
+  DeviceGuard device_guard{Device(DeviceType::PrivateUse1, device_index)};
   for (const auto i : c10::irange(kStreamsPerPool)) {
     for (const auto p : c10::irange(max_compile_time_stream_priorities)) {
       initSingleDeviceStream(p, device_index, i);
@@ -137,18 +132,15 @@ static void initDeviceStreamState(DeviceIndex device_index) {
   }
 }
 
-static void initOpenRegStreamsOnce() {
+void initOpenRegStreamsOnce() {
   c10::call_once(init_flag, initGlobalStreamState);
-
   for (const auto i : c10::irange(num_devices)) {
     c10::call_once(
         device_flags[i], initDeviceStreamState, static_cast<DeviceIndex>(i));
   }
-
   if (current_streams) {
     return;
   }
-
   // Inits current streams (thread local) to the last queue in the "normal
   // priority" queue pool. Note: the queue pool have not been initialized yet.
   // It will be initialized in initDeviceStreamState for the specified device.
@@ -158,9 +150,19 @@ static void initOpenRegStreamsOnce() {
   }
 }
 
-static uint32_t get_idx(std::atomic<uint32_t>& counter) {
-  auto raw_idx = counter++;
-  return raw_idx % kStreamsPerPool;
+inline void check_device(DeviceIndex device_index) {
+  TORCH_CHECK(
+      device_index >= 0 && device_index < num_devices,
+      "Device index value ",
+      static_cast<int>(device_index),
+      " is out of index range [0, ",
+      static_cast<int>(num_devices),
+      ")");
+}
+
+uint32_t get_idx(std::atomic<uint32_t>& counter) {
+  auto raw = counter++;
+  return raw % kStreamsPerPool;
 }
 
 OpenRegStream OpenRegStreamForId(DeviceIndex device_index, StreamId stream_id) {
@@ -183,8 +185,9 @@ orStream_t OpenRegStream::stream() const {
   switch (st) {
     // The index 0 stream is default as well.
     case StreamIdType::DEFAULT:
-    case StreamIdType::NORMAL:
       return streams[device_index][static_cast<uint8_t>(st)][si];
+    case StreamIdType::NORMAL:
+      return streams[device_index][static_cast<uint8_t>(st) - 1][si];
     case StreamIdType::EXT:
       return reinterpret_cast<orStream_t>(stream_id);
     default:
@@ -207,10 +210,9 @@ OpenRegStream getStreamFromPool(const int priority, DeviceIndex device_index) {
   if (device_index == -1) {
     device_index = current_device();
   }
-  auto pri_idx =
-      std::clamp(priority, 0, max_compile_time_stream_priorities - 1);
+  auto pri_idx = std::clamp(priority, 0, max_compile_time_stream_priorities);
   const auto idx = get_idx(priority_counters[device_index][pri_idx]);
-  auto id_type = static_cast<StreamIdType>(pri_idx);
+  auto id_type = static_cast<StreamIdType>(pri_idx + 1);
   return OpenRegStreamForId(device_index, makeStreamId(id_type, idx));
 }
 
@@ -232,6 +234,7 @@ OpenRegStream getDefaultOpenRegStream(DeviceIndex device_index) {
   if (device_index == -1) {
     device_index = current_device();
   }
+  check_device(device_index);
   return OpenRegStreamForId(
       device_index, makeStreamId(StreamIdType::DEFAULT, 0));
 }
@@ -241,6 +244,7 @@ OpenRegStream getCurrentOpenRegStream(DeviceIndex device_index) {
   if (device_index == -1) {
     device_index = current_device();
   }
+  check_device(device_index);
   return OpenRegStreamForId(device_index, current_streams[device_index]);
 }
 
--- a/test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
+++ b/test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
@@ -18,7 +18,7 @@ c10::once_flag init_flag;
 DeviceIndex num_devices = -1;
 constexpr int kStreamsPerPoolBits = 5;
 constexpr int kStreamsPerPool = 1 << kStreamsPerPoolBits;
-constexpr int kStreamTypeBits = 2;
+constexpr int kStreamTypeBits = 3;
 
 /*
  * The stream pools are lazily initialized when the first queue is requested
@@ -29,10 +29,11 @@ constexpr int kStreamTypeBits = 2;
 std::deque<c10::once_flag> device_flags;
 std::vector<std::array<
     std::array<orStream_t, kStreamsPerPool>,
-    c10::openreg::max_compile_time_stream_priorities>>
+    c10::openreg::max_compile_time_stream_priorities + 1>>
     streams;
+std::vector<orStream_t> default_streams;
 std::deque<
-    std::array<std::atomic<uint32_t>, max_compile_time_stream_priorities>>
+    std::array<std::atomic<uint32_t>, max_compile_time_stream_priorities + 1>>
     priority_counters;
 
 thread_local std::unique_ptr<StreamId[]> current_streams = nullptr;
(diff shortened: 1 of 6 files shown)


docs/source/accelerator/stream.md
```diff
@@ -0,0 +1,88 @@
+# Stream
+
+## Background
+
+Streams are the asynchronous work queues on an accelerator. They allow multiple independent command queues to make progress concurrently on the same device, while the runtime preserves per-thread and per-device “current stream” state. In PyTorch, a stream API must:
+- Provide a per-device default stream and a reusable pool of non-default streams
```
guangyey: Hi, I think not every backend provides a default stream, and the concept of a “default stream” can vary between different backends.
In PyTorch, backends are required to support 'non-default' streams, while the 'default stream' is optional.

docs/source/accelerator/stream.md
```diff
@@ -0,0 +1,88 @@
...
+
+Goal: deliver a minimal, production-ready Stream implementation aligned with the OpenReg reference, with correct context restoration and full `torch.Stream` interop.
+
+## Design
+
+The accelerator needs to support the following basic functionalities.
+
+| Functionality | Description | When to use |
+| :-- | :-- | :-- |
+| Default stream | Retrieve the default stream of a device. | Implicit work submission when no explicit stream is set. |
```
guangyey: Could we remove Default Stream here?

torch/csrc/Stream.cpp
```diff
@@ -82,11 +84,14 @@ static PyObject* THPStream_pynew(
...
   std::optional<c10::Stream> stream_opt;
   if (r.idx == 0) {
+    // LITERALINCLUDE START: PYTORCH STREAM CTOR GETNEW
     c10::impl::VirtualGuardImpl impl{device_type};
     stream_opt = impl.getNewStream(
         c10::Device(device_type, device_index), static_cast<int>(priority));
+    // LITERALINCLUDE END: PYTORCH STREAM CTOR GETNEW
   } else {
     stream_opt = c10::Stream::unpack3(stream_id, device_index, device_type);
+
```
guangyey:
```suggestion
```

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegGuard.h
```diff
@@ -108,6 +110,7 @@ struct OpenRegGuardImpl final : public c10::impl::DeviceGuardImplInterface {
     const auto old_stream = getCurrentOpenRegStream(s.device().index());
     setCurrentOpenRegStream(stream);
     return old_stream.unwrap();
+
```
guangyey:
```suggestion
```

docs/source/accelerator/stream.md
zhudada0120: The default stream has been changed to an optional implementation.

docs/source/accelerator/stream.md
zhudada0120: The default stream has been moved to the end of the table and marked as optional.

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegGuard.h
zhudada0120: Fixed

torch/csrc/Stream.cpp
zhudada0120: Fixed

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
```diff
@@ -153,9 +154,19 @@ static void initOpenRegStreamsOnce() {
...
-  return raw_idx % kStreamsPerPool;
+inline void check_device(DeviceIndex device_index) {
+  TORCH_CHECK(
+      device_index >= 0 && device_index < num_devices,
+      "Device index value ",
+      static_cast<int>(device_index),
+      " is out of index range [0, ",
+      static_cast<int>(num_devices),
+      ")");
+}
```
KarhouTam, line 182: I think this basic function should better be in `OpenRegFunctions.h/cpp`?

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
```diff
@@ -14,29 +14,28 @@ namespace c10::openreg {
...
+std::vector<std::array<
+  std::array<orStream_t, kStreamsPerPool>,
+  c10::openreg::max_compile_time_stream_priorities>>
+  streams;
+std::deque<
+  std::array<std::atomic<uint32_t>, max_compile_time_stream_priorities>>
+  priority_counters;
 
-static thread_local std::unique_ptr<StreamId[]> current_streams = nullptr;
+thread_local std::unique_ptr<StreamId[]> current_streams = nullptr;
```
KarhouTam, line 40: A dumb question, why here is not `static`?

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
```diff
@@ -178,8 +189,9 @@ orStream_t OpenRegStream::stream() const {
   switch (st) {
     // The index 0 stream is default as well.
     case StreamIdType::DEFAULT:
+      return nullptr;
     case StreamIdType::NORMAL:
-      return streams[device_index][static_cast<uint8_t>(st)][si];
+      return streams[device_index][static_cast<uint8_t>(st) - 1][si];
```
KarhouTam: Another dumb question, why `-1` here?

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
```diff
@@ -207,9 +220,10 @@ OpenRegStream getStreamFromPool(const int priority, DeviceIndex device_index) {
   auto pri_idx =
       std::clamp(priority, 0, max_compile_time_stream_priorities - 1);
   const auto idx = get_idx(priority_counters[device_index][pri_idx]);
-  auto id_type = static_cast<StreamIdType>(pri_idx);
+  auto id_type = static_cast<StreamIdType>(pri_idx + 1);
```
KarhouTam: Ditto, why `+1` here?

docs/source/accelerator/stream.md
```diff
@@ -0,0 +1,92 @@
...
+.. literalinclude:: ../../../test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
+    :language: c++
+    :start-after: LITERALINCLUDE START: OPENREG GET STREAM FROM POOL
+    :end-before: LITERALINCLUDE END: OPENREG GET STREAM FROM POOL
+    :linenos:
+```
+
+### Summary
+
+This implementation ensures efficient and thread-safe allocation of streams from the pool, leveraging PyTorch’s abstraction layers and OpenReg’s backend capabilities.
```
KarhouTam: Maybe no need?

KarhouTam (commented):
Hi, @zhudada0120 . Just leave some comments (most are dumb questions) here. PTAL.

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
zhudada0120: This references the CUDA implementation, where check_device in CUDA is part of CUDAStream.cpp.

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
zhudada0120: The anonymous namespace itself implicitly has the property of internal linkage, and its effect is exactly the same as that of static at file scope. There is no need to explicitly add static.

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
zhudada0120: In the encoding of OpenReg Stream Id, the default stream and the ordinary stream should be distinguished, as the original implementation caused conflicts between the IDs of these two types of streams.

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegGuard.cpp
```diff
@@ -2,6 +2,8 @@
 
 namespace c10::openreg {
 
+// LITERALINCLUDE START: OPENREG GUARD REGISTRATION
 C10_REGISTER_GUARD_IMPL(PrivateUse1, OpenRegGuardImpl);
+// LITERALINCLUDE END: OPENREG GUARD REGISTRATION
```
fffrog, line 7: The registration of Guard will be shown in the Guard docs.

docs/source/accelerator/stream.md
```diff
@@ -0,0 +1,92 @@
...
+- Interoperate with Python via the device-agnostic `torch.Stream`
+- Compose with Guards so that higher-level PyTorch components can route to the backend correctly
+
+```{note}
+A per-device default stream may be provided by some backends, but the concept and semantics of a “default stream” are backend-dependent and therefore optional. The minimal implementation of the backend only needs to support non-default streams.
+```
+
+Goal: deliver a minimal, production-ready Stream implementation aligned with the OpenReg reference, with correct context restoration and full `torch.Stream` interop.
+
+## Design
```
fffrog: It would be better if we can illustrate three point informations there:
- Give a introduction of StreamID composition
- Show what StreamID perform to achieve device-agnostic
- Show the key functionalities about Stream class and hierarchy

docs/source/accelerator/stream.md
```diff
@@ -0,0 +1,92 @@
...
+
+| Functionality | Description | When to use |
+| :-- | :-- | :-- |
+| Stream pool | Obtain a reusable, non-default stream from a per-device pool (round-robin). | Overlap independent work or isolate async regions. |
+| Current stream (TLS) | Get/set the current stream per thread and device. | Enter/exit contexts, ensure ops submit to intended stream. |
+| External stream wrap | Wrap a foreign runtime handle as a PyTorch stream (no ownership transfer). | Interop with external libraries or pre-existing streams. |
+| Guard integration | Route `VirtualGuardImpl(DeviceType)` hooks to backend (stream creation/switch). | Make `torch.Stream` and context managers work uniformly. |
+| Default stream (optional) | If provided by the backend, retrieve the per-device default stream. | Used for implicit work submission when no explicit stream is set; otherwise, backends may require explicit streams or use a different mechanism. |
+
+## Implementation
```
fffrog: PyTorch provider two way to support Stream

- torch.Stream
- torch.xxx.Stream

The former is device-agnostic, which is alse recommended by communities and help to reduce the developement cost for new accelerator. And the latter is device specific implementation, which is used for compatibality.

There we need to show both two ways for developers, including C++ side, exposing C++ to Python and Python side

docs/source/accelerator/stream.md
```diff
@@ -0,0 +1,92 @@
...
+.. literalinclude:: ../../../torch/csrc/Stream.cpp
+    :language: c++
+    :start-after: LITERALINCLUDE START: PYTORCH STREAM CTOR GETNEW
+    :end-before: LITERALINCLUDE END: PYTORCH STREAM CTOR GETNEW
+    :linenos:
+```
+
+- `VirtualGuardImpl` is PyTorch’s abstraction layer for device-specific operations. It routes the request to the appropriate backend implementation.
+
+### Guard Layer
```
fffrog: `Guard` will serve as a dependency, demonstrating how `Stream` is integrated into `Guard` for use by PyTorch's internal mechanisms.

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
```diff
@@ -81,74 +80,74 @@ inline std::ostream& operator<<(std::ostream& stream, StreamIdType s) {
...
-static inline StreamIdType streamIdType(StreamId s) {
+inline StreamIdType streamIdType(StreamId s) {
   // Externally allocated streams have their id being the orStream_ptr
   // so the last bit will be 0
-  if (!(s & 1)) {
+  if (!(s & 1) && s) {
     return StreamIdType(StreamIdType::EXT);
   }
-
+  TORCH_CHECK(
```
fffrog: From my point of view, the `s` can equal to 0, because default stream is 0.

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
```diff
@@ -81,74 +80,74 @@ inline std::ostream& operator<<(std::ostream& stream, StreamIdType s) {
...
 
-static inline size_t streamIdIndex(StreamId s) {
+inline size_t streamIdIndex(StreamId s) {
   return static_cast<size_t>(
       (s >> (kStreamTypeBits + 1)) & ((1 << kStreamsPerPoolBits) - 1));
 }
 
 StreamId makeStreamId(StreamIdType st, size_t si) {
-  if (st == StreamIdType::EXT) {
+  if (st == StreamIdType::DEFAULT) {
```
fffrog: The CPU-based OpenReg accelerator does not support default streams with a value of nullptr because OpenReg simulates stream behavior by creating an orStream structure, which creates a new thread in its constructor.

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
```diff
@@ -183,8 +192,9 @@ orStream_t OpenRegStream::stream() const {
   switch (st) {
     // The index 0 stream is default as well.
     case StreamIdType::DEFAULT:
+      return nullptr;
     case StreamIdType::NORMAL:
-      return streams[device_index][static_cast<uint8_t>(st)][si];
+      return streams[device_index][static_cast<uint8_t>(st) - 1][si];
```
fffrog: In theory, the default stream and ordinary stream should be separate, so we do not need to subtract 1 additionally.

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
```diff
@@ -207,10 +210,9 @@ OpenRegStream getStreamFromPool(const int priority, DeviceIndex device_index) {
   if (device_index == -1) {
     device_index = current_device();
   }
-  auto pri_idx =
-      std::clamp(priority, 0, max_compile_time_stream_priorities - 1);
+  auto pri_idx = std::clamp(priority, 0, max_compile_time_stream_priorities);
   const auto idx = get_idx(priority_counters[device_index][pri_idx]);
-  auto id_type = static_cast<StreamIdType>(pri_idx);
+  auto id_type = static_cast<StreamIdType>(pri_idx + 1);
```
fffrog: Hey, it is a little weird to adjust offset by +1/-1;

A reachable way to fix it is we can redefine the StreamIdType.

```C++
enum class StreamIdType : uint8_t {
  NORMAL = 0x0,
  DEFAULT = 0x6,
  EXT = 0x7,
};
```

Therefore, the priority of stream will range from 0x0 to 0x6, we do not need to adjust the offset manually, and as of `DEFAULT`, it can be treated as same ligic as `EXT`.

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
```diff
@@ -14,43 +14,46 @@ namespace c10::openreg {
...
- * Where StreamIdType:
- *  00 = default stream
- *  01 = normal stream
- *  11 = external stream
+ * StreamIdType:
+ *  000-001 = OpenReg priority stream
+ *  110 = default stream
+ *  111 = external stream
+ * Priorities 000-101 are reserved for the normal stream pool and can be extended as per practical requirements.
+ * OpenReg currently supports priorities 0 and 1.
```
fffrog:
```suggestion
 * StreamIdType:
 *  000 = normal stream
 *  001 = high stream
 *  110 = default stream
 *  111 = external stream

 * The range 000 to 101 is reserved for stream pools of different priorities and can be expanded as needed. (OpenReg currently supports two priorities: 0 and 1)
```

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
```diff
@@ -60,95 +63,104 @@ static thread_local std::unique_ptr<StreamId[]> current_streams = nullptr;
...
+  switch (s.getStreamType()) {
     case StreamIdType::DEFAULT:
       return stream << "DEFAULT";
-    case StreamIdType::NORMAL:
-      return stream << "NORMAL";
     case StreamIdType::EXT:
       return stream << "EXT";
     default:
-      break;
+      return stream << "NORMAL" << static_cast<int>(s.getStreamType());
```
fffrog:
```suggestion
      return stream << "PRIORITY" << static_cast<int>(s.getStreamType());
```

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
zhudada0120: I have changed to the format you mentioned. Thanks for the advice. However, during the modification, I swapped the stream descriptions of 000 and 001 because stream 000 has a higher priority than stream 001.

test/cpp_extensions/open_registration_extension/torch_openreg/csrc/runtime/OpenRegStream.cpp
zhudada0120: I have changed to the format you mentioned. Thanks for the advice.

test/cpp_extensions/open_registration_extension/torch_openreg/third_party/openreg/csrc/stream.cpp
```diff
@@ -210,7 +210,7 @@ orError_t orStreamCreate(orStream_t* stream) {
   int min_p, max_p;
   orDeviceGetStreamPriorityRange(&min_p, &max_p);
 
-  return orStreamCreateWithPriority(stream, 0, max_p);
```
fffrog, line 213: Hey, the second parameter is not for low priority, it is just used for a flag indicating something else.

test/cpp_extensions/open_registration_extension/torch_openreg/third_party/openreg/csrc/stream.cpp
```diff
@@ -178,6 +177,7 @@ orError_t orStreamCreateWithPriority(
     orStream_t* stream,
     [[maybe_unused]] unsigned int flag,
     int priority) {
+  priority = -priority;
```
fffrog: Hey!

if  priority is -1 now, then the value of priority will be changed to 1 after this expression, then it will raise an error at line 187, because the priority is greater than  max_p whose value is 0, right?

So, in order to help understand, we can just set the priority to 0 and 1, and the priority is higher when the number is bigger, and delete this line, is that OK?

test/cpp_extensions/open_registration_extension/torch_openreg/third_party/openreg/csrc/stream.cpp
```diff
@@ -283,8 +283,8 @@ orError_t orDeviceGetStreamPriorityRange(
     return orErrorUnknown;
   }
 
-  // OpenReg have only one priority now.
-  *leastPriority = 0;
+  // OpenReg priority levels are -1 and 0
+  *leastPriority = -1;
```
fffrog: Please refet to the comments located at line 180 of file(test/cpp_extensions/open_registration_extension/torch_openreg/third_party/openreg/csrc/stream.cpp)

fffrog (dismissed):
SGTM, thank you for updaing.

fffrog (approved):
LGTM, thank you for updating.

test/cpp_extensions/open_registration_extension/torch_openreg/tests/test_streams.py
```diff
@@ -30,6 +29,21 @@ def test_stream_context(self):
         with torch.Stream(device="openreg:1") as stream:
             self.assertEqual(torch.accelerator.current_stream(), stream)
 
+    def test_stream_context_exception_restore(self):
+        prev_stream = torch.Stream(device="openreg:0")
+        torch.accelerator.set_stream(prev_stream)
```
fffrog: Hey, from my point of view, I do not need those two line.

PR #166119, opened 2025-10-23
Fix device mismatch in sdpa docstring example

Danielohayon:
Fixes #166117 

This PR corrects the Python code example in the `scaled_dot_product_attention` docstring using the fix I suggested in the issue.

The Python code example in the docstring for torch.nn.functional.scaled_dot_product_attention can cause a device mismatch error when the input tensors are on a GPU.

Problematic line: `temp_mask = torch.ones(L, S, dtype=torch.bool).tril(diagonal=0)`


sladyn98 (changes requested):
I think if you add the release note label we can get this merged in .

PR #166138, opened 2025-10-23
[inductor][ez] add overridable env var for disabling fx graph cache

shunting314:
* __->__ #166138

I set TORCHINDUCTOR_FX_GRAPH_CACHE=0 a lot to make sure the compilation
happens by disabling fx graph caching. I even put this in my .bashrc.
But this cause a simple vllm script fail:
https://gist.github.com/shunting314/4253b2b5ab5e7d1b0fc9516c84054904

Error log:
https://gist.github.com/shunting314/1d04bbeb58bc486f975684f56d65615d

The root cause is,
1. vllm patch inductor_config.fx_graph_cache to True here:
   https://github.com/vllm-project/vllm/blob/e255d929902dcf8968541d2cbf0d18f0fe3f9c49/vllm/compilation/compiler_interface.py#L308

   The code in vllm relies fx graph cache is on (unless
   VLLM_DISABLE_COMPILE_CACHE is overriden to false)
2. setting TORCHINDUCTOR_FX_GRAPH_CACHE=0 will cause
   inductor_config.fx_graph_cache not overridable.

I add TORCHINDUCTOR_FX_GRAPH_CACHE_DEFAULT so that we can still use it to skip fx
graph cache while still allow project like vllm to override it.

### DIFF
--- a/torch/_inductor/config.py
+++ b/torch/_inductor/config.py
@@ -92,6 +92,7 @@ precompilation_timeout_seconds: int = 60 * 60
 # use fx aot graph codegen cache
 fx_graph_cache: bool = Config(
     justknob="pytorch/remote_cache:enable_local_fx_graph_cache",
+    env_name_default="TORCHINDUCTOR_FX_GRAPH_CACHE_DEFAULT",
     env_name_force="TORCHINDUCTOR_FX_GRAPH_CACHE",
     default=True,
 )


shunting314:
The setting like TORCHINDUCTOR_FX_GRAPH_CACHE=0 causes the following behavior:
```
inductor_config.fx_graph_cache = True
print(inductor_config.fx_graph_cache)  # get False since TORCHINDUCTOR_FX_GRAPH_CACHE is assigned as env_name_force field
```
The new envvar will avoid that and is safe to put in .bashrc

eellison (approved):

PR #166144, opened 2025-10-23
[Inductor-FX] Don't flatten constant args

chenmillie:
Summary: Fallback kernels are created with flattened constant args and an `unflatten` utility to unflatten them when needed. Apply it in FXConverter to preserve the original structure

Test Plan: added new CI tests

Differential Revision: D85347589

### DIFF
--- a/torch/_inductor/codegen/wrapper_fxir.py
+++ b/torch/_inductor/codegen/wrapper_fxir.py
@@ -1112,11 +1112,15 @@ class FxConverter:
         # Get FX nodes corresponding to the call args.
         assert ir.is_node_sequence(kernel.inputs)
         tensor_nodes = tuple(self._generate_buffer(arg) for arg in kernel.inputs)
-        args = tensor_nodes + tuple(kernel.constant_args)
+        if hasattr(kernel, "unflatten_args"):
+            args, _ = kernel.unflatten_args(tensor_nodes, kernel.constant_args)
+        else:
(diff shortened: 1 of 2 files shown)


blaine-rister (approved):
LGTM! I left one comment; please see that before merging.

torch/_inductor/codegen/wrapper_fxir.py
```diff
@@ -1113,11 +1113,15 @@ def _generate_extern_kernel_common(
         # Get FX nodes corresponding to the call args.
         assert ir.is_node_sequence(kernel.inputs)
         tensor_nodes = tuple(self._generate_buffer(arg) for arg in kernel.inputs)
-        args = tensor_nodes + tuple(kernel.constant_args)
+        if hasattr(kernel, "unflatten_args"):
```
blaine-rister, line 1115: Have we seen any cases where `kernel` doesn't have this attr? Since `kernel` has type `ir.ExternKernel`, it seems like it should have an `unflatten_args` method. Then again, Python type annotations can be misleading.

chenmillie: Sometimes it can be an `ir.ExternKernelOut`, which doesn't have the attribute.

blaine-rister: Makes sense. Thanks.

Issue #164562, opened 2025-10-03
Update FlashInfer version on PyTorch CI vllm tests to v0.3.1


huydhn:
This is to match what vLLM is using after https://github.com/vllm-project/vllm/pull/25782, we should also explore the option to use the new `tools/flashinfer-build.sh` script there to simplify the build process.  This is an important dependency that we need to update to match the behavior of vLLM CI.

https://github.com/pytorch/pytorch/pull/164361 attempted to do this while trying to fix trunk, but we encountered several issues there that we need to abandon the approach:
* Missing dependencies like `cuda-python` and `pynvml`
* Compiling FlashInfer now requires linking with CUDA driver `-lcuda`, which makes it impossible to build on CPU-only runner.  We need to understand why

PR #166146, opened 2025-10-23
vllm test build

yangw-dev:
FIx the vllm test build it's broken due to the flashinfer dependency

### DIFF
--- a/.github/ci_configs/vllm/Dockerfile
+++ b/.github/ci_configs/vllm/Dockerfile
@@ -285,7 +285,7 @@ RUN --mount=type=bind,source=${TORCH_WHEELS_PATH},target=/dist \
 
 # Install the vllm wheel from previous stage
 RUN --mount=type=cache,target=/root/.cache/uv \
-    uv pip install --system /wheels/vllm/*.whl --verbose
+    uv pip install --system /wheels/vllm/*.whl uv pip install --system --prerelease=allow /wheels/vllm/*.whl --verbose
 
 # Install xformers wheel from previous stage
 RUN --mount=type=cache,target=/root/.cache/uv \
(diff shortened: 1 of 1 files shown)


.github/ci_configs/vllm/Dockerfile
```diff
@@ -296,7 +300,7 @@ ARG torch_cuda_arch_list='8.0;8.9;9.0a;10.0a;12.0'
 ENV TORCH_CUDA_ARCH_LIST=${torch_cuda_arch_list}
 
 ARG FLASHINFER_GIT_REPO="https://github.com/flashinfer-ai/flashinfer.git"
```
yangw-dev, line 303: @huydhn 
it sees like uv pip install --system /wheels/vllm/*.whl search for flashinfer 0.4.0 automatically.  I might move the flashinfer wheel build before the vllm wheel build. 
wonder if build flashinfer from source might not the case anymore? A little doubt about it

.github/ci_configs/vllm/Dockerfile
```diff
@@ -296,7 +300,7 @@ ARG torch_cuda_arch_list='8.0;8.9;9.0a;10.0a;12.0'
...
-ARG FLASHINFER_GIT_REF="v0.2.14.post1"
+ARG FLASHINFER_GIT_REF="v0.4.1"
```
huydhn: This might not work IIRC, you can do it later in the context of https://github.com/pytorch/pytorch/issues/164562 if you only want to fix trunk.   We don't need to build flashinfer after https://github.com/pytorch/pytorch/pull/165274#issuecomment-3408531784 lands and can just pip install it from pypi

huydhn (changes requested):
Let's keep `v0.2.14.post1` in this PR https://github.com/pytorch/pytorch/pull/166146/files#r2457590754

.github/ci_configs/vllm/Dockerfile
yangw-dev: ah i see, sounds good!

huydhn (approved):

PR #166149, opened 2025-10-23
transform fr traces for ft

tushar00jain:
Summary:
- the ranks in the default pg config are local ranks
- however fr trace analysis requires them to be global ranks
- so we transform the local ranks to global ranks before the analysis kicks in based on a cli flag

### DIFF
--- a/tools/flight_recorder/components/builder.py
+++ b/tools/flight_recorder/components/builder.py
@@ -374,6 +374,22 @@ def build_collectives(
     return tracebacks, collectives, nccl_calls
 
 
+def transform_ft(
+    details: dict[str, dict[str, Any]], group_world_size: int
+) -> dict[str, dict[str, Any]]:
+    for dump_key, dump in details.items():
+        rank = dump["rank"]
+        for key, pg_config in dump["pg_config"].items():
+            if pg_config["desc"] == "default_pg":
+                ranks = eval(pg_config["ranks"])
+                replica_id = rank // group_world_size
+                first_rank = replica_id * group_world_size
+                new_ranks = [r + first_rank for r in ranks]
+                details[dump_key]["pg_config"][key]["ranks"] = f"{new_ranks}"
+
+    return details
+
+
 def build_db(
     details: dict[str, dict[str, Any]], args: argparse.Namespace, version: str
 ) -> Database:
--- a/tools/flight_recorder/components/config_manager.py
+++ b/tools/flight_recorder/components/config_manager.py
@@ -74,6 +74,17 @@ class JobConfig:
             default=10,
             help="Maximum number of mismatches we print (from earliest).",
         )
+        self.parser.add_argument(
+            "--transform-ft",
+            action="store_true",
+            help="Transform PG config to use global ranks to analyze traces produced by torchft",
+        )
+        self.parser.add_argument(
+            "--group-world-size",
+            type=int,
+            default=None,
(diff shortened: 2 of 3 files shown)


tools/flight_recorder/components/config_manager.py
```diff
@@ -74,6 +74,18 @@ def __init__(self: "JobConfig"):
...
             help="Maximum number of mismatches we print (from earliest).",
         )
+        self.parser.add_argument(
+            "--transform-ft",
+            type=bool,
+            default=False,
+            help="Transform PG config to use global ranks to analyze traces produced by torchft",
+        )
+        self.parser.add_argument(
+            "--world-size",
```
fduwjj: shall we not call it world-size, because this sounds more like a replica size?

tools/flight_recorder/components/builder.py
```diff
@@ -374,6 +374,22 @@ def build_collectives(
...
+) -> dict[str, dict[str, Any]]:
+    for dump_key, dump in details.items():
+        rank = dump["rank"]
+        for key, pg_config in dump["pg_config"].items():
+            if pg_config["desc"] == "default_pg":
+                ranks = eval(pg_config["ranks"])
+                replica_id = rank // world_size
+                first_rank = replica_id * world_size
+                new_ranks = [r + first_rank for r in ranks]
+                details[dump_key]["pg_config"][key]["ranks"] = f"{new_ranks}"
```
fduwjj, line 388: hmmm wait I thought based on our discussion, this should still be global ranks? Or because now different replica has different pg names so it should be fine?

Also I remember we use global ranks inside analysis script, no?

tools/flight_recorder/components/config_manager.py
tushar00jain: we're currently using this terminology in torchft, so better to keep it consistent. world refers to the number of ranks in 1 replica

tools/flight_recorder/components/builder.py
tushar00jain: yes, it should be global rank. hence the need for this pr. we can't really have global ranks set in the default pg in pytorch. there's a lot of pytorch code that relies on default pg having local ranks. probably not the best to interfere with that code

tools/flight_recorder/components/config_manager.py
fduwjj: but world size has a different meaning in other trainers as well right? I mean literally "world" means the whole world no matter there are replicas or not??

tools/flight_recorder/components/builder.py
fduwjj: then is this function only for default PG? because within a replica, we really can still use global ranks.

tushar00jain: @fduwjj currently, yes. we only use default pg and torchft pg. for torchft pg we set the global ranks ourself (I have a pr for that on torchft and there's an if condition here)

tools/flight_recorder/components/config_manager.py
tushar00jain: > but world size has a different meaning in other trainers as well right

that's not the current meaning of the term in the code

> "world" means the whole world no matter there are replicas or not

not true in the code. it just refers to ranks within one replica

tushar00jain: in manager.py, we also use `group_world_size` for this. so maybe use that also here if you don't want "world_size" as it is?

fduwjj (approved):

PR #166160, opened 2025-10-24
[random] Add `generator` arg to `rand*_like` APIs 

### LINKED ISSUE #165865: randn_like should take a generator.

https://docs.pytorch.org/docs/stable/generated/torch.randn.html#torch.randn allows us to bring our own generator but https://docs.pytorch.org/docs/stable/generated/torch.randn_like.html does not which reduces it's convenience. 

### Alternatives

_No response_

### Additional context

_No response_

cc @pbelevich @albanD

KarhouTam:
Fixes #165865

## What this PR does?

- [x] Add `generator` arg to `rand*_like` APIs (`rand_like()`, `randn_like()`, `randint_like()`).
- [x] Add unit tests for  `rand*_like` APIs
- [x] Add corresponding arg docs
- [x] Refactor `rand*_like()` codes in `TensorFactories.cpp`
- [x] Add corresponding and former missed items in `VmapModeRegistrations.cpp`

## Example (using `rand_like()`)

```python
gen0 = torch.Generator()
gen1 = torch.Generator()
gen2 = torch.Generator()

gen0.manual_seed(42)
gen1.manual_seed(42)
gen2.manual_seed(2025)

tensor = torch.empty(10)

t0 = torch.rand_like(tensor, generator=gen0)
t1 = torch.rand_like(tensor, generator=gen1)
t2 = torch.rand_like(tensor, generator=gen2)

assert t0 == t1
assert t2 != t0
assert t2 != t1
```

cc @pbelevich @albanD


KarhouTam:
I think this PR is ready for review!
PTAL. @albanD @pbelevich

aten/src/ATen/native/TensorFactories.cpp
```diff
@@ -1089,6 +1090,7 @@ Tensor& rand_out(
 
 Tensor rand_like(
     const Tensor& self,
+    std::optional<Generator> generator,
```
cyyever, line 1093: It is better to append the new parameter `generator` to the parameter list.

aten/src/ATen/native/TensorFactories.cpp
```diff
@@ -1197,7 +1216,9 @@ Tensor& randint_out(
 
 Tensor randint_like(
     const Tensor& self,
+    int64_t low,
```
cyyever, line 1219: Is `low` really required in this PR? It is another new parameter.

aten/src/ATen/native/TensorFactories.cpp
```diff
@@ -1197,7 +1216,9 @@ Tensor& randint_out(
...
     int64_t high,
+    std::optional<Generator> generator,
```
cyyever, line 1221: Append `generator`?

aten/src/ATen/native/TensorFactories.cpp
```diff
@@ -1089,6 +1090,7 @@ Tensor& rand_out(
 
 Tensor rand_like(
     const Tensor& self,
+    std::optional<Generator> generator,
     std::optional<ScalarType> dtype,
     std::optional<Layout> layout,
     std::optional<Device> device,
```
cyyever, line 1096: What happens if `device` and `generator`'s device are different?

KarhouTam: Will raise error, just like `torch.rand()` and other random APIs. For example
```python
g = torch.Generator("cuda:0")
a = torch.eye(3) # on cpu
torch.rand_like(a, generator=g)
# RuntimeError: Expect a 'cpu' device type for generator but found 'cuda'
```

aten/src/ATen/native/TensorFactories.cpp
KarhouTam: Sorry. A little bit confused. To which parameter list?

aten/src/ATen/native/TensorFactories.cpp
KarhouTam: The `randint_like()` with `low` argument was added before this PR. Because I slightly refactored these functions (let overrides reuse the same function) and git blamed them to me.

aten/src/ATen/native/TensorFactories.cpp
KarhouTam: If is the html doc, then I already done it. Please check [built html preview](https://docs-preview.pytorch.org/pytorch/pytorch/166160/generated/torch.rand_like.html) out.

cyyever: > Sorry. A little bit confused. To which parameter list?

rand_like gets a new parameter `generator`, it is better to put it at the end.

aten/src/ATen/native/TensorFactories.cpp
cyyever: Should we add an override such that it accepts generator but no device and put all the new functionality there?

aten/src/ATen/native/TensorFactories.cpp
KarhouTam: My thought is prepending `generator` ensures consistency with functions like `torch.rand()`, `torch.randn()`, and `torch.randint()`, which place the `generator` argument at the start of their keyword arguments. This change won't break existing user code, as all arguments following `*` need to be passed as keywords, such as `torch.rand_like(..., generator=g, layout=...)`. Please correct me if I miss something.

cyyever: This is a C++ function and there is no kwarg in C++. I mean because of poor linter/ code analysis support in C++, we shouldn't change the positions of parameters if possible.

aten/src/ATen/native/TensorFactories.cpp
KarhouTam: I think the `device` is for explicitly ensure which device the output tensor belongs to. If `device` is not set, then torch will decide it up to `input.device`. For example
```python
device = torch.device("cuda:0")
g = torch.Generator(device)
cpu_tensor = torch.eye(3)
cuda_tensor = torch.eye(3, device=device)
torch.rand_like(cpu_tensor, generator=g) # RuntimeError: Expect a 'cpu' device type for generator but found 'cuda'
torch.rand_like(cpu_tensor, generator=g, device=device) # Good! 
torch.rand_like(cuda_tensor, generator=g) # Also good! 
```
In fact the output's device is always the same as `generator`'s (if is set).

KarhouTam: But we can notice that `torch.rand_like(cpu_tensor, generator=cuda_generator, device=cpu_device)` is not work. I think the `device` argument is used to trans tensor to the target device before generating random numbers. The same behavior as `torch.rand()`, ...

aten/src/ATen/native/TensorFactories.cpp
KarhouTam: Oh, I see. Sound fair.

KarhouTam: A question, if we move `generator` to the end of the list. Should we also change the corresponding python function parameter list? Which will break the consistency to funcitons without `_like` and I think is not worth to do that. 🤔

cyyever: Is it possible to change only the C++ part?

KarhouTam: I will try it locally.

KarhouTam: Have tried. It's not gonna make it. If the c++ function signature changed, the `native_functions.yaml` need to be changed, which means the python frontend functions will be changed finally. 

More importantly, place `generator` at the end will fail the build (somewhere hard coding `return at::native::rand_like(self, dtype, layout, device, pin_memory, memory_format);` and becuase `generator` is at the end and no value is passed. So the call to `rand_like` is ambiguous.

```
 /root/pytorch/build/aten/src/ATen/RegisterCompositeExplicitAutograd_0.cpp: In function 'at::Tensor at::{anonymous}::{anonymous}::wrapper_CompositeExplicitAutograd__rand_like(const at::Tensor&, std::optional<c10::ScalarType>, std::optional<c10::Layout>, std::optional<c10::Device>, std::optional<bool>, std::optional<c10::MemoryFormat>)':
  /root/pytorch/build/aten/src/ATen/RegisterCompositeExplicitAutograd_0.cpp:8982:31: error: call of overloaded 'rand_like(const at::Tensor&, std::optional<c10::ScalarType>&, std::optional<c10::Layout>&, std::optional<c10::Device>&, std::optional<bool>&, std::optional<c10::MemoryFormat>&)' is ambiguous
   8982 |   return at::native::rand_like(self, dtype, layout, device, pin_memory, memory_format);
```
```c++
namespace at {
namespace native {
TORCH_API at::Tensor rand_like(const at::Tensor & self, ::std::optional<at::ScalarType> dtype={}, ::std::optional<at::Layout> layout={}, ::std::optional<at::Device> device={}, ::std::optional<bool> pin_memory={}, ::std::optional<at::MemoryFormat> memory_format=::std::nullopt);
TORCH_API at::Tensor & rand_like_out(const at::Tensor & self, ::std::optional<at::MemoryFormat> memory_format, at::Tensor & out);
TORCH_API at::Tensor rand_like(const at::Tensor & self, ::std::optional<at::ScalarType> dtype={}, ::std::optional<at::Layout> layout={}, ::std::optional<at::Device> device={}, ::std::optional<bool> pin_memory={}, ::std::optional<at::MemoryFormat> memory_format=::std::nullopt, ::std::optional<at::Generator> generator=::std::nullopt);
TORCH_API at::Tensor & rand_like_generator_out(const at::Tensor & self, ::std::optional<at::MemoryFormat> memory_format, ::std::optional<at::Generator> generator, at::Tensor & out);
} // namespace native
} // namespace at
```

cyyever (approved):

albanD (approved):
SGTM!
Thanks for taking the time to look at this one!

PR #166174, opened 2025-10-24
[BE]: Update NCCL version to 2.28.7

Skylion007:
Update NCCL version to 2.28.7. Fixes issues that prevented 2.28.3 update in https://github.com/pytorch/pytorch/pull/162351


eqy:
We had a number of issues with fixes in-flight, requested a review from @nWEIdia to confirm if 2.28.7 is a good target or if we should wait for a slightly later version.

PR #166182, opened 2025-10-24
set pg name based on ranks

tushar00jain:
Summary:
- in torchft we have multiple default pg's, 1 for each task group
- for flight recorder to work, each of these need to have a different name, so entries can be matched
- change the `init_process_group` api to optionally take a list of ranks. if provided, we use the hash of the ranks as the name of the pg. for torchft, we'll pass global ranks here so the default pg have a different name on each task group


fduwjj (commented):
I think this change looks weird to me. Any chance we can do that inside torchft?

fduwjj (approved):

PR #166185, opened 2025-10-24
Improve device and code compute capability check

### LINKED ISSUE #164596: [CUDA 13] Incorrect warning messages are printed out upon importing torch on newer devices like Spark
### 🐛 Describe the bug

Example warning:

0: /usr/local/lib/python3.12/dist-packages/torch/cuda/__init__.py:283: UserWarning:  0:     Found GPU0 NVIDIA GB10 which is of cuda capability 12.1. 0:     Minimum and Maximum cuda capability supported by this version of PyTorch is 0:     (8.0) - (12.0)

The above warning is incorrect because pytorch binary compiled with 12.0 can directly run on 12.1 (Spark)

Workaround PR: https://github.com/pytorch/pytorch/pull/164590  

### Versions

Nightly cuda13 binary. 

cc @ptrblck @eqy @tinglvv @atalman @malfet

### LINKED ISSUE #164708: `torch.cuda._check_capability` could raise a false-negative warning
The function in question is defined here -- 
https://github.com/pytorch/pytorch/blob/9fff8155c362da777e7ce31b85fb2dc7cfced2d5/torch/cuda/__init__.py#L247-L283

In my environment, PyTorch is installed with the following env vars: 
- `TORCH_CUDA_ARCH_LIST="8.9 9.0+PTX"`
- `CUDA_HOME=/usr/local/cuda-12.9`
- `CUDNN_LIB_DIR=/usr/local/cuda-12.9/targets/x86_64-linux/lib`
- `CUDNN_INCLUDE_DIR=/usr/local/cuda-12.9/targets/x86_64-linux/include`

Thus my source build PyTorch works in my environment but when I do something on CUDA then I see a warning as follows:
```
$ python -c 'import torch; print(torch.ones((1,), device="cuda"))'
/path/to/torch/cuda/__init__.py:284: UserWarning:
    Found GPU0 NVIDIA RTX 6000 Ada Generation which is of cuda capability 8.9.
    Minimum and Maximum cuda capability supported by this version of PyTorch is
    (8.9) - (9.0)

  warnings.warn(
/path/to/torch/cuda/__init__.py:302: UserWarning:
    Please install PyTorch with a following CUDA
    configurations:  12.6 12.8 13.0 following instructions at
    https://pytorch.org/get-started/locally/

  warnings.warn(matched_cuda_warn.format(matched_arches))
tensor([1.], device='cuda:0')
```

I'm wondering if this check should respect the minor version of the device being checked, not only the major version.

cc @ptrblck @msaroufim @eqy @jerryzh168 @malfet

lakshayg:
Here's what the revised warning looks like

```
torch/cuda/__init__.py:340: UserWarning: Found GPU0 ... which is of compute capability (CC) 9.0.
The following list shows the CCs this version of PyTorch was built for and the hardware CCs it supports:
- 7.5 which supports hardware CC >=7.5,<8.0
- 8.0 which supports hardware CC >=8.0,<9.0 except {8.7}
- 8.6 which supports hardware CC >=8.6,<9.0 except {8.7}
- 10.0 which supports hardware CC >=10.0,<11.0 except {10.1}
- 10.1 which supports hardware CC {10.1, 11.0}
Please follow the instructions at https://pytorch.org/get-started/locally/ to install a PyTorch release that supports one of these CUDA versions: 12.6, 12.8, 13.0
```

fixes: #164708

Reference: NVIDIA Compute Capability to Device Mapping: https://developer.nvidia.com/cuda-gpus 

cc: @ptrblck @nWEIdia @tinglvv

### DIFF
--- a/test/test_cuda_compatibility.py
+++ b/test/test_cuda_compatibility.py
@@ -1,0 +1,158 @@
+# Owner(s): ["module: cuda"]
+
+import warnings
+from unittest.mock import patch
+
+import torch
+import torch.cuda
+from torch.testing._internal.common_utils import run_tests, TestCase
+
+
+class TestCodeCompatibleWithDevice(TestCase):
+    def test_compatible_cases(self):
+        self.assertTrue(
+            torch.cuda._code_compatible_with_device(device_cc=80, code_cc=80)
+        )
+        self.assertTrue(
+            torch.cuda._code_compatible_with_device(device_cc=86, code_cc=80)
+        )
+
+    def test_backward_incompatible(self):
+        self.assertFalse(
+            torch.cuda._code_compatible_with_device(device_cc=80, code_cc=86)
+        )
+
+    def test_cross_major_incompatible(self):
+        self.assertFalse(
+            torch.cuda._code_compatible_with_device(device_cc=90, code_cc=80)
+        )
+        self.assertFalse(
+            torch.cuda._code_compatible_with_device(device_cc=75, code_cc=80)
+        )
+
+    def test_igpu_cases(self):
+        self.assertFalse(
+            torch.cuda._code_compatible_with_device(device_cc=53, code_cc=50)
+        )
+        self.assertFalse(
+            torch.cuda._code_compatible_with_device(device_cc=87, code_cc=80)
+        )
+        self.assertTrue(
+            torch.cuda._code_compatible_with_device(device_cc=53, code_cc=53)
+        )
+
+    def test_special_case_sm101_on_sm110(self):
+        self.assertTrue(
+            torch.cuda._code_compatible_with_device(device_cc=110, code_cc=101)
+        )
+
+    def test_unknown_code_cc(self):
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
+            result = torch.cuda._code_compatible_with_device(device_cc=990, code_cc=990)
+            self.assertTrue(result)
+            self.assertEqual(len(w), 1)
+            self.assertIn("unknown compute capability", str(w[0].message))
+
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
+            result = torch.cuda._code_compatible_with_device(device_cc=991, code_cc=990)
+            self.assertTrue(result)
+            self.assertEqual(len(w), 1)
+
+
+@patch("torch.cuda.get_device_name", return_value="NVIDIA MOCK DEVICE")
+@patch("torch.cuda.device_count", return_value=1)
+@patch("torch.version.cuda", "12.6")
+class TestCheckCapability(TestCase):
+    def test_rocm_skips_check(self, *args):
+        with (
+            patch("torch.version.cuda", None),
+            warnings.catch_warnings(),
+        ):
+            warnings.simplefilter("error")
+            self.assertIsNone(torch.version.cuda)
+            torch.cuda._check_capability()
+
+    @patch("torch.cuda.get_arch_list", return_value=["sm_70", "sm_80", "sm_90"])
+    @patch("torch.cuda.get_device_capability", return_value=(8, 0))
+    def test_compatible_device_no_warning(self, *args):
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            torch.cuda._check_capability()
+
+    @patch("torch.cuda.get_arch_list", return_value=["sm_80"])
+    @patch("torch.cuda.get_device_capability", return_value=(7, 5))
+    def test_incompatible_device_warns(self, *args):
+        with self.assertWarnsRegex(
+            UserWarning, r"Found GPU0.*which is of compute capability.*7\.5"
+        ):
+            torch.cuda._check_capability()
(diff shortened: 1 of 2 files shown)


torch/cuda/__init__.py
```diff
@@ -245,65 +245,113 @@ def _extract_arch_version(arch_string: str) -> int:
...
+    103: CompatInterval(start=103),
+    110: CompatInterval(start=110),
+    120: CompatInterval(start=120),
+}
+
+
+# TORCH_CUDA_ARCH_LIST for PyTorch releases
+PYTORCH_RELEASES_CODE_CC: dict[str, set[int]] = {
+    "12.6": {50, 60, 70, 80, 86, 90},
+    "12.8": {70, 80, 86, 90, 100, 110, 120},
```
nWEIdia: According to current setup: 
https://github.com/pytorch/pytorch/blob/fdcf402d82d32cc5122b4c39dc674ab6f8e211ce/.ci/aarch64_linux/aarch64_ci_build.sh#L9-L14 

110 would only be available in standard build from 13.0 binary. So the warning message, when on THOR, should only prompt users to download cu130 binary, not cu128.

nWEIdia:
> @pytorchbot label "topic: not user facing"

I would argue this is `user facing` , as potentially this code path could produce something the user might not be seeing previously. Or it fixes something user was seeing (e.g. on Spark), which was the main motivation of this PR. The user won't see the warning anymore e.g. while running pytorch nightly cu13.0 binary with this PR in.

torch/cuda/__init__.py
lakshayg: I have fixed this. I see that this scripts also says that 12.6 only supports SM>=8.0. Should I update that as well or is there something I'm missing?

lakshayg:
> > @lakshayg: @pytorchbot label "topic: not user facing"
> 
> @nWEIdia: I would argue this is `user facing` , ...

I have added the "release notes: cuda" label. Is that the right one or do you recommend a different one?

nWEIdia (approved):
LGTM, thanks!

atalman (approved):
lgtm, do we have a link to Nvidia documentation where device and code compute capability is defined ? This way will be easier to update the code for future devices

nWEIdia:
Yes @atalman : I updated the PR description with this official link: https://developer.nvidia.com/cuda-gpus

lakshayg:
@nWEIdia I believe @ptrblck wanted to review it as well. He wanted to think through the case when someone is building pytorch with custom arch list.

atalman:
Hi @lakshayg is this ready to be landed ?

lakshayg:
@atalman it's ready IMO but @ptrblck might want to take a look

test/test_cuda_compatibility.py
```diff
@@ -0,0 +1,154 @@
...
+
+    @patch("torch.cuda.get_arch_list", return_value=["sm_60"])
+    @patch("torch.cuda.get_device_capability", return_value=(7, 0))
+    def test_warning_suggests_compatible_pytorch_release(self, *args):
+        with warnings.catch_warnings(record=True) as w:
+            warnings.simplefilter("always")
+            torch.cuda._check_capability()
+            self.assertEqual(len(w), 1)
+            msg = str(w[0].message)
+            self.assertIn("12.6", msg)
```
ptrblck, line 139: Could generalize the tests a it more as they are hard-coding the current support matrix and will break in the future e.g. when 13.0 is replaced with 13.1 etc.?

torch/cuda/__init__.py
```diff
@@ -253,68 +253,136 @@ def _extract_arch_version(arch_string: str) -> int:
...
+    110: _CompatInterval(start=110),
+    120: _CompatInterval(start=120),
+}
+
+
+# TORCH_CUDA_ARCH_LIST for PyTorch releases
+PYTORCH_RELEASES_CODE_CC: dict[str, set[int]] = {
+    "12.6": {50, 60, 70, 80, 86, 90},
+    "12.8": {70, 80, 86, 90, 100, 120},
+    "13.0": {75, 80, 86, 90, 100, 110, 120},
```
ptrblck, line 321: Do we need to split between x86 and SBSA since `sm_110` is SBSA only?

lakshayg: I don't think this part needs any change because it's only used to report the compatible pytorch release. If the user has sm_110, they'll see pytorch cuda 13.0 as the suggested release. This mapping only ever goes from right to left.

test/test_cuda_compatibility.py
lakshayg: I have mocked the relevant dict so the test will not need to be changed

busFred:
i'm on spark with torch 2.9.1 cu13.0 and i'm also receiving the same error. i skim through this discussion chain, correct me if i'm wrong, it seems this warning is just a false alarm and does not affect actual usability; I guess I could just ignore this until the next update comes out(?)

ptrblck:
@busFred Yes, ignore the warning assuming you are already using any of our PyTorch builds with CUDA 13.x on Spark.

PR #166194, opened 2025-10-24
[ROCm][tunableop] Added MX-FP8 and MX-FP4 datatype support to Tunableop

sarthaktandon9amd:
- We want to add mx-fp4 and mx-fp8 support to tunableop allowing us to tune scaled workloads.

### DIFF
--- a/aten/src/ATen/cuda/tunable/GemmHipblaslt.h
+++ b/aten/src/ATen/cuda/tunable/GemmHipblaslt.h
@@ -525,7 +525,7 @@ class HipblasltGemmOp : public Callable<ParamsT> {
         hipblasLtMatmulDescAttributes_t a_scale_ptr_desc = HIPBLASLT_MATMUL_DESC_A_SCALE_POINTER;
         hipblasLtMatmulDescAttributes_t b_scale_ptr_desc = HIPBLASLT_MATMUL_DESC_B_SCALE_POINTER;
 
-        bool is mx_fp8 = false;
+        bool is_mx_fp8 = false;
 #if ROCM_VERSION >= 70000
         if constexpr (std::is_same_v<ParamsT, ScaledGemmParams<CT>>) {
           auto* scaled_params = static_cast<const ScaledGemmParams<CT>*>(params);
@@ -737,7 +737,7 @@ auto GetHipBlasLtTypeStringAndOps() {
         computeType,
         heuristic_result));
   }
-  
+
   TORCH_HIPBLASLT_CHECK(hipblasLtDestroy(handle));
 
   int returned_algo_count = heuristic_result.size();
--- a/aten/src/ATen/cuda/tunable/GemmHipblaslt.h
+++ b/aten/src/ATen/cuda/tunable/GemmHipblaslt.h
@@ -10,7 +10,6 @@
 #include <c10/cuda/CUDACachingAllocator.h>
 #include <c10/util/StringUtil.h>
 #include <fmt/printf.h>
-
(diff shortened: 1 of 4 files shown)


aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -1233,62 +1233,118 @@ _tunable_scaled_gemm_rocm(
...
-              BLASOP_A, BLASOP_B> scaledgemm{};                       \
-          scaledgemm(&params);                                        \
-        }                                                             \
-        else if (mat2.scalar_type() == ScalarType::Float8_e5m2) {     \
-          static at::cuda::tunable::ScaledGemmTunableOp<              \
-              at::Float8_e5m2, at::Float8_e5m2, scalar_t,             \
-              BLASOP_A, BLASOP_B> scaledgemm{};                       \
-          scaledgemm(&params);                                        \
-        }                                                             \
+#define TUNABLE_DISPATCH(BLASOP_A, BLASOP_B)                                      \
```
naromero77amd, line 1236: Let's add some comments about what this code is doing, what is supported and what is not supported.

aten/src/ATen/cuda/tunable/TunableGemm.h
```diff
@@ -326,4 +326,28 @@ class ScaledGemmTunableOp : public TunableOp<ScaledGemmParams<CT>> {
   }
 };
 
+template <typename AT, typename BT, typename CT, BlasOp ALayout, BlasOp BLayout>
```
naromero77amd, line 329: If I am following the logic, we need two ScaledGemm<blah>TunableOp classes:
- Need to still support traditional FP8 as supported in MI300 series
- Vesus supporting MX as in FP8 and FP4.

If this class can support Fp4 as well, can we please get rid of `Fp8` in the name and just call it
`ScaledGemmMxTunableOp` (just drop `Fp8` from the name).

aten/src/ATen/cuda/tunable/GemmHipblaslt.h
```diff
@@ -520,6 +523,20 @@ class HipblasltGemmOp : public Callable<ParamsT> {
       if (mat1_scale_ptr && mat2_scale_ptr) {
         hipblasLtMatmulDescAttributes_t a_scale_ptr_desc = HIPBLASLT_MATMUL_DESC_A_SCALE_POINTER;
         hipblasLtMatmulDescAttributes_t b_scale_ptr_desc = HIPBLASLT_MATMUL_DESC_B_SCALE_POINTER;
+
+        bool is_mx_fp8 = false;
+#if ROCM_VERSION >= 70000
```
naromero77amd, line 528: Is it because `HIPBLASLT_MATMUL_MATRIX_SCALE_VEC32_UE8M0`
 only exists in ROCm 70000 or later?

aten/src/ATen/cuda/tunable/GemmHipblaslt.h
```diff
@@ -682,6 +702,94 @@ auto GetHipBlasLtScaledGemmTypeStringAndOps() {
   return GetHipBlasLtTypeStringAndOps<AT, BT, CT, ALayout, BLayout, ScaledGemmParams<CT>>();
 }
 
+#if ROCM_VERSION >= 70000
```
naromero77amd, line 705: Not following why this #if #else block is needed.

Did you intend to commit this code block? I think the code should look similar to lines 700-703.

naromero77amd (commented):
There is some code that doesn't look right, can you please double-check?

Also had some other minor code comments.

PR #166201, opened 2025-10-24
[cuDNN][SDPA] Handle `c10:Error` when checking device capability for prefer-cuDNN SDPA check

eqy:
Fake device test can execute this function when the number of visible CUDA devices is 0, fix to unblock #165922

### DIFF
--- a/aten/src/ATen/native/transformers/cuda/sdp_utils.cpp
+++ b/aten/src/ATen/native/transformers/cuda/sdp_utils.cpp
@@ -81,9 +81,13 @@ bool check_prefer_cudnn_attention() {
     return false;
   }
 #if (defined(CUDNN_VERSION) && (CUDNN_VERSION >= 90900))
-  auto dprops = at::cuda::getCurrentDeviceProperties();
-  auto major = dprops->major;
-  return (major == 9 || major == 10) && !dprops->minor;
+  try {
+    auto dprops = at::cuda::getCurrentDeviceProperties();
+    auto major = dprops->major;
+    return (major == 9 || major == 10) && !dprops->minor;
+  } catch (c10::Error const& e) {
+    return false;
+  }
 #else
   return false;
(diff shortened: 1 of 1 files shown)


Skylion007 (approved):

aten/src/ATen/native/transformers/cuda/sdp_utils.cpp
```diff
@@ -76,14 +76,18 @@ bool priority_order_init_ = false;
...
   }
 #if (defined(CUDNN_VERSION) && (CUDNN_VERSION >= 90900))
-  auto dprops = at::cuda::getCurrentDeviceProperties();
-  auto major = dprops->major;
-  return (major == 9 || major == 10) && !dprops->minor;
+  try {
+    auto dprops = at::cuda::getCurrentDeviceProperties();
+    auto major = dprops->major;
+    return (major == 9 || major == 10) && !dprops->minor;
+  } catch (c10::Error const& e) {
```
Skylion007, line 88: Is there not a more specific error we can use?

eqy: I tried but it looks like `CUDAError` won't do it

Skylion007: @eqy Can we at least log the error then with DEBUG info?

eqy: For this PR or for debugging? I think this is just an edge case of `TestExportOnFakeCudaCUDA` where there's nothing we can/should do other than skip the device properties call

Skylion007: Errors should always be logged at some high verbosity level. Otherwise, how do we know debug if we accidentally trigger this condition.

PR #166203, opened 2025-10-24
Remove likely unnecessary _EXPAND trick for non-windows in HIDDEN_NAMESPACE_BEGIN

janeyx99:
I've learned that the EXPAND trick is needed mostly for an MSVC quirk to properly expand arguments. I tested on Linux locally and suspect that we don't need the _EXPAND for non-windows.  This PR is BE to minimalize what we need and remove what we don't, but I'm also okay not landing this if @malfet tells me that this quirk goes beyond MSVC.

* __->__ #166203
* #166079
* #166078
* #166077
* #166076

### DIFF
--- a/torch/headeronly/macros/Macros.h
+++ b/torch/headeronly/macros/Macros.h
@@ -619,10 +619,7 @@ __host__ __device__
 // loading custom extensions compiled against different libtorch
 // versions where these APIs may have changed.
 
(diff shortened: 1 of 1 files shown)


malfet (approved):
LGTM, as preprocessor macros have no effect in the runtime, i.e. if it compiles, ship it

PR #166224, opened 2025-10-25
Fix flake8 B028 warnings

cyyever:
This PR fixes flake8 B028 warning by specifying stacklevel=2 in `warnings.warn`. The advantage is that users can know more contextual information about PyTorch warnings.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @EikanWang @jgong5 @wenzhe-nrv @sanchitintel @mcarilli @ptrblck @leslie-fang-intel @ezyang

### DIFF
--- a/torch/__init__.py
+++ b/torch/__init__.py
@@ -2659,7 +2659,8 @@ def compile(
     if torch.compiler.is_exporting():
         warnings.warn(
             "You are calling torch.compile inside torch.export region. "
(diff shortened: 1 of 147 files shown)


ezyang (approved):
stacklevel=2 may not always be correct in all contexts but it sure as heck is better than not having it at all

PR #166232, opened 2025-10-25
Add FP32 upcasting for half-precision cross-entropy on large tensors

### LINKED ISSUE #85791: nn.CrossEntropyLoss overflow with FP16 and minibatch
### 🐛 Describe the bug

Using nn.CrossEntropyLoss with FP16 and long sequence is stable. However, introducing minibatch dimension would led to overflow and `CrossEntropyLoss` would output `inf`. 

To reproduce:

```Python
import torch

ce = torch.nn.CrossEntropyLoss().cuda().half()

inp = torch.rand((20, 14749, 1025))
inp = inp.cuda().half()
t = torch.randint(low=0, high=14749, size=[20, 1025]).cuda()

loss = ce(inp, t)
print(loss)

ce = torch.nn.CrossEntropyLoss().cuda()
inp = torch.rand((20, 14749, 1025))
inp = inp.cuda()

loss = ce(inp, t)
print(loss)

ce.half()
inp = inp.cuda().half()
inp = inp.transpose(1,2)
inp = inp.flatten(start_dim=0, end_dim=1)
t = t.flatten(start_dim=0, end_dim=1)

loss = ce(inp, t)
print(loss)
```

The first loss would be `inf`. Both the second and the third would be correct.

### Versions

I tested on 1.8.2 and 1.12.1, both are the same.

vishalgoyal316:
Automatically upcast FP16/BF16 inputs to FP32 when computing cross-entropy loss on large tensors (>10k elements) with reduction, then downcast the result. This improves numerical stability while maintaining performance for smaller tensors.

Fixes #85791

### DIFF
--- a/aten/src/ATen/native/LossNLL.cpp
+++ b/aten/src/ATen/native/LossNLL.cpp
@@ -656,12 +656,21 @@ Tensor cross_entropy_loss_symint(
     ret = cross_entropy_loss_label_smoothing(self, target, weight_, reduction, std::move(ignore_index), label_smoothing);
   } else {
     auto class_dim = self.dim() == 1 ? 0 : 1;
+
+    bool is_half_precision = (self.scalar_type() == ScalarType::Half ||
+                              self.scalar_type() == ScalarType::BFloat16);
+    bool needs_fp32_compute = is_half_precision && reduction != Reduction::None && self.numel() > 10000;
+
+    auto compute_input = needs_fp32_compute ? self.to(ScalarType::Float) : self;
+
+    auto log_probs = at::log_softmax(compute_input, class_dim, compute_input.scalar_type());
+
     ret = at::nll_loss_nd_symint(
-        at::log_softmax(self, class_dim, self.scalar_type()),
-        target,
-        weight,
-        reduction,
-        std::move(ignore_index));
+      log_probs, target, weight, reduction, std::move(ignore_index));
+
+    if (needs_fp32_compute) {
+      ret = ret.to(self.scalar_type());
+    }
   }
   return ret;
 }
--- a/aten/src/ATen/native/LossNLL.cpp
+++ b/aten/src/ATen/native/LossNLL.cpp
@@ -316,8 +316,23 @@ void nll_loss_forward_out_cpu_template(
   const auto input_dtype = input.scalar_type();
   const bool is_half_precision = (input_dtype == ScalarType::Half || input_dtype == ScalarType::BFloat16);
 
-  // Fast path for FP32 or reduction='none' (common case)
-  if (C10_LIKELY(!is_half_precision || reduction == Reduction::None)) {
+  // Fast path for FP32/FP64 (common case)
+  if (C10_LIKELY(!is_half_precision)) {
+    AT_DISPATCH_FLOATING_TYPES(
+        input_dtype, "nll_loss_out_frame", [&] {
+          if (target.scalar_type() == kByte) {
+            nll_loss_out_frame<scalar_t, uint8_t>(
+                output, total_weight, input, target, weight, reduction, ignore_index);
+          } else {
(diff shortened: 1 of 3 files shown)


vadimkantorov:
Might it be better to simply always upcast? (and fuse the upcast inside the CE nll_loss function itself to avoid allocation of a temp tensor - can be a huge memory cost for large vocabs)

vishalgoyal316:
> Might it be better to simply always upcast? (and fuse the upcast inside the CE nll_loss function itself to avoid allocation of a temp tensor - can be a huge memory cost for large vocabs)

@vadimkantorov, you are right, I have updated the code based on your feedback, thanks!

vishalgoyal316:
> Why not just do the accumulation in high precision but do the rest of the computation in low precision?

@Skylion007, it is also great idea. I explored it, can I push the code with this approach?

aten/src/ATen/native/LossNLL.cpp
```diff
@@ -652,12 +652,54 @@ Tensor cross_entropy_loss_symint(
...
+      if (weight.has_value()) {
+        weight_half = weight.value().to(self.scalar_type());
+      }
+
+      auto unreduced_loss = at::nll_loss_nd_symint(
+          log_probs, target, weight_half, Reduction::None, ignore_index);
+
+      // upcast to FP32 only for the reduction step
+      auto loss_fp32 = unreduced_loss.to(ScalarType::Float);
+      auto sum_loss = loss_fp32.sum();
```
isuruf:
```suggestion
      auto sum_loss = unreduced_loss.sum(ScalarType::Float);
```

isuruf:
Is this issue specific to `nn.CrossEntropyLoss`? Is it there for `nn.NLLLoss` too? If so, I think the issue should be fixed in `nll_loss*` functions.

aten/src/ATen/native/LossNLL.cpp
```diff
@@ -308,33 +308,63 @@ void nll_loss_forward_out_cpu_template(
...
     int64_t reduction,
     int64_t ignore_index) {
+
+  const auto input_dtype = input.scalar_type();
+  const bool is_half_precision = (input_dtype == ScalarType::Half || input_dtype == ScalarType::BFloat16);
+
+  // Fast path for FP32 or reduction='none' (common case)
+  if (C10_LIKELY(!is_half_precision || reduction == Reduction::None)) {
+    AT_DISPATCH_FLOATING_TYPES_AND2(
+        ScalarType::BFloat16, ScalarType::Half, input_dtype, "nll_loss_out_frame", [&] {
```
isuruf, line 337: We don't need dispatching to `BFloat16` and `Half` here.

aten/src/ATen/native/LossNLL.cpp
```diff
@@ -308,33 +308,63 @@ void nll_loss_forward_out_cpu_template(
...
+  // FP16/BF16 with reduction: compute element-wise loss in native precision
+  // but perform reduction in FP32 to prevent overflow
+
+  // Convert weight to input dtype for element-wise operations if needed
+  const Tensor& weight_for_loss = weight.defined() && weight.scalar_type() != input_dtype
+      ? weight.to(input_dtype) : weight;
+
+  // Compute unreduced loss in native precision
+  auto output_unreduced = at::empty_like(input.select(-1, 0));
   AT_DISPATCH_FLOATING_TYPES_AND2(
```
isuruf: Here, it's the other way. We don't need dispatching to other floating types.

aten/src/ATen/native/LossNLL.cpp
vishalgoyal316: Thanks, let me do the same.

aten/src/ATen/native/LossNLL.cpp
vishalgoyal316: Sure.

PR #166235, opened 2025-10-25
[OpenReg] Update Installation in README.md

can-gaa-hou:
It is recommended to use `python -m pip install --no-build-isolation .` instead of `pip3 install --no-build-isolation .` because most of us use a virtual environment, and the latter probably relies on the system `pip3` rather than the conda or uv. We need to make it consistent with the Python we use, and it is also consistent with how `torch` is installed.

cc @FFFrog @albanD


fffrog (dismissed):
SGTM, thank you.

fffrog (approved):

ezyang (approved):

PR #166246, opened 2025-10-26
[BugFix] Fix exponential distribution generating zero values on CPU backend

### LINKED ISSUE #166243: [BugReport] Exponential distribution may generate zero values on CPU backend
### 🐛 Describe the bug

aspired by https://github.com/pytorch/pytorch/pull/159386
I find https://github.com/pytorch/pytorch/blob/6494cdc40c55a7ff5d2c94788f7fbf5d4822ae7c/aten/src/ATen/native/Distributions.cpp#L614-L620
and then I find https://github.com/pytorch/pytorch/blob/6494cdc40c55a7ff5d2c94788f7fbf5d4822ae7c/aten/src/ATen/core/TransformationHelper.h#L128-L145

This is a historical problem.

On CPU, there is a very low probability to generate 0 from exponential.Though it's pretty low, it must be resolved and unified.

### Versions

None

lingebeng:
Fix exponential distribution generating zero values on CPU backend and unified the two backends.

Fixes [#166243](https://github.com/pytorch/pytorch/issues/166243)

### DIFF
--- a/aten/src/ATen/core/TransformationHelper.h
+++ b/aten/src/ATen/core/TransformationHelper.h
@@ -127,21 +127,10 @@ C10_HOST_DEVICE inline double cauchy(double val, double median, double sigma) {
 template <typename T>
 C10_HOST_DEVICE inline T exponential(T val, T lambda) {
   // https://en.wikipedia.org/wiki/Exponential_distribution#Generating_exponential_variates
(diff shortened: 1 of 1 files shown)


ngimel (changes requested):
Please don't touch cuda implementation

lingebeng:
I am so sorry. I won't do that.

PR #166249, opened 2025-10-26
Fix error suppression syntax in onnx, jit, _dynamo

maggiemoss:
Ensures pyrefly will only silence one specific error code 

pyrefly check 
lintrunner

cc @EikanWang @jgong5 @wenzhe-nrv @sanchitintel @ezyang @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela


oulgen (approved):

torch/onnx/_internal/exporter/_tensors.py
```diff
@@ -30,16 +30,16 @@ def __init__(
 
     @property
     def rank(self) -> int | None:
-        # pyrefly: ignore  # missing-attribute
+        # pyrefly: ignore [missing-attribute]
         if self.shape is None:
             return None
-        # pyrefly: ignore  # bad-argument-type
+        # pyrefly: ignore [bad-argument-type]
```
justinchuby, line 36: Curious why there is a type issue in the first place?

torch/onnx/_internal/exporter/_registration.py
```diff
@@ -64,11 +64,11 @@ def __post_init__(self) -> None:
                 if isinstance(self.onnx_function, onnxscript.OnnxFunction):
                     signature = _schemas.OpSignature.from_function(  # type: ignore[attr-defined]
                         self.onnx_function,
-                        # pyrefly: ignore  # missing-attribute
+                        # pyrefly: ignore [missing-attribute]
```
justinchuby, line 67: Is the idea to replace mypy, or have both linters running?

maggiemoss: The idea is to eventually replace mypy, fixing these syntax issues first

PR #166254, opened 2025-10-26
Enable flake8 B028 check

cyyever:
This PR is a follow-up of #166224. It fixes remaining flake8 B028 warning by specifying `stacklevel=2` in warnings.warn and enables the check. The advantage is that users can know more contextual information about PyTorch warnings.

cc @EikanWang @jgong5 @wenzhe-nrv @sanchitintel @ezyang @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @Lucaskabela @chenyang78

### DIFF
--- a/torch/_dynamo/variables/higher_order_ops.py
+++ b/torch/_dynamo/variables/higher_order_ops.py
@@ -2199,7 +2199,6 @@ class CondHigherOrderVariable(TorchHigherOrderOperatorVariable):
                 "Pred is a Python constant. When used with torch.cond, it specializes on one of the branches."
                 " If you want torch.cond to preserve two branches, please make the predicate a boolean tensor or a SymBool.",
                 UserWarning,
-                stacklevel=2,
             )
             if pred.as_python_constant():
                 return true_fn.call_function(tx, operands.unpack_var_sequence(tx), {})
--- a/torch/_dynamo/variables/higher_order_ops.py
+++ b/torch/_dynamo/variables/higher_order_ops.py
@@ -2195,7 +2195,7 @@ class CondHigherOrderVariable(TorchHigherOrderOperatorVariable):
         # Specialize into one of the branches since pred is constant
         pred, true_fn, false_fn, operands = args
         if type(args[0]) is ConstantVariable:
-            warnings.warn(
+            warnings.warn(  # noqa: B028
                 "Pred is a Python constant. When used with torch.cond, it specializes on one of the branches."
                 " If you want torch.cond to preserve two branches, please make the predicate a boolean tensor or a SymBool.",
                 UserWarning,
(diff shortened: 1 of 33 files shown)


ezyang (changes requested):
Tests should not be modified for this lint. Exclude this lint from the test folder.

.github/scripts/filter_test_configs.py
```diff
@@ -124,7 +124,7 @@ def get_pr_info(pr_number: int) -> dict[str, Any]:
     )
 
     if not json_response:
-        warnings.warn(f"Failed to get the labels for #{pr_number}")
+        warnings.warn(f"Failed to get the labels for #{pr_number}", stacklevel=2)
```
ezyang: need to exclude .github and scripts as well

ezyang (changes requested):
please review your PR first

.github/scripts/filter_test_configs.py
cyyever: I was not sure because the .github was not excluded before. But it's reasonable to exclude it.

cyyever:
@ezyang  `.github` was not excluded before in flake8, should we also exclude it for existing checks? IMO it should be excluded, maybe we should put the config changes for other checks in another PR.

torch/_dynamo/variables/higher_order_ops.py
```diff
@@ -1804,6 +1804,7 @@ def _call_function(
                 "Pred is a Python constant. When used with torch.cond, it specializes on one of the branches."
                 " If you want torch.cond to preserve two branches, please make the predicate a boolean tensor or a SymBool.",
                 UserWarning,
+                stacklevel=2,
```
ezyang: This one looks wrong. But all of the Dynamo warnings are trouble for a different reason: we don't want to report the Dynamo stack, we want to report the user stack.

torch/random.py
```diff
@@ -203,5 +203,5 @@ def fork_rng(
         yield
     finally:
         torch.set_rng_state(cpu_rng_state)
-        for device, device_rng_state in zip(devices, device_rng_states):
+        for device, device_rng_state in zip(devices, device_rng_states, strict=True):
```
ezyang: unrelated

torch/_dynamo/variables/higher_order_ops.py
cyyever: Reverted

torch/random.py
cyyever: reverted

cyyever:
@ezyang fixed

PR #166257, opened 2025-10-26
[2/N] Add strict parameter to Python zip calls 

cyyever:
This PR adds `strict=True/False` to zip calls in test utils. strict=True is passed when possible.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @nikitaved @pearu @cpuhrsch @amjames @bhosmer @jcaip @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @aditew01 @robieta @chaekit @guotuofeng @guyang3532 @dzhulgakov @davidberard98 @briancoutinho @sraikund16 @sanrise @alexsamardzic


torch/nn/modules/adaptive.py
```diff
@@ -273,7 +273,9 @@ def _get_full_log_prob(self, input, head_output):
 
         out[:, : self.shortlist_size] = head_logprob[:, : self.shortlist_size]
 
-        for i, (start_idx, stop_idx) in enumerate(zip(self.cutoffs, self.cutoffs[1:])):
+        for i, (start_idx, stop_idx) in enumerate(
+            zip(self.cutoffs, self.cutoffs[1:], strict=True)
```
Skylion007: This should actually itertools pairwise? And it will fail when strict is True anyway?

torch/nn/modules/container.py
```diff
@@ -150,7 +150,9 @@ def __delitem__(self, idx: Union[slice, int]) -> None:
             delattr(self, key)
         # To preserve numbering
         str_indices = [str(i) for i in range(len(self._modules))]
-        self._modules = OrderedDict(list(zip(str_indices, self._modules.values())))
+        self._modules = OrderedDict(
+            list(zip(str_indices, self._modules.values(), strict=True))
```
Skylion007: I don't know if the list call is necessary for OrderedDict

torch/sparse/_triton_ops_meta.py
```diff
@@ -289,7 +297,7 @@ def to_key(parameters):
         return tuple(parameters[k] for k in sorted(parameters))
 
     def from_key(key, parameters):
-        return dict(zip(sorted(parameters), key))
+        return dict(zip(sorted(parameters), key, strict=True))
```
Skylion007, line 300: Is this a typo in the original code, why are the parameters sorted without sorting with keys? This will put the keys in a non deterministic order from parameters?

torch/nn/modules/adaptive.py
cyyever: Yes, I should fix it.

torch/nn/modules/container.py
cyyever: No need.

torch/nn/modules/adaptive.py
```diff
@@ -273,7 +273,9 @@ def _get_full_log_prob(self, input, head_output):
 
         out[:, : self.shortlist_size] = head_logprob[:, : self.shortlist_size]
 
-        for i, (start_idx, stop_idx) in enumerate(zip(self.cutoffs, self.cutoffs[1:])):
+        for i, (start_idx, stop_idx) in enumerate(
+            zip(self.cutoffs, self.cutoffs[1:], strict=False)
```
Skylion007: Replace with itertools.pairwise(self.cutoffs), it's more efficient, readable, and equivalent.

torch/sparse/_triton_ops_meta.py
cyyever: I checked the call site. From the context I found that the passed `key` is a sorted sequence.

torch/nn/modules/adaptive.py
cyyever: Fixed

janeyx99 (approved):

PR #166258, opened 2025-10-26
[1/N] Remove unused loop variables

cyyever:
This PR removes unused loop variables.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @EikanWang @jgong5 @wenzhe-nrv @sanchitintel @ezyang @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela @xmfan

### DIFF
--- a/torch/_dynamo/guards.py
+++ b/torch/_dynamo/guards.py
@@ -317,7 +317,7 @@ class GuardManagerWrapper:
             is_diff_guard_node = (
                 node.get_source() in self.diff_guard_sources or node.fail_count() > 0
             )
-            for idx, (key_mgr, val_mgr) in sorted(
+            for _idx, (key_mgr, val_mgr) in sorted(
                 node.get_key_value_managers().items()
             ):
                 is_diff_guard_node |= visit(key_mgr) | visit(val_mgr)
@@ -440,7 +440,7 @@ class GuardManagerWrapper:
             is_subtree_tag_safe = True
 
             # Recurse to get the tag safe roots from subtree.
-            for idx, (key_mgr, val_mgr) in sorted(
+            for _idx, (key_mgr, val_mgr) in sorted(
                 node.get_key_value_managers().items()
             ):
                 if key_mgr is not None:
@@ -448,9 +448,7 @@ class GuardManagerWrapper:
                 if val_mgr is not None:
                     tag_safe_roots.extend(visit(val_mgr))
 
-            for idx, (key_mgr, val_mgr) in sorted(
-                node.get_key_value_managers().items()
-            ):
+            for key_mgr, val_mgr in node.get_key_value_managers().values():
                 if key_mgr:
                     is_subtree_tag_safe &= key_mgr.is_tag_safe()
 
--- a/torch/_inductor/comms_debug.py
+++ b/torch/_inductor/comms_debug.py
@@ -46,7 +46,7 @@ def _debug_iterative_memory_recompute(
     if iter_cm != new_cm:
         log = "ITERATIVE CURR MEMORY CANDIDATE DOES NOT MATCH"
         iterative_recompute_error = True
-    for i, gn in enumerate(gns):
+    for gn in gns:
         iter_gnm = iter_curr_memory[gn]
         new_gnm = est_curr_memory[gn]
         if iter_gnm != new_gnm:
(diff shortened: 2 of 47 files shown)


torch/_subclasses/fake_tensor.py
```diff
@@ -3118,7 +3118,7 @@ def _validate_symbolic_output_for_caching(
     if is_tracing:
         # Check for SymNode types in PROXY mode - this should bypass caching
         # regardless of whether symbols are known or not
-        for node in _iterate_nodes(output):
+        for _node in _iterate_nodes(output):
```
Lucaskabela: Should this be replaced here to `for _ in _iterate_nodes(output)`?

torch/_inductor/scheduler.py
```diff
@@ -3339,7 +3339,7 @@ def compile_kernel(
             future_choices: list[tuple[Any, Optional[LambdaFuture], ModuleType]] = []
             for hint_override in config.multi_kernel_hints:
                 choice_timings = multi_node.choice_timings(hint_override)
-                for choice, unfused_time in sorted(
+                for choice, _unfused_time in sorted(
```
Lucaskabela: Similar here - should `unfused_time` be entirely replaced to `_`

torch/_inductor/comms_debug.py
```diff
@@ -65,7 +65,7 @@ def _debug_iterative_memory_recompute(
             f"\nCANDIDATE_NEW_ALLOCFREE:{snodes_allocfree[candidate]}"
         )
         peak_log = ""
-        for i, (pre, post) in enumerate(snodes_curr_memory):
+        for i, (pre, _post) in enumerate(snodes_curr_memory):
```
Lucaskabela, line 68: Similar though on unused (let's replace with `_`

torch/_inductor/fx_passes/post_grad.py
```diff
@@ -1265,7 +1265,7 @@ def decomp(*flat_args):
 
     graph_pass.apply(graph)
 
-    for node in graph.find_nodes(
+    for _node in graph.find_nodes(
```
Lucaskabela: -> `_`

torch/_dynamo/guards.py
```diff
@@ -317,7 +317,7 @@ def visit_dict_manager(node: DictGuardManager) -> bool:
             is_diff_guard_node = (
                 node.get_source() in self.diff_guard_sources or node.fail_count() > 0
             )
-            for idx, (key_mgr, val_mgr) in sorted(
+            for _idx, (key_mgr, val_mgr) in sorted(
```
Lucaskabela, line 320: -> `_`

Lucaskabela (changes requested):
See comments - I think we can use `_` as opposed to just prefixing the variables here

I am also curious - how did we find and fix these errors? Turn on the rule then manual? Or did we use some rule/programmatic way?

Thanks!

torch/_inductor/comms_debug.py
Skylion007: I mean more descriptive names are always good, and the underscore prefix is pretty standard here for unused variables in structure binding outputs.

torch/_dynamo/guards.py
Skylion007: @Lucaskabela Why the previous name is more descriptive and the underscore indicates the variable is unused per the pep documentation

cyyever:
@Lucaskabela I think keeping the original index name is better except in simple cases such as `for i in rang(10)`, where I replaced `i` with `_`. The check was done by flake8 B007 rule and I used ruff implementation to scan the code base.

torch/_inductor/fx_passes/post_grad.py
cyyever: Fixed

torch/_subclasses/fake_tensor.py
cyyever: Fixed

torch/_inductor/comms_debug.py
cyyever: I prefer keeping it

torch/_dynamo/guards.py
cyyever: I prefer keeping it

mlazos (approved):

Lucaskabela (approved):
I think that is a valid point - everything else lgtm so approving

PR #166263, opened 2025-10-26
[PP] Free send ops earlier

H-Huang:
* __->__ #166263

Release send buffers as soon as the async send operations complete, rather than waiting until the end of the step. This allows the send buffers to be released earlier.

### DIFF
--- a/torch/distributed/pipelining/schedules.py
+++ b/torch/distributed/pipelining/schedules.py
@@ -2191,6 +2191,17 @@ BACKWARD_INPUT, BACKWARD_WEIGHT, and OVERLAP_F_B are supported."
         # count either full_backward or backward_weight together, to determine when to sync DP grads
         self.backward_counter.clear()
         for time_step, action in enumerate(self.pipeline_order_with_comms[self.rank]):
+            # Remove completed send operations from list, this frees the memory used for the send buffer earlier
+            for work_batch in send_ops:
+                completed_ops = []
+                for idx, send_op in enumerate(work_batch):
+                    if send_op.is_completed():
+                        send_op.wait()
+                        completed_ops.append(idx)
+                # Remove completed ops (reverse to maintain indices)
+                for idx in reversed(completed_ops):
+                    del work_batch[idx]
+
             try:
                 with record_function(_get_profiler_function_name(action)):
                     if action.computation_type in self._comp_type_to_function_map:


Skylion007 (approved):

torch/distributed/pipelining/schedules.py
```diff
@@ -2191,6 +2191,17 @@ def _perform_action(action: _Action) -> None:
         # count either full_backward or backward_weight together, to determine when to sync DP grads
         self.backward_counter.clear()
         for time_step, action in enumerate(self.pipeline_order_with_comms[self.rank]):
+            # Remove completed send operations from list, this frees the memory used for the send buffer earlier
+            for work_batch in send_ops:
+                completed_ops = []
+                for idx, send_op in enumerate(work_batch):
```
wconstab, line 2197: @kwen2501 is it logical to wait on, and later delete, individual send ops that are part of a batch?  if they are batched, they would all complete together, and i'm not sure if the individual handles even get updated.  Should we instead query if the whole batch is completed?

kwen2501: A couple things:
1. 
```
                    if send_op.is_completed():
                        send_op.wait()
```
`is_completed()` is a CPU side check IIRC. I am not sure why we issue `wait()` **after** the CPU observes completion. 

2.
```
                # Remove completed ops (reverse to maintain indices)
                for idx in reversed(completed_ops):
                    del work_batch[idx]
```
This changes the `work_batch` list in the middle of the iteration. Not entirely recommended but maybe the `reversed` helped.

3. Finally to your question, I think today a `Work` holding a reference to the output buffer is somewhat problematic. 

https://github.com/pytorch/pytorch/blob/99e07c39ec3e9cf4573da6e6737c3d0362bc6b13/torch/csrc/distributed/c10d/ProcessGroupNCCL.cpp#L3695-L3696

I guess that is one reason for the deletion here.

wconstab: I can say that the goals here are (1) free the send buffer (by deleting the work obj), and (2) NOT blocking the CPU on the send (hence, never calling .wait() until completion is known).

To your first point, should we not call .wait() after observing completion?  Technically, I think skipping a wait would be unsafe due to not synchronizing the streams, except this is a send operation, so I think this omission is harmless.

It might be OK to just delete the work objects after determining they are completed?

kwen2501: My point is: there is rarely a need to observe completion from CPU perspective -- so we should not call `is_completed()` at all. We only need to call `.wait()` before we'd need to modify the send buffer again. So yes, no need to call it immediately after send, if you don't want to block computation.

kwen2501: You can even delete a `Work` object before it `is_completed()`. 
```
work.wait()
del work
```
This is totally fine.

wconstab: we can't call work.wait() eagerly, becuase if the work is not done yet we'd block subsequent GPU operations on this rank's PP schedule, which should be issued before the send is completed.

The trouble is, there is no point in our schedule where we 'know' the send has completed on GPU, since we are running ahead.  At the same time, it is too late if we leave all the send buffers unfreed until the end of the whole schedule step.

This PR is an attempt to find a middle ground, by freeing send buffers as soon as they are finished with, but without ever blocking the GPU on any of them.  

> there is rarely a need to observe completion from CPU perspective 

maybe this is one of those rare cases?  :) 

we do not ever want the GPU to wait on the send; but we do want the CPU to free the send buffer as soon as it has been used.

kwen2501: But at today's call point would `is_completed()` be always false since the CPU is running way ahead?

PR #166271, opened 2025-10-26
[MPS] fix large matmul test device

Isalia20:
PR is self explanatory
Test was introduced by https://github.com/pytorch/pytorch/pull/143095 and was always running on CPU

### DIFF
--- a/test/test_mps.py
+++ b/test/test_mps.py
@@ -634,6 +634,7 @@ class MatmulTest(TestCaseMPS):
     def test_batched_matrix_x_broadcasted_matrix(self):
         self._helper((10, 3, 4), (4, 5))
 
+    @serialTest()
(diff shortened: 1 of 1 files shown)


kulinseth (approved):

test/test_mps.py
```diff
@@ -636,8 +636,8 @@ def test_batched_matrix_x_broadcasted_matrix(self):
 
     def test_large_matmul(self):
```
malfet, line 638:
```suggestion
    @serialTest()
    def test_large_matmul(self):
```

malfet (approved):

test/test_mps.py
Skylion007: Don't we have a big test pytest guard too?

PR #166272, opened 2025-10-26
[MPS] Better error checking for FFT ops

malfet:
* #166273
* #166215
* #166214
* __->__ #166272

Namely, error out rather than crash when out dtype is of an unexpected type
Resize output tensor to the expected size in `_out` operation, to prevent crash when tensor of an unexpected size is passed.
Preserve symbolic shapes whenever possible

Test plan: Run `python test_ops.py -v -k test_out_warning_fft_hfft_mps` for MPS device, without this change it crashes with `Error: Invalid KernelDAG, equalShape for destination failed'`, run `python ../test/test_ops.py -v -k test_dtypes_stft_mps`, without this change it crashes with `A complex mlir::Type does not have a corresponding complex MPSDataType"`, when input dtype is bfloat16


kulinseth (approved):

aten/src/ATen/native/mps/operations/FastFourierTransform.mm
```diff
@@ -112,6 +110,12 @@ Tensor _fft_c2c_mps(const Tensor& self, IntArrayRef dim, int64_t normalization,
                          int64_t normalization,
                          int64_t last_dim_size,
                          Tensor& out) {
+  TORCH_CHECK(self.is_complex(), "Input must be complex");
+  TORCH_CHECK(out.scalar_type() == c10::toRealValueType(self.scalar_type()), "Unexpected output type");
+  auto in_sizes = self.sizes();
```
Skylion007: Why not ensure symbolic shape support with sym_sizes? And SymDimVector?

PR #166275, opened 2025-10-26
[Inductor] Prevent kernel fusion with too many unique inputs and outputs

andyanwang:
MTIA triton currently has a limit that it can't support the cases when there are too many input/output buffers. This PR adds the limitation to prevent large fusion with many input/output buffer.

* #166276
* __->__ #166275
* #166274

Differential Revision: [D85509351](https://our.internmc.facebook.com/intern/diff/D85509351/)

### DIFF
--- a/torch/_inductor/scheduler.py
+++ b/torch/_inductor/scheduler.py
@@ -3882,6 +3882,15 @@ class Scheduler:
         if V.graph.sizevars.statically_known_gt(memory_overhead, 32 * bw_saving):
             return True
         return False
+    
+    def fusion_prevent_too_many_reads_and_writes(
+        self, node1: BaseSchedulerNode, node2: BaseSchedulerNode, threshold: int
+    ) -> bool:
+        unique_reads = (node1.read_writes.reads | node2.read_writes.reads) - (
+            node1.read_writes.writes | node2.read_writes.writes
+        )
+        unique_writes = (node2.read_writes.writes) - (node1.read_writes.reads | node2.read_writes.reads)
+        return len(unique_reads) + len(unique_writes) > threshold
 
     def are_long_distant_nodes(
         self, node1: BaseSchedulerNode, node2: BaseSchedulerNode
--- a/torch/_inductor/choices.py
+++ b/torch/_inductor/choices.py
@@ -527,6 +527,17 @@ class InductorChoices:
             WhyNoFuse(node1, node2)("Fusion will increase peak memory")
             return False
 
+        if (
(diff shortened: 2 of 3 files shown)


torch/_inductor/scheduler.py
```diff
@@ -3882,6 +3882,15 @@ def _find_single_user_inputs(
...
             return True
         return False
+    
+    def fusion_prevent_too_many_reads_and_writes(
+        self, node1: BaseSchedulerNode, node2: BaseSchedulerNode, threshold: int
+    ) -> bool:
+        unique_reads = (node1.read_writes.reads | node2.read_writes.reads) - (
+            node1.read_writes.writes | node2.read_writes.writes
+        )
+        unique_writes = (node2.read_writes.writes) - (node1.read_writes.reads | node2.read_writes.reads)
```
eellison: We want all unique reads & writes. To dedpulicate reads/writes, we can use a set as you are doing. But we also need to account for all of the read/writes that would be removed through fusion. 

the reads that are removed through fusion are `node2 reads - node1 writes`.

For the writes, you need to check which of node1's write are [removed through fusion](https://github.com/pytorch/pytorch/blob/f89a7e9fe80fb0f74af52bad892e1afafa57be9c/torch/_inductor/scheduler.py#L4700-L4711). A node1 write may have other readers besides node2. 

once you have the unique reads and writes, you can get all of the unique buffer names to get total io bufs.

eellison (commented):
test ?

torch/_inductor/scheduler.py
```diff
@@ -3883,6 +3883,54 @@ def _find_single_user_inputs(
             return True
         return False
 
+    def fusion_prevent_too_many_reads_and_writes(
```
mlazos, line 3886: This computation is identical to what we compute when fusing two nodes, I think you can probably use the same logic as https://github.com/pytorch/pytorch/blob/37d57ac9cb7f538b812cf1d9851b55b46213fe15/torch/_inductor/dependencies.py#L441 right?

torch/_inductor/scheduler.py
andyanwang: Updated!

torch/_inductor/scheduler.py
andyanwang: let me look into that

andyanwang: @mlazos , I'm not sure if I understand correctly, but the logic of `merge` seems a bit different, as it doesn't remove node2 writes that can be removed.

eellison (approved):

PR #166276, opened 2025-10-26
[Inductor] Prevent large fusion for aten::cat

andyanwang:
Same context as in https://github.com/pytorch/pytorch/pull/166275 :
MTIA triton currently has a limit that it can't support the cases when there are too many input/output buffers. 

This PR adds the limitation to prevent large fusion with many input/output buffer, specifically for aten::cat operator.

* __->__ #166276

in addition to D84571005

Differential Revision: [D84569512](https://our.internmc.facebook.com/intern/diff/D84569512/)


torch/_inductor/lowering.py
```diff
@@ -1846,6 +1859,26 @@ def op_count(x):
...
         return op in (aten.cat.default, aten.constant_pad_nd.default)
 
+    def count_input_buffers(x):
+        """Count the number of input buffers that would be read after fusion."""
+        if isinstance(x, (TensorBox, ir.StorageBox)):
+            return count_input_buffers(unwrap_tensor(x))
+
+        # If it's not a Pointwise operation, it will be a single input buffer
+        if not isinstance(x, ir.Pointwise):
+            return 1
```
eellison: We will also fuse in reductions.. check for `ir.Loops`. See torch.cat(softmax, softmax) + 2

torch/_inductor/lowering.py
```diff
@@ -1846,6 +1859,26 @@ def op_count(x):
...
+
+        # For Pointwise operations, count all the buffers they read
+        total = 0
+        for read in x.get_read_names():
+            total += count_input_buffers(V.graph.get_buffer(read))
```
eellison: Don't we need to recursively check these buffers inputs ?

torch/_inductor/lowering.py
```diff
@@ -1846,6 +1859,26 @@ def op_count(x):
     def additional_pointwise_ops(op: torch._ops.OpOverload):
         return op in (aten.cat.default, aten.constant_pad_nd.default)
 
+    def count_input_buffers(x):
```
eellison: I think we should be getting unique name inputs, then taking sum

andyanwang: I've changed to use a set `counted_buffers` to dedup the input already counted

torch/_inductor/lowering.py
andyanwang: Got it. Changed to check `ir.Loops` instead

torch/_inductor/lowering.py
```diff
@@ -1847,6 +1860,39 @@ def op_count(x):
     def additional_pointwise_ops(op: torch._ops.OpOverload):
         return op in (aten.cat.default, aten.constant_pad_nd.default)
 
+    def count_input_buffers(x, counted_buffers=None):
+        """Count the number of unique input buffers that would be read after fusion."""
+        if counted_buffers is None:
+            counted_buffers = OrderedSet()
```
eellison, line 1866: can we always pass in counted_buffers, and not return anythign ? This function is written in a way that's sort of unclear if we're writing to the input, or returning an output.

torch/_inductor/lowering.py
```diff
@@ -1847,6 +1860,39 @@ def op_count(x):
...
+
+        if isinstance(x, (TensorBox, ir.StorageBox)):
+            return count_input_buffers(unwrap_tensor(x), counted_buffers)
+
+        # If it's not a fusable operation (Loops which includes Pointwise and Reduction),
+        # it will be a single input buffer that we need to count if not already counted
+        if not isinstance(x, ir.Loops):
+            # Get a unique identifier for this buffer
+            buffer_name = getattr(x, "name", id(x))
```
eellison, line 1875: do you want to factor out htis and call it ? https://github.com/pytorch/pytorch/blob/95a25fd072b1a36654a8c6d059b9b253143ca099/torch/_inductor/lowering.py#L3970-L3977

torch/_inductor/lowering.py
```diff
@@ -1847,6 +1860,39 @@ def op_count(x):
...
+                counted_buffers.add(buffer_name)
+                return 1
+            return 0
+
+        # For fusable operations (Pointwise, Reduction), recursively count their input buffers
+        total = 0
+        for read in x.get_read_names():
+            total += count_input_buffers(V.graph.get_buffer(read), counted_buffers)
+
+        return total
```
eellison, line 1886: We should be recursively gathering all inputs read, and then counting, so that we dont double count.

eellison (commented):
Not sure we're doing deduping correctly

PR #166278, opened 2025-10-26
Remove not needed code path. 

laithsakka:
* __->__ #166278

I accepted a PR that added this code, but re-examining it now, I'm questioning the approach. It seems like we're working around an issue with the inductor generating incorrect sizes. A comment suggests it might be related to unsqueezed u0 values. Removing this code didn't cause any failures, so I'll take it out and address the root issue if it arises.


Lucaskabela (approved):
Thanks for minimizing - I think we are correct in saying this change wasn't needed

PR #166294, opened 2025-10-27
[hops] fix cond subgraph runtime asserts from polluting main graph

bobrenjc93:
* __->__ #166294

The main issue addressed by this PR can be illustrated by the following example:

```
def compute(x, w):
    return torch.nn.functional.linear(x, w)

def nop(x, w):
    torch._check(x.shape[0] == 0)
    return torch.empty_like(x) 

def chunked_compute(x, w):
    return torch.cond(x.shape[0] > 0, compute, nop, (x, w))
```

Here, the runtime assertion `x.shape[0] == 0` should only apply in the false branch of the conditional. However, the current behavior incorrectly “leaks” this information into the global ShapeEnv during tracing. This happens because the runtime assert in the subgraph updates the global shape state unconditionally.

This PR fixes that by making all runtime asserts inside subgraphs conditional on the predicate of the outer torch.cond. In the example above, the traced subgraph now includes an assert equivalent to:

```
torch._check(x.shape[0] <= 0 and x.shape[0] == 0)
```

This ensures the assertion remains valid in its local context without polluting global shape information.

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @ezyang @EikanWang @wenzhe-nrv @voznesenskym @penguinwu @Guobing-Chen @zhuhaozhe @blzheng @jiayisunx @chenyang78 @kadeng @chauhang @amjames @Lucaskabela


torch/fx/experimental/symbolic_shapes.py
```diff
@@ -7939,6 +7954,21 @@ def constrain_symbol_range(
...
+    @contextmanager
+    def patch_ra_prelude(self, prelude: sympy.Expr) -> Iterator[None]:
+        """
+        Context manager that ensures all runtime asserts generated while this context manager
+        is active include a prelude expression. This is mainly used in torch.cond to guarantee
+        that runtime asserts in subgraphs are guarded by the original cond predicate, preventing
+        them from leaking into the main graph.
+        """
+        prev = self.ra_prelude
+        self.ra_prelude = prelude
```
aorenste, line 7970: This needs to 'and' with the previous ra_prelude, doesn't it (i'm thinking about nested torch.cond statements)

PR #166302, opened 2025-10-27
Add input_size validation for RNN modules with clear error messages

### LINKED ISSUE #136936: `input_size` argument of `nn.RNN()` gets indirect error messages
### 🐛 Describe the bug

Setting the float value `3.` to `input_size` of [nn.RNN()](https://pytorch.org/docs/stable/generated/torch.nn.RNN.html) gets the indirect error message as shown below:

```python
import torch
from torch import nn

my_tensor = torch.tensor([[8., -3., 5.]])

torch.manual_seed(42)
                      # ↓↓
rnn = nn.RNN(input_size=3., hidden_size=2) # Error

rnn(input=my_tensor)
```

> TypeError: empty(): argument 'size' failed to unpack the object at pos 2 with error "type must be tuple of ints,but got float"

And, setting the boolean value `True` to `input_size` of `nn.RNN()` gets the indirect error message as shown below:

```python
import torch
from torch import nn

my_tensor = torch.tensor([[8., -3., 5.]])

torch.manual_seed(42)
                      # ↓↓↓↓
rnn = nn.RNN(input_size=True, hidden_size=2)

rnn(input=my_tensor) # Error
```

> RuntimeError: input.size(-1) must be equal to input_size. Expected True, got 3

So, the error messages should be something direct like as shown below:

> TypeError: `input_size` argument must be `int` but got `float`

> TypeError: `input_size` argument must be `int` but got `bool`

### Versions

```python
import torch

torch.__version__ # '2.3.0'
```

vishalgoyal316:
Validates that input_size is an int > 0, providing clear TypeError/ValueError instead of cryptic torch.empty() errors. Applies to nn.RNN, nn.LSTM, nn.GRU.

Fixes #136936

### DIFF
--- a/torch/nn/modules/rnn.py
+++ b/torch/nn/modules/rnn.py
@@ -134,6 +134,12 @@ class RNNBase(Module):
             raise TypeError(
                 f"batch_first should be of type bool, got: {type(batch_first).__name__}"
             )
+        if not isinstance(input_size, int):
+            raise TypeError(
+                f"input_size should be of type int, got: {type(input_size).__name__}"
+            )
+        if input_size <= 0:
+            raise ValueError("input_size must be greater than zero")
         if not isinstance(hidden_size, int):
             raise TypeError(
                 f"hidden_size should be of type int, got: {type(hidden_size).__name__}"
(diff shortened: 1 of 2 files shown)


test/test_nn.py
```diff
@@ -4049,6 +4049,27 @@ def test_projections_errors_on_gru_and_rnn(self):
             with self.assertRaisesRegex(ValueError, error_msg):
                 rnn = getattr(nn, mode)(30, 20, 2, proj_size=10)
 
+    def test_rnn_invalid_input_size_and_hidden_size(self):
```
mikaylagawarecki: Similar ErrorModuleInput comment

mikaylagawarecki (approved):
Thanks

PR #166165, opened 2025-10-24
[Inductor] addmm with bias -> unfuse bias if there is a pointwise/reduction consumer

nikitaved:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* __->__ #166165

Prefer unfused addmm when there is at least a single elemwise/reduction consumer..

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @chenyang78

cc @eqy

PR #166307, opened 2025-10-27
[CUDA][cuBLASLt] addmm -- extend bias fusions to cases with (1 by n) shapes

nikitaved:
* #166300
* #166170
* #166402
* #166165
* __->__ #166307


aten/src/ATen/native/cuda/Blas.cpp
```diff
@@ -329,10 +329,14 @@ static bool isInputCompliesAddmmCudaLt(Tensor& result, const Tensor& self, const
...
   return (beta.toComplexDouble() == 1.0
-    // self.dim() == 1 && result.dim() == 2 && self.sizes()[0] == mat2_sizes[1]
-    // is to use lt interface only when self is bias.
-    && self.dim() == 1 && self.sizes()[0] == mat2_sizes[1] && self.is_contiguous()
     && result.dim() == 2 && result.is_contiguous()
+    // Conditions for bias to be fusable
+    && (
+      self.is_contiguous() &&
+      // NOTE: fine to have 1-len dims to the left from the leading one
+      self.dim() <= result.dim() && self.squeeze().dim() == 1 &&
```
eqy: cool use of `squeeze` here

test/test_linalg.py
```diff
@@ -7259,9 +7259,11 @@ def _test_addmm_impl(self, func, activation, device, dtype):
         m2 = torch.randn(50, 25, device=device).to(dtype)
         self._test_addmm_addmv(func, M, m1, m2, activation=activation)
 
-        # vector-shaped bias and beta=1 result in epilogue fusion in CUDA
+        # vector-shaped bias (or with 1-len dims on the left from the leading dim)
```
eqy, line 7331: is this an "or" or have we changed the case from vector-shaped bias to 1 x n one?

eqy (approved):
is it worth parametrizing the test(s) to try both 1xm and n-shaped bias or is that excessive?

test/test_linalg.py
nikitaved: It is an or. Motivated by the "expected fusions" tests from Inductor -- and we can fuse these broadcast biases safely.

PR #166308, opened 2025-10-27
NVFP4 grouped gemm support via. FBGEMM kernels

slayton58:
* __->__ #166308

Summary:

* Add NVFP4 (1x16 block e4m3, tensor-wise fp32) scaled grouped gemm
* Extend testing to add nvfp4 support

Test Plan:

```
pytest -svv -k grouped test/test_scaled_matmul_cuda.py
```

Reviewers:

Subscribers:

Tasks:

Tags:


test/test_scaled_matmul_cuda.py
```diff
@@ -541,95 +677,62 @@ def test_mxfp8_scaled_grouped_mm_2d_2d(self, G, M, N, K, wrap_v2):
...
+            xh, wh.t(), offs=input_group_end_offsets, out_dtype=torch.bfloat16
         )
 
         # Assert no NaNs
-        assert not y_mxfp8.isnan().any(), "mxfp8 output contains NaN"
+        assert not y_lp.isnan().any(), "mxfp8 output contains NaN"
 
         # Assert outputs are close
-        torch.testing.assert_close(y_mxfp8, y_bf16, atol=8.0e-2, rtol=8.0e-2)
+        torch.testing.assert_close(y_lp, y_bf16, atol=8.0e-2, rtol=8.0e-2)
```
slayton58, line 724: Note @drisspg @danielvegamyhre we can probably be more strict here now that the reference calc has changed

test/test_scaled_matmul_cuda.py
```diff
@@ -643,60 +746,120 @@ def test_mxfp8_scaled_grouped_mm_2d_3d(self, G, M, N, K, wrap_v2):
...
+            **kwargs
+        )
 
         # Compute reference bf16 grouped gemm.
+        # Note: Reference result should be on reconstructed, not original values.
+        #       as-in float(fp4(t)) not t itself.
         y_bf16 = torch._grouped_mm(
-            X,
-            W.transpose(-2, -1),
+            xh,
```
slayton58, line 852: Same again, should be able to make these stricter

aten/src/ATen/native/cuda/GroupedBlas.cpp
```diff
@@ -208,6 +208,35 @@ _f8_f8_bf16_rowwise_grouped_mm(
...
+      const Tensor& scale_a,
+      const Tensor& global_scale_a,
+      const Tensor& scale_b,
+      const Tensor& global_scale_b,
+      const std::optional<Tensor>& offs,
+      const std::optional<Tensor>& bias,
+      Tensor& out) {
+#if !defined(USE_ROCM) && defined(USE_FBGEMM_GENAI)
+  // Checks
+  auto o = fbgemm_gpu::f4f4bf16_grouped_mm(
```
ngimel, line 237: ooc, does fbgemm fn perform all the necessary checks on shapes/values? Also what contraction dimension is assumed here? For fp8 it was different between pytorch and fbgemm

slayton58: Checking in fbgemm is admittedly pretty light - I'll add some more heavy-handed ones.

Contraction dim is implied to be `(-1, -1)`

aten/src/ATen/native/cuda/GroupedBlas.cpp
```diff
@@ -305,18 +326,23 @@ void _check_scales_mxfp8(const Tensor& mat, const Tensor& scale, const int dim,
     // We'll need to support 3d-3d and 3d-2d cases once mxfp8 grouped gemm supports them.
     int64_t G = mat.size(0);
     int64_t K = mat.size(1);
+    if (is_nvfp4) {
+      K *= 2;
```
danielvegamyhre: nit: i assume this *2 is because 2 fp4 elems per byte right? if so, I think adding a short comment or adding some code like this would make it more easily interpretable:

```cpp
if (is_nvfp4) {
    int fp4_elems_per_byte = 2;
    K = K * fp4_elems_per_byte;
}
```

aten/src/ATen/native/cuda/GroupedBlas.cpp
```diff
@@ -568,16 +595,19 @@ _scaled_grouped_mm_cuda_v2(
...
+      // scale shape checks
+      _check_scales_blocked(mat_a, scale_a[0], 0 /* dim */, 0 /* arg_idx */);
+      _check_scales_blocked(mat_b, scale_b[0], 1 /* dim */, 1 /* arg_idx */);
+      return _f4_f4_bf16_grouped_mm_fbgemm(
+          mat_a,
+          mat_b,
+          scale_a[0], /* block-scale A */
+          scale_a[1], /* global-scale A */
+          scale_b[0], /* block-scale B */
+          scale_b[1], /* global-scale A */
```
danielvegamyhre: Comment should say `/* global-scale B */`

danielvegamyhre (commented):
LGTM, just 2 minor comments. Will let Natalia approve. Thanks for adding this, looking forward to using it in torchao

aten/src/ATen/native/cuda/GroupedBlas.cpp
ngimel: :-( so this is different from -1, -2 that all other mm apis have

slayton58: To be very clear to make sure we're definitely talking about the same things:

PyT starts with (for 2d/3d): `mat_a: [M, K], mat_b: [G, K, N]`, and passes into the fbgemm code `mat_a, mat_b.transpose(-2, -1)`, or `[M, K], [G, N, K]` (which is `(-1, -1)` contraction dims)

aten/src/ATen/native/cuda/GroupedBlas.cpp
slayton58: Added a bit more explanation - comments (and compile-time constants) are cheap :)

aten/src/ATen/native/cuda/GroupedBlas.cpp
slayton58: To summarise some offline working-through:
```
# Tensors initially stored as:
mat_a: [M, K], mat_b: [G, N, K]
# Passed to torch.nn.functional.scaled_grouped_mm as: mat_a, mat_b.transpose(-2, -1):
mat_a: [M, K], mat_b: [G, K, N]
# This is passed through to at::_scaled_grouped_mm_v2
# And checked that mat_a.size(-1) == mat_b.size(-2), i.e. contraction_dim=(-1, -2)
# This eventually passes down into fbgemm::f4f4bf16_grouped_mm
# which calls into dispatch as (mat_a, mat_b.transpose(-2, -1), ...), and so the final call gets:
mat_a: [M, K], mat_b: [G, N, K]
# which implies contraction_dim=(-1, -1) for the final kernel.
```

aten/src/ATen/native/cuda/GroupedBlas.cpp
```diff
@@ -245,25 +287,35 @@ void _check_scales_fp8_rowwise(const Tensor& mat, const Tensor& scale, const int
   }
 }
 
-void _check_scales_mxfp8(const Tensor& mat, const Tensor& scale, const int dim, const int arg_idx) {
+void _check_scales_blocked(const Tensor& mat, const Tensor& scale, const int dim, const int arg_idx) {
+  // if {mx,nv}fp4, will need to modify K later
+  bool is_fp4 = (mat.scalar_type() == kFloat4_e2m1fn_x2);
+  int blocksize = 32;
+  // check for nvfp4 vs. mxfp4 to fix blocksize
+  if (is_fp4 && scale.scalar_type() == kFloat8_e4m3fn) {
```
danielvegamyhre: if data is fp4 then per-block scales should always be e4m3 right?  do we have validation for this?

ngimel (approved):

PR #166318, opened 2025-10-27
support batch size=0 for flash attention

### LINKED ISSUE #165944: Support Batchsize 0 for flash attention by early return
# Summary

We should support the flashattention backend when bsz = 0 by early returning 

We should ensure that the lse is working correctly as well as bwd -> I guess no grads should flow and bwd should also ealry return 0

liangel-02:
Fixes #165944

**Summary**

Today, if we attempt to run flash attention with batch_size 0, we get error `Runtime Error: batch size must be positive`. This PR fixes this by returning early with empty tensors in the fwd and bwd.

**Test plan**
`python test/test_transformers.py -k test_scaled_dot_product_attention` - added case for batch_size=0

### DIFF
--- a/aten/src/ATen/native/transformers/cuda/flash_attn/flash_api.cpp
+++ b/aten/src/ATen/native/transformers/cuda/flash_attn/flash_api.cpp
@@ -22,6 +22,7 @@
 #else
 #include <ATen/ops/empty.h>
 #include <ATen/ops/empty_like.h>
+#include <ATen/ops/zeros_like.h>
 #include <ATen/ops/reshape.h>
 #include <ATen/ops/scalar_tensor.h>
 #include <ATen/ops/sum.h>
@@ -41,7 +42,6 @@ C10_DIAGNOSTIC_POP()
 #include <namespace_config.h>
 #include <static_switch.h>
 #include <ATen/native/transformers/cuda/flash_attn/flash_api.h>
-
 
 #include <c10/util/Exception.h>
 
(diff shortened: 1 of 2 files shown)


aten/src/ATen/native/transformers/cuda/flash_attn/flash_api.cpp
```diff
@@ -863,6 +881,20 @@ mha_bwd(const at::Tensor &dout,  // batch_size x seqlen_q x num_heads, x head_si
     const int head_size = sizes[3];
     const int seqlen_k = k.size(1);
     const int num_heads_k = k.size(2);
+
+    if (batch_size == 0) {
+        auto opts = q.options();
+        at::Tensor dq = at::empty_like(q).zero_();
```
drisspg: call zeros_like ?

aten/src/ATen/native/transformers/cuda/flash_attn/flash_api.cpp
```diff
@@ -417,6 +416,26 @@ mha_fwd(const at::Tensor &q,         // batch_size x seqlen_q x num_heads x head
     const int head_size_og = sizes[3];
     const int seqlen_k = k.size(1);
     const int num_heads_k = k.size(2);
+
+    if (batch_size == 0) {
```
drisspg, line 421: @soulitzer what is the semantic for outputs/grads for empty are the 0s or empty?

aten/src/ATen/native/transformers/cuda/flash_attn/flash_api.cpp
```diff
@@ -417,6 +417,26 @@ mha_fwd(const at::Tensor &q,         // batch_size x seqlen_q x num_heads x head
...
+        at::Tensor rng_state = at::empty({2}, at::dtype(c10::kUInt64).device(at::kCUDA));
+        at::Tensor _unused = at::empty({}, at::dtype(c10::kUInt64).device(at::kCUDA));
+        at::Tensor p = at::empty({0}, opts);
+        if (return_softmax) {
+            auto round_multiple = [](int x, int m) { return (x + m - 1) / m * m; };
+            const int seqlen_q_rounded = round_multiple(seqlen_q, 128);
+            const int seqlen_k_rounded = round_multiple(seqlen_k, 128);
+            p = at::empty({0, num_heads, seqlen_q_rounded, seqlen_k_rounded}, opts);
+        }
+        return {out, q_padded, k_padded, v_padded, softmax_lse, rng_state, _unused, p};
```
Skylion007: If we with those, all these values should be moved into the std::tuple.

liangel-02: could you clarify what you mean by this

Skylion007:
```suggestion
        return {std::move(out), std::move(q_padded), std::move(k_padded), std::move(v_padded), std::move(softmax_lse), std::move(rng_state), _unused, std::move(p)};
```

aten/src/ATen/native/transformers/cuda/flash_attn/flash_api.cpp
soulitzer: hmm probably not a huge difference when tensors are zero-numel

liangel-02: ill change to `empty_like` since filling with zeros after isnt necessary

drisspg (approved):

PR #166320, opened 2025-10-27
[DO NOT REVIEW] Inductor lite mode with CUDAGraph support

BoyuanFeng:
This PR tracks the inductor fallback cudagraph backend. The code changes will be landed in smaller PRs.

- [x] Cache `get_free_symbol_uses` for faster compilation time of inductor graph partition #166338
- [x] Land inductor lite mode without cudagrah #167115
- [x] Receive right static input indices when tensor subclass is used (#167127 by @bdhirsh and #166761 by @angelayi)

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @aditew01 @ezyang @EikanWang @wenzhe-nrv @voznesenskym @penguinwu @Guobing-Chen @zhuhaozhe @blzheng @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela @xmfan

### DIFF
--- a/torch/_inductor/__init__.py
+++ b/torch/_inductor/__init__.py
@@ -332,6 +332,21 @@ def list_mode_options(
 
     mode_options: dict[str, dict[str, bool]] = {
         "default": {},
+        # lightweight backend
+        "light": {
+            "fallback_by_default": True,
+            "use_dce": False,
+            "allow_buffer_reuse": False,
+            "reorder_for_peak_memory": False,
+            "reorder_for_compute_comm_overlap": False,
+            "triton.reorder_for_reducing_graph_partitions": False,
+            "use_pre_grad_passes": False,
+            "use_joint_graph_passes": False,
+            "use_post_grad_passes": False,
+            "aten_distributed_optimizations.enable_overlap_scheduling": True,
+            "use_decomposition": False,
+            # "triton.cudagraphs": True,
+        },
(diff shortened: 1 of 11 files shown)


torch/_inductor/__init__.py
```diff
@@ -332,6 +332,21 @@ def list_mode_options(
 
     mode_options: dict[str, dict[str, bool]] = {
         "default": {},
+        # lightweight backend
```
wconstab, line 335: @BoyuanFeng  just a question about how the config works.  in autoparallel, we directly call `compile_fx_inner` from `aot_export_joint_...`'s compile_fn.    Can I directly use this 'mode' for light somewhere?  or do i just need to manually set these same set of options in the inductor.config module?

BoyuanFeng: for a workaround, we can just set the config for autoparallel for now. I will try to land these config soon.

In general, autoparallel uses `aot_export_join_...` taking fw_compiler and bw_compiler, which is different from compile_fx. We need a bit refactor on inductor side to provide these api directly.

PR #166322, opened 2025-10-27
Improve eig tests in preparation for new eig backends

johannesz-codes:
### Summary
Improves validation of `torch.linalg.eig` results by verifying the eigen decomposition identity **A v − v λ = 0**.

### Motivation
Eigenvectors are not unique, and numerical differences between backends (cuSOLVER, MAGMA, CPU)
can cause false test failures. This PR replaces direct elementwise comparisons with a mathematical
identity check, improving robustness across devices.

### Details
- Introduces `fulfills_eigen_decomposition_identity()` in `test_eig_compare_backends()` to validate the eigen equation.
- Uses CPU matmul for high-precision verification.
- Handles zero-sized matrices explicitly.
- Tolerances derived from numerical comparisons between cuSOLVER and NumPy.  
  See discussion: [dev-discuss.pytorch.org link](https://dev-discuss.pytorch.org/t/cusolver-dnxgeev-faster-cuda-eigenvalue-calculations/3248/6)

### Impact
- Improves test stability and correctness across eig backends.
- No change to public API.
- All tests pass; lintrunner reports no issues.
- Enables introduction of new eig backends without false test failures.

### DIFF
--- a/test/test_linalg.py
+++ b/test/test_linalg.py
@@ -2138,32 +2138,6 @@ class TestLinalg(TestCase):
     @skipCUDAIfNoMagma
     @dtypes(*floating_and_complex_types())
     def test_eig_compare_backends(self, device, dtype):
-        def fulfills_eigen_decomposition_identity(a, eig, dtype):
-
-            # check correctness using eigendecomposition identity
-            if dtype in [torch.float32, torch.complex64]:
-                atol = 1e-3  # less accurate results for float32 and complex64 (one to two OOM from NumPy results)
-            else:
-                atol = 1e-13  # Same OOM for NumPy
-
-
-            w, v = eig
-
-            # move all tensors to CPU to avoid issues with GPU matmul
-            a = a.cpu().to(v.dtype)
-            v = v.to("cpu")
-            w = w.to("cpu")
-
-            if a.numel() == 0 and v.numel() == 0 and w.numel() == 0:
-                return True
-            elif a.numel() == 0 or v.numel() == 0 or w.numel() == 0:
-                return False
-
-            diff = (a @ v) - (v * w.unsqueeze(-2))
-            diff = diff.abs()
-            diff = torch.max(diff)
-            print(f"diff: {diff}")
-            return diff <= atol
 
         def run_test(shape, *, symmetric=False):
             from torch.testing._internal.common_utils import random_symmetric_matrix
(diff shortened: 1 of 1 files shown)


test/test_linalg.py
```diff
@@ -2138,6 +2138,33 @@ def run_test(shape, *, symmetric=False):
...
+            # check correctness using eigendecomposition identity
+            if dtype in [torch.float32, torch.complex64]:
+                atol = 1e-3  # less accurate results for float32 and complex64 (one to two OOM from NumPy results)
+            else:
+                atol = 1e-13  # Same OOM for NumPy
+
+
+            w, v = eig
+
+            # move all tensors to CPU to avoid issues with GPU matmul
```
lezcano: What are the issues with GPU matmul?

lezcano: I guess it's multiplying a CPU and a GPU tensor, SG.

johannesz-codes: My concern was tensor core multiplication might mess with accuracy when using fp32

test/test_linalg.py
```diff
@@ -2154,7 +2181,7 @@ def run_test(shape, *, symmetric=False):
             # compare with CPU
             expected = torch.linalg.eig(a.to(complementary_device))
             self.assertEqual(expected[0], actual[0])
```
lezcano, line 2157: Remove this line as well and inline the function in the test. Otherwise LGTM

test/test_linalg.py
lezcano: Nah, it doesn't use tensor cores by default for fp32, it's all SIMD

johannesz-codes: That case I would move the entire evaluation on device

test/test_linalg.py
johannesz-codes: Did you mean to remove the entire complementary block?

lezcano: nvm, sorry, I'm still waking up.
I see, you keep the comparison of the eigenvalues, which are well defined, and you use the definition to check the eigenvectors.

lezcano (commented):
A few things:
1. Could you inline the function you just wrote?
2. Don't move things to CPU unnecessarily. Those computations can be performed on the device.
3. Use `assertEqual(a@v, v - w)` rather than rolling out your own comparison (let alone printing unconditionally)

lezcano (approved):
Great! Thank you for the clean-up!

PR #166324, opened 2025-10-27
Make bucketing aware of collective LIFO semantics

eellison:
* #166528
* #166527
* __->__ #166324
* #166309

In the initial pr for overlapping preserving bucketing, for a graph like:

```
def foo(...):
     ag = all_gather(...)
     hiding_compute = mm(...)
     wait(ag)
```

We would add dependencies from mm -> ag, and wait from wait -> hiding_compute, to prevent bucketing reordering these collectives so that overlap no long occurred. however, there is an additional way for bucketing to prevent overlap.

If we were to reorder another collective so the graph looked like: 

```
def foo(...):
     ag = all_gather(...)
     ar = all_reduce(...)
     wait(ar)
     hiding_compute = mm(...)
     wait(ag)
```

Overlap would not occur, because the wait for the all reduce would also force realization of every collective enqueued on the same stream prior to the all reduce. NCCL uses a single stream per process group.

To model, we set a set a strict ordering of all collective starts, waits, and hiding compute initially when bucketing. Then, when trying to add a collective to a bucket, we will see if we interfere with overlap for all of the following possible bucketings:

[move collective start to bucket start, move bucket start to collective start] x [move collective wait to bucket wait x move bucket wait to collective wait].

For any of these positions, we check if overlap would have been interfered with because of stream queue semantics. Then, if not, we remove the moving start and wait from the constrained ordering of collectives, and see if it's topologically valid to merge the nodes.

### DIFF
--- a/test/distributed/test_overlap_bucketing_unit.py
+++ b/test/distributed/test_overlap_bucketing_unit.py
@@ -106,29 +106,6 @@ def compute_ancestors(graph):
     return node_ancestors
 
 
-# @requires_accelerator_dist_backend(["nccl", "xccl"])
-# @instantiate_parametrized_tests
-# class TestCollectivesMultiProc(DynamoDistributedMultiProcTestCase):
-#     """
-#     Run correctness checks in multi-proc runner, mark with minimum # GPUs to run under
-#     """
-
-#     device = acc.type if (acc := torch.accelerator.current_accelerator()) else "cpu"
-
-#     def get_world_trs(self):
-#         return {
-#             "tag": "",
-#             "ranks": list(range(self.world_size)),
-#             "group_size": self.world_size,
-#         }
-
-#     @property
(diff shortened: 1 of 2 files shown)


test/distributed/test_overlap_bucketing_unit.py
```diff
@@ -0,0 +1,436 @@
...
+#             "tag": "",
+#             "ranks": list(range(self.world_size)),
+#             "group_size": self.world_size,
+#         }
+
+#     @property
+#     def world_size(self) -> int:
+#         # hack: no matter whether we have 2 or 3 or 4 gpus, just run on 2
+#         # works around issue with skipif<2 and workers with unpredictable #s gpu
+#         return 2
```
IvanKobzarev: cleanup?

IvanKobzarev (approved):

PR #166328, opened 2025-10-27
[dynamo] Dont guard on numpy Cython functions

anijain2305:
* #166321
* #166329
* __->__ #166328


Lucaskabela (approved):
Can we add unit test checking this (only because I worry in the future someone may try and fold these conditionals together with an and, which may be wrong here so we should be sure to have some signal covering)

Other than that lgtm so approving

PR #166331, opened 2025-10-27
Add conv ops to operator microbenchmark

jainapurva:
Adding `conv` (conv1d, conv2d, conv3d) to the list of operator microbenchmarks run in the CI script (`.ci/pytorch/test.sh`), ensuring convolution operators are now benchmarked alongside existing ones.

### DIFF
--- a/benchmarks/operator_benchmark/pt/configs.py
+++ b/benchmarks/operator_benchmark/pt/configs.py
@@ -141,8 +141,6 @@ conv_3d_configs_long = op_bench.cross_product_configs(
     D=[128],
     H=[128],
     W=[128],
-    G=[1],
-    pad=[0],
     device=["cpu", "cuda"],
     tags=["long"],
 )
--- a/benchmarks/operator_benchmark/pt/configs.py
+++ b/benchmarks/operator_benchmark/pt/configs.py
@@ -11,6 +11,11 @@ def remove_cuda(config_list):
     return [config for config in config_list if cuda_config not in config]
 
 
+def remove_cpu(config_list):
+    cpu_config = {"device": "cpu"}
+    return [config for config in config_list if cpu_config not in config]
+
+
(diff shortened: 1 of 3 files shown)


benchmarks/operator_benchmark/pt/configs.py
```diff
@@ -11,6 +11,11 @@ def remove_cuda(config_list):
     return [config for config in config_list if cuda_config not in config]
 
 
+def remove_cpu(config_list):
```
huydhn, line 14: Just curious: Why do you need to remove CPU device on those conv ops?  Because they are all on backward path?

huydhn (approved):

benchmarks/operator_benchmark/pt/configs.py
```diff
@@ -127,6 +132,18 @@ def remove_cuda(config_list):
...
+    OC=[32, 64],
+    kernel=[3, 5],
+    stride=[1, 2],
+    N=[1],
+    D=[128],
+    H=[128],
+    W=[128],
+    device=["cpu", "cuda"],
+    tags=["long"],
+)
```
jbschlosser, line 146: out of curiosity, does this cover the shapes relevant to the recent regression?

jbschlosser (approved):

benchmarks/operator_benchmark/pt/configs.py
jainapurva: Yes, as backward tests are not needed for cpu benchmarking, I'm removing cpu tests from the final list

benchmarks/operator_benchmark/pt/configs.py
jainapurva: Yes, the shapes of regression are covered, but the enablement of AMP is not covered

PR #166335, opened 2025-10-27
 Raise error for 1D (size > 1) -> 0D parameter loads

### LINKED ISSUE #165873: No error raised despite shape mismatch in load_state_dict
### 🐛 Describe the bug

When I initialize a Parameter of a Module to be a scalar, and then load in a state_dict containing a 1D tensor, there is no error; instead, the Parameter takes on the first value of the 1D tensor without complaint. In this scenario, I would expect there to be a RuntimeError. Indeed, if you you try other combinations (initializing the parameter to be a size [1] 1D tensor, or loading in tensors of 2 or more dimensions), a RuntimeError is raised.

See the following code for an example; the assertion passes:

```python
import torch
import torch.nn as nn


class SimpleModule(nn.Module):
    def __init__(self, threshold_value):
        super().__init__()
        self.threshold = nn.Parameter(torch.tensor(threshold_value))
    
    def forward(self, x):
        return x

large_tensor = torch.randn(32000)
state_dict = {"threshold": large_tensor}

module = SimpleModule(0.0)
module.load_state_dict(state_dict)
assert module.threshold.item() == state_dict['threshold'][0].item() 
```

I tested this on version 2.8 on macOS, but this error occurs on CUDA too. Also see [this short thread](https://discuss.pytorch.org/t/no-error-raised-despite-shape-mismatch-in-load-state-dict/223641/2) where another forumgoer, KFrank, confirms that this occurs in 2.9 and did some more testing of when this bug occurs.

### Versions

PyTorch version: 2.8.0
Is debug build: False

OS: macOS 15.6.1 (arm64)
CMake version: Could not collect
Libc version: N/A

Is CUDA available: False
CUDA runtime version: No CUDA
cuDNN version: No CUDA
Is XPU available: False
HIP runtime version: N/A
MIOpen runtime version: N/A
Is XNNPACK available: True

CPU:
Apple M3

[pip3] Could not collect
[conda] Could not collect

cc @malfet @mruberry @mikaylagawarecki

dsashidh:
Fixes #165873 

# Title
Fix load_state_dict: raise error for 1D (size > 1) -> 0D parameter loads

## Summary
This PR fixes a bug where loading a 1D tensor (size > 1) into a scalar (0D) parameter would silently take the first element instead of raising an error. The fix preserves backward compatibility for 1D tensors of size 1 while catching genuine shape mismatches.

## Motivation
Previously, loading a 1D tensor like torch.randn(32000) into a 0D scalar parameter would silently slice the first element, leading to silent data loss and potential bugs. This change ensures users get a clear error when there's a genuine shape mismatch.

## Behavior change

Before: 
1D tensor (any length) -> 0D scalar -> silently coerced using input_param[0]

After: 
- 1D tensor (size == 1) -> 0D scalar -> allowed (backward compatibility)
- 1D tensor (size > 1) -> 0D scalar -> raises RuntimeError with size mismatch message

In torch/nn/modules/module.py, _load_from_state_dict, added input_param.shape[0] == 1 check to the backward compatibility condition to only allow single-element 1D tensors.

## Tests
Added test_scalar_param_1d_tensor_raises to verify that loading 1D tensors of size > 1 raises an error, while size 1 loads successfully.

### DIFF
--- a/torch/nn/modules/module.py
+++ b/torch/nn/modules/module.py
@@ -2438,12 +2438,9 @@ class Module:
                     not is_param_lazy
                     and len(param.shape) == 0
                     and len(input_param.shape) == 1
+                    and input_param.shape[0] == 1
                 ):
-                    error_msgs.append(
-                        f"size mismatch for {key}: copying a param with shape {input_param.shape} from checkpoint, "
-                        f"the shape in current model is {param.shape}."
-                    )
-                    continue
+                    input_param = input_param[0]
 
                 if not is_param_lazy and input_param.shape != param.shape:
                     # local shape should match the one in checkpoint
--- a/torch/nn/modules/module.py
+++ b/torch/nn/modules/module.py
@@ -2439,7 +2439,11 @@ class Module:
                     and len(param.shape) == 0
                     and len(input_param.shape) == 1
                 ):
-                    input_param = input_param[0]
+                    error_msgs.append(
+                        f"size mismatch for {key}: copying a param with shape {input_param.shape} from checkpoint, "
+                        f"the shape in current model is {param.shape}."
+                    )
+                    continue
 
                 if not is_param_lazy and input_param.shape != param.shape:
                     # local shape should match the one in checkpoint
--- a/test/nn/test_load_state_dict.py
+++ b/test/nn/test_load_state_dict.py
@@ -66,20 +66,22 @@ class TestLoadStateDict(NNTestCase):
         class SimpleModule(nn.Module):
             def __init__(self):
                 super().__init__()
-                self.threshold = nn.Parameter(torch.tensor(0.0))  # 0D
+                self.threshold = nn.Parameter(torch.tensor(0.0))
 
             def forward(self, x):
                 return x
 
         m = SimpleModule()
 
+        # Test that [3] -> scalar raises error
         sd = {"threshold": torch.randn(3)}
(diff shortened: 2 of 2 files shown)


torch/nn/modules/module.py
morrison-turnansky, line 2437: @dsashidh  It looks like this case was here for backwards compatibility, but from a long time ago. 

If there is a decision to no longer support this backward compatibility, the cleaner fix would be to remove this if statement and let if fall back to if not is_param_lazy and input_param.shape != param.shape, where it will be caught in this case and currently has the same body.

mikaylagawarecki (commented):
I think my comment below might be an acceptable fix

torch/nn/modules/module.py
```diff
@@ -2436,7 +2436,11 @@ def _load_from_state_dict(
                     and len(param.shape) == 0
                     and len(input_param.shape) == 1
```
mikaylagawarecki, line 2440: Perhaps add `and input_param.shape[0] == 1`, so the unexpected case will not fall into this if statement

torch/nn/modules/module.py
dsashidh: Thanks for the feedback, I commented below.

torch/nn/modules/module.py
dsashidh: Thanks for the feedback, I commented below.

mikaylagawarecki (approved):

PR #166338, opened 2025-10-27
[GraphPartition] cache get_free_symbol_uses

BoyuanFeng:
Graph partition relies on `get_free_symbol_uses()` to collect symbol inputs.
https://github.com/pytorch/pytorch/blob/ee7434be822cf6e75b4566d8159f550ee233d8ae/torch/_inductor/scheduler.py#L4869-L4885

I empirically observed that `get_free_symbol_uses()` becomes slower for larger graphs. Specifically, I tried to aten fallback for torchtitan which results in 10k+ aten nodes. When processing the 600-th node, it takes seconds to `get_free_symbol_uses()` for 1 node.

Why? Because `get_free_symbol_uses()` may recursively call another `get_free_symbol_uses()`, which could recursively run many times.
https://github.com/pytorch/pytorch/blob/ee7434be822cf6e75b4566d8159f550ee233d8ae/torch/_inductor/ir.py#L4541-L4543

This PR fixes the issue by caching the results of `get_free_symbol_uses()`. I validated on torchtitan that the issue is fixed.


torch/_inductor/utils.py
```diff
@@ -709,6 +709,21 @@ def cache_property_on_self(fn: Callable[P, RV]) -> CachedMethod[P, RV]:
...
+def cache_on_self_and_args(
+    fn: Callable[Concatenate[Any, P], RV],
+) -> Callable[Concatenate[Any, P], RV]:
+    cache = {}
+
+    @functools.wraps(fn)
+    def wrapper(self: Any, *args: P.args, **kwargs: P.kwargs) -> RV:
+        key = (id(self), args, tuple(sorted(kwargs.items())))
+        if key not in cache:
+            cache[key] = fn(self, *args, **kwargs)
```
desertfire: Looking at how `cache_on_self` was implemented, I think we should do something similar here to further improve the performance.

eellison (commented):
Right, flexible layout might change stride, although it's unlikely it would induce a new symbol use. Would it be safer to only cache if the layout is fixed ?

torch/_inductor/ir.py
```diff
@@ -3973,6 +4010,33 @@ def same_ordered(
...
+        self.assert_free_symbol_uses_unchanged("stride", value)
+        self._stride = value
+
+    @property
+    def offset(self) -> Expr:
+        return self._offset
+
+    @offset.setter
+    def offset(self, value: Expr) -> None:
+        self.assert_free_symbol_uses_unchanged("offset", value)
```
BoyuanFeng, line 4037: error if free symbols are added or deleted after initialization.

eellison (commented):
sorry, one last question - at the point we call it all these nodes should have fixed layout. should we just only cache in the fixed layout case ? i think that will be a bit simpler.

desertfire:
> sorry, one last question - at the point we call it all these nodes should have fixed layout. should we just only cache in the fixed layout case ? i think that will be a bit simpler.

That is simpler but probably will have performance implication. @BoyuanFeng , I wonder how much performance difference it will be.

eellison (approved):
Looks good - cc @bobrenjc93 for other thoughts

PR #166339, opened 2025-10-27
[Inductor] Lower fallback nodes annotated with "should_fallback"

chenmillie:
Summary:
This PR introduces an inductor-level fallback mechanism that gives users control over which operations or subgraphs Inductor should lower and which should fall back to preexisting kernels. This has similar motivation as #164776 in providing flexibility to selectively disable Inductor lowering for specific nodes.

The implementation simply adds a check for the `"should_fallback"` metadata annotation on FX graph nodes. If this is set to `True`, the lowerer falls back before attempting the normal lowering path. Note that since these are user-directed fallbacks dependent upon specific, customized conditions, use `add_to_fallback_set=False` to avoid permanent overwrites of inductor's lowering/fallback rules. 

Simple example marking nodes for fallback based on custom predicates:

```
def should_fallback_predicate(node: torch.fx.Node, pred: Callable[torch.fx.Node, bool]):
    # Apply predicate and mark for fallback if needed
    if self.predicate(node):
         node.meta["should_fallback"] = True
```

Test Plan: added a CI test

Differential Revision: D85347587

### DIFF
--- a/test/inductor/test_selective_lowering.py
+++ b/test/inductor/test_selective_lowering.py
@@ -1,0 +1,91 @@
+# Owner(s): ["module: inductor"]
+"""
+Test selective lowering control via node metadata annotations.
+"""
+
+from collections.abc import Callable
+
+import torch
+from torch._inductor.test_case import TestCase as InductorTestCase
+from torch.testing._internal.common_utils import instantiate_parametrized_tests
+from torch.testing._internal.inductor_utils import GPU_TYPE, HAS_GPU
+
+
+@instantiate_parametrized_tests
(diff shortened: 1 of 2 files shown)


blaine-rister (approved):
LGTM! This is a very important feature for MTIA, where we want to customize the fallback logic based on node arguments and not only the node target. Past experience has shown FX passes to be a very convenient way to control fallbacks. Though I'll defer to @BoyuanFeng @eellison or @ezyang  to comment on the Inductor implementation.

eellison (approved):

PR #166341, opened 2025-10-27
[PT2 Compiler] Add annotation for dynamo disabled callables

malaybag:
Summary: To make torch.export compatible with PT2 compile (which is done on top of exported model) we need to store torch._dynamo.disable attributes in exported model and later restore this after unflattening of exported model. This diff will add annotations to all nodes with torch._dynamo.disable, which will be preserved during exporting.

Test Plan:
```
buck test mode/opt caffe2/test:test_export -- 'test_dynamo_disable_annotations'
```
https://www.internalfb.com/intern/testinfra/testrun/6473924770741560

Differential Revision: D85302730

### DIFF
--- a/torch/_dynamo/eval_frame.py
+++ b/torch/_dynamo/eval_frame.py
@@ -78,7 +78,7 @@ from torch.export.dynamic_shapes import (
     _RelaxedConstraint,
     Constraint,
 )
-from torch.fx import GraphModule
+from torch.fx import GraphModule, traceback as fx_traceback
 from torch.fx.experimental._dynamism import (
     clone_and_convert_to_meta,
     track_dynamism_across_examples,
@@ -1134,6 +1134,17 @@ class DisableContext(_TorchDynamoContext):
             try:
                 _maybe_set_eval_frame(_callback_from_stance(self.callback))
                 try:
+                    if torch.compiler.is_exporting():
+                        with fx_traceback.annotate(
+                            {
+                                "_torchdynamo_disable": True,
+                                "_torchdynamo_disable_recursive": True,
+                                "_torchdynamo_disable_method": getattr(
+                                    fn, "__name__", type(fn).__name__
+                                ),
+                            }
+                        ):
+                            return fn(*args, **kwargs)
                     return fn(*args, **kwargs)
                 finally:
                     set_eval_frame(None)
--- a/torch/_dynamo/external_utils.py
+++ b/torch/_dynamo/external_utils.py
@@ -196,6 +196,10 @@ def get_nonrecursive_disable_wrapper(fn: Callable[_P, _R]) -> Callable[_P, _R]:
     # this function is in external_utils so that convert_frame doesn't skip it.
     @functools.wraps(fn)
     def nonrecursive_disable_wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _R:
+        if torch.compiler.is_exporting():
+            raise RuntimeError(
+                "Non-recursive torch.compiler.disable is not supported with torch.export."
+            )
         return fn(*args, **kwargs)
 
     return nonrecursive_disable_wrapper
--- a/test/dynamo/test_decorators.py
+++ b/test/dynamo/test_decorators.py
@@ -2155,6 +2155,43 @@ Detected recompile when torch.compile stance is 'fail_on_recompile'. filename: '
             torch.compile(model)
             torch.compile(other_model)
 
+    def test_dynamo_disable_annotations(self):
+        class SimpleModel(torch.nn.Module):
+            def __init__(self) -> None:
+                super().__init__()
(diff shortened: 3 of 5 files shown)


angelayi (commented):
@williamwen42 do you mind taking a look at the dynamo changes

torch/_dynamo/external_utils.py
```diff
@@ -196,7 +197,18 @@ def get_nonrecursive_disable_wrapper(fn: Callable[_P, _R]) -> Callable[_P, _R]:
     # this function is in external_utils so that convert_frame doesn't skip it.
     @functools.wraps(fn)
     def nonrecursive_disable_wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _R:
-        return fn(*args, **kwargs)
+        with fx_traceback.annotate(
```
williamwen42: the entire with block here will be traced by dynamo

torch/_dynamo/eval_frame.py
```diff
@@ -1133,7 +1133,18 @@ def _fn(*args: Any, **kwargs: Any) -> Any:
             try:
                 _maybe_set_eval_frame(_callback_from_stance(self.callback))
                 try:
-                    return fn(*args, **kwargs)
+                    with fx_traceback.annotate(
```
williamwen42: the with block here might not be traced because eval_frame captures new frames. If you want to trace the with block, you will have to create another frame (by creating another local function).

williamwen42 (commented):
made comments on dynamo parts

torch/_dynamo/external_utils.py
malaybag: @williamwen42 - what will be the side effect of the tracing this block?

torch/_dynamo/eval_frame.py
malaybag: Sorry, n00b questions, do we need to trace the with block? The goal here is just to update custom meta, which the is done by fx_traceback.annotate.

torch/_dynamo/external_utils.py
```diff
@@ -196,7 +197,18 @@ def get_nonrecursive_disable_wrapper(fn: Callable[_P, _R]) -> Callable[_P, _R]:
...
+            {
```
angelayi: @malaybag since we decided to not support recursive=False, let's just throw a RuntimeError here if torch.compiler.is_exporting()?

torch/_dynamo/eval_frame.py
williamwen42: Never mind, my previous comment is not relevant because dynamo will be disabled as we run this code.

However, I do have some performance concerns since we're on the `torch.compile` hot path. Can you try to do this instead?
```python
if torch.compiler.is_exporting():
  with fx_traceback.annotate(...):
    return fn(...)
return fn(...)
```

This should prevent `fx_traceback` overhead from affecting `torch.compile` overhead (non-export case)

torch/_dynamo/external_utils.py
williamwen42: I'm wrong here since Dynamo shouldn't trace through `fx_traceback.annotate`.

Offline discussion w/ @angelayi - it might be better to error here if export really shouldn't support disable(recursive=False).

Also see here (https://github.com/pytorch/pytorch/blob/e69aaaf45a8018004aa91d58bef77199acbb888e/torch/_dynamo/convert_frame.py#L673) for how we determine if a frame has nonrecursive disable applied to it

malaybag: @williamwen42 added runtime error for recursive=False case

torch/_dynamo/eval_frame.py
malaybag: @williamwen42 updated the logic. Please take a look.

torch/_dynamo/external_utils.py
malaybag: @angelayi updated to add runtime error.

torch/_dynamo/external_utils.py
```diff
@@ -196,6 +196,8 @@ def get_nonrecursive_disable_wrapper(fn: Callable[_P, _R]) -> Callable[_P, _R]:
     # this function is in external_utils so that convert_frame doesn't skip it.
     @functools.wraps(fn)
     def nonrecursive_disable_wrapper(*args: _P.args, **kwargs: _P.kwargs) -> _R:
+        if torch.compiler.is_exporting():
+            raise RuntimeError("Recursive dynamo disable is not supported with torch.export.")
```
williamwen42:
```suggestion
            raise RuntimeError("Non-recursive torch.compiler.disable is not supported with torch.export.")
```

williamwen42 (commented):
ok lgtm on dynamo side

williamwen42 (approved):

angelayi (approved):

PR #166342, opened 2025-10-27
Rework PyObject preservation

colesbury:
Make the PyObject preservation scheme thread-safe with free threaded (nogil) Python. The general idea is:

* Python Tensor and Storage objects always hold a strong reference to their underlying c10 object
* c10 objects hold a strong reference to their Python objects if there's at least one other reference to the c10 object

This is implemented in `intrusive_ptr`:

* The top most bit (`kHasPyObject`) from the weakref count is now used to indicate if the `intrusive_ptr_target` has an associated PyObject. So `kHasPyObject` is one bit, the weakref count is now 31 bits and the strong refcount remains 32 bits.
* When the reference count increases from one to two and `kHasPyObject` is set, we incref the associated Python object to ensure that it's kept alive. 
* When the reference count decreases from two to one (i.e., there are no C++ reference to the `intrusive_ptr_target` other than from the Python object), we decre the associated Python object to break the cycle.

Other benefits:

* We can delete a lot of the copypasta from Python internal `subtype_dealloc`
* This fixes the weakref and GC bugs we had in the previous scheme. Python weakrefs on Tensors and Storages should just work as expected now.

Risks:

* Extra branch for reference count operations on `intrusive_ptr<TensorImpl>`, `intrusive_ptr<StorageImpl>`, and the generic `intrusive_ptr<intrusive_ptr_target>` even when we're not using Python.
* It's a big change


c10/util/intrusive_ptr.h
```diff
@@ -241,6 +247,12 @@ class C10_API intrusive_ptr_target {
     return *this;
   }
 
+  bool set_has_pyobject() noexcept {
```
swolchok: not sure what the return value of this function is indicating semantically

c10/util/intrusive_ptr.h
```diff
@@ -33,6 +33,8 @@ constexpr uint64_t kImpracticallyHugeWeakReferenceCount =
 constexpr uint64_t kReferenceCountOne = 1;
 constexpr uint64_t kWeakReferenceCountOne = (kReferenceCountOne << 32);
 constexpr uint64_t kUniqueRef = (kReferenceCountOne | kWeakReferenceCountOne);
+// Indicates whether the object has a PyObject wrapper.
+constexpr uint64_t kHasPyObject = (uint64_t(1) << 63);
```
swolchok, line 41: can/should we define a compile-time check to rule out some types T such that `intrusive_ptr<T>` cannot possibly have a pyobject and 1) save the extra instructions for those types and 2) check this compile-time assumption (ideally at compile time, but runtime is OK) by failing if the new pyobject stuff is called? I imagine that those types exist (do we ever back e.g. c10::List with a PyObject?) and would at least appreciate the code size savings

swolchok: oops, I realize now that I should have suggested that instead we allowlist those types that require PyObject support rather than ruling out specific types.

colesbury: I was thinking about something like:

```c++
template <class T>
struct TargetTraits {
  static constexpr bool has_pyobject = false;
  static inline void incref_pyobject(T* self) = delete;
  static inline void decref_pyobject(T* self) = delete;
};

template <>
struct TargetTraits<c10::TensorImpl> {
  static constexpr bool has_pyobject = true;

  static inline void incref_pyobject(c10::TensorImpl* self) noexcept {
    ...
  }
  static inline void decref_pyobject(c10::TensorImpl* self) noexcept {
    ...
  }
};
```

But in a few places we use the generic `intrusive_ptr<intrusive_ptr_target>` and I think that would end up (incorrectly) ignoring the PyObject wrappers when the actual target is a Tensor/Storage.

swolchok: > But in a few places we use the generic intrusive_ptr<intrusive_ptr_target> and I think that would end up (incorrectly) ignoring the PyObject wrappers when the actual target is a Tensor/Storage.

Couldn't we solve that by opting intrusive_ptr_target into the has_pyobject behavior? this shouldn't affect subclasses; instantiating TargetTraits<T> shouldn't care about inheritance

colesbury: Yeah, I think that should work

c10/core/StorageImpl.h
```diff
@@ -105,6 +105,30 @@ struct C10_API StorageImpl : public c10::intrusive_ptr_target {
     data_ptr_.clear();
   }
 
+  void incref_pyobject() const override final {
+    // Because intrusive_ptr incref uses relaxed memory order, we need to
+    // do an acquire fence to ensure that the kHasPyObject bit was
+    // observed before the load of the PyObject* below.
+    // NB: This is a no-op on x86/x86-64
+    std::atomic_thread_fence(std::memory_order_acquire);
```
swolchok: how bad is this on aarch64? is it reasonable to suppose that there might be workloads that trigger this frequently? @mcfi any thoughts?

colesbury: Benchmark: https://gist.github.com/colesbury/27ed5b8bbe01dc2eabbfe5285d0aa1d7

Apple M4: ~0.38 ns (so maybe ~1-2 cycles)
AWS t4g: ~2.85 ns (so maybe 8 cycles)

In my experience, all the non Apple aarch64 chips have pretty similar performance for memory barriers (with Apple's chips performing much better). 

We could also get away with a reload `combined_refcount_.load(std::memory_order_acquire)`. That compiles to `ldar`/`ldapr` instead of ~~`dmb ish`~~ `dmb ishld` and is a bit faster. It's just a bit awkward in the code, and GCC won't optimize the atomic load away on x86-64 like it will the fence:

Apple M4: ~0.22 ns (so maybe ~1 cycle)
AWS t4g: ~0.82 ns (so maybe ~2-3 cycles)

colesbury: > is it reasonable to suppose that there might be workloads that trigger this frequently

I think the main way it gets triggered if you call from Python to C++ and stash that Tensor somewhere. For example:

1) When autograd saves a c10::Tensor for backwards
2) If you have a kernel that looks like:

```c++
void my_kernel(const Tensor & input) {
   Tensor x = input; // may incref the PyObject
   ...
}
```

If you already have an owning reference in C++ (i.e., a `Tensor` not a `const Tensor&`) then further increfs won't trigger it.

swolchok: > Benchmark: https://gist.github.com/colesbury/27ed5b8bbe01dc2eabbfe5285d0aa1d7
> 
> Apple M4: ~0.38 ns (so maybe ~1-2 cycles)
> AWS t4g: ~2.85 ns (so maybe 8 cycles)
> 
> In my experience, all the non Apple aarch64 chips have pretty similar performance for memory barriers (with Apple's chips performing much better).

we are rapidly leaving the set of things I have actual experience with, but isn't the problem with memory barriers that they affect the entire system, whereas load-acquire and store-release are more narrow? seems like that effect wouldn't necessarily show up in a benchmark. would defer to @mcfi on this topic though

mcfi: I think a reload of the counter with acquire seems to be the best choice for arm64 since we already loaded the counter so the reload (ldar if targeting Armv8.0 and ldapr if targeting Armv8.2 + LRCPC) should hit L1 cache and be the fastest compared to a dmb fence on Arm CPUs.

For x86, the acquire fence is a nop, so it might be good to have the following defined in intrusive_ptr.h and used everywhere.
#ifdef x86
#define acquire_fence std::atomic_thread_fence(std::memory_order_acquire)
#elif arm64
#define acquire_fence combined_refcount_.load(std::memory_order_acquire)
#endif

colesbury: I've updated the benchmark with:

fetch_add

```c++
x.fetch_add(1, std::memory_order_relaxed);
```

fetch_add + reload

```c++
x.fetch_add(1, std::memory_order_relaxed);
x.load(std::memory_order_acquire);
```

Apple M4: 3.5 ns/iter
AWS t4g: 8.2 ns/iter

fetch_add + acquire fence (`dmb ishld`)

```c++
x.fetch_add(1, std::memory_order_relaxed);
std::atomic_thread_fence(std::memory_order_acquire);
``` 

The fence is no slower, sometimes faster:

|          | fetch_add | fetch_add + reload | fetch_add + fence |
|----------|-----------|--------------------|-------------------|
| Apple M4 | 1.6 ns    | 3.6 ns             | 1.6 ns            |
| AWS t4g  | 5.2 ns    | 8.2 ns             | 8.0 ns            |

(Also I remembered to compile with `-march=armv8.3-a` for the aws graviton this time, although it doesn't make much of a difference)

colesbury: I think all these calls are going to be cheaper than the `pybind11::gil_scoped_acquire gil`  called indirectly a few lines below

mcfi: ok, thanks for the new microbenchmark data. I'm fine either way.

c10/core/impl/PyObjectSlot.h
```diff
@@ -8,117 +8,70 @@
...
-  // with other data structures are still correctly synchronized, so that
-  // we fall in the "Single-Location Data Structures" case as described in
-  // http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2020/p2055r0.pdf
-  // However, on x86, it doesn't matter if I use acquire or relaxed on the load
-  // as I get the same assembly in both cases.  So I just use the more
-  // conservative acquire (which will impede compiler optimizations but I don't
-  // care)
+  // This is now always the global interpreter if the PyObject is set.
+  // Maybe we can remove this field some day...
   std::atomic<PyInterpreter*> pyobj_interpreter_;
```
ezyang, line 67: https://github.com/pytorch/pytorch/pull/162659 but it got yanked because... idk, unclear, potentially someone spookily depending on TensorImpl ABI details....

test/test_torch.py
```diff
@@ -10407,6 +10402,31 @@ def callback(w):
...
+            barrier = threading.Barrier(NUM_THREADS)
+            with ThreadPoolExecutor(max_workers=NUM_THREADS) as executor:
+                futures = [
+                    executor.submit(access_untyped_storage, tensor, barrier)
+                    for _ in range(NUM_THREADS)
+                ]
+
+                # Check that all the storages returned were the same
+                for future in futures:
+                    self.assertEqual(future.result(), id(tensor.untyped_storage()))
```
albanD: nit: given that we don't keep the storage object alive, could this be passing by mistake because it re-uses the same memory space for all the storage objects one after the other?
Should we use weakrefs here to check for identity?

torch/csrc/autograd/utils/grad_layout_contract.h
```diff
@@ -65,7 +65,9 @@ inline at::Tensor clone_obey_contract(
                          .new_empty_strided_symint(
                              variable.sym_sizes(),
                              variable.sym_strides(),
-                             variable.options().memory_format(std::nullopt))
+                             variable.options()
+                                 .memory_format(std::nullopt)
+                                 .dtype(new_grad.dtype()))
```
albanD, line 70: the new_* function will use the input Tensor's properties when not specified here, so this shouldn't change anything right?

torch/csrc/autograd/functions/accumulate_grad.h
```diff
@@ -180,7 +180,9 @@ struct TORCH_API AccumulateGrad : public Node {
       if (!GradMode::is_enabled() && !new_grad.is_sparse() &&
           !new_grad.is_sparse_csr() &&
           !(variable.is_sparse_csr() && new_grad.layout() == at::kStrided) &&
-          at::caching::adjusted_use_count(new_grad) <= num_expected_refs &&
+          impl::is_tensor_stealable(
+              new_grad,
+              num_expected_refs + at::caching::is_cached_tensor(new_grad)) &&
```
albanD, line 185: Why the addition of is_cached_tensor here?

torch/csrc/autograd/python_variable.cpp
```diff
@@ -257,152 +257,65 @@ void activateGPUTrace() {
...
-}
-
-// returns true if successfully rezzed; if so, cancel the
-// rest of deallocation
-static bool THPVariable_tryResurrect(THPVariable* self) {
-  const auto& tensor = THPVariable_Unpack(self);
 
-  if (!isResurrectable(self)) {
-    return false;
+  if (check_has_torch_dispatch(obj)) { // Boo!
```
albanD: Boo?

torch/csrc/autograd/variable.h
```diff
@@ -197,6 +197,22 @@ TORCH_API std::unique_ptr<PostAccumulateGradHook>& post_acc_grad_hooks(
 TORCH_API void create_cpp_hook(
     const at::TensorBase& /*self*/,
     bool is_retains_grad_hooks = false);
+
+inline bool is_tensor_stealable(
+    const at::Tensor& new_grad,
+    size_t num_expected_refs = 1) {
+  size_t use_count = new_grad.use_count();
+  if (use_count <= num_expected_refs) {
```
albanD, line 205: Can it ever be < ? Or should that be considered a bug?

torch/csrc/utils/python_compat.h
```diff
@@ -33,12 +33,67 @@ static inline int PyCode_GetNFreevars(PyCodeObject* code) {
 #endif
 }
 
-// Provided by CPython but getting the header for them is very hard
-#if IS_PYTHON_3_11_PLUS
-// NOLINTNEXTLINE(readability-redundant-declaration)
-PyAPI_FUNC(void) _PyWeakref_ClearRef(PyWeakReference* self);
+#if !IS_PYTHON_3_14_PLUS
+static inline int PyUnstable_TryIncRef(PyObject* obj) {
```
albanD: these landed in capi compat right?

torch/csrc/PyInterpreter.cpp
```diff
@@ -292,6 +254,33 @@ void ConcretePyInterpreterVTable::incref(PyObject* pyobj) const {
...
+bool ConcretePyInterpreterVTable::try_incref(
+    const c10::impl::PyObjectSlot& pyobj_slot) const {
+  if (!Py_IsInitialized())
+    return false;
+  pybind11::gil_scoped_acquire gil;
+  PyObject* pyobj = pyobj_slot.load_pyobj();
+  if (!pyobj) {
+    return false;
+  }
+#if IS_PYTHON_3_14_PLUS
```
albanD: Not needed given the api compat addition

albanD (commented):
Very nice cleanup!

test/test_torch.py
colesbury: Yeah, I can do that

torch/csrc/autograd/functions/accumulate_grad.h
colesbury: It's the same logic as before, just on the other side of the `<=` comparison.

* `at::caching::adjusted_use_count(new_grad)` previous subtracted 1 from the left hand side. Instead we add one here to the right hand side, because it works better with the way `is_tensor_stealable` is written
* Ideally, we would have accounted for `at::caching::is_cached_tensor()` inside `is_tensor_stealable`, but that breaks some Meta-internal stuff with linking :(

torch/csrc/autograd/variable.h
colesbury: I'm not sure. Seems like it might be a bug to me, but that's the condition we had before.

torch/csrc/utils/python_compat.h
colesbury: Yes, removed now

torch/csrc/autograd/python_variable.cpp
colesbury: Uh... [Boo-urns](https://www.youtube.com/watch?v=2Kpb8eu1pEY)

I meant to pull this check out to the callers. I'll adjust it now

torch/csrc/autograd/utils/grad_layout_contract.h
colesbury: This fixes a semi-related bug. I can pull it out into a separate PR if you prefer. Basically `grad_dtype` was ignored if we copied instead of stole.

albanD: Ho right.
FYI @soulitzer

albanD (approved):
Looks good.
Thanks for the simplification!

albanD (approved):
SGTM

PR #166347, opened 2025-10-27
[DebugMode] store stringify args by default

pianpwk:
DebugMode currently stores dispatch call args & kwargs, which is all intermediate tensors and more. This quickly OOMed on GPU when trying to debug some torchtitan / llama 8b models.

This defaults to storing the stringified version, adding a flag `DebugMode(store_original_args=True)` if users want to store the original args as-is (and for BC).

* #166440
* #166348
* __->__ #166347

### DIFF
--- a/test/distributed/tensor/debug/test_debug_mode.py
+++ b/test/distributed/tensor/debug/test_debug_mode.py
@@ -64,6 +64,10 @@ class TestDTensorDebugMode(TestCase):
         self.assertTrue(isinstance(debug_mode.operators[2], _RedistributeCall))
         self.assertEqual(next(iter(debug_mode.operators[1])), torch.ops.aten.mm.default)
 
+        # check stringification
+        self.assertTrue(hasattr(debug_mode.operators[0], "args_str"))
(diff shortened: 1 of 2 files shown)


torch/utils/_debug_mode.py
```diff
@@ -188,6 +237,7 @@ def __init__(
         record_realtensor=True,
         record_tensor_attributes=None,
         record_nn_module=False,
+        store_original_args=False,
```
yushangdi, line 240: nit: since we're getting more and more inputs, can we add some docstring on what does each of these input args do? Doesn't have to be in this PR.

test/distributed/tensor/debug/test_debug_mode.py
```diff
@@ -279,6 +284,8 @@ def test_tensor_attributes(self):
       aten::_unsafe_view(t: f32[64, 8], [8, 8, 8])""",
         )
 
+        self.assertTrue(hasattr(debug_mode.operators[0], "args"))
```
yushangdi, line 287: nit: check that the content of the args is correct?

yushangdi (approved):
Some minor comments, otherwise LGTM.

PR #166348, opened 2025-10-27
[DebugMode] dispatch call hooks

pianpwk:
Adds customizable hooks on `__torch_dispatch__` calls for logging/recording arbitrary values.

Recording hooks store the hook outputs for each call at `debug_mode.operators[*].record`
```python
with DebugMode() as debug_mode, DebugMode.dispatch_hooks(record_hook = some_func):
    # some compute    
    ...
```

Logging hooks annotate the string dump:
```python
with DebugMode() as debug_mode, DebugMode.dispatch_hooks(log_hook = some_func):
    ...
```

Adds default hooks `DebugMode.record_outputs()` and `DebugMode.log_tensor_hashes()`, for checking numerical equivalence. The hashing hook borrows from the Observer. Example dump:
```
aten::sum(dt: f32[8, 32]| S(0))
  aten::sum(t: f32[1, 32])  # {'hash': 3.2215590476989746}
  _c10d_functional::all_gather_into_tensor(t: f32[1, 32], 8, 0)  # {'hash': 204.8783062621951}
  _c10d_functional::wait_tensor(t: f32[8, 32])  # {'hash': 204.8783062621951}
  aten::mm(t: f32[1, 8], t: f32[8, 32])  # {'hash': 12.014171155635267}
  aten::sum(t: f32[1, 32])  # {'hash': 3.2215590476989746}
  aten::t(t: f32[1, 8])  # {'hash': 3.7167285680770874}
  aten::detach(t: f32[8, 1])  # {'hash': 3.7167285680770874}
...
```

On the FSDP2 / simple FSDP NE in https://github.com/pytorch/pytorch/pull/164939, with hashing, this produces 2 log dumps (FSDP2: P2010198620, simple FSDP: P2010198963). I asked Claude to check the hashes, it wrote an analysis script, and was able to guess RMS norm as the root cause: P2010195076

Another throw-away example for logging per-op memory usage: https://gist.github.com/pianpwk/372082bf29467aa4aa25cb26dee24aea

* #166440
* __->__ #166348


yushangdi (approved):

torch/utils/_debug_mode.py
```diff
@@ -86,7 +86,9 @@ def to_str(x):
 
 def default_hash_fn(t: torch.Tensor, use_scalar: bool = False) -> torch.Tensor:
     """
-    from Observer. Computes a hash for a tensor by converting it to float (if needed), making it contiguous,
+    from Observer (genai/llama4x/tools/hash/observer.py)
```
yushangdi: maybe leave the path out?

torch/utils/_debug_mode.py
```diff
@@ -383,6 +451,56 @@ def debug_string(self) -> str:
...
+    @contextlib.contextmanager
+    def log_output_hashes(hash_fn: Optional[Callable] = None):
+        if hash_fn is None:
+            hash_fn = functools.partial(default_hash_fn, use_scalar=True)
+
+        def _dispatch_hash_hook(func, types, args, kwargs, result):
+            with torch._C._DisablePythonDispatcher():
+                out = tree_map(
+                    lambda x: hash_fn(x) if isinstance(x, torch.Tensor) else None,
+                    result,
```
mikaylagawarecki: Does `out` here ever come from c10d collectives? If so I expect you might have to .wait() the Work object returned by the collective before hashing the outputs of the c10d collectives themselves to match across runs.

I couldn't tell whether this was the case in P2010198963 and P2010198620 as the sequence of ops seem to be different

I suppose if you are using claude to compare hashes this does not matter too much, but if you want hashes to match exactly to make sure collectives aren't the cause of the issue (e.g.if NVLS is turned on, then I think this matters)

mikaylagawarecki (commented):
As a more nit comment I think there are classes of ops that you can just skip hashing (e.g. empty, profiler ops etc.) [[link](https://www.internalfb.com/code/fbsource/[82330672c52b3a6babf491c4d7d7c81a753cca28]/genai/llama4x/tools/hash/observer.py?lines=101-102)] 


A nice to have might be to pass a file with the previous sequence of hashes to DebugMode and compare the hashes at runtime like Observer provides an option to, but I don't think that's necessary

torch/utils/_debug_mode.py
pianpwk: Ah yes, we definitely see c10d collectives here. I've been debugging assuming collective hashes are unreliable, but probably we should have this option.

Did you ever run into issues with .wait() calls in the Observer, interfering with the actual model execution? I don't know if that's a legitimate concern, but I'm worried waiting by default will cause unexpected weirdness.

pianpwk: will handle this in a followup PR :)

PR #166349, opened 2025-10-27
[dynamo] Add __iter__ for iterable VariableTrackers

rtimpe:
* #166350
* #162768
* __->__ #166349

This is in preparation for implementing iter with a polyfill


torch/_dynamo/variables/dicts.py
```diff
@@ -787,12 +794,16 @@ def unpack_var_sequence(self, tx):
...
-        # OrderedDict though requires side effects tracking because it supports arbitrary setattr.
-        if self.user_cls is dict:
-            if name in self.user_cls.__dict__:
+        # dict not allow setting arbitrary attributes.  OrderedDict and
+        # defaultdict allow arbitrary setattr, but not deletion of default attrs
+        if any(
+            self.user_cls is t
+            for t in (dict, collections.OrderedDict, collections.defaultdict)
+        ):
+            if hasattr(self.user_cls, name):
```
guilhermeleobas, line 803: Can `hasattr` execute user-code? It might be safer to check `self.user_cls.__dict__`?

guilhermeleobas (changes requested):

torch/_dynamo/variables/dicts.py
rtimpe: I think it's safe as written since we just checked that `self.user_cls` is a builtin type.  But it's risky in that someone could rewrite that if statement and not update the `hasattr`.

Checking `__dict__` would be safer, but for `defaultdict` and `OrderedDict` it doesn't include methods inherited from the base `dict`.  For example
```
>>> 'keys' in collections.defaultdict.__dict__
False
```

guilhermeleobas (approved):

PR #166351, opened 2025-10-27
[precompile] Pickle and check closure variable properly.

zhxchen17:
Summary:

Previously we didn't correctly handle closure tuple when there's content in it. Adding additional code for serializing the tuple and merge it with guard manager local scope.

Test Plan:

pytest test/dynamo/test_aot_compile.py

Reviewers:

Subscribers:

Tasks:

Tags:

Fixes #ISSUE_NUMBER

### DIFF
--- a/test/dynamo/test_aot_compile.py
+++ b/test/dynamo/test_aot_compile.py
@@ -84,6 +84,21 @@ class RepeatInterleaveModule(torch.nn.Module):
         return y_repeat
 
 
+class MultiModalMixin(torch.nn.Module):
+    def forward(self, x):
+        return super().forward(x)
+
+
+class TextModel(torch.nn.Module):
+    def forward(self, x):
+        return x + 1
+
+
+class TestVLLMModel(MultiModalMixin, TextModel):
+    def forward(self, x):
+        return super().forward(x)
+
+
 @torch._dynamo.config.patch("enable_aot_compile", True)
 @instantiate_parametrized_tests
 class TestAOTCompile(torch._inductor.test_case.TestCase):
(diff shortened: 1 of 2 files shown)


test/dynamo/test_aot_compile.py
```diff
@@ -532,6 +532,25 @@ def foo(x):
         )
         self.assertEqual(compiled_foo(inputs), foo(inputs))
 
+    def test_aot_compile_with_closure_save_and_load(self):
```
Lucaskabela, line 550: Can we add a test with the `self.__class__.mro()` call in it?

This is currently what fails in vLLM (https://github.com/vllm-project/vllm/blob/53a56e658b9ab1eabb7339d3001dc3fd9178dd21/vllm/model_executor/models/transformers/multimodal.py#L317)

Maybe compiling an nn.Module that calls it will properly trigger the path

torch/_dynamo/aot_compile.py
```diff
@@ -58,13 +59,36 @@ def check_compatibility(self) -> None:
...
+
+
 @dataclass
 class AOTCompiledFunction:
     _artifacts: CompileArtifacts
     _guard_check_enabled: bool = True
 
     def guard_check(self, *args: Any, **kwargs: Any) -> bool:
-        f_locals = bind_locals(self._artifacts.signature, *args, **kwargs)
+        assert len(self._artifacts.closure) == len(self._artifacts.bytecode.co_freevars)
```
Lucaskabela: Think we need to add a `or None` check here as this causes error

Lucaskabela (changes requested):
Minor comments here on passing test and adding coverage

torch/_dynamo/aot_compile.py
```diff
@@ -58,13 +59,41 @@ def check_compatibility(self) -> None:
...
 @dataclass
 class AOTCompiledFunction:
     _artifacts: CompileArtifacts
     _guard_check_enabled: bool = True
 
     def guard_check(self, *args: Any, **kwargs: Any) -> bool:
-        f_locals = bind_locals(self._artifacts.signature, *args, **kwargs)
+        f_locals: dict[str, Any] = {}
+        if self._artifacts.closure:
+            assert self._artifacts.bytecode.co_freevars
```
Lucaskabela: nit: can we combine these asserts with `and`?

test/dynamo/test_aot_compile.py
```diff
@@ -84,6 +84,21 @@ def forward(self, x):
...
+    def forward(self, x):
+        return super().forward(x)
+
+
+class TextModel(torch.nn.Module):
+    def forward(self, x):
+        return x + 1
+
+
+class VLLMModel(MultiModalMixin, TextModel):
```
Lucaskabela: nit: We should avoid calling this VLLMModel directly to avoid confusion from users - lets prefix it with `Test` -> `TestVLLMModel`

Lucaskabela (approved):
Two minor nits, but those can be fixed without follow up review - thanks for helping to support this!

PR #166357, opened 2025-10-28
[ROCm] Enable multi-arch compilation and unit tests for AOT Inductor

chinmaydk99:
## Summary
This PR adds multi-architecture kernel compilation support for ROCm in PyTorch's AOT Inductor module, enabling a single compiled model to run across multiple AMD GPU architectures (MI200, MI300, MI350, etc.) without recompilation.

## Implementation
- **Multi-arch compilation pipeline**: Compiles LLVM IR to multiple GPU architectures and bundles them using `clang-offload-bundler`
- **Architecture detection**: Automatically detects target architectures from `torch.cuda.get_arch_list()`, with overrides via `PYTORCH_ROCM_ARCH` environment variable
- **ROCm-specific utilities**: New `rocm_multiarch_utils.py` module handles ROCm toolchain integration
- **Test infrastructure**: Adapted AOT Inductor tests to support both CUDA and ROCm compilation paths

## Testing
Successfully tested on:
- MI200
- MI300 

**Enabled tests:**
- `test_simple_multi_arch`
- `test_compile_after_package_multi_arch`
- `test_compile_with_exporter`
- `test_compile_with_exporter_weights`

cc @jeffdaily @sunway513 @jithunnair-amd @pruthvistony @ROCmSupport @dllehr-amd @jataylo @hongxiayang @naromero77amd @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben


test/inductor/test_aot_inductor_package.py
```diff
@@ -515,10 +513,9 @@ def default(*args, **kwargs):
                         )
 
     @unittest.skipIf(
-        _get_torch_cuda_version() < (12, 6), "Test is only supported on CUDA 12.6+"
+        torch.version.hip is None and _get_torch_cuda_version() < (12, 6), "Test is only supported on CUDA 12.6+"
```
jataylo: Just wondering why we need this change, seems that skipping if torch.version.hip is None is same as skipping on NV. We should just need to remove the skipIfRocm line IIUC.

torch/_inductor/rocm_multiarch_utils.py
```diff
@@ -0,0 +1,301 @@
...
+            if archs:
+                log.info(f"Using ROCm architectures from config: {archs}")
+                return archs
+    except Exception as e:
+        log.debug(f"Could not read config.rocm.target_archs: {e}")
+
+    # Default to common MI300/MI450 architectures
+    default_archs = ["gfx90a", "gfx942", "gfx1100", "gfx1101"]
+    log.info(f"Using default ROCm architectures: {default_archs}")
+    return default_archs
```
jeffdaily: When torch is built, it saves the gfx targets used for the build.

```suggestion
    return torch.cuda.get_arch_list()
```

torch/export/experimental/__init__.py
```diff
@@ -420,13 +420,15 @@ def _compiled_and_package(
                     path = Path(base_directory) / f"{name}_input_{i}.pt"
                     torch.save(t, path)
 
-        cmake_file_str = _get_make_file(package_name, model_names, use_cuda)
+        # Detect if ROCm is being used
+        is_hip = bool(getattr(torch.version, "hip", None))
```
jeffdaily: Why not just `torch.version.hip`?  Or `torch.version.hip is not None`?  Are you finding cases where `hip` as an attr doesn't exist?

torch/_inductor/codecache.py
```diff
@@ -1680,30 +1680,42 @@ def set(
...
-                specified_dir=split_aot_inductor_output_path(
-                    config.aot_inductor.output_path
-                )[0],
-                # make sure asm file has the same basename
-                key=basename,
-            )
+            # Allow ROCm single-arch to skip (asm=None OK), require for everything else
+            if torch.version.hip is None or (asm and asm_type):
+                assert asm, "Missing kernel assembly code"
+                assert asm_type, "Missing kernel assembly type"
```
jeffdaily, line 1702: If `torch.version.hip is None`, then we're not ROCm.  You say (asm=None OK) on ROCm, but the `or (asm and asm_type)` confused me.  If `asm=None` on ROCm, then the `asm and asm_type` is also `None`, so does this clause ever evaluate for ROCm or was it never supposed to?

jeffdaily (changes requested):

torch/_inductor/codecache.py
chinmaydk99: Yes, it evaluates True for ROCm multi-arch because we capture LLVM IR in `triton_heuristics.py` (so `asm` is NOT None). 

For ROCm single-arch, `asm` IS None, so it correctly skips.

The logic is:
- CUDA/XPU: Always enters (left side True)
- ROCm multi-arch: Enters (right side True - has LLVM IR)  
- ROCm single-arch: Skips (right side False - no asm)

torch/export/experimental/__init__.py
chinmaydk99: You're right, this was an oversight from my end. I'll change it to torch.version.hip to stay consistent with rest of the codebase

jataylo (commented):
^

test/inductor/test_aot_inductor_package.py
jataylo: Oh I get this now, we want to make sure tests are running on ROCm when _get_torch_cuda_version() < (12, 6), makes sense.

jeffdaily (approved):

PR #166369, opened 2025-10-28
Add DTensor dispatch key (but don't do anything special with it yet)

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #166370
* __->__ #166369
* #166368

Just a baby step to make sure we and CI are happy with this before moving on to a combined fast path.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @ezyang @gchanan

PR #166368, opened 2025-10-28
Refactor: extract OperatorArgsKwargsView from parseIValuesToPyArgsKwargs

swolchok:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #167051
* __->__ #166368

Intended to make it easier to reuse this logic for processing operator arguments as IValues in following PR(s).

Testing: python test/test_python_dispatch.py (broke during development, seems to work now)

PR #166370, opened 2025-10-28
Do our own call into Python for DTensor dispatch

swolchok:
* __->__ #166370
* #166369
* #166368

Another incremental step: don't deal with custom ops on the critical
path, let the dispatcher do that. Take control of calling into Python
on the critical path as well.


torch/csrc/autograd/python_variable.cpp
```diff
@@ -771,7 +771,10 @@ static Tensor make_tensor_for_subclass_helper(
   Storage storage{
       Storage::use_byte_size_t{},
       size_bytes,
-      at::DataPtr{nullptr, options.device()},
+      // REVIEW: without the device change here,
+      // test_dtensor_save_load breaks. I'm not at all sure why it was
+      // OK before.
```
swolchok: based on test failures, changing the device to meta here clearly isn't correct . will go back.

swolchok:
running into some weird coupling here. `test_dtensor_save_load` is broken. (repro: `python test/distributed/tensor/test_dtensor.py -k test_dtensor_save_load`, you need to patch the test to reduce world_size to 1 if you have only one GPU)
Apparently, the presence of `DTensor.__torch_dispatch__` somehow causes DTensors' storage to have a valid data_ptr and thus be compatible with serialization. This PR deletes `DTensor.__torch_dispatch__` and breaks it; I know it's the mere presence of `DTensor.__torch_dispatch__` because 1) putting it back fixes the test 2) printing inside `DTensor.__torch_dispatch__` shows that it's not called before or during serialization.

torch/_tensor.py
```diff
@@ -462,7 +467,10 @@ def _reduce_ex_internal(self, proto):
             return (torch._utils._rebuild_nested_tensor, args_nested)
         elif (
             type(self) is not torch.Tensor
-            and type(self).__torch_dispatch__ is not torch.Tensor.__torch_dispatch__
+            and (
+                type(self).__torch_dispatch__ is not torch.Tensor.__torch_dispatch__
+                or _is_dtensor(self)
+            )
```
swolchok: not sure how to structurally avoid is_dtensor special cases in N things that currently just care about `__torch_dispatch__`

swolchok: thinking about whether we put DTensor.__torch_dispatch__ back and have it just upcall to the DTensor dispatch key stuff

swolchok:
looks like the test failure is telling us we need to respect monkey-patching of custom_op_handlers. this will cost us a little performance but I guess it's what we have to do

swolchok:
noting that basic distributed tests (the two jobs whose names mention `test(distributed, `) are green

torch/csrc/autograd/python_variable.cpp
```diff
@@ -844,12 +847,21 @@ static PyObject* THPVariable_make_wrapper_subclass(
...
+
+  // DTensor is known to have __torch_dispatch__ and we have to check
+  // for it so that we correctly set up the DTensor dispatch key
+  // anyway.
+  const auto dtensor = get_dtensor_class();
+  const bool is_dtensor =
+      cls == dtensor.ptr() || checked_issubclass(cls, dtensor.ptr());
+  if (!is_dtensor) {
+    py::object attr = PyObject_FastGetAttrString(cls, "__torch_dispatch__");
+    TORCH_CHECK_TYPE(
```
albanD, line 859: What is this guarding? That we don't remove the torch_dispatch handler for DTensor?

torch/csrc/autograd/python_variable.cpp
```diff
@@ -865,13 +877,19 @@ static PyObject* THPVariable_make_wrapper_subclass(
   // data
   auto sym_sizes = r.symintlist(1);
   auto sym_strides_own = r.symintlistOptional(2);
+  std::optional<DispatchKeySet> extra_dispatch_keys =
+      r.toDispatchKeySetOptional(13);
+  if (is_dtensor) {
+    extra_dispatch_keys = extra_dispatch_keys.value_or(DispatchKeySet())
```
albanD, line 883: You have both Python and DTensor keys here?
Why not just one?

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1010,6 +1094,70 @@ static c10::SymDimVector tuple_to_symintlist(PyObject* obj) {
...
 }
 
+static constexpr c10::DispatchKeySet after_Python_keyset =
+    c10::DispatchKeySet(c10::DispatchKeySet::FULL) ^
+    (c10::DispatchKeySet(
+         c10::DispatchKeySet::FULL_AFTER,
+         c10::DispatchKey::Python) |
+     c10::DispatchKeySet(c10::DispatchKey::Python));
+
+// This is much simpler than the __torch_function__ paths precisely
```
albanD: I'm a bit confused on how this handles the `return NotImplemented` pattern to ensure that `DTensor + OtherClass` can have DTensor return NotImplemented, to get the OtherClass handler to do it's thing, before trying DTensor again

torch/csrc/autograd/python_variable.cpp
swolchok: it is the same check as before -- that subclasses passed to make_wrapper_subclass must define `__torch_dispatch__`. The comment on lines 851-853 was intended to clarify that we don't check in the case of the class being DTensor because we already know that DTensor defines `__torch_dispatch__`.

torch/csrc/autograd/python_variable.cpp
swolchok: because the `__torch_dispatch__` path needs to continue to work for DTensor. see breadcrumbs in my [previous comments](https://github.com/pytorch/pytorch/pull/166370#discussion_r2478850657) where I discovered during debugging that we have a lot of code that simply checks for `__torch_dispatch__`.

torch/csrc/autograd/python_variable.cpp
swolchok: hmm. Do we really have 0 tests for that? I'll try to find one and then find out whether they ran. (If we have tests and they ran, well, tests are green :) )

torch/csrc/autograd/python_variable.cpp
```diff
@@ -1010,6 +1094,70 @@ static c10::SymDimVector tuple_to_symintlist(PyObject* obj) {
...
+// because it doesn't have to worry about anything to do with dispatch
+// or __torch_function__.
+void callDTensorCustomOpHandler(
```
swolchok: hmm, given that we have to respect monkey patching of _custom_op_handlers, I think that having the explicit hardcoded custom op handler path 1) is confusing 2) doesn't respect monkey patching that removes the hardcoded op handlers. I'll remove it.

torch/csrc/autograd/python_variable.cpp
swolchok: grepping test/ for NotImplemented, it looks like we probably don't have DTensor-specific NotImplemented tests. will have to add one. Do you have any examples of code you would expect to work (and ideally a pointer to how it works in the `__torch_dispatch__` regime)?

albanD: I wouldn't be surprised if we don't.
cc @drisspg for FP8Tensor combination and @amjames for ComplexTensor combinations. Do either of you have minimal example we could add as a sanity check here?

torch/csrc/autograd/python_variable.cpp
albanD: Some code also check for the Python key being set IIRC

albanD: I'm not sure how we can make the disable logic that exclude the python key also work though.
We should check https://github.com/pytorch/pytorch/blob/main/c10/core/impl/TorchDispatchModeTLS.cpp , https://github.com/pytorch/pytorch/blob/main/torch/csrc/utils/disable_torch_function.h#L24 and
 https://github.com/pytorch/pytorch/blob/main/torch/csrc/utils/tensor_new.cpp#L382

but @ezyang should we make Python and alias key that point to both DTensor key and the oldPython key? :D

torch/csrc/autograd/python_variable.cpp
drisspg: So we have the code: https://github.com/pytorch/ao/blob/01374eb58a4f9ba52780efbe0ef4e056e36d338c/torchao/float8/float8_training_tensor.py#L370-L371

 but tbh I dont remember if this is used anymore or we have unpacked by the time we get here these days

amjames: Based on a fairly quick review of what this PR and others in the stack are doing I am convinced there will be compatibility issues if we don't stand up some kind of testing to cover that before moving forward.

@hameerabbasi is working to fold in DTensor composability tests at https://github.com/openteams-ai/pytorch-complex-tensor/pull/18 for our ComplexTensor subclass, but even just running those tests we would need to diff the errors since there are currently many failures hameer is working to diagnose he has ID's some upstream fixes already but we do not have issues open yet. 

We probably should set up a test for composability b/t DTensor and user defined subclasses AFAICT we don't have any. It should be possible to create a variant of the DTensor tests that applies another subclass as a wrapper or applies DTensor as a wrapper over another subclass. We have some subclasses defined for [testing purposes](https://github.com/pytorch/pytorch/blob/d77c24caac4b42c56b4fa6a156ce85fb4907643e/torch/testing/_internal/common_subclass.py#L269) and the subclass tests give an [example](https://github.com/pytorch/pytorch/blob/397d9fe2aea0dc60ea19ffddf6ac750420362867/test/test_subclass.py#L37) of how to parameterize a test case on that list.

swolchok: My tentative plan for sidestepping these issues is to experiment with abandoning the DTensor dispatch key and instead patching the Python dispatch key path to detect DTensor and, if found, replace the final call into Python to execute the `__torch_dispatch__` function with a call to the C++ fast path. My hope is that such a change will make it simple and obvious to keep DTensor working the same with respect to `__torch_dispatch__` modes/other subclasses/etc. as any other class, while still having a fast path.

torch/distributed/tensor/_dispatch.py
```diff
@@ -135,6 +135,16 @@ def _allow_implicit_replication(self) -> bool:
     def _allow_implicit_replication(self, value: bool) -> None:
         return torch._C._set_dtensor_allow_implicit_replication(value)
 
+    def custom_op_handler(
```
swolchok, line 138: whoops, forgot to remove this

torch/csrc/autograd/python_variable.cpp
swolchok: preview is up as https://github.com/pytorch/pytorch/pull/167051

swolchok:
superseded by #167051

PR #166371, opened 2025-10-28
DTensor dispatch fast path: C++ unwrap_to_op_info, OpSchema creation, & cached sharding prop

swolchok:
* #166808
* #166372
* __->__ #166371
* #166370
* #166369
* #166368

Incremental addition of C++ fast path, taking advantage of the DTensor
dispatch key to let us work with IValues without Python conversion for
the fast path. I tried to just port unwrap_to_op_info to C++, but that
didn't get much of a win; the nice thing here seems to be the fusion
of unwrap_to_op_info with recompute_comparison_key.

This + the following PR appear to reduce DTensor dispatch time for
detach() from 36-40 usec to 24-26 usec (using a benchmark similar to the one on #160580 on a Skylake machine.


torch/csrc/autograd/python_variable.cpp
```diff
@@ -1124,36 +1167,242 @@ void callDTensorCustomOpHandler(
...
+      const c10::OperatorHandle& op,
+      c10::SmallVector<IValueOrDTensorSpec, 8> comparison_key,
+      std::size_t comparison_key_hash,
+      std::size_t args_schema_len)
+      : op_(op),
+        hash_(hash_combine(
+            hash_combine(
+                std::hash<c10::OperatorHandle>()(op),
+                comparison_key_hash),
+            args_schema_len)),
```
swolchok, line 1218: note that including args_schema_len in the hash is a behavior change, but I believe it's an excellent one: args_schema_len was in the `OpSchema.__eq__` implementation all along, so not including it in the hash would just have been leading to unnecessary collisions.

PR #166373, opened 2025-10-28
Move static from_ivalue/to_ivalue to new shim_common.cpp

mikaylagawarecki:
Move `from_ivalue` and `to_ivalue` and their dependents `StableIValueBoxedKernel`, `aoti_torch_library_impl` `aoti_torch_call_dispatcher` into new (non-aoti shim_common.cpp)

This is in prep for the above PRs where I add v2s (`torch_call_dispatcher` and `torch_library_impl`) that are versioning aware

* #165285
* #165284
* #165283
* #164332
* #163683
* __->__ #166373
* #164356

### DIFF
--- a/build_variables.bzl
+++ b/build_variables.bzl
@@ -87,6 +87,7 @@ core_sources_common = [
     "torch/csrc/jit/mobile/promoted_prim_ops.cpp",
     "torch/csrc/jit/mobile/prim_ops_registery.cpp",
     "torch/csrc/profiler/util.cpp",
+    "torch/csrc/shim_common.cpp",
 ]
 
 torch_unpickler_common = [
--- a/torch/csrc/shim_common.cpp
+++ b/torch/csrc/shim_common.cpp
@@ -1,0 +1,220 @@
+#include <c10/core/DispatchKey.h>
+#include <c10/util/Exception.h>
+#include <torch/csrc/inductor/aoti_runtime/utils.h>
+#include <torch/csrc/inductor/aoti_torch/c/shim.h>
+#include <torch/csrc/inductor/aoti_torch/tensor_converter.h>
+#include <torch/csrc/inductor/aoti_torch/utils.h>
+#include <torch/csrc/jit/serialization/pickle.h>
+#include <torch/csrc/stable/library.h>
+#include <torch/library.h>
+
+using namespace torch::aot_inductor;
+
+static StableIValue from_ivalue(
+    const c10::TypePtr& type,
+    const c10::IValue& ivalue) {
+  switch (type->kind()) {
+    case c10::TypeKind::TensorType: {
+      AtenTensorHandle ath = torch::aot_inductor::new_tensor_handle(
+          std::move(const_cast<at::Tensor&>(ivalue.toTensor())));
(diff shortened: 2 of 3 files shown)


build_variables.bzl
```diff
@@ -482,6 +482,7 @@ inductor_core_resources = [
     "torch/csrc/inductor/aoti_torch/oss_proxy_executor.cpp",
     "torch/csrc/inductor/inductor_ops.cpp",
     "torch/csrc/jit/serialization/pickle.cpp",
+    "torch/csrc/shim_common.cpp",
```
janeyx99, line 485: should this actually go in core_sources_common? or core_sources_full instead? since this doesn't have to do with inductor.

torch/csrc/shim_common.cpp
```diff
@@ -0,0 +1,220 @@
...
+#include <c10/util/Exception.h>
+#include <torch/csrc/inductor/aoti_runtime/utils.h>
+#include <torch/csrc/inductor/aoti_torch/c/shim.h>
+#include <torch/csrc/inductor/aoti_torch/tensor_converter.h>
+#include <torch/csrc/inductor/aoti_torch/utils.h>
+#include <torch/csrc/jit/serialization/pickle.h>
+#include <torch/csrc/stable/library.h>
+#include <torch/library.h>
+
+using namespace torch::aot_inductor;
```
janeyx99: how come we need this?

build_variables.bzl
mikaylagawarecki: I had put it in core_sources_common previously and imported, but that caused all the mobile builds to fail

janeyx99 (commented):
looks fine to me, barring two questions

build_variables.bzl
janeyx99: hmmmmmmmmmmmmmmmmmmmmmmmm.

i think it may be important to figure out why that is 😛  otherwise this is a bit weird why it needs to be in inductor.

mikaylagawarecki: I think the reason why is that this file includes the aoti shim (whose cpp file is only in the inductor sources)

So if we put this in core_sources_common, it would be included by the mobile builds ([specifically this line](https://www.internalfb.com/code/fbsource/[609be73ed60908657fad3d41d167064f53e28c0d]/fbcode/caffe2/ovrsource_defs.bzl?lines=1438)), but then the aoti shim would be excluded

Separately, this new shim_common.cpp now implements some of aoti's shim.h, so I believe it's needed here https://github.com/pytorch/pytorch/blob/5e47b4dd60ff9efb253286af5a2479d9d800ce6a/caffe2/CMakeLists.txt#L426-L428

torch/csrc/shim_common.cpp
mikaylagawarecki: you're right this is unnecessary, removed

build_variables.bzl
janeyx99: Still not sure I fully understand if this is the right move, but given that it seems ok + that this is undoable, I am approving.

janeyx99 (approved):

PR #166361, opened 2025-10-28
Avoid DDE in narrow with unbacked start

laithsakka:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #166379
* __->__ #166361

Slice knows how to handle unbacked start, we do not need to offset start before calling slice, we can leave it for slice.
The only edge case is when start<0 and start+length ==0 in that case slice and narrow would deviate, 
for that case we shall pass dim_size instead of start+length 

cc @jgong5 @mingfeima @XiaobingSuper @sanchitintel @ashokei @jingxu10 @jerryzh168 @aditew01 @voznesenskym @penguinwu @EikanWang @Guobing-Chen @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben

PR #166379, opened 2025-10-28
make narrow_tensor_symint DDE-free

laithsakka:
* __->__ #166379
* #166361

 https://github.com/pytorch/pytorch/issues/158081


Lucaskabela (changes requested):
This change seems to have caused some odd test failures related to formatting, but otherwise the content looks good.  Requesting changes for viz on the test failures

test/test_proxy_tensor.py
```diff
@@ -1946,75 +2234,89 @@ def f(t):
...
 }
 
 only_real_tensor_failures = {
     xfail('narrow'),
-    xfail('tensor_split'),
+    xfail("tensor_split"),
 }
 
 only_fake_tensor_failures = {
-    xfail('narrow'),
```
Lucaskabela, line 1990: This is the only substantial change I could find in this file - correct me if wrong :)

Lucaskabela (approved):
Lol so many changes to `test_proxy_tensor` XD I am assuming linter did a lot 

Other than that though, this change looks good to me - I don't really understand the failure from the cpp.jit test but assuming it is a change in the positive direction we can approve

test/test_proxy_tensor.py
laithsakka: right but this is k we do not fail now.

laithsakka: ok why is that let me look into that lol.

PR #160078, opened 2025-08-07
Move prioritized text linker optimization code from setup.py to cmake

robert-hardwick:
Note. This is a replica PR of #155901 which will be closed. I had to create a new PR in order to add it into my ghstack as there are some later commits which depend on it.

### Summary

🚀 This PR moves the prioritized text linker optimization from setup.py to cmake ( and enables by default on Linux aarch64 systems )

This change consolidates what was previously manual CI logic into a single location (cmake), ensuring consistent behavior across local builds, CI pipelines, and developer environments.

### Motivation
Prioritized text layout has measurable performance benefits on Arm systems by reducing code padding and improving cache utilization. This optimization was previously triggered manually via CI scripts (.ci/aarch64_linux/aarch64_ci_build.sh) or user-set environment variables. By detecting the target architecture within setup.py, this change enables the optimization automatically where applicable, improving maintainability and usability.

Note:

Due to ninja/cmake graph generation issues we cannot apply the linker file globally to all targets to the targets must be manually defined. See CMakeLists.txt the main libraries torch_python, torch, torch_cpu, torch_cuda, torch_xpu have been targetted which should be enough to maintain the performance benefits outlined above.

Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #162810
* #160079
* #159737
* __->__ #160078

Co-authored-by: Usamah Zaheer <usamah.zaheer@arm.com>

cc @malfet @snadampal @milpuz01 @aditew01 @nikhil-arm @fadara01 @usamahz

PR #166380, opened 2025-10-28
Revert: "Move prioritized text linker optimization code from setup.py to cmake #160078"

Aidyn-A:
This PR reverts #160078 as it does not add any flags to both linker and compiler.


robert-hardwick (changes requested):
prefer not to merge this until we have checked again after moving the summary print

robert-hardwick (approved):

PR #166381, opened 2025-10-28
Replace c10::call_once with static initialization

cyyever:
This PR replaces c10::call_once calls with static initialization when possible. C++11 semantics guarantees that static initialization is atomic. Static initialization also has lower cost than using c10::call_once.

### DIFF
--- a/aten/src/ATen/cuda/detail/BLASConstants.cu
+++ b/aten/src/ATen/cuda/detail/BLASConstants.cu
@@ -2,8 +2,6 @@
 #include <ATen/Tensor.h>
 #include <ATen/cuda/Exceptions.h>
 
-#include <mutex>
-
 namespace at {
 namespace cuda {
 namespace detail {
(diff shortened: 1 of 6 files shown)


aten/src/ATen/cuda/detail/BLASConstants.cu
```diff
@@ -2,36 +2,20 @@
...
-#include <mutex>
-
 namespace at {
 namespace cuda {
 namespace detail {
 
-__device__ __constant__ float cublas_one_device;
-__device__ __constant__ float cublas_zero_device;
+__device__ __constant__ float cublas_one_device{1.f};
+__device__ __constant__ float cublas_zero_device{0.f};
```
malfet: IMO this one would not work, this is why you still need to have `cudaMemcpyToSymbol`

malfet (changes requested):
Everything looks good but change to BLASConstants.cu (not sure if are they still used anywhere though)

aten/src/ATen/cuda/detail/BLASConstants.cu
cyyever: That is interesting. I believe you, and it has to be reverted

cyyever:
@malfet Your suggestions are applied.

malfet (approved):
LGTM

PR #166053, opened 2025-10-22
[inductor] track reduction before splitting

shunting314:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #166669
* #166697
* #166675
* #166585
* #166461
* #166382
* __->__ #166053

Keep tracking of the reduction before splitting.

In the mix-order reduction context, if one of the reduction is split, it makes it much harder to fuse with the other reduction. Tracking the metadata of the reduction before splitting to make the fusion possible.

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben

PR #166382, opened 2025-10-28
[inductor] more aggressive mix order reduction

shunting314:
* #166669
* #166697
* #166675
* #166585
* #166461
* __->__ #166382
* #166053

More aggressive mix order reductions so that when rnumel is larger than 1024 we can still generate the fused kernel. Also use more warps in that case.


shunting314:
Checked the models failing accuracy test. 

For deit_base_distilled_patch16_224, I can repro on my h100. But the fused kernel looks correct: https://gist.github.com/shunting314/428f36ad11c7da9731113159f24c3bb2 . (unfused kernel for reference: https://gist.github.com/shunting314/5566f845acf85676bf6606cec94cb4f8 , https://gist.github.com/shunting314/3618b566b2b76a08845565f5e2d43157). Also using `--float` rather than `--amp ` solve the issue. I'll just raise tolerance to fix. Similar to vit_base_patch16_siglip_256 .

beit_base_patch16_224 is quite tricky. I can not repro on H100 dev server. While I can repro on the A10G used to run CI jobs. The issue is reproed even with `--float`. But if I change the wrapper and run both fused and non-fused kernel and compare the result, they are very close:
  ```
  # first iteration
  (out[0] - side[0]).abs().max()=tensor(2.7940e-09, device='cuda:0')
  (out[1] - side[1]).abs().max()=tensor(4.6566e-10, device='cuda:0')
  (out[0] - side[0]).abs().max()=tensor(8.1491e-10, device='cuda:0')
  (out[1] - side[1]).abs().max()=tensor(1.1642e-10, device='cuda:0')
  # second iteration
  (out[0] - side[0]).abs().max()=tensor(2.3283e-09, device='cuda:0')
  (out[1] - side[1]).abs().max()=tensor(3.4925e-10, device='cuda:0')
  (out[0] - side[0]).abs().max()=tensor(2.3283e-09, device='cuda:0')
  (out[1] - side[1]).abs().max()=tensor(8.1491e-10, device='cuda:0')
  ```
These small difference should be acceptable. Maybe small difference get amplified by the later operators.

jansel (approved):

v0i0 (approved):

PR #166386, opened 2025-10-28
Fixes absolute tolerance scaling for complex backpropagation in fast gradcheck

redwrasse:
Fixes #166385 

This PR updates the `_adjusted_atol` for complex tangent input vectors `u` and also adds a regression test to `test_autograd.py` for the updated behavior (compare with the script in https://github.com/pytorch/pytorch/issues/166385).

The current absolute tolerance adjustment (`atol`) for fast gradcheck doesn't incorporate the complex component of the input tangent vector. The bound should be corrected from a factor of the sum over absolute value of real components to sum of absolute values of both real and imaginary components. The end effect of this change would be a multiplicative factor of `2`.

I haven't checked, but this bug may be relevant if there are any flaky OpInfo gradient checks, particularly over large complex matrices, since the `atol` transform also scales like the dim. `d` of the vector `u`.

### DIFF
--- a/test/test_autograd.py
+++ b/test/test_autograd.py
@@ -6667,25 +6667,28 @@ Done""",
         torch.manual_seed(97)
 
         def sample_func(z):
-            return 1. / torch.norm(z)
+            return 1.0 / torch.norm(z)
 
         # Input needs to be at least 2-dim. to trigger
         # in gradcheck an input projection vector u
         # that is not all 1s.
         eps = 10e-3  # eps distance factor from origin, to get
         # some interesting numerical vs analytic discrepancy.
-        z = eps * torch.rand(2,
-                             dtype=torch.complex128,
-                             requires_grad=True,
-                             )
+        z = eps * torch.rand(
+            2,
+            dtype=torch.complex128,
+            requires_grad=True,
+        )
         atol = 8.3e-6
         rtol = 1e-9
 
         # check both fast and slow gradcheck pass after the fix to _adjusted_atol()
-        self.assertTrue(gradcheck(sample_func, (z,), fast_mode=True,
-                      atol=atol, rtol=rtol))
-        self.assertTrue(gradcheck(sample_func, (z,), fast_mode=False,
-                                  atol=atol, rtol=rtol))
+        self.assertTrue(
+            gradcheck(sample_func, (z,), fast_mode=True, atol=atol, rtol=rtol)
+        )
+        self.assertTrue(
+            gradcheck(sample_func, (z,), fast_mode=False, atol=atol, rtol=rtol)
+        )
 
     def test_gradcheck_get_numerical_jacobian(self):
         # get_numerical_jacobian is deprecated and no longer used internally by gradcheck
--- a/test/test_autograd.py
+++ b/test/test_autograd.py
@@ -6656,6 +6656,12 @@ Done""",
         c = torch.ones(2, 2, requires_grad=True, dtype=torch.complex128)
         self.assertTrue(gradcheck(fn2, (c)))
 
+    def test_gradcheck_adjusted_atol_complex_inputs(self):
+        # FIXME implement as regression test
+        # a certain input (proably large-min) with certain atol should fail the projection-specific check ('fast gradcheck') but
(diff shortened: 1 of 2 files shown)


hameerabbasi:
I didn't check the math here, but it seems that there are still cases where slow gradcheck passes, but fast fails, which should never happen according to #166385.

Example:

```
# Checkout and install PyTorch from this PR

pip install 'complex-tensor @ git+https://github.com/openteams-ai/pytorch-complex-tensor.git@03b8d7e15b5c8567c9d7e6deb5bf4bcbe6f7283c'
pytest --pyargs complex_tensor -k "test_fn_grad_rsub_cpu_complex128" # Succeeds

# Patch `site-packages/complex_tensor/tests/test_ops.py`, comment line 188
# Line 188 runs the tests in slow mode, and those succeed

pytest --pyargs complex_tensor -k "test_fn_grad_rsub_cpu_complex128" # Fails
```

redwrasse:
> I didn't check the math here, but it seems that there are still cases where slow gradcheck passes, but fast fails, which should never happen according to #166385.
> 
> Example:
> 
> ```
> # Checkout and install PyTorch from this PR
> 
> pip install 'complex-tensor @ git+https://github.com/openteams-ai/pytorch-complex-tensor.git@03b8d7e15b5c8567c9d7e6deb5bf4bcbe6f7283c'
> pytest --pyargs complex_tensor -k "test_fn_grad_rsub_cpu_complex128" # Succeeds
> 
> # Patch `site-packages/complex_tensor/tests/test_ops.py`, comment line 188
> # Line 188 runs the tests in slow mode, and those succeed
> 
> pytest --pyargs complex_tensor -k "test_fn_grad_rsub_cpu_complex128" # Fails
> ```
> 
> Edit: Specifically, it seems for an operation as simple as subtraction, the numerical gradient somehow goes to `nan+nanj`.

@hameerabbasi thanks for taking a look. Any way you could share here directly a seed and gradcheck input testcases(s) that are failing? That would help me investigate.

hameerabbasi:
> Any way you could share here directly a seed and gradcheck input testcases(s) that are failing? That would help me investigate.

@redwrasse It's hard to do this without a dependency on `complex_tensor`. Here's a test case that succeeds in slow mode and fails in fast mode with your PR, with the `complex_tensor` dep.

Install it with:

`pip install 'complex-tensor @ git+https://github.com/openteams-ai/pytorch-complex-tensor.git@03b8d7e15b5c8567c9d7e6deb5bf4bcbe6f7283c'`

```python
from __future__ import annotations

import torch
from torch.testing._internal.common_methods_invocations import op_db
from torch.testing._internal.common_utils import (
    TestGradients,
    run_tests,
    unMarkDynamoStrictTest,
)

from complex_tensor.ops._common import ComplexDispatchMode

aten = torch.ops.aten

@unMarkDynamoStrictTest
class TestComplexBwdGradients(TestGradients):
    def test_fn_grad(self) -> None:
        torch.manual_seed(42)
        op = next(op_candidate for op_candidate in op_db if op_candidate.name == "rsub")
        with ComplexDispatchMode():
            # op.gradcheck_fast_mode = False  # Uncomment for slow mode
            self._grad_test_helper(torch.device("cpu"), torch.complex128, op, op.get_op())


if __name__ == "__main__":
    run_tests()
```

albanD (approved):
Sounds good. Thanks!

redwrasse:
> Sounds good. Thanks!

Thanks! I'm seeing couple CI errors on lint and test `test/dynamo_expected_failures/TestNNParametrization.test_new_spectral_norm_forward_swap_True`, though I don't think they're related.

redwrasse:
> > Sounds good. Thanks!
> 
> Thanks! I'm seeing a couple CI errors on lint and test `test/dynamo_expected_failures/TestNNParametrization.test_new_spectral_norm_forward_swap_True`, though I don't think they're related.
> 
> (just rebased)

bump @albanD, in case you missed this :)

hameerabbasi (approved):
This LGTM, thanks @redwrasse. The subclass issue was fixed by #168248.

redwrasse:
@albanD @soulitzer,  bumping to get this merged, just in case it got lost.

PR #166391, opened 2025-10-28
Remove outdated skip conditons of CUDA and ROCm

cyyever:
This PR removes outdated skip conditions of CUDA 11 and ROCm 5


albanD (approved):
Thanks!

PR #166396, opened 2025-10-28
[xpu][test][1/N] Enable tests of test_nn.py on Intel GPU - instantiate TestNN with instantiate_device_type_tests

daisyden:
For https://github.com/pytorch/pytorch/issues/114850, we will port aten unit tests to Intel GPU. This PR will work on test/test_nn.py TEST_NN class for single GPU test only. We could enable Intel GPU with following methods and try the best to keep the original code styles:

1. Use torch.accelerator to extend cude specific test to XPU.
2. Added skipIfXPU decorator for cases with known issues on Intel GPU
3. Enabled 'xpu' for some test pathes

### DIFF
--- a/test/test_nn.py
+++ b/test/test_nn.py
@@ -35,9 +35,9 @@ from torch.testing._internal.common_utils import dtype_name, freeze_rng_state, r
     skipIfNoLapack, skipIfRocm, MI300_ARCH, skipIfRocmArch, \
     TEST_NUMPY, TEST_SCIPY, TEST_WITH_CROSSREF, TEST_WITH_ROCM, \
     download_file, get_function_arglist, load_tests, skipIfMPS, \
-    IS_PPC, IS_ARM64, IS_CPU_CAPABILITY_SVE256, IS_CPU_EXT_SVE_SUPPORTED, xfailIf, TEST_GPU, \
+    IS_PPC, IS_ARM64, IS_CPU_CAPABILITY_SVE256, IS_CPU_EXT_SVE_SUPPORTED, xfailIf, \
     parametrize as parametrize_test, subtest, \
-    skipIfTorchDynamo, gcIfJetson, set_default_dtype, HAS_ACCELERATOR
+    skipIfTorchDynamo, gcIfJetson, set_default_dtype, ACCELERATOR_TYPE
 from torch.testing._internal.common_cuda import TEST_CUDA, TEST_MULTIGPU, TEST_CUDNN, \
     _get_torch_rocm_version
 from torch.testing._internal.common_nn import NNTestCase, NewModuleTest, CriterionTest, \
@@ -231,13 +231,13 @@ class TestNN(NNTestCase):
         self.assertEqual(m.double(), m.to(torch.float64))
         self.assertRaises(RuntimeError, lambda: m.to('cpu', copy=True))
 
-        if HAS_ACCELERATOR:
+        if ACCELERATOR_TYPE == device:
             for gpu in [device,
                         str(torch.device(0)) if torch.accelerator.device_count() == 1 else str(torch.device(1))]:
-                m2 = m.to(device)
-                self.assertIs(m2, m2.to(device))
+                m2 = m.to(gpu)
+                self.assertIs(m2, m2.to(gpu))
                 self.assertEqual(m, m2.to('cpu'))
-                self.assertEqual(m2, m.to(device))
+                self.assertEqual(m2, m.to(gpu))
                 self.assertIs(m2, m2.to(dtype=torch.float32))
                 self.assertEqual(m2.double(), m2.to(dtype=torch.float64))
 
@@ -2670,30 +2670,28 @@ tensor(..., device='meta', size=(1,), requires_grad=True)""")
         expected_loss = torch.tensor(0.25)
         self.assertTrue(torch.isclose(loss, expected_loss), f"Expected {expected_loss}, but got {loss}")
 
-    def test_mse_loss_mixed_dtype_grad(self):
-        devices = ['cpu'] + (['cuda'] if TEST_CUDA else [])
-        for device in devices:
-            for reduction in ['mean', 'sum', 'none']:
-                x = torch.tensor([1.0, 2.0], dtype=torch.float32, requires_grad=True, device=device)
-                y = torch.tensor([1.5, 2.5], dtype=torch.float64, requires_grad=True, device=device)
-                loss = F.mse_loss(x, y, reduction=reduction)
-                if reduction == 'mean':
-                    expected_loss = torch.tensor(0.25, dtype=torch.float64, device=device)
-                    expected_grad_x = torch.tensor([-0.5, -0.5], dtype=torch.float32, device=device)
-                    expected_grad_y = torch.tensor([0.5, 0.5], dtype=torch.float64, device=device)
-                elif reduction == 'sum':
-                    expected_loss = torch.tensor(0.5, dtype=torch.float64, device=device)
-                    expected_grad_x = torch.tensor([-1.0, -1.0], dtype=torch.float32, device=device)
-                    expected_grad_y = torch.tensor([1.0, 1.0], dtype=torch.float64, device=device)
-                else:  # 'none'
-                    expected_loss = torch.tensor([0.25, 0.25], dtype=torch.float64, device=device)
-                    expected_grad_x = torch.tensor([-1.0, -1.0], dtype=torch.float32, device=device)
-                    expected_grad_y = torch.tensor([1.0, 1.0], dtype=torch.float64, device=device)
-                self.assertEqual(loss, expected_loss, atol=1e-6, rtol=0)
-                grad = torch.ones_like(loss) if reduction == 'none' else None
-                loss.backward(grad)
-                self.assertEqual(x.grad, expected_grad_x, atol=1e-6, rtol=0)
-                self.assertEqual(y.grad, expected_grad_y, atol=1e-6, rtol=0)
+    def test_mse_loss_mixed_dtype_grad(self, device):
+        for reduction in ['mean', 'sum', 'none']:
+            x = torch.tensor([1.0, 2.0], dtype=torch.float32, requires_grad=True, device=device)
+            y = torch.tensor([1.5, 2.5], dtype=torch.float64, requires_grad=True, device=device)
+            loss = F.mse_loss(x, y, reduction=reduction)
+            if reduction == 'mean':
+                expected_loss = torch.tensor(0.25, dtype=torch.float64, device=device)
+                expected_grad_x = torch.tensor([-0.5, -0.5], dtype=torch.float32, device=device)
+                expected_grad_y = torch.tensor([0.5, 0.5], dtype=torch.float64, device=device)
+            elif reduction == 'sum':
+                expected_loss = torch.tensor(0.5, dtype=torch.float64, device=device)
+                expected_grad_x = torch.tensor([-1.0, -1.0], dtype=torch.float32, device=device)
+                expected_grad_y = torch.tensor([1.0, 1.0], dtype=torch.float64, device=device)
+            else:  # 'none'
+                expected_loss = torch.tensor([0.25, 0.25], dtype=torch.float64, device=device)
+                expected_grad_x = torch.tensor([-1.0, -1.0], dtype=torch.float32, device=device)
+                expected_grad_y = torch.tensor([1.0, 1.0], dtype=torch.float64, device=device)
+            self.assertEqual(loss, expected_loss, atol=1e-6, rtol=0)
+            grad = torch.ones_like(loss) if reduction == 'none' else None
+            loss.backward(grad)
+            self.assertEqual(x.grad, expected_grad_x, atol=1e-6, rtol=0)
+            self.assertEqual(y.grad, expected_grad_y, atol=1e-6, rtol=0)
 
     def test_weighted_l1_loss_with_weights(self):
         inputs = torch.tensor([1.0, 2.0, 3.0, 4.0], requires_grad=True)
@@ -2854,8 +2852,6 @@ tensor(..., device='meta', size=(1,), requires_grad=True)""")
         self.assertEqual(res_cpu, res_gpu, atol=1e-3, rtol=0)
 
     def test_CTCLoss_zero_lengths(self, device):
-        devices = ['cpu']
-        devices += [device] if HAS_ACCELERATOR else []
         N = 3
         S = 2
         C = 200
@@ -2863,19 +2859,20 @@ tensor(..., device='meta', size=(1,), requires_grad=True)""")
         target = torch.randint(low=1, high=C, size=(N, S), dtype=torch.int)
         input_lengths = torch.full(size=(N,), fill_value=0, dtype=torch.int)
         target_lengths = torch.full(size=(N,), fill_value=0, dtype=torch.int)
-        for device in devices:
-            inp = torch.randn(T, N, C, dtype=torch.float, device=device).log_softmax(2).requires_grad_()
-            res = torch.nn.functional.ctc_loss(inp, target, input_lengths, target_lengths, reduction='none')
-            self.assertTrue((res == 0).all().item())
-            res.sum().backward()
-            self.assertTrue((inp.grad == 0).all().item())
+
+        inp = torch.randn(T, N, C, dtype=torch.float, device=device).log_softmax(2).requires_grad_()
+        res = torch.nn.functional.ctc_loss(inp, target, input_lengths, target_lengths, reduction='none')
+        self.assertTrue((res == 0).all().item())
+        res.sum().backward()
+        self.assertTrue((inp.grad == 0).all().item())
+
         target_lengths = torch.full(size=(N,), fill_value=1, dtype=torch.int)
-        for device in devices:
-            inp = torch.randn(T, N, C, dtype=torch.float, device=device).log_softmax(2).requires_grad_()
-            res = torch.nn.functional.ctc_loss(inp, target, input_lengths, target_lengths, reduction='none')
-            self.assertTrue((res == torch.inf).all().item())
-            res.sum().backward()
-            self.assertTrue((inp.grad == 0).all().item())
+
+        inp = torch.randn(T, N, C, dtype=torch.float, device=device).log_softmax(2).requires_grad_()
+        res = torch.nn.functional.ctc_loss(inp, target, input_lengths, target_lengths, reduction='none')
+        self.assertTrue((res == torch.inf).all().item())
+        res.sum().backward()
+        self.assertTrue((inp.grad == 0).all().item())
 
     @onlyAccelerator
     def test_CTCLoss_zero_infinity(self, device):
@@ -3631,7 +3628,7 @@ tensor(..., device='meta', size=(1,), requires_grad=True)""")
 
     @unittest.skipIf(not TEST_CUDNN, 'CUDNN not available')
     @tf32_on_and_off
-    def test_cudnn_weight_tying(self, device):
+    def test_cudnn_weight_tying(self):
         rnns = [
             nn.LSTM(10, 20, batch_first=True, bidirectional=True),
             nn.LSTM(10, 20, batch_first=True, bidirectional=True, proj_size=10),
@@ -3929,8 +3926,8 @@ tensor(..., device='meta', size=(1,), requires_grad=True)""")
         num_directions = 1
         bad_size = 7  # prime number so that no size can divide it.
 
-        def test(input_shape, hidden_h_shape, hidden_c_shape, device='cuda'):
-            for input, hidden in get_inputs(input_shape, hidden_h_shape, hidden_c_shape):
+        def test(input_shape, hidden_h_shape, hidden_c_shape, device):
+            for input, hidden in get_inputs(input_shape, hidden_h_shape, hidden_c_shape, device):
                 model = nn.LSTM(input_size, hidden_size, num_layers, proj_size=proj_size, device=device)
                 self.assertRaises(RuntimeError, lambda: model(input, hidden))
 
@@ -3943,12 +3940,12 @@ tensor(..., device='meta', size=(1,), requires_grad=True)""")
             new_shape[dim] = new_dim_size
             return tuple(new_shape)
 
-        def get_inputs(input_shape, hidden_h_shape, hidden_c_shape):
+        def get_inputs(input_shape, hidden_h_shape, hidden_c_shape, device):
             '''returns list( tuple(input, hidden) )
             where input, hidden are inputs to a model'''
-            input = torch.randn(input_shape)
-            hidden_h = torch.randn(hidden_h_shape)
-            hidden_c = torch.randn(hidden_c_shape)
+            input = torch.randn(input_shape, device=device)
+            hidden_h = torch.randn(hidden_h_shape, device=device)
+            hidden_c = torch.randn(hidden_c_shape, device=device)
             return [(input, (hidden_h, hidden_c))]
 
         # Incorrect input batch size
@@ -4449,16 +4446,14 @@ tensor(..., device='meta', size=(1,), requires_grad=True)""")
 
     def test_RNN_input_size_zero(self, device):
         for module in (nn.RNN, nn.LSTM, nn.GRU):
-            for device in get_all_device_types():
-                input = torch.zeros((5, 0, 3))
-                rnn = module(input_size=3, hidden_size=4)
-                if device in ['cuda', 'xpu', 'mps']:
-                    rnn.to(device)
-                    input = input.to(device)
-                outs = rnn(input)
-                self.assertEqual(outs[0].shape, torch.Size([5, 0, 4]))
-                # Check that backward does not cause a hard error
-                outs[0].sum().backward()
+            input = torch.zeros((5, 0, 3))
+            rnn = module(input_size=3, hidden_size=4)
+            rnn.to(device)
+            input = input.to(device)
+            outs = rnn(input)
+            self.assertEqual(outs[0].shape, torch.Size([5, 0, 4]))
+            # Check that backward does not cause a hard error
+            outs[0].sum().backward()
 
     @unittest.skipIf(not TEST_CUDNN, "needs cudnn")
     def test_RNN_dropout_state(self):
@@ -4717,14 +4712,11 @@ tensor(..., device='meta', size=(1,), requires_grad=True)""")
         gradgradcheck(func, [v])
 
     def test_PReLU_backward_requires_grad_false(self, device):
-        devices = ['cpu']
-        devices += [device] if HAS_ACCELERATOR else []
-        for d in devices:
-            m = nn.PReLU().to(d)
-            x = torch.randn(2, 3, 4, 5, device=d, requires_grad=False)
-            y = m(x)
-            y.mean().backward()
-            self.assertEqual(x.grad, None)
+        m = nn.PReLU().to(device)
+        x = torch.randn(2, 3, 4, 5, device=device, requires_grad=False)
+        y = m(x)
+        y.mean().backward()
+        self.assertEqual(x.grad, None)
 
     def test_bce_loss_always_nonnegative(self):
         target = torch.ones(5)
@@ -5072,7 +5064,7 @@ tensor(..., device='meta', size=(1,), requires_grad=True)""")
         helper(self, torch.bfloat16)
         helper(self, torch.float16)
 
-    @onlyAccelerator
+    @unittest.skipIf(not TEST_CUDA, "CUDA unavailable")
     @unittest.skipIf(not TEST_CUDNN, "needs cudnn")
     def test_batchnorm_cudnn_nhwc(self):
         def run_test(input, grad_output):
@@ -5257,25 +5249,27 @@ tensor(..., device='meta', size=(1,), requires_grad=True)""")
         name_fn=lambda f, b, m, t: f"{f}_vs_{b}{'_mixed' if m else ''}_{dtype_name(t)}"
     )
     def test_batchnorm(self, device, dims, mode, memory_format, ref_backend, mixed, dtype):
-        if HAS_ACCELERATOR:
-            _device = device.split(':')[0]
+        _dtype = dtype_name(dtype)
+        if torch.version.cuda or device == 'xpu':
             skip_tests = [
-                f"test_batchnorm_2D_train_NCHW_vs_cpu_mixed_bfloat16_{_device}_{dtype_name(dtype)}",
-                f"test_batchnorm_3D_train_NCHW_vs_cpu_mixed_bfloat16_{_device}_{dtype_name(dtype)}",
-                f"test_batchnorm_2D_train_NHWC_vs_NCHW_mixed_bfloat16_{_device}_{dtype_name(dtype)}",
-                f"test_batchnorm_3D_train_NHWC_vs_NCHW_mixed_bfloat16_{_device}_{dtype_name(dtype)}"]
-            skip_tests = skip_tests + (["test_batchnorm_3D_train_NCHW_vs_native_mixed_float16"] if TEST_CUDA else [])
+                f"test_batchnorm_2D_train_NCHW_vs_cpu_mixed_bfloat16_{device}_{_dtype}",
+                f"test_batchnorm_3D_train_NCHW_vs_cpu_mixed_bfloat16_{device}_{_dtype}",
+                f"test_batchnorm_2D_train_NHWC_vs_NCHW_mixed_bfloat16_{device}_{_dtype}",
+                f"test_batchnorm_3D_train_NHWC_vs_NCHW_mixed_bfloat16_{device}_{_dtype}"]
+            skip_tests = skip_tests + (
+                [f"test_batchnorm_3D_train_NCHW_vs_native_mixed_float16_{device}_{_dtype}"] if torch.version.cuda else []
+            )
             if self._testMethodName in skip_tests:
                 self.skipTest(f"Failed on {device.upper()}")
 
         if torch.version.hip:
-            if self._testMethodName in ("test_batchnorm_2D_train_NCHW_vs_native_mixed_bfloat16",
-                                        "test_batchnorm_3D_train_NCHW_vs_native_mixed_bfloat16") \
+            if self._testMethodName in (f"test_batchnorm_2D_train_NCHW_vs_native_mixed_bfloat16_{device}_{_dtype}",
+                                        f"test_batchnorm_3D_train_NCHW_vs_native_mixed_bfloat16_{device}_{_dtype}") \
                     and _get_torch_rocm_version() >= (6, 4):
                 # https://github.com/pytorch/pytorch/issues/156513
                 self.skipTest("bfloat16 NCHW train failed due to native tolerance issue")
 
-            if self._testMethodName == "test_batchnorm_3D_train_NCHW_vs_native_mixed_float16":
+            if self._testMethodName == f"test_batchnorm_3D_train_NCHW_vs_native_mixed_float16_{device}_{_dtype}":
                 self.skipTest("3D float16 NCHW train failed on ROCm")
 
         if dims == 3 and memory_format in ("NHWC", "NCHW"):
@@ -5438,30 +5432,27 @@ tensor(..., device='meta', size=(1,), requires_grad=True)""")
         self.assertTrue(gradcheck(lambda x, y: F.pairwise_distance(x, y), (input1, input2)))
 
     # TODO: Create an OpInfo for pdist
-    def test_pdist(self):
-        for device, trans in itertools.product(device_(), [False, True]):
+    def test_pdist(self, device):
+        for trans in [False, True]:
             inp = torch.randn(4, 5, dtype=torch.double, device=device, requires_grad=True)
             if trans:
                 inp = inp.transpose(0, 1)
             for p in [0, 1, 2, 0.5, 1.5, 2.5, float('inf')]:
                 self.assertTrue(gradcheck(lambda x: F.pdist(x, p), (inp,)))
 
-    def test_pdist_zeros(self):
+    def test_pdist_zeros(self, device):
         """Test that grad is still valid when dist is 0"""
-        for device in device_():
-            inp = torch.randn(1, 3, dtype=torch.double, device=device, requires_grad=True).repeat([2, 1])
-            for p in [0, 1, 2, 0.5, 1.5, 2.5, float('inf')]:
-                self.assertTrue(gradcheck(lambda x: F.pdist(x, p), (inp,)))
+        inp = torch.randn(1, 3, dtype=torch.double, device=device, requires_grad=True).repeat([2, 1])
+        for p in [0, 1, 2, 0.5, 1.5, 2.5, float('inf')]:
+            self.assertTrue(gradcheck(lambda x: F.pdist(x, p), (inp,)))
 
-    def test_pdist_empty_row(self):
-        for device in device_():
-            inp = torch.randn(1, 3, dtype=torch.double, device=device, requires_grad=True)
-            self.assertTrue(gradcheck(F.pdist, (inp,)))
+    def test_pdist_empty_row(self, device):
+        inp = torch.randn(1, 3, dtype=torch.double, device=device, requires_grad=True)
+        self.assertTrue(gradcheck(F.pdist, (inp,)))
 
-    def test_pdist_empty_col(self):
-        for device in device_():
-            inp = torch.randn(4, 0, dtype=torch.double, device=device, requires_grad=True)
-            self.assertTrue(gradcheck(F.pdist, (inp,)))
+    def test_pdist_empty_col(self, device):
+        inp = torch.randn(4, 0, dtype=torch.double, device=device, requires_grad=True)
+        self.assertTrue(gradcheck(F.pdist, (inp,)))
 
     @unittest.expectedFailure
     def test_pdist_cpu_gradgrad_unimplemented(self):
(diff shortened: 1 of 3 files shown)


test/test_nn.py
```diff
@@ -227,12 +231,13 @@ def test_to(self):
         self.assertEqual(m.double(), m.to(torch.float64))
         self.assertRaises(RuntimeError, lambda: m.to('cpu', copy=True))
 
-        if torch.cuda.is_available():
-            for cuda in ['cuda', 'cuda:0' if torch.cuda.device_count() == 1 else 'cuda:1']:
-                m2 = m.cuda(device=cuda)
-                self.assertIs(m2, m2.to(cuda))
+        if torch.cuda.is_available() or torch.xpu.is_available():
```
guangyey:
```suggestion
        if TEST_GPU :
```

test/test_nn.py
```diff
@@ -2784,15 +2797,15 @@ def test_CTCLoss_critical_target_len(self):
...
+        inp = torch.randn(T, N, C, dtype=torch.float, device=device_type).log_softmax(2).requires_grad_()
         with cudnn.flags(enabled=True):
             res_gpu = torch.nn.functional.ctc_loss(inp, target, input_lengths, target_lengths, reduction='none')
         res_cpu = torch.nn.functional.ctc_loss(inp.cpu(), target, input_lengths, target_lengths, reduction='none')
         self.assertEqual(res_cpu, res_gpu, atol=1e-3, rtol=0)
 
     def test_CTCLoss_zero_lengths(self):
         devices = ['cpu']
-        devices += ['cuda'] if TEST_CUDA else []
+        devices += [device_type] if TEST_CUDA or TEST_XPU else []
```
guangyey:
```suggestion
        devices += [device_type] if TEST_GPU else []
```

test/test_nn.py
```diff
@@ -2814,16 +2827,16 @@ def test_CTCLoss_zero_lengths(self):
             res.sum().backward()
             self.assertTrue((inp.grad == 0).all().item())
 
-    @unittest.skipIf(not TEST_CUDA, 'CUDA not available')
+    @unittest.skipIf(not TEST_CUDA and not TEST_XPU, 'CUDA and XPU not available')
```
guangyey:
```suggestion
    @unittest.skipIf(not TEST_GPU, 'CUDA and XPU not available')
```

test/test_nn.py
```diff
@@ -2894,12 +2907,12 @@ def test_LSTM_cell_forward_hidden_size(self):
         self.assertRaises(Exception, lambda: lstm(input, (cx, hx)))
 
 
-    @unittest.skipIf(not TEST_CUDA, 'CUDA not available')
+    @unittest.skipIf(not TEST_CUDA and not TEST_XPU, 'CUDA and XPU not available')
```
guangyey:
```suggestion
    @unittest.skipIf(not TEST_GPU, 'CUDA and XPU not available')
```

test/test_nn.py
```diff
@@ -5008,17 +5021,17 @@ def helper(self, dtype):
         helper(self, torch.bfloat16)
         helper(self, torch.float16)
 
-    @unittest.skipIf(not TEST_CUDA, "CUDA unavailable")
-    @unittest.skipIf(not TEST_CUDNN, "needs cudnn")
+    @unittest.skipIf(not TEST_CUDA and not TEST_XPU, "CUDA and XPU unavailable")
```
guangyey:
```suggestion
    @unittest.skipIf(not TEST_GPU, "CUDA and XPU unavailable")
```

test/test_nn.py
```diff
@@ -5008,17 +5021,17 @@ def helper(self, dtype):
...
+    @unittest.skipIf(not TEST_CUDNN and not TEST_XPU, "needs cudnn or xpu")
```
guangyey:
```suggestion
    @unittest.skipIf(not TEST_CUDNN, "needs cudnn")
```

test/test_nn.py
```diff
@@ -5155,7 +5168,7 @@ def test_batchnorm_buffer_update_when_stats_are_not_tracked(self):
         self.assertTrue(torch.equal(running_var, bn.running_var))
 
 
-    @unittest.skipIf(not torch.cuda.is_available(), "CUDA not available")
+    @unittest.skipIf(not torch.cuda.is_available() and not torch.xpu.is_available(), "CUDA and XPU not available")
```
guangyey:
```suggestion
    @unittest.skipIf(not TEST_GPU, "CUDA and XPU not available")
```

test/test_nn.py
```diff
@@ -5193,13 +5206,13 @@ def test_batchnorm_buffer_update_when_stats_are_not_tracked(self):
         name_fn=lambda f, b, m, t: f"{f}_vs_{b}{'_mixed' if m else ''}_{dtype_name(t)}"
     )
     def test_batchnorm(self, dims, mode, memory_format, ref_backend, mixed, dtype):
-        if torch.version.cuda:
+        if torch.version.cuda or torch.version.xpu:
```
guangyey: Originally, it means CUDA skips these cases, but ROCM doesn't. I think we should run these cases and see what will happen.

test/test_nn.py
```diff
@@ -5336,13 +5349,13 @@ def _inference(memory_format_name, ref_backend, mixed, dtype):
         else:
             _inference(memory_format, ref_backend, mixed, dtype)
 
-    @unittest.skipIf(not torch.cuda.is_available(), "CUDA not available")
+    @unittest.skipIf(not torch.cuda.is_available() and not torch.xpu.is_available(), "CUDA and XPU not available")
```
guangyey:
```suggestion
    @unittest.skipIf(not TEST_GPU, "CUDA and XPU not available")
```

test/test_nn.py
daisyden: ok, let me try.

test/test_nn.py
guangyey: Here don't need `TEST_XPU`, right

daisyden: The case is general and can work on XPU so I didn't skip it.

test/test_nn.py
```diff
@@ -2814,16 +2827,16 @@ def test_CTCLoss_zero_lengths(self):
             res.sum().backward()
             self.assertTrue((inp.grad == 0).all().item())
 
-    @unittest.skipIf(not TEST_CUDA, 'CUDA not available')
+    @unittest.skipIf(not TEST_GPU, 'CUDA and XPU not available')
```
guangyey:
```suggestion
    @unittest.skipIf(not TEST_GPU, 'The current accelerator is not available')
```

test/test_nn.py
```diff
@@ -5199,7 +5212,14 @@ def test_batchnorm(self, dims, mode, memory_format, ref_backend, mixed, dtype):
                                         "test_batchnorm_2D_train_NHWC_vs_NCHW_mixed_bfloat16",
                                         "test_batchnorm_3D_train_NHWC_vs_NCHW_mixed_bfloat16",
                                         "test_batchnorm_3D_train_NCHW_vs_native_mixed_float16"):
-                self.skipTest("Failed on CUDA")
+                self.skipTest(f"Failed on {device_type.upper()}")
+
+        if torch.version.xpu:
```
guangyey:
```suggestion
        if TEST_XPU:
```

guangyey (dismissed):
lgtm.

test/test_nn.py
```diff
@@ -4388,9 +4401,9 @@ def test_RNN_input_size_zero(self):
             for device in get_all_device_types():
                 input = torch.zeros((5, 0, 3))
                 rnn = module(input_size=3, hidden_size=4)
-                if device == 'cuda':
-                    rnn.cuda()
-                    input = input.cuda()
+                if device == 'cuda' or device == 'xpu':
```
malfet:
```suggestion
                if device in ['cuda', 'xpu']:
```

test/test_nn.py
```diff
@@ -75,14 +75,18 @@
 if TEST_NUMPY:
     import numpy as np
 
+TEST_GPU = torch.cuda.is_available() or torch.xpu.is_available()
+TEST_MULTIGPU = TEST_GPU and torch.accelerator.device_count() >= 2
```
malfet: Why not change the definition in `common_cuda.py`?

malfet (changes requested):
Is it possible to split this change into few smaller ones? For example, change to `common_device_types` and `test_namedtensor.py` looks totally fine and majority of the changes to `test_nn` are also OK, but multigpu changes seems a bit risky

Perhaps let's do single GPU enablement in 1 PR and than multi-GPU in another

test/test_nn.py
daisyden: Can I add TEST_GPU and TEST_MULTIGPU definition in common_utils.py?

test/test_nn.py
daisyden: Thanks for the suggestions! I will split the PR into two, one for single GPU and one for multi-gpu.

test/test_nn.py
daisyden: Hi, @malfet, I removed the code updates for multi-gpu from this PR and will create another PR for them. I also moved the TEST_GPU definition to common_utils. Please further review this PR.  Thanks.

daisyden: Hi, @malfet, TEST_GPU is not a general definition. After consideration I feel it is not appropriate to put under common_utils or common_cuda. I define it as
``` 
TEST_GPU = TEST_CUDA or TEST_XPU
```
because other accelerators have limitation on test_nn, so I only port those cases for XPU.

test/test_nn.py
daisyden: Done, this PR is now only for single gpu.

test/test_nn.py
```diff
@@ -75,14 +75,17 @@
...
 # WARNING: If you add a new top-level test case to this file, you MUST
 # update test/run_test.py to list it, otherwise it will NOT be run in
 # CI.
 
 class TestNN(NNTestCase):
-    _do_cuda_memory_leak_check = True
-    _do_cuda_non_default_stream = True
+    if TEST_CUDA:
+        _do_cuda_memory_leak_check = True
+        _do_cuda_non_default_stream = True
```
EikanWang:
```python
    _do_cuda_memory_leak_check = TEST_CUDA
    _do_cuda_non_default_stream = TEST_CUDA
```

test/test_nn.py
```diff
@@ -1923,6 +1927,7 @@ def test_weight_norm_pickle(self):
         m = pickle.loads(pickle.dumps(m))
         self.assertIsInstance(m, nn.Linear)
 
+    @skipIfXpu  # https://github.com/intel/torch-xpu-ops/issues/2228
```
EikanWang: `torch._C._scatter` is dedicated to `cuda`. For XPU, it is not a priority unless users specifically request it for XPU.

```suggestion
    @skipIfXpu
```

test/test_nn.py
```diff
@@ -2290,34 +2295,37 @@ def test_normalize(self):
         inputs = torch.randn((), requires_grad=True)
         self.assertTrue(gradcheck(lambda x: F.normalize(x, p=1, dim=-1), (inputs,)))
 
+    @skipIfXpu  # https://github.com/intel/torch-xpu-ops/issues/2228
```
EikanWang: ditto

test/test_nn.py
```diff
@@ -2290,34 +2295,37 @@ def test_normalize(self):
...
-                   torch.randn(4, 4, device='cuda', requires_grad=True, dtype=torch.double),
-                   torch.randn(4, 4, device='cuda', requires_grad=True, dtype=torch.double))
+        tensors = (torch.randn(4, 4, device=device_type, requires_grad=True, dtype=torch.double),
+                   torch.randn(4, 4, device=device_type, requires_grad=True, dtype=torch.double),
+                   torch.randn(4, 4, device=device_type, requires_grad=True, dtype=torch.double))
         # TODO(#50743): the following segfaults with check_batched_grad=True
         _assertGradAndGradgradChecks(self, lambda *i: Broadcast.apply((0, 1), *i), tensors,
                                      check_batched_grad=False)
 
+    @skipIfXpu  # https://github.com/intel/torch-xpu-ops/issues/2228
```
EikanWang: ditto

test/test_nn.py
```diff
@@ -2290,34 +2295,37 @@ def test_normalize(self):
...
+            torch.randn(1, 2, device=device_type, requires_grad=False),
+            torch.randn(1, 2, device=device_type, requires_grad=True),
+            torch.randn(1, 2, device=device_type, requires_grad=True),
         ]
         broadcasted_variables = Broadcast.apply((0, 1), *variables)
         for output_idx, broadcasted_var in enumerate(broadcasted_variables):
             input_var = variables[output_idx % len(variables)]
             self.assertEqual(input_var.requires_grad, broadcasted_var.requires_grad)
 
+    @skipIfXpu  # https://github.com/intel/torch-xpu-ops/issues/2228
```
EikanWang: ditto

test/test_nn.py
```diff
@@ -2768,13 +2776,17 @@ def test_CTCLoss_long_targets(self):
...
+                                                   input_lengths,
+                                                   target_lengths,
+                                                   reduction='sum',
+                                                   zero_infinity=True)
+            grad_gpu, = torch.autograd.grad(res_gpu, log_probs, grad_out.to(device_type))
         self.assertEqual(res_cpu, res_gpu, atol=1e-4, rtol=0)
         self.assertEqual(grad_cpu, grad_gpu, atol=1e-4, rtol=0)
 
-    @unittest.skipIf(not TEST_CUDA, 'CUDA not available')
+    @unittest.skipIf(not TEST_GPU, 'The current accelerator not available')
```
EikanWang:
```suggestion
    @unittest.skipIf(not TEST_GPU, f"{device_type } not available")
```

test/test_nn.py
```diff
@@ -2814,16 +2826,16 @@ def test_CTCLoss_zero_lengths(self):
             res.sum().backward()
             self.assertTrue((inp.grad == 0).all().item())
 
-    @unittest.skipIf(not TEST_CUDA, 'CUDA not available')
+    @unittest.skipIf(not TEST_GPU, 'The current accelerator not available')
```
EikanWang: ditto

test/test_nn.py
```diff
@@ -2894,12 +2906,12 @@ def test_LSTM_cell_forward_hidden_size(self):
         self.assertRaises(Exception, lambda: lstm(input, (cx, hx)))
 
 
-    @unittest.skipIf(not TEST_CUDA, 'CUDA not available')
+    @unittest.skipIf(not TEST_GPU, 'The current accelerator not available')
```
EikanWang: ditto

test/test_nn.py
```diff
@@ -3173,7 +3185,7 @@ def get_a_test_layer(use_cuda, activation, batch_first=False):
             nhead = 2
             dim_feedforward = 16
             dropout = 0.0
-            device = torch.device("cuda" if use_cuda else "cpu")
+            device = torch.device(device_type if use_cuda else "cpu")
```
EikanWang:
```suggestion
            device = torch.device(device_type)
```

test/test_nn.py
```diff
@@ -3199,8 +3211,8 @@ def get_a_test_layer(use_cuda, activation, batch_first=False):
             def perm_fn(x):
                 return x.transpose(1, 0) if batch_first else x
             activation = F.relu
-            use_cuda = torch.cuda.is_available()
-            device = torch.device("cuda" if use_cuda else "cpu")
+            use_cuda = TEST_GPU
```
EikanWang:
```suggestion
```

test/test_nn.py
```diff
@@ -3199,8 +3211,8 @@ def get_a_test_layer(use_cuda, activation, batch_first=False):
...
+            device = torch.device(device_type if use_cuda else "cpu")
```
EikanWang:
```suggestion
            device = torch.device(device_type)
```

test/test_nn.py
```diff
@@ -3199,8 +3211,8 @@ def get_a_test_layer(use_cuda, activation, batch_first=False):
...
 
             decoder_layer = get_a_test_layer(use_cuda=use_cuda, activation=activation,
```
EikanWang:
```suggestion
            decoder_layer = get_a_test_layer(activation=activation,
```

test/test_nn.py
EikanWang:
```suggestion
        def get_a_test_layer(activation, batch_first=False):
```

test/test_nn.py
```diff
@@ -3424,8 +3436,8 @@ def perm_fn(x):
 
             # gelu activation test cases
             activation = "gelu"
-            use_cuda = torch.cuda.is_available()
-            device = torch.device("cuda" if use_cuda else "cpu")
+            use_cuda = TEST_GPU
+            device = torch.device(device_type if use_cuda else "cpu")
 
             decoder_layer = get_a_test_layer(use_cuda=use_cuda, activation=activation,
```
EikanWang:
```suggestion
            decoder_layer = get_a_test_layer(activation=activation,
```

test/test_nn.py
```diff
@@ -3424,8 +3436,8 @@ def perm_fn(x):
 
             # gelu activation test cases
             activation = "gelu"
-            use_cuda = torch.cuda.is_available()
-            device = torch.device("cuda" if use_cuda else "cpu")
+            use_cuda = TEST_GPU
```
EikanWang:
```suggestion
```

test/test_nn.py
```diff
@@ -3424,8 +3436,8 @@ def perm_fn(x):
...
+            device = torch.device(device_type if use_cuda else "cpu")
```
EikanWang:
```suggestion
            device = torch.device(device_type)
```

test/test_nn.py
```diff
@@ -4077,17 +4089,17 @@ def forward_backward(cuda, rnn, input_val, grad_output, weights_val, hx_val, gra
                 hx = hx_val.clone().requires_grad_(True)
 
             if cuda:
```
EikanWang, line 4136: Please refine the function parameter as well.

```suggestion
            if accelerator:
```

test/test_nn.py
```diff
@@ -4286,8 +4298,8 @@ def check_weight_norm(m, name):
             m = torch.nn.utils.weight_norm(m, name=name)
 
             # moves to CUDA
```
EikanWang, line 4345:
```suggestion
            # moves to the accelerator
```

test/test_nn.py
```diff
@@ -4315,10 +4327,10 @@ def test_partial_flat_weights(self):
         self.assertFalse(hasattr(m, "weight_hh_l0"))
         # verifies that moving to CUDA with only some attributes defined
```
EikanWang:
```suggestion
        # verifies that moving to Accelerator with only some attributes defined
```

test/test_nn.py
```diff
@@ -5336,13 +5355,13 @@ def _inference(memory_format_name, ref_backend, mixed, dtype):
         else:
             _inference(memory_format, ref_backend, mixed, dtype)
 
-    @unittest.skipIf(not torch.cuda.is_available(), "CUDA not available")
+    @unittest.skipIf(not TEST_GPU, "The current accelerator not available")
```
EikanWang:
```suggestion
    @unittest.skipIf(not TEST_GPU, f"{device_type} not available")
```

test/test_nn.py
```diff
@@ -5199,7 +5211,14 @@ def test_batchnorm(self, dims, mode, memory_format, ref_backend, mixed, dtype):
...
                                         "test_batchnorm_3D_train_NHWC_vs_NCHW_mixed_bfloat16",
                                         "test_batchnorm_3D_train_NCHW_vs_native_mixed_float16"):
-                self.skipTest("Failed on CUDA")
+                self.skipTest(f"Failed on {device_type.upper()}")
+
+        if TEST_XPU:
+            if self._testMethodName in ("test_batchnorm_2D_train_NCHW_vs_cpu_mixed_bfloat16",
+                                        "test_batchnorm_3D_train_NCHW_vs_cpu_mixed_bfloat16",
+                                        "test_batchnorm_2D_train_NHWC_vs_NCHW_mixed_bfloat16",
+                                        "test_batchnorm_3D_train_NHWC_vs_NCHW_mixed_bfloat16"):
```
EikanWang: @daisyden , `test_batchnorm_3D_train_NCHW_vs_native_mixed_float16` can pass on XPU, right?

EikanWang:
```python
if TEST_GPU:
    skip_tests = [
        "test_batchnorm_2D_train_NCHW_vs_cpu_mixed_bfloat16",
        "test_batchnorm_3D_train_NCHW_vs_cpu_mixed_bfloat16",
        "test_batchnorm_2D_train_NHWC_vs_NCHW_mixed_bfloat16",
        "test_batchnorm_3D_train_NHWC_vs_NCHW_mixed_bfloat16"]
    skip_tests = skip_tests + (["test_batchnorm_3D_train_NCHW_vs_native_mixed_float16"] if TEST_CUDA  else [])
    if self._testMethodName in skip_tests:
        self.skipTest(f"Failed on {device_type.upper()}")
```

test/test_nn.py
```diff
@@ -5031,25 +5043,25 @@ def run_test(input, grad_output):
...
-        input = torch.randint(1, 10, (2, 8, 8, 1), dtype=torch.float32, device="cuda")
+        input = torch.randint(1, 10, (2, 8, 8, 1), dtype=torch.float32, device=device_type)
         input = input.contiguous(memory_format=torch.channels_last).detach().requires_grad_()
-        grad = torch.randint(1, 10, (2, 8, 8, 1), dtype=torch.float32, device="cuda")
+        grad = torch.randint(1, 10, (2, 8, 8, 1), dtype=torch.float32, device=device_type)
         grad = grad.permute(0, 2, 1, 3)
         run_test(input, grad)
 
-    @unittest.skipIf(not TEST_CUDA, "CUDA unavailable")
+    @unittest.skipIf(not TEST_GPU, "The current accelerator unavailable")
```
EikanWang:
```suggestion
    @unittest.skipIf(not TEST_GPU, f"{device_type} unavailable")
```

test/test_nn.py
```diff
@@ -5065,10 +5077,10 @@ def test_batchnorm_cudnn_half(self):
             self.assertEqual(cudnn_output, thnn_output)
             self.assertEqual(cudnn_input_grad, thnn_input_grad, atol=1e-3, rtol=0)
 
-    @unittest.skipIf(not TEST_CUDA, "CUDA unavailable")
+    @unittest.skipIf(not TEST_GPU, "The current accelerator unavailable")
```
EikanWang: ditto

test/test_nn.py
```diff
@@ -5155,7 +5167,7 @@ def test_batchnorm_buffer_update_when_stats_are_not_tracked(self):
         self.assertTrue(torch.equal(running_var, bn.running_var))
 
 
-    @unittest.skipIf(not torch.cuda.is_available(), "CUDA not available")
+    @unittest.skipIf(not TEST_GPU, "The current accelerator not available")
```
EikanWang: ditto

test/test_nn.py
```diff
@@ -4444,8 +4456,8 @@ def test_RNN_change_dropout(self):
             rnn = nn.RNN(100, 100, 2, dropout=0, nonlinearity='relu')
             input = torch.rand(3, 2, 100)
             if cuda:
```
EikanWang, line 4505: Please refine the variable name.

test/test_nn.py
```diff
@@ -4376,9 +4388,9 @@ def test_error_RNN_seq_len_zero(self):
                 for device in get_all_device_types():
                     input = torch.ones(0, 10, 5)
                     rnn = module(5, 6, bidirectional=bidirectional)
-                    if device == 'cuda':
-                        rnn.cuda()
-                        input = input.cuda()
+                    if device == device_type:
```
EikanWang: `device_type` could be `cpu`. I suppose this check intends to gate `cuda` and `xpu` only.

test/test_nn.py
```diff
@@ -4333,7 +4345,7 @@ def test_RNN_dropout(self):
                 for cuda in (True, False):
                     rnn = nn.RNN(10, 1000, 2, bias=False, dropout=p, nonlinearity='relu')
                     if cuda:
```
EikanWang, line 4396: Pls. refine the variable name.

EikanWang (changes requested):

test/test_nn.py
daisyden: Yes, Intel GPU can pass with test_batchnorm_3D_train_NCHW_vs_native_mixed_float16, will refine the code.

EikanWang (approved):

test/test_nn.py
```diff
@@ -2318,30 +2323,32 @@ def forward(self, x):
...
+        tensors = (torch.randn(4, 4, device=device_type, requires_grad=True, dtype=torch.double),
+                   torch.randn(4, 4, device=device_type, requires_grad=True, dtype=torch.double),
+                   torch.randn(4, 4, device=device_type, requires_grad=True, dtype=torch.double))
         # TODO(#50743): the following segfaults with check_batched_grad=True
         _assertGradAndGradgradChecks(self, lambda *i: Broadcast.apply((0, 1), *i), tensors,
                                      check_batched_grad=False)
 
+    @skipIfXpu
     @unittest.skipIf(not TEST_MULTIGPU, "multi-GPU not supported")
     def test_broadcast_not_requiring_grad(self):
```
malfet: Why not migrate this test to standard device instantiation?

test/test_nn.py
```diff
@@ -75,14 +75,16 @@
 if TEST_NUMPY:
     import numpy as np
 
+TEST_GPU = TEST_CUDA or TEST_XPU
```
malfet: Why not introduce `HAS_GPU` (or `HAS_ACCELERATOR`) prefix somewhere in torch.testing._internal.common_device_types and use it throughout the codebase?

malfet (dismissed):
This PR achieves the goal of enabling tests for XPU, but imo it's goes contrary to the goal of generalizing the tests/using torch.accelerator API. I.e. why those tests should not be running on some PrivateUse1 devices or say MPS?


I.e. perhaps it could be split into two PRs:
 - one that adds `device` argument to the tests and decorates CUDA only tests with `@onlyAccelerator` and 2nd one that instantiates subset of tests to GPU

test/test_nn.py
daisyden: @malfet  Thanks for the suggestion, I updated this PR to instantiate TestNN with instantiate_device_type_tests(). Please help review the code.  I will enable XPU test in another PR when the first PR is completed.

test/test_nn.py
daisyden: _test_RNN_cpu_vs_cudnn is cudnn test, we could not change it.

test/test_nn.py
daisyden: cudnn test only, so keep the code unchanged.

test/test_nn.py
```diff
@@ -224,19 +224,20 @@ def forward(self, inp):
...
         self.assertIs(m, m.to('cpu'))
         self.assertIs(m, m.to('cpu', dtype=torch.float32))
         self.assertEqual(m.double(), m.to(torch.float64))
         self.assertRaises(RuntimeError, lambda: m.to('cpu', copy=True))
 
-        if torch.cuda.is_available():
-            for cuda in ['cuda', 'cuda:0' if torch.cuda.device_count() == 1 else 'cuda:1']:
-                m2 = m.cuda(device=cuda)
-                self.assertIs(m2, m2.to(cuda))
+        if HAS_ACCELERATOR:
```
mikaylagawarecki: It could be the case that `HAS_ACCELERATOR`  is true even if the `device` argument passed is `cpu`. This would mean that the cpu version of `test_to` would still enter the GPU.

In PyTorch CI, I don't believe this happens because CPU tests would run on CPU-only machines and GPU tests would run on GPU-only machines. However, we can't guarantee that would be the case in other CI setups that run our test suite

Perhaps we could do something like the following instead which should achieve the same purpose and avoid the potential issue -- 

```suggestion
        acc = torch.accelerator.current_accelerator()
        if acc is not None and torch.device(device).type == acc.type: 
```

what do you think?

test/test_nn.py
```diff
@@ -2815,15 +2821,15 @@ def test_CTCLoss_critical_target_len(self):
...
         with cudnn.flags(enabled=True):
             res_gpu = torch.nn.functional.ctc_loss(inp, target, input_lengths, target_lengths, reduction='none')
         res_cpu = torch.nn.functional.ctc_loss(inp.cpu(), target, input_lengths, target_lengths, reduction='none')
         self.assertEqual(res_cpu, res_gpu, atol=1e-3, rtol=0)
 
-    def test_CTCLoss_zero_lengths(self):
+    def test_CTCLoss_zero_lengths(self, device):
         devices = ['cpu']
-        devices += ['cuda'] if TEST_CUDA else []
+        devices += [device] if HAS_ACCELERATOR else []
```
mikaylagawarecki: now that we have migrated this test to use `instantiate_device_type_tests`, we no longer need a list of `devices` to loop over right?

test/test_nn.py
```diff
@@ -4418,14 +4421,14 @@ def test_error_RNN_seq_len_zero(self):
                     with self.assertRaisesRegex(RuntimeError, "Expected sequence length to be larger than 0 in RNN"):
                         rnn(input)
 
-    def test_RNN_input_size_zero(self):
+    def test_RNN_input_size_zero(self, device):
         for module in (nn.RNN, nn.LSTM, nn.GRU):
             for device in get_all_device_types():
```
mikaylagawarecki: the test no longer needs to iterate over `get_all_device_types` when it has been instantiated with instantiate_device_type_tests

test/test_nn.py
```diff
@@ -4418,14 +4421,14 @@ def test_error_RNN_seq_len_zero(self):
...
                 input = torch.zeros((5, 0, 3))
                 rnn = module(input_size=3, hidden_size=4)
-                if device == 'cuda':
-                    rnn.cuda()
-                    input = input.cuda()
+                if device in ['cuda', 'xpu', 'mps']:
+                    rnn.to(device)
+                    input = input.to(device)
```
mikaylagawarecki:
```suggestion
                rnn.to(device)
                input = input.to(device)
```

test/test_nn.py
```diff
@@ -2815,15 +2821,15 @@ def test_CTCLoss_critical_target_len(self):
...
-        inp = torch.randn(T, N, C, dtype=torch.float, device='cuda').log_softmax(2).requires_grad_()
+        inp = torch.randn(T, N, C, dtype=torch.float, device=device).log_softmax(2).requires_grad_()
         with cudnn.flags(enabled=True):
             res_gpu = torch.nn.functional.ctc_loss(inp, target, input_lengths, target_lengths, reduction='none')
         res_cpu = torch.nn.functional.ctc_loss(inp.cpu(), target, input_lengths, target_lengths, reduction='none')
         self.assertEqual(res_cpu, res_gpu, atol=1e-3, rtol=0)
 
-    def test_CTCLoss_zero_lengths(self):
+    def test_CTCLoss_zero_lengths(self, device):
         devices = ['cpu']
```
mikaylagawarecki: you can delete `devices`, there's no need to loop over all devices if we already use `instantiate_device_type_tests`

test/test_nn.py
mikaylagawarecki:
```suggestion
x1 = torch.tensor(-1.6437e+10, dtype=torch.float64, device=device)
x2 = torch.full((3, 8, 2, 6), float('nan'), dtype=torch.float16, device=device)
```

test/test_nn.py
```diff
@@ -5756,9 +5762,9 @@ def test_cosine_similarity_mixed_precision(self):
         with self.assertRaisesRegex(RuntimeError, "eps must be non-negative"):
             F.cosine_similarity(x1, x2, dim=0, eps=eps_negative)
 
-        if torch.cuda.is_available():
-            x1_cuda = x1.cuda()
-            x2_cuda = x2.cuda()
+        if HAS_ACCELERATOR:
```
mikaylagawarecki: As long as we create `x1` and `x2` on device this becomes code duplication. Everything in this if block can be deleted now that we use `instantiate_device_type_tests`.

To be clear, `instantiate_device_type_tests` will run the test with the `device` argument given to it.

test/test_nn.py
```diff
@@ -6418,13 +6424,13 @@ def test_shape(N, C, ID, IH, IW, D, H, W, mode, padding_mode, align_corners):
...
-                        test(N, C, D, H, W, mode, padding_mode, align_corners, input_requires_grad)
+                        test(N, C, D, H, W, mode, padding_mode, align_corners, input_requires_grad, device=device)
 
-    def test_grid_sample_nearest_neighbor_rounding_mode_consistency(self):
+    def test_grid_sample_nearest_neighbor_rounding_mode_consistency(self, device):
 
         device_list = ['cpu']
-        if TEST_CUDA:
-            device_list.append('cuda')
+        if HAS_ACCELERATOR:
```
mikaylagawarecki: similar comment here, `device_list` and any inner loop over `device_list` can be deleted

test/test_nn.py
```diff
@@ -6748,7 +6754,7 @@ def test_upsamplingLinear1d_spatial_invariance(self):
         self.assertEqual(out_t_9[:, :, :15], out_t_5)
 
     @set_default_dtype(torch.double)
-    def test_upsampling_not_recompute_scale_factor(self):
+    def test_upsampling_not_recompute_scale_factor(self, device):
```
mikaylagawarecki, line 6746: remove the internal loop over device_list and instantiate tensors within the function on device as appropriate

test/test_nn.py
```diff
@@ -6983,8 +6989,8 @@ def _make_input(dim, device):
             return torch.ones(size, requires_grad=True, device=device)
 
         device_list = ['cpu']
-        if TEST_CUDA:
```
mikaylagawarecki, line 7014: similar comment, remove device_list and the loop over it

test/test_nn.py
mikaylagawarecki: instantiate these on device and delete everything in the `if TEST_CUDA` block

test/test_nn.py
```diff
@@ -3601,7 +3605,7 @@ def test_cudnn_weight_format(self):
 
     @unittest.skipIf(not TEST_CUDNN, 'CUDNN not available')
     @tf32_on_and_off
-    def test_cudnn_weight_tying(self):
+    def test_cudnn_weight_tying(self, device):
```
mikaylagawarecki: As a separate PR stacked on top of this one, I think we should move all the cudnn tests to a separate test class that only runs on CUDA

test/test_nn.py
```diff
@@ -5929,11 +5934,11 @@ def test_affine_grid_backward_cl_cf_consistency(self, device, nd):
         self.assertEqual(theta_grad_cf, theta_grad_cl)
 
     @set_default_dtype(torch.double)
-    def test_grid_sample(self):
+    def test_grid_sample(self, device):
         # Backward pass of native C++ and CUDA kernels branch depending on whether input requires gradient,
         # so we test both cases.
-        def test(N, C, H, W, mode, padding_mode, align_corners, input_requires_grad):
-            def test_shape(N, C, IH, IW, H, W, mode, padding_mode, align_corners):
+        def test(N, C, H, W, mode, padding_mode, align_corners, input_requires_grad, device='cuda'):
```
mikaylagawarecki: nit: perhaps no need for default here 
```suggestion
        def test(N, C, H, W, mode, padding_mode, align_corners, input_requires_grad, device):
```

torch/testing/_internal/common_device_type.py
```diff
@@ -1682,6 +1682,10 @@ def onlyHPU(fn):
     return onlyOn("hpu")(fn)
 
 
+def onlyAccelerator(fn):
```
mikaylagawarecki, line 1685: I much prefer this to onlyOn(['xpu', 'cuda']) in particular I like that this will help us distinguish tests that should run on all accelerators explicitly :)

We can update this to include PrivateUse1 in the future

test/test_nn.py
```diff
@@ -14231,7 +14234,8 @@ def __init__(self) -> None:
 
 
 instantiate_device_type_tests(TestNNDeviceType, globals(), allow_mps=True)
-instantiate_parametrized_tests(TestNN)
+instantiate_device_type_tests(TestNN, globals())
```
mikaylagawarecki: As a side effect, does this make all the tests that are not passed the `device` arg run the exact same code twice? Once for TestNNCPU and once for TestNN{Device}?

mikaylagawarecki (commented):
Thanks for your work on this PR! There are 3 high level comments that I have 

- Several of the tests in this PR have an additional internal loop that loops over a `device_list` and runs code on each device, despite having instantiated the tests for each device with `instantiate_device_type_tests`. The internal loop can be removed
- See [this comment](https://github.com/pytorch/pytorch/pull/166396/changes#r2891973959) re `HAS_ACCELERATOR` within the test body 
- Some of the tests marked as onlyAccelerator seem to be intended to be cudnn tests per their name.

mikaylagawarecki:
@claude This PR intends to make the `TestNN` class device-generic. Review the changes in this PR, paying particular attention to 1. whether any existing test functionality was changed 2. whether the contents of the modified tests still make sense given that the test class was migrated to use `instantiate_device_type_tests` 3. whether the tests marked onlyAccelerator should run on all accelerators

test/test_nn.py
daisyden: @mikaylagawarecki You are right, I will do an update.

test/test_nn.py
daisyden: Good suggestion.

test/test_nn.py
daisyden: Yes, if no device arg passed, the case will run twice, maybe we can create a new PR stacked on this one to move those case into a new test class.

daisyden:
> Thanks for your work on this PR! There are 3 high level comments that I have
> 
> * Several of the tests in this PR have an additional internal loop that loops over a `device_list` and runs code on each device, despite having instantiated the tests for each device with `instantiate_device_type_tests`. The internal loop can be removed
> * See [this comment](https://github.com/pytorch/pytorch/pull/166396/changes#r2891973959) re `HAS_ACCELERATOR` within the test body
> * Some of the tests marked as onlyAccelerator seem to be intended to be cudnn tests per their name.

@mikaylagawarecki  thanks for your valuable suggestions. Basically I agree with you on the three points and I will update my code.

test/test_nn.py
```diff
@@ -5227,14 +5226,17 @@ def test_batchnorm_buffer_update_when_stats_are_not_tracked(self):
...
-    def test_batchnorm(self, dims, mode, memory_format, ref_backend, mixed, dtype):
-        if torch.version.cuda:
-            if self._testMethodName in ("test_batchnorm_2D_train_NCHW_vs_cpu_mixed_bfloat16",
-                                        "test_batchnorm_3D_train_NCHW_vs_cpu_mixed_bfloat16",
-                                        "test_batchnorm_2D_train_NHWC_vs_NCHW_mixed_bfloat16",
-                                        "test_batchnorm_3D_train_NHWC_vs_NCHW_mixed_bfloat16",
-                                        "test_batchnorm_3D_train_NCHW_vs_native_mixed_float16"):
-                self.skipTest("Failed on CUDA")
+    def test_batchnorm(self, device, dims, mode, memory_format, ref_backend, mixed, dtype):
+        if ACCELERATOR_TYPE == device:
```
guangyey: This code change changes ROCM's behavior.

test/test_nn.py
daisyden: Cudnn test, we could not change it in this PR.

daisyden: CUDNN test, we will not update it.

test/test_nn.py
daisyden: cudann test, will not update it in the PR.

test/test_nn.py
daisyden: updated

test/test_nn.py
daisyden: I will simplify test_cosine_similarity_mixed_precision.

test/test_nn.py
```diff
@@ -5252,23 +5225,28 @@ def test_batchnorm_buffer_update_when_stats_are_not_tracked(self):
...
+    def test_batchnorm(self, device, dims, mode, memory_format, ref_backend, mixed, dtype):
+        _dtype = dtype_name(dtype)
+        if torch.version.cuda or device == 'xpu':
+            skip_tests = [
+                f"test_batchnorm_2D_train_NCHW_vs_cpu_mixed_bfloat16_{device}_{_dtype}",
+                f"test_batchnorm_3D_train_NCHW_vs_cpu_mixed_bfloat16_{device}_{_dtype}",
+                f"test_batchnorm_2D_train_NHWC_vs_NCHW_mixed_bfloat16_{device}_{_dtype}",
+                f"test_batchnorm_3D_train_NHWC_vs_NCHW_mixed_bfloat16_{device}_{_dtype}"]
+            skip_tests = skip_tests + (
+                ["test_batchnorm_3D_train_NCHW_vs_native_mixed_float16_{device}_{_dtype}"] if torch.version.cuda else []
```
mikaylagawarecki: claude's comment here is valid
```suggestion
                [f"test_batchnorm_3D_train_NCHW_vs_native_mixed_float16_{device}_{_dtype}"] if torch.version.cuda else []
```

test/test_nn.py
```diff
@@ -224,19 +224,20 @@ def forward(self, inp):
...
-                m2 = m.cuda(device=cuda)
-                self.assertIs(m2, m2.to(cuda))
+        if ACCELERATOR_TYPE == device:
+            for gpu in [device,
+                        str(torch.device(0)) if torch.accelerator.device_count() == 1 else str(torch.device(1))]:
+                m2 = m.to(device)
+                self.assertIs(m2, m2.to(device))
                 self.assertEqual(m, m2.to('cpu'))
-                self.assertEqual(m2, m.to(cuda))
+                self.assertEqual(m2, m.to(device))
```
mikaylagawarecki: claude's comment here is also valid

```suggestion
                m2 = m.to(gpu)
                self.assertIs(m2, m2.to(gpu))
self.assertEqual(m, m2.to('cpu'))
                self.assertEqual(m2, m.to(gpu))
```

test/test_nn.py
```diff
@@ -6844,7 +6792,7 @@ def test_upsampling_small_scale(self):
         expected_out_t = torch.tensor([[[[2.5]]]])
         self.assertEqual(expected_out_t, out_t)
 
-    def test_upsampling_bfloat16(self, dtype=torch.bfloat16):
+    def test_upsampling_bfloat16(self, device, dtype=torch.bfloat16):
```
mikaylagawarecki, line 6814: claude's comment here also seems to be valid (?) the loop in 6811 makes this test only test cpu

test/test_nn.py
```diff
@@ -5068,7 +5041,7 @@ def helper(self, dtype):
         helper(self, torch.bfloat16)
         helper(self, torch.float16)
 
-    @unittest.skipIf(not TEST_CUDA, "CUDA unavailable")
+    @onlyAccelerator
```
mikaylagawarecki: nit: I think no need to use `@onlyAccelerator` here since this is a cudnn test

But it's fine since the CUDNN skip will skip the test either way

test/test_nn.py
```diff
@@ -1757,7 +1758,7 @@ def test_type(self):
             self.assertIsInstance(l.weight.data, torch.cuda.DoubleTensor)
             self.assertIsInstance(l.bias.data, torch.cuda.DoubleTensor)
             self.assertIsInstance(net.indices, torch.cuda.LongTensor)
-            net.to(torch.empty(1, device="cuda:0", dtype=torch.half))
+            net.to(torch.empty(1, device="cuda", dtype=torch.half))
```
mikaylagawarecki: nit: no need to change this

(also could you add a comment above the `TEST_CUDA` block that we don't make this device-generic as this is testing legacy features please)

test/test_nn.py
```diff
@@ -7570,11 +7510,11 @@ def test_layer_norm_backwards_eps(self):
                 if bias and elementwise_affine:
                     self.assertEqual(ln.bias.grad, ln_cuda.bias.grad, f"bias grad failed: {m=} {n=}", rtol=rtol, atol=atol)
 
-    @unittest.skipIf(not TEST_CUDA, "CUDA not available")
+    @onlyAccelerator
     @largeTensorTest("40GB", device="cuda")
```
mikaylagawarecki: since these tests are now device-generic, I think it might also be reasonable to remove the `device=` argument from `largeTensorTest`, wdyt?

That said I'm also okay with it if we handle that in a separate PR

test/test_nn.py
```diff
@@ -4443,18 +4420,17 @@ def test_error_RNN_seq_len_zero(self):
...
-                    input = input.cuda()
-                outs = rnn(input)
-                self.assertEqual(outs[0].shape, torch.Size([5, 0, 4]))
-                # Check that backward does not cause a hard error
-                outs[0].sum().backward()
+            input = torch.zeros((5, 0, 3))
+            rnn = module(input_size=3, hidden_size=4)
+            if device in ['cuda', 'xpu', 'mps']:
+                rnn.to(device)
+                input = input.to(device)
```
mikaylagawarecki: No need to hardcode this list here
```suggestion
            rnn.to(device)
            input = input.to(device)
```

test/test_nn.py
```diff
@@ -6443,13 +6396,13 @@ def test_shape(N, C, ID, IH, IW, D, H, W, mode, padding_mode, align_corners):
...
+                        test(N, C, D, H, W, mode, padding_mode, align_corners, input_requires_grad, device=device)
 
-    def test_grid_sample_nearest_neighbor_rounding_mode_consistency(self):
+    def test_grid_sample_nearest_neighbor_rounding_mode_consistency(self, device):
 
         device_list = ['cpu']
-        if TEST_CUDA:
-            device_list.append('cuda')
+        if ACCELERATOR_TYPE == device:
+            device_list.append(device)
```
mikaylagawarecki: claude's comment is also accurate here:
why does this test still have the loop over `device_list`?

test/test_nn.py
```diff
@@ -7523,9 +7463,9 @@ def test_layer_norm_grads_with_create_graph_flag(self):
 
         self.assertEqual(grads1, grads2, rtol=rtol, atol=atol)
 
-        if TEST_CUDA:
-            x = x.to('cuda')
-            layer_norm = layer_norm.to('cuda')
+        if ACCELERATOR_TYPE == device:
```
mikaylagawarecki: hm maybe I'm missing something, but for this one isn't it ok to just delete this path and run the code above with tensors / LayerNorm that is instantiated on device?

test/test_nn.py
```diff
@@ -14277,7 +14214,14 @@ def __init__(self) -> None:
...
 
 instantiate_device_type_tests(TestNNDeviceType, globals(), allow_mps=True)
-instantiate_parametrized_tests(TestNN)
+
+# https://github.com/pytorch/pytorch/issues/177119
+if os.environ.get('PYTORCH_TEST_WITH_DYNAMO', '0') == '0':
+    device_list = ('cuda', 'cpu')
+else:
+    device_list = ('cuda',)
+instantiate_device_type_tests(TestNN, globals(), only_for=tuple(device_list))
```
mikaylagawarecki: why did we need to add mps skips above if this is only_for=('cpu', 'cuda')?

test/test_nn.py
```diff
@@ -14277,7 +14214,14 @@ def __init__(self) -> None:
 
 
 instantiate_device_type_tests(TestNNDeviceType, globals(), allow_mps=True)
-instantiate_parametrized_tests(TestNN)
+
+# https://github.com/pytorch/pytorch/issues/177119
+if os.environ.get('PYTORCH_TEST_WITH_DYNAMO', '0') == '0':
+    device_list = ('cuda', 'cpu')
```
mikaylagawarecki: Should this be 
```suggestion
if os.environ.get('PYTORCH_TEST_WITH_DYNAMO', '0') == '0':
    device_list = None
```
(otherwise the `only_for` will prevent 3rd party backends from using this unless its overidden via an env var)

test/test_nn.py
mikaylagawarecki: q: why did we need to add mps skips above if this is only_for=('cpu', 'cuda')?

mikaylagawarecki (commented):
Thanks for your effort on this, I know it is not easy to go make the entire class device-generic!

I went over the PR manually and I think most of my comments ended up matching the claude comments.  I think this is pretty close, just a couple more small comments

test/test_nn.py
daisyden: I tried to keep the original logic, but you are right, accelerator should also work on the code. I will remove the loop.

test/test_nn.py
daisyden: The skipIfMps in this PR are for TestNNDeviceType which is instantiate with instantiate_device_type_tests(TestNNDeviceType, globals(), allow_mps=True).

daisyden: The skipIfMPS in this PR are for TestNNDeviceType which is instantiate with instantiate_device_type_tests(TestNNDeviceType, globals(), allow_mps=True).

test/test_nn.py
daisyden: Can I use except_for? 
```
if os.environ.get('PYTORCH_TEST_WITH_DYNAMO', '0') == '1':
    device_list = ('cpu', )

instantiate_device_type_tests(TestNN, globals(), except_for=device_list)
```

test/test_nn.py
daisyden: I will do the update.

daisyden:
@claude Could you do another pass on this PR, taking into account previous review comments. Make sure that the functionality being tested was not changed.

guangyey (approved):

mikaylagawarecki (approved):
Thank you! 

For my understanding -- the main change in this PR was to instantiate `TestNN` with `instantiate_device_type_tests`. `TestNN` now no longer has `@onlyCUDA`. It seems like this PR also migrated some (but not all) `@onlyCUDA` in `TestNNDeviceType` to `@onlyAccelerator`. Will there be a follow-up PR that removes more `@onlyCUDA` from `TestNNDeviceType` (or moves them into a separate class if they are truly cuda-specific)?

test/test_nn.py
```diff
@@ -5307,23 +5247,28 @@ def test_batchnorm_buffer_update_when_stats_are_not_tracked(self):
...
-                                        "test_batchnorm_3D_train_NCHW_vs_cpu_mixed_bfloat16",
-                                        "test_batchnorm_2D_train_NHWC_vs_NCHW_mixed_bfloat16",
-                                        "test_batchnorm_3D_train_NHWC_vs_NCHW_mixed_bfloat16",
-                                        "test_batchnorm_3D_train_NCHW_vs_native_mixed_float16"):
-                self.skipTest("Failed on CUDA")
+    def test_batchnorm(self, device, dims, mode, memory_format, ref_backend, mixed, dtype):
+        _dtype = dtype_name(dtype)
+        if torch.version.cuda or device == 'xpu':
+            skip_tests = [
+                f"test_batchnorm_2D_train_NCHW_vs_cpu_mixed_bfloat16_{device}_{_dtype}",
```
malfet, line 5255: Err, why do you need to parametrize test name of of a sudden?

test/test_nn.py
```diff
@@ -11514,14 +11393,16 @@ def test_grid_sample_large_index_3d(self, device, dtype):
             small_image.grad.zero_()
             large_view.grad.zero_()
 
-    @onlyCUDA
-    def test_grid_sample_half_precision(self):
+    @onlyAccelerator
+    # TypeError: Cannot convert a MPS Tensor to float64 dtype as the MPS framework doesn't support float64
+    @skipIfMPS
```
malfet, line 11401: Never use `skipIfMPS`, rather `@expectedFailureMPS` or something of that nature

test/test_nn.py
```diff
@@ -14332,7 +14216,14 @@ def __init__(self) -> None:
...
 
 instantiate_device_type_tests(TestNNDeviceType, globals(), allow_mps=True)
-instantiate_parametrized_tests(TestNN)
+
+device_list = None
+# https://github.com/pytorch/pytorch/issues/177119
+if os.environ.get('PYTORCH_TEST_WITH_DYNAMO', '0') == '1':
+    device_list = ('cpu', )
+
+instantiate_device_type_tests(TestNN, globals(), except_for=device_list)
```
malfet, line 14229: Can you elaborate why this is needed?

malfet (changes requested):
Two comments:
 - Test plan for changes like that should include links to some runs before and after the change to guarantee there are no significant changes in the number of tests run before/after change (as this have a potential of skipping lots of those for no good reasion
 - Never use `skipIf` decorator unless it hard crashes, instead use `extectedFailureIf`

mikaylagawarecki:
@daisyden I am sorry it is my bad that some of my comments led to more changes being put into this PR. Perhaps we can break it up like

1. Instantiate TestNN with instantiate_device_type_tests, change 'cuda' → device                                                                                                         
2. Replace unittest.skipIf(TEST_CUDA) / onlyCUDA with onlyAccelerator as appropriate
3. Remove unnecessary internal loops over device

 @malfet let me know if this does not sound reasonable

test/test_nn.py
daisyden: @malfet my CI triggered dynamo_wrapped test on CPU, which detected a torch CPU bug that can be reproduced with nightly build, see https://github.com/pytorch/pytorch/issues/177119, as it impacted a lot of tests I skipped it. Sorry maybe it is not a good solution.

test/test_nn.py
daisyden: Hi, @malfet, when TestNN is instantiated with instantiate_device_type_tests() 'device' parameter is added to test_batchnorm(), so the names of functions to skip are changed. And as xpu can pass one more case than cuda I added a conditional statement.

PR #166398, opened 2025-10-28
Remove clang-tidy type conversion suppressions

cyyever:
This PR fixes and removes type conversion suppressions of clang-tidy.

### DIFF
--- a/aten/src/ATen/native/IndexingUtils.cpp
+++ b/aten/src/ATen/native/IndexingUtils.cpp
@@ -16,8 +16,7 @@ bool canUse32BitIndexMath(const TensorBase& t, int64_t max_elem) {
   auto linearId = elements - 1;
 
   // NOTE: Assumes all strides are positive, which is true for now
-  // NOLINTNEXTLINE(bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions)
-  for (int i = t.dim() - 1; i >= 0; --i) {
+  for (int64_t i = t.dim() - 1; i >= 0; --i) {
     auto curDimIndex = linearId % t.sym_size(i);
     auto curDimOffset = curDimIndex * t.sym_stride(i);
     offset += curDimOffset;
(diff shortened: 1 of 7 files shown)


aten/src/ATen/native/quantized/cpu/kernels/QuantizedOpKernels.cpp
```diff
@@ -1326,23 +1313,20 @@ void qadd_scalar_kernel(Tensor& out, const Tensor& self, const Scalar& other) {
 template <bool ReLUFused = false>
 void qadd_kernel(Tensor& out, const Tensor& self, const Tensor& other) {
   int64_t zero_point = out.q_zero_point();
-  // NOLINTNEXTLINE(bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions)
-  float scale = out.q_scale();
+  float scale = static_cast<float>(out.q_scale());
```
Skylion007, line 1316: So these should be casted to float right?

aten/src/ATen/native/IndexingUtils.cpp
```diff
@@ -16,8 +16,7 @@ bool canUse32BitIndexMath(const TensorBase& t, int64_t max_elem) {
   auto linearId = elements - 1;
 
   // NOTE: Assumes all strides are positive, which is true for now
-  // NOLINTNEXTLINE(bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions)
-  for (int i = t.dim() - 1; i >= 0; --i) {
+  for (int64_t i = t.dim() - 1; i >= 0; --i) {
```
Skylion007: No reason not to use auto here?

Skylion007 (approved):

aten/src/ATen/native/IndexingUtils.cpp
cyyever: fixed

aten/src/ATen/native/quantized/cpu/kernels/QuantizedOpKernels.cpp
cyyever: Yes, because they are passed below to functions which only accept float.

PR #166400, opened 2025-10-28
[inductor] Respect pg topology for nccl collectives estimations

IvanKobzarev:
* #166521
* __->__ #166400

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @chenyang78

### DIFF
--- a/torch/_inductor/comm_analysis.py
+++ b/torch/_inductor/comm_analysis.py
@@ -260,7 +260,7 @@ class NVLinkConnection(Connection):
         "2.0": 50.0,  # 25 GB/s unidirectional × 2
         "3.0": 50.0,  # 25 GB/s unidirectional × 2 (A100)
         "4.0": 50.0,  # 25 GB/s unidirectional × 2 (H100, H200)
-        "5.0": 400,  # 200 GB/s unidirectional x 2 (GB200)
+        "5.0": 400.0,  # 200 GB/s unidirectional x 2 (GB200)
     }
 
     def __post_init__(self) -> None:
--- a/torch/_inductor/comm_analysis.py
+++ b/torch/_inductor/comm_analysis.py
@@ -2,7 +2,7 @@
 import logging
 import math
 from abc import ABC, abstractmethod
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from enum import Enum, IntEnum
 from functools import lru_cache
 from typing import ClassVar, Optional
@@ -254,6 +254,7 @@ class NVLinkConnection(Connection):
 
     version: str
     num_links: int
+    _bandwidth: float = field(init=False)
     # Bidirectional bandwidth per link (GB/s)
     BANDWIDTH_PER_LINK: ClassVar[dict[str, float]] = {
         "1.0": 40.0,  # 20 GB/s unidirectional × 2
@@ -294,6 +295,7 @@ class InfiniBandConnection(Connection):
 
     rate: float = 200.0  # Default (4xHDR), ibstat Rate output, Gbps per direction
     num_ports: int = 4
+    _bandwidth: float = field(init=False)
 
     def __post_init__(self) -> None:
         self._bandwidth = self.rate / 8 * self.num_ports * 2
--- a/torch/_inductor/comm_analysis.py
+++ b/torch/_inductor/comm_analysis.py
@@ -1,12 +1,13 @@
-import functools
 import logging
 import math
 from enum import IntEnum
+from functools import lru_cache
 from typing import Optional
 
 import sympy
 
 import torch
+from torch.distributed import ProcessGroup
 from torch.fx.operator_schemas import normalize_function
 
 from . import ir
@@ -30,7 +31,7 @@ class NVIDIA_GPU_TYPE(IntEnum):
     HOPPER = 2
 
 
-@functools.lru_cache
+@lru_cache
 def get_gpu_type() -> NVIDIA_GPU_TYPE:
     gpu_info = torch.utils.collect_env.get_gpu_info(torch.utils.collect_env.run) or ""
     if "V100" in gpu_info:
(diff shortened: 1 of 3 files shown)


torch/_inductor/comm_analysis.py
```diff
@@ -216,8 +217,170 @@ def estimate_nccl_collective_runtime_nccl_estimator(snode) -> Optional[float]:
...
+
+
+@dataclass
+class NVLinkConnection(Connection):
+    """NVLink connection for intra-node GPU communication"""
+
+    version: str
+    num_links: int
+    # Bidirectional bandwidth per link (GB/s)
+    BANDWIDTH_PER_LINK = {
```
Skylion007: Use a dataclass.Field factory lambda. This is a class variable not an instance variable. All instances will share the same dictionary, which is probably not intended. If one of the instances mutates it will lead to unexpected behavior. If this is expected behavior, annotate with typing.ClassVar

torch/_inductor/comm_analysis.py
```diff
@@ -216,8 +217,170 @@ def estimate_nccl_collective_runtime_nccl_estimator(snode) -> Optional[float]:
...
+        "1.0": 40.0,  # 20 GB/s unidirectional × 2
+        "2.0": 50.0,  # 25 GB/s unidirectional × 2
+        "3.0": 50.0,  # 25 GB/s unidirectional × 2 (A100)
+        "4.0": 50.0,  # 25 GB/s unidirectional × 2 (H100, H200)
+        "5.0": 400,  # 200 GB/s unidirectional x 2 (GB200)
```
Skylion007: For consistency, 400 should probably be float. 400.0

torch/_inductor/comm_analysis.py
```diff
@@ -216,8 +217,170 @@ def estimate_nccl_collective_runtime_nccl_estimator(snode) -> Optional[float]:
...
+
+@dataclass
+class InfiniBandConnection(Connection):
+    """InfiniBand connection for inter-node communication"""
+
+    rate: float = 200.0  # Default (4xHDR), ibstat Rate output, Gbps per direction
+    num_ports: int = 4
+
+    def __post_init__(self) -> None:
+        self._bandwidth = self.rate / 8 * self.num_ports * 2
```
Skylion007, line 301: Nit: have it be listed along with the other instance variables like rate and num_ports but set Type dataclass.InitVar[float]. That way static type checking will work and it knows the attribute exists. You would still manually est it in the __post_init__ and it would complain if it's not assigned there.

torch/_inductor/compile_fx.py
```diff
@@ -1327,6 +1327,19 @@ def codegen_and_compile(
                     },
                     payload_fn=lambda: inductor_post_grad_graph_str,
                 )
+                fd = io.StringIO()
```
Skylion007: StringIO file descriptor not closed, use with context here?

torch/_inductor/comm_analysis.py
```diff
@@ -216,8 +217,170 @@ def estimate_nccl_collective_runtime_nccl_estimator(snode) -> Optional[float]:
...
+    for r in range(size):
+        uuid = gathered[r][0]
+        if r == rank:
+            conn.append(CurrentDevice())
+            continue
+        if uuid in uuid_to_nvlinkconn:
+            conn.append(uuid_to_nvlinkconn[uuid])
+            continue
+
+        # TODO: get number of ports and ib rate, parsing ibstat?
```
Skylion007, line 371: For Nvidia specifically, I think Nvidia SMI bindings also provide it at least for local ranks.

torch/_inductor/compile_fx.py
IvanKobzarev: Oh this got here by mistake. Will clean it.

torch/_inductor/comm_analysis.py
IvanKobzarev: I switched to field(init=False)

torch/_inductor/comm_analysis.py
```diff
@@ -216,8 +220,164 @@ def estimate_nccl_collective_runtime_nccl_estimator(snode) -> Optional[float]:
...
+
+@dataclass
+class CurrentDevice(Connection):
+    """Dummy Implementation"""
+
+    type: ConnectionType = ConnectionType.CURRENT_DEVICE
+    bandwidth: float = 0.0
+
+
+@dataclass
```
Skylion007, line 251: Do you want people to create the data classes using positional args? Might be a good idea to specify kwargs only to the dataclass decorator prevent accidental swapping of ctor args since they are both numbers.

torch/_inductor/comm_analysis.py
```diff
@@ -1,12 +1,16 @@
-import functools
+# mypy: disable-error-code=attr-defined
```
Skylion007, line 1: I would do inline suppressions. This is removing one of the most useful mypy error codes

torch/_inductor/comm_analysis.py
```diff
@@ -216,8 +220,164 @@ def estimate_nccl_collective_runtime_nccl_estimator(snode) -> Optional[float]:
...
+    version: str
+    num_links: int
+    _bandwidth: float = field(init=False)
+    # Bidirectional bandwidth per link (GB/s)
+    BANDWIDTH_PER_LINK: ClassVar[dict[str, float]] = {
+        "1.0": 40.0,  # 20 GB/s unidirectional × 2
+        "2.0": 50.0,  # 25 GB/s unidirectional × 2
+        "3.0": 50.0,  # 25 GB/s unidirectional × 2 (A100)
+        "4.0": 50.0,  # 25 GB/s unidirectional × 2 (H100, H200)
+        "5.0": 400.0,  # 200 GB/s unidirectional x 2 (GB200)
```
eellison, line 264: What does 5.0 mean here ? what is the key ?

torch/_inductor/comm_analysis.py
```diff
@@ -216,8 +220,164 @@ def estimate_nccl_collective_runtime_nccl_estimator(snode) -> Optional[float]:
...
+
+    def __repr__(self) -> str:
+        return f"NVLinkConnection(version='{self.version}', num_links={self.num_links}, bandwidth={self.bandwidth:.0f})"
+
+
+@dataclass
+class InfiniBandConnection(Connection):
+    """InfiniBand connection for inter-node communication"""
+
+    rate: float = 200.0  # Default (4xHDR), ibstat Rate output, Gbps per direction
```
eellison, line 296: Where are we getting these numbers ?

torch/_inductor/comm_analysis.py
```diff
@@ -216,8 +220,164 @@ def estimate_nccl_collective_runtime_nccl_estimator(snode) -> Optional[float]:
...
+
+    props = torch.cuda.get_device_properties(current_device_idx)
+    uuid = str(props.uuid)
+    gathered: list[list[str]] = [[] for _ in range(size)]
+    torch.distributed.all_gather_object(gathered, [uuid], pg)
+    nvlink_n = len(nvlink_matrix)
+    uuid_to_nvlinkconn = {}
+    for dev_idx in range(nvlink_n):
+        # TODO: retrieve nvlink version
+        version = "4.0"
```
eellison, line 355: this is nvlink version ? does it vary by hardware?

torch/_inductor/comm_analysis.py
```diff
@@ -236,13 +396,25 @@ def estimate_nccl_collective_runtime_impl(
 
     # Currently assumes each node has 8 gpus. And when >1 node is used, assumes each node uses all 8 gpus.
     # TODO: Need to find a way to get accurate "gpus per node" and "# nodes" info.
```
eellison, line 398: I think this is more reason to just use the nccl estimator directly.

torch/_inductor/comm_analysis.py
```diff
@@ -216,8 +220,164 @@ def estimate_nccl_collective_runtime_nccl_estimator(snode) -> Optional[float]:
...
+    devices_uuids = [
+        str(torch.cuda.get_device_properties(i).uuid)
+        for i in range(torch.cuda.device_count())
+    ]
+    current_device_idx = torch.cuda.current_device()
+
+    props = torch.cuda.get_device_properties(current_device_idx)
+    uuid = str(props.uuid)
+    gathered: list[list[str]] = [[] for _ in range(size)]
+    torch.distributed.all_gather_object(gathered, [uuid], pg)
```
eellison, line 350: Is it worth specifying somewhere that this does an all gather ? If this is called in non-spmd this will hang. Why do we need to do this?

eellison (commented):
Would it make sense to just use the nccl estimator instead? I dont think the inductor config values are kept very up to date..

eellison (commented):
Would it make sense to just use the nccl estimator instead? I dont think the inductor config values are kept very up to date..

eellison (commented):
Would it make sense to just use the nccl estimator instead? I dont think the inductor config values are kept very up to date..

eellison (commented):
Would it make sense to just use the nccl estimator instead? I dont think the inductor config values are kept very up to date..

torch/_inductor/comm_analysis.py
IvanKobzarev: The NVLink version

torch/_inductor/comm_analysis.py
IvanKobzarev: So far I hardcoded what we have on our devservers, using `ibstat` output

torch/_inductor/comm_analysis.py
IvanKobzarev: Yes, nvlink version. Mostly not, but there is variance for some limited versions of hardware. Here hardcoded what we have for H100, but we can retrieve it with API.
For now we can also check the model of gpu and set the version according to gpu.

torch/_inductor/comm_analysis.py
IvanKobzarev: We need to gather unique id of GPUs from all process_group. And then we match this unique ids with local node devices to find which nodes are on NVLink, others will be on IB.

PR #160079, opened 2025-08-07
Make manylinux build.sh work for AArch64 and AArch64+CUDA builds

robert-hardwick:
### Modify .ci/manylinux/build.sh for AArch64 builds

Most platforms currently use scripts in `.ci/manylinux` to produce whl files for releases e.g. nightly. AArch64 cpu and AArch64 + CUDA manylinux wheel builds, however currently use scripts in `.ci/linux-aarch64` to produce equivalent wheels.

Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #160079

cc @malfet @snadampal @milpuz01 @aditew01 @nikhil-arm @fadara01

PR #166407, opened 2025-10-28
Fix -ffunction-sections, -fdata-sections not being added on aarch64.

robert-hardwick:
Preferred solution to #166380

Changes:

- Moved summary print to bottom of CMakeLists.txt
- Fix the problem 'add_compile_options' should be called before targets defined, so opted for `append_cxx_flag_if_supported` and `append_c_flag_if_supported` ( new ).
- Added extra verbosity so it can be seen when linker script added.

( unfortunately linker script has to be added per-target rather than globally due to ninja/cmake depdendency tracking ).

* #160079
* __->__ #166407

Also move summary print to bottom of CMakeLists.txt and improve logging


Aidyn-A:
>The linker script target generate_linker_script needs to be run before any of the build targets, so effectively needs to be a global dependency of all the targets.

Right `add_custom_target` requires a target, but instead you may use `execute_process` to prepare a linker script at the CMake stage, long before Ninja starts doing anything.

Like here for example: https://github.com/pytorch/pytorch/blob/2de4cf2102fdc35da00cc8070fdbcaf57198a576/cmake/Codegen.cmake#L38-L43

CMakeLists.txt
```diff
@@ -726,6 +726,44 @@ if(NOT DEFINED USE_BLAS)
...
+
+  execute_process(
+    COMMAND ${Python_EXECUTABLE}
+            ${CMAKE_SOURCE_DIR}/tools/setup_helpers/generate_linker_script.py
+            --filein "${LINKER_SCRIPT_FILE_IN}"
+            --fout  "${LINKER_SCRIPT_FILE_OUT}"
+    RESULT_VARIABLE _gen_result
+    OUTPUT_VARIABLE _gen_output
+    ERROR_VARIABLE  _gen_error
+  )
```
Aidyn-A, line 743: Awesome!

CMakeLists.txt
```diff
@@ -726,6 +726,44 @@ if(NOT DEFINED USE_BLAS)
...
+      "Failed to generate linker script:\n${_gen_output}\n${_gen_error}")
+  endif()
+
+  append_cxx_flag_if_supported("-ffunction-sections" CMAKE_CXX_FLAGS)
+  append_cxx_flag_if_supported("-fdata-sections" CMAKE_CXX_FLAGS)
+  append_c_flag_if_supported("-ffunction-sections" CMAKE_C_FLAGS)
+  append_c_flag_if_supported("-fdata-sections" CMAKE_C_FLAGS)
+
+  set(CMAKE_SHARED_LINKER_FLAGS "${CMAKE_SHARED_LINKER_FLAGS} -T${LINKER_SCRIPT_FILE_OUT}")
+  set(CMAKE_MODULE_LINKER_FLAGS "${CMAKE_MODULE_LINKER_FLAGS} -T${LINKER_SCRIPT_FILE_OUT}")
```
Aidyn-A, line 756: I see it is getting applied in the Summary:
```
2025-11-03T15:18:21.9325120Z --   Shared LD flags       :  -Wl,--no-as-needed -T/pytorch/cmake/linker_script.ld -rdynamic
2025-11-03T15:18:21.9325515Z --   Static LD flags       : 
2025-11-03T15:18:21.9325813Z --   Module LD flags       :  -T/pytorch/cmake/linker_script.ld
```

Aidyn-A (approved):
Looks good to me now!
 
@malfet could you please review?

atalman (approved):
lgtm, thank you

PR #166416, opened 2025-10-28
ci: Add aarch64 docker builds for modern clang

seemethere:
* #166441
* __->__ #166416

Should enable us to build using some arm optimizations that are only
available on the newest versions of clang.


.github/workflows/docker-builds.yml
```diff
@@ -77,6 +77,10 @@ jobs:
         include:
           - docker-image-name: pytorch-linux-jammy-aarch64-py3.10-gcc11
             runner: linux.arm64.m7g.4xlarge
+          - docker-image-name: pytorch-linux-jammy-aarch64-py3.10-clang19
```
malfet: Why do we need clang-19 and clang-21?

malfet (approved):
If it compiles, ship it, but I would probably just stick to the latest clang currently supported

.github/workflows/docker-builds.yml
seemethere: ~This is actually borked, jammy doesn't have either 19 or 21 so we'll need to build from source.~

I figured out how to do this without actually needing to build from source.

I think we aligned that only 21 is needed for now.

seemethere: I removed 19 and only kept 21

PR #166423, opened 2025-10-28
Fix MSAN errors in get_env_mutex

pganssle-google:
When running MSAN over PyTorch, this is detected as `use-of-uninitialized-value` and should be heap-allocated.

Also noticed that both `std::optional` and `std::string` are being used without the headers directly included, so that has been updated in this PR as well.

### DIFF
--- a/c10/util/env.cpp
+++ b/c10/util/env.cpp
@@ -3,7 +3,9 @@
 #include <fmt/format.h>
 #include <cstdlib>
 #include <mutex>
+#include <optional>
 #include <shared_mutex>
+#include <string>
 
 namespace c10::utils {
 
--- a/c10/util/env.cpp
+++ b/c10/util/env.cpp
@@ -9,9 +9,10 @@
 
 namespace c10::utils {
 
-static std::shared_mutex& get_env_mutex() {
-  static std::shared_mutex env_mutex;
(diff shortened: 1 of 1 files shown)


jkammerl:
I believe it is actually a true positive case. By defining the static object std::shared_mutex env_mutex, its destructor among other destructors will be called in an arbitrary order and it can be freed while other threads may still access it. We do see a unit test that runs into that problem. Creating the object with new, we ensure that the destructor is actually never called.

c10/util/env.cpp
```diff
@@ -3,13 +3,15 @@
...
 #include <shared_mutex>
+#include <string>
 
 namespace c10::utils {
 
-static std::shared_mutex& get_env_mutex() {
-  static std::shared_mutex env_mutex;
-  return env_mutex;
+std::shared_mutex& get_env_mutex() {
+  static std::shared_mutex* env_mutex = new std::shared_mutex();
```
cyyever, line 14: Add a comment for pointing out the deliberate leakage?

cyyever (approved):

c10/util/env.cpp
pganssle-google: Oops, sorry I missed this. Done.

PR #166299, opened 2025-10-27
[xpu][feature] Introduce ExpandableSegment for XPU

guangyey:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #166495
* #166424
* #166292
* __->__ #166299
# Motivation
This PR intends to add `ExpandableSegment` struct, which is used to help support the expandable segment feature. I split it to a single PR to facilitate the code review.

PR #166292, opened 2025-10-27
[xpu][feature] Support expandable segment feature for XPU

guangyey:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #166495
* #166424
* __->__ #166292
* #166299
# Motivation
This PR intends to add expandable segment feature support on XPU. This will help
- Reduce memory fragmentation;
- Gradually map physical pages into virtual address space as needed.

# Additional Context
The traditional caching allocator frequently allocates and frees device memory blocks. However, over time, with varying tensor size, the device address space becomes fragmented. Even when there's enough total free memory, a lack of contiguous space can cause large allocations to fail.
The **expandable segment** feature addresses this by dynamically extending physical memory within a reserved virtual address range, reducing fragmentation and minimizing reallocation overhead.
The potential drawbacks are
- Virtual memory overhead;
- Potential page mapping overhead;
- Increased complexity.

PR #166424, opened 2025-10-28
[xpu][feature] Introduce PeerToPeerAccess API for XPU

guangyey:
* #166495
* __->__ #166424
* #166292
* #166299
# Motivation
This PR introduces support for peer-to-peer (P2P) access between devices, including querying and enabling P2P connections between two devices.
It supports two categories of allocations:
- Regular allocations;
- Expandable segment allocations.

# Additional Context
The follow-up is that we should use this feature to optimize our copy kernel when P2P is supported.


c10/xpu/XPUCachingAllocator.cpp
```diff
@@ -1209,6 +1224,14 @@ class XPUAllocator : public DeviceAllocator {
     assertValidDevice(device);
     device_allocators[device]->resetAccumulatedStats();
   }
+
+  void enablePeerAccess(c10::DeviceIndex dev, c10::DeviceIndex dev_to_access) {
+    assertValidDevice(dev);
+    assertValidDevice(dev_to_access);
+    c10::xpu::get_raw_device(dev).ext_oneapi_enable_peer_access(
```
gujinghui, line 1239: Should we handle the exception if failed? Or, you assume the ext_oneapi_can_access_peer already checked before calling here?

aten/src/ATen/xpu/PeerToPeerAccess.cpp
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  // Query the hardware to determine if P2P access is supported
+  cache = static_cast<int8_t>(
+      c10::xpu::get_raw_device(dev).ext_oneapi_can_access_peer(
+          c10::xpu::get_raw_device(dev_to_access),
+          sycl::ext::oneapi::peer_access::access_supported));
+
+  if (cache) {
+    XPUCachingAllocator::enablePeerAccess(dev, dev_to_access);
```
gujinghui, line 57: why enablePeerAccess belongs to XPUCachingAllocator?

c10/xpu/XPUCachingAllocator.cpp
```diff
@@ -1047,6 +1047,21 @@ class DeviceCachingAllocator {
       stats.requested_bytes[statType].reset_peak();
     }
   }
+
+  void addPeerAccess(c10::DeviceIndex dev_to_access) {
```
gujinghui: better to move all related APIs to PeerToPeerAccess files?

gujinghui:
Any existing case for P2P?

c10/xpu/XPUCachingAllocator.cpp
guangyey: Yes, see the following code in PeerToPeerAccess.cpp, we only enable peer access when they are supported.
```cpp
if (cache) {
    XPUCachingAllocator::enablePeerAccess(dev, dev_to_access);
}
```

aten/src/ATen/xpu/PeerToPeerAccess.cpp
guangyey: 1. The major reason is that it follows CUDA's design. It facilitates further code unification.
2. The memory blocks are managed by CachingAllocator, which are the internal member that CachingAllocator belongs to.

c10/xpu/XPUCachingAllocator.cpp
guangyey: `expandable_segments_` is an internal member of CachingAllocator. In addition, in the future, other internal states (such as those tracking IPC status) are also maintained within CachingAllocator. Therefore, moving this API to another file would be difficult. And following the current design helps us to unify code better.

guangyey:
> Any existing case for P2P?

For copy path.
https://github.com/pytorch/pytorch/blob/695cb0d342f67db563fb57ce1a5f3261b073b790/aten/src/ATen/native/cuda/Copy.cu#L364-L369

c10/xpu/XPUCachingAllocator.cpp
gujinghui: > helps us to unify code better.

So, for now, we are following CUDA impl to reuse the code in allocator.
Sounds like we need better/much clear design when abstract the allocator next.

gujinghui (approved):

albanD (approved):
SGTM!

PR #166440, opened 2025-10-28
[DebugMode] add stack traces

pianpwk:
Captures stack trace for torch_dispatch calls, under `with DebugMode(record_stack_trace=True)`: Traces aren't rendered in debug string, but are in `.stack_trace` for each log.

* __->__ #166440


yushangdi (approved):

torch/utils/_debug_mode.py
```diff
@@ -103,6 +106,17 @@ def default_hash_fn(t: torch.Tensor, use_scalar: bool = False) -> torch.Tensor:
     return out
 
 
+def _get_stack_trace(cpp=False) -> str:
```
mikaylagawarecki: Just happened to look at this as I was looking at https://github.com/pytorch/pytorch/pull/166348

Curious qn if we're in backward -- does this show "which line in forward triggered a certain op in backward to hit TorchDispatchMode"? If not, I think that could be a useful feature

I had to set `torch.autograd.set_detect_anomaly(True)` and do something like [this](https://www.internalfb.com/code/fbsource/[82330672c52b3a6babf491c4d7d7c81a753cca28]/genai/llama4x/tools/hash/observer.py?lines=398-401) to get that, but not sure if the same applies here

pianpwk: ah this is a great point, not yet, but I think that'll be a separate field (fwd_stack_trace?) - there's merit with just the raw bwd trace too.

I'll spend more time trying to disentangle it from anomaly mode - I don't think DebugMode should always raise on NaN gradients, and DTensor doesn't seem to love the NaN checks: seeing stuff like `NotImplementedError: Operator aten._is_any_true.default does not have a sharding strategy registered.`

mikaylagawarecki: Ah there's a `check_nan=False` flag in set_detect_anomaly, but I assume you already tried that

pianpwk: my bad, didn't realize that before landing... Following up in #166842

PR #166447, opened 2025-10-28
[ez] Print some more test timing info in the logs

clee2000:
You can just subtract timestamps, but this makes it easier

### DIFF
--- a/test/run_test.py
+++ b/test/run_test.py
@@ -1826,13 +1826,13 @@ def run_test_module(
         test_name = test.name
 
         # Printing the date here can help diagnose which tests are slow
-        start = datetime.now()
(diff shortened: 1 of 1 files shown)


test/run_test.py
```diff
@@ -1826,9 +1826,14 @@ def run_test_module(
         test_name = test.name
 
         # Printing the date here can help diagnose which tests are slow
-        print_to_stderr(f"Running {str(test)} ... [{datetime.now()}]")
+        start = datetime.now()
```
Skylion007: Use a monotonic, high resolution clock like time.perf_counter()

Skylion007 (approved):

PR #166449, opened 2025-10-28
[pp] Add reduce_grad Action

IvanKobzarev:
* __->__ #166449


test/distributed/pipelining/test_schedule.py
```diff
@@ -574,6 +575,45 @@ def test_unshard_reshard(self, test_info):
...
+                ],
+            },
+        ],
+    )
+    def test_reduce_grad(self, test_info):
+        compute_sch = self._parse_actions(test_info["compute"])
+        expected_comms_sch = self._parse_actions(test_info["comms"])
+
+        comms_sch = _add_reduce_grad(compute_sch)
+        for expected, actual in zip(expected_comms_sch, comms_sch):
```
Skylion007:
```suggestion
        for expected, actual in zip(expected_comms_sch, comms_sch, strict=True):

File: torch/distributed/pipelining/schedules.py [hunk 2]
```diff
@@ -1059,6 +1064,35 @@ def _get_pipeline_order(self) -> Optional[dict[int, list[Optional[_Action]]]]:
         return pipeline_order
 
 
+def _requires_reduce_grad(action_type: _ComputationType):
```
Skylion007:
```suggestion
def _requires_reduce_grad(action_type: _ComputationType) -> bool:

File: torch/distributed/pipelining/schedules.py [hunk 3]
```diff
@@ -1055,10 +1060,60 @@ def _get_pipeline_order(self) -> Optional[dict[int, list[Optional[_Action]]]]:
...
                     remaining_backward -= 1
 
-            pipeline_order[rank] = actions
+            pipeline_order[rank] = _add_reduce_grad(actions)
         return pipeline_order
 
 
+def _requires_reduce_grad(action_type: _ComputationType) -> bool:
+    return action_type in (I, W, B)
+
```
sanketpurandare, line 1069: the `I` action wouldn't require a reduce_grad, only `B` and `W` would require it right?

torch/distributed/pipelining/schedules.py
```diff
@@ -1055,10 +1060,60 @@ def _get_pipeline_order(self) -> Optional[dict[int, list[Optional[_Action]]]]:
...
+
+def _add_reduce_grad(actions: list[Optional[_Action]]) -> list[Optional[_Action]]:
+    """
+    REDUCE_GRAD refers to joint across minibatches grad reduction.
+    reduce_grad frees memory and we want to schedule it just after the last "backward"-like stage.
+    """
+    stage_with_bw_count: dict[int, int] = defaultdict(int)
```
wconstab: what's the point of counting, every stage should have the same number of backwards, equal to the number of microbatches, right?  is this just for validation sake?

torch/distributed/pipelining/schedules.py
```diff
@@ -1055,10 +1060,60 @@ def _get_pipeline_order(self) -> Optional[dict[int, list[Optional[_Action]]]]:
...
+            _count_leaf_action(a)
+
+    actions_with_reduce_grad: list[Optional[_Action]] = []
+    cnt: dict[int, int] = defaultdict(int)
+
+    def _leaf_action(a, to_schedule):
+        if _requires_reduce_grad(a.computation_type):
+            stage_index = a.stage_index
+            cnt[stage_index] += 1
+            if cnt[stage_index] == stage_with_bw_count[stage_index]:
```
wconstab: related to my comment about counting, i think what you did is fine but overly defensive, and probably better to follow the existing pattern from elsewhere in this file:
```               
last_backward = self.backward_counter[stage_idx] == self._n_microbatches
```

although you don't have access to self in here, so maybe pass n_microbatches into the add_reduce_grad helper, or maybe just ignore me and keep the counting method

torch/distributed/pipelining/schedules.py
```diff
@@ -1596,12 +1651,6 @@ def step(
             args_split, kwargs_split, targets_split, losses, return_outputs
         )
 
-        # Stage post processing
-        # TODO: remove this section and include as part of the schedule IR?
-        for stage in self._stages:
-            grad_scale_factor = self._n_microbatches if self.scale_grads else 1
-            stage._post_backward(grad_scale_factor)
```
wconstab, line 1603: note: this _post_backward call is renamed to 'perfom_reduce_grad' and called as part of the schedule IR

torch/distributed/pipelining/schedules.py
```diff
@@ -2185,6 +2238,9 @@ def _perform_action(action: _Action) -> None:
                     mb_index,
                     last_backward=last_backward,
                 )
+            elif comp_type == REDUCE_GRAD:
```
wconstab, line 2229: i'm probably ok with landing as-is for now, but i am wondering if we eventually get away from hardcoding the fsdp-specific logic into the stage.

for autoparallel, we're relying on someone to override stage.perform_reduce_grad, right?

wconstab (commented):
LGTM, but how have you tested this? in particular, i'd be a little worried about breaking existing FSDP/DDP compatibility.

IvanKobzarev:
> LGTM, but how have you tested this? in particular, i'd be a little worried about breaking existing FSDP/DDP compatibility.

Tested our tests that run schedules and do numerics checks.

I do not know if our tests cover DDP and FSDP cases, need to verify.

torch/distributed/pipelining/schedules.py
IvanKobzarev: I see, yes. I thought about very abstract schedules without invariants. Will change it to just n_microbatches.

torch/distributed/pipelining/schedules.py
IvanKobzarev: Yes, autoparallel will override all the stages.

IvanKobzarev:
> LGTM, but how have you tested this? in particular, i'd be a little worried about breaking existing FSDP/DDP compatibility.

@wconstab Tested torchtitan integration_tests - ok, 
and llama 3 with `--parallelism.data_parallel_shard_degree=4, --parallelism.pipeline_parallel_degree=2`

wconstab (approved):
there are explicit composability tests in here:
https://github.com/pytorch/pytorch/blob/main/test/distributed/_composable/test_composability/test_pp_composability.py
Just make sure they were run on your PR or add more tags to get CI to run all the tests if not

sanketpurandare (approved):

PR #166450, opened 2025-10-28
Add model code stack trace to cuda.memory._snapshot

yushangdi:
We store a mapping between generated fx graph code and original model code stack trace in `fx.traceback._FX_METADATA_REGISTRY`. And we do a post-processing on the memory snapshot to append the original model stack trace information. 

To achieve this, the biggest change we had to do in `aot_eager` mode is to give each generated fx graph a unique stack trace, i.e. it cannot just be `<eval_with_key>`. We set co_filename to **pretend** that the code is from `co_filename` file. Now instead of `<eval_with_key>` in stack trace, we get something like `fx_generated_3a4b5c6d7e8f9a0.py`. 

`augment_with_fx_traces` arg is added to `torch.cuda.memory._snapshot` and `_dump_snapshot`. When the arg is set to True, a post-processing will run to populate the original model stack trace to the snapshot frames. 

Alternative:

Instead of setting co_filename, we can also do it like below:
Note that if we do it this way, we will need to dump the file to make the graph module torch-scriptable. TorchScript requires source access in order to carry out compilation, so we need to make sure original .py files are available. 
```
        key = filename
        globals_copy = globals.copy()
        globals_copy["__file__"] = key
        globals_copy["__name__"] = key
        linecache.lazycache(key, globals_copy)
        exec(compile(src, key, "exec"), globals)
````

Other changes: 

- Update `MemoryViz.js` to display fx node information and original model code if exist

```
python test/test_fx.py -k test_lineno_map
python test/test_fx.py -k test_custom_traceback_raised
python test/test_public_bindings.py 
python test/test_cuda.py -k test_fx_memory
python test/test_fx.py -k test_informative_co_filename
python test/test_fx.py -k test_autowrap_functions
python test/dynamo/test_utils.py -k test_inductor_provenance
```

```python
# Profile with memory snapshot
torch.cuda.memory._record_memory_history()

compiled = torch.compile(mod, backend="aot_eager", fullgraph=True)
result = compiled(torch.randn(10, 10, device="cuda:0"))

torch.cuda.memory._dump_snapshot("memory_snapshot.pickle", augment_with_fx_traces=True)
torch.cuda.memory._record_memory_history(enabled=None)
```

Screenshot 2025-10-30 at 10 40 44 AM

### DIFF
--- a/torch/cuda/_utils.py
+++ b/torch/cuda/_utils.py
@@ -1,4 +1,7 @@
 import ctypes
+import os
+import pickle
+import re
 import sys
 from typing import Any, Optional, Union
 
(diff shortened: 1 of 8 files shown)


torch/fx/graph_module.py
```diff
@@ -828,7 +839,45 @@ def recompile(self) -> PythonCode:
...
+
+        if torch.fx.config.run_graph_module_code_from_file:
+
+            debug_dir = torch.fx.config.fx_graph_module_dir
+            if debug_dir is None:
+                debug_dir = tempfile.gettempdir()
+            os.makedirs(debug_dir, exist_ok=True)
+
+            # Generate a unique filename based on module class name and id
+            module_name = f"{FX_GRAPH_MODULE_FILE_PREFIX}{cls.__name__}_{id(self)}"
```
ezyang: don't do this. Instead, hash the source code so you have something like a content addressed store

yushangdi: updated now!

torch/fx/graph_module.py
```diff
@@ -825,10 +864,49 @@ def recompile(self) -> PythonCode:
...
         cls = type(self)
         co_fields = self._graph._co_fields if hasattr(self._graph, "_co_fields") else {}
-        cls.forward = _forward_from_src(self._code, python_code.globals, co_fields)
+
+        if co_fields:
+            cls.forward = _forward_from_src(self._code, python_code.globals, co_fields)
+        else:
+            node_metadata: dict[int, dict[str, Any]] = {}
+            for i, node in enumerate(self._graph.nodes):
+                node_metadata[i] = {
```
yushangdi, line 877: Question: Do we want to guard this part of the code behind a flag or make it the default behavior?

The trace in dumped snapshot is already guarded by the `augment_with_fx_traces` flag in `torch.cuda.memory._dump_snapshot("memory_snapshot.pickle", augment_with_fx_traces=True)`, it defaults to False.

torch/fx/graph.py
```diff
@@ -854,7 +856,13 @@ def _tensor_annotation(t: torch.Tensor) -> str:
 
 {prologue}
 {code}"""
-        return PythonCode(fn_code, globals_, _lineno_map=lineno_map)
+        prologue_start = wrap_stmts.count("\n") + 4
```
ezyang, line 859: where does the 4 come from?

torch/fx/graph_module.py
```diff
@@ -353,6 +362,36 @@ def _print_readable(
...
+    Returns:
+        A 51-character base32-encoded hash
+    """
+    import json
+
+    # Create a deterministic string representation of all components
+    # We use JSON to ensure consistent serialization
+    hash_data = {
+        "code": code,
+        "node_metadata": node_metadata,
```
ezyang, line 383: Remind me why the metadata needs to be hashed?

torch/fx/graph_module.py
```diff
@@ -353,6 +362,36 @@ def _print_readable(
...
+    }
+    hashing_str = json.dumps(hash_data).encode("utf-8")
+
+    # [:51] to strip off the "Q====" suffix common to every hash value.
+    return (
+        base64.b32encode(hashlib.sha256(hashing_str).digest())[:51]
+        .decode("utf-8")
+        .lower()
```
ezyang, line 391: ask an llm to do this better, it's in distribution

torch/fx/graph_module.py
ezyang: The main problem with making this default behavior is you're going to affect ALL of the FX codegen happening everywhere in the company, and so if there are any bugs you are getting yanked. The second problem is that this code might actually be temporary if we go from aot eager to inductor, inductor takes care of making the files anyway, so we wouldn't need this code. So I definitely vote flag. It's also not obvious to me that we are adequately flagged right now, what exactly is `co_field`?

torch/cuda/_utils.py
```diff
@@ -527,3 +530,121 @@ def _get_device_index(
         if isinstance(device, torch.cuda.device):
             return device.idx
     return _torch_get_device_index(device, optional, allow_cpu)
+
+
+# =============================================================================
+# Memory Profiler Utilities
+# =============================================================================
+# Utility functions for augmenting CUDA memory profiler snapshots with FX graph metadata.
```
ezyang, line 538: Is this really the right spot to put these?  Why not just in memory.py?

torch/cuda/_utils.py
```diff
@@ -527,3 +530,121 @@ def _get_device_index(
...
+
+
+def _augment_frames(frames: list[dict[str, Any]]) -> int:
```
ezyang, line 541: ask claude to improve the type signature, it should be easy to have a better one here than Any

ezyang:
This seems basically reasonable but I want to be careful about unintended consequences form the change.

torch/fx/graph.py
yushangdi: There are some empty lines between `wrap_stmts` and prologue: 

```
fn_code = f"""
{wrap_stmts}

{prologue}
{code}"""
```

torch/fx/graph_module.py
yushangdi: In the unit test `test_fx_memory_profiler_augmentation` example, the MLPModule and MLPModule2 would produce exactly the same graph code. However, executing them corresponds to different original model stack traces. So I added node_metadata in the hash to distinguish them.

torch/fx/graph_module.py
yushangdi: this is written by lllm, but it's the same code as what's used in `codecache` haha

https://github.com/pytorch/pytorch/blob/c3b71d54995a16fd09c18a5da5d598a23388bbd6/torch/_inductor/codecache.py#L342-L352

torch/cuda/_utils.py
yushangdi: makes sense, moved to memory.py now

torch/cuda/_utils.py
yushangdi: fixed now.

PR #166455, opened 2025-10-28
[ez] Remove maybe unused var in common_utils.py

clee2000:
I think we don't actually need this anymore since we don't use inspect.getfile to get the filename anymore


huydhn (approved):
LGTM! Not sure why lint fails in your case, maybe do a rebase

PR #161015, opened 2025-08-19
[DTensor] fix DTensorTestCase.destroy_pg() when device_type is "cpu" but CUDA device is available

XilunWu:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* __->__ #161015

**Summary**
When `device_id` is not None, barrier() will choose the accelerator of the most
pripority, which means if the test specifies to use CPU for testing while CUDA is
available on the host, the barrier() will use CUDA. To avoid this and better respect
`self.device_type`, we add this branch to enforce barrier() to use CPU when
`self.device_type` is CPU and other accelerator is also available.

**Test**
`pytest test/distributed/tensor/test_dtensor_testbase.py`

**Debugging Output**
```
# from init_process_group()
init pg: backend=gloo, device_id = None
default_pg has backend: gloo, device_types: [device(type='cuda'), device(type='cpu')]

# from barrier()
barrier: device_ids = [10], devices = [], device = None, PG=[device(type='cuda'), device(type='cpu')]
```

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta

PR #166463, opened 2025-10-28
[ptd] Fix test config in destroy_pg

fduwjj:
Summary: When device_type is CPU we will not use device id from CUDA which is enabled in https://github.com/pytorch/pytorch/pull/161015. However, we should not exclude the case when the accelerator itself is CPU. This PR fixes it.

Test Plan: UT

Differential Revision: D85714901

### DIFF
--- a/torch/testing/_internal/distributed/_tensor/common_dtensor.py
+++ b/torch/testing/_internal/distributed/_tensor/common_dtensor.py
@@ -447,7 +447,7 @@ class DTensorTestBase(MultiProcessTestCase):
                 torch.cuda.current_device() if self.device_type == "cuda" else self.rank
             )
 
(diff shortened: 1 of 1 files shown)


mori360 (approved):

torch/testing/_internal/distributed/_tensor/common_dtensor.py
```diff
@@ -447,7 +447,7 @@ def destroy_pg(self, device_id: Optional[int] = None) -> None:
                 torch.cuda.current_device() if self.device_type == "cuda" else self.rank
             )
 
-        if self.device_type == "cpu" and torch._C._get_accelerator().type != "cpu":
```
fegin, line 450: Do we know the reason why the condition was added? cc., @XilunWu

fegin (approved):

PR #166465, opened 2025-10-28
compile_worker: Make a timer class

c00w:
* #166584
* #166581
* #166467
* __->__ #166465

This subclass allows us to trigger an action after we haven't seen any activity
for a certain amount of seconds.


c00w:
@masnesral - The big issue is that threading.Timer can't be extended (i.e. reset the timer to be 30s in the future on each subproc call). If it could be, we'd use it. (The other way to implement this, is each time we get a new compile request,  we make a new timer, but that would require making a new thread for each request, which is slow).

torch/_inductor/compile_worker/timer.py
```diff
@@ -0,0 +1,53 @@
+from threading import Lock, Thread
+from time import sleep, time
+from typing import Any, Optional, Union
+
+
+class Timer:
+    """
+    This measures how long we have gone since last receiving a event and if it is grater than a set interval, calls a function.
```
masnesral: nits:
a event -> an event
grater -> greater

torch/_inductor/compile_worker/timer.py
```diff
@@ -0,0 +1,53 @@
...
+    """
+
+    def __init__(
+        self,
+        duration: Union[int, float],  # Duration in seconds
+        call: Any,  # Function to call when we expire
```
masnesral: Can we make this a better type? At least a callable?

torch/_inductor/compile_worker/timer.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def check(self) -> None:
+        while True:
+            # We have to be sensitive on checking here, to avoid too much impact on cpu
+            sleep(self.sleep_time)
+            with self.lock:
+                if self.exit:
+                    return
+                assert self.last_called is not None
+                if self.last_called + self.duration >= time():
```
masnesral: `time.monotonic()`?

torch/_inductor/compile_worker/timer.py
```diff
@@ -0,0 +1,53 @@
...
+    ) -> None:
+        # We don't start the background thread until we actually get an event.
+        self.background_thread: Optional[Thread] = None
+        self.last_called: Optional[float] = None
+        self.duration = duration
+        self.sleep_time = 60
+        self.call = call
+        self.exit = False
+
+        # Technically GIL should ensure only one call, but let's be explicit here
```
masnesral: Does it? I thought the GIL would only guarantee simple assignment is atomic and a lock would be required.

torch/_inductor/compile_worker/timer.py
```diff
@@ -0,0 +1,53 @@
...
+        self.lock = Lock()
+
+    def record_call(self) -> None:
+        with self.lock:
+            if self.background_thread is None:
+                self.background_thread = Thread(target=self.check)
+                self.background_thread.start()
+            self.last_called = time()
```
masnesral: the internet told me to use time.monotonic() because time() can jump forward or backward. I didn't know about time.monotonic(), but maybe that's more robust?

torch/_inductor/compile_worker/timer.py
```diff
@@ -0,0 +1,53 @@
...
+        self.call = call
+        self.exit = False
+
+        # Technically GIL should ensure only one call, but let's be explicit here
+        self.lock = Lock()
+
+    def record_call(self) -> None:
+        with self.lock:
+            if self.background_thread is None:
+                self.background_thread = Thread(target=self.check)
```
masnesral: Should this be a daemon thread? We don't have cancel functionality in this timer, so I think we could block the main process after exit until the thread wakes up and dies?

masnesral (changes requested):
minor question about daemon thread and timer usage

torch/_inductor/compile_worker/timer.py
c00w: Done.

torch/_inductor/compile_worker/timer.py
c00w: Technically I think it only swaps on I/O or calls to c libraries that release it. Either way just deleted the comment.

torch/_inductor/compile_worker/timer.py
c00w: Probably fine, switched to monotonic since I like it more.

torch/_inductor/compile_worker/timer.py
c00w: Good catch, fixed, also gave it a decent name.

torch/_inductor/compile_worker/timer.py
```diff
@@ -0,0 +1,54 @@
...
+    def __init__(
+        self,
+        duration: Union[int, float],  # Duration in seconds
+        call: Callable[[], None],  # Function to call when we expire
+    ) -> None:
+        # We don't start the background thread until we actually get an event.
+        self.background_thread: Optional[Thread] = None
+        self.last_called: Optional[float] = None
+        self.duration = duration
+        self.sleep_time = 60
```
masnesral, line 20: Do you think it would make sense to set the sleep time according to the duration time instead of hard-coding to 60?

masnesral (approved):

PR #166467, opened 2025-10-28
subproc_pool: Add support for enabling quiesce via a timer

c00w:
* #166584
* #166581
* __->__ #166467

This adds the capability to subproc pool to enable quiesce via a timer


test/inductor/test_compile_worker.py
```diff
@@ -75,13 +79,17 @@ def test_logging(self):
...
             os.environ["TORCHINDUCTOR_WORKER_LOGPATH"] = temp_log.name
-            pool = SubprocPool(2)
+            pool = self.make_pool(2)
             try:
                 pool.submit(operator.add, 100, 1)
                 self.assertEqual(os.path.exists(temp_log.name), True)
             finally:
                 pool.shutdown()
 
+class TestCompileWorkerWithTimer(TestCase):
```
masnesral: I guess I don't know what this does. Did you mean to derive from `TestCompileWorker`?

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -130,6 +131,7 @@ def __init__(
         nprocs: int,
         pickler: Optional[SubprocPickler] = None,
         kind: SubprocKind = SubprocKind.FORK,
+        quiesce = False,
```
masnesral: fix lint?

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -197,6 +199,11 @@ def __init__(
 
         self.running = True
 
+        if quiesce:
+            self.timer = Timer(60, self.quiesce)
```
masnesral: Can we make this duration configurable, i.e., in _inductor/config.py. I kinda think we'll want to set it to 2 or 3 min for some internal usage.

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -197,6 +199,11 @@ def __init__(
...
+        else:
+            self.timer = None
```
masnesral, line 225: fix lint?

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -255,6 +262,8 @@ def _read_thread(self) -> None:
             with self.futures_lock:
                 if not self.running:
                     return
+                if self.timer:
+                    timer.record_call()
```
masnesral: `self.timer`

torch/_inductor/compile_worker/subproc_pool.py
masnesral: I think we're going to have to make quiesce() a little better before we can just call it directly from the timer callback. Two things I'm thinking about, maybe related:
* We didn't have to worry about races before because the queisce call would only come after compile was finished. But now the quiesce() call could come any time, right? I think there might be a problem where we shut down the pool when something is running?
* Compile can take a long time in principle, so we might surpass the threshold and call quiesce while jobs are still running. So I think it's not enough to rely on the timer expiration alone - we also somehow need to check if there are jobs running.

masnesral (changes requested):

torch/_inductor/compile_worker/subproc_pool.py
masnesral: Maybe the timer API should be that the callback returns a bool where if False, the timer sleeps and tries again. The quiesce() needs to check if there are outstanding jobs.

test/inductor/test_compile_worker.py
c00w: Yes sorry, fixed.

torch/_inductor/compile_worker/subproc_pool.py
c00w: Done.

torch/_inductor/compile_worker/subproc_pool.py
c00w: Done.

torch/_inductor/compile_worker/subproc_pool.py
c00w: done.

torch/_inductor/compile_worker/subproc_pool.py
c00w: Done.

c00w: I've added this in, if we get a quiesce with pending futures, we record a call, and then ignore it. I'll add some tests to confirm the behaviour here.

test/inductor/test_compile_worker.py
```diff
@@ -75,14 +78,19 @@ def test_logging(self):
...
+            pool = self.make_pool(2)
             try:
                 pool.submit(operator.add, 100, 1)
                 self.assertEqual(os.path.exists(temp_log.name), True)
             finally:
                 pool.shutdown()
 
 
+class TestCompileWorkerWithTimer(TestCompileWorker):
+    def make_pool(self, size):
```
masnesral, line 92: I wonder if we'd get a little better coverage by also setting the quiesce() timeout at a low value for all of these tests?

torch/_inductor/config.py
```diff
@@ -954,6 +954,10 @@ def decide_compile_threads() -> int:
     default=False,
 )
 
+quiesce_async_compile_time: int = Config(
```
masnesral, line 964: nit: comment that this is in seconds?

torch/_inductor/compile_worker/subproc_pool.py
```diff
@@ -306,6 +317,9 @@ def _read_thread(self) -> None:
                 del self.pending_futures[job_id]
 
     def quiesce(self) -> None:
+        if self.timer is not None and len(self.pending_futures) > 0:
+            self.timer.record_call()
+            return
```
masnesral: I feel like we still need a lock to protect when we're killing the pool at the same time we're trying to use it: https://github.com/pytorch/pytorch/blob/dfebdcab86acbaa0eaa996b47595e5f27a66492e/torch/_inductor/compile_worker/subproc_pool.py#L373
Maybe acquiring the write_lock in that method is enough. It looks like shutdown calls quiesce() and also grabs the lock, but seems to release it first.

masnesral (changes requested):
I'm worried that there's a possible race between killing the internal pool and trying to start a new job. Can you take a look at my comment there?

torch/_inductor/compile_worker/subproc_pool.py
masnesral: Looking closer, I don't think it's enough to grab the lock only there because the code to check and re-initialize the pool doesn't grab the lock. But hopefully you know what I mean

torch/_inductor/config.py
c00w: Done.

test/inductor/test_compile_worker.py
c00w: Done.

torch/_inductor/compile_worker/subproc_pool.py
c00w: Chatted offline, the shutdown(wait=false) call will actually wait for pending futures to finish before it shutsdown. This should resolve any issues with quiesce locking (with the only side effect that 1) if we're spamming quiesce we'll eat startup costs, and 2) we may end up with more threads / processes for a bit if we're flapping).

https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown

torch/_inductor/compile_worker/subproc_pool.py
c00w: Repeating what was mentioned below - TrackedProcessPoolExecutor, inherits from ProcessPoolExecutor, and the shutdown call, retains the processes that are in use, until they finish their pending futures.

torch/_inductor/config.py
```diff
@@ -960,6 +960,11 @@ def decide_compile_threads() -> int:
     default=False,
 )
 
+# Time in seconds to wait before quiescing
+quiesce_async_compile_time: int = Config(
+    default=60,
```
masnesral, line 965: How do you feel about making this longer for internal? If is_fbcode(), 2 or 3 min? Or do you want trainers to decide that themselves?

masnesral (approved):

torch/_inductor/config.py
c00w: I'll stack a diff to make this a config item.

PR #166470, opened 2025-10-28
[ROCm][inductor] More configs for pointwise kernels.

naromero77amd:
This config improves performance by 250% on some kernels that contain `t1.atomic_add(...)`. Again, we conditionalize for ROCm/HIP, so there is no impact to NV.

cc @jeffdaily @sunway513 @jithunnair-amd @pruthvistony @ROCmSupport @dllehr-amd @jataylo @hongxiayang @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @mlazos

### DIFF
--- a/torch/_inductor/runtime/triton_heuristics.py
+++ b/torch/_inductor/runtime/triton_heuristics.py
@@ -2621,6 +2621,12 @@ def pointwise(
                             num_stages=2,
                             waves_per_eu=1,  # 20% improvement
                         ),
+                        triton_config_with_settings(
+                            size_hints,
+                            64,
+                            num_warps=1,
+                            num_stages=1,  # 250% improvement
+                        ),
                     ]
                 )
     if len(size_hints) == 2:
--- a/torch/_inductor/runtime/triton_heuristics.py
+++ b/torch/_inductor/runtime/triton_heuristics.py
@@ -2621,14 +2621,19 @@ def pointwise(
                             num_stages=2,
                             waves_per_eu=1,  # 20% improvement
                         ),
-                        triton_config_with_settings(
-                            size_hints,
-                            64,
-                            num_warps=1,
-                            num_stages=1,  # 250% improvement
-                        ),
                     ]
                 )
+                if inductor_meta.get("atomic_add_found"):
+                    configs.extend(
+                        [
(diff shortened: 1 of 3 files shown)


PaulZhang12 (approved):
LGTM! Seems like another one of those layout issues with num_warps=1....

torch/_inductor/runtime/triton_heuristics.py
```diff
@@ -2621,6 +2621,12 @@ def pointwise(
                             num_stages=2,
                             waves_per_eu=1,  # 20% improvement
                         ),
+                        triton_config_with_settings(
```
eellison: Can we conditionalize this if atomic add is actually present ?

mlazos (approved):

eellison (commented):
Hmm, we really need a repo, where we can check in kernels that are used for any of benchmark, and make sure we capture, dont regress, and can optimize in future. tritonbench sort of covers this but might be too general.

torch/_inductor/runtime/triton_heuristics.py
naromero77amd: I don't see how we can do this.

@jataylo any ideas?

eellison: See, `num_stores` in heuristics: https://github.com/pytorch/pytorch/blob/e0604d3170f1123ebacbced66be400c7f162621f/torch/_inductor/codegen/triton.py#L5052

naromero77amd: At the moment, `t1.atomic_add` are counted as `stores`. So I could not distinguish kernel with one atomic_add vs one store. So I had to add another field into this `inductor_meta` structure.

eellison: Sounds good - was just pointing to where we do similar analysis. makes sense you need new field.

eellison (approved):

jansel (approved):

Issue #146108, opened 2025-01-31
[associative_scan] Support lifted arguments for inductor


bohnstingl:
### 🚀 The feature, motivation and pitch

Currently we support lifted arguments for `associative_scan` only in dynamo, but not in inductor. This issue is a follow-up on https://github.com/pytorch/pytorch/pull/140043

cc @ydwu4

bohnstingl:
The lifted arguments have landed for the front-end and dynamo. https://github.com/pytorch/pytorch/pull/140043

PR #166474, opened 2025-10-29
[associative_scan] Support for lifted parameters in combine_mode='pointwise'

bohnstingl:
This PR introduces support for lifted parameters to `associative_scan` in the `combine_mode='pointwise'`. In particular, it adds the functionality to the `tl.associative_scan` by wrapping the `combine_fn` with the additional arguments and returning them directly. For example, a pseudo code of this approach is:
```
def combine_fn(x, y, add_inp):
  return x + y + add_inp

def combine_fn_wrapper(x, add_inpx, y, add_inpy):
  return combine_fn(x, y, add_inp), add_inpx

res = tl.associative_scan(combine_fn_wrapper, (x, add_inp), ...)
```

This PR addresses the issue https://github.com/pytorch/pytorch/issues/146108.

Note, in the current implementation, the additional arguments are broadcasted to the same shapes as the inputs `xs`, because for tl.associative_scan all input tensors need to be of the same shape. The current handling for this does not feel right, may cause functional issues and needs to be improved.

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @ipiszy @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @jataylo @chenyang78 @ydwu4 @zou3519


torch/_higher_order_ops/associative_scan.py
```diff
@@ -94,7 +94,18 @@ def __call__(self, combine_fn, xs, additional_inputs):
...
+            # Try to broadcast the shapes of the additional arguments to the inputs
+            # This is necessary as the triton kernel for the lowering during
+            # in combine_mode='pointwise', requires already broadcasted shapes
+            additional_inputsb = torch.broadcast_tensors(*xs, *additional_inputs)[
+                len(xs) :
+            ]
+            additional_inputsb = [add_inpsb[0, :] for add_inpsb in additional_inputsb]
+        except:  # noqa: E722
+            additional_inputsb = additional_inputs
+        return super().__call__(combine_fn, xs, additional_inputsb)
```
bohnstingl: This is where and how the broadcasting of the lifted parameters is done currently.
It is only needed for the `tl.associative_scan` and if there are better places to do this, it would certainly make sense to move it.

PR #166483, opened 2025-10-29
[DTensor] support DTensor view (flatten/unflatten) with _StridedSharding

weifengpy:
DTensor view (flatten, unflatten) is the last piece to support nn.Linear(DTensor). this is needed by if we do global spmd. I have to split the PR into 2 because of 2000 lines hard limit

Refactor `propagate_shape_and_sharding` into a two-phase `_ViewShardingPropagator`

## Flatten: Shard → _StridedShard

Consider flattening [2, 4] → [8] with Shard(1) on a mesh of size 2. The DimMap rule is `Flatten(InputDim(0), InputDim(1))`
                                                                                                                                                                                                          
Phase 1 — analyze(): Shard(1) falls inside the Flatten range. Sharded dim's size (4) is divisible by the mesh size (2), marking `shard_allowed[1] = [True]` and records the dim mapping: input dims {0, 1} → output dim 0.
                                                                                                                                                                                                          
Phase 2 — rewrite_output_placements(): Shard(1) gets mapped to output dim 0, but Dim 1 is not the first dim in the Flatten (dim 0 is), so flattening introduces a _StridedShard. It computes `split_factor = product(local_shapes[0:1]) = 2` and outputs _StridedShard(dim=0, split_factor=2)

Note: The sharded dim does not need to be divisible by the mesh size **if it is the last dim in the flatten range**. For example, flatten(dims=(0,1)) on [2, 3] → [6] with Shard(1), mesh size 2: 3 % 2 != 0, it can still represented as _StridedShard(dim=0, split_factor=2)

## Unflatten: _StridedShard → Shard

Consider the reverse: unflattening [8] → [2, 4] with _StridedShard(0, split_factor=2) on mesh size 2. The rule is `(Split(InputDim(0), (2, 4), 0), Split(InputDim(0), (2, 4), 1))`
                                                                                                                                                                                                          
Phase 1 — analyze(): InputDim(0) has _StridedShard(0, split_factor=2). The split_factor matches split_id=1 (where prod(group_shape[:1]) = 2). Setting shard_allowed[0] is set to [True]. In general, non-last split dims require even divisibility, otherwise raising RuntimeError "must redistribute first"

Phase 2 — rewrite_output_placements(): resolve _StridedShard(split_factor=2) back to plain Shard. Output dim 0 (split_id=0) has product(group_shape[:0]) = 1 — doesn't match 2. Output dim 1 (split_id=1) has product(group_shape[:1]) = 2 — match. The unflatten resolves _StridedShard into Shard(dim=1). If cannot match, throw runtime error "split_factor does not match any output dimension"

## Callstack
<img width="761" height="457" alt="Screenshot 2026-03-04 at 01 30 05" src="https://github.com/user-attachments/assets/e817f9f6-914f-400a-96ed-022e755bb3e6" />

* #178454
* #176151
* __->__ #166483



test/distributed/tensor/test_view_ops.py
```diff
@@ -670,6 +758,7 @@ def test_squeeze_(self):
     skipped_tests=[
         # Comparing data pointers is not supported for local tensor
         "test_dtensor_view_op_uneven",
+        "test_dtensor_flatten",
```
wconstab: why is this skipped?  the test case looked ok to me, so i would hope it passes!

torch/distributed/tensor/_ops/_matrix_ops.py
```diff
@@ -93,6 +94,9 @@ def _mm_like_strategy(
                 generate_redistribute_costs(mat2_strategy, mat2_spec),
             ]
             strtg.redistribute_cost = redistribute_cost
+            if len(self_strategy.strategies) == 1 and len(self_strategy.strategies[0].output_specs.placements) == 1 and len(self_spec.placements) == 1 and self_spec.placements[0].is_shard() and isinstance(self_strategy.strategies[0].output_specs.placements[0], _StridedShard) and self_spec.placements[0].dim == self_strategy.strategies[0].output_specs.placements[0].dim and strtg.output_specs == strtg.input_specs[0]:
```
wconstab: at least, make this more readable please.  now i'm trying to figure out what it actually means..

test/distributed/tensor/test_view_ops.py
weifengpy: no need to skip. this is for LocalTensor. will add it back

torch/distributed/tensor/_ops/_matrix_ops.py
weifengpy: yeah, it's so long. just trying to check they only have single placement and they are sharded on the same dim. will make it more readable

torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -519,183 +524,633 @@ def propagate_shape_and_sharding(
...
     Sharding propagation follows mapped dimensions:
     - An output dimension that maps directly to an input dimension is sharded equally
-    - An output dimension that is a flattened set of input dimensions can only be
-      sharded if only the leftmost flattened dimension is sharded.
+    - An output dimension that is a flattened set of input dimensions can be sharded:
+      the first sharded dim stays as Shard, non-first sharded dims become _StridedShard
     - An output dimension that is a split of the input dimension can only be sharded
       if the leftmost split size is divisible by the mesh dimension
     """
-    if not len(input_src_placements) == len(mesh_sizes):
```
weifengpy, line 566: claude-checked the assertion is redundant.

```
 It's enforced at DTensor construction boundaries:
  - _api.py:877 — distribute_tensor checks len(placements) != device_mesh.ndim
  - _api.py:1179 — DTensor init checks device_mesh.ndim != len(placements)
  - _redistribute.py:745 — redistribute checks len(src_placement) != mesh.ndim

By the time we reach propagate_shape_and_sharding, the placements come from input_src_spec.placements (a DTensorSpec) and mesh.shape — both derived from the same DeviceMesh. The DTensorSpec is constructed with that mesh, and all upstream code validates the invariant. So yes, len(placements) == len(mesh_sizes) is already guaranteed by the call stack
```

test/distributed/tensor/test_view_ops.py
```diff
@@ -787,13 +1413,5 @@ def test_storage_offset_shard_dim1_slice_dim0(self):
         self.assertEqual(sliced_dtensor.full_tensor(), expected)
 
 
-TestViewOpsWithLocalTensor = create_local_tensor_test_class(
```
weifengpy, line 1108: will add this back. waiting for CI now to make sure we are not timing out already

wconstab:
claude review issues 1, 2 look worth fixing to me.  3 I don't understand yet, and 5 seems like a good idea if its not hard to include, but probably ok to keep as a TODO.

test/distributed/tensor/test_view_ops.py
```diff
@@ -660,7 +658,656 @@ def test_dtensor_view_op_uneven(self):
...
+                                    tensor_dims,
+                                    flatten_start,
+                                    flatten_end,
+                                    mesh,
+                                    placements,
+                                )
+
+                    # Replicate
+                    if self.is_local_tensor_enabled:
+                        all_dims = [(even,) * tensor_ndim]
```
wconstab: hm, is it better to make localtensor just skip the test case rather than change the test case to be a different case?

torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -558,186 +559,663 @@ def propagate_shape_and_sharding(
...
+                else:
+                    input_to_output_tensor_dims[in_dim.input_dim].append(output_dim)
+            elif isinstance(cmd, Split):
+                # Split(split_id>0): _analyze_split returned [], so chase the
+                # root input dim and append this output dim to its existing entry.
+                #
+                # Flatten+Split example: view([2, 3], [3, 2])
+                #   rule = (Split(Flatten(InputDim(0), InputDim(1)), (3,2), 0),
+                #           Split(Flatten(InputDim(0), InputDim(1)), (3,2), 1))
+                #   output_dim=0 (split_id=0): same as Split example above.
```
wconstab, line 682: why in these cases do we automatically look at the left dim from the Flatten?

torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -558,186 +559,663 @@ def propagate_shape_and_sharding(
...
+        self,
+    ) -> tuple[Sequence[Placement], dict[int, list[int]]]:
+        """Phase 1: walk the DimMap rule, return (input_tgt_placements, input_to_output_tensor_dims)."""
+        input_dims_in_rule = self._input_dims_in_rule(self.rule)
+
+        # Default: shardable if the dim appears in the rule. Refined by _analyze_*.
+        for dim in range(len(self.global_input_shape)):
+            self.shard_allowed[dim] = [dim in input_dims_in_rule] * self.mesh_ndim
+
+        # Walk the rule to fill shard_allowed and build input_to_output_tensor_dims.
```
wconstab: except we don't actually fill out shard_allowed as we're walking the rule?
edit: nvm, i guess you refer to the `self._input_dims_in_rule(self.rule)` above- probably this comment should be removed

torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -558,186 +559,663 @@ def propagate_shape_and_sharding(
...
+                input_tgt_placements.append(p)
+        return input_tgt_placements, input_to_output_tensor_dims
+
+    def rewrite_output_placements(
+        self,
+        input_tgt_placements: Sequence[Placement],
+        input_to_output_tensor_dims: dict[int, list[int]],
+    ) -> list[Placement]:
+        """Phase 2: consume analyze() outputs, return final output placements."""
+        # Output dims already assigned to a mesh dim by _StridedShard rewriting.
```
wconstab: confused by this comment, i didn't see StridedShard rewriting happen in phase1 (analyze)

wconstab (commented):
overall PR structure and tests look good, and i reviewed the top level 2 phase code. still need to review the analyze/rewrite helpers which are quite significant. will keep reviewing tmrw

torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -558,189 +559,662 @@ def propagate_shape_and_sharding(
...
+        Unlike _rewrite_strided_shard, this does NOT update
+        claimed_output_dims.  When multiple mesh dims have plain Shard on the
+        same input dim (e.g. [Shard(0), Shard(0)]), they legitimately
+        target the same output dim; coordination is handled by
+        the analysis phase's shard_allowed check which validates divisibility.
+        """
+        tgt_shard_dims = [
+            d
+            for d in input_to_output_tensor_dims[p.dim]
+            if (p.dim, d) not in claimed_output_dims
```
wconstab: hmm. -  p.dim not in claimed_output_dims... ?

torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -558,189 +559,662 @@ def propagate_shape_and_sharding(
...
+                ):
+                    return candidate_dim
+        return None
+
+    def _rewrite_plain_shard(
+        self,
+        p: Shard,
+        mesh_dim: int,
+        placements: Sequence[Placement],
+        claimed_output_dims: set[tuple[int, int]],
```
wconstab: hmm, this is a pairing of input to output dim, or what? needs a better docstring

wconstab: make this a set of NamedTuples or Dataclasses or something

torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -558,189 +559,662 @@ def propagate_shape_and_sharding(
...
+        local_tensor_shapes: list[int],
+        input_to_output_tensor_dims: dict[int, list[int]],
+    ) -> Placement:
```
wconstab: Given a plain Shard(dim=X) input placement on a specific mesh dim, determine what output placement it maps to after the view op

test/distributed/tensor/test_view_ops.py
weifengpy: that might be clearer indeed. I skipped the test for local tensor mode

torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -558,189 +559,662 @@ def propagate_shape_and_sharding(
...
+        a non-first dim produces _StridedShard — this is where _StridedShard
+        placements originate (consumed later by _rewrite_strided_shard).
+
+        Unlike _rewrite_strided_shard, this does NOT update
+        claimed_output_dims.  When multiple mesh dims have plain Shard on the
+        same input dim (e.g. [Shard(0), Shard(0)]), they legitimately
+        target the same output dim; coordination is handled by
+        the analysis phase's shard_allowed check which validates divisibility.
+        """
+        tgt_shard_dims = [
```
wconstab, line 1074: suggested comment:

looks up input_to_output_tensor_dims[p.dim] — the output dims that input dim p.dim maps to — and filters out any already claimed by _StridedShard rewriting on earlier mesh dims

Also, `claimed_output_dims` is a bit generic. is the entire purpose of this set to make sure StridedShard claimed dims can't get reused by Shard later?  Should we rename this to stridedshard_output_dims instead?

torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -558,189 +559,662 @@ def propagate_shape_and_sharding(
...
+        ]
+        if len(tgt_shard_dims) == 0:
+            raise AssertionError(
+                f"No output dim available for Shard(dim={p.dim}) on mesh dim "
+                f"{mesh_dim}. All output dims already claimed by earlier mesh dims."
+            )
+        if len(tgt_shard_dims) == 1:
+            tgt_shard_dim = tgt_shard_dims[0]
+        else:
+            tgt_shard_dim = next(
```
wconstab, line 1090: wondering if this next thing is too general: do we have cases where there are multiple Splits to consider, or can we just assert that tgt_shard_dims is len 2 here and refers to a split and grab its .split_id 0?  (that way seems both more strict and more readable IMO, but not sure if this extra flexibility is useful)

torch/distributed/tensor/_ops/_view_ops.py
weifengpy: The outer rule is a Split. both output dims (3 and 2) come from the same flattened value 6 (= 2 * 3), so they should be grouped under one key in `input_to_output_tensor_dims`. We pick `InputDim(0)` because `_analyze_flatten` always returns the first input dim as its first element — even when multiple dims are sharded, `_analyze_split` takes in_dims[0].

torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -558,189 +559,662 @@ def propagate_shape_and_sharding(
...
+            if tgt_shard_dim is None:
+                raise AssertionError(
+                    f"No Split(split_id=0) found among unclaimed output dims "
+                    f"{tgt_shard_dims} for Shard(dim={p.dim}) on mesh dim {mesh_dim}."
+                )
+        cmd = self.rule[tgt_shard_dim]
+        if isinstance(cmd, (Split, InputDim)):
+            # Split/InputDim: 1:1 dim mapping, sharding transfers directly.
+            # Flatten needs stride computation below (multiple dims merge).
+            local_tensor_shapes[p.dim] = (
```
wconstab: nit: slightly prefer returning new copy of local_tensor_shapes rather than mutating. And also don't love having duplicate sites where we do the new local_tensor_shape computation. not sure if theres an easy way to do it with a single return site though.  The function looks correct to me.

torch/distributed/tensor/_ops/_view_ops.py
weifengpy: Good catch. This is actually refining shard_allowed instead of filling/initializing it. I updated the comment.

torch/distributed/tensor/_ops/_view_ops.py
weifengpy: The "already" refers to earlier mesh dims in the same Phase 2 loop. I updated the comment to avoid confusion

test/distributed/tensor/test_view_ops.py
```diff
@@ -264,16 +265,18 @@ def test_double_shard_split_validation(self):
...
+        # out_size=3 is not divisible by mesh dim 0 size=2 → error
+        with self.assertRaisesRegex(
+            RuntimeError, "not evenly divisible by mesh dimension"
+        ):
             propagate_shape_and_sharding(
                 [Shard(0), Shard(0)],
                 (12,),
                 dim_maps[torch.Tensor.view](torch.empty(12), [3, 4]),
                 (2, 3),
+                strict_view=True,
```
weifengpy, line 277: @stmcgovern @wconstab I believe what https://github.com/pytorch/pytorch/pull/177973 wants to cover is strict_view=True . Otherwise strict_view=False allows redistributing input [S, S] to [R, S]

torch/distributed/tensor/_ops/_view_ops.py
```diff
@@ -558,189 +559,662 @@ def propagate_shape_and_sharding(
...
+        placement as-is.
+        """
+        tgt_shard_dims = [
+            d
+            for d in input_to_output_tensor_dims[p.dim]
+            if (p.dim, d) not in claimed_output_dims
+        ]
+        # Phase 1: resolve SS → Shard.  If an output dim's Split has a
+        # group_shape prefix matching the split_factor, the strided pattern
+        # is fully captured by the Split, so SS simplifies to Shard.
```
wconstab, line 1176: example: unflattening a tensor with shape [6,4] into [2,3,4], with original placement SS(0, sf=2) on mesh shape [3] gives output placement Shard(1) 

The original SS sf=2 means there are 2 groups of contiguous data within the outer dim of 6.  We find out whether a trivial shard 'fits exatly' into this pattern, turns out S(1) does (and we know because prod(2) == sf) where 2 is group_shape[:1] and 1 is the new Shard dim.

wconstab (approved):
ok, looks good to me! made some comments mostly about code style or adding comments. Please address the ones that make sense

test/distributed/tensor/test_view_ops.py
stmcgovern: @weifengpy yes we see that here where the redistribution causes a sync bug for squeeze. https://github.com/pytorch/pytorch/pull/175798#discussion_r2961340758

torch/distributed/tensor/_ops/_view_ops.py
weifengpy: one example is, Unflatten (24,) → (4, 6) on 2D mesh (4, 6) with [_StridedShard(0, sf=1), _StridedShard(0, sf=4)] 

p.dim = 0 appears twice. We need (p.dim, d) to support 2 mesh both sharding on dim0                                                                                                                                
1. Mesh dim 0: _rewrite_strided_shard for _StridedShard(0, sf=1) — sf=1 matches prod(group_shape[:0]) = 1, resolves to output dim 0. Adds (0, 0) to claimed_output_dims.                                                                                                                                                  
2. Mesh dim 1: _rewrite_strided_shard for _StridedShard(0, sf=4) — sf=4 matches prod(group_shape[:1]) = 4, resolves to output dim 1. Adds (0, 1).                   
3. Result: [Shard(0), Shard(1)] — correct

torch/distributed/tensor/_ops/_view_ops.py
weifengpy: good example! I added it to the comment

torch/distributed/tensor/_ops/_view_ops.py
weifengpy: never thought of inplace muation vs new copy. good call out. I updated the PR to use copy appaorach. the return signature is difffernt though, it becomes "output placement, new local_tensor_shapes"

torch/distributed/tensor/_ops/_view_ops.py
weifengpy: good question. I updated the code comment on the need to consdier multiple Splits

torch/distributed/tensor/_ops/_view_ops.py
weifengpy: I see why `claimed_output_dims` is confusing now. Updated the PR to use NamedTuples

torch/distributed/tensor/_ops/_view_ops.py
weifengpy: renamed to strided_shard_claimed_dims and let it be NamedTuple

test/distributed/tensor/test_view_ops.py
weifengpy: got you. I was just making sure adding strict_view=True is correct

PR #166497, opened 2025-10-29
Enable verify_dynamo on Python 3.13

cyyever:
Dynamo now supports Python 3.13.

### DIFF
--- a/tools/dynamo/verify_dynamo.py
+++ b/tools/dynamo/verify_dynamo.py
@@ -216,8 +216,8 @@ def main() -> None:
         f"ROCM version: {rocm_ver}\n"
     )
     for args in _SANITY_CHECK_ARGS:
-        if sys.version_info >= (3, 13):
-            warnings.warn("Dynamo not yet supported in Python 3.13. Skipping check.")
+        if sys.version_info >= (3, 14):
+            warnings.warn("Dynamo not yet supported in Python 3.14. Skipping check.")
             continue
         check_dynamo(*args)
     print("All required checks passed")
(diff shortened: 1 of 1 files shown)


tools/dynamo/verify_dynamo.py
```diff
@@ -216,9 +216,6 @@ def main() -> None:
         f"ROCM version: {rocm_ver}\n"
     )
     for args in _SANITY_CHECK_ARGS:
-        if sys.version_info >= (3, 13):
```
Lucaskabela, line 219: shouldn't we just bump this version to `3.14` (next unsupported version) instead of remove?

Lucaskabela (changes requested):
I think we should simply edit the version as opposed to completely removing

tools/dynamo/verify_dynamo.py
cyyever: Yes, that is better.

williamwen42 (changes requested):
Can we remove the `continue` as well? We should probably crash if Dynamo indeed does not support 3.14 (this will also help to remind me to update this once 3.14 support is completed).

cyyever:
I don't think deliberately making it crash is suitable here...

williamwen42 (approved):
I still think we can remove the `continue` since this script is really only used for CI sanity checks, but stamping to unblock.

cyyever:
@williamwen42 Fine, let's remove it.

cyyever:
@Lucaskabela Have a look?

Lucaskabela (approved):
LGTM!

PR #166500, opened 2025-10-29
[2/N] Fix unused loop variables

cyyever:
This PR removes unused loop variables.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @ezyang @EikanWang @jgong5 @wenzhe-nrv @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela @xmfan

### DIFF
--- a/test/torch_np/numpy_tests/lib/test_function_base.py
+++ b/test/torch_np/numpy_tests/lib/test_function_base.py
@@ -2596,7 +2596,7 @@ class TestBincount(TestCase):
         intp_refcount = sys.getrefcount(np.dtype(np.intp))
         double_refcount = sys.getrefcount(np.dtype(np.double))
 
-        for _j in range(10):
+        for _ in range(10):
             np.bincount([1, 2, 3])
         assert_equal(sys.getrefcount(np.dtype(np.intp)), intp_refcount)
         assert_equal(sys.getrefcount(np.dtype(np.double)), double_refcount)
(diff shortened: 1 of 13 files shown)


test/torch_np/numpy_tests/lib/test_function_base.py
```diff
@@ -2596,12 +2596,12 @@ def test_dtype_reference_leaks(self):
         intp_refcount = sys.getrefcount(np.dtype(np.intp))
         double_refcount = sys.getrefcount(np.dtype(np.double))
 
-        for j in range(10):
+        for _j in range(10):
```
mlazos: Why is this different from the others?

torch/_dynamo/variables/iter.py
```diff
@@ -400,8 +400,9 @@ def get_item(it):
             else:
                 return it.next_variable(tx)
 
+        idx = -1
```
mlazos: I would say None is more appropriate here, but this is more of a nit

mlazos (approved):

test/torch_np/numpy_tests/lib/test_function_base.py
cyyever: Oh, I should use `_`

torch/_dynamo/variables/iter.py
cyyever: None is indeed better

PR #166504, opened 2025-10-29
[xpu][test] Enable int4 and int8 test on Intel GPU

xiaowangintel:
Enable int4 and int8 test to enhance testing on Intel GPU. 

### DIFF
--- a/torch/_meta_registrations.py
+++ b/torch/_meta_registrations.py
@@ -3695,7 +3695,7 @@ def meta__convert_weight_to_int4pack_for_cpu(w, inner_k_tiles):
 @register_meta([aten._weight_int4pack_mm])
 def meta__weight_int4pack_mm(x, w, q_group_size, q_scale_and_zeros):
     torch._check(x.dim() == 2, lambda: "x must be a 2D tensor")
-    expected_dim = 2 if "xpu" in w.fake_device.type else 4
+    expected_dim = 2 if w.fake_device.type == "xpu" else 4
     torch._check(w.dim() == expected_dim, lambda: f"w must be a {expected_dim}D tensor")
     torch._check(
         x.dtype in [torch.float32, torch.float16, torch.bfloat16],
@@ -3705,7 +3705,7 @@ def meta__weight_int4pack_mm(x, w, q_group_size, q_scale_and_zeros):
         w.dtype is torch.int32,
         lambda: f"expected w to be int32, got {w.dtype}",
     )
-    dim_n = w.size(0) if "xpu" in w.fake_device.type else w.size(0) * 8
+    dim_n = w.size(0) if w.fake_device.type == "xpu" else w.size(0) * 8
     return x.new_empty(x.size(0), dim_n, dtype=x.dtype)
 
 
--- a/torch/_meta_registrations.py
+++ b/torch/_meta_registrations.py
@@ -3695,7 +3695,7 @@ def meta__convert_weight_to_int4pack_for_cpu(w, inner_k_tiles):
 @register_meta([aten._weight_int4pack_mm])
 def meta__weight_int4pack_mm(x, w, q_group_size, q_scale_and_zeros):
     torch._check(x.dim() == 2, lambda: "x must be a 2D tensor")
-    torch._check(w.dim() == 4, lambda: "w must be a 4D tensor")
+    torch._check(w.dim() == 4 or w.dim() == 2, lambda: "w must be a 4D or 2D tensor")
     torch._check(
         x.dtype in [torch.float32, torch.float16, torch.bfloat16],
         lambda: f"expected x to be f32/f16/bf16, got {x.dtype}",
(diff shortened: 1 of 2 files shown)


torch/_meta_registrations.py
```diff
@@ -3679,7 +3682,10 @@ def meta__weight_int4pack_mm(x, w, q_group_size, q_scale_and_zeros):
         w.dtype is torch.int32,
         lambda: f"expected w to be int32, got {w.dtype}",
     )
-    return x.new_empty(x.size(0), w.size(0) * 8, dtype=x.dtype)
+    if w.dim() == 4:
+        return x.new_empty(x.size(0), w.size(0) * 8, dtype=x.dtype)
+    elif w.dim() == 2:
```
etaf: In my person, It's better to set the condition as if xpu.

xiaowangintel: I don't think so, the output shape is directly related to the weight shape or dim, not the device.

liangan1: > In my person, It's better to set the condition as if xpu.

it is meta device in this stage, whether there is any information to judge it is xpu now?

etaf: w.fake_device.type:

xiaowangintel: https://github.com/pytorch/pytorch/pull/166504/files#diff-6dffed1ade0ba3e887f9a4eafa3bfcec267ab2365b8adcb91bd391f49b3fd2e3R3673
@liangan1 Open this link, we can use w.fake_device to get original device type.

etaf: > I don't think so, the output shape is directly related to the weight shape or dim, not the device.

Maybe you are right, it just not easy to see the mapping between dim 4: w.size(0) * 8, and  dim 2: w.size(0) for me.

EikanWang: > I don't think so, the output shape is directly related to the weight shape or dim, not the device.

Then my question will be why one dim of the output tensor should be `w.size(0) * 8` if `w.dim()` is `4`. Is it not due to the detailed implementation for different devices?

torch/_meta_registrations.py
```diff
@@ -3670,7 +3670,10 @@ def meta__convert_weight_to_int4pack_for_cpu(w, inner_k_tiles):
 @register_meta([aten._weight_int4pack_mm])
 def meta__weight_int4pack_mm(x, w, q_group_size, q_scale_and_zeros):
     torch._check(x.dim() == 2, lambda: "x must be a 2D tensor")
-    torch._check(w.dim() == 4, lambda: "w must be a 4D tensor")
+    if "xpu" in w.fake_device.type:
+        torch._check(w.dim() == 2, lambda: "w must be a 2D tensor on XPU")
+    else:
+        torch._check(w.dim() == 4, lambda: "w must be a 4D tensor")
```
EikanWang:
```suggestion
    expected_dim = 2 if "xpu" == w.fake_device.type else 4
    torch._check(w.dim() == expected_dim, lambda: f"w must be a {expected_dim}D tensor")
```

torch/_meta_registrations.py
```diff
@@ -3679,7 +3682,10 @@ def meta__weight_int4pack_mm(x, w, q_group_size, q_scale_and_zeros):
         w.dtype is torch.int32,
         lambda: f"expected w to be int32, got {w.dtype}",
     )
-    return x.new_empty(x.size(0), w.size(0) * 8, dtype=x.dtype)
+    if w.dim() == 4:
+        return x.new_empty(x.size(0), w.size(0) * 8, dtype=x.dtype)
+    elif w.dim() == 2:
+        return x.new_empty(x.size(0), w.size(0), dtype=x.dtype)
```
EikanWang:
```suggestion
    dim_n = w.size(0) if w.fake_device.type == "xpu" else w.size(0) * 8
    return x.new_empty(x.size(0), dim_n, dtype=x.dtype)
```

torch/_meta_registrations.py
xiaowangintel: Yes, It's. But, the other devices aligned to CUDA or XPU also need to modify this code if they want to support this interface. And if cuda need to aligned to xpu and versus, this code need to be modified, too.

torch/_meta_registrations.py
xiaowangintel: done

torch/_meta_registrations.py
xiaowangintel: done

torch/_meta_registrations.py
```diff
@@ -3670,7 +3670,8 @@ def meta__convert_weight_to_int4pack_for_cpu(w, inner_k_tiles):
 @register_meta([aten._weight_int4pack_mm])
 def meta__weight_int4pack_mm(x, w, q_group_size, q_scale_and_zeros):
     torch._check(x.dim() == 2, lambda: "x must be a 2D tensor")
-    torch._check(w.dim() == 4, lambda: "w must be a 4D tensor")
+    expected_dim = 2 if "xpu" in w.fake_device.type else 4
```
EikanWang: "xpu" in `w.fake_device.type` means the literal of the fake device type contains `xpu` rather than being `xpu`. It would be better to ensure a clear semantics here.
```suggestion
    expected_dim = 2 if "xpu" == w.fake_device.type else 4
```

torch/_meta_registrations.py
```diff
@@ -3679,7 +3680,8 @@ def meta__weight_int4pack_mm(x, w, q_group_size, q_scale_and_zeros):
         w.dtype is torch.int32,
         lambda: f"expected w to be int32, got {w.dtype}",
     )
-    return x.new_empty(x.size(0), w.size(0) * 8, dtype=x.dtype)
+    dim_n = w.size(0) if "xpu" in w.fake_device.type else w.size(0) * 8
```
EikanWang: ditto

EikanWang (approved):

jansel (approved):

PR #166509, opened 2025-10-29
[3/N] Fix unused loop variables

cyyever:
This PR removes unused loop variables in tests.

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @EikanWang @jgong5 @wenzhe-nrv @sanchitintel @ezyang @voznesenskym @penguinwu @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @jiayisunx @ipiszy @chenyang78 @kadeng @muchulee8 @amjames @chauhang @aakhundov @coconutruben @Lucaskabela @xmfan

### DIFF
--- a/test/distributed/_composable/fsdp/test_fully_shard_training.py
+++ b/test/distributed/_composable/fsdp/test_fully_shard_training.py
@@ -827,7 +827,7 @@ class TestFullyShardShardPlacementFnMultiProcess(FSDPTest):
 
         torch.manual_seed(42 + self.rank)
         inp = torch.randint(0, model_args.vocab_size, (2, 16), device=device_type.type)
-        for iter_idx in range(5):
+        for _ in range(5):
             ref_loss = ref_model(inp).sum()
             loss = model(inp).sum()
             self.assertEqual(ref_loss, loss)
--- a/test/dynamo/test_modules.py
+++ b/test/dynamo/test_modules.py
@@ -2575,7 +2575,7 @@ class OptimizedModuleTest(torch._dynamo.test_case.TestCase):
         def save_activations(mod, inp, out):
             activations.append(inp)
 
-        for _name, module in model.named_modules():
+        for module in model.modules():
             module.register_forward_hook(save_activations)
 
         cnt = torch._dynamo.testing.CompileCounter()
--- a/test/dynamo/test_modules.py
+++ b/test/dynamo/test_modules.py
@@ -262,7 +262,7 @@ class ConstLoop(torch.nn.Module):
         self.count = 3
 
     def forward(self, x):
-        for i in range(self.count):
+        for _ in range(self.count):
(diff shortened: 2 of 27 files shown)


torch/_dynamo/variables/iter.py
```diff
@@ -400,8 +400,9 @@ def get_item(it):
             else:
                 return it.next_variable(tx)
 
+        idx: int | None = None
```
Lucaskabela, line 423: This change seems very different than the rest - why are we adding this as opposed to removing idx? 

In the case we cannot remove idx, why aren't we just keeping this code the same?

torch/_dynamo/compiled_autograd.py
```diff
@@ -969,7 +969,8 @@ def dce(self) -> None:
 
         # Dynamo guards will error instead of creating aliasing guards unless we unpack them in the graph
         unpack_nodes: OrderedSet[torch.fx.Node] = OrderedSet()
-        for i, node in enumerate(self.fx_tracer.graph.find_nodes(op="placeholder")):
+        i: int | None = None
```
Lucaskabela, line 972: See above

torch/ao/quantization/fx/convert.py
```diff
@@ -297,6 +293,8 @@ def add_dequantize_op_kwargs(dequantize_op, input_node):
             zero_point_node = graph.create_node(
                 "call_function", operator.getitem, (choose_qparams_node, 1), {}
             )
+            # we have quant_min, quant_max and dtype, all should be stored
```
Lucaskabela, line 296: Seems outside the scope of changes

Lucaskabela (changes requested):
Few minor comments on changes made in the non test files

torch/ao/quantization/fx/convert.py
cyyever: Because the comment is not so related to the change location than this line.

torch/_dynamo/variables/iter.py
cyyever: Because `idx==0` is checked in the exception handling. We want to distinguish between an empty sequence and the first element using `idx=None`

torch/_dynamo/compiled_autograd.py
cyyever: The reason here is also to distinguish between an empty sequence and the first element, for a more robust assertion used below.

Lucaskabela (approved):
I have a similar thought to our other PR - namely is have we/will we enable B007 shortly after these PRs?

Other than that LGTM

Skylion007 (approved):

cyyever:
Our aim is finally enabling it after these fixes.

PR #166515, opened 2025-10-29
Fix issue with inductor pointwise kernels always autotuning

jataylo:
We noticed for pointwise examples currently we are always running into autotune path

Example
```
import torch
@torch.compile()
def add_seven(x):
    return x + 7
x = torch.randn(65534, device='cuda')
y = add_seven(x)
```

Logs:
```
V1029 11:30:39.839000 739 site-packages/torch/_inductor/runtime/triton_heuristics.py:315] [0/0] CachingAutotuner gets 5 configs for triton_poi_fused_add_0
V1029 11:30:39.839000 739 site-packages/torch/_inductor/runtime/triton_heuristics.py:321] [0/0] XBLOCK: 512, num_warps: 2, num_ctas: 1, num_stages: 1, maxnreg: None
V1029 11:30:39.839000 739 site-packages/torch/_inductor/runtime/triton_heuristics.py:321] [0/0] XBLOCK: 256, num_warps: 4, num_ctas: 1, num_stages: 1, maxnreg: None
V1029 11:30:39.839000 739 site-packages/torch/_inductor/runtime/triton_heuristics.py:321] [0/0] XBLOCK: 8192, waves_per_eu: 2, num_warps: 4, num_ctas: 1, num_stages: 1, maxnreg: None
V1029 11:30:39.839000 739 site-packages/torch/_inductor/runtime/triton_heuristics.py:321] [0/0] XBLOCK: 4096, num_warps: 4, num_ctas: 1, num_stages: 1, maxnreg: None
V1029 11:30:39.839000 739 site-packages/torch/_inductor/runtime/triton_heuristics.py:321] [0/0] XBLOCK: 2048, waves_per_eu: 1, num_warps: 8, num_ctas: 1, num_stages: 2, maxnreg: None
```

The root cause of this seems to be a regression from https://github.com/pytorch/pytorch/pull/164905 which introduced inductor_meta.get("autotune_pointwise") into the autotune logic.

This PR simplifies these autotune conditions slightly to ensure tuning is only performed when relevant env vars are set

### DIFF
--- a/test/inductor/test_max_autotune.py
+++ b/test/inductor/test_max_autotune.py
@@ -2835,7 +2835,7 @@ class TestMaxAutotuneRemoteCache(TestCase):
                 torch.compile(mm, dynamic=dynamic)(a, b)
                 reset()
             global_stats.report()
-            self.assertEqual(global_stats.autotune_remote, Stats(2, 3, 2))
+            self.assertEqual(global_stats.autotune_remote, Stats(0, 0, 0))
 
 
 class _TestTritonTemplateCaller(TritonTemplateCaller):
--- a/torch/_inductor/runtime/triton_heuristics.py
+++ b/torch/_inductor/runtime/triton_heuristics.py
@@ -2590,7 +2590,7 @@ def pointwise(
 
     configs = None
     if len(size_hints) == 1:
-        if not inductor_meta.get("autotune_pointwise", True) and not (
+        if not (
             inductor_meta.get("max_autotune")
             or inductor_meta.get("max_autotune_pointwise")
         ):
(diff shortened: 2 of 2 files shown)


jataylo:
> The cited PR #164905 would change pointwise autotuning only if you have the deterministic mode set by sth like `torch.use_deterministic_algorithms(True)`. Without setting this mode, the PR should not add more pointwise autotuning.
> 
> Also, why running into an autotune path is an issue here?

@shunting314 I would have assumed it would have been antithetical to the design of max_autotune and max_autotune_pointwise to be always autotuning and take a compilation time hit. If we want to autotune by default I'm fine to close this PR, in that case we should probably just remove the if/else entirely, but this seemed like a mistake to me cc: @eellison

jataylo:
General consensus I've gotten is that pointwise tuning time isn't a problem so we can leave as default, closing this PR.

PR #166521, opened 2025-10-29
[inductor] Collectives estimations: option to use nccl estimator for fx node

IvanKobzarev:
* __->__ #166521


torch/_inductor/comm_analysis.py
```diff
@@ -561,13 +555,52 @@ def estimate_nccl_collective_runtime_from_fx_node(
...
     coll = get_collective_type_from_kernel_name(fx_node.target.name())
 
+    def _nccl_estimate() -> Optional[float]:
+        # TODO: Refactor with estimate_nccl_collective_runtime_nccl_estimator
+        import torch.utils._pytree as pytree
+        from torch.distributed.distributed_c10d import _resolve_process_group
+
+        flat_args, flat_args_pytree_spec = pytree.tree_flatten((args, kwargs))
+
+        def _tensor(size, dtype, device) -> torch.Tensor:  # type: ignore[no-untyped-def]
```
Skylion007, line 403: Just type the args here or use a functools.partial if you want to be lazy

torch/_inductor/comm_analysis.py
```diff
@@ -561,13 +555,52 @@ def estimate_nccl_collective_runtime_from_fx_node(
...
+            return torch.empty(size, dtype=dtype, device=device)
+
+        def to_real_tensor(e: Any) -> Any:
```
Skylion007, line 413: Return type is a union of input TypeVar, a Tensor or a RealTensor? Or realistically, should always just return type of a real Tensor right?

torch/_inductor/comm_analysis.py
```diff
@@ -561,13 +555,52 @@ def estimate_nccl_collective_runtime_from_fx_node(
...
+            if isinstance(e, torch.fx.Node):
+                return to_real_tensor(e.meta["val"])
+            if isinstance(e, torch.Tensor):
+                return _tensor(e.size(), e.dtype, e.device)
```
eellison: do we need to deal with symints here /size hints ?

torch/_inductor/comm_analysis.py
```diff
@@ -533,6 +526,7 @@ def estimate_fx_collective_size(fx_node: torch.fx.Node) -> int:
 def estimate_nccl_collective_runtime_from_fx_node(
     fx_node: torch.fx.Node,
     override_size: Optional[int] = None,
+    use_nccl_estimator: bool = False,
```
eellison: should we turn this on ?

torch/_inductor/comm_analysis.py
```diff
@@ -561,13 +555,52 @@ def estimate_nccl_collective_runtime_from_fx_node(
...
     assert opt_args_kwargs is not None
-    _, kwargs = opt_args_kwargs
+    args, kwargs = opt_args_kwargs
 
     group_name = kwargs["group_name"]
     group_size = _get_group_size_by_name(group_name)
     assert isinstance(fx_node.target, torch._ops.OpOverload)
     coll = get_collective_type_from_kernel_name(fx_node.target.name())
 
+    def _nccl_estimate() -> Optional[float]:
```
eellison, line 398: nit: any reason to not factor this out ?

torch/_inductor/comm_analysis.py
```diff
@@ -561,13 +555,52 @@ def estimate_nccl_collective_runtime_from_fx_node(
...
+        # TODO: Refactor with estimate_nccl_collective_runtime_nccl_estimator
+        import torch.utils._pytree as pytree
+        from torch.distributed.distributed_c10d import _resolve_process_group
+
+        flat_args, flat_args_pytree_spec = pytree.tree_flatten((args, kwargs))
+
+        def _tensor(size, dtype, device) -> torch.Tensor:  # type: ignore[no-untyped-def]
+            return torch.empty(size, dtype=dtype, device=device)
```
eellison: if `override_size` is set, change this to [override_size]

eellison (approved):
Looks good, a few comments

torch/_inductor/comm_analysis.py
```diff
@@ -561,13 +555,52 @@ def estimate_nccl_collective_runtime_from_fx_node(
...
-    _, kwargs = opt_args_kwargs
+    args, kwargs = opt_args_kwargs
 
     group_name = kwargs["group_name"]
     group_size = _get_group_size_by_name(group_name)
     assert isinstance(fx_node.target, torch._ops.OpOverload)
     coll = get_collective_type_from_kernel_name(fx_node.target.name())
 
+    def _nccl_estimate() -> Optional[float]:
+        # TODO: Refactor with estimate_nccl_collective_runtime_nccl_estimator
```
eellison, line 399: what is the difference with `estimate_nccl_collective_runtime_nccl_estimator` ? why cant we reuse ?

torch/_inductor/comm_analysis.py
eellison, line 398: if this has overhead - which i assume it does - should we lru cche around the estimator ?

torch/_inductor/comm_analysis.py
IvanKobzarev: estimate_nccl_collective_runtime_nccl_estimator is for SchedulerNode

torch/_inductor/comm_analysis.py
IvanKobzarev: This function will be called for non-Tensor, non fx.Node args too. In this case it can return any constant arg type, 'str', int.

torch/_inductor/comm_analysis.py
IvanKobzarev: Yes, for sym sizes we should use size_hint in other case we will have recompilations.

IvanKobzarev: Yes, I will add size_hint to avoid recompilations.

torch/_inductor/comm_analysis.py
IvanKobzarev: Mostly just extracted it to simplify the logic to fallback on our runtime if nccl estimate fails.

PR #166533, opened 2025-10-29
Type torch/_dynamo/variables/[sdpa/script_object/optimizer].py

Lucaskabela:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #166535
* #166534
* __->__ #166533
* #166532

PR #166532, opened 2025-10-29
Type torch/_dynamo/variables/torch_function.py

Lucaskabela:
Stack from [ghstack](https://github.com/ezyang/ghstack) (oldest at bottom):
* #166535
* #166534
* #166533
* __->__ #166532

PR #166534, opened 2025-10-29
Type torch/_dynamo/variables/[constant/distributed/iter].py

Lucaskabela:
* #166535
* __->__ #166534
* #166533
* #166532

### DIFF
--- a/torch/_dynamo/variables/constant.py
+++ b/torch/_dynamo/variables/constant.py
@@ -1,5 +1,3 @@
-# mypy: ignore-errors
-
 """
 Constant and enum variable tracking in Dynamo.
 
@@ -8,8 +6,9 @@ values during compilation, ensuring proper handling of Python literals and
 maintaining type safety through the compilation process.
 """
 
+import enum
 import operator
-from typing import TYPE_CHECKING
+from typing import Any, Optional, TYPE_CHECKING, Union
 
 import torch
 from torch._dynamo.source import AttrSource, GetItemSource
@@ -40,7 +39,7 @@ class ConstantVariable(VariableTracker):
     """
 
     @staticmethod
-    def create(value, **kwargs) -> VariableTracker:
+    def create(value: Any, **kwargs: Any) -> VariableTracker:
         """
         Create a `ConstantVariable` based on the given value, and supports
         automatic routing for collection types like `tuple` (in which case we'd
@@ -76,7 +75,7 @@ class ConstantVariable(VariableTracker):
 
         return ConstantVariable(value, **kwargs)
 
-    def __init__(self, value, **kwargs) -> None:
+    def __init__(self, value: Any, **kwargs: Any) -> None:
         super().__init__(**kwargs)
         assert ConstantVariable.is_base_literal(value), f"""
 Cannot construct `ConstantVariable` for value of type {type(value)}.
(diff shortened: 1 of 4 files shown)


torch/_dynamo/variables/distributed.py
```diff
@@ -67,24 +66,24 @@ def __init__(self, value, **kwargs) -> None:
...
 
     @staticmethod
-    def is_available():
+    def is_available() -> bool:
         # check if the distributed package is available or not
         return torch.distributed.is_available()
 
 
-def is_from_local(value):
+def is_from_local(value: Any) -> bool:
```
Skylion007, line 78: Input should be object not Any, return type should probably be a TypeGuard or that function.

torch/_dynamo/variables/distributed.py
```diff
@@ -136,7 +137,7 @@ def var_getattr(self, tx: "InstructionTranslator", name: str) -> VariableTracker
 
 class PlacementClassVariable(DistributedVariable):
     @staticmethod
-    def is_placement_type(value):
+    def is_placement_type(value: Any) -> bool:
```
Skylion007, line 140: Most of these is method should take in object not Any. Subtle difference for things that should never be put into these args like Never,NoReturn or any function that always throws.

torch/_dynamo/variables/distributed.py
```diff
@@ -256,7 +257,7 @@ def reconstruct(self, codegen):
 
 class DeviceMeshVariable(DistributedVariable):
     @staticmethod
-    def is_device_mesh(value):
+    def is_device_mesh(value: Any) -> bool:
```
Skylion007, line 260: I think this can be TypeGuard, use TYPE_CHECKING block to resolve import loop if necessary.

torch/_dynamo/variables/distributed.py
```diff
@@ -361,7 +362,7 @@ def var_getattr(self, tx: "InstructionTranslator", name):
         return super().var_getattr(tx, name)
 
     @staticmethod
-    def is_process_group(value):
+    def is_process_group(value: Any) -> bool:
```
Skylion007, line 365: Same could be TypeGuard on the union if the IsType check

torch/_dynamo/variables/iter.py
```diff
@@ -362,16 +370,18 @@ def __init__(
         self.index = 0
         self.strict = strict
 
-    def python_type(self):
+    def python_type(self) -> type:
```
Skylion007, line 373: This could be type[zip] as return type.

Skylion007 (approved):
Some nits

torch/_dynamo/variables/constant.py
```diff
@@ -92,48 +91,52 @@ def __init__(self, value, **kwargs) -> None:
...
+        self, tx: "InstructionTranslator", arg: VariableTracker
+    ) -> VariableTracker:
         return ConstantVariable.create(
             self.value[arg.as_python_constant()],
         )
 
     @staticmethod
-    def is_base_literal(obj):
+    def is_base_literal(obj: object) -> bool:
         return type(obj) in common_constant_types
```
Skylion007, line 123: Can be a n isinstance check, no need to call type explicitly unless you are overriding the possibility of someone overwriting the __'isntance_check__' magic method?

torch/_dynamo/variables/constant.py
```diff
@@ -92,48 +91,52 @@ def __init__(self, value, **kwargs) -> None:
...
 
     def __repr__(self) -> str:
         return f"ConstantVariable({type(self.value).__name__}: {repr(self.value)})"
 
-    def as_python_constant(self):
+    def as_python_constant(self) -> Any:
         return self.value
 
-    def is_python_constant(self):
+    def is_python_constant(self) -> bool:
```
Skylion007, line 103: Slight nit, if you don't think this will ever subclasses, you can have it return type Literal[True]

Skylion007 (approved):

torch/_dynamo/variables/distributed.py
Lucaskabela: Only issue with TypeGuard in these cases is the imports needed may not be available and so this could cause issues (i.e the torch.distributed namespace is not always populated)

So I think it is worth keeping these as bool to simplify the code as otherwise we will need lots of import handling at the top level

PR #166535, opened 2025-10-29
Type torch/_dynamo/variables/base.py

Lucaskabela:
* __->__ #166535
* #166534
* #166533
* #166532

cc @voznesenskym @penguinwu @EikanWang @jgong5 @Guobing-Chen @XiaobingSuper @zhuhaozhe @blzheng @wenzhe-nrv @jiayisunx @chenyang78 @kadeng @chauhang @amjames

### DIFF
--- a/torch/_dynamo/source.py
+++ b/torch/_dynamo/source.py
@@ -111,9 +111,9 @@ def is_constant_source(source: Source) -> bool:
     return False
 
 
-def _get_source_debug_name(source: Source) -> str:
+def _get_source_debug_name(source: Optional[Source]) -> str:
     try:
-        return source.name()
+        return source.name()  # type: ignore[union-attr]
     except NotImplementedError:
         return "<unknown source>"
 
--- a/torch/_dynamo/variables/base.py
+++ b/torch/_dynamo/variables/base.py
@@ -1,5 +1,3 @@
-# mypy: ignore-errors
-
 """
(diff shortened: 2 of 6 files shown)


torch/_dynamo/source.py
```diff
@@ -111,9 +111,9 @@ def is_constant_source(source: Source) -> bool:
     return False
 
 
-def _get_source_debug_name(source: Source) -> str:
+def _get_source_debug_name(source: Optional[Source]) -> str:
     try:
-        return source.name()
+        return source.name()  # type: ignore[union-attr]
```
Skylion007, line 116: If none is valid, should have handling for it. Add early return.

torch/_dynamo/variables/base.py
```diff
@@ -636,12 +640,12 @@ def __init__(
                 assert source is not None
 
 
-def raise_type_error_exc(tx: "InstructionTranslator", msg_str: str) -> None:
+def raise_type_error_exc(tx: Any, msg_str: str) -> None:
```
Skylion007, line 643: Doesn't this always return an exception? Should be NoReturn or Never return type from typing/typing_extensions. That tells mypy/pyrefly it always throws.

This can also be applied to method if none of the subclasses will return a value, but always raise an exception.

Lucaskabela:
Seems this stack has some funky stuff - will be abandoning and moving to https://github.com/pytorch/pytorch/pull/166564 where I have applied suggestions

PR #166536, opened 2025-10-29
Add API to annotate disjoint backward and handle in AC

soulitzer:
* #167251
* #167006
* #166610
* __->__ #166536

This adds zero-bubble / DualPipeV support for (S)AC

Before:
- AC will always retrigger recompute upon every distinct backward.

After:
- Any checkpointed regions encountered by backward under the same instance of this context manager will only trigger recompute at most once, even if there are multiple calls to backward.
- Backward calls under the same instance of this context manager must execute over non-overlapping regions of the backward graph even if retain_graph=True.


torch/autograd/graph.py
```diff
@@ -254,6 +255,39 @@ def increment_version(tensor: Union[torch.Tensor, Iterable[torch.Tensor]]) -> No
     torch._C._increment_version(tensor)
 
 
+class GraphExecutionGroup:
```
soulitzer: Name bikesheddable, could be something more related to disjointness

torch/utils/checkpoint.py
```diff
@@ -1587,6 +1592,40 @@ def recompute_fn(*inputs):
 
     return
 
+
+class GraphExecGroup:
+    """Any checkpointed regions encountered by backward under the same instance
+    of this context manager will only trigger recompute at most once, even if
```
albanD:
```suggestion
    of this context manager will trigger recompute at most once, even if
```

torch/utils/checkpoint.py
```diff
@@ -1587,6 +1592,40 @@ def recompute_fn(*inputs):
...
+    there are multiple calls to backward.
+
+    Backward calls under the same instance of this context manager must execute
+    over non-overlapping regions of the backward graph even if retain_graph=True.
```
albanD, line 1616: Why? If they don't overlap, then they would never trigger the recompute more than once, hence according to the paragraph above, this context manager is a noop? :D

torch/utils/checkpoint.py
```diff
@@ -1587,6 +1592,40 @@ def recompute_fn(*inputs):
 
     return
 
+
+class GraphExecGroup:
```
albanD, line 1610: nit: for the name, feel free to ignore, but why are we talking of "graph" here?
Also we should mention that this is only relevant for non-reentrant checkpoint

test/test_autograd.py
```diff
@@ -7362,6 +7362,61 @@ def test_checkpoint_sequential_warns_if_use_reentrant_not_passed_explcitly(self)
...
+
+        # Test the not actually disjoint case (using retain_graph=True since
+        # otherwise autograd itself will catch this)
+        def fn(x):
+            return x.sin().cos()
+
+        x = torch.randn(3, 3, requires_grad=True)
+        out = checkpoint(fn, x, use_reentrant=False)
+        group = torch.utils.checkpoint.GraphExecGroup()
+        with group:
```
albanD: Note that usually in python you would inline the context manager creation there.

torch/utils/checkpoint.py
```diff
@@ -1587,6 +1592,40 @@ def recompute_fn(*inputs):
...
+
+    Backward calls under the same instance of this context manager must execute
+    over non-overlapping regions of the backward graph even if retain_graph=True.
+    """
+    _current_group: Optional[int] = None
+    _counter: int = 0
+
+    def __init__(self) -> None:
+        self._id: int = GraphExecGroup._counter
+        GraphExecGroup._counter += 1
```
albanD: is there a thread safe version of this?

torch/utils/checkpoint.py
```diff
@@ -1587,6 +1592,40 @@ def recompute_fn(*inputs):
...
+    of this context manager will only trigger recompute at most once, even if
+    there are multiple calls to backward.
+
+    Backward calls under the same instance of this context manager must execute
+    over non-overlapping regions of the backward graph even if retain_graph=True.
+    """
+    _current_group: Optional[int] = None
+    _counter: int = 0
+
+    def __init__(self) -> None:
```
albanD: Since we never expect this to be set globally, you most likely can make this a regular context manager with something like https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager that will make your life a lot easier

torch/utils/checkpoint.py
soulitzer: I think there's two concepts of overlap here:
- two backward passes resulting in unpacking the same saved activation
- two backward passes unpacking two different saved activation, but those two saved activation were saved while under the same checkpoint region.

In this description I mean that we don't allow the first kind of overlap! 
This context manager is needed due to the second kind of overlap, which will cause recompute can be triggered twice.

test/test_autograd.py
soulitzer: Yup, although I think I also wanted to emphasize that the incrementing is done in the construction, and I'm able to use it in this manner (but I forgot to do so in this test!)
```
group = ...
with group:
    ...
with group:
    ...
```

torch/utils/checkpoint.py
soulitzer: Not super opinionated on this, but graph exec group seemed analogous to graph task id, and it's a group of (disjoint) backward graph execution.

Good catch, let me mention non-reentrant in the description.

torch/utils/checkpoint.py
soulitzer: IIUC if we use that, each instance of that one can only be entered/exited only one time?
I might prefer to have the ability to create the object representing the group once, and then wrap any .grads individually. This would be useful when I'm implementing pipelining schedules with split dW/dI because the dW and dI are called kind of independently and it may be hard to find a place to have a single context region.

torch/utils/checkpoint.py
soulitzer: Getting rid of the counter entirely. Now we're keyed on the GraphExecGroup object itself. 
Also made the state thread local.

soulitzer: Tests are failing because we probably need to add it to the tls state to so that autograd can ferry it to backward.

soulitzer: Updated now.

torch/utils/checkpoint.py
soulitzer: (non-reentrant has been updated in the description)

c10/core/AutogradState.h
```diff
@@ -61,13 +68,18 @@ struct C10_API AutogradState {
...
+  }
+
  private:
   bool grad_mode_ : 1;
   bool inference_mode_ : 1;
   bool fw_grad_mode_ : 1;
   bool multithreading_enabled_ : 1;
   // NOLINTNEXTLINE(cppcoreguidelines-use-default-member-init)
   bool view_replay_enabled_ : 1;
+  std::optional<SafePyObject> graph_exec_group_;
```
albanD: Move above the boolean field to prevent extra padding

albanD (approved):
Small thin, but otherwise good to land!

PR #166537, opened 2025-10-29
Replace decltype(auto) with auto

cyyever:
This PR replaces `decltype(auto)` with `auto` for C++ return type deduction and simplifies some templates.  

cc @H-Huang @awgu @wanchaol @fegin @fduwjj @wz337 @wconstab @d4l3k @pragupta @msaroufim @dcci @EikanWang @jgong5 @wenzhe-nrv @sanchitintel

### DIFF
--- a/c10/test/util/TypeList_test.cpp
+++ b/c10/test/util/TypeList_test.cpp
@@ -239,7 +239,7 @@ struct Class2 {
 
 struct mapper_call_func {
   template <class T>
-  decltype(auto) operator()(T) {
+  auto operator()(T) {
     return T::type::func();
   }
 };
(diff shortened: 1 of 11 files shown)


Skylion007 (approved):
When did this stop being necessary, C++17?

aten/src/ATen/core/boxing/impl/WrapFunctionIntoRuntimeFunctor.h
```diff
@@ -35,7 +34,6 @@ class WrapFunctionIntoRuntimeFunctor_<
 template <class FuncType>
 using WrapFunctionIntoRuntimeFunctor = detail::WrapFunctionIntoRuntimeFunctor_<
     FuncType,
-    typename guts::infer_function_traits_t<FuncType>::return_type,
```
Skylion007, line 38: Removing this seems unrelated?

cyyever: It can be simplified as a result of refactoring.

cyyever: This is reverted to pass CI.

PR #166540, opened 2025-10-29
LocalTensor for random_ops tests

dolpm:
adds support for randomness in localtensor. tl;dr it needs to be able to handle RNG the same way (i.e., rng tracking/syncing across shards, user-defined seeds, user-defined generators, etc.

we extend the existing OffsetBasedRNGTracker to play nicely with localtensor's setup, creating a few small subclasses and patching the core RNG logic to manage the per-rank seeds and offsets correctly.

i still haven't done the per-rank generator support (since the existing tests imply a globally-seeded generator), but that it something that should be done.

### DIFF
--- a/test/distributed/tensor/test_random_ops.py
+++ b/test/distributed/tensor/test_random_ops.py
@@ -6,8 +6,8 @@ import itertools
 import torch
 import torch.distributed._functional_collectives as funcol
 import torch.distributed.tensor._random as random
+from torch.distributed._local_tensor import LocalTensor, maybe_run_for_local_tensor
 from torch.distributed.device_mesh import init_device_mesh
-from torch.distributed.distributed_c10d import broadcast_object_list
 from torch.distributed.fsdp import fully_shard
 from torch.distributed.tensor import (
     DeviceMesh,
@@ -26,6 +26,7 @@ from torch.distributed.tensor.debug import CommDebugMode
 from torch.distributed.tensor.parallel import ColwiseParallel, parallelize_module
 from torch.testing._internal.common_utils import run_tests
 from torch.testing._internal.distributed._tensor.common_dtensor import (
+    create_local_tensor_test_class,
     DTensorTestBase,
     skip_if_lt_x_gpu,
     skip_unless_torch_gpu,
@@ -34,9 +35,12 @@ from torch.testing._internal.distributed._tensor.common_dtensor import (
 from torch.utils._typing_utils import not_none
 
 
-def get_generator_seed_for_device_type(device_type: str) -> int:
-    device_module = torch.get_device_module(device_type)
-    return device_module.get_rng_state()[:8].view(torch.int64).item()
+def get_generator_seed_for_device_type(device_type: str):
+    from torch.distributed._local_tensor import (
+        get_generator_seed_for_device_type as _get_seed,
+    )
+
+    return _get_seed(device_type)
 
 
 class DistTensorRandomInitTest(DTensorTestBase):
@@ -134,9 +138,6 @@ class DistTensorRandomInitTest(DTensorTestBase):
             torch.empty(*size, device="meta"), device_mesh, [Replicate()]
         )
 
-        # the tensor slice on the current rank
-        self_slice = slice(1024 * self.rank, 1024 * self.rank + 1024)
-
         # Test 1: enable the distribute region for RNG (by default)
         self.assertTrue(meta_dtensor.is_meta)
         # Tensor meta init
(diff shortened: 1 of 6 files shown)


torch/distributed/tensor/_random.py
```diff
@@ -239,6 +242,20 @@ def _set_device_state(self, state: torch.Tensor):
...
     ):
+        from torch.distributed._local_tensor import local_tensor_mode
+
+        lm = local_tensor_mode()
+
+        if lm is not None and not lm._disable:
+            from torch.distributed._local_tensor import _LocalOffsetBasedRNGTracker
+
+            local_tracker = _LocalOffsetBasedRNGTracker(self._device.type)
+            local_tracker.distribute_region_enabled = self.distribute_region_enabled
```
dzmitry-huba: Can you please refactor this into a function in local tensor that if local tensor mode is enabled returns context from local_tracker._distribute_region and otherwise None. This way we put all the details back in the local tensor module and call here a single method, something like this:

if local_tracker_context := maybe_enable_local_tracker(self.distribute_region_enabled):
   with local_tracker_context:
       yield
   return

test/distributed/tensor/test_random_ops.py
```diff
@@ -34,9 +36,34 @@
 from torch.utils._typing_utils import not_none
 
 
-def get_generator_seed_for_device_type(device_type: str) -> int:
-    device_module = torch.get_device_module(device_type)
-    return device_module.get_rng_state()[:8].view(torch.int64).item()
+def get_generator_seed_for_device_type(device_type: str):
```
dzmitry-huba, line 38: I think we should extract this functionality into local tensor module

torch/distributed/_local_tensor/__init__.py
```diff
@@ -256,21 +266,31 @@ def _for_each_rank_run_func(
...
     ]
 
-    # NB: Before invoking an op we are collecting rng states from CPU and
-    # CUDA devices such that we can reset to the same before invoking op
-    # for each rank. This is not very efficient and will likely be revisited
-    # to support per rank rng state.
-    rng_state = _get_rng_state()
+    lm: LocalTensorMode | None = local_tensor_mode()
+    use_per_rank_rng = lm is not None and len(lm._per_rank_rng_states) > 0
+
```
dzmitry-huba, line 271: Can you remind when this (_for_each_rank_run_func) function is called and local tensor mode is None? I thought there are two cases when it is called: torch_dispatch and @maybe_run_for_local_tensor and in both cases it is present.

torch/distributed/_local_tensor/__init__.py
```diff
@@ -437,6 +457,236 @@ def wrap_int(self, num: int) -> "LocalIntNode | ConstantIntNode":
...
+    def get_rng_state(self):
+        """
+        Get RNG state, automatically returning LocalTensor in LocalTensor mode.
+
+        Returns:
+            LocalTensor in LocalTensor mode, regular Tensor otherwise
+        """
+        lm = local_tensor_mode()
+        if lm is None or lm._disable:
```
dzmitry-huba: This pattern or its opposite is used quite frequently that it warrants a separate convenience function.

torch/distributed/_local_tensor/__init__.py
```diff
@@ -437,6 +457,236 @@ def wrap_int(self, num: int) -> "LocalIntNode | ConstantIntNode":
...
+
+        original_state = _get_rng_state()
+        per_rank_states = {}
+
+        try:
+            for rank in lm.ranks:
+                if rank in lm._per_rank_rng_states:
+                    _set_rng_state(*lm._per_rank_rng_states[rank])
+
+                per_rank_states[rank] = self._device_handle.get_rng_state()
```
dzmitry-huba, line 508: Do you mind adding comment here explaining why we need to first set rng state and then immediately get it

torch/distributed/_local_tensor/__init__.py
```diff
@@ -437,6 +457,236 @@ def wrap_int(self, num: int) -> "LocalIntNode | ConstantIntNode":
...
+            state: Regular Tensor or LocalTensor with per-rank states
+        """
+        lm = local_tensor_mode()
+
+        if isinstance(state, LocalTensor):
+            assert lm is not None and not lm._disable
+
+            for rank, rank_state in state._local_tensors.items():
+                self._device_handle.set_rng_state(rank_state.to("cpu"))
+                lm._per_rank_rng_states[rank] = _get_rng_state()
```
dzmitry-huba, line 534: Same as above a comment would be helpful to understand the set followed by get pattern.

torch/distributed/_local_tensor/__init__.py
```diff
@@ -437,6 +457,236 @@ def wrap_int(self, num: int) -> "LocalIntNode | ConstantIntNode":
...
+    def _set_pre_op_offset(self, state, spec) -> None:
+        """Compute and set per-rank offsets before the random operation."""
+        from torch.distributed.tensor._ops.utils import prod
+        from torch.distributed.tensor._utils import (
+            _compute_local_shape_and_global_offset,
+        )
+        from torch.distributed.tensor.placement_types import Shard
+
+        lm = local_tensor_mode()
+        assert lm is not None and not lm._disable
```
dzmitry-huba: This is the "opposite" pattern I mentioned in the comment above

torch/distributed/_local_tensor/__init__.py
```diff
@@ -949,6 +1220,138 @@ def _unpatch_device_mesh(self) -> None:
...
+
+
+class _LocalRandom:
+    """
+    Holds implementations of random functionality that must be patched while running
+    under LocalTensorMode.
+    """
+
+    @staticmethod
+    def dtensor_manual_seed(seed: int, device_mesh: DeviceMesh) -> None:
```
dzmitry-huba: In torch_manual_seed we assumed that the seed can be a SymInt with LocalInt node representing seeds for every rank. Why are we not assuming the same for the dtensor_manual_seed?

torch/distributed/_local_tensor/__init__.py
```diff
@@ -1183,3 +1586,113 @@ def current() -> "LocalRunnerMode":
...
+                rank_offset = offset.node._local_ints[rank]
+                offset_tensor = torch.tensor(
+                    [rank_offset], dtype=torch.uint64, device="cpu"
+                ).view(torch.uint8)
+                state[8:] = offset_tensor
+        else:
+            offset_int = int(offset) if isinstance(offset, SymInt) else offset
+            offset_tensor = torch.tensor(
+                [offset_int], dtype=torch.uint64, device="cpu"
+            ).view(torch.uint8)
```
dzmitry-huba: I'm curious if _LocalPhiloxState can derive from _PhiloxState and when implementing the methods like this we replace the "else: " branch with something like super().<method>(<args>) to avoid having to repeat the original implementation.

dolpm: yeah so we will run into some circular dependency stuff if we have it inherit from _PhiloxState -- what we can do is just construct PhiloxState per-rank in each class method. it's not efficient nor pretty, but this doesn't really need to be and at least we don't have to do all the bitpacking stuff all over again.

torch/distributed/_local_tensor/__init__.py
dolpm: can actually remove this now entirely.

torch/distributed/_local_tensor/__init__.py
dolpm: these checks are so the linter doesn't act up...

dzmitry-huba (approved):

PR #166541, opened 2025-10-29
address DDE in matmul decomp

laithsakka:
* __->__ #166541

Address https://github.com/pytorch/pytorch/issues/165081


aten/src/ATen/native/TensorShape.cpp
```diff
@@ -1,5 +1,6 @@
 #include <ATen/core/ATen_fwd.h>
 #include <c10/core/ScalarType.h>
+#include <c10/core/SymInt.h>
 #define TORCH_ASSERT_ONLY_METHOD_OPERATORS
```
laithsakka, line 4: NO why thise changes are here wth

mlazos (approved):
lgtm, one question when you `guard_or_false` are we just optimizing the case when it equals 1? I'm guessing we can do something when the value is 1, and recompile if it's not 1 later. But in the other direction we'll just use the unpsecialized code for all cases.

laithsakka:
> lgtm, one question when you `guard_or_false` are we just optimizing the case when it equals 1? I'm guessing we can do something when the value is 1, and recompile if it's not 1 later. But in the other direction we'll just use the unpsecialized code for all cases.

yeh when we know one of them is for sure 1 we take the opt path , not that we already know they are not both 1s.
now in the case that we do not know we take the general path and that would be our unbacked semantics for this.

PR #166544, opened 2025-10-29
[ROCm][CI] Create periodic-rocm-mi200.yml

amdfaa:
* We are separating out the rocm jobs of the periodic workflow
* We are introducing a new label `ciflow/periodic-rocm-mi200` to allow us to run distributed tests only on ROCm runners, without triggering many other jobs on the `periodic.yml` workflow (via `ciflow/periodic`)
* This new workflow will also be triggered via the `ciflow/periodic`, thus maintaining the old status quo.
* We are reverting to the `linux.rocm.gpu.4` label since it targets a lot more CI nodes at this point than the K8s/ARC-based `linux.rocm.gpu.mi250.4` label, as that is still having some network/scaling issues.


.github/workflows/periodic-rocm-mi200.yml
jithunnair-amd, line 1: This needs to have a .yml extension, and name needs to be `periodic-rocm-mi200.yml` to be consistent with https://github.com/pytorch/pytorch/blob/main/.github/workflows/periodic-rocm-mi300.yml?

amdfaa: Yeah, I was wondering why the syntax highlighting was off.

.github/workflows/periodic-mi200
```diff
@@ -0,0 +1,83 @@
...
+    # We have several schedules so jobs can check github.event.schedule to activate only for a fraction of the runs.
+    # Also run less frequently on weekends.
+    - cron: 45 0,8,16 * * 1-5
+    - cron: 45 4 * * 0,6
+    - cron: 45 4,12,20 * * 1-5
+    - cron: 45 12 * * 0,6
+    - cron: 29 8 * * *  # about 1:29am PDT, for mem leak check and rerun disabled tests
+  push:
+    tags:
+      - ciflow/periodic-mi200/*
```
jithunnair-amd: should be `ciflow/periodic-rocm-mi200` to be consistent with https://github.com/pytorch/pytorch/blob/main/.github/workflows/periodic-rocm-mi300.yml

jeffdaily (approved):

.github/workflows/periodic-rocm-mi200.yml
```diff
@@ -0,0 +1,83 @@
...
+    # We have several schedules so jobs can check github.event.schedule to activate only for a fraction of the runs.
+    # Also run less frequently on weekends.
+    - cron: 45 0,8,16 * * 1-5
+    - cron: 45 4 * * 0,6
+    - cron: 45 4,12,20 * * 1-5
+    - cron: 45 12 * * 0,6
+    - cron: 29 8 * * *  # about 1:29am PDT, for mem leak check and rerun disabled tests
+  push:
+    tags:
+      - ciflow/periodic-rocm-mi200/*
```
pragupta, line 15: @jeffdaily  / @amdfaa / @jithunnair-amd  -- If we completely remove rocm from periodic job, then we are removing default rocm testing on distributed jobs, which will lead to many issues coming our way down the road. Also, perhaps others won't recognize they need to add a separate label. So, if we are not concerned about capacity, I'd like if we can add `ciflow/periodic` here and keep mi200 testing on periodic jobs.

jeffdaily: Heard loud and clear.  We should consider moving mi300 into ciflow/periodic.  But let's wait and see how many failures slip into main using a land-but-revert-on-error strategy.  IMHO having mi200 as part of ciflow/periodic hasn't seemed to help us prevent merges that break rocm.

amdfaa: The workflow has been updated to add the periodic tag. They can now be tested separately or together.

